PR #6533, opened 2015-01-05
[TIMOB-18173] iOS: Fix adding date property for Ti.Contacts.createPerson

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-18173


vishalduggal:
CR + Fr ok. __APPROVED__

PR #6535, opened 2015-01-05
[TIMOB-17404] iOS: Fixes Error when focusing TextField in HeaderView of empty TableView

jonalter:
Fixes [TIMOB-17404](https://jira.appcelerator.org/browse/TIMOB-17404)
Test case in ticket.


vishalduggal:
FR failed. Additional test case added to ticket

PR #6537, opened 2015-01-06
[TIMOB-16270] Fix startIntentSenderForResult for Android < Jelly Bean (Level 16)

bogdansmc:
JIRA: https://jira.appcelerator.org/browse/TIMOB-16270

Function was added in API level 16 and doesn't work for pre Jelly Bean android versions.
Version added in API level 5 used for previous versions.
http://developer.android.com/reference/android/app/Activity.html#startIntentSenderForResult(android.content.IntentSender,%20int,%20android.content.Intent,%20int,%20int,%20int)


hansemannn:
Approved!

PR #6538, opened 2015-01-06
[TIMOB-17561] Android: Fixed - When adding swipe event to a tabGroup, th...

ashcoding:
...e inside tableview's inside won't scroll nice (parity issue)

Jira: https://jira.appcelerator.org/browse/TIMOB-17561

In the SimpleOnGestureListener, if onFling returns `True`, it takes it as that the event is over and will stop the flinging action of the view as the event is consumed. To allow it to continue to scroll with inertia, it needs to always return false.

```
// create table view 1
    title:'Tab 1',
    backgroundColor:'#000'
});
var tab1 = Titanium.UI.createTab({  
    icon:'KS_nav_views.png',
    title:'Tab 1',
    window:win1
});

var rows1 = [];
for(var i = 0; i < 50; i++) {
    rows1.push({ title: 'Row 1:' + i});
}

var table1 = Titanium.UI.createTableView({
    data: rows1,
    width: 'auto',
    height: 'auto'
});

win1.add(table1);

// create table view 2
var win2 = Titanium.UI.createWindow({  
    title:'Tab 2',
    backgroundColor:'#000'
});
var tab2 = Titanium.UI.createTab({  
    icon:'KS_nav_ui.png',
    title:'Tab 2',
    window:win2
});

var rows2 = [];
for(var i = 0; i < 50; i++) {
    rows2.push({ title: 'Row 2:' + i});
}

var table2 = Titanium.UI.createTableView({
    data: rows2,
    width: 'auto',
    height: 'auto'
});

win2.add(table2);

//  create tabgroup and tabs
var tabGroup = Ti.UI.createTabGroup();
tabGroup.addTab(tab1);  
tabGroup.addTab(tab2);  

// Event handler to change tab on left/right swipe
// This is where the problem comes in.
// Comment out the 'swipe' listener and normal table view scroll behavior will be back

tabGroup.addEventListener('swipe', function(e) {
    var activeTabIndex = tabGroup.tabs.indexOf(tabGroup.activeTab);

    if(e.direction === 'right' && activeTabIndex > 0) {
        tabGroup.setActiveTab(activeTabIndex - 1);
    } else if(e.direction === 'left' && activeTabIndex < tabGroup.tabs.length - 1) {
        tabGroup.setActiveTab(activeTabIndex + 1);
    }
    // Is there something missing in this event handler to have native behavior when the swipe
    // direction is not left or right?
});

// open tab group
tabGroup.open();
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
@@ -1436,7 +1436,7 @@ public abstract class TiUIView
 					} else {
 						data.put(TiC.EVENT_PROPERTY_DIRECTION, velocityY > 0 ? "down" : "up");
 					}
-					return fireEvent(TiC.EVENT_SWIPE, data);
+					fireEvent(TiC.EVENT_SWIPE, data);
 				}
 				return false;
 			}


hieupham007:
This change affects all apps, not just apps with tableview, which will cause behavioral changes to app that expect the current touch architecture. I would advise not to merge this to master unless we are planning to redefine Android gesture architecture.

ingo:
I agree. I would not want to change this behavior without careful consideration.

ashcoding:
Okay. Understood. Shall I close this PR?

ingo:
If you can fix it without this behavior change, is that a small change to this PR, or a new approach?

ashcoding:
Probably a whole new approach. Thus, I think this PR can be rejected.

hieupham007:
Closing PR

PR #6539, opened 2015-01-06
Fixes switch animation on when programmatically setting value

FokkeZB:
JIRA: https://jira.appcelerator.org/browse/TIMOB-2908


hansemannn:
@FokkeZB Is this still happening? Would be happy to get this in.

FokkeZB:
Still true, but as you pointed out on Flow you wouldn't always want it not to animate so it would be better if we'd override `setValue(val)` to allow:

```
mySwitch.setValue(true, {
  animated: false
});
```

skypanther:
I commented on the ticket. Simply disabling the animation doesn't fully resolve the current problem with the switch.

hansemannn:
Closed in favor of #7456

PR #6540, opened 2015-01-06
[TIMOB-18317] Use view bounds instead of center

vishalduggal:
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-18317)
Regression path->KS


jonalter:
FT **pass**

PR #6541, opened 2015-01-06
[TIMOB-18317] (3_5_X) Use view bounds instead of center

vishalduggal:
Cherry Pick PR #6540 into 3_5_X
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-18317)
Regression path->KS


jonalter:
FT **pass**

PR #6544, opened 2015-01-07
[TIMOB-18193] iOS: adding flag to xcodebuild to force building all architectures in ARCHS

jonalter:
Test case in ticket [TIMOB-18193](https://jira.appcelerator.org/browse/TIMOB-18193)


cheekiatng:
FR passed. Merged.

PR #6547, opened 2015-01-07
[TIMOB-18334] Fixed debugger termination

pec1985:
https://jira.appcelerator.org/browse/TIMOB-18334
https://jira.appcelerator.org/browse/TIMOB-18335


pinnamur:
The new library has addressed removing execution pointer after the launch is terminated and removed. However, when the breakpoint condition is hit, then the thread suspends at "Top-level Script", instead of app.js file.

Similarly, also noticed that, if a new breakpoint is added in a button click event handler (after the app is launched in debug mode), then the thread is not suspended at the new breakpoint.

pec1985:
Updated library with the fixes

cheekiatng:
FR passed. Merging.

PR #6548, opened 2015-01-08
[TIMOB-16946] static analyzer to ignore intentional warnings

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-16946


jonalter:
CR & FT **pass**

PR #6552, opened 2015-01-08
Timob 17357: Android: Support Material Theme

pingwang2011:
https://jira.appcelerator.org/browse/TIMOB-17357


android/modules/ui/AndroidManifest.xml
```diff
@@ -4,7 +4,7 @@
 	android:versionCode="1"
 	android:versionName="1.0">
 
-	<uses-sdk android:minSdkVersion="8" android:targetSdkVersion="11" />
+	<uses-sdk android:minSdkVersion="11" android:targetSdkVersion="21" />
```
mokesmokes: @pingwang2011 why is the minSdkVersion 11?

pingwang2011: Because addlayoutlistener is a method added from API 11

mokesmokes: So can we make that specific functionality dependent on API > 10? So all we would lose is the postlayout event for these widgets for API10, which few people probably need in any case? It's better than dropping API 10 altogether, in my opinion.Thanks!

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -77,47 +79,36 @@
...
-		}
+		super(proxy);
+		Log.d(TAG, "Creating a text field", Log.DEBUG_MODE);
 		
-		/** 
-		 * Check whether the called view is a text editor, in which case it would make sense to 
-		 * automatically display a soft input window for it.
-		 */
-		@Override
-		public boolean onCheckIsTextEditor () {
```
hieupham007: why is this removed?

hieupham007:
Please merge with master

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
mokesmokes: Because we're no longer extending EditText, thus we're not overriding any of its methods. Instead, we're instantiating an EditText from XML and calling methods on the instance. That's the only documented/official way for Appcompat to generate the properly styled widgets for all API versions.

mokesmokes:
@pingwang2011 @hieupham007 prior to merge can you please address the API 10 issue above? Thanks

ingo:
@mokesmokes We deprecated support for API 10 in 3.5.0 as the market share has dropped below 10% (7.8% as of last count), and will remove it in the next major version of the Titanium SDK.

mokesmokes:
@ingo OK, then perhaps set the minimum version at 15 since 11-14 are each below 0.1%? AFAIK all those versions were upgradeable to at least 15, with the vast majority running Jellybean by now.

pingwang2011:
Merged with master. Ready for review.

hieupham007:
Test case looks good, though KS looks different - posted screenshots of before and after in ticket.

PR #6555, opened 2015-01-08
[TIMOB-18345] 3.5.x back port iOS ListView - fix index out of bounds

pec1985:
https://jira.appcelerator.org/browse/TIMOB-18345


jonalter:
CR & FT **pass**

PR #6559, opened 2015-01-09
[TIMOB-18332] iOS: Fixes app crash when deleting last row of tableview

jonalter:
Cherry-pick of #6558 
Test case in ticket: [TIMOB-18332](https://jira.appcelerator.org/browse/TIMOB-18332)


vishalduggal:
CR + FR ok. __APPROVED__

PR #6563, opened 2015-01-12
[TIMOB-18365] iOS: TextField - Fixed value not updated when autocorrect ...

jonalter:
...is up and return is pressed

Test case in ticket: [TIMOB-18365](https://jira.appcelerator.org/browse/TIMOB-18365)

Cherry-pick of #6562


vishalduggal:
CR + FR ok. __APPROVED__

PR #6567, opened 2015-01-14
[TIMOB-18368] Fix missing path.join in _build.js

skypanther:
https://jira.appcelerator.org/browse/TIMOB-18368


cb1kenobi:
CR'd. Looks good!

PR #6571, opened 2015-01-15
[TIMOB-16909]: Fix anti-aliasing on views

hieupham007:
testing steps in JIRA

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
@@ -110,8 +110,10 @@ public class TiBorderWrapperView extends FrameLayout
 				float innerRadii[] = new float[8];
 				Arrays.fill(innerRadii, radius - padding);
 				innerPath.addRoundRect(innerRect, innerRadii, Direction.CW);
+				borderPath.addRoundRect(innerRect, innerRadii, Direction.CCW);
 			} else {
 				innerPath.addRect(innerRect, Direction.CW);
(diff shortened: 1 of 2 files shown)


ashcoding:
I tried this and it works nicely!

I have another scenario though which still has aliasing. :( 
File is in jira 'another_app.js' and resource image 'square.png'.

If I were to use:-
```
backgroundImage:'square.png',
```

There will still be aliasing.

android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
```diff
@@ -111,17 +125,27 @@ private void updateBorderPath()
...
+		setAlphaAndColor();
+		canvas.drawPath(borderPath, paint);
+	}
+
+	private void setAlphaAndColor()
+	{
 		paint.setColor(color);
 		if (alpha > -1) {
 			paint.setAlpha(alpha);
     	}
```
ashcoding: Minor thing: Curly bracket indentation is wrong.

hieupham007:
Yes, this fix doesn't fix backgroundDrawable (clipping will cause aliasing for drawables). There's really no way for fixing drawables unless we create bitmaps from canvas and use PorterDuff xfer modes. That method will cause memory fragmentation and is not really viable.

This PR will fix backgroundColor for most cases. The aliasing is still there, but a lot less worse than before.

ashcoding:
I came to a similar conclusion with respect to Drawables as what you mentioned. I guess there's no ideal way to solve that yet.

PR #6574, opened 2015-01-15
[TIMOB-18278] Update contentSize

vishalduggal:
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-18278)


jonalter:
CR & FT **pass**

PR #6582, opened 2015-01-17
[TIDOC-2089] Add parity report support to docgen.js

garymathews:
Add the ability to generate a parity report from the JSCA file.

Usage:

```
docgen.js -f parity
```


apidoc/lib/parity_generator.js
```diff
@@ -0,0 +1,89 @@
+if (!String.prototype.contains) {
```
ingo: Do we have copyright headers at the top of these? A file description?

ingo:
Approved. Merging.

PR #6583, opened 2015-01-19
[TIMOB-18338] Add command to append windows apis to yaml and jsca

garymathews:
This should append all windows APIs into the YAML documentation and JSCA file via the 'docgen -a' command.
- Fixed bug with incorrect ivi and windows parameters in the 'package.build' method


ingo:
Approved. Merging.

PR #6585, opened 2015-01-20
[TIMOB-18399] Windows: Projects can not be deleted if it's built once

skypanther:
https://jira.appcelerator.org/browse/TIMOB-18399

### DIFF
--- a/cli/commands/build.js
+++ b/cli/commands/build.js
@@ -257,9 +257,11 @@ exports.run = function (logger, config, cli, finished) {
 					logger.error(msg);
 				});
 				logger.log();
+				logger.log.end();
 				process.exit(1);
 			} else {
 				logger.info(__('Project built successfully in %s', delta.cyan) + '\n');
+				logger.log.end();
 			}
 
 			finished();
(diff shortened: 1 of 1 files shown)


pinnamur:
FR failed. I'm still unable to delete the project (enable the option to delete from disk as well) after it is run on Android device. The delete failed as it was unable to delete build_android.log file.

cli/commands/build.js
```diff
@@ -315,7 +317,7 @@ function patchLogger(logger, cli) {
 			// log it to our log file, stripping out the color codes
 			logger.log.filestream.write('\n' + prefix + (args.length > 2 ? sprintf.apply(null, args.slice(1)) : args[1]).replace(/\x1B\[\d+m/g, ''));
 		} else {
-			logger.log.buffer += '\n' + prefix + args[1].replace(/\x1B\[\d+m/g, '');
+			!logger.fileWriteEnabled && (logger.log.buffer += '\n' + prefix + args[1].replace(/\x1B\[\d+m/g, ''));
```
pinnamur: You are trying to buffer the data only if the fileWriteEnabled is false. Why do you need to buffer the data if we close the file stream, but probably never open it again ?

Shouldn't the entire block be guarded against fileWriteEnabled condition ?

pinnamur:
FR **passed**.

pinnamur:
CR **passed**.

PR #6586, opened 2015-01-20
[TIMOB-16660]: Add support for enableJavascriptInterface

hieupham007:
Testing steps in JIRA


ashcoding:
Everything works well as expected.

Just want to check on something. For device Android API 16 and below, when `enableJavascriptInterface: false` is set. There will be this error that is shown as info in the logs:-
`01-21 04:03:15.613: I/TiWebChromeClient.console(698): (main) [253,331095] ReferenceError: Can't find variable: TiApp (1:undefined)`.

I'm assuming this is due to https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java#L78 where the script for TiApp is being appended to the website and it is still trying to reference TiApp although it no longer can be accessed due to the object not being added via `addJavascriptInterface`.

Is this behavior expected and okay?

hieupham007:
That's a bug, I fixed it with the refactor

ashcoding:
Code reviewed and functionally tested. **Accepted**.

PR #6588, opened 2015-01-20
[TIMOB-18401] 3.5.x back port iOS - updated debugger library

pec1985:
Added a compiler flag to debugger to avoid a potential crash with third party modules
https://jira.appcelerator.org/browse/TIMOB-18401


jonalter:
FT **pass**

PR #6590, opened 2015-01-20
[TIMOB-18407] changed attributedString font property

cheekiatng:
JIRA here: https://jira.appcelerator.org/browse/TIMOB-18407


vishalduggal:
CR + FR ok. __APPROVED__

PR #6591, opened 2015-01-21
Fix rgbHex in parity report

garymathews:
- Minor fix to _rgbHex_ in parity report template, prevents _rgbHex_ from generating invalid values.


ingo:
Approved. Merging.

PR #6593, opened 2015-01-22
[TIMOB-18113] Specify Android build tools version

skypanther:
https://jira.appcelerator.org/browse/TIMOB-18113

### DIFF
--- a/node_modules/titanium-sdk/lib/android.js
+++ b/node_modules/titanium-sdk/lib/android.js
@@ -793,17 +793,16 @@ function findSDK(dir, config, androidPackageJson, callback) {
 		tasks = {},
 		buildToolsDir = path.join(dir, 'build-tools');
 
-	// check if we're we have a build-tools directory
+	/*
+		Determine build tools version to use based on either config setting
+		(android.buildTools.selectedVersion) or latest version
+	*/
 	if (fs.existsSync(buildToolsDir)) {
-		var files = fs.readdirSync(buildToolsDir).sort(),
-			i = 0,
-			len = files.length,
-			ver, file;
-
-		for (; i < len; i++) {
-			ver = files[i];
+		var file,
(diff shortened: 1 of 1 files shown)


eric34:
Some items for consideration:
I suggest we use the latest build tools as the default. This would allow the latest build to be used which would continue current behavior. If the user wishes to select a different installed version, that would satisfy the needs of the original ticket reporter. 

In regards to the supported range, there will be an issue if the range includes non-working versions. In our case, this would be the situation at ship time if we include version 21. Version 21.0.1 fails our build, as does 21.0.2. 21.0.3 works however. Since we have no mechanism to allow some or disallow others, I wonder if we feel this is a valid approach to track what is supported. 

There is one other consideration: The dependency chain that includes the Ti SDK, the Android SDK and the Build tools version must be taken into consideration. For example, with material theme support, the min compile target is API 21, and this requires build tools 21. So selecting another version of build tools would fail to build in this case.

skypanther:
Our thought behind not using the latest build tools as the default was so that we could control the tooling versions a bit more tightly. In the past, developers update tooling as soon as Google releases it and if there's an issue, we get hammered with issues. In this way, we have to release a new SDK that specifically supports the new build tools.

As implemented, the PR enables developers to select the newer build tools version manually. However, by default, only versions within the supported range are auto-selected.

Would you file a new ticket to address the "non-contiguous range" issue you identified? The appc.version.satisfies logic will need to be updated to support this sort of situation.

As for point three, the dependency chain is accounted for. The build tool's Pkg Revision file is checked against supported version information in the package.json file. The version is marked as not-supported if required versions don't match.

skypanther:
Per conversation with Ingo, Eric, and Praveen, I'm merging this PR. We will consider additional enhancements (segmented ranges, warnings) in separate tickets.

PR #6598, opened 2015-01-23
[TIMOB-18359] iOS: Fix bug for retrieving birthday in iOS8 and other contact updates

cheekiatng:
Fixed bug for retrieving birthday for iOS8
Added support for Social Profile
Added support for more Instant Messengers
Added support for Alternative Birthday (ios8 only)
Edited Contacts yml docs
Test case in JIRA: https://jira.appcelerator.org/browse/TIMOB-18359

### DIFF
--- a/apidoc/Titanium/Contacts/Person.yml
+++ b/apidoc/Titanium/Contacts/Person.yml
@@ -85,6 +85,24 @@ properties:
     type: String
     platforms: [android, iphone, ipad, tizen, blackberry]
     
+  - name: alternateBirthday
+    summary: |
+        Alternate birthday of the person. Single Dictionary.
+    description: |
+        The format of the dictionary accepted by this property is as follows:
+        
+        * Keys: `calendarIdentifier`, `era`, `year`, `month`, `day` and `isLeapMonth`.
+        * Values: Use `chinese`, `hebrew` and `islamic-civil` for `calendarIdentifier`.
+        Use `number` type for `era`, `year`, `month` and `day`. These must be consistent with
+        corresponding `calendarIdentifier`.
+        Use `boolean` type for `isLeapMonth`.
+
+
+    type: Dictionary
+    platforms: [iphone, ipad]
+    osver: {ios: {min: "8.0"}}
+    since: 3.6.0
+
   - name: created
     summary: Date and time that the person record was created. Single value.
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/Contacts/Person.yml
```diff
@@ -160,15 +178,31 @@ properties:
...
     type: Dictionary
     platforms: [android, iphone, ipad, blackberry]
     
+  - name: socialProfile
+    summary: |
+        Social profile information of the person. Multi-value.
+    description: |
+        The format of the dictionary accepted by this property is as follows:
+        
+        * Keys: any of `home`, `work` and/or `other`.
```
vishalduggal: I think the keys are incorrect. Shouldn't they be `facebookProfile`, `flickrProfile` etc?

apidoc/Titanium/Contacts/Person.yml
```diff
@@ -160,15 +178,31 @@ properties:
         `string` type value.
         
         The `service` key value may be one of `AIM`, `Facebook`, `GaduGadu`, `GoogleTalk`, `ICQ`, 
-        `MSN`, `QQ`, `Skype`, or `Yahoo`.
+        `Jabber`, `MSN`, `QQ`, `Skype`, or `Yahoo`.
```
vishalduggal: Is Jabber supported on Android?

vishalduggal:
The code works fine but the documentation and sample need work.

Code reviewed. Please address comments

apidoc/Titanium/Contacts/Person.yml
cheekiatng: I followed the same format as the existing instantMessage. Even the docs describes that these `keys` (`facebookProfile`,`flickrProfile`) have to be a subset of `home`, `work` and `other` keys. Not sure if it's because of achieving parity with Android. I do agree with you that there are predefined values though, will look into this further.

apidoc/Titanium/Contacts/Person.yml
cheekiatng: Added support for Jabber in Android in next commit. The keys `home`, `work`, `others` were there before this PR, and if changed, will break the parity for android. The predefined values are actually exactly these `service` key values written in the doc, and are case sensitive.

vishalduggal:
CR + FR ok. __APPROVED__

PR #6602, opened 2015-01-26
[TIMOB-17446] CLI args to specify i18n and platform locations

skypanther:
https://jira.appcelerator.org/browse/TIMOB-17446

### DIFF
--- a/node_modules/titanium-sdk/lib/i18n.js
+++ b/node_modules/titanium-sdk/lib/i18n.js
@@ -12,6 +12,13 @@ var appc = require('node-appc'),
 	DOMParser = require('xmldom').DOMParser;
 
 exports.load = function (projectDir, logger, opts) {
+	if(process.argv.indexOf('--i18n-directory')) {
+		// Enable developers to specify i18n directory location with build flag
+		var customI18n = process.argv[process.argv.indexOf('--i18n-directory')+1];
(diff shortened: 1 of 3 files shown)


pinnamur:
To keep the arguments consistent with the existing arguments such as --project-dir or --output-dir, I would recommend to keep the existing arguments as --i18n-dir <path> and 
--platform-dir <path>

node_modules/titanium-sdk/lib/i18n.js
```diff
@@ -12,6 +12,13 @@ var appc = require('node-appc'),
 	DOMParser = require('xmldom').DOMParser;
 
 exports.load = function (projectDir, logger, opts) {
+	if(process.argv.indexOf('--i18n-directory')) {
+		// Enable developers to specify i18n directory location with build flag
+		var customI18n = process.argv[process.argv.indexOf('--i18n-directory')+1];
+		if(customI18n && fs.existsSync(path.join(path.resolve(projectDir), customI18n))) {
+			projectDir = path.join(projectDir, customI18n);
+		}
+	}
```
pinnamur: Just being nitty-gritty. Can you format the above block ? For ex. spacing issues with if conditions - there should be a space between if and (

pinnamur:
FR **passed**. Should be good to merge after changing the arguments.

PR #6604, opened 2015-01-27
[TIMOB-17619] Clear margin flags when in sandbox

vishalduggal:
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-17619)

Regression path
KS Layout tests
TIMOB-11483


jonalter:
CR & FT **pass**

PR #6607, opened 2015-01-28
[TIDOC-2101] Display default values for properties in JsDuck

bhatfield:
Generate jsduck file

`node docgen -f jsca`

Open the `titanium_mobile/dist/titanium.js` file.

Properties that have a default value should display the default value next to the property name like
`@property [<PROPERTY_NAME>=<DEFAULT_VALUE>]`

Look for bubbleParent, accessibilityLabel, accessibilityValue, backgroundColor, etc.

### DIFF
--- a/apidoc/lib/jsduck_generator.js
+++ b/apidoc/lib/jsduck_generator.js
@@ -228,6 +228,17 @@ function exportReturns (api) {
 
 }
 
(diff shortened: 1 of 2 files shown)


tstatler:
Confirmed output in titanium.js. ***APPROVED!***

PR #6609, opened 2015-01-29
[TIDOC-2058] Added GitHub @editurl field to JSDuck generator.

tstatler:
Updated docgen.py to expose a new field on each class that contained the path to the current API file. The jsduck generator was updated to include an @editurl field in each class that pointed to the "Github edit URL". Because different APIs live in different locations, the specific URL needs to be computed based on whether it's a Titanium SDK API, a module API, an Alloy API, or a Tizen API.
- Since some modules are not public, they are 'blacklisted' from their URLs being included in the output.
- Since Alloy docs are passed as-is to JSDuck, it seemed easier just to manually add @editurl tags to Alloy file (see https://github.com/appcelerator/alloy/pull/659).
### Testing steps
1. Run `docgen.py -f jsduck -o ../build/`.
2. View build/titanium.js.

Spot check a few @class tags and make sure each has a corresponding @editurl tag. Copy/paste the @editurl URL into a browser and make sure it opens the expected edit page on Github.

### DIFF
--- a/apidoc/docgen.py
+++ b/apidoc/docgen.py
@@ -196,7 +196,6 @@ def process_yaml(source_dirs, options=None):
 				if os.path.splitext(filename)[-1] != ".yml" or filename in ignore_files:
 					continue
 				filepath = os.path.join(root, filename)
-				relative_path = os.path.relpath(filepath)
 				log.trace("Processing: %s" % filepath)
 				types = None
 				types = load_one_yaml(filepath)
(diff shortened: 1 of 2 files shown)


jhaynie:
this is awesome.

bhatfield:
@tstatler This should be added to the JavaScript version of the scripts too.

bhatfield:
@tstatler 
So it's generating the URLs, but they are not working.  The one's I check look like this:
~~~
/**
     * @class TableViewIndexEntry
     * @typestr Object
     * @pseudo
     * @editurl https://github.com/appcelerator-modules/TableView.yml/edit/master/l
     * <p>A simple object that represents an index entry in a <code>TableView</code>.</p>
     */
~~~

tstatler:
@bhatfield, sorry about that, I've updated the PR with the fix. Please try again.

Also, yes, the JavaScript version is coming shortly.

bhatfield:
@tstatler Link is working now.  Will merge with doctools PR is reviewed.

bhatfield:
**Approved.**

PR #6615, opened 2015-02-03
[TIDOC-2105] Ported @editurl JSDuck exporter feature to JavaScript.

tstatler:
Also removed extra forward slashes (/) from default output file paths.
### Testing steps
1. Run `node docgen.js -f jsduck`
2. Open ../dist/titanium.js.
3. Spot check a few @class tags and make sure each has a corresponding @editurl tag. Copy/paste the @editurl URL into a browser and make sure it opens the expected edit page on Github.

Also check that last line of output from docgen.js is "Generated output at ../dist**/**titanium.js" and not "Generated output at ../dist**//**titanium.js" (notice double forward slashes).

### DIFF
--- a/apidoc/docgen.js
+++ b/apidoc/docgen.js
@@ -771,17 +771,17 @@ formats.forEach(function (format) {
 			break;
 		case 'jsca' :
 			render = JSON.stringify(exportData, null, '    ');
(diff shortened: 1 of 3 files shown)


bhatfield:
**Approved.**

PR #6638, opened 2015-02-11
[TIMOB-18245] Ti.Media.takeScreenshot() does not render the selected iOS TabbedBar element

cheekiatng:
JIRA here: https://jira.appcelerator.org/browse/TIMOB-18245


vishalduggal:
CR + FR ok. __APPROVED__

PR #6639, opened 2015-02-11
IOS Code cleanup

vishalduggal:
Deletes a bunch of code that is not executed due to version checks.
Also removes a bunch of if conditionals that always evaluate to true due to minimum OS support.


pec1985:
Code reviewed and ran Kitchen Sink. **APPROVED!**

PR #6643, opened 2015-02-16
[TIMOB-18396] CLI : analytic events are not sent if you pass --url flag multiple times

feons:
https://jira.appcelerator.org/browse/TIMOB-18396


skypanther:
**APPROVED**

PR #6649, opened 2015-02-18
[TIMOB-18573] warn, but build, if Build Tools newer than known supported

skypanther:
https://jira.appcelerator.org/browse/TIMOB-18573


feons:
**Approved.**

PR #6652, opened 2015-02-19
TIMOB-18538 Update node-appc; args to spawn must be Array

pinnamur:
Cherry picking the changes to 3_5_X branch for the node 0.12 compatibility changes.


skypanther:
Be sure that the LiveView plug in is removed from your ~/.titanium/config.json file before testing this change or you'll get the same error as on the original ticket. With that said, this works as expected.

**APPROVED**

PR #6654, opened 2015-02-19
[TIMOB-17367] iOS - avoid left and right nav buttons on window to get GC'ed

pec1985:
Jira ticket https://jira.appcelerator.org/browse/TIMOB-17367


vishalduggal:
CR + FR ok. __APPROVED__

PR #6656, opened 2015-02-20
Text widget cleanup

vishalduggal:
General cleanup of TextWidget classes
Also fixes [TIMOB-16496](https://jira.appcelerator.org/browse/TIMOB-16496)


pec1985:
Code reviewed and tested it. **APPROVED!**

PR #6660, opened 2015-02-24
[TIMOB-13545]: Use httpClient to get image when decoder fails

hieupham007:
Testing steps in JIRA
https://jira.appcelerator.org/browse/TIMOB-13545


pec1985:
Merging

PR #6667, opened 2015-02-26
[TIMOB-18592]: Update code and doc for 'canceledOnTouchOutside'

hieupham007:
Testing steps in JIRA.

### DIFF
--- a/apidoc/Titanium/UI/Android/ProgressIndicator.yml
+++ b/apidoc/Titanium/UI/Android/ProgressIndicator.yml
@@ -44,9 +44,9 @@ properties:
 
   - name: canceledOnTouchOutside
     summary: |
-        Sets whether this dialog is canceled when touched outside the window's bounds.
+        When `cancelable` is set to `true` and this is set to `true`, the dialog is canceled when touched outside the window's bounds.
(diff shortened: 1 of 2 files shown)


apidoc/Titanium/UI/Android/ProgressIndicator.yml
```diff
@@ -44,9 +44,9 @@ properties:
 
   - name: canceledOnTouchOutside
     summary: |
-        Sets whether this dialog is canceled when touched outside the window's bounds.
+        When `cancelable` is set to `true` and this is set to `true`, the dialog is canceled when touched outside the window's bounds.
```
pec1985: The word cancelled is spelled with two L's, like `cancelled`

apidoc/Titanium/UI/Android/ProgressIndicator.yml
```diff
@@ -44,9 +44,9 @@ properties:
...
     type: Boolean
-    default: true
+    default: false
```
pec1985: Document that the `default` changed in the next version of Ti, like.... "starting in 4.0.0, this will be set to `true` by default"

PR #6675, opened 2015-03-02
[TIMOB-18355] Create objective-c formatting file

ingo:
DO NOT MERGE

This file is based on the google code style and is basically identical to our C++ version. I've formatted some sample files to give a selection of the results.

To Try:
- Install Alcatraz: https://github.com/supermarin/Alcatraz
- Install Clang-format
- Set format option to “File”
- Select files and format


iphone/Classes/GDataXMLNode.h
```diff
@@ -104,56 +103,61 @@ typedef NSUInteger GDataXMLNodeKind;
...
++ (id)namespaceWithName:(NSString*)name stringValue:(NSString*)value;
 
-+ (id)textWithStringValue:(NSString *)value;
-+ (id)cDataSectionWithStringValue:(NSString *)value;
++ (id)textWithStringValue:(NSString*)value;
++ (id)cDataSectionWithStringValue:(NSString*)value;
 
-- (NSString *)stringValue;
-- (void)setStringValue:(NSString *)str;
+- (NSString*)stringValue;
```
cheekiatng: Inconsistent changes made by Clang Format. Here it removed the space between NSString and * in stringValue, but in GDataXMLNode.m it maintains the space instead. Similarly in other files, it even intentionally adds the spaces between the type and *. From observation, the format will be consistent on 1 file, but not necessarily across different files in this case.

ingo: It is possible we are using an older version of clang-format. Perhaps updating to a newer compiled version would help.

It also seems to treat .h and .m files a bit differently.

How about http://objclean.com/index.php? It seems like it might work.

jonalter: @ingo that one looks better.
I'm pretty sure we want to have a space between the type and the `*` since that is Apple's standard.
```
(NSString *)stringValue; // Correct
(NSString*)stringValue;  // Incorrect
```

ingo: I tried objclean. I can't say it's that much better. Let me try a new version of clang-format

jonalter: Found a .clang-format file that is very close to Apple's standards https://github.com/haaakon/Apple-clang-format
It looks pretty good. All we need to do is save the file in the root of the project and then select "File" in clang format.

ingo: @jonalter We may not want to follow Apple's style, though. For instance, we specifically use tabs over spaces.

jonalter: @ingo Then we could take that format file and make our own that we use for all projects.

ingo: Sure. Note that the format file I have above is "BasedOnStyle: Google" So that's just a shorthand of all the settings. You can export the style for Google and then tweak whatever you like.

ingo:
@cheekiatng @jonalter I figured out DerivePointerAlignment. That was making it switch between files. What do you think now?

cheekiatng:
Think you are a genius. Using your clang-format I tried it on my own files too, and you managed to standardize the spacing for type and *. I also noticed that you managed to standardize a spacing between the '+' and '(' here: +(TiOrientationFlags) TiOrientationFlagsFromObject:(id)args. Maybe we should consider changing our stand about not having that spacing. 
Everything else looks ok.

jonalter:
Looks good

ingo:
@pec1985 @vishalduggal further thoughts? If not, I'll add the clang-format file and you can start using it in any new code you write. We'll reformat old code after 4.0.0 is released.

ingo:
Closing this in favor of #6701

PR #6676, opened 2015-03-02
[TIMOB-18433] Exposing hint text color for EditText

ashcoding:
...r for Hint to appear in EditText
Jira:https://jira.appcelerator.org/browse/TIMOB-18433

To test this out, you need to scons and build the SDK and install it inside studio. Then use it in a new app as follows.

Sample Code:

``` javascript
var win = Ti.UI.createWindow({backgroundColor: '#fff'});
var text = "Can you see me?";
var text1 = Ti.UI.createTextField({hintText: 'Can you see me',color:'#000',top: 10,});
var text2 = Ti.UI.createTextField({hintText: 'Can you see me',color:'#000',top: 50,});
win.add(text1);
win.add(text2);
win.open();
```


ashcoding:
@hieupham007 For this change, would https://github.com/appcelerator/titanium_mobile/blob/master/android/cli/commands/_build.js#L3385 require changes too?

Changing it to:

```javascript
		var flags = 'Theme.AppCompat.Light.DarkActionBar';
		if (this.tiapp.fullscreen || this.tiapp['statusbar-hidden']) {
			flags = 'Theme.AppCompat.Fullscreen';
		}
```

ashcoding:
Upon further investigation, this is not the solution. Will investigate this further.

ashcoding:
I exposed the hintColor property. 

Perhaps setting it as '#a8a8a8' (Grey) as default might also be something that should be considered. BUT that might not be good in the case where the user changes the Theme in the manifest. I'm typing this out to just get my flow of thought out.

Essentially,  the white hint color is a native behavior as the theme "Theme.AppCompat" sets the default background to be something that is dark and not white.

Perhaps a custom Theme for AppCompat for Titanium should be considered. Which uses the Theme.AppCompat as a base but sets the hintColor to grey. Which can be set here: https://github.com/ashcoding/titanium_mobile/commit/a4d6f38728fd6d01d6c6451f6c5e4f66d27985c1

I think this might be the best solution. I'll investigate this more.

hieupham007:
We need to document 'hintTextColor'. This is theme related, I see you already have a few workarounds in the ticket so I don't think there is anything else that needs to be done.

ashcoding:
For review.

Test:
```javascript
var win = Ti.UI.createWindow({backgroundColor: '#fff'});
var text = "Can you see me?";
var text1 = Ti.UI.createTextField({hintText: 'Can you see me', hintTextColor:'#0FF', color:'#000',top: 10,});
var text2 = Ti.UI.createTextField({hintText: 'Can you see me', hintTextColor:'#FF0', color:'#000',top: 50,});
win.add(text1);
win.add(text2);
win.open();
```

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -187,6 +187,26 @@ properties:
     default: No hint text.
     platforms: [android, blackberry]
     
+  - name: hintTextColor
+    summary: Hint text color.
```
hieupham007: hmm summary seems incomplete, maybe "Color of hint text that displays when field is empty", similar to hintText docs.

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -187,6 +187,26 @@ properties:
...
+    platforms: [android]
+    since: 4.1.0
+    description: |
+        Hint text color in Android is determined by the theme of the application. By default, the
+        theme is 'Theme.AppCompat' which is a Dark theme. When you create the background to be white,
```
hieupham007: `Theme.Appcompat`

apidoc/Titanium/UI/TextArea.yml
ashcoding: will do.

apidoc/Titanium/UI/TextArea.yml
ashcoding: Will do.

ashcoding:
@hieupham007 Changes done. Please review. Thank you.

hieupham007:

PR #6697, opened 2015-03-08
[TC-5357] Support destructive alert style

hansemannn:
JIRA reference: https://jira.appcelerator.org/browse/TC-5357

### DIFF
--- a/apidoc/Titanium/UI/AlertDialog.yml
+++ b/apidoc/Titanium/UI/AlertDialog.yml
@@ -173,7 +173,7 @@ properties:
     summary: Index to define the destructive button.
     platforms: [iphone, ipad]
     description: |
-        Note that this property available on iOS SDK 8 or above. 
+        Note that this property is only available on iOS 8 or above. Setting to -1 disables this option.
     type: Number
     default: -1
     
@@ -194,7 +194,7 @@ properties:
         <Titanium.UI.iPhone.AlertDialogStyle>. Using styles other than default one can break
         your dialog layout if more than two buttons used. All styles can handle up to two
         buttons comfortably, except for default style can handle up to six buttons when `title`
-        and `message` is empty or not given. Note that this property available on 
+        and `message` is empty or not given. Note that this property is only available on
         iOS SDK 5 or above. 
     type: Number
     default: <Titanium.UI.iPhone.AlertDialogStyle.DEFAULT>
--- a/apidoc/Titanium/UI/AlertDialog.yml
+++ b/apidoc/Titanium/UI/AlertDialog.yml
@@ -168,6 +168,14 @@ properties:
         On iOS and Mobile Web, set to `-1` to disable the cancel option.
     type: Number
     default: undefined (Android), -1 (iOS, Mobile Web)
+
+  - name: destructive
+    summary: Index to define the destructive button.
+    platforms: [iphone, ipad]
+    description: |
+        Note that this property available on iOS SDK 8 or above. 
+    type: Number
+    default: -1
     
   - name: message
     summary: Dialog message.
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIAlertDialogProxy.m
```diff
@@ -134,7 +136,7 @@ -(void)show:(id)args
                 NSString* btnName = [TiUtils stringValue:btn];
                 if (!IS_NULL_OR_NIL(btnName)) {
                     UIAlertAction* theAction = [UIAlertAction actionWithTitle:btnName
-                                                                        style:((curIndex == cancelIndex) ? UIAlertActionStyleCancel : UIAlertActionStyleDefault)
+                                                                        style:((curIndex == cancelIndex) ? UIAlertActionStyleCancel : (curIndex == destructiveIndex) ? UIAlertActionStyleDestructive :UIAlertActionStyleDefault)
```
pec1985: This line is a bit confusing and not easy to follow due to the multiple ternary operators. If would be better to spread the out for clarity purposes, something like this:
```
UIAlertActionStyle style;
if (curIndex == cancelIndex) {
	style = UIAlertActionStyleCancel;
} else if (curIndex == destructiveIndex) {
	style = UIAlertActionStyleDestructive;
} else {
	style  = UIAlertActionStyleDefault;
}
```

apidoc/Titanium/UI/AlertDialog.yml
```diff
@@ -168,6 +168,14 @@ properties:
...
     type: Number
     default: undefined (Android), -1 (iOS, Mobile Web)
+
+  - name: destructive
+    summary: Index to define the destructive button.
+    platforms: [iphone, ipad]
+    description: |
+        Note that this property available on iOS SDK 8 or above. 
+    type: Number
+    default: -1
```
pec1985: Would be nice to specify that setting it to `-1` disables this option, or does nothing. Having a default value of `-1` without explaining what it does makes it unclear

pec1985:
PR reviewed, left a couple of comments.

iphone/Classes/TiUIAlertDialogProxy.m
pec1985: Also, if this is iOS 8 an above only, make sure you check for that with:
```
if ([TiUtils isIOS8OrGreater]) { ..... }
```

hansemannn:
Thank you! Comments are reasonable, will have a look at it!

hansemannn:
Updated docs and code readability as well as a bounding-check like it is done with the cancel-button. I think the iOS 8 check is not necessary in this case, because the code is already in the conditional iOS 8 block to support the UIAlertController.

pec1985:
One more thing... the destructive index is missing from the click event. For consistency, add it here as well `fireClickEventWithAction`:
```
NSMutableDictionary *event = [NSMutableDictionary dictionaryWithObjectsAndKeys:
                               NUMUINTEGER(indexOfAction),@"index",
                               [NSNumber numberWithInt:cancelIndex],@"cancel",
                               nil];
```

hansemannn:
Added the destructive index and implemented a quick sample example code to test:

``` javascript
	backgroundColor: "#fff"
});

	title: "Show destructive dialog"
});

	title: "Proceed",
	message: "Do you really want to proceed?",
	destructive: 0,
	cancel: 1,
	buttonNames: ["Destructive", "Cancel", "Proceed"]
});

	dialog.show();
});

dialog.addEventListener("click", function(e) {
	Ti.API.info(e);
});

win.add(button);

win.open();
```

pec1985:
i have opened a new PR with your commits for Titanium master branch #6842

pec1985:
There is one minor thing missing in the documentation. The `since` tag (for Ti SDK) and the `osver` for the iOS version. You can simple cherry pick from master since I already merged that
```
git cherry-pick cfcbc96
```

hansemannn:
Closing this outdated PR in favor of #6842

PR #6698, opened 2015-03-09
Fix(TiShadow): Android service dont working in tishadow

rotorgames:
Android Service, does not work with tishadow.

Titanium uses the directory `android_asset`, and tishadow retains all the resources in the `appdata-private`, so running the service was impossible.

The changes allow us to determine when the compilation comes with tishadow, and allow you to take the source files for the service of `appdata-private`.

Please accept these changes, it is important.

P.S. I apologize for my bad English.

### DIFF
--- a/android/modules/android/src/java/ti/modules/titanium/android/TiJSIntervalService.java
+++ b/android/modules/android/src/java/ti/modules/titanium/android/TiJSIntervalService.java
@@ -21,6 +21,7 @@ import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.proxy.IntentProxy;
 import org.appcelerator.titanium.proxy.ServiceProxy;
 import org.appcelerator.titanium.util.TiBindingHelper;
+import org.appcelerator.titanium.util.TiShadowHelper;
 
 import android.app.Service;
 import android.content.Intent;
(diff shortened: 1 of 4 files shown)


dbankier:
+1

caspahouzer:
+1

ingo:
Hi All. Thank you for the PR, and we'd like to help. A few notes:

1. For an item like this, we need a JIRA ticket and test case
2.This change needs to be product-agnostic. So we need a way of adding this so that tishadow can take advantage of it, but is not tishadow-specific.
3. Documentation updates for areas where this changes behavior.

Thanks!

rotorgames:
@ingo @dbankier Thanks for the quick reply.

Can be done to take any titanium files from `appdata-private` directory, and it will be included by using `<property name = "useDataDirectory" type = "bool"> true </ property>`.

TiShadow will add this property to `tiapp.xml`.

How do you like this idea?

ingo:
@rotorgames Is there a need to make this property work on other platforms too? It might be better to specify _which_ directory to use, rather than just to use a different one. @ashcoding @hieupham007 @pec1985 thoughts?

rotorgames:
@ingo As far as I know, this problem is only on android, but it can do as you want. Probably it will be even better and more convenient.

ashcoding:
@Ingo I'll go and investigate things out first.

rotorgames:
Issue jira: https://jira.appcelerator.org/browse/TC-5428

rotorgames:
@ingo @ashcoding @hieupham007 @pec1985 jira Issue: https://jira.appcelerator.org/browse/TC-5428

hansemannn:
Closing PR since it wants to merge into the stale branch `3_5_X`. Please submit a new one for master and we can take care of it, thank you!

PR #6699, opened 2015-03-09
[TIMOB-18662] Bump Titanium SDK Version to 3.5.2

jonalter:
[TIMOB-18662](https://jira.appcelerator.org/browse/TIMOB-18662)

### DIFF
--- a/build/titanium_version.py
+++ b/build/titanium_version.py
@@ -1,2 +1,2 @@
-version = '3.5.1'
+version = '3.5.2'
 module_apiversion = '2'
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "titanium-mobile",
 	"description": "Appcelerator Titanium Mobile",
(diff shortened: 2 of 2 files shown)


ingo:
Approved. Merging.

jhaynie:
do we really need to do PR for stuff as simple as this?

ingo:
Yes, because we've forgotten to update both files. The fact we have both files is super-annoying. Once we solve https://jira.appcelerator.org/browse/TIMOB-18508, then no.

jhaynie:
ok, can we just fix that? that seems like a quick fix.

ingo:
Yes--problem is it is reused in a bunch of places. Trying to get to it. I am actively trying to refine as much process as quickly as possible :)

jhaynie:
can't we just put it in the top level build to generate the right value in both places as-is. at least then you don't hard code it in two places.

PR #6701, opened 2015-03-10
[TIMOB-18355] Add clang-format and updated js formatting rules

ingo:


ingo:
Same as #6675, but without the Obj-C file changes. We'll start making those changes once 4.0 ships so as not to make backports complicated.  Note that to actually format the files, you would need to run "grunt format". I chose not to format as part of npm test (yet) because it would be possibly destructive.

PR #6702, opened 2015-03-10
Grunt package

ingo:
This is meant to solve what was discussed in #6699. We make it easier to bump versions in titanium_mobile.

To use and test:
1. Bump to next patch: grunt bump:patch (should go to X.Y.Z + 1)
2. Bump to next minor: grunt bump:minor (should go to X.Y + 1.Z)
3. Bump to next major: grunt bump:major (should go to X + 1.Y.Z)

Note that we use grunt-version here over grunt-bump because we need to bump package.json and titanium_version. However, we use the grunt-bump syntax as I believe it is the better long-term plugin and there is a discussion to fix this here: https://github.com/vojtajina/grunt-bump/issues/131

See https://jira.appcelerator.org/browse/TIMOB-18662 as well.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,6 +8,17 @@ module.exports = function (grunt) {
         clangFormat: {
             src: [] // unused ATM
         },
+		version: {
+			patchTitaniumVersion: {
+				options: {
+					prefix: '^version\\s+=\\s+[\'"]'
+				},
+				src: ['build/titanium_version.py']
+			},
+			patchPackage: {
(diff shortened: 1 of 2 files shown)


jonalter:
Looks good!

ingo:
@jonalter feel free to merge unless you have objections :)

jhaynie:
yeah

hansemannn:
Hi @ingo, do you still want this change in? We should update it to work with master and latest node deps before. Thanks!

PR #6706, opened 2015-03-10
[TIMOB-18464]: Fix crash when app exits.

hieupham007:
Testing steps: Run KS, hit 'back' button at splash screen, app shouldn't crash.


ashcoding:
Code reviewed and functionally tested with Kitchen Sink. Request **accepted**

PR #6718, opened 2015-03-12
[TIMOB-18680] Check both cached and proxy values

vishalduggal:
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-18680)
Regression path
KS->ListView


pec1985:
Reviewed and tested. **APPROVED!**

PR #6719, opened 2015-03-12
[TIMOB-18495] Pass guid to titanium_prep - update libraries

pec1985:
Epic: https://jira.appcelerator.org/browse/TIMOB-18494
Specific: https://jira.appcelerator.org/browse/TIMOB-18495

PR #6726, opened 2015-03-14
[TIMOB-18495] Pass guid to titanium_prep - update libraries

cb1kenobi:
This is an updated version of #6719 that fixes a bug in the Android build when titanium_prep errors out.


ingo:
Approved. Merging.

PR #6730, opened 2015-03-16
[TIMOB-18706] Fix error when Windows Phone SDK is NOT installed

garymathews:
- Initialize _results.windowsphone_ to not be _null_

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-18706)


sgtcoolguy:
Closing since the fix needs to be done in appcelerator/windowslib

PR #6737, opened 2015-03-20
[TIMOB-18391] Fix error when using String.format with multiple parameters

jonalter:
[TIMOB-18391](https://jira.appcelerator.org/browse/TIMOB-18391)

Strings that come from L('key') already have %s replaced with %@


garymathews:
CR:FT pass

PR #6738, opened 2015-03-20
[TIMOB-18391] 4_0_X Fix error when using String.format with multiple parameters

jonalter:
[TIMOB-18391](https://jira.appcelerator.org/browse/TIMOB-18391)

Strings that come from L('key') already have %s replaced with %@


garymathews:
CR:FT pass

Gerst20051:
@jonalter @garymathews any idea what commit broke this? and when this will be released?

Gerst20051:
I also wanna point out that this is only affecting iphone simulators.

PR #6746, opened 2015-03-26
[TIMOB-16244] Inital implementation of scrollstart and scrollend for listview

pec1985:
Jira ticket: https://jira.appcelerator.org/browse/TIMOB-16244
Note:
This adds two events to the Ti.UI.ListView for **Android** and **iOS** only: `scrollstart` and `scrollend`
The event arguments contain the following information:

```
eventArgs = {
    firstVisibleSectionIndex: Number
    firstVisibleItemIndex: Number
    firstVisibleSection: [object Ti.UI.ListSection]
    firstVisibleItem: [object Object]
    visibleItemCount: Number
}
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
@@ -34,6 +34,7 @@ import ti.modules.titanium.ui.android.SearchViewProxy;
 import ti.modules.titanium.ui.widget.searchbar.TiUISearchBar;
 import ti.modules.titanium.ui.widget.searchbar.TiUISearchBar.OnSearchChangeListener;
 import ti.modules.titanium.ui.widget.searchview.TiUISearchView;
+import ti.modules.titanium.ui.widget.tableview.TiTableView;
 import android.app.Activity;
 import android.content.Context;
 import android.graphics.Color;
@@ -52,6 +53,7 @@ import android.widget.FrameLayout;
 import android.widget.ListView;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
+import android.widget.AbsListView.OnScrollListener;
 
 public class TiListView extends TiUIView implements OnSearchChangeListener {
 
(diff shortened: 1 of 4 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -302,7 +304,68 @@ public TiListView(TiViewProxy proxy, Activity activity) {
...
+						KrollDict eventArgs = new KrollDict();
+						Pair<ListSectionProxy, Pair<Integer, Integer>> info = getSectionInfoByEntryIndex(_firstVisibleItem);
+						
+						
+						int itemIndex = info.second.second;
+						ListSectionProxy section = info.first;
+						
+						if(section.getHeaderTitle() == null || section.getHeaderView() == null) {
+							if(itemIndex > 0) itemIndex -= 1;
+							_visibleItemCount -=1;
```
ashcoding: This line might be causing this problem:- 

For edge case, when you are either at the top of the list and continue scrolling up or at the bottom of the list and scrolling down, it continues to count and go negative.

```
03-29 23:30:38.900: I/TiAPI(9211):  END firstVisibleSectionIndex: 29 firstVisibleItemIndex: 11 firstVisibleSection: [object ListSection] firstVisibleItem: [object Object] visibleItemCount: -16
```

cheekiatng:
on the iOS end, CR and FT all ok.

pec1985:
@ashcoding Thanks! I just pushed a commit that fixes that issue

iphone/Classes/TiUIListView.m
```diff
@@ -1653,18 +1653,71 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
...
+        
+        TiUIListSectionProxy* section = [[self listViewProxy] sectionForIndex: [indexPath section]];
+        NSDictionary* eventArgs = @{
+                                    @"firstVisibleSectionIndex": NUMINTEGER([indexPath section]),
+                                    @"firstVisibleItemIndex": NUMINTEGER([indexPath row]),
+                                    @"firstVisibleSection": section,
+                                    @"firstVisibleItem": [section itemAtIndex:[indexPath row]],
+                                    @"visibleItemCount": NUMUINTEGER(visibleItemCount)
+                                    };
+        [[self proxy] fireEvent:eventName withObject:eventArgs propagate:NO];
```
vishalduggal: This event is set to not propagate while the event on Android propagates. Suggest moving both platform to non propagation. Also suggest that the check for existence of event listener check this proxy only and not parents. Both iOS and Android have variants for this in base TiViewProxy

iphone/Classes/TiUIListView.m
```diff
@@ -1653,18 +1653,71 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
...
+
+
+// For now, this is fired on `scrollstart` and `scrollend`
+- (void)fireScrollEvent:(NSString*)eventName
+{
+    if([[self proxy] _hasListeners:eventName])
+    {
+        UITableView* tableView = [self tableView];
+        
+        NSIndexPath* indexPath = [self firstVisibleIndexPath:0];
```
vishalduggal: I would not do this at all. Just rely on the system supplied `indexPathsForVisibleRows` for the tableView. Less computation and fewer errors

vishalduggal:
What is the behavior of the event parameters when search is active(ie the scrollView is the tableView from SearchDisplayController or searchText is a valid string and the tableView is showing filtered results)

vishalduggal:
Code reviewed. Please address comments.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -809,6 +872,8 @@ protected void scrollToItem(int sectionIndex, int sectionItemIndex, boolean anim
 		final int position = findItemPosition(sectionIndex, sectionItemIndex);
 		if (position > -1) {
 			if (animated) {
+				fireScrollEvent(TiC.EVENT_SCROLLSTART);
```
hieupham007: you shouldn't fire the event on automated scrolling.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -303,6 +307,34 @@ public TiListView(TiViewProxy proxy, Activity activity) {
 		listView.setFocusableInTouchMode(true);
 		listView.setDescendantFocusability(ViewGroup.FOCUS_AFTER_DESCENDANTS);
 
+		listView.setOnScrollListener(new OnScrollListener()
+		{
+			@Override
+			public void onScrollStateChanged(AbsListView view, int scrollState)
+			{
+				switch(scrollState) {
```
hieupham007: why not just use if/else?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -472,6 +504,37 @@ public void processProperties(KrollDict d) {
 		super.processProperties(d);
 		
 	}
+	
+	private void fireScrollEvent(String eventName) {
+		if(listView == null) return;
```
hieupham007: if (...) {
}

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -472,6 +504,37 @@ public void processProperties(KrollDict d) {
...
+		if(!proxy.hasListeners(TiC.EVENT_SCROLLSTART) && !proxy.hasListeners(TiC.EVENT_SCROLLEND)) return;
```
hieupham007: same here

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -440,6 +440,11 @@
 	/**
 	 * @module.api
 	 */
+	public static final String EVENT_SCROLLSTART = "scrollstart";
```
hieupham007: this needs to be placed alphabetically

apidoc/Titanium/UI/ListView.yml
```diff
@@ -397,6 +397,62 @@ events:
...
+        the `firstVisibleItemIndex` to determine the position of the listview.
+    platforms: [iphone, ipad, android]
+    since: 4.1.0
+    properties: 
+      - name: visibleItemCount
+        summary: The number of visible items in the listView when the event fires
+        type: Number
+
+      - name: firstVisibleItem
+        summary: The first visible item in the listView when the event fires, this item might not be fully visible
```
bhatfield: These are two separate sentences.  You can either change the comma to a semicolon, or change the comma to a period and capitalize 'this'.

Repeat for firstVisibleItemIndex and the same for the scroll end event.

bhatfield:
General note:
Change listView and listview to 'list view'

vishalduggal:
The PR in general works well. There is one use case where the FR fails.

Scroll the ListView in either direction and before the scrolling animation completes tap the status bar. In that case scrollend is fired with the wrong parameters. Expect the firstVisibleItemIndex to be 0. Instead it is always the index of item which is at the top when the status bar is tapped.

Once this is fixed will accept PR for iOS

vishalduggal:
CR + FR ok. Accepted for iOS

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -303,6 +306,59 @@ public TiListView(TiViewProxy proxy, Activity activity) {
...
+				}
+
+				KrollDict eventArgs = new KrollDict();
+				Pair<ListSectionProxy, Pair<Integer, Integer>> info = getSectionInfoByEntryIndex(_firstVisibleItem);
+				int visibleItemCount = _visibleItemCount;
+				
+				int itemIndex = info.second.second;
+				ListSectionProxy section = info.first;
+				
+				if (section.getHeaderTitle() == null || section.getHeaderView() == null) {
```
hieupham007: getSectionInfoByEntryIndex should take header and footer into account already. This will cause index to be 1 off when header is present

hieupham007:
CR + FR. Index is off by 1 when header is present. Left a comment

hieupham007:
CR + FR. Request **accepted**

PR #6752, opened 2015-03-27
[TIMOB-18143] iOS Fixed searchbar background image

pec1985:
The commit should have said "background image" instead of "background color"
ticket: https://jira.appcelerator.org/browse/TIMOB-18143

### DIFF
--- a/iphone/Classes/TiUISearchBar.m
+++ b/iphone/Classes/TiUISearchBar.m
@@ -141,17 +141,19 @@
 -(void)setBackgroundImage_:(id)arg
 {
     UIImage *image = [self loadImage:arg];
-    if([TiUtils isIOS7OrGreater]) {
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUISearchBar.m
```diff
@@ -141,11 +141,22 @@ -(void)setBarColor_:(id)value
 -(void)setBackgroundImage_:(id)arg
 {
     UIImage *image = [self loadImage:arg];
-    [[self searchBar] setBackgroundImage:image];
+    if([TiUtils isIOS7OrGreater]) {
```
vishalduggal: iOS7 is our minimum supported version, so this if condition will always return true. You can remove it

cheekiatng:
CR and FT Passed. Merging.

PR #6754, opened 2015-03-30
[TIMOB-18767] Android: Add support for xxxhigh to displayCaps.density

FokkeZB:


hieupham007:
Code reviewed. Looks good

hieupham007:
Functionally tested. The nexus 6 is reporting '560' DPI instead of '640'.

FokkeZB:
@hieupham007, yes and TVs report 213 and because we're having 160/medium as the default that gives odd results. We should do >= instead of ==

hieupham007:
I just submit https://github.com/appcelerator/titanium_mobile/pull/6776, added support for a few densities. I think this should fix our problem

cheekiatng:
Closing PR since TIMOB-18767 is resolved and closed. Thanks!

PR #6757, opened 2015-03-31
Fix minor formatting errors exposed in latest build.

ingo:
The last build (https://travis-ci.org/appcelerator/titanium_mobile/builds/56451004) appears to have errored (though it does not appear to fail locally). Correcting issues.


ingo:
Approved. Merging.

PR #6768, opened 2015-04-07
[TIMOB-18781] Updated ioslib to 0.2.14 which adds support for iOS simrun...

cb1kenobi:
...times introduced by Xcode 6.2. Fixed bad grammar in iOS build error message.


muhammaddadu:
FT: **PASSED**

PR #6776, opened 2015-04-10
[TIMOB-18767]: Added support for a few densities.

hieupham007:
testing steps in JIRA

### DIFF
--- a/android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
+++ b/android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
@@ -67,15 +67,19 @@ public class DisplayCapsProxy extends KrollProxy
 				return "high";
 			case DisplayMetrics.DENSITY_MEDIUM :
 				return "medium";
-			case DisplayMetrics.DENSITY_XHIGH :
+			case 213: //TV
 				return "xhigh";
-			case DisplayMetrics.DENSITY_400:
+			case 280: //Introduce in API 22.
+				return "xhigh";
+			case 320:
+				return "xhigh";
+			case 400:
 				return "xxhigh";
-			case DisplayMetrics.DENSITY_XXHIGH :
+			case 480 :
 				return "xxhigh";
-			case DisplayMetrics.DENSITY_560:
+			case 560:
 				return "xxxhigh";
-			case DisplayMetrics.DENSITY_XXXHIGH:
+			case 640:
 				return "xxxhigh";
 			case DisplayMetrics.DENSITY_LOW :
 				return "low";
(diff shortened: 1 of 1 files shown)


android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -67,10 +67,16 @@ public String getDensity() {
 				return "high";
 			case DisplayMetrics.DENSITY_MEDIUM :
 				return "medium";
-			case 320 : // DisplayMetrics.DENSITY_XHIGH (API 9)
+			case DisplayMetrics.DENSITY_XHIGH :
```
FokkeZB: Are you sure we can use this constant? The original code said something about this only being available in API 9 and higher? Or is it now available in all API's we support?

If we can use the constants, why not implement all of them (http://developer.android.com/reference/android/util/DisplayMetrics.html) and map in-betweens like `DENSITY_560` to `560` and `DENSITY_TV` to `tv` to stay in line with the original platform values?

hieupham007: XHIGH should be fine since min support is 10 now. I'll add the TV and switch them to numbers.

android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -67,10 +67,20 @@ public String getDensity() {
 				return "high";
 			case DisplayMetrics.DENSITY_MEDIUM :
 				return "medium";
-			case 320 : // DisplayMetrics.DENSITY_XHIGH (API 9)
+			case 213: //TV
```
FokkeZB: Shouldn't this return 'tv' to be in line with http://developer.android.com/reference/android/util/DisplayMetrics.html?

android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -67,10 +67,20 @@ public String getDensity() {
...
 				return "xhigh";
+			case 280: //Introduce in API 22.
```
FokkeZB: Shouldn't this return '280' to be in line with http://developer.android.com/reference/android/util/DisplayMetrics.html?

android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -67,10 +67,20 @@ public String getDensity() {
...
+				return "xhigh";
+			case 320:
+				return "xhigh";
+			case 400:
```
FokkeZB: Shouldn't this return '400' to be in line with http://developer.android.com/reference/android/util/DisplayMetrics.html?

android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -67,10 +67,20 @@ public String getDensity() {
...
+				return "xxhigh";
 			case 480 :
 				return "xxhigh";
+			case 560:
```
FokkeZB: Shouldn't this return '560' to be in line with http://developer.android.com/reference/android/util/DisplayMetrics.html?

PR #6778, opened 2015-04-14
[TIMOB-18473]: Remove density scaling for background drawables to preven...

hieupham007:
...t memory overload.

Testing steps in JIRA.

http://jira.appcelerator.org/browse/TIMOB-18473


pec1985:
Code reviewed and tested. **APPROVED!**

PR #6817, opened 2015-05-01
[TIMOB-18830] iOS: ImageView delays first image of animation

vishalduggal:
Test is in [JIRA](https://jira.appcelerator.org/browse/TIMOB-18830)


cheekiatng:
CR and FT passed. Merging.

PR #6408, opened 2014-11-27
[TIMOB-18868] Do not overwrite Info.plist keys set by user

FokkeZB:
- Reads `CFBundleVersion` and `CFBundleShortVersionString` from `tiapp.xml`
- Only sets `CFBundleVersion` if target is iTunes or there is no custom value in either `tiapp.xml` or a custom `Info.plist`
- Only sets `CFBundleShortVersionString` if there is no custom value in either `tiapp.xml` or a custom `Info.plist`
- Only generates `CFBundleIconFiles` if there is no custom value in either `tiapp.xml` or a custom `Info.plist`
- Only generates `UILaunchImages` if there is no custom value in either `tiapp.xml` or a custom `Info.plist`, also fixing needing `ti clean` after adding new launch images.
- Only generates `UIAppFonts` if there is no custom value in either `tiapp.xml` or a custom `Info.plist`

PR #6818, opened 2015-05-03
[TIMOB-17993] Correct issues with CFBundleVersion and CFBundleShortVersionString

FokkeZB:
A version of https://github.com/appcelerator/titanium_mobile/pull/6408 that only has changes directly related to the original request in https://jira.appcelerator.org/browse/TIMOB-17993
- Reads CFBundleVersion and CFBundleShortVersionString from `tiapp.xml/ios/plist`
- Only sets `CFBundleVersion` if target is iTunes (which needs a unique version) or there is no custom value in either `tiapp.xml/ios/plist` or a custom `Info.plist`
- Only sets `CFBundleShortVersionString` if there is no custom value in either `tiapp.xml/ios/plist` or a custom `Info.plist`

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -1975,10 +1975,11 @@ iOSBuilder.prototype.createInfoPlist = function createInfoPlist(next) {
 	}
 
 	// if the user has a Info.plist in their project directory, consider that a custom override
+	var custom;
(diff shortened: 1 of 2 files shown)


cb1kenobi:
Closing as this functionality was already resolved by #6408.

PR #6821, opened 2015-05-06
[TIMOB-18765] iOS: Fix for uiimageview regression project name or file name has a space

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-18765

### DIFF
--- a/iphone/Classes/TiUtils.m
+++ b/iphone/Classes/TiUtils.m
@@ -773,34 +773,7 @@ If the new path starts with / and the base url is app://..., we have to massage 
 		return [NSURL URLWithString:relativeString];
 	}
 
-	NSURL *result = nil;
-		
-	// don't bother if we don't at least have a path and it's not remote
-	//TODO: What is this mess? -BTH
-	if ([relativeString hasPrefix:@"http://"] || [relativeString hasPrefix:@"https://"])
-	{
-		NSRange range = [relativeString rangeOfString:@"/" options:0 range:NSMakeRange(7, [relativeString length]-7)];
-		if (range.location!=NSNotFound)
-		{
-			NSString *firstPortion = [relativeString substringToIndex:range.location];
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiUtils.m
```diff
@@ -794,8 +794,13 @@ +(NSURL*)toURL:(NSString *)relativeString relativeToURL:(NSURL *)rootPath
 			}
 		}
 	}
-
-	result = [NSURL URLWithString:[relativeString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding] relativeToURL:rootPath];
+	//only add percentescape if there are spaces in relativestring
+	if ([[relativeString componentsSeparatedByString:@" "] count] -1 == 0) {
```
vishalduggal: This will work for the fix but it is not the right fix. For example if the image was specified as 'image%1.png' this if condition will fail and you would get a nil result. 

The real problem is in ImageView loadUrl function where we are using 
`NSURL *url_ = [TiUtils toURL:[img absoluteString] proxy:self.proxy];`

What we sould be using is 
`NSURL *url_ = [TiUtils toURL:[img path] proxy:self.proxy];`

vishalduggal:
Code reviewed. __CR/FR Failed__.

vishalduggal:
You might also want to cleanup the 
`+(NSURL*)toURL:(NSString *)relativeString relativeToURL:(NSURL *)rootPath` method in `TiUtils.m`
We have special handling of URL's with http and https schemes.

If `- (NSString *)stringByAddingPercentEscapesUsingEncoding:(NSStringEncoding)enc;` is performing the same functionality we should remove the block of code that transforms these remote URL's

cheekiatng:
Good point. Thanks!

vishalduggal:
CR + FR ok. __APPROVED__

PR #6835, opened 2015-05-11
[TIMOB-18877] Geocoder sample from Studio dashboard fails with uncaught NSException on 7.1 simulators

muhammaddadu:
- The binary that packaged with titanium_mobile did not reflect the released version of the module

[TIMOB-18877](https://jira.appcelerator.org/browse/TIMOB-18877)


cheekiatng:
Tested ok. Merging.

PR #6837, opened 2015-05-12
[TIMOB-18824] Android: ListView crashes app when scrolled to end of list

alexgbernier:


alexgbernier:
I signed the CLA earlier today, BTW.

cheekiatng:
Thanks! we'll look into this.
JIRA: https://jira.appcelerator.org/browse/TIMOB-18824

ashcoding:
@alexgbernier Just checking, did you sign the CLA here? https://cla.appcelerator.com/

alexgbernier:
Ah. Signed the one in some documentation. Maybe old documentation. Just
signed this one, too.

BTW, when I build with scons, my apk is about 23mb bigger than with the
builds from continuous.

ashcoding:
@alexgbernier Thanks for signing the CLA.
Not sure if the size being bigger more than normal. My current scons size is 155.1 MB on an iOS.

On another note, tried the test code that was in Jira on Nexus 6 (5.0) and a HTC Desire Eye (5.0.2) with the Master branch and SDK 4.0.0v20150429154210. I can't get it to crash. Seems to be working normally.

alexgbernier:
Hey @ashcoding, thank you for looking into this. Regarding size, I was referring to the APK built with a built-from-source SDK. My app APK built with a continuous integration SDK is 23mb. My APK built with an SDK built with scons is 43.3mb. What command do you use to build your SDK? I used `scons build_jsca=0 android=1`. If the APK were same size as with continuous integration build APK I could release my fix while we discuss this pull request. Any tips appreciated. At this risk of 

With regard to the bug, I actually can't reproduce it now on my device either, but we've been getting reports of it from our users, and we have ~110 crashes a day from Crittercism:

```
java.lang.NullPointerException: Attempt to read from field 'java.lang.Object android.util.Pair.second' on a null object reference at ti.modules.titanium.ui.widget.listview.TiListView$1.onScrollStateChanged(TiListView.java:334)
```

Seeing as the getSectionInfoByEntryIndex method can return null in several situations, it seems sound to have a null check of its response, crash reports aside. No doubt I agree it would be ideal to be able to reproduce the crash, then verify the fix. Let me know your thoughts.

hieupham007:
Code reviewed. Looks good.

ashcoding:
Code reviewed and functionally tested. All good. **Accepted**
:thumbsup:

alexgbernier:
Thank you!

PR #6842, opened 2015-05-12
[TC-5357] iOS - Support destructive alert style

pec1985:
PR originally from @hansemannn #6697 
This is against master


pec1985:
Code reviewed and tested. Thanks @hansemannn!

PR #6845, opened 2015-05-13
[TIMOB-17573] Parity Android: Set Multiple Cookies in Request Header

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-17573

### DIFF
--- a/android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
+++ b/android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
@@ -791,8 +791,23 @@ public class TiHTTPClient
 	public void setRequestHeader(String header, String value)
 	{
 		if (readyState <= READY_STATE_OPENED) {
-			headers.put(header, value);
-
+			if (value == null) {
+				// If value is null, remove header
+				headers.remove(header);
+			} else {		
+				if (headers.containsKey(header)){
+					// Appends a value to a header
+					// If it is a cookie, use ';'. If not, use ','.
+					String seperator = ("Cookie".equalsIgnoreCase(header))? "; " : ", ";
+					StringBuffer val = new StringBuffer(headers.get(header));
+					val.append(seperator+value);
+					headers.put(header, val.toString());
+				} else {
+					// Set header for the first time
+					headers.put(header, value);
+				}
+			}
+			
 		} else {
 			throw new IllegalStateException("setRequestHeader can only be called before invoking send.");
 		}


android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -791,8 +791,23 @@ public void clearCookies(String url)
...
-			headers.put(header, value);
-
+			if (value == null) {
+				// If value is null, remove header
+				headers.remove(header);
+			} else {		
+				if (headers.containsKey(header)){
+					// Appends a value to a header
+					// If it is a cookie, use ';'. If not, use ','.
+					String seperator = ("Cookie".equalsIgnoreCase(header))? "; " : ", ";
```
hieupham007: separator typo

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -791,8 +791,23 @@ public void clearCookies(String url)
 	public void setRequestHeader(String header, String value)
 	{
 		if (readyState <= READY_STATE_OPENED) {
-			headers.put(header, value);
-
+			if (value == null) {
+				// If value is null, remove header
+				headers.remove(header);
```
hieupham007: not sure if this is a good idea. Why aren't we accepting null values?

ashcoding: The idea was, if you set cookies, it will be appended and if you set null, it is cleared. A null value is being used as a way to clear all the values since we only have a set method.

Should this behavior be changed?

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
ashcoding: Thanks for catching this. Will change this to "Separator"

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
hieupham007: Not sure I understand your point. This code sets the headers, not cookies?

ashcoding: The cookies are being (can be) set in the headers.

Also the null value is following this behavior: https://github.com/appcelerator/APSHTTPClient/blob/c8531ff81a381c90ad445ceff891ded99214e6a9/APSHTTPClient/APSHTTPRequest.m#L172

https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/Network/HTTPClient.yml#L249

If I misunderstood it somehow, do let me know.

hieupham007:

PR #6849, opened 2015-05-14
[TIMOB-18865] iOS: support toolbar for undocked keyboard

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-18865


garymathews:
FR: **PASS**

PR #6852, opened 2015-05-15
[TIMOB-18765] Only escape if it is a local path and has spaces 

vishalduggal:
Titanium calls `[TiUtils toURL:(NSString*)relativeString relativeToURL:(NSURL *)rootPath]` many times and in some scenarios it will call this method more than once on the same parameter. This PR is essentially a cherry pick of a commit that was previously [reverted](https://github.com/cheekiatng/titanium_mobile/commit/8caea4da94f6c0fac48686f9fe38c7a5c8dcdf45).


garymathews:
FR: **PASS**

PR #6853, opened 2015-05-15
[TIMOB-18765] Only escape if it is a local path and has spaces 

vishalduggal:
Cherry Pick PR #6852 into 4_0_X


garymathews:
FR: **PASS**

PR #6855, opened 2015-05-16
[CLI-629] Fix invalid 'titanium' path for Windows 4_0_X

garymathews:
- Rather than discard the node path using _slice()_, instead _pop()_ the titanium path we want
- This prevents paths with spaces being incorrectly split and discarded

[JIRA Ticket](https://jira.appcelerator.org/browse/CLI-629)


ingo:
Approved. Merging.

PR #6857, opened 2015-05-17
[CLI-629] Fix invalid 'titanium' path for Windows

pinnamur:
- In case if there is any space in the titanium path, then this will handle to get the entire path after we split by spaces.


muhammaddadu:
CR, FT: **PASSED**

**APPROVED**

PR #6860, opened 2015-05-18
[TIMOB-18915] Fixed bug where archived appstore builds were being add…

cb1kenobi:
…ed to 'other items' instead of 'ios apps'.


garymathews:
FR: **PASS**

Gerst20051:
:+1:

PR #6864, opened 2015-05-20
Fix reference to unexisting "ti sdk update"

FokkeZB:


hansemannn:
@FokkeZB Still existing, right? Can you update the branch and we get this in?

FokkeZB:
Took some fiddling, but should be up to date now ;)

PR #6873, opened 2015-05-26
[CLI-658] Encryption fails on Windows when using 64-bit Java

feons:
https://jira.appcelerator.org/browse/CLI-658


pinnamur:
FR and CR passed. I have re-initiated travis build. Can be merged after the build succeed.

PR #6874, opened 2015-05-26
[CLI-658] Encryption fails on Windows when using 64-bit Java 4_0_X

feons:
https://jira.appcelerator.org/browse/CLI-658


pinnamur:
Merging PR, as CR is passed and @eharrisappcel has verified FR.

The tests are failing, but we will address it laster to give build to QE.

PR #6875, opened 2015-05-27
[TIMOB-16488] iOS: Add FileProtectionKey Feature

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-16488

### DIFF
--- a/iphone/Classes/TiFilesystemFileProxy.m
+++ b/iphone/Classes/TiFilesystemFileProxy.m
@@ -1,6 +1,6 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2009-2014 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-2015 by Appcelerator, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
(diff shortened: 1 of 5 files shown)


iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -153,6 +153,37 @@ -(id)spaceAvailable:(id)args
...
+			return @"IOS_FILE_PROTECTION_COMPLETE";
+		}
+		if ([obj isEqualToString:NSFileProtectionCompleteUnlessOpen]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNLESS_OPEN";
+		}
+		if ([obj isEqualToString:NSFileProtectionCompleteUntilFirstUserAuthentication]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNTIL_FIRST_USER_AUTHENTICATION";
+		}
+	}
+}
```
pec1985: Missing `return` after the ifs this will crash if `obj == nil`

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -153,6 +153,37 @@ -(id)spaceAvailable:(id)args
...
+		}
+		if ([obj isEqualToString:NSFileProtectionComplete]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE";
+		}
+		if ([obj isEqualToString:NSFileProtectionCompleteUnlessOpen]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNLESS_OPEN";
+		}
+		if ([obj isEqualToString:NSFileProtectionCompleteUntilFirstUserAuthentication]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNTIL_FIRST_USER_AUTHENTICATION";
+		}
```
pec1985: You should return the actual constant's here, for example `NSFileProtectionCompleteUntilFirstUserAuthentication`, when checked agains the `FilesystemModule`, we need to make sure they match.

`Ti.Filesystem. IOS_FILE_PROTECTION_COMPLETE_UNTIL_FIRST_USER_AUTHENTICATION` will return `NSFileProtectionCompleteUntilFirstUserAuthentication`'s value

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -153,6 +153,37 @@ -(id)spaceAvailable:(id)args
 	return [resultDict objectForKey:NSFileSystemFreeSize];
 }
 
+-(id)getProtectionKey:(id)args
+{
+	NSError *error = nil;
+	NSDictionary * resultDict = [fm attributesOfItemAtPath:path error:&error];
+	if (error!=nil) return NUMBOOL(NO);
```
pec1985: If there's an error, maybe we should print it out to the console. Also, document that if this fails, it will return `false` (or `0`)

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -153,6 +153,37 @@ -(id)spaceAvailable:(id)args
...
+		}
+	}
+}
+
+-(id)setProtectionKey:(id)args
+{
+	ENSURE_SINGLE_ARG(args, NSString);
+	NSError *error = nil;
+	BOOL result = [fm setAttributes:[NSDictionary dictionaryWithObjectsAndKeys:args, NSFileProtectionKey, nil] ofItemAtPath:path error:&error];
+	if (error!=nil) return NUMBOOL(NO);
```
pec1985: Print error (if any) to the console

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -153,6 +153,37 @@ -(id)spaceAvailable:(id)args
...
+		if ([obj isEqualToString:NSFileProtectionCompleteUnlessOpen]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNLESS_OPEN";
+		}
+		if ([obj isEqualToString:NSFileProtectionCompleteUntilFirstUserAuthentication]) {
+			return @"IOS_FILE_PROTECTION_COMPLETE_UNTIL_FIRST_USER_AUTHENTICATION";
+		}
+	}
+}
+
+-(id)setProtectionKey:(id)args
```
pec1985: If the return type is know (in this case `NSNumber*`), declare the function as such:
`-(NSNumber*)setProtectionKey:(id)args`

PR #6879, opened 2015-05-28
[TIMOB-15468] iOS: Add error event for AudioPlayer

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-15468

### DIFF
--- a/iphone/Classes/TiMediaAudioPlayerProxy.m
+++ b/iphone/Classes/TiMediaAudioPlayerProxy.m
@@ -353,8 +353,10 @@ MAKE_SYSTEM_PROP(STATE_PAUSED,AS_PAUSED);
 
 -(void)errorReceived:(id)sender
 {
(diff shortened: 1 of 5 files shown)


iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -351,6 +351,11 @@ - (void)updateProgress:(NSTimer *)updatedTimer
 	}
 }
 
+-(void)errorReceived:(id)sender
+{
+	NSDictionary *event = [TiUtils dictionaryWithCode:player.errorCode message:[AudioStreamer stringForErrorCode:player.errorCode]];
+	[self fireEvent:@"error" withObject:event];
```
pec1985: Check if event exists before firing it.
```
if ([self _hasListeners:@"error"])
```

pec1985:
We should also remove the `presentAlertWithTitle:message:` method since it is not used after this PR, and it's very old code, and we shouldn't be showing alerts on errors anyway.

PR #6894, opened 2015-06-07
Fix non-existing variable

FokkeZB:
`id` does not exist, `module.manifest.moduleid` does.


hansemannn:
Thanks!

FokkeZB:
Cleaning up? Thanks!

PR #6905, opened 2015-06-11
[CLI-556] Alert user if Android SDK version is below supported minimum 4_0_X

feons:
https://jira.appcelerator.org/browse/CLI-556


eharrisappcel:
**FR passed**

Uninstalled Android SDKs so that only API Level 19 was on my system. When trying to build with no targetSK set in the tiapp.xml the user is shown *Unable to find a suitable installed Android SDK that is >=21 and <=22* and the build errors out.

PR #6910, opened 2015-06-14
Timob 19015 - Add isIOS9OrGreater

benbahrenburg:
Add isIOS9OrGreater util method to determine if we are running on iOS9 or greater

Jira Ticket: https://jira.appcelerator.org/browse/TIMOB-19015

Tested : iOS 8 / iOS 9


cheekiatng:
Approved!

PR #6922, opened 2015-06-18
Load remote background image in view

ludolphus:
This should solve jira ticket [iOS: backgroundImage of view from remote URL not loading](https://jira.appcelerator.org/browse/TIMOB-13986)


hansemannn:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/7644, doing it like the TiUIImageView does it.

PR #6921, opened 2015-06-17
[TIMOB-18851]: Implement the filtering of analytics events for Android.

hieupham007:
https://jira.appcelerator.org/browse/TIMOB-18851

PR #6923, opened 2015-06-18
[TIMOB-18851] Implement the filtering of analytics events for iOS

pec1985:
PR follows #6921

https://jira.appcelerator.org/browse/TIMOB-18851

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -958,7 +958,7 @@ MAKE_SYSTEM_PROP(ACTIVITYTYPE_OTHER_NAVIGATION, CLActivityTypeOtherNavigation);
 
 -(void)fireApplicationAnalyticsIfNeeded:(NSArray *)locations
 {
-    if ([AnalyticsModule isEventFltered:@"ti.geo"]) {
+    if ([AnalyticsModule isEventFiltered:@"ti.geo"]) {
         return;
     }
     static BOOL analyticsSend = NO;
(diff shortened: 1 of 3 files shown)


iphone/Classes/GeolocationModule.m
```diff
@@ -955,7 +956,11 @@ -(void)setPurpose:(NSString *)reason
 
 #pragma mark Geolacation Analytics
 
--(void)fireApplicationAnalyticsIfNeeded:(NSArray *)locations{
+-(void)fireApplicationAnalyticsIfNeeded:(NSArray *)locations
+{
+    if ([AnalyticsModule isEventFltered:@"ti.geo"]) {
+        return;
+    }
     static BOOL analyticsSend = NO;
```
cheekiatng: Was trying to test and realised ti.geo is never fired anyway, both with and without the new code. have to remove 'static' here, otherwise analyticsSend will always be set to YES.

pec1985: I don't know why that's there, it was there already. Let me do some research and find out

lokeshchdhry:
**PR Approved**
Tested using the custom 4.1.0 build given by @pec1985 .
As of now only `ti.geo` events are filtered by using Ti.Analytics.filterEvents(['ti.geo']) in the app.
Checked the following:
1. We get `ti.geo` events with out the filterEvents method -- Passed. We got `ti.geo` as expected.
2. We get `ti.geo` events with the filterEvents method -- Passed. We did not get `ti.geo` which is expected as they are filtered.
3. Any other events other than `ti.geo` are filtered -- Passed. No other events are filtered apart from ti.geo.

cheekiatng:
Approved! Merging.

PR #6927, opened 2015-06-21
TIMOB-19003: Remove watchkit openParentApp

benbahrenburg:
This is no longer supported in iOS9, so we will use watch kit instead

Jira #: https://jira.appcelerator.org/browse/TIMOB-19003


cheekiatng:
@benbahrenburg, unfortunately for consistency, we have to do deprecation instead of complete removal. instead of removing, we have to put a warning to inform the developers to move to watch kit connectivity instead.

PR #6944, opened 2015-06-30
Update name and osname for Windows

FokkeZB:
Based on:
https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/Ti/src/Platform.cpp


hansemannn:
Approved!

PR #6945, opened 2015-07-01
[TIMOB-19002] iOS9: Update Ti.Contacts to support new API

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19002

### DIFF
--- a/apidoc/Titanium/Contacts/Contacts.yml
+++ b/apidoc/Titanium/Contacts/Contacts.yml
@@ -137,6 +137,7 @@ methods:
       - name: id
         summary: Contact identifier.
         type: Number
+    since: "4.3.0"      
     osver: {ios: {min: "9.0"}}
 
   - name: removeGroup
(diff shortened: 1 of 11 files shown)


hansemannn:
Results:
- The app crashes, when you try to remove a contact which has already been removed.
- SortedMembers (by Ti.Contacts.CONTACTS_SORT_FIRST_NAME and Ti.Contacts.CONTACTS_SORT_LAST_NAME) is not working as mentioned before.

@cheekiatng Can you check that?

cheekiatng:
Yes the app crashes when trying to remove already removed contact. We have stated that in the docs here http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Contacts-method-removePerson that this will result in unexpected behaviors. And unfortunately i couldn't figure a way to make it not crash either.

hansemannn:
Crashs resolved, can be merged.

PR #6948, opened 2015-07-01
Android: Removed unnecessary cast in onPageFinished of TiWebViewClient

PSanetra:
Hi,

sdk 4.0.0 introduced a breaking change, caused by an unnecessary cast in the onPageFinished method in the TiWebViewClient class. For example that change caused the following issue: https://github.com/viezel/NappJockey/issues/8

This PR removes that cast.

PSanetra


ashcoding:
@PSanetra, thanks for the PR! Is there an example code to show the crash?

hansemannn:
Closing this PR, since there haven't been any response.

PR #6957, opened 2015-07-06
TIMOB-18098: check putExtra param for null

skypanther:
https://jira.appcelerator.org/browse/TIMOB-18098

This change probably doesn't address the original condition reported by this ticket. However, it did resolve some related problems for our app.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/IntentProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/IntentProxy.java
@@ -176,6 +176,9 @@ public class IntentProxy extends KrollProxy
 	@Kroll.method
 	public void putExtra(String key, Object value)
 	{
+		if (value == null) {
+			return;
+		}
 		if (value instanceof String) {
 			intent.putExtra(key, (String) value);
 		} else if (value instanceof Boolean) {


hansemannn:
@skypanther Can you explain what exact problems it fixes? The null-check looks reasonable, but want to make sure we think the same.

skypanther:
We were getting the error I noted in https://jira.appcelerator.org/browse/TIMOB-18098?focusedCommentId=355123&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-355123 when users would tap on a push notification. The "UPDATE 15/09/2015" comments at http://iamyellow.net/post/40100981563/gcm-appcelerator-titanium-module gives additional background on the problem. We ended up using a custom build of the 3.5.1 SDK with my change here integrated and that resolved most of our crashes.

We have since switched GCM modules because the author of the iamyellow module has stopped maintaining his module. The new one we're using doesn't have the same issues because it fires the intents differently. However, I don't think checking for null is an unreasonable thing to do.

ashcoding:
This fix by @skypanther was already merged here: https://github.com/appcelerator/titanium_mobile/commit/42fc687f64f234447e816098f880ab6050f8a925

Will be closing this PR.

PR #6960, opened 2015-07-08
Revert "[TIMOB-19002] iOS9: Update Ti.Contacts to support new API"

pec1985:
Reverts appcelerator/titanium_mobile#6945


pec1985:
We'll need to wait for Xcode 7 to be a requirement (not optional) or place some pre-compile `ifdef`s
screen shot 2015-07-08 at 10 18 59 am

cb1kenobi:
Killing this PR as it is not the right thing to do.

PR #6968, opened 2015-07-11
iOS: Add recursive cell dequeue notifier

animecyc:
Allows for module maintainers to know when a cell is dequeued and respond accordingly.


animecyc:
Will be opening another PR in a bit, cleaning up my fork

PR #6971, opened 2015-07-13
iOS: Add recursive cell dequeue notifier

animecyc:
Allows for module maintainers to know when a cell is dequeued and respond accordingly.


pec1985:
Neither the `ListView` nor the `TableView` are part of the public module API.
PR rejected.

PR #6972, opened 2015-07-13
iOS: TiUIListView does not resume the ImageLoader

animecyc:
If a user scrolls a list and does not _fling_ the table the shared image loader does not decelerate thus causing the image loader to remain suspended.


hansemannn:
Hi @animecyc, thank you for providing the PR! We will likely use it as part of #7320

PR #6973, opened 2015-07-14
[TIDOC-2255] fix the method returns correct value in api.json

k0sukey:
JIRA https://jira.appcelerator.org/browse/TIDOC-2255


hansemannn:
Was fixed as part of https://github.com/appcelerator/titanium_mobile/pull/6974, thanks!

PR #6975, opened 2015-07-15
[TIMOB-19211] iOS: Tapping "undo" will get a range or index out of bounds on TiUITextField

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19211
Only merge after 4.2.0


hansemannn:
CR passed and approved!

PR #6978, opened 2015-07-19
Update AppModule.m

jyounus:
Removed annoying and unnecessary debug log.
If developers want to print out logs, they can manually do that in JS. No need to force this inside of the SDK. During development, if you have an app-wide fireEvent running after every x interval, the console output becomes really polluted. Android doesn't do this, would be nice if iOS wouldn't do this either.


hansemannn:
As the log is only on the debug level, we would like to remain it there. You can change your log level inside the studio or using ``--log-level info`` in the CLI. Thanks!

PR #6981, opened 2015-07-22
[TC-5633] iOS: Fix crash with an empty list view and a register event…

wachri:
Fixes a issues described in https://jira.appcelerator.org/browse/TC-5633


hansemannn:
We should also update the docs to mention the indexes "-1" when there are no items.

hansemannn:
Code reviewed. Functionally tested. **Accepted**.

Cherry-picked it and committed one PR to update docs and fire the event. **New PR: #7046**

Closing this PR in favor of that. Thank you @wachri!

PR #6983, opened 2015-07-24
[TIMOB-19214][TIMOB-19209][TIMOB-19249] iOS Fixed race condition

pec1985:
Jira Tickets:
[TIMOB-19214](https://jira.appcelerator.org/browse/TIMOB-19214)
[TIMOB-19209](https://jira.appcelerator.org/browse/TIMOB-19209)
[TIMOB-19249](https://jira.appcelerator.org/browse/TIMOB-19249)

PR #6984, opened 2015-07-24
[TIMOB-19214][TIMOB-19209][TIMOB-19249] iOS Fixed race condition

pec1985:
Back port from #6983


cheekiatng:
CR and FT passed. *APPROVED!*

PR #6990, opened 2015-07-29
[TIMOB-19274] CLI: Fix for when building for iOS device the app will always be 32-bit

cheekiatng:
JIRA Here: https://jira.appcelerator.org/browse/TIMOB-19274


cb1kenobi:
**APPROVED**

PR #6994, opened 2015-07-29
[TIMOB-19265] Fixed sort order of simulators so that it sensibly auto…

cb1kenobi:
…selects a simulator.


cb1kenobi:
**APPROVED**

PR #6995, opened 2015-07-29
[TIMOB-19268] Xcode 7 - Fix app crashes on launch if app is installed on an iOS device < 9.0

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19268


cb1kenobi:
Tested and works like a champ! **APPROVED**

PR #7011, opened 2015-08-06
[TIMOB-19320] Renamed the Apple Watch template so that we can actuall…

cb1kenobi:
…y create Apple Watch apps. Some minor tweaks. Added team support as apart of TIMOB-19311.


cb1kenobi:
**APPROVED**

PR #7013, opened 2015-08-07
[TIMOB-19311] Added sqlite3 to main package.json so that we could eas…

cb1kenobi:
…ily inject the pre-built binaries.


cb1kenobi:
**APPROVED**

PR #7047, opened 2015-08-19
[TIMOB-12246] Fixed TextField with KEYBOARD_DECIMAL_PAD doesn't allow…

collinprice:
… decimal point.


collinprice:
Oops. Not actually a fix.

PR #7050, opened 2015-08-19
[TIMOB-19314]: Update map module

hieupham007:


lokeshchdhry:
Verified the updated map module has 2 jar files : google play services base jar & google play services map jar. 
No issues found while building with the module.

PR #7058, opened 2015-08-21
[TIMOB-19381] Fixed bug with resolving relative paths to extensions. …

cb1kenobi:
…Also fixed small bug where the context was not being passed in when processing Info.plist UIAppFonts.


cb1kenobi:
**APPROVED**

PR #7060, opened 2015-08-21
[TIMOB-18199] iOS: iOS modules can not build if there any JS files inside assets folder

feons:
https://jira.appcelerator.org/browse/TIMOB-18199


cheekiatng:
APPROVED.

PR #7070, opened 2015-08-25
[TIMOB-19405] Fixed bug where iPhone specific app icons and launch im…

cb1kenobi:
…ages were being required when building iPad-only apps (and vice versa).


cb1kenobi:
**APPROVED**

PR #7092, opened 2015-08-30
TIMOB: 19026 Fix Init Problem

benbahrenburg:
Fix problem where progress view color is not applied on init in some cases.


iphone/Classes/TiUIActivityIndicatorProxy.m
```diff
@@ -11,6 +11,18 @@
 
 @implementation TiUIActivityIndicatorProxy
 
+-(NSArray *)keySequence
+{
+    return [NSArray arrayWithObjects:
+            @"visibile",
```
cheekiatng: visible

cheekiatng:
Approved!

PR #7099, opened 2015-08-31
[TIMOB-19424] Fixed bug where app icons and launch images were gettin…

cb1kenobi:
…g lost during differential builds. Also fixed bug where build-only simulator builds were actually doing device builds.


cb1kenobi:
**APPROVED**

PR #7108, opened 2015-09-01
[TIMOB-19447] CommonJS modules are not copied to the correct location when require encryption

feons:
https://jira.appcelerator.org/browse/TIMOB-19447


cb1kenobi:
Looks great! **APPROVED**

PR #7110, opened 2015-09-01
CLI: Fixed default flags for running on main thread and and using JSCore

pec1985:


iphone/cli/commands/_build.js
```diff
@@ -2051,9 +2051,9 @@ iOSBuilder.prototype.initialize = function initialize() {
 	// This is default behavior for now. Move this to true in phase 2.
 	// Remove the debugHost/profilerHost check when we have debugging/profiling support with JSCore framework
 	// TIMOB-17892
-	this.currentBuildManifest.useJSCore = this.useJSCore = !this.debugHost && !this.profilerHost && this.tiapp.ios['use-jscore-framework'];
+	this.currentBuildManifest.useJSCore = this.useJSCore = !this.debugHost && !this.profilerHost && (this.tiapp.ios['use-jscore-framework'] || false);
```
cb1kenobi: If all you're looking for is truthiness, then you don't need to change the code. If you need a boolean, then do `!!this.tiapp.ios['use-jscore-framework']` so that truthy === true and falsey === false.

cheekiatng:
APPROVED!

PR #7113, opened 2015-09-01
[CLI] fix max call stack when iterating through many js files

pec1985:
Wrapped some code in a `setImmediate` to avoid max call stack. When having too many files this message appears and the build fails

```
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.

/usr/local/lib/node_modules/titanium/node_modules/longjohn/dist/longjohn.js:194
        throw e;
              ^
RangeError: Maximum call stack size exceeded
```

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -4493,52 +4493,50 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			this.logger.info(__('Processing JavaScript files'));
 
 			async.eachSeries(Object.keys(jsFiles), function (file, next) {
-				var info = jsFiles[file];
-				if (this.encryptJS) {
-					if (file.indexOf('/') === 0) {
-						file = path.basename(file);
+				setImmediate(function(){
+					var info = jsFiles[file];
+					if (this.encryptJS) {
+						file = file.replace(/\./g, '_');
+						info.dest = path.join(this.buildAssetsDir, file);
+						this.jsFilesToEncrypt.push(file);
 					}
-					file = file.replace(/\./g, '_');
-					info.dest = path.join(this.buildAssetsDir, file);
-					this.jsFilesToEncrypt.push(file);
-				}
+					this.cli.createHook('build.ios.copyResource', this, function (from, to, cb) {
+						try {
+							// parse the AST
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -4493,52 +4493,50 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			this.logger.info(__('Processing JavaScript files'));
 
 			async.eachSeries(Object.keys(jsFiles), function (file, next) {
-				var info = jsFiles[file];
-				if (this.encryptJS) {
-					if (file.indexOf('/') === 0) {
```
cb1kenobi: We need this line. Please restore it.

pec1985: Not sure what happened there

pec1985:
Take a look at the changes and then I'll indent everything

iphone/cli/commands/_build.js
```diff
@@ -4493,6 +4493,7 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			this.logger.info(__('Processing JavaScript files'));
 
 			async.eachSeries(Object.keys(jsFiles), function (file, next) {
+				setImmediate(function(){
```
cb1kenobi: Last thing, then you can indent. We do spaces between `function` and `()` and `{`. The space between `function` and `(` is because we always put a space after `function` regardless if the function is named or anonymous. Since functions are no different than other control blocks such as `if` and `while`, we do a space between the `)` and `{`.

pec1985:
done

cb1kenobi:
Looks good. **APPROVED**.

@pec1985 do you have a backport too?

pec1985:
@cb1kenobi Here it is https://github.com/appcelerator/titanium_mobile/pull/7124

ingo:
@pec1985 I need a JIRA ticket for this.

PR #7117, opened 2015-09-02
Fix type of setNeedsSave arg

FokkeZB:
Should be NSNumber instead of NSString since it's a boolean

Backport of https://github.com/appcelerator/titanium_mobile/pull/7116


cheekiatng:
APPROVED!

PR #7120, opened 2015-09-02
[TIMOB-19184] iOS9: Expose new iOS9 "behavior" property

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19184

### DIFF
--- a/apidoc/Titanium/App/iOS/UserNotificationAction.yml
+++ b/apidoc/Titanium/App/iOS/UserNotificationAction.yml
@@ -27,6 +27,18 @@ properties:
     type: Number
     availability: creation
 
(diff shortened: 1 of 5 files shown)


cheekiatng:
Since it's 2 constants, maybe we should consider, instead of asking developers to set a constant for the behavior property, we can have a BOOL property, named 'allowsLockscreenReply'

hansemannn:
Thought about it and am not sure, if there might come some different behaviors in the feature that will require to use constants. Therefore, I preferred a "future-save" implementation. Not sure whats best here.

cheekiatng:
Approved!

PR #7124, opened 2015-09-04
[CLI] Backport - fix max call stack when iterating through many js files

pec1985:
Backport for PR #7113


cb1kenobi:
**APPROVED**!

PR #7126, opened 2015-09-04
[TIMOB-19469] Fixed bug when writing an entitlements file but the par…

cb1kenobi:
…ent diretory doesn't exist.


cheekiatng:
APPROVED!

PR #7129, opened 2015-09-07
[TIMOB-17595] iOS: The app fails to install on iOS device after a second iOS device is also plugged in

FokkeZB:
**NOTE:** This is not a fix but an improvement of the error message

https://jira.appcelerator.org/browse/TIMOB-17595


hansemannn:
I think @cb1kenobi is already working on a fix on the CLI side to make this message obsolet.

FokkeZB:
He asked me for a PR (see JIRA) but a full fix is always better ;)

hansemannn:
Great! I would prefer checking against the error code (0xe800007f) rather then the else-case for this. But Chris will review I think.

cb1kenobi:
I'm fine with the messaging. If we don't handle a specific error code, then we have no idea what's up and the only thing we can recommend is to reconnect and try again.

**APPROVED**

@FokkeZB do you have a 5_0_X backport?

FokkeZB:
@cb1kenobi I have now:
https://github.com/appcelerator/titanium_mobile/pull/7143

PR #7130, opened 2015-09-07
Remove glance and notification scenes and controllers

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-19472


FokkeZB:
Outdated

PR #7134, opened 2015-09-08
[TIMOB-19435] Change Pref Activity to Pref Frag

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-19435


android/modules/ui/src/java/ti/modules/titanium/ui/android/TiPreferencesFragment.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    private static final String TAG = "TiPreferencesFragment";
+    private static final String DEFAULT_PREFS_RNAME = "preferences";
+    
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        
+        String prefsName = DEFAULT_PREFS_RNAME;
+        if (getActivity().getIntent().hasExtra("prefsName")) {
```
hieupham007: what if getActivity() or getIntent() is null?

android/modules/ui/src/java/ti/modules/titanium/ui/android/TiPreferencesFragment.java
```diff
@@ -0,0 +1,42 @@
...
+            String name = getActivity().getIntent().getExtras().getString("prefsName");
```
hieupham007: please try to avoid duplicate method calling: getActivity().getIntent() is called twice back to back

ashcoding:
Addressed the comments. I'm putting the logic to get the pref name in the activity and passing it to the fragment as an argument via a bundle.

I think this might be better.

hieupham007:
This is what I see: 
screenshot_2015-09-16-14-46-05

Is this expected?

ashcoding:
That is not the expected outcome. May I know what device you are using and the android version?

hieupham007:
Galaxy Nexus 4.1.2 and Android M Preview 3. Same results.

ashcoding:
Not sure why but it's working correctly for me. I pulled and merge master to make this updated with master. Could you try it again?

hieupham007:
I'm seeing this now: 
screenshot_2015-09-17-13-42-09

ashcoding:
I'm assuming the title `215` is not intended? This seems quite strange. I'll look into it more.

ashcoding:
Updated the PR. The title is now taken from the first Preference Screen title. Eg, this will be `TITLE` in the following example.
```xml
<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android" android:title="TITLE">
    <CheckBoxPreference
        android:key="key1"
        android:title="this is key 1"
        android:summary="key 1 summary"
        android:defaultValue="false" />
    <CheckBoxPreference
        android:key="key2"
        android:title="this is key 2"
        android:summary="key 2 summary"
        android:defaultValue="false" />
</PreferenceScreen>
```

hieupham007:

PR #7135, opened 2015-09-08
[TIMOB-19473] iOS9: Fix memory issues in Contacts and Geolocation Modules

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19473


hansemannn:
CR & FT **approved**!

PR #7137, opened 2015-09-08
[TIMOB-19154] iOS9: Deprecate NSURLConnection and replace with NSURLSession API

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19154


hansemannn:
@cheekiatng I ran the tests you provided and got some errors:
- "TLS v1.0": FAILS - Returned TLS 1.2 instead
- JSON-Multipart: FAILS - [__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[2]
- TIMOB-15380: FAILS - Server not found (the url provided in the ticket just does not work anymore, not our bug)

Can you check the first 2 list points before I merge? Want to make sure not to rush. Thank you!

cheekiatng:
@hansemannn oh ya i forgot to mention those 3 are expected to fail, the tests were extracted from android implementation so some aren't supposed to work on iOS:
1. iOS9 support only TLS 1.2
2. If you look into the JS code, there's a missing token string. It needs that string to run, you can get it from the corresponding jira ticket.
3. As you say, not out bug.

hansemannn:
Approved!

PR #7139, opened 2015-09-08
[TIMOB-19475] Android: Build fails if target SDK is lower than API 23

feons:
https://jira.appcelerator.org/browse/TIMOB-19475

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -16,7 +16,7 @@
 	],
 	"minSDKVersion": "14",
 	"vendorDependencies": {
-		"android sdk": ">=21 <=23",
+		"android sdk": "23",
 		"android build tools": ">=17 <23.x",
 		"android platform tools": ">=17 <=23.x",
 		"android tools": "<=24.3.x",


android/cli/commands/_build.js
```diff
@@ -1102,7 +1102,7 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 		// target sdk is too old
 		if (this.realTargetSDK < this.minTargetApiLevel) {
 			logger.error(__('The target SDK %s is not supported by Titanium SDK %s', this.targetSDK + (this.targetSDK !== this.realTargetSDK ? ' (' + this.realTargetSDK + ')' : ''), ti.manifest.version));
-			logger.error(__('The target SDK version must be %s or newer', this.minTargetApiLevel) + '\n');
+			logger.error(__('The target SDK version must be %s', this.minTargetApiLevel) + '\n');
```
cb1kenobi: I'm not sure about this. So, yes we only support api 23, but if the target sdk is greater than 23, we don't really error out... only if it's less than 23. So, it should be safe to leave the "or newer".

cb1kenobi:
**APPROVED**

PR #7141, opened 2015-09-08
[TIMOB-19465] Fixed bug where CleanCSS was stripping @import statemen…

cb1kenobi:
…ts from the minified CSS files.


cb1kenobi:
**APPROVED**

PR #7143, opened 2015-09-08
[TIMOB-17595] iOS: The app fails to install on iOS device after a second iOS device is also plugged in

FokkeZB:
NOTE: This is not a fix but an improvement of the error message

https://jira.appcelerator.org/browse/TIMOB-17595

Backport of https://github.com/appcelerator/titanium_mobile/pull/7129


cb1kenobi:
**APPROVED**

PR #7144, opened 2015-09-08
[TIMOB-19475] Android: Build fails if target SDK is lower than API 23 backport

feons:
https://jira.appcelerator.org/browse/TIMOB-19475


cb1kenobi:
**APPROVED**

PR #7157, opened 2015-09-10
[TIMOB-19477] Added support for DefaultIcon.png

cb1kenobi:


cheekiatng:
APPROVED!

node_modules/titanium-sdk/lib/builder.js
```diff
@@ -373,3 +374,99 @@ Builder.prototype.validateTiModules = function validateTiModules(platformName, d
...
 };
+
+/**
+ * Generates missing app icons based on the DefaultIcon.png.
+ *
+ * @param {Array<Object>} icons - An array of objects describing the icon size to generate and the destination
+ * @param {Function} callback - A function to call after the icons have been generated
+ */
+Builder.prototype.generateAppIcons = function generateAppIcons(icons, callback) {
+	var defaultIcon = path.join(this.projectDir, 'DefaultIcon.png'),
```
FokkeZB: Any reason we have this in the project root instead of images? Is that because we don't want the actual `DefaultIcon.png` in the IPA? Could we still use platform specific icons once we also bring this generating of icons to Android and other platforms?

PR #7173, opened 2015-09-12
Parse windows tag in tiapp.xml, fix for TIMOB-19178

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19178


cb1kenobi:
CR'd. I'm good with this. @ingo is good with this. **APPROVED**!

PR #7179, opened 2015-09-14
[TIMOB-19446] iOS9: Support block callback on watchOS 2

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19446


iphone/Classes/WatchSessionModule.m
```diff
@@ -108,8 +153,20 @@ -(void)sendMessage:(id)value
         DebugLog(@"[ERROR] No watch paired");
         return;
     }
-    ENSURE_SINGLE_ARG(value, NSDictionary)
-    [[self watchSession] sendMessage:value replyHandler:nil errorHandler:nil];
+    ENSURE_SINGLE_ARG(args, NSDictionary)
+    NSDictionary *value = args;
+    KrollCallback *replyHandler = [value objectForKey:@"onReply"];
```
hansemannn: Can we rename it to `reply` to keep the naming conventions like `success`, `error`and `cancel` callbacks?

hansemannn:
Callbacks tested, but receiving the following error using the simulator:
`[ERROR] :  [TestApp WatchApp] (Error) WatchKit: <SPApplicationDelegate.m __55-[SPApplicationDelegate _startActivatingAssertionTimer]_block_invoke:2063> We've held the CSLSActivatingUIAssertion for 30 secs. Giving up`

As I said in the related ticket, there might be a problem that the native Xcode 7 apps do not work in the simulator, too. Is that is true, we should provide a better error message to the user that they should test it using a device.

cheekiatng:
Addressed comment about `onReply` to `reply`
However, I can't reproduce the error you saw. Did you try with my sample app here?
https://github.com/cheekiatng/Titanium-WatchOS2-Sample-App

hansemannn:
Will test again using your sample app

hansemannn:
CR & FT done, works perfectly. Thanks @cheekiatng!

PR #7182, opened 2015-09-14
[TIMOB-19513](5_0_X) Default UIRequiresFullScreen to true. You can still ove…

cb1kenobi:
…rride it with a custom Info.plist or the <ios><plist> in the tiapp.xml.


hansemannn:
APPROVED!

PR #7187, opened 2015-09-15
[TIMOB-19518] Updated to node-ios-device v0.7.0 which no longer unins…

cb1kenobi:
…talls the app prior to installing the new app for iOS Simulator builds.


cheekiatng:
APPROVED!

PR #7206, opened 2015-09-18
[TIMOB-19549] Fixed bug where minified JS files were overriding the o…

cb1kenobi:
…riginals in the Resources dir for device and dist builds.


cheekiatng:
APPROVED.

PR #7225, opened 2015-09-23
[TIMOB-19562](5_0_X) Fixed iOS 9 crash due to Ti.Network.Socket.TCP#close() call

prop:


hansemannn:
Approved! Thank you @prop!

PR #7241, opened 2015-09-29
[TIMOB-19612] Fixed a bunch of bugs with differential builds. Specifi…

cb1kenobi:
…cally with ApplicationRouting.m, defines.h, resized app icons, and static titanium libs. Fixed bug where Xcode generated files were being deleted accidentially. Fixed some unnecessary file copying. Fixed regression where subsequent builds had Xcode compiling all files.


cheekiatng:
CR and FT passed. APPROVED.

PR #7244, opened 2015-09-29
[TIMOB-19566]-Ti.Locale.currentLanguage returns locale on iOS 9

AngelkPetkov:
JIRA:https://jira.appcelerator.org/browse/TIMOB-19566?filter=-1
Simply just truncated the string to 2 characters.


iphone/Classes/TiLocale.m
```diff
@@ -44,6 +44,11 @@ +(NSString*)defaultLocale
 +(void)setLocale:(NSString*)locale
 {
 	TiLocale *l = [TiLocale instance];
+	NSString *var = [[UIDevice currentDevice] systemVersion];
+	if (var.floatValue >= 9.0) {
+		//[TIMOB-19566]:Truncate the current locale for parity between iOS versions
+		locale =[locale substringToIndex:2];
+	}
 	l.currentLocale = locale;
 	NSString *path = [[ NSBundle mainBundle ] pathForResource:locale ofType:@"lproj" ];
```
cheekiatng: 'locale' is used here for referencing a path, so i'm not sure it's a good idea to change this variable, since the ticket wants to address currentLanguage only. Considering the return value of [TiLocal defaultLocale] is only used for getCurrentLanguage, it maybe a better idea to change the make this correction here instead https://github.com/AngelkPetkov/titanium_mobile/blob/TIMOB-19566-currentLanguage/iphone/Classes/TiLocale.m#L41

cheekiatng:
Please address comments and maybe squash the commits if you can? Thanks!

PR #7245, opened 2015-09-29
Fix TIMOB-19648 - Fix Android SDK compilation against JDK 8

sgtcoolguy:
Use [android.util.Base64](http://developer.android.com/reference/android/util/Base64.html) class in place of the commons codec Base64 implementation which won't compile on JDK 8. Should be ok since the android version was added in API level 8 - a long time ago.

https://jira.appcelerator.org/browse/TIMOB-19648

### DIFF
--- a/android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
+++ b/android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
@@ -14,7 +14,6 @@ import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
-import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.appcelerator.kroll.KrollModule;
 import org.appcelerator.kroll.annotations.Kroll;
@@ -23,6 +22,8 @@ import org.appcelerator.titanium.TiBlob;
 import org.appcelerator.titanium.TiContext;
 import org.appcelerator.titanium.TiFileProxy;
 import org.appcelerator.titanium.util.TiMimeTypeHelper;
+
+import android.util.Base64;
 
 @Kroll.module
 public class UtilsModule extends KrollModule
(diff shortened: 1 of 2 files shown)


android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
```diff
@@ -66,7 +66,7 @@ public TiBlob base64encode(Object obj)
 		String data = convertToString(obj);
 		if (data != null) {
 			try {
-				return TiBlob.blobFromString(new String(Base64.encodeBase64(data.getBytes("UTF-8")), "UTF-8"));
+				return TiBlob.blobFromString(new String(Base64.encode(data.getBytes("UTF-8"), Base64.DEFAULT), "UTF-8"));
```
ashcoding: I did a check on this with the following code
```java
				//Current
				Log.d("LogCheck1", new String(Base64.encode(data.getBytes("UTF-8"), Base64.DEFAULT), "UTF-8"));
				//Old and Correct
				Log.d("LogCheck2", new String(org.apache.commons.codec.binary.Base64.encodeBase64(data.getBytes("UTF-8")), "UTF-8"));
				//New and Correct
				Log.d("LogCheck3", new String(Base64.encode(data.getBytes("UTF-8"), Base64.NO_WRAP), "UTF-8"));
```

Using Base64.DEFAULT produces a slightly different String that include line terminators.

I tested it with this code:
```javascript

var encoded = Ti.Utils.base64encode("Hi there Is There\n /n \n THis is \n /n long long \n   long /n helooooooooo Hi "); 

```

Stackoverflow link that is relevant: http://stackoverflow.com/questions/17912119/is-there-any-difference-between-apaches-base64-encodebase64-and-androids-base6

android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
```diff
@@ -80,7 +80,7 @@ public TiBlob base64decode(Object obj)
 		String data = convertToString(obj);
 		if (data != null) {
 			try {
-				return TiBlob.blobFromData(Base64.decodeBase64(data.getBytes("UTF-8")));
+				return TiBlob.blobFromData(Base64.decode(data.getBytes("UTF-8"), Base64.DEFAULT));
```
ashcoding: See comments above on `Base64.DEFAULT`.

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -562,7 +561,7 @@ public TiFileProxy getFile()
 	@Kroll.method
 	public String toBase64()
 	{
-		return new String(Base64.encodeBase64(getBytes()));
+		return android.util.Base64.encodeToString(getBytes(), android.util.Base64.DEFAULT);
```
ashcoding: See comments above on `Base64.DEFAULT`.

PR #7246, opened 2015-09-29
[TIMOB-19594] iOS: Cannot build iOS Modules since 5.0.0.GA TiSDK

feons:
https://jira.appcelerator.org/browse/TIMOB-19594


hansemannn:
Tested and **approved**, thank you @feons !

PR #7248, opened 2015-09-30
[TIMOB-19296] update all image-related methods to support assets catalog

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19296

### DIFF
--- a/iphone/Classes/ImageLoader.m
+++ b/iphone/Classes/ImageLoader.m
@@ -505,10 +505,10 @@ DEFINE_EXCEPTIONS
 				scaleUp = YES;
 			}
 			UIImage *resultImage = nil;
-			NSRange range = [urlString rangeOfString:@".app"];
+			NSRange range = [path rangeOfString:@".app"];
 			NSString *imageArg = nil;
 			if (range.location != NSNotFound) {
-				imageArg = [urlString substringFromIndex:range.location+5];
+				imageArg = [path substringFromIndex:range.location+5];
 			}
 			if (imageArg != nil) {
 				unsigned char digest[CC_SHA1_DIGEST_LENGTH];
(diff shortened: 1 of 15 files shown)


iphone/Classes/ImageLoader.m
```diff
@@ -504,7 +504,30 @@ -(ImageCacheEntry *)entryForKey:(NSURL *)url
...
-			UIImage * resultImage = [UIImage imageWithContentsOfFile:path];
+			UIImage *resultImage = nil;
+			NSRange range = [urlString rangeOfString:@".app"];
+			NSString *imageArg = nil;
+			if (range.location != NSNotFound) {
+				imageArg = [urlString substringFromIndex:range.location+5];
+			}
+			if (imageArg != nil) {
+				unsigned char digest[CC_SHA1_DIGEST_LENGTH];
+				NSData *stringBytes = [imageArg dataUsingEncoding: NSUTF8StringEncoding];
```
cb1kenobi: Close, so close.

`imageArg` originates from a URL which means it's URL encoded. Spaces become `%20` and such. We need `imageArg` to be URL decoded first.

NSString has a handy dandy function `stringByReplacingPercentEscapesUsingEncoding` that decodes the string, however this function has been deprecated in favor of `stringByRemovingPercentEncoding`.

I may be mistaken, but I think we're going to need to do an `#if` and use the appropriate method:

```objc
#if IS_XCODE_7
				imageArg = [imageArg stringByRemovingPercentEncoding];
#else
				imageArg = [imageArg stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
#endif
```

I've tested the code above by placing it before line 515 and it is able to decode and sha correctly.

cheekiatng: Yikes you are right. A better way to do this is instead of using ```urlString = [url absoluteString]```, we use ```urlString = [url path]``` instead. since path is the recommended by Apple when dealing with local file paths, and it won't convert spaces to %20. PR coming soon.

cb1kenobi:
This PR has made all of my wildest dreams come true! **APPROVED**

PR #7257, opened 2015-10-01
[TIMOB-19615] iOS: Suppress missing UIBackgroundModes "fetch" and "remote-notification" warnings

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19615


iphone/cli/commands/_build.js
```diff
@@ -4945,6 +4945,13 @@ iOSBuilder.prototype.processTiSymbols = function processTiSymbols() {
 		if (this.useJSCore) {
 			contents += '\n#define USE_JSCORE_FRAMEWORK';
 		}
+		var infoPlist = this.infoPlist;
+		if (Array.isArray(infoPlist.UIBackgroundModes) && infoPlist.UIBackgroundModes.indexOf('remote-notification') !== -1) {
+			contents += '\n#define USE_TI_SILENTPUSH';
+		}
+		if (Array.isArray(infoPlist.UIBackgroundModes) && infoPlist.UIBackgroundModes.indexOf('fetch') !== -1) {
+			contents += '\n#define USE_TI_FETCH';
+		}
```
cb1kenobi: This will only add these #defines if you're building for simulator or including all ti modules. That means device and dist builds will never have these set. These new lines should probably be moved after this if/else block and before the `if (!destExists || contents !== fs.readFileSync(dest).toString()) {` line.

cheekiatng: But device builds were already addressed here at this line:
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L4959

cb1kenobi:
CR'd and FR'd. **APPROVED**

PR #7260, opened 2015-10-01
[TIMOB-19658] Removed the JDK 1.6 requirement and wired up the new ti…

cb1kenobi:
…tanium_prep.


sgtcoolguy:
CR and FR passed with both JDK 8 and JDK 6 on my El Capitan laptop.

PR #7268, opened 2015-10-05
[TIMOB-19011] Android: MapView is blank if the container view has rou…

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-19011

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
@@ -1266,9 +1266,8 @@ public abstract class TiUIView
 					if (radiusDim != null) {
 						radius = (float) radiusDim.getPixels(getNativeView());
 					}
-					if (radius > 0f && HONEYCOMB_OR_GREATER) {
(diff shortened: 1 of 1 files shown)


hieupham007:
You are reversing this change: https://jira.appcelerator.org/browse/TIMOB-5461. Have you check for possible side effects?

sriks:
Thats true. But HW acceleration had issues with honeycomb, future versions are good enough. But to fix this issue, map view requires HW acceleration turned on. I checked the view with rounded corners in Android 4.1,4.4, 5.0 and 6.0, but cannot rule out slight chance of unwanted side effects in honeycomb.

Or else can we disable HW acceleration on Honeycomb only?

hieupham007:
Well, we don't support HC anymore, but I wouldn't rule out possible side effects of newer versions unless we run a full suite of tests on this change. The effects may be limited since we only enable HA on one view. If you look at the tables here, some operations are not supported when HA is on: http://developer.android.com/guide/topics/graphics/hardware-accel.html

hieupham007:
Ran a few tests, not seeing any regressions. Merging.

PR #7272, opened 2015-10-07
[TIMOB-19680] Fixed bug where invalid app icons were breaking the bui…

cb1kenobi:
…ld when it should just warn and try to continue.


cheekiatng:
APPROVED.

PR #7275, opened 2015-10-07
TIMOB-19687 Developer Setup Improvements - Work in progress

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19687

This PR is intended to make life easier in getting a working Android SDK dev environment. This still assumes you have Java, Android SDK/NDK, scons and all the SDK build tools already installed. It could certainly be grown to do more of that for the developer (i.e. check for homebrew, grab and install it, use it to install scons/python/ant - should already have git if they have the file...).

Here's what it does so far:
- include the Android formatting prefs directly into the module projects themselves so they're on by default for Eclipse users and developers don't need to hunt down the formatting XML file, and set it as their workspace default.
- It also includes the xml formatting settings in the repo just in case/for completeness sake. 
- It includes a shell script (setup.sh) for Mac users that will download Eclipse 4.5.1, unzip it, install ADT into it, and copy the pre-setup Eclipse into /Applications.
- It adds two eclipse projects that should get imported alongside all the other projects: titanium-dist and TitaniumTest. Their setup is pre-done, but came from using [this wiki page](https://wiki.appcelerator.org/display/pe/Android+Titanium+SDK+Development+Setup) and some trial and error and much gnashing of teeth.


ashcoding:
:+1:

ashcoding:
I think there might be some issues with the settings in eclipse in some project being set to 1.7 JDK instead of 1.6 JDK which results in some projects to say:
kroll-common and kroll-v8 `Using 1.7 requires compiling with Android 4.4 (KitKat)l currently using API 10`

If I'm not wrong, this 2 project in Eclipse should change their Project Build Target to Android 6.0 to allow for usage of 1.7 JDK.

sgtcoolguy:
I don't believe any of the projects actually have prefs checked in regarding the Java version. I think this happens if your workspace is set up for Java 7+. We can certainly check in prefs/settings per-project to a Java 6 SDK to avoid this, though.

ashcoding:
I think that's happening with mine. Thanks for the clarification. :)

PR #7276, opened 2015-10-07
[TIMOB-19510] iOS: Added UserActivity.isSupported()

AngelkPetkov:
:Jira: https://jira.appcelerator.org/browse/TIMOB-19510?filter=-1
Added a isSupported() method for parity reasons

### DIFF
--- a/iphone/Classes/TiAppiOSUserActivityProxy.m
+++ b/iphone/Classes/TiAppiOSUserActivityProxy.m
@@ -184,6 +184,12 @@
     return NUMBOOL(_isValid);
 }
 
+-(NSNumber*)supported
+{
+    DEPRECATED_REPLACED(@"Ti.App.IOS.UserActivity.getSupported()" ,@"5.1.0",@"Ti.App.IOS.UserActivity.isSupported()")
+    return NUMBOOL(_supported);
+}
+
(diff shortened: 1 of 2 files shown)


hansemannn:
The docs are missing. Please add them and I will continue reviewing. Thank you!

iphone/Classes/TiAppiOSUserActivityProxy.m
```diff
@@ -175,11 +184,6 @@ -(NSNumber*)isValid
     return NUMBOOL(_isValid);
 }
 
--(NSNumber*)supported
```
hansemannn: We should not remove the existing ones, since they are already used by developers.

iphone/Classes/TiAppiOSUserActivityProxy.m
```diff
@@ -51,6 +51,15 @@ -(BOOL)activityTypeValid:(NSString*)activityType
...
 }
 
+-(id)isSupported:(id)unused
+{
+    if([TiUtils isIOS8OrGreater]){
+        return NUMBOOL(_supported);
+    }else{
+        return NUMBOOL(NO);
+    }
+}
```
hansemannn: Please make sure to use spaces between `}else{` and `if(`

PR #7280, opened 2015-10-08
[TIMOB-19621]: tiImage() complete callback is not fired on Android

sriks:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19621
The java proxy method was not accepting params. So even though callback is provided in js, the result is retuned synchronously. Added the method params and handing accordingly.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -37,6 +37,8 @@ import org.appcelerator.titanium.view.TiUIView;
 import android.animation.Animator;
 import android.animation.AnimatorListenerAdapter;
 import android.app.Activity;
+import android.graphics.Bitmap;
+import android.graphics.Bitmap.Config;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Message;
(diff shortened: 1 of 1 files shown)


android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -916,11 +916,11 @@ protected void handleFocus()
 	}
 
 	@Kroll.method
-	public TiBlob toImage()
+	public TiBlob toImage(@Kroll.argument(optional=true) KrollFunction callback)
 	{
-		if (TiApplication.isUIThread()) {
-			return handleToImage();
-
+		if (callback != null) {
```
hieupham007: The logic of this if statement is flawed b/c it doesn't take into account TiApplication.isUIThread() clause. Also, you should return the TiBlob, not null even when there's a callback present.

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -916,11 +916,11 @@ protected void handleFocus()
...
+			callback.callAsync(getKrollObject(), new Object[] {handleToImage()});
+			return null;
 		} else {
 			return (TiBlob) TiMessenger.sendBlockingMainMessage(getMainHandler().obtainMessage(MSG_TOIMAGE), getActivity());
```
hieupham007: You would also need to pass the callback here and handle it in handleMessage().

hieupham007:
Code reviewed. Please address comments

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
sriks: Thanks for pointing out. In case callback exists then the rendering should happen in a background thread but the method should return a 'non-null' blob immediately ? The documentation seems misleading about return value.

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -916,14 +918,40 @@ protected void handleFocus()
...
+			}
+		} 
+		
+		/*
+		 * Callback exists. Perform async rendering and return an empty blob.
+		 */
+		else {
+			// Create a non-null empty blob to return. 
+			blob  = TiBlob.blobFromImage(Bitmap.createBitmap(1, 1, Config.ARGB_8888));
+			Runnable renderRunnable = new Runnable() {
```
hieupham007: you can use TiMessenger.sendBlockingMainMessage here and pass an Object as argument.

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -916,14 +918,40 @@ protected void handleFocus()
...
-
-		} else {
-			return (TiBlob) TiMessenger.sendBlockingMainMessage(getMainHandler().obtainMessage(MSG_TOIMAGE), getActivity());
+		final boolean waitForFinish = (callback == null);
+		TiBlob blob;
+		
+		/*
+		 * Callback don't exist. Just render on main thread and return blob. 
+		 */
+		if (waitForFinish) {
```
hieupham007: You can just use (callback == null) instead of making a new variable and only use it once

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -916,14 +918,40 @@ protected void handleFocus()
...
+				blob = handleToImage();
+			} else {
+				blob = (TiBlob) TiMessenger.sendBlockingMainMessage(getMainHandler().obtainMessage(MSG_TOIMAGE), getActivity());
+			}
+		} 
+		
+		/*
+		 * Callback exists. Perform async rendering and return an empty blob.
+		 */
+		else {
```
hieupham007: Please follow code format guideline: 
if (...) {

} else {

}

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
sriks: Calling TiMessenger.sendBlockingMainMessage() is performing the msg synchronously while blocking on the main thread. This essentially makes the method execution synchronous. But according to doc and iOS implementation - if a callback is passed, the method is supposed to execute asynchronously. So Im using a runnable to render and send the result through callback. 

Am I missing anything here ?

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
sriks: Noted

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
sriks: Noted

hieupham007:

PR #7299, opened 2015-10-13
[TIMOB-19716] Support contact icon in UIApplicationShortcuts

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19716

### DIFF
--- a/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
+++ b/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
@@ -62,7 +62,7 @@ methods:
         type: String
 
 examples:
-  - title: Example app.js
+  - title: Full example (get shortcuts, add shortcuts, remove shortcuts, check shortcuts).
     example: |
         The following code excerpt demonstrates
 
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
```diff
@@ -185,8 +247,9 @@ properties:
 
   - name: icon
     summary: |
-        The icon to be displayed on the application shortcut, for example <Titanium.UI.iOS.SHORTCUT_ICON_TYPE_COMPOSE>.
-        Note: You can also use a local image specified by the image path.
-    type: Number
+        The icon to be displayed on the application shortcut. You can either use one of the contastants like
```
cheekiatng: spelling: not contestants, but constants

hansemannn: Addressed your comment in the latest commit, thanks!

cheekiatng:
CR and FT passed. 2 questions:
in your example app:
1. i tried doing "Add Shortcut" twice with 2 different contacts. but there's only 1 shortcut appearing. which is the first add. is this expected behavior?
2. Assuming it is the expected behavior, shouldn't the 2nd "add shortcut" overwrite the first? because the first contact always appears (until i chose remove)

hansemannn:
The first button is "Add Ti.Contacts Application Shortcut" to create a shortcut and the second is "Remove Ti.Contacts Application Shortcut" to remove the previous created shortcut to allow the tester to test some contacts.

cheekiatng:
APPROVED!

PR #7323, opened 2015-10-20
[TIMOB-19737] Error when selecting an iOS certificate without provisioning profiles

feons:
https://jira.appcelerator.org/browse/TIMOB-19737


cheekiatng:
APPROVED!

PR #7324, opened 2015-10-20
[TIMOB-19753] Fix UserActivity macros

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19753


iphone/Classes/TiAppiOSSearchableItemProxy.h
```diff
@@ -5,10 +5,10 @@
  * Please see the LICENSE included with this distribution for details.
  */
 #if IS_XCODE_7
+#ifdef USE_TI_APPIOS
```
cheekiatng: should this be 
```
#ifdef USE_TIAPPIOSSEARCHABLEITEM
```
instead?

iphone/Classes/TiAppiOSSearchableItemProxy.m
```diff
@@ -5,12 +5,11 @@
  * Please see the LICENSE included with this distribution for details.
  */
 #if IS_XCODE_7
+#ifdef USE_TI_APPIOS
```
cheekiatng: should this be 
```
#ifdef USE_TIAPPIOSSEARCHABLEITEM
```
instead?

iphone/Classes/TiAppiOSSearchableItemProxy.h
hansemannn: The legacy plan was to not replace every constant as part of this PR, but yes it makes sense. Will update the remaining constants. Thanks!

cheekiatng:
APPROVED!

PR #7325, opened 2015-10-21
[TIMOB-19739] iOS: Handle null case where phone number is added via phone dialer

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19739


hansemannn:
Approved!

PR #7330, opened 2015-10-22
`[TIMOB-19757] Cannot access files in Resources directory with 5.2.x

feons:
https://jira.appcelerator.org/browse/TIMOB-19757

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -4056,7 +4056,7 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 					case 'png':
 					case 'jpg':
 						if (useAppThinning) {
-							imageAssets[relPath] = info
+							imageAssets[relPath] = info;
 						} else {
 							resourcesToCopy[relPath] = info;
 						}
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -4535,7 +4542,13 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 		},
 
 		function createAssetImageSets() {
-			this.logger.info(__('Creating assets image set'));
+			if (this.useAppThinning) {
+				this.logger.info(__('Creating assets image set'));
+			} else {
+				this.logger.info(__('Skip creating assets image set'));
+				return;
```
cb1kenobi: This is a bit hard to read. The "then" falls through, but the "else" returns. Please reorder this to be easier to follow. Also, I think it would be a good idea to state why we're skipping the creation of the asset image sets.

```javascript
if (!this.useAppThinning) {
    this.logger.info(__('App thinning disabled, skipping asset image sets'));
}
```

iphone/cli/commands/_build.js
```diff
@@ -4048,7 +4055,7 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 
 					case 'png':
 					case 'jpg':
-						imageAssets[relPath] = info;
+						useAppThinning ? imageAssets[relPath] = info : resourcesToCopy[relPath] = info;
```
cb1kenobi: This is hard to read. Just use a plain old if/else please. I have no problem doing inline if's, however you're assigning a l-value and it's easy for people to get confused.

cb1kenobi:
Looks good. CR'd and FR'd. **APPROVED**

PR #7333, opened 2015-10-22
[TIMOB-19766] Android: Modifying current/root activity's content view in app.js gets overwritten by default layout

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19766

This PR is specifically to support hyperloop usage on the UI thread. If the user modifies the root/current activity's content view, we will actually blow it away with our "default layout" when we run Kroll in the UI thread. The easiest hack around this I found was to override the setContentView calls to toggle a flag. Then when we're about to set the default layout we skip doing so if the flag is set to true.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -64,6 +64,7 @@ import android.view.Surface;
 import android.view.View;
 import android.view.Window;
 import android.view.WindowManager;
+import android.view.ViewGroup.LayoutParams;
 
 import com.appcelerator.analytics.APSAnalytics;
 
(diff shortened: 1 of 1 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -114,6 +115,8 @@
 	public TiWindowProxy lwWindow;
 	public boolean isResumed = false;
 
+    private boolean overridenLayout;
```
hieupham007: indentation is off here.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -502,6 +505,25 @@ protected void windowCreated(Bundle savedInstanceState)
 		}
 	}
 
+	// Record if user has set a content view manually from hyperloop code during require of app.js!
+	@Override
+	public void setContentView(View view) {
```
hieupham007: Indentation for these methods seem to be a little bit off.

sgtcoolguy:
@hieupham007 I pushed the indentation fix. If the PR is OK, let me know and I'll squash merge it so it's not two commits.

PR #7334, opened 2015-10-22
[TIMOB-19768] iOS: changes related to Hyperloop

jhaynie:
For FR:
- build a Titanium out without Hyperloop plugin and module. Should not fail and run as normal.
- add Hyperloop plugin and module and then make sure that it works. This project is setup to run out of the box: https://github.com/appcelerator/hyperloop-examples


pec1985:

PR #7339, opened 2015-10-23
[TIMOB-19707] If the DefaultIcon.png or DefaultIcon-ios.png contains …

cb1kenobi:
…an alpha channel, the alpha channel is stripped before generating app icons.


feons:
Looks great! **Approved!**

PR #7357, opened 2015-10-28
[TIMOB-19567] iOS: Deprecate useractivitywillsave event

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19567

### DIFF
--- a/apidoc/Titanium/App/iOS/UserActivity.yml
+++ b/apidoc/Titanium/App/iOS/UserActivity.yml
@@ -85,7 +85,7 @@ properties:
 
   - name: needsSave
     summary: |
-        Set to true everytime the user activity needs to be updated before handing it off to another device.
+        Set to true everytime you have updated the user activity and need the changes to be saved before handing it off to another device.
     type: Boolean
     osver: {ios: {min: "8.0"}}
 
--- a/apidoc/Titanium/App/iOS/UserActivity.yml
+++ b/apidoc/Titanium/App/iOS/UserActivity.yml
@@ -85,17 +85,7 @@ properties:
 
   - name: needsSave
     summary: |
-        Determines if iOS should fire the `useractivitywillsave` event for the application to update
-        the state or content of the user activity.
-    description: |
-        Set the property to `true` if the user activity needs to be updated before handing it off to
-        another device, then listen for the `useractivitywillsave` event to update the state or content
-        of the user activity.
-
-        iOS will reset the property to false after the `useractivitywillsave` event fires.
-
-        Note that when the user activity first becomes active, the `useractivitywillsave` event
-        will fire once even if the property is set to false.
+        Set to true everytime the user activity needs to be updated before handing it off to another device.
     type: Boolean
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiAppiOSUserActivityProxy.m
```diff
@@ -186,7 +186,7 @@ -(NSNumber*)isValid
 
 -(NSNumber*)supported
 {
-    DEPRECATED_REPLACED(@"App.IOS.UserActivity.getSupported()" ,@"5.1.0",@"App.IOS.UserActivity.isSupported()")
+    DEPRECATED_REPLACED(@"App.IOS.UserActivity.getSupported()" ,@"5.1.0",@"Titanium.App.IOS.UserActivity.isSupported()")
```
FokkeZB: Should the first line be prefixed with "Titanium" as well?

iphone/Classes/TiAppiOSUserActivityProxy.m
```diff
@@ -197,6 +197,7 @@ -(NSNumber*)supported
 - (void)userActivityWillSave:(NSUserActivity *)userActivity
 {
     if([self _hasListeners:@"useractivitywillsave"]){
+        DebugLog(@"[WARN] Titanium.App.IOS.UserActivity.useractivitywillsave event is deprecated. Update user activity by setting Titanium.App.IOS.UserActivity.needsSave property to true instead.");
```
FokkeZB: Just setting needsSave is not gonna do it. I think it's better to put it this way:

> Update user activity and then set the Titanium.App.iOS.UserActivity.needsSave property to true every time you have a change.

apidoc/Titanium/App/iOS/UserActivity.yml
```diff
@@ -85,17 +85,7 @@ properties:
...
-    description: |
-        Set the property to `true` if the user activity needs to be updated before handing it off to
-        another device, then listen for the `useractivitywillsave` event to update the state or content
-        of the user activity.
-
-        iOS will reset the property to false after the `useractivitywillsave` event fires.
-
-        Note that when the user activity first becomes active, the `useractivitywillsave` event
-        will fire once even if the property is set to false.
+        Set to true everytime the user activity needs to be updated before handing it off to another device.
```
FokkeZB: I think this is more clear:

> Set to true everytime you have updated the user activity and need the changes to be saved before handing it off to another device.

iphone/Classes/TiAppiOSUserActivityProxy.m
hansemannn: We have a weird behavior in the macro that I discovered last week. The prefix is added to the first part, but not to the second one. So this is right for now and will be refactored along with the other used macros in the future to unify the macro.

hansemannn:
Thanks @FokkeZB for checking the docs. Will merge after @cheekiatng is back.

hansemannn:
Approved!

PR #7362, opened 2015-10-28
[TIMOB-19799] Updated ioslib to v0.8.2 to fix bug when simulators are…

cb1kenobi:
… found from previous install, but the old Xcode was removed and the new Xcode was installed and the EULA was not accepted.


cb1kenobi:
Fire in the hole! **APPROVED**

PR #7363, opened 2015-10-28
[TIMOB-19741] iOS: Fixed blank webViews

AngelkPetkov:
Fixed bug ,where cashed webViews could not be loaded. Will not set url for the webView unless there is a url property set in the js.

JIRA: https://jira.appcelerator.org/browse/TIMOB-19741


hansemannn:
CR + FT **approved**, merging.

PR #7374, opened 2015-10-29
[TC-5746] iOS: Added optionsImages in OptionDialog

kopiro:
https://jira.appcelerator.org/browse/TC-5746

### DIFF
--- a/iphone/Classes/TiUIOptionDialogProxy.m
+++ b/iphone/Classes/TiUIOptionDialogProxy.m
@@ -54,6 +54,8 @@
 		[options addObject:NSLocalizedString(@"OK",@"Alert OK Button")];
 	}
 
+    NSMutableArray *imagesOptions = [self valueForKey:@"imagesOptions"];
+
     forceOpaqueBackground = [TiUtils boolValue:[self valueForKey:@"opaquebackground"] def:NO];
     persistentFlag = [TiUtils boolValue:[self valueForKey:@"persistent"] def:YES];
     cancelButtonIndex = [TiUtils intValue:[self valueForKey:@"cancel"] def:-1];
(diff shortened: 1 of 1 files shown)


hansemannn:
Hi @kopiro, also thanks for this PR! Please provide a sample code before we do a further review. Also check the removed (red) lines again.

iphone/Classes/TiUIOptionDialogProxy.m
```diff
@@ -99,17 +101,22 @@ -(void)show:(id)args
...
             UIAlertAction* theAction = [UIAlertAction actionWithTitle:@"Cancel"
                                                                 style:UIAlertActionStyleCancel
                                                               handler:^(UIAlertAction * action){
                                                                   [self fireClickEventWithAction:action];
                                                               }];
+            if (!IS_NULL_OR_NIL(imagesOptions)) {
+                UIImage *image = [ [UIImage imageNamed:[imagesOptions objectAtIndex:curIndex]] imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
+                [theAction setValue:image forKey:@"image"];
+            }
+
```
hansemannn: This code would only be executed, if the current device is an iPad and no ``cancel`` property is set. It also does not check for array bounds (use case: options = ["test1","test2"] and imageOptions = ["image.png"]). So:
- Rename ``imagesOptions`` to ``imageOptions``
- Check the array bounds
- Use the TiUtils helper to create an image. It makes it save for the new asset catalog coming by default in 5.2.0.

Example:
```
NSMutableArray *imagesOptions = [self valueForKey:@"imageOptions"];

...
 
if (imagesOptions != nil && curIndex < [imagesOptions count]) {
  UIImage *image = [ [TiUtils image:[imagesOptions objectAtIndex:curIndex] proxy:self] imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
  [theAction setValue:image forKey:@"image"];
}
```

kopiro: @hansemannn I tried to understand TiHelpers, can you provide me sample / doc to create images in a right way?
Thanks!

hansemannn:
Closing this PR for now after discussing with @kopiro, that we should wrap the image options as part of a refactoring that includes an object instead of string values inside the ``options`` property.

PR #7378, opened 2015-11-02
Fix Ti.Calendar.* constants for permission model to Ti.Contacts.*

FokkeZB:


FokkeZB:
Would you review and merge @hansemannn ?

PR #7386, opened 2015-11-03
[TIMOB-19851] Android: Media - Crash and other bugs in new permissions architecture

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-19851

**NOTE:** This _only_ resolves the documentation part of the JIRA ticket. Crash is still there.

### DIFF
--- a/apidoc/Titanium/Media/Media.yml
+++ b/apidoc/Titanium/Media/Media.yml
@@ -5,6 +5,17 @@ description: |
     The Media module is used to access the device's media-related functionality, such
     as using the device's camera and photo gallery, playing media files, or recording
     audio or video.
+    
+    #### Android Platform Notes
+
+    Currently, on Android, camera permissions must be explicitly configured in
+    `tiapp.xml` in order to access the camera:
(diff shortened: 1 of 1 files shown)


hieupham007:
Fokke, our CLI automatically add the camera permission when certain methods are used in the app, i.e: showCamera(). In your example, you did not use any camera related methods (besides the methods requesting for permissions), therefore the 'camera' permission is not added to the manifest. This is correct behavior.

FokkeZB:
@hieupham007 I see:
https://github.com/appcelerator/titanium_mobile/blob/1607237763f9e490e41d98b8d711ac49eeb01043/android/cli/commands/_build.js#L3368

Well, then surely if you use a permission method those permissions should be added right?

FokkeZB:
Will create a JIRA ticket to add the permission request methods to 
https://github.com/appcelerator/titanium_mobile/blob/1607237763f9e490e41d98b8d711ac49eeb01043/android/cli/commands/_build.js#L3368

PR #7392, opened 2015-11-03
[TIMOB-19828] [5_1_X] node-ios-device, ioslib, and TiSDK to work with Node.js 5.0

ingo:
We update ioslib to 0.9.1 (and its dependencies as appropriate).

We set the maximum allowable node.JS version as 5.x

The part of interest is the ABI (or modules) version here, which would
require a recompile:

https://nodejs.org/en/download/releases/

Note it only changes on major version bumps


pinnamur:
FR passed. I'm able to build an iOS app onto SDK on different versions of Node (4.x and 5.0)

PR #7393, opened 2015-11-04
[TIMOB-19777] iOS: Disable new UITableView property by default

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-19777
cellLayoutMarginsFollowReadableWidth

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -197,13 +197,15 @@ static TiViewProxy * FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoin
             [_tableView setLayoutMargins:UIEdgeInsetsZero];
         }
         
+#if IS_XCODE_7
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiUIListView.m
```diff
@@ -201,6 +201,11 @@ - (UITableView *)tableView
     if ([_tableView superview] != self) {
         [self addSubview:_tableView];
     }
+    
+    if ([TiUtils isIOS9OrGreater]) {
+        _tableView.cellLayoutMarginsFollowReadableWidth = NO;
+    }
+    
     return _tableView;
```
hansemannn: Please do this check before line 200 in the initial creation of the ``UITableView``. Also make sure to use the ``IS_XCODE_7`` macro to not break Xcode 6 builds.

hansemannn:
CR + FT passed, APPROVED!

PR #7407, opened 2015-11-06
Insist titanium_prep.macos to return standard out.

atrovis:
In my environment (osx 10.10.5 / TiSDK 5.0.2.GA / JDK 1.6.0_65), titanium_prep.macos returns no output. 
It causes android module build failure with common.js file in assets folder.
So, I added "-v" option to force standard out and it worked fine.

Thank you.

### DIFF
--- a/support/android/jspacker.py
+++ b/support/android/jspacker.py
@@ -108,7 +108,7 @@ class Crypt(object):
     output = open(os.path.join(target_dir, 'AssetCryptImpl.java'), 'w')
 
     sys.stdout.flush()
-    cmdargs = [titanium_prep, package, asset_dir]
+    cmdargs = [titanium_prep, "-v", package, asset_dir]
     cmdargs.extend(self.files)
     so, process = run.run(cmdargs, return_process=True)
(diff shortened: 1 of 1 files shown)


hansemannn:
@atrovis Can you please sign the [CLA](https://cla.appcelerator.com/) first?

atrovis:
@hansemannn I've just signed. Thanks for your reply.

hansemannn:
Thanks! Would be great to have @cb1kenobi take a look at it.

cb1kenobi:
For starters, the Python scripts are deprecated. We don't support them anymore.

Second, the Android version of titanium_prep has no `-v` flag. Furthermore, all output is written to stdout. I can't show the code, but believe me, it's littered with `System.out.println()` calls.

I don't care if we merge the PR. This PR is a reminder that the Python scripts should be removed.

hansemannn:
Alright! So since we will most likely remove the python scripts in the upcoming 6.0 timeframe, I'll close the PR at least for now. Thanks anyway!

PR #7412, opened 2015-11-09
[TIMOB-19694] Added support for launch screen storyboards for iOS 9

cb1kenobi:
https://jira.appcelerator.org/browse/TIMOB-19694


cb1kenobi:
Fixed merge conflicts.

iphone/Classes/TiRootViewController.m
```diff
@@ -209,7 +211,19 @@ - (void)remoteControlReceivedWithEvent:(UIEvent *)event
...
+	if (chosenColor == nil) {
+#if defined(DEFAULT_BGCOLOR_RED) && defined(DEFAULT_BGCOLOR_GREEN) && defined(DEFAULT_BGCOLOR_BLUE)
+		chosenColor = [UIColor colorWithRed: DEFAULT_BGCOLOR_RED
+									  green: DEFAULT_BGCOLOR_GREEN
+									   blue: DEFAULT_BGCOLOR_BLUE
+									  alpha: 1.0f];
+#else
+		chosenColor = [UIColor blackColor];
+#endif
+	}
```
hansemannn: If `bgColor` is not nil, it will not be assigned here. Is that intended?

cb1kenobi: Not sure what you mean. `chosenColor` is either `bgColor` or black. If `bgColor` is `nil`, then `chosenColor` is `nil`. If `chosenColor` is `nil`, then we used to set it to black, but now we check if the colors are defined otherwise fallback to black. The code is good. Not sure what your concern is.

cheekiatng:
CR and FT passed! APPROVED!

PR #7419, opened 2015-11-11
[TIMOB-19667] Support new 3D-Touch events and properties

AngelkPetkov:
:JIRA:  https://jira.appcelerator.org/browse/TIMOB-19667
:JIRA: https://jira.appcelerator.org/browse/TIMOB-19492?filter=-1

### DIFF
--- a/apidoc/Titanium/UI/TableView.yml
+++ b/apidoc/Titanium/UI/TableView.yml
@@ -581,21 +581,28 @@ events:
         type: Number
 
       - name: force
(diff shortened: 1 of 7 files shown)


cheekiatng:
Did a CR. looks good. Only comment on the yml docs is, for the new properties for the event listeners, it'll be best if you also include:
since: 5.2.0
platforms: [iphone, ipad]
osver: {ios: {min: "9.1"}}

PR #7421, opened 2015-11-11
[TIMOB-19896] (5_1_X) iOS: Fix contact where email label is nil in vcf card scenario

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19896


hansemannn:
Approved!

PR #7423, opened 2015-11-11
[TIMOB-19898] Fixed abi parsing to be fool proof...

cb1kenobi:
if such a thing exists.


cheekiatng:
CR and FT passed. APPROVED.

PR #7424, opened 2015-11-11
[TIMOB-19544] Added support for iPad Pro app icons. Didn't add launch…

cb1kenobi:
… images because they are the same as iPad Air. Probably because they want you to use storyboards.

https://jira.appcelerator.org/browse/TIMOB-19544


cheekiatng:
CR and FT passed. Approved!

PR #7436, opened 2015-11-13
[TIMOB-19941] iOS: ScrollableView set pageControl color to transparent

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-19941

While showPagingControl is disabled , there should now be no longer a black footer.


hansemannn:
Closing in favor of #7443 and #7444.

PR #7443, opened 2015-11-16
[TIMOB-19761] iOS: ScrollableView produces 20px black paging control by default

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19761


hansemannn:
CR + FT passed, PR **approved**!

PR #7444, opened 2015-11-16
[TIMOB-19761](5_1_X) iOS: ScrollableView produces 20px black paging control by default

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19761


hansemannn:
Approved!

PR #7447, opened 2015-11-16
[TIDOC-2252] Docs on Interactive Notifications needs to be updated for Apple Watch

FokkeZB:
https://jira.appcelerator.org/browse/TIDOC-2252


hansemannn:
Approved!

PR #7448, opened 2015-11-16
Fix TIMOB-19903 Android: Fix crash due to GC of UI native proxies in hyperloop

sgtcoolguy:
This is actually related to running on main thread, not specifically hyperloop. Also, while this does fix the issue, it raises the question on whether we want to always call the on\* callbacks sync, even if we're not running on main thread. It's entirely possible that the crash could happen if the timing is right when fired async for onStop/onDestroy.
- Need to fire on\* property callbacks synchronously when run on main thread. Otherwise they fire _after_ we've already released the activity proxy on the native side and we get a fatal crash.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -72,7 +72,7 @@ import com.appcelerator.analytics.APSAnalytics;
  * The base class for all non tab Titanium activities. To learn more about Activities, see the
  * <a href="http://developer.android.com/reference/android/app/Activity.html">Android Activity documentation</a>.
  */
-public abstract class TiBaseActivity extends AppCompatActivity 
+public abstract class TiBaseActivity extends AppCompatActivity
 	implements TiActivitySupport/*, ITiWindowHandler*/
 {
 	private static final String TAG = "TiBaseActivity";
@@ -94,7 +94,7 @@ public abstract class TiBaseActivity extends AppCompatActivity
 
 	public static KrollObject cameraCallbackContext, contactsCallbackContext, oldCalendarCallbackContext, calendarCallbackContext, locationCallbackContext;
 	public static KrollFunction cameraPermissionCallback, contactsPermissionCallback, oldCalendarPermissionCallback, calendarPermissionCallback, locationPermissionCallback;
-	
+
 	protected View layout;
 	protected TiActivitySupportHelper supportHelper;
 	protected int supportHelperId = -1;
@@ -108,7 +108,7 @@ public abstract class TiBaseActivity extends AppCompatActivity
 	protected int msgActivityCreatedId = -1;
 	protected int msgId = -1;
 	protected static int previousOrientation = -1;
-	//Storing the activity's dialogs and their persistence 
+	//Storing the activity's dialogs and their persistence
 	private CopyOnWriteArrayList<DialogWrapper> dialogs = new CopyOnWriteArrayList<DialogWrapper>();
 	private Stack<TiWindowProxy> windowStack = new Stack<TiWindowProxy>();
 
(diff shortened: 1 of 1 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -1093,7 +1093,15 @@ private void dispatchCallback(String name, KrollDict data) {
...
 
-		activityProxy.callPropertyAsync(name, new Object[] { data });
+		// TIMOB-19903
+		if (TiApplication.getInstance().runOnMainThread()) {
+			// We must call this synchornously to ensure it happens before we release the Activity reference on the V8/Native side!
+			activityProxy.callPropertySync(name, new Object[] { data });
+		} else {
+			// This hopefully finishes before we release the reference on the native side?! I have seen it crash because it didn't before though...
+			// Not sure it's safe to keep this behavior...
+			activityProxy.callPropertyAsync(name, new Object[] { data });
```
hieupham007: under what circumstances would this method be called on KrollRuntime thread? This is used exclusively on on* methods, which run on the main thread.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -1093,7 +1093,15 @@ private void dispatchCallback(String name, KrollDict data) {
 
 		data.put("source", activityProxy);
 
-		activityProxy.callPropertyAsync(name, new Object[] { data });
+		// TIMOB-19903
+		if (TiApplication.getInstance().runOnMainThread()) {
+			// We must call this synchornously to ensure it happens before we release the Activity reference on the V8/Native side!
+			activityProxy.callPropertySync(name, new Object[] { data });
```
hieupham007: Aren't we only releasing the references in onDestroy()? Making it sync on every on* methods seem a bit excessive.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
sgtcoolguy: I'm not sure I follow. With recent changes (required by hyperloop) the KrollRuntime thread may be the same as the UI thread or separate depending on a tiapp property.

callPropertyAsync always sticks a message in the queue to fire the JS method. When the KrollRuntime thread is the main/UI thread, that means it's guaranteed not to fire until after all the work it's currently doing - so it won't ever run until after we've already destroyed the activity proxy during onDestroy flows.

When there are separate main and UI threads, I'm not sure that there's any guarantee of ordering between the onDestroy/onStop callback and the release of the activity proxy. So I think it will effectively be a race condition. I have anecdotally seen this when debugging and not running on the main thread - where closing a window would cause a crash. But perhaps I'm wrong and we can guarantee/know that the on* callback will always be fired and finished before the release call later in the onDestroy flow?

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
sgtcoolguy: That's why I limited the change to only the cases where we're running Kroll on the UI thread - because I wasn't sure if there were other instances of us assuming ordering of these callbacks versus other work. When we're running in that mode, I would assume we'd want to ensure that the onStart would be guaranteed to run while we're doing our onStart() stuff, and not _after_. If we don't run it sync and we're on the UI thread, it's guaranteed as-is to run after we're fully done with all of our setup for the on* calls. When run sync, it's guaranteed to run exactly when we decide to fire this callback in that process.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
hieupham007: Ah, I'm not up to date with the recent hyperloop changes. If the callbacks are async then there's no guarantee that it will be finished before release call. This change should be fine.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
hieupham007: Ok.

PR #7457, opened 2015-11-17
[TIMOB-2908](5_1_X) iOS: Switch - Disable animation on init but allow developer to enable/disable animation post init

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-2908


hansemannn:
CR + FT passed, PR approved!

PR #7458, opened 2015-11-17
[TIMOB-20003] Permission request error response message prefixed with namespace

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-20003


hansemannn:
Why did you remove the prefix? I think a proper formatting would have been better.

FokkeZB:
Because you get the message in the callback for the specific request you made so it doesn't add anything useful.

hansemannn:
Checked the attached logs inside the ticket, makes sense. Approved!

PR #7464, opened 2015-11-18
[TIMOB-20010] iOS9: FIX remove contacts causing app to crash

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20010


hansemannn:
Approved!

PR #7468, opened 2015-11-18
[TIMOB-20013](5_1_X) Ti.Contacts - Fix relatedNames property causing crash

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20013


hansemannn:
CR + FT passed, PR approved!

PR #7472, opened 2015-11-19
[TIMOB-20014] Fixed differential build bug when JavaScript minification is disabled.

cb1kenobi:


cheekiatng:
CR and FT passed. APPROVED!

PR #7475, opened 2015-11-19
[TIMOB-19692] iOS: Kitchen Sink Update

AngelkPetkov:


pec1985:
We are going to get rid of the Kitchen Sink project in **this** repo eventually. We'll just leave a Resources folder and a simple app.js.
No need to have multiple copies of Kitchen Sink.

PR #7486, opened 2015-11-23
[TIMOB-20048](5_1_X) iOS: Revert APSHTTPClient library to use back NSURLConnection

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20048


hansemannn:
Tested against the given examples, PR approved!

PR #7488, opened 2015-11-24
[TIMOB-18827] Android: Allow for specifiying image to cache

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-18827
See test code in ticket.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIImageView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIImageView.java
@@ -761,6 +761,18 @@ public class TiUIImageView extends TiUIView implements OnLifecycleEvent, Handler
 		}
 	}
 
+	private void cacheImage(TiDrawableReference imageRef) {
+	    synchronized(releasedLock) {
+            int hash = imageRef.hashCode();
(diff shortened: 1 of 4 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIImageView.java
```diff
@@ -761,6 +761,18 @@ private void setImageInternal() {
...
 	}
 
+	private void cacheImage(TiDrawableReference imageRef) {
+	    synchronized(releasedLock) {
+            int hash = imageRef.hashCode();
+            Bitmap b = mMemoryCache.get(hash);
+            if (b == null) {
+                Log.i(TAG, "Image isn't cached");
+                b = imageRef.getBitmap(true);
+                mMemoryCache.put(hash, b);
```
hieupham007: what if the image is UHD, like 100mb?

hieupham007: caching an image b/c it flickers the first time it loads doesn't seem like a good solution. Also, we already have a caching system in place...

ashcoding:
Closing and rejecting this PR cause workaround is acceptable.

PR #7495, opened 2015-11-25
[TIMOB-20061] (5_1_X) Android: Set TLS properly when validatesSecureCert is true.

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-20061


ashcoding:
Code reviewed and tested functionally. **Accepted**

PR #7496, opened 2015-11-25
[TIMOB-19455] Add Swift versions of the applewatch templates

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-19455

Includes the following:

**cli/lib/creators/applewatch.js**
- Changes default applewatch template from `watchos1` to `watchos2-swift`.
- Changes target names to be like Xcode creates them.

**iphone/cli/commands/_build.js**
- Removes swift-flag check since it checks on the wrong place (for some reason Xcode only sets the flag for the WatchKit App and not the Extension that contains Swift as you'd expect) and it's the developers's responsibility anyway.
- Supports using `$(PRODUCT_BUNDLE_IDENTIFIER)'` as the `CFBundleIdentifier` in the `Info.plist`, which is what Xcode does.

**templates/applewatch/**
- Removes code like GlanceController that Xcode doesn't put there by default.
- Fixes spaces in `IBSC_MODULE`.
- Improves the use of variables for folder names and some variables.
- Adds `PRODUCT_BUNDLE_IDENTIFIER` to the `.pbxproj` and uses it in `Info.plist` as Xcode does.
- Adds a `watchos2-swift` template.
- Copies `watchos2` to `watchos2-objc` so we can remove `watchos2` in 6.0


FokkeZB:
Ready for you @cheekiatng

cheekiatng:
FT Passed. APPROVED!

PR #7497, opened 2015-11-25
[TIMOB-19452] Extensions should default to the app's provisioning profile if not set

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-19452


cheekiatng:
APPROVED!

PR #7499, opened 2015-11-25
[TIMOB-19725] iOS : Toggled action buttons display over keyboard

AngelkPetkov:
The action buttons can now only be set upon creation.

:JIRA: https://jira.appcelerator.org/browse/TIMOB-19725?filter=-1


cheekiatng:
CR and FT Passed. APPROVED.

PR #7500, opened 2015-11-26
[TIMOB-20063]: iOS: Ti.Contacts.showContacts crashes

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-20063

### DIFF
--- a/apidoc/Titanium/Contacts/Contacts.yml
+++ b/apidoc/Titanium/Contacts/Contacts.yml
@@ -538,6 +538,8 @@ properties:
       property.
       Note: If ringtone or texttone is selected, null values are returned, since these are unsupported
       by Apple.
+      Since iOS 9.0, there is a native apple issue whereby it may return null if the birthday property 
+      is selected on certain device models.
   platforms: [iphone, ipad]
   type: Callback<Object>
   optional: true
--- a/iphone/Classes/TiContactsPerson.m
+++ b/iphone/Classes/TiContactsPerson.m
@@ -55,7 +55,7 @@ static NSDictionary* iOS9propertyKeys;
 	if (self = [super _initWithPageContext:context]) {
 		person = [person_ retain];
 		module = module_;
-		iOS9contactProperties = [self getiOS9ContactProperties: person_];
+		iOS9contactProperties = [[self getiOS9ContactProperties: person_] retain];
 	}
 	return self;
 }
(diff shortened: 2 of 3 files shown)


cheekiatng:
CR and FT passed. APPROVED!

iphone/Classes/TiContactsPerson.m
```diff
@@ -84,33 +84,26 @@ -(NSDictionary*)getiOS9ContactProperties: (CNMutableContact*) contact
...
-		contact.nickname, @"nickname",
-		contact.phoneticGivenName, @"firstPhonetic",
-		contact.phoneticFamilyName, @"lastPhonetic",
-		contact.phoneticMiddleName, @"middlePhonetic",
-		contact.organizationName, @"organization",
-		contact.jobTitle, @"jobTitle",
-		contact.departmentName, @"department",
-		contact.emailAddresses, @"email",
-		contact.note, @"note",
-//		contact.birthday, @"birthday", birthday property has some problems in iOS9 Beta, have to be exclusively handled
```
cheekiatng: There was a special handle case for birthday here. Not sure if Apple has already fixed this since Beta, but it's best to revisit it. Right now, when you do showContacts and select birthday, it returns undefined. 
Please also check alternate birthday to make sure it works correctly. as well as anniversary.
For reference, the behavior should be the same as when run on iOS8.4

cheekiatng:
I also noticed that there was a mistake in the method name in ContactsModule.m line 192. it should be
DEPRECATED_REPLACED(@"Contacts.requestAuthorization()", @"5.1.0", @"Contacts.requestContactsPermissions()");
Please change this line as well as part of this PR.

iphone/Classes/TiContactsPerson.m
sriks: Birthdays seems to have a strange issue. We are getting the birthday from selected CNContact. But CNContactPropery.key is returning "birthdays" with suffix s. This is tossing the lookup and returning the birthday as empty. Not sure if this is an issue with iOS 9. 

iOS 8 works as expected.

sriks: iOS9: The CNContactProperty for gregorian date is retiring an undocumented string "_systemCalendar". Accommodated this check as well. Now returns the gregorian date correctly.

iphone/Classes/TiContactsPerson.m
```diff
@@ -618,7 +611,7 @@ -(id)valueForUndefinedKey:(NSString *)key
 #if IS_XCODE_7
 	if ([TiUtils isIOS9OrGreater]) {
 		//birthday property managed seperately
-		if ([key isEqualToString:@"birthday"]) {
+		if ([key isEqualToString:@"birthday"] && [person isKeyAvailable:CNContactBirthdayKey]) {
```
cheekiatng: I just tested by selecting the person's birthday in showContacts. It doesn't return the date because isKeyAvailable is returning no. It looks like if you select birthday, apple doesn't return you the value. If this is the case, and it's native behavior, we can just document about it under `selectedProperty` in apidoc

sriks: Yes thats true. But the behaviour is inconsistent. It works well in iPhone 6 (9.1) simulator only! Not sure about device. Anyways I will add this note to docs.

apidoc/Titanium/Contacts/Contacts.yml
```diff
@@ -538,6 +538,7 @@ properties:
       property.
       Note: If ringtone or texttone is selected, null values are returned, since these are unsupported
       by Apple.
+      Selecting Birthday returns null in iOS 9. This is a native issue.
```
cheekiatng: Since iOS 9.0, there is a native apple issue whereby it may return null if the birthday property is selected on certain device models.

cheekiatng:
@sriks Change the doc as what i have described, and also update the back port with this change.

sriks:
@cheekiatng Updated docs accordingly. Updated back port at https://github.com/appcelerator/titanium_mobile/pull/7504

cheekiatng:
CR and FT passed. Approved.

PR #7512, opened 2015-11-30
[TC-5781] iOS: Support for `selected` event on TabGroup when tab already selected

kopiro:
https://jira.appcelerator.org/browse/TC-5781


hansemannn:
Status update here: PR looks great, but as it is a change of behavior, we will make it likely part of 6.0.0.

hansemannn:
**Approved**, thank you @kopiro !

PR #7513, opened 2015-11-30
[TIMOB-20079] iOS: Style Ti.UI.iPad.Popover using backgroundColor

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20079


cheekiatng:
CR and FT Passed. Approved!

PR #7518, opened 2015-12-01
[TIMOB-19802] (5_1_X) Android: CardView should use standard property names

ashcoding:
https://jira.appcelerator.org/browse/TIMOB-19802


ashcoding:
**Merging**

PR #7528, opened 2015-12-02
[TIMOB-20004] iOS: Differences between new permission request responses

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-20004

### DIFF
--- a/iphone/Classes/CalendarModule.m
+++ b/iphone/Classes/CalendarModule.m
@@ -244,10 +244,10 @@ typedef void(^EKEventStoreRequestAccessCompletionHandler)(BOOL granted, NSError 
                                  completion:^(BOOL granted, NSError *error){
                                      NSDictionary* propertiesDict;
                                      if (error == nil) {
-                                         propertiesDict = [TiUtils dictionaryWithCode:(granted ? 0 : 1) message:nil];
+                                         NSString* errorMsg = granted ? nil : @"The user has denied access to events in Calendar.";
+                                         propertiesDict = [TiUtils dictionaryWithCode:(granted ? 0 : 1) message:errorMsg];
                                      } else {
-                                         propertiesDict = [TiUtils dictionaryWithCode:[error code]
-                                                                              message:[TiUtils messageFromError:error]];
+                                         propertiesDict = [TiUtils dictionaryWithCode:[error code] message:[TiUtils messageFromError:error]];
                                      }
                                      KrollEvent * invocationEvent = [[KrollEvent alloc] initWithCallback:callback eventObject:propertiesDict thisObject:self];
                                      [[callback context] enqueue:invocationEvent];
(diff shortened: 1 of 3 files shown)


iphone/Classes/ContactsModule.m
```diff
@@ -231,7 +231,8 @@ -(void)requestContactsPermissions:(id)args
         TiThreadPerformOnMainThread(^(){
             CNContactStore *ourContactStore = [self contactStore];
             [ourContactStore requestAccessForEntityType:CNEntityTypeContacts completionHandler:^(BOOL granted, NSError *error) {
-                NSDictionary * propertiesDict = [TiUtils dictionaryWithCode:[error code] message:[TiUtils messageFromError:error]];
+                NSString* errorMessage = (granted) ? (nil) : (@"The user has denied access to the address book");
+                NSDictionary * propertiesDict = [TiUtils dictionaryWithCode:[error code] message:errorMessage];
```
hansemannn: Shouldn't we use the internal error message here?

iphone/Classes/CalendarModule.m
```diff
@@ -244,7 +244,9 @@ -(void) requestAuthorization:(id)args forEntityType:(EKEntityType)entityType
                                  completion:^(BOOL granted, NSError *error){
                                      NSDictionary* propertiesDict;
                                      if (error == nil) {
-                                         propertiesDict = [TiUtils dictionaryWithCode:(granted ? 0 : 1) message:nil];
+                                         NSString* errorMsg = (granted ? nil : @"The user has denied access to events in Calendar.");
+                                         propertiesDict = [TiUtils dictionaryWithCode:(granted ? 0 : 1)
```
hansemannn: The dictionary format is shifted here.

iphone/Classes/ContactsModule.m
sriks: This error msg is part of api response and hence not supposed to be displayed to end user. Because, we can essentially change this to a framework error msg in future when OS returns it to be.

iphone/Classes/CalendarModule.m
sriks: Not clear. Can you elaborate please.

hansemannn: Check the ``propertiesDict``, the TiUtils call is shifted (http://abload.de/img/bildschirmfoto2015-12p0jkn.png)

sriks: Because of objc-c styles method formatting. Changed to single line. Please check.

hansemannn:
PR looks good, will merge after the holidays. But please consider to unify the coding style for inline if-statements in the future. You used both
```
(condition) ? (@"true-condition") : (@"false-condition")
```
and
```
(condition ? @"true-condition" : @"false-condition")
```
mixed. The [usual best practise](http://stackoverflow.com/a/8760625/5537752) as part of the existing iOS source is
```
condition ? @"true-condition" : @"false-condition"
```
so no brackets at all. No critic, just a possible improvement :+1:

AngelkPetkov:
Will pass Jira review ones Hans merges the PR

sriks:
Actually I use braces when there is an expression evaluation condition. This improves readability and makes it safe. 

```objc
(x > y) ? (@"It works") : (@"Fails!")
```
But noted I should have been consistent for simple conditions.

PR #7529, opened 2015-12-02
Fix TC-5784 - setting a view layout to null on android crashes the app

xavierlacot:
This PR fixes the bug described in the ticket [TC-5784](https://jira.appcelerator.org/browse/TC-5784).

The `TiView` class constructor does not correctly check that the layout property is not null, and it crashes  when calling the `equals()` method. This PR assumes that, when a null layout is passed, the layout will default to a composite layout (the default value).


ashcoding:
@xavierlacot thanks for the PR.
Moved ticket to [TIMOB-20089](https://jira.appcelerator.org/browse/TIMOB-20089) for review.f

ashcoding:
Code reviewed and functionally tested. Request **Accepted**.

:+1:

PR #7534, opened 2015-12-02
[TIMOB-16355] iOS: SeparatorInsets should not be used for header/sect…

AngelkPetkov:
Setting the separatorInsets can now be done by 2 properties., rowSeparatorInsets and tableSeparatorInsets. If rowSeparatorInsets is not the cell insets will be set to the tableSeparatorInsets.

:JIRA: https://jira.appcelerator.org/browse/TIMOB-16355?filter=-1

### DIFF
--- a/iphone/Classes/TiUITableView.m
+++ b/iphone/Classes/TiUITableView.m
@@ -300,6 +300,7 @@
 		filterAnchored = NO; // defaults to false on search
 		searchString = @"";
 		defaultSeparatorInsets = UIEdgeInsetsZero;
+        rowSeparatorInsets = UIEdgeInsetsZero;
 	}
 	return self;
 }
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUITableView.m
```diff
@@ -2116,7 +2140,11 @@ - (UITableViewCell *)tableView:(UITableView *)ourTableView cellForRowAtIndexPath
     if ([TiUtils isIOS8OrGreater] && (tableview == ourTableView)) {
         [cell setLayoutMargins:UIEdgeInsetsZero];
     }
-	
+    
+    if (rowSeparatorInsets.left && rowSeparatorInsets.right != 0) {
```
cheekiatng: if ((rowSeparatorInsets.left != 0) && (rowSeparatorInsets.right != 0))

cheekiatng:
CR and FT done. Please address comments

iphone/Classes/TiUITableView.m
AngelkPetkov: Yeah i went or not and , my bad. I do the check because if the variable is not set that way whatever  
tableSeparatorInsets is set too the cell insets will be the same. That way we keep the original behavior as well as the new one.

cheekiatng:
APPROVED!

PR #7539, opened 2015-12-03
[TIMOB-20002]: iOS: Ti.Geolocation.requestLocationPermissions is not …

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-20002
This addresses the issue of invoking location permissions not in main thread.

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -869,7 +869,9 @@ MAKE_SYSTEM_PROP(ACTIVITYTYPE_OTHER_NAVIGATION, CLActivityTypeOtherNavigation);
                (currentPermissionLevel == kCLAuthorizationStatusAuthorized)) {
                 errorMessage = @"Cannot change already granted permission from AUTHORIZATION_ALWAYS to AUTHORIZATION_WHEN_IN_USE";
             } else if (!userDidApproveOrReject) {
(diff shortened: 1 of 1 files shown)


iphone/Classes/GeolocationModule.m
```diff
@@ -869,7 +869,9 @@ -(void)requestLocationPermissions:(id)args
                (currentPermissionLevel == kCLAuthorizationStatusAuthorized)) {
                 errorMessage = @"Cannot change already granted permission from AUTHORIZATION_ALWAYS to AUTHORIZATION_WHEN_IN_USE";
             } else if (!userDidApproveOrReject) {
-                [[self locationPermissionManager] requestWhenInUseAuthorization];
+                dispatch_async(dispatch_get_main_queue(), ^{
```
AngelkPetkov: Should use                 TiThreadPerformOnMainThread(^{
                    [[self locationPermissionManager] requestWhenInUseAuthorization];
                }, NO); instead. As it does a couple of checks regarding the main thread to prevent  crashes.

AngelkPetkov:
Closing PR because it doesn't fix the whole of the issue, sorry. The callback will still get called regardless if the permissions are undetermined or false.

AngelkPetkov:
Closing the PR sorry ,because it doesn't fully fix the issue, the callback will still print Success even if we disallow  access.

PR #7541, opened 2015-12-03
[TIMOB-20002] iOS: RequestLocationPermissions is not always called?

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-20002
The callback will not be called before the user allows or denys premissions. The callback now returns the correct value 1 ("False') if the premissions are not granted

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -859,14 +859,14 @@ MAKE_SYSTEM_PROP(ACTIVITYTYPE_OTHER_NAVIGATION, CLActivityTypeOtherNavigation);
     
     CLAuthorizationStatus requested = [TiUtils intValue: value];
     CLAuthorizationStatus currentPermissionLevel = [CLLocationManager authorizationStatus];
-    
-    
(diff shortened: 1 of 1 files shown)


iphone/Classes/GeolocationModule.m
```diff
@@ -859,8 +859,17 @@ -(void)requestLocationPermissions:(id)args
...
-    // Corelocation shows permission alert only in undetermined case.
-    BOOL userDidApproveOrReject = (currentPermissionLevel != kCLAuthorizationStatusNotDetermined);
+    
+    
+    BOOL permissionsGranted = (currentPermissionLevel == kCLAuthorizationStatusAuthorizedAlways) || (currentPermissionLevel == kCLAuthorizationStatusAuthorizedWhenInUse);
+    if (permissionsGranted) {
+        [self executeAndReleaseCallbackWithCode:0 andMessage:nil];
+        return;
+    } else if (currentPermissionLevel == kCLAuthorizationStatusDenied) {
+        [self executeAndReleaseCallbackWithCode:1 andMessage:nil];
```
pec1985: The message should be "The user denied access to use location services."

pec1985:
Left one commend, address it and I'll merge it. Make sure the fix also goes into the back port PR

pec1985:

PR #7548, opened 2015-12-04
[TIMOB-20030] iOS: Removing repeated eventListeners

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-20030?filter=-1

### DIFF
--- a/iphone/Classes/KrollObject.m
+++ b/iphone/Classes/KrollObject.m
@@ -1348,8 +1348,7 @@ TI_INLINE TiStringRef TiStringCreateWithPointerValue(int value)
 		TiValueRef jsCallbackArrayLength = TiObjectGetProperty(jsContext, jsCallbackArray, kTiStringLength, &exception);
 		int arrayLength = (int)TiValueToNumber(jsContext, jsCallbackArrayLength, &exception);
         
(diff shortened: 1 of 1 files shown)


iphone/Classes/KrollObject.m
```diff
@@ -1347,7 +1347,17 @@ -(void)storeListener:(id)eventCallbackOrWrapper forEvent:(NSString *)eventName
 	{
 		TiValueRef jsCallbackArrayLength = TiObjectGetProperty(jsContext, jsCallbackArray, kTiStringLength, &exception);
 		int arrayLength = (int)TiValueToNumber(jsContext, jsCallbackArrayLength, &exception);
-		
+        
+		double lenght = TiValueToNumber(jsContext, jsCallbackArrayLength, NULL);
```
cheekiatng: isn't this the same as arrayLength that was previously declared? please explain why this has to be declared again?
and length has to be spelt correctly.

cheekiatng:
FT passed. Please address comments.

iphone/Classes/KrollObject.m
pec1985: Yep, it's the same.... :D

cheekiatng:
CR and FT Passed. approved!

PR #7554, opened 2015-12-08
[TIMOB-20028] Layout fails on iOS8 with iOS9-only Picker height set

AngelkPetkov:
[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-20028?filter=-1)


pec1985:
PR reviewed and approved

PR #7557, opened 2015-12-08
[TIMOB-20118] iOS: Contacts property fullName does not return appropr…

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-20118
Re-arranged the order of the strings, removed organization from full name and updated docs.


pec1985:
Reviewed. **APPROVED!!**

PR #7560, opened 2015-12-08
[TIMOB-20118] iOS (5_1_2): Contacts property fullName does not return…

AngelkPetkov:
:JIRA: https://jira.appcelerator.org/browse/TIMOB-20118

### DIFF
--- a/apidoc/Titanium/Contacts/Person.yml
+++ b/apidoc/Titanium/Contacts/Person.yml
@@ -131,6 +131,9 @@ properties:
     
   - name: fullName
     summary: Localized full name of the person. Single value. Read-only on Android.
+    description: | 
+        The full name will be printed in the order of.
+        Full name, middle name, family name, suffix.
     type: String
     permission: read-only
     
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiContactsPerson.m
```diff
@@ -477,18 +477,19 @@ -(NSString*)fullName
 		if ([person.namePrefix length]) {
 			[compositeName appendFormat:@"%@ ", person.namePrefix];
 		}
-		if ([person.nameSuffix length]) {
-			[compositeName appendFormat:@"%@ ", person.nameSuffix];
-		}
-		if ([person.organizationName length]) {
```
hansemannn: Why did we remove the `organisationName` property?

apidoc/Titanium/Contacts/Person.yml
```diff
@@ -131,6 +131,9 @@ properties:
     
   - name: fullName
     summary: Localized full name of the person. Single value. Read-only on Android.
+    description: | 
+        The full name will be printed in the order of.
+        Full name, middle name, family name, suffix.
```
hansemannn: Shouldn't it be "First name" instead of "Full name"?

PR #7588, opened 2015-12-16
[TIMOB-19486] Android: Generate missing app icons based on DefaultIcon.png

feons:
https://jira.appcelerator.org/browse/TIMOB-19486

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2555,32 +2555,22 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
 			defaultIconChanged = prevDefaultIconHash !== currDefaultIconHash;
 		}
 
-		function copyGenAppIcons() {
-			// copy generated app icons from ../build/android/bin/assets/res to ../build/android/res
-			var source = path.join(_t.buildBinAssetsDir, 'res');
-			if (fs.existsSync(source)) {
-				copyDir.call(_t, {
-					src: source,
-					dest: _t.buildResDir
-				}, cb);
-			} else {
-				cb();
-			}
-		}
-
 		Object.keys(lookup).forEach(function (key) {
 			var icon = lookup[key],
 				source,
 				dest,
+				destFinal,
 				description,
 				resDrawable;
 
 			if (key === 'default') {
 				dest = path.join(this.buildBinAssetsResourcesDir, this.tiapp.icon);
+				destFinal = dest;
 				description = path.join('Resources', 'android', this.tiapp.icon);
 			} else {
 				resDrawable = path.join('res', 'drawable' + '-' + key, this.tiapp.icon);
 				dest = path.join(this.buildBinAssetsDir, resDrawable);
+				destFinal = path.join(this.buildDir, resDrawable);
 				description = path.join('platform', 'android', resDrawable);
 			}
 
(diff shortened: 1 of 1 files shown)


android/cli/commands/_build.js
```diff
@@ -2510,6 +2510,113 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+				defaultIcon = icon;
+				return true;
+			}
+		});
+
+		if (defaultIcon) {
+			var currDefaultIconHash = this.hash(fs.readFileSync(defaultIcon)),
+				defaultIconRelative = path.relative(this.projectDir, defaultIcon),
+				prevDefaultIcon = previousBuildManifestFiles && previousBuildManifestFiles[defaultIconRelative],
+				prevDefaultIconHash = prevDefaultIcon && prevDefaultIcon.hash,
```
cb1kenobi: Comma should be a semi-colon.

android/cli/commands/_build.js
```diff
@@ -2510,6 +2510,113 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+
+			this.currentBuildManifestFiles[defaultIconRelative] = {
+				hash: currDefaultIconHash
+			};
+
+			defaultIconChanged = prevDefaultIconHash !== currDefaultIconHash;
+		}
+
+		function copyGenAppIcons() {
```
cb1kenobi: This is sort of confusing. Maybe we just inline this function in the `this.generateAppIcons()` call?

android/cli/commands/_build.js
```diff
@@ -2510,6 +2510,113 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
 		}
 	}, this);
 
+	// genereate missing app icons
+	tasks.push(function (cb) {
+		this.currentBuildManifestFiles = this.currentBuildManifestFiles || {};
```
cb1kenobi: Android build doesn't use differential builds... yet. There is no need for `this.currentBuildManifestFiles`. All references should be removed.

android/cli/commands/_build.js
```diff
@@ -2510,6 +2510,113 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+			// the app icon was previously resized
+			if (!defaultIconChanged && fs.existsSync(dest)) {
+				this.logger.trace(__('Found %sx%s app icon: %s', icon.width, icon.height, dest.cyan));
+				return;
+			}
+
+			// generated icons will be placed under ../build/android/bin/assets/res/drawable-[density]
+			fs.existsSync(path.dirname(dest)) || wrench.mkdirSyncRecursive(path.dirname(dest));
+			missingIcons.push({
+				description: __('%s', description),
```
cb1kenobi: Not sure why you can't just set `description: description,`.

android/cli/commands/_build.js
feons: That would mean, we will regenerate the app icons on every build/rebuild?

android/cli/commands/_build.js
feons: me neither :s

android/cli/commands/_build.js
cb1kenobi: Part of the problem is `this.currentBuildManifestFiles` is something you created. It doesn't exist before this PR. It is never persisted to disk. So, it has no effect.

The way I see it, you can either remove `this.currentBuildManifestFiles` in which case the missing icons will be generated every build -or- hack something to track this state.

Simply checking if the destination image exists would be a step in the right direction. Combine that with checking the mtime on the source DefaultIcon*.png with the mtime of the generated images and you should have an OK mechanism.

cheekiatng:
APPROVED

cheekiatng:
Sorry I thought it was addressed. Reverting PR. I would prefer checking the mtime mechanism.

cheekiatng:
@feons can you create another PR that addresses that issue? I'll reopen the JIRA ticket.

android/cli/commands/_build.js
```diff
@@ -2510,6 +2510,120 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+		this.currentBuildManifestFiles = this.currentBuildManifestFiles || {};
+
+		var lookup = {
+				// ../Resources/android/appicon.png
+				'default':	{ height: 128, width: 128, required: true },
+				// ../platform/android/res/drawable-[key]/appicon.png
+				'ldpi':		{ height: 36, width: 36, required: false },
+				'mdpi':		{ height: 48, width: 48, required: false },
+				'hdpi':		{ height: 72, width: 72, required: false },
+				'xhdpi':	{ height: 96, width: 96, required: false },
```
cb1kenobi: Missing padding between `:` and value.

PR #7593, opened 2015-12-18
[TIMOB-20156] iOS: Return typedText value for Remote Notifications Text Actions

realmikebutler:
5.1 added support for TextActions.  Works fine for local notifications, unfortunately for Remote notifications the typedText value is ignored and not accessible from Javascript.   This pull request simply takes the code that was added for returning 'typedText' in local notifications and adds it to the remote notification callback.

### DIFF
--- a/iphone/Classes/TiApp.m
+++ b/iphone/Classes/TiApp.m
@@ -496,7 +496,7 @@ TI_INLINE void waitForMemoryPanicCleared();   //WARNING: This must never be run 
 	completionHandler();
 }
 
-- (void) application:(UIApplication *)application handleActionWithIdentifier:(NSString *)identifier forRemoteNotification:(NSDictionary *)userInfo completionHandler:(void (^)())completionHandler {
+- (void) application:(UIApplication *)application handleActionWithIdentifier:(NSString *)identifier forRemoteNotification:(NSDictionary *)userInfo withResponseInfo:(NSDictionary *)responseInfo completionHandler:(void (^)())completionHandler {
     RELEASE_TO_NIL(remoteNotification);
     [self generateNotification:userInfo];
     NSMutableDictionary *event = [[NSMutableDictionary alloc] init];
@@ -508,10 +508,18 @@ TI_INLINE void waitForMemoryPanicCleared();   //WARNING: This must never be run 
     if (category != nil) {
         event[@"category"] = category;
     }
+
+#if IS_XCODE_7
+    if([TiUtils isIOS9OrGreater] == YES) {
+        [event setValue:responseInfo[UIUserNotificationActionResponseTypedTextKey] forKey:@"typedText"];
+    }
+#endif
+    
     [[NSNotificationCenter defaultCenter] postNotificationName:kTiRemoteNotificationAction object:event userInfo:nil];
     [event autorelease];
     completionHandler();
 }
+
 
 
 #pragma mark Apple Watchkit handleWatchKitExtensionRequest


realmikebutler:
This PR simply adds 5 lines the TiApp.m  file.  Why would a tiny change to an Objective C file cause problems building for Android?

hansemannn:
Hi @realmikebutler, CR looks good! I also moved a JIRA ticket for you so our QE can verify and close the ticket afterwards (https://jira.appcelerator.org/browse/TIMOB-20156). 

Can you please provide a few lines of code so it's easier to test the changes? Thanks!

realmikebutler:
Not sure it'd be that easy to do in a couple of lines - you'd have to set up something to push a notification through the APNS

hansemannn:
@realmikebutler Did you see the code snippet I provided above?

hansemannn:
@realmikebutler Please address the comments above in the next 7 days to get this merged, thank you!

ludolphus:
I've tested this. It does not work when the app is not already loaded in background and you have activationMode set to Ti.App.iOS.USER_NOTIFICATION_ACTIVATION_MODE_BACKGROUND This should really work because that's the whole point of background interactive notifications: they come in while you are in another app and the app that the incoming notification belongs to does not necessarily have to be in memory at that time. See also https://jira.appcelerator.org/browse/TC-5830

hansemannn:
@ludolphus Did you test my proposed solution in addition to this PR (which is not merged, yet)?

ludolphus:
@hansemannn yes I used yours. It works when the app was already loaded before the notification comes in, meaning the app is visible in application switcher and still 'active' in memory.
It does not work when app is not in memory. This is more an issue of handling the notifications in that situation, the typedText feature works when app is in mem, like all other interactive notifications work.
I tested this with push notifications as well as with local notifications. Same issue with local notifications: when app not in memory it doesn't work. See test case in jira ticket I mentioned in my previous comment.

hansemannn:
As far as I know, it's not supposed to work with background notifications. If you want to use the typed text, you need to use the foreground mode. Same in native apps.

ludolphus:
Works for WhatsApp just fine for example.

hansemannn:
Because the action is configured in the foreground mode I guess. So you can still receive them from the background / notification center / banner. If you think it's still a bug, I'd be happy to have you filed a bug in JIRA with a reproducible test case with the background mode. Is that suitable?

ludolphus:
After using quick reply in those apps, they do not go to foreground and are not visible in application switcher but the reply is sent to the other party.

I already created a Jira ticket for this a couple of days ago https://jira.appcelerator.org/browse/TC-5830 But it got closed without a solution. I even included a Gif movie showing what happens with the test app in all cases. The test case uses local notifications, but like I said the background actions there also do not work when app is not already in memory.

Thanks for looking into this.

ludolphus:
Telegram is open source, as you can see here https://github.com/peter-iakovlev/Telegram/blob/316bfd9598400d4571f62b80bcccce65eb3bd0ea/Telegraph/TGAppDelegate.mm#L1701 they also use a background activation mode for the textInput behavior.

hansemannn:
Alright. So we need to figure out, how we can receive the notifications in background mode, even when the app was not opened before. Let me do some research on how it's done natively. Thanks!

ludolphus:
Yes, awesome! Thanks for looking into this. Btw you can also reach me on TiSlack, I'll ping you in a DM

hansemannn:
Closing this PR in favor of the discussed solution being implemented the next few weeks. Thanks!

hansemannn:
Hey @ludolphus, I want to fix this asap; this week! I noticed that the `application:handleActionWithIdentifier:identifier:userInfo:completionHandler ` should definitely get called when the app is in the background. The event triggered from there is `remotenotificationaction`, would that one be the appropriate one to expect the value? And what others as well? `localnotificationaction` ? If I am correct, that one is already fired correctly.

hansemannn:
Update: New PR is [here](https://github.com/appcelerator/titanium_mobile/pull/9033), check it out!

ludolphus:
typedText is now available for remote notifications. But the app already has to be started and available in memory. Won't work when the app is not in memory and a remote push notification comes in and the user uses this option to reply by typing in the text area of the notification. This is another issue of course, but without it having typedText and interactive notifications for remote notifications is almost useless (from user point of view, because it won't work all the time).

PR #7594, opened 2015-12-20
[TIMOB-20055] Handle nulled preview in proxy

hansemannn:
Improvements to prevent the preview view on peek, if the preview is set to null. Use case: We want the preview only be shown in a tableview, if the cell contains an image to be previewed. This improvements make that possible. Before, a blank preview would have been shown.

Addition to [this JIRA](https://jira.appcelerator.org/browse/TIMOB-20055).


pec1985:
PR reviewed and approved.

PR #7596, opened 2015-12-21
[TIMOB-19241] iOS: Change the default height and width of the textField.

AngelkPetkov:
Changed default height and width to 100 if their not initialized.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19241?filter=-1)


hansemannn:
Why should we force a height and width on the textfield? It would break tons of existing apps and act against the native behavior of iOS. Please consider.

pec1985:
This is not a fix. PR rejected.

PR #7597, opened 2015-12-22
[TIMOB-20165] iOS Ti Geolocation callback thread issue.

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-20165

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -1103,15 +1103,15 @@ MAKE_SYSTEM_PROP(ACTIVITYTYPE_OTHER_NAVIGATION, CLActivityTypeOtherNavigation);
                 errorStr = @"The user denied access to use location services.";
         }
         
-        NSMutableDictionary * propertiesDict = [TiUtils dictionaryWithCode:code message:errorStr];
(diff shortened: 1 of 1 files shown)


pec1985:
I don't it's necessary to call it on the main thread since event queue is thread safe. You may also want to test the new `[KrollCallback callAsync:thisObject:]` method.

I will test it as is and without the `TiThreadPerformOnMainThread` function and see what happens.

sriks:
@pec1985 

I did a quick experiment with ``[KrollCallback callAsync:thisObject:]`` but that don't seem to be working as the context is different. But invoking event is thread safe so it don't need to explicitly running on main thread. 

I would like to wait for you to test and confirm. Do define ``TI_USE_KROLL_THREAD``

collinprice:
I believe @sriks is correct in saying that the callback needs to be invoked on the main thread. I am getting the following warning, which causes an error, when I attempt to create a window in the callback.

    [WARN] Creating [object TiUIView] in a different context than the calling function.

collinprice:
@pec1985 After looking through the source the method `[KrollCallback callAsync:thisObject:]` was only added to `master` and does not exist in 5.1.X so in this case it cannot be used to fix this bug.

pec1985:

PR #7598, opened 2015-12-23
[TIMOB-17951] Handle undefined parameter in Ti.Media.switchCamera

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-17951

### DIFF
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -969,6 +969,8 @@ MAKE_SYSTEM_PROP(VIDEO_TIME_OPTION_EXACT,MPMovieTimeOptionExact);
 
 -(void)switchCamera:(id)args
 {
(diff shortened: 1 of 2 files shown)


iphone/Classes/MediaModule.m
```diff
@@ -969,6 +969,11 @@ -(void)stopVideoCapture:(id)args
 
 -(void)switchCamera:(id)args
 {
+    // TIMOB-17951
+    if ([args objectAtIndex:0] == [NSNull null]) {
```
pec1985: Make sure that `args` is not `nil` and that it is an `NSArray`

hansemannn: `ENSURE_TYPE(args, NSArray)` should validate it properly now.

PR #7600, opened 2015-12-27
[TIMOB-20178] Support Peek and Pop in Ti.UI.WebView

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20178

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -322,7 +322,7 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 	return url;
 }
 
--(id)setAllowsLinkPreview_:(id)value
+-(void)setAllowsLinkPreview_:(id)value
 {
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIWebView.m
```diff
@@ -322,6 +322,17 @@ -(id)url
...
+{
+#if IS_XCODE_7
+    if ([TiUtils isIOS9OrGreater] == NO) {
+        return;
+    }
+    ENSURE_TYPE(value, NSNumber);
+    [webview setAllowsLinkPreview:[TiUtils boolValue:value]];
+#endif
+}
+
```
pec1985: Return type is `id` but returns nothing, this will crash on device

hansemannn: Didn't crash on the iPhone 6S test device, but of course it needs to be `void`. Fixed.

pec1985:
PR reviewed. **APPROVED!**

PR #7608, opened 2016-01-04
[TIMOB-19773] iOS: include catching exception when enabling analytics

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19773


cheekiatng:
Approved!

PR #7609, opened 2016-01-04
[TIMOB-18411] iOS: A ListView's SearchBar doesn't filter when program…

sriks:
JIRA https://jira.appcelerator.org/browse/TIMOB-18411


hansemannn:
PR approved!

PR #7612, opened 2016-01-05
[TIMOB-20027] Enable App-thinning Splicing by default for new projects

feons:
https://jira.appcelerator.org/browse/TIMOB-20027


cheekiatng:
CR and FT passed. APPROVED!

PR #7615, opened 2016-01-06
[TIDOC-2413] Docs: Explain the images property properly

Topener:
In the documentation it is not clear when using the `images` property, a `start` method needs to be called. This is very useful and saves a lot of time when documented properly


hansemannn:
Approved!

PR #7553, opened 2015-12-07
[TIMOB-20087] iOS - Fixed regression with ScrollableView

pec1985:
https://jira.appcelerator.org/browse/TIMOB-20087

PR #7617, opened 2016-01-06
[TIMOB-20196] iOS: ScrollableView is not possible to change views

AngelkPetkov:
:JIRA:https://jira.appcelerator.org/browse/TIMOB-20196?filter=-1

### DIFF
--- a/iphone/Classes/TiUIScrollableViewProxy.m
+++ b/iphone/Classes/TiUIScrollableViewProxy.m
@@ -71,7 +71,7 @@
 
 -(void)setViews:(id)args
 {
-#ifndef TI_USE_AUTOLAYOUT
+#ifdef TI_USE_AUTOLAYOUT
     ENSURE_UI_THREAD(setViews, args)
 #endif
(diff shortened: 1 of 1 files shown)


pec1985:
Good catch! That also fixes another bug. This PR #7553 is wrong and needs to be corrected. If you could, do these changes:
From
```
#ifndef TI_USE_AUTOLAYOUT
    ENSURE_UI_THREAD(setViews, args)
#endif
```
to 
```
#ifdef TI_USE_AUTOLAYOUT
    ENSURE_UI_THREAD(setViews, args)
#endif
```
And then move this line to the end of the function
```
[self replaceValue:args forKey:@"views" notification:YES];
```
Let me know if you have questions

pec1985:
Fixes https://jira.appcelerator.org/browse/TIMOB-20138 as well

AngelkPetkov:
Thanks! Addressed comments.

pec1985:
Thanks!

PR #7622, opened 2016-01-08
[TIMOB-19833] iOS: Allow users to access images via Ti.Filesystem when app thinning is enabled

cheekiatng:
JIRA: https://jira.appcelerator.org/browse/TIMOB-19833


AngelkPetkov:
FT and CT approved. Merging. Looks like the travis fail is because there is no unit test for android.

PR #7628, opened 2016-01-12
Fix: "ti clean" does not scan SDK hooks

infosia:
Fix for [TIMOB-20085](https://jira.appcelerator.org/browse/TIMOB-20085)

Currently there's no chance for Titanium SDK to load "hooks" while `ti clean` even when you subscribe `clean.post` event. It is required from [Windows SDK](https://github.com/appcelerator/titanium_mobile_windows/pull/505) to load hooks before `ti clean` because it uses special directory for Visual Studio.

This PR basically gives platform-SDK a chance to load hooks before `ti clean`, via `scanHooks` function in `_clean.js`. platform-SDK that needs clean-hook should scan hooks in `cli/commands/_clean.js`, and then subscribe `clean.pre`, `clean.post` and so on.

_cli/commands/_clean.js_

``` javascript
var path = require('path');

exports.scanHooks = function(logger, config, cli) {
    cli.scanHooks(path.join(__dirname, '..', 'hooks'));
}
```

_cli/hooks/clean.js_

``` javascript
 exports.init = function (logger, config, cli) {
    cli.on('clean.post', {
        post: function (builder, finished) {
... // platform-specific clean-up
            finished();
        }
    });
};
```

See discussion in https://github.com/appcelerator/titanium_mobile_windows/pull/505 for more details.

### DIFF
--- a/cli/commands/clean.js
+++ b/cli/commands/clean.js
@@ -89,6 +89,8 @@ exports.run = function (logger, config, cli) {
 	if (cli.argv.platforms) {
 		async.series(cli.argv.platforms.map(function (platform) {
 			return function (next) {
+				// scan platform SDK specific clean hooks
+				cli.scanHooks(path.join(__dirname, '..', '..', platform, 'cli', 'hooks'));
 				cli.fireHook('clean.pre', function () {
 					cli.fireHook('clean.' + platform + '.pre', function () {
 						var dir = path.join(buildDir, platform);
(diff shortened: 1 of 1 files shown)


cli/commands/clean.js
```diff
@@ -89,6 +89,11 @@ exports.run = function (logger, config, cli) {
 	if (cli.argv.platforms) {
 		async.series(cli.argv.platforms.map(function (platform) {
 			return function (next) {
+				// scan platform SDK specific clean hooks
```
sgtcoolguy: Do we really need to look for a platform-specific cli/commands/_clean.js, load that and call scanHooks on it? Shouldn't we just call:
```javascript
cli.scanHooks(path.join(__dirname, '..', '..', platform, 'cli', 'hooks'));
```

Basically just say load all hooks from the platform specific cli hooks directory.

cli/commands/clean.js
```diff
@@ -117,6 +122,16 @@ exports.run = function (logger, config, cli) {
 	} else if (appc.fs.exists(buildDir)) {
 		logger.debug(__('Deleting all platform build directories'));
 
+		// scan platform SDK specific clean hooks
+		if (ti.targetPlatforms) {
+			ti.targetPlatforms.forEach(function(platform) {
+				var platformCleanModule = path.join(__dirname, '..', '..', platform, 'cli', 'commands', '_clean.js');
```
sgtcoolguy: same as above. I don't think we need the special _clean.js middle-man for loading hooks.

cli/commands/clean.js
infosia: Yeah, I just wanted to make it flexible for platform SDK to choose hook directory but I might be thinking too much.

PR #7634, opened 2016-01-14
[TIMOB-19826] Added error code and moved validation code to Analytics …

ashcoding:
…class
Jira: https://jira.appcelerator.org/browse/TIMOB-19826

### DIFF
--- a/android/modules/analytics/src/java/ti/modules/titanium/analytics/AnalyticsModule.java
+++ b/android/modules/analytics/src/java/ti/modules/titanium/analytics/AnalyticsModule.java
@@ -7,15 +7,16 @@
 package ti.modules.titanium.analytics;
 
 import java.util.HashMap;
+import java.util.Iterator;
 
 import org.appcelerator.kroll.KrollDict;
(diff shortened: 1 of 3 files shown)


ashcoding:
Code was moved out of TiConvert as it is something that is only used in Analytics. Hence moved to Analytics namespace.

zackehh:
Out of interest, is there any reason the SDK uses `org.json` and `JSONObject` rather than using [Jackson](https://github.com/FasterXML/jackson-core)? Is it just to avoid the external dependency? 

We built a module which is useful for these types of validations for PEM; https://github.com/zackehh/dot-notes-java but it requires Jackson JSON rather than JSONObject.

ashcoding:
You are right. It is to not depend on an external dependency (and prevent bloat in Android).

Android has a 65K Reference Limit (methods) that Titanium is limited by (at the moment).

sriks:
FT Passed.

zackehh:
@ashcoding aw, that's lame. Makes sense though :) thanks.

PR #7649, opened 2016-01-29

FokkeZB:


feons:
@FokkeZB, this also need to be backported.

FokkeZB:
Here's the BP: https://github.com/appcelerator/titanium_mobile/pull/7669

cb1kenobi:
This PR has problems and needs to be fixed. It should not have been merged.

You still need to check if `player` is truthy in every `if` statement. The path you are setting is OS X specific, so just compare to `darwin`. Use `!==` instead of `!=`.

```javascript
if (process.platform === 'darwin' && (!player || !fs.existsSync(player))) {
    player = path.join(dir, 'player.app', 'Contents', 'MacOS', 'player');
}
if (player && !fs.existsSync(player)) {
    player = null;
}
```

PR #7656, opened 2016-01-31
[TIMOB-20217] iOS: Ti.Geolocation.hasGeolocationPermission() and Ti.Geolocation.getCurrentPosition() are not working on IOS 7 - fix for TI_USE_KROLL_THREAD

sriks:
This requires that creation of location delegate and subsequent permission alert are presented on main thread. This is because by default JS runs on its own thread enabled by `TI_USE_KROLL_THREAD`.

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -869,7 +869,10 @@ MAKE_SYSTEM_PROP(ACTIVITYTYPE_OTHER_NAVIGATION, CLActivityTypeOtherNavigation);
 -(void)requestLocationPermissions:(id)args
 {
     if (![TiUtils isIOS8OrGreater]) {
-        [self requestLocationPermissioniOS7:args];
+        // It is required that delegate is created and permission is presented in main thread.
+        TiThreadPerformOnMainThread(^{
+            [self requestLocationPermissioniOS7:args];
+        }, NO);
         return;
     }
     


iphone/Classes/GeolocationModule.m
```diff
@@ -869,7 +869,10 @@ - (void)requestLocationPermissioniOS7:(id)args {
 -(void)requestLocationPermissions:(id)args
 {
     if (![TiUtils isIOS8OrGreater]) {
-        [self requestLocationPermissioniOS7:args];
+        // It is required that delegate is created and permission is presented in main thread.
+        TiThreadPerformOnMainThread(^{
+            [self requestLocationPermissioniOS7:args];
+        }, NO);
```
hansemannn: I don't see any dependency to the `TI_USE_KROLL_THREAD`, did you test it with both `<run-on-mainthread/>` and without? Just want to double-check.

sriks: Actually when we run on Studio the JS is executed in a separate thread. `<run-on-mainthread/>` is false by default. But the permission are required to be presented on main thread. 

But when we run in XCode `TI_USE_KROLL_THREAD` is not defined, which makes JS run on main thread. So it very easy to miss on background thread related issues. This fix is one of them.
This can be tested by defining and un-defining `TI_USE_KROLL_THREAD` in defines.h. Should work in both cases.

hansemannn: Alright. But if thats the case, we definitely need to make the Xcode builds behave the same as the actual `ti build -p ios` builds. Would it make sense to define it by default in the Xcode project? Should be the same with `<use-autolayout/>` that triggers a different behavior.

sriks: I highly recommend that `TI_USE_KROLL_THREAD` is defined by default in our code base. This makes the Xcode build behave as expected. Actually it is quite easy to miss these thread related issues :(
About this got to check with Kiat.

hansemannn: Yes. At least if we keep the main thread enabled by default. But as far as I know, I should get enabled by default starting in 6.0.0 in favor of Hyperloop. /cc @cheekiatng

cheekiatng: Sure we can make TI_USE_KROLL_THREAD default for now in our code base. However, there's a blocker for setting it to default starting in 6.0.0 due to https://jira.appcelerator.org/browse/TIMOB-19762.

hansemannn:
As discussed in the standup, we'll go with this fix and have the `TI_USE_KROLL_THREAD` in mind for 6.0.0. Approved!

PR #7657, opened 2016-01-31
[TIMOB-20294] Resolve Xcode compiler warnings

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20294

### DIFF
--- a/iphone/Classes/FilesystemModule.m
+++ b/iphone/Classes/FilesystemModule.m
@@ -24,7 +24,7 @@ extern NSString * TI_APPLICATION_RESOURCE_DIR;
 {
 	if ([arg isKindOfClass:[TiFilesystemFileProxy class]])
 	{
-		return [arg path];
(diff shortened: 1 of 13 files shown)


sriks:
@hansemannn Its good to see the warnings are addressed!
I understand most of them has to do with proper casting. But can I get some test cases to make sure we did not miss any edge cases ?

hansemannn:
@sriks Unfortunetely, we would need to track it back to the proxies that call the methods. I can setup tests today and provide them. Since it's 6.0.0 timeline anyway, we have enough time!

sriks:
@hansemannn Will kitchensink serve for now ?

hansemannn:
@sriks I write test cases today to exactly trigger all methods being used there. Although I'm pretty sure everything works, that's the most save way to test it.

hansemannn:
The test cases can be found in the JIRA ticket.

PR #7669, opened 2016-02-02

FokkeZB:

Backport for https://github.com/appcelerator/titanium_mobile/pull/7649

### DIFF
--- a/node_modules/titanium-sdk/lib/emulators/genymotion.js
+++ b/node_modules/titanium-sdk/lib/emulators/genymotion.js
@@ -286,7 +286,7 @@ function findGenymotion(dir, config, callback) {
 	if (!player || !fs.existsSync(player)) {
 		player = path.join(dir, 'player' + exe);
 	}
-	if (!fs.existsSync(player) && process.platform != 'win32') {
+	if (!fs.existsSync(player) && process.platform === 'darwin') {
 		player = path.join(dir, 'player.app', 'Contents', 'MacOS', 'player');
 	}
 	if (!fs.existsSync(player)) {
--- a/node_modules/titanium-sdk/lib/emulators/genymotion.js
+++ b/node_modules/titanium-sdk/lib/emulators/genymotion.js
@@ -286,7 +286,10 @@ function findGenymotion(dir, config, callback) {
 	if (!player || !fs.existsSync(player)) {
 		player = path.join(dir, 'player' + exe);
 	}
-	if (player && !fs.existsSync(player)) {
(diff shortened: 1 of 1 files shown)


node_modules/titanium-sdk/lib/emulators/genymotion.js
```diff
@@ -286,7 +286,10 @@ function findGenymotion(dir, config, callback) {
 	if (!player || !fs.existsSync(player)) {
 		player = path.join(dir, 'player' + exe);
 	}
-	if (player && !fs.existsSync(player)) {
+	if (!fs.existsSync(player) && process.platform != 'win32') {
```
cb1kenobi: You still need to check if `player` is truthy in every `if` statement. The path you are setting is OS X specific, so just compare to `darwin`. Use `!==` instead of `!=`.

```javascript
if (process.platform === 'darwin' && (!player || !fs.existsSync(player))) {
    player = path.join(dir, 'player.app', 'Contents', 'MacOS', 'player');
}
if (player && !fs.existsSync(player)) {
    player = null;
}
```

cb1kenobi:
Since this is the backport, I suggest fixing master, then backporting both this commit and the fix.

node_modules/titanium-sdk/lib/emulators/genymotion.js
FokkeZB: I get the darwin fix, but why the truthy check? After line https://github.com/FokkeZB/titanium_mobile/blob/patch-37/node_modules/titanium-sdk/lib/emulators/genymotion.js#L287 it will always be truthy since `path.join()` does not resolve and just joins the arguments right?

FokkeZB: I [blame](https://github.com/FokkeZB/titanium_mobile/blame/patch-37/node_modules/titanium-sdk/lib/emulators/genymotion.js#L25) you for the `==` btw, thought I should have corrected it indeed ;)

cb1kenobi: I'm fine with omitting the `player` truthiness checks. At least now it works on Linux.

I've slowly been migrating `==` to `===`. There's a lot of code. :)

Can you also please fix master?

FokkeZB: Well, it works on Linux as it was since the `fs.existsSync` on the next line will simply return false, but here's the, uh...  "forward port" for master:

https://github.com/appcelerator/titanium_mobile/pull/7670

cb1kenobi:
Thank you! **APPROVED**

PR #7670, opened 2016-02-02

FokkeZB:

A forward port (?) for a fix in https://github.com/appcelerator/titanium_mobile/pull/7649


cb1kenobi:
Thank you! **APPROVED**

PR #7677, opened 2016-02-04
[TIMOB-2000]_typoFIX iOS:

AngelkPetkov:


hansemannn:
The ticket does match the actual problem: https://jira.appcelerator.org/browse/TIMOB-2000

PR #7680, opened 2016-02-04
[TIMOB-20260] Android: Android M Camera Overlay crash fix

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-20260
Test project in Jira.

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
@@ -34,6 +34,7 @@ import android.hardware.Camera.PictureCallback;
 import android.hardware.Camera.ShutterCallback;
 import android.hardware.Camera.Size;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.v7.app.ActionBar;
 import android.view.Gravity;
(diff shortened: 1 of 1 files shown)


android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
```diff
@@ -495,24 +496,33 @@ private static File writeToFile(byte[] data, boolean saveToGallery) throws Throw
...
+	                .equals(Parameters.FOCUS_MODE_INFINITY))) {
+	            AutoFocusCallback focusCallback = new AutoFocusCallback()
+	            {
+	                public void onAutoFocus(boolean success, Camera camera)
+	                {
+	                    camera.takePicture(shutterCallback, null, jpegCallback);
+	                    if (!success) {
+	                        Log.w(TAG, "Unable to focus.");
+	                    }
+	                    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
hieupham007: I think we should leave a comment here saying "cancelAutoFocus" causes the camera to crash on M (probably due to discontinued support of android.hardware.camera) and we should move to android.hardware.camera2.

hieupham007:

PR #7682, opened 2016-02-05
[TIMOB-20038] Android: TableView crash fix

ashcoding:
Jira: https://jira.appcelerator.org/browse/TIMOB-20038

To test, change numberOfRows to be lesser than maxClassname. If maxClassname crashes when numberOfRows are more than maxClassname, this is expected. numberOfRows has to be lesser than maxClassname.

Test code:

``` javascript
var section = Ti.UI.createTableViewSection();
var numberOfRows = 4;   
for (var i=0; i < numberOfRows; i++)     
{    
    //fill section with rows    
    section.add(Ti.UI.createTableViewRow({
        height: 50,
        width: Ti.UI.FILL,
        backgroundColor: 'white',
        color:'black',
        className: 'row' + i,
        title: 'row' + i
    }));
};

var table = Ti.UI.createTableView({
    top: 0,
    left: 0,
    maxClassname: 4,
    height: Ti.UI.FILL,
    width: Ti.UI.FILL,
    data: [ section ]
});

    fullscreen: true        
});

window.add(table);
window.open();
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
@@ -161,7 +161,9 @@ public class TiTableView extends FrameLayout
 
 		@Override
 		public int getViewTypeCount() {
-			return maxClassname;
+		        // Fix for TIMOB-20038. Seems that there are 3 more
+		        // hidden views that needs to be recreated onLayout
+			return maxClassname + 3;
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -48,7 +48,8 @@
 	public static final int TI_TABLE_VIEW_ID = 101;
 	private static final String TAG = "TiTableView";
 
-	protected int maxClassname = 32;
+	// Default maxClassname is 32. It is set to 35 for bug in TIMOB-20038
+	protected int maxClassname = 35;
```
hieupham007: since you already do +3, this doesn't need to be 35, 32 should be good, otherwise you'd add twice.

hieupham007:
Code reviewed. Looks good.

PR #7686, opened 2016-02-05
[TIMOB-20298] iOS: Classic Titanium app with tabs crashes when closed

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20298


hansemannn:
Approved!

PR #7691, opened 2016-02-09
Update SearchableItemAttributeSet.yml

Gertjancom:
The example has the wrong key for 'uniqueIdentifier'. Took me a while to figure out the name in the example is not the right key :)


cheekiatng:
Thanks @GertjanSmits !

hansemannn:
Hi @GertjanSmits, great catch! Can you try to sign the CLA? We currently have issues with processing the requests, but using the private mode of your browser could resolve the issue for now. If you still experiencing problems, let me know and we take care of that. Thanks again!

GertjanSmits:
I did, I think. But the check error is still there, not sure what I'm doing wrong.

hansemannn:
Ok, no problem! It's a great addition for the SDK, so we just merged it for now. Thanks!

PR #7719, opened 2016-02-12
[TIMOB-20393] Removed launch screen min version check since it makes no sense.

cb1kenobi:


cheekiatng:
CR and FT passed. APPROVED!

PR #7726, opened 2016-02-15
[TIMOB-20400] WWDR certificate can not be found in keychain other than system

FokkeZB:
https://jira.appcelerator.org/browse/TIMOB-20400

### DIFF
--- a/node_modules/ioslib/lib/certs.js
+++ b/node_modules/ioslib/lib/certs.js
@@ -166,7 +166,7 @@ function detect(options, callback) {
 								if (!code) {
 									var tmp = [];
 									parseCerts(out, tmp);
-									results.certs.wwdr = tmp.length && tmp[0].invalid === false;
+									results.certs.wwdr = results.certs.wwdr || (tmp.length && tmp[0].invalid === false);
 								}
 								next();
 							});


node_modules/ioslib/lib/certs.js
```diff
@@ -166,7 +166,7 @@ function detect(options, callback) {
 								if (!code) {
 									var tmp = [];
 									parseCerts(out, tmp);
-									results.certs.wwdr = tmp.length && tmp[0].invalid === false;
+									results.certs.wwdr = results.certs.wwdr || (tmp.length && tmp[0].invalid === false);
```
cb1kenobi: `results.certs.wwdr` will never be truthy because if it was, the parent function would return on line 163.

FokkeZB: Mmm, right... but since the tasks for all keychains run in parallel the same task for another keychain might have set it to true between the test in 163 and when the subprocess calls back 166.

cb1kenobi: I see your point. I'm cool with the fix, but this is the wrong repo. The fix should go in ioslib, not titanium_mobile.

I went ahead and made the fix: https://github.com/appcelerator/ioslib/pull/30.

PR #7729, opened 2016-02-15
Clarify that tintColor sets the ProgressBar's filled part

FokkeZB:
Also fix mixed usage of text and message.

### DIFF
--- a/apidoc/Titanium/UI/ProgressBar.yml
+++ b/apidoc/Titanium/UI/ProgressBar.yml
@@ -56,7 +56,7 @@ methods:
 properties:
 
   - name: color
-    summary: Color of the progress bar text, as a color name or hex triplet.
+    summary: Color of the progress bar message, as a color name or hex triplet.
     description: |
         For information about color values, see the "Colors" section of <Titanium.UI>. 
     type: String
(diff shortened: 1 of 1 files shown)


FokkeZB:
@clinseman

apidoc/Titanium/UI/ProgressBar.yml
```diff
@@ -71,13 +71,13 @@ properties:
   - name: message
     summary: Progress bar text.
     type: String
-
+  - name: tintColor
+    summary: The color shown for the portion of the progress bar that is filled.
```
hansemannn: Missing `type: String` and `platforms: [iphone,ipad]` keys.

FokkeZB: Doesn't it inherit that from the parent? (Ti.UI.View)

hansemannn: When you redefine the property, it will (as far as I know) override the "parent" key.

FokkeZB: Are you sure? This View also overrides the add-method and here it doesn't redefine properties it doesn't change either:
https://github.com/FokkeZB/titanium_mobile/blob/patch-41/apidoc/Titanium/UI/ProgressBar.yml#L40

hansemannn:
You are correct, just checked it by generating the docs. Approved!

PR #7730, opened 2016-02-15
No need to call show() for ProgressBar

FokkeZB:
See discussion in https://jira.appcelerator.org/browse/TIMOB-3754

/cc @jawa9000


hansemannn:
Approved!

PR #7732, opened 2016-02-15
Remove skitch that gave 404

FokkeZB:
/cc @jawa9000


hansemannn:
Can we provide a new image source here instead of just removing it?

FokkeZB:
@jawa9000 I think this was an image displaying what a popover looks like, as we have with other APIs as well. Not sure what your policy is on where to host these images but I think we'd better use S3 instead of skitch.

FokkeZB:
Also see https://jira.appcelerator.org/browse/TIDOC-2458

jawa9000:
For documentation that goes public, images must be attached to the wiki page and never sourced from any external locations. The publication scripts either don't have permission to fetch images from remote locations or doesn't know how to handle externally linked images.

PR #7736, opened 2016-02-16
[TIMOB-5812] (5_2_X) Run Ti.UI.iOS.MenuPopup UI-methods on main thread

hansemannn:
JIRA: https://jira.appcelerator.org/browse/TIMOB-5812


cheekiatng:
CR and FT passed. APPROVED.

PR #7748, opened 2016-02-17
Use calibrated white instead of rgb(1,1,1,1)

FokkeZB:
@cb1kenobi this seems to make more sense, you agree?


FokkeZB:
Never mind, I see how you use it to inject the color set in tiapp.xml

PR #7756, opened 2016-02-18
[TIMOB-14953] iOS: Ti.Calendar getEventsBetweenDates is not working

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-14953?filter=-1)
Fixed spelling and change the code so the method takes in Date objects just like its suppose to acording to the docs,  instead of just strings.

### DIFF
--- a/apidoc/Titanium/Calendar/CalendarProxy.yml
+++ b/apidoc/Titanium/Calendar/CalendarProxy.yml
@@ -34,11 +34,11 @@ methods:
     parameters:
       - name: date1
         summary: Start date.
-        type: Date
+        type: [Date,String]
         
       - name: date2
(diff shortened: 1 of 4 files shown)


hansemannn:
CR looks fine! The [docs](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar.Calendar-method-getEventsBetweenDates) state that only a JS date is accepted, so we either need to update the docs to support both or only accept the native JS date and deprecate the String args.

AngelkPetkov:
I'll Update the docs as every one has been using strings till now. That way we except both no need for deprecating.

hansemannn:
Approved!

PR #7760, opened 2016-02-19
[TIMOB-20087] iOS: Breaks ability to bind one collection…

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-20087)  Revert ifdef to ifndef


hansemannn:
Approved!

PR #7769, opened 2016-02-23
[TIMOB-20405] (5_2_X) iOS: Calculate Ti.UI.Tab iconInsets internally

hansemannn:
https://jira.appcelerator.org/browse/TIMOB-20405

### DIFF
--- a/apidoc/Titanium/UI/Tab.yml
+++ b/apidoc/Titanium/UI/Tab.yml
@@ -224,7 +224,8 @@ properties:
     summary: The icon inset or outset for each edge.
     description: | 
         Use this property for example to center an icon without providing a title. To do that, set the
-        insets to `{top:6, bottom:-6, right:0, left:0}`.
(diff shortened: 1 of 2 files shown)


cheekiatng:
APPROVED!

apidoc/Titanium/UI/Tab.yml
```diff
@@ -224,7 +224,8 @@ properties:
     summary: The icon inset or outset for each edge.
     description: | 
         Use this property for example to center an icon without providing a title. To do that, set the
-        insets to `{top:6, bottom:-6, right:0, left:0}`.
+        insets to `{top:6, left: 0}`. The `right` and `bottom` are calculated internally to prevent the
```
FokkeZB: I'd leave out `left:0` since that's the default (and you would mostly use `top` anyway)

hansemannn: Well it's an example. If you don't set left, it will be 0 anyway. We have more critical issues for now I guess :-)

PR #7773, opened 2016-02-24
[TIDOC-2454] API reference for Activity & Shared Elements transitions out of place and incorrect

sriks:
JIRA https://jira.appcelerator.org/browse/TIDOC-2454

### DIFF
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -1023,6 +1023,139 @@ properties:
     constants: Titanium.UI.Android.PIXEL_FORMAT_*
     since: "1.8.0"
 
+  - name: activityExitTransition
+    summary: The type of transition used when activity is exiting. 
+    description: |
+        Activity A's exit transition determines how views in A are animated when A starts B.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false or 
+        there is no shared element between the participating activities.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.  
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_EXPLODE, Titanium.UI.Android.TRANSITION_FADE_*, 
+                 Titanium.UI.Android.TRANSITION_SLIDE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: If not specified uses platform theme transition. 
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activityEnterTransition
+    summary: The type of transition used when activity is entering. 
+    description: |
+        Activity B's enter transition determines how views in B are animated when A starts B.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false or 
+        there is no shared element between the participating activities.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_EXPLODE, Titanium.UI.Android.TRANSITION_FADE_*, 
+                 Titanium.UI.Android.TRANSITION_SLIDE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: If not specified uses platform theme transition. 
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activityReturnTransition
+    summary: The type of transition used when returning from a previously started activity. 
+    description: |
+        Activity B's return transition determines how views in B are animated when B returns to A.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false or 
+        there is no shared element between the participating activities.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_EXPLODE, Titanium.UI.Android.TRANSITION_FADE_*, 
+                 Titanium.UI.Android.TRANSITION_SLIDE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: If not specified uses `activityEnterTransition`. 
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activityReenterTransition
+    summary: The type of transition used when reentering to a previously started activity. 
+    description: |
+        Activity A's reenter transition determines how views in A are animated when B returns to A.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false or 
+        there is no shared element between the participating activities.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_EXPLODE, Titanium.UI.Android.TRANSITION_FADE_*, 
+                 Titanium.UI.Android.TRANSITION_SLIDE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: If not specified uses `activityExitTransition`. 
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementExitTransition
+    summary: The type of exit transition used when animating shared elements between two activities.
+    description: |
+        Activity A's shared element exit transition animates shared elements before they transition from A to B
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: | 
+        Defaults to android platform's [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml) transition.
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementEnterTransition
+    summary: The type of enter transition used when animating shared elements between two activities.
+    description: |
+        Activity B's shared element enter transition determines how shared elements animate from A to B.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: |
+        Defaults to android platform's [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml) transition.
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementReturnTransition
+    summary: The type of return transition used when animating shared elements between two activities.
+    description: |
+        Activity B's shared element return transition determines how shared elements animate from B to A. 
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: |
+        Defaults to android platform's [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml) transition.
+    since: "5.2.0"
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementReenterTransition
+    summary: The type of reenter transition used when animating shared elements between two activities.
+    description: |
+        Activity A's shared element reenter transition animates shared elements after they have transitioned from B to A. 
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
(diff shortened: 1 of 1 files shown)


apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementExitTransition
+    summary: The type of exit transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
```
FokkeZB: Can we clarify the activitySharedElement*Transition descriptions a bit?

- Exit: Activity A's shared element exit transition animates shared elements before they transition from A to B.
- Enter: Activity B's shared element entry transition determines how shared elements animate from A to B.
- Return: Activity B's shared element return transition determines how shared elements animate from B to A. Defaults to the enter transition.
- Reenter: Activity A's shared element reenter transition animates shared elements after they have transitioned from B to A. Defaults to the exit transition.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activityReenterTransition
+    summary: The type of transition used when reentering to a previously started activity. 
+    description: |
+        Activity A's reenter transition determines how views in A are animated when B returns to A.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
```
FokkeZB: I would add: Defaults to the exit transition.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activityReturnTransition
+    summary: The type of transition used when returning from a previously started activity. 
+    description: |
+        Activity B's return transition determines how views in B are animated when B returns to A.
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
```
FokkeZB: I would add: Defaults to the enter transition.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementExitTransition
+    summary: The type of exit transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
```
FokkeZB: I actually can't find it documented anywhere if shared elements cannot also use the explode, slide etc. Unfortunately I don't notice differences between the change transitions either so hard to tell. Any idea?

FokkeZB:
@sriks I left some comments on the PR code

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+
+  - name: activitySharedElementReenterTransition
+    summary: The type of reenter transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: Titanium.UI.Android.TRANSITION_CHANGE_BOUNDS
```
FokkeZB: I think by default there's no reenter transition and since it happens after the return transition and the shared element is already on its place you wouldn't see it anyway.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+
+  - name: activitySharedElementExitTransition
+    summary: The type of exit transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: Titanium.UI.Android.TRANSITION_CHANGE_BOUNDS
```
FokkeZB: I think by default there's no exit transition and since it happens before the enter transition and the shared element is still on its place you wouldn't see it anyway.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementReenterTransition
+    summary: The type of reenter transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
```
FokkeZB: activitySharedElementReenterTransition can take any transition, not just CHANGE_

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementReturnTransition
+    summary: The type of return transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
```
FokkeZB: activitySharedElementReenterTransition can take any transition, not just CHANGE_

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,127 @@ properties:
...
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementEnterTransition
+    summary: The type of enter transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
```
FokkeZB: activitySharedElementReenterTransition can take any transition, not just CHANGE_

apidoc/Titanium/UI/Window.yml
sriks: The android docs don't specify any order. This is because the transition is a visual clue happening over the scene. Not a state change.

apidoc/Titanium/UI/Window.yml
sriks: As per android docs, it takes CHANGE_ transitions only. But programatically it accepts all classes derived from Transition. So what we mentioned in our docs should hold good. 

> Android 5.0 (API level 21) also supports these shared elements transitions:
> 
> changeBounds - Animates the changes in layout bounds of target views.
> changeClipBounds - Animates the changes in clip bounds of target views.
> changeTransform - Animates the changes in scale and rotation of target views.
> changeImageTransform - Animates changes in size and scale of target images.
>

apidoc/Titanium/UI/Window.yml
sriks: The sharedElementReenterTransition will be picked after the element is moved to its start state. Since we are providing just the bridge to native API, it is difficult to count on what the developers may or may not use. I think we have this property and let the developer decide how to use it. 

Also since this API is designed with auto defaults in place, the developer don't have to specify the reenter/exit transition explicitly unless they want to completely customize it.

apidoc/Titanium/UI/Window.yml
sriks: Updated.

apidoc/Titanium/UI/Window.yml
sriks: Updated

apidoc/Titanium/UI/Window.yml
FokkeZB: It does at:
https://developer.android.com/reference/android/view/Window.html#attr_android:windowSharedElementExitTransition

* exit: "prior to transferring to the called Activity"
* reenter: "used for shared elements transferred back to a calling Activity" (past tense)

It's also very clearly explained at:
http://www.androiddesignpatterns.com/2015/01/activity-fragment-shared-element-transitions-in-depth-part3a.html

And this with videos shows it as well:
https://halfthought.wordpress.com/2014/12/08/what-are-all-these-dang-transitions/

But as I say in the blog with the transition animations we have (no shadow stuff like the above video uses) exit and reeenter are useless. Still, we should describe them :D

apidoc/Titanium/UI/Window.yml
FokkeZB: :+1:

apidoc/Titanium/UI/Window.yml
FokkeZB: [Agreed](https://github.com/appcelerator/titanium_mobile/pull/7773#discussion_r54843126)

Could there be a way to allow developers to define transitions in XML (like we do with custom themes) and use them here some how (reference via `Ti.Android.R` for example)? Not for now, but as a later improvement?

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,131 @@ properties:
...
+    availability: creation
+    platforms: [android]
+    osver: {android: {min: 5.0}}
+
+  - name: activitySharedElementEnterTransition
+    summary: The type of enter transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
```
FokkeZB: Defaults to [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml), which is all of them combined.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,131 @@ properties:
...
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: Titanium.UI.Android.TRANSITION_CHANGE_BOUNDS
```
FokkeZB: It actually defaults to [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml), which is all of them combined.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1023,6 +1023,131 @@ properties:
...
+
+  - name: activitySharedElementReturnTransition
+    summary: The type of return transition used when animating shared elements between two activities.
+    description: |
+        Applicable for Android 5.0 and above. This value will be ignored if `animated` is set to false.
+        See "Material design activity transitions in Android" in the main description of Titanium.UI.Window
+        for more information.
+    type: Number
+    constants: [ Titanium.UI.Android.TRANSITION_CHANGE_*, Titanium.UI.Android.TRANSITION_NONE ]
+    default: Titanium.UI.Android.TRANSITION_CHANGE_BOUNDS
```
FokkeZB: It actually defaults to [move](https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/res/res/transition/move.xml), which is all of them combined.

apidoc/Titanium/UI/Window.yml
sriks: Noted. But we don't get any sort of callbacks on this. But we should describe them :)

apidoc/Titanium/UI/Window.yml
sriks: Yes. We can investigate this for future enhancement.

apidoc/Titanium/UI/Window.yml
sriks: Documented anywhere? Else we can mention as platform default.

FokkeZB: It's in the source code for the material theme:
https://github.com/android/platform_frameworks_base/blob/d59921149bb5948ffbcb9a9e832e9ac1538e05a0/core/res/res/values/themes_material.xml#L173-L174

apidoc/Titanium/UI/Window.yml
FokkeZB: https://jira.appcelerator.org/browse/TIMOB-20514

apidoc/Titanium/UI/Window.yml
sriks: Updated docs.

FokkeZB:
@sriks I see the descriptions for the shared element transitions are still identical, could we improve those?

- Exit: Activity A's shared element exit transition animates shared elements before they transition from A to B.
- Enter: Activity B's shared element entry transition determines how shared elements animate from A to B.
- Return: Activity B's shared element return transition determines how shared elements animate from B to A. Defaults to the enter transition.
- Reenter: Activity A's shared element reenter transition animates shared elements after they have transitioned from B to A. Defaults to the exit transition.

sriks:
@FokkeZB Added more description to shared element transitions.

FokkeZB:
**APPROVED**

Let's merge (I can't)

sriks:
Merging.

PR #7775, opened 2016-02-24
Do not log alerts

FokkeZB:
I don't see why we have to log UI alerts to the console.


hansemannn:
Makes sense, good catch yooo!

PR #7780, opened 2016-02-26
[TIMOB-19919] Android: Reverting changes for onBackPressed 

ashcoding:
https://jira.appcelerator.org/browse/TIMOB-19919


sriks:
Should we remove onBack from docs and property accessors here https://github.com/manumaticx/titanium_mobile/blob/fddcb6fbf90425c4d392157ecedf6f4e90f49e77/android/titanium/src/java/org/appcelerator/titanium/proxy/TiWindowProxy.java#L36 ?

ashcoding:
@sriks Addressed your comment by removing that property in this latest commit.

PR #7781, opened 2016-02-27
[TIMOB-20464] iOS: Picker cant be set as a percentage

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-20464?filter=-1)

### DIFF
--- a/iphone/Classes/TiUIPicker.m
+++ b/iphone/Classes/TiUIPicker.m
@@ -26,48 +26,40 @@
 
 USE_PROXY_FOR_VERIFY_AUTORESIZING
 
--(CGFloat)verifyHeight:(CGFloat)suggestedHeight
-{
-	// pickers have a forced height so we use it's height
-	// instead of letting the user set it
-	return picker.frame.size.height;
-}
-
--(CGFloat)verifyWidth:(CGFloat)suggestedWidth
-{
-    if (suggestedWidth <= 0 && picker != nil) {
-        return picker.frame.size.width;
-    }
-    else {
-        return suggestedWidth;
-    }
-}
-
 -(UIControl*)picker 
 {
 	if (picker==nil)
 	{
-        float width = 320;
-        float height = 216;
-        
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIPicker.m
```diff
@@ -26,55 +26,55 @@ -(void)dealloc
...
+        float width  =  self.bounds.size.height;
+        float height = self.bounds.size.width;
+        if (![TiUtils isIOS9OrGreater])
+        {
+            width  = 320;
+            height = 216;
+            //TIMOB-20028
+            [[self proxy ]setValue:NUMDOUBLE(width) forKey:@"width"];
+            [[self proxy ]setValue:NUMDOUBLE(height) forKey:@"height"];
+        }
```
pec1985: Do not call `setValue:ForKey:` since it adds an unnecessary js property

AngelkPetkov: Prior to iOS 9 it doesnt matter what we set the size too it will never get scaled. The reason im setting the value for key is since this was added as a change to TIMOB-20028. If I dont it reverts  the change to a  ticket weve already closed as fixed.
https://jira.appcelerator.org/browse/TIMOB-20028?filter=-1&jql=assignee%20%3D%20currentUser()

hansemannn: As discussed in 1:1, we should migrate the Picker to use the layout engine instead of handling all dimensions manually. For example, what do we do if the user sets `Ti.UI.FILL` on the picker?

pec1985:
Approved.

inakiabt:
Could be this issue related to [TIMOB-20518](https://jira.appcelerator.org/browse/TIMOB-20518)?

AngelkPetkov:
Yes it does , if you take a look at the JiRa ticket ive related them both , unfortanetly it does not fully fix the issue. It will be fixed soon!

PR #7789, opened 2016-03-01
[TIMOB-20431] (3_5_X) Remove NonValidatingTrustManager.

hieupham007:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-20431


adoblas:
+1

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -549,16 +549,7 @@ public int getReadyState()
 
 	public boolean validatesSecureCertificate()
 	{
-		if (proxy.hasProperty("validatesSecureCertificate")) {
```
FokkeZB: We need a deprecation warning here so we can remove it in 6.0

hieupham007:
Merging

PR #7791, opened 2016-03-01
[TIMOB-20489] (5_2_X) Added support for a 'Resource' dir for iOS modules that…

cb1kenobi:
- [TIMOB-20489] Added support for a 'Resource' dir for iOS modules that copy files into output app directory.
- Added deprecation warning for the old iOS module python-based build script.
- Removed iOS module build's config() method since it's never invoked.
- Removed archiver monkey patch for older version that is no longer needed.
- Added support for build-only iOS module builds.
- Fixed bug where the build didn't synchronize the completion of both xcodebuild calls.
- Added a log output token to xcodebuild to identify what output belonged to which xcodebuild call.
- Cleaned up module documentation generation for iOS only.
- General iOS module build cleanup.

### DIFF
--- a/cli/commands/build.js
+++ b/cli/commands/build.js
@@ -36,7 +36,7 @@ exports.title = __('Build');
 exports.desc = __('builds a project');
 exports.extendedDesc = __('Builds an existing app or module project.');
 
-exports.config = function (logger, config, cli) {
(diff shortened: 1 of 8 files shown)


feons:
Looks good. **Approved.**

PR #7795, opened 2016-03-02
[TIMOB-20453] Show launchImage in RootViewController unless we're using storyboard

FokkeZB:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-20453

**Additional description (optional)**:


cheekiatng:
APPROVED

PR #7800, opened 2016-03-02
Improve parity report size/performance

sgtcoolguy:
There is no JIRA ticket for this change.

This stemmed from my frustrations with the current parity report - mainly that it's long to load (from the web) and slow to toggle showing members/pseudo-types.

As a first cut, this PR cut the size of the parity report HTML file on my Mac from 14,672,146 bytes to 5,798,335 bytes. Initial page load time went from 703ms to 424ms. I attempted to time showing members (without pseudo-types), and there are so many events it's hard to know the exact number, but it's over 15s from click to redraw. After this change it's now 1.8645s.

Future: Ideally I'd love to further improve it to be able to hide platform-specific APIs and change the API coverage percentage reported as a result (since there are a large number of Android/iOS specific APIs that make the API coverage look bad for something like Mobileweb or Windows).

### DIFF
--- a/apidoc/templates/parity.ejs
+++ b/apidoc/templates/parity.ejs
@@ -1,115 +1,215 @@
-<!-- Copyright (c) 2015 Appcelerator, Inc. All Rights Reserved. -->
+<!-- Copyright (c) 2015-2016 Appcelerator, Inc. All Rights Reserved. -->
 <!-- Licensed under the terms of the Apache Public License.     -->
-<% var green = '#3CC850', red = '#C83232' %>
-<% function contains(a,b){for(i in a)if(a[i]==b)return true;return false;} %>
-<% function rgbHex(r,g,b){return "#"+((1<<24)+(r<<16)+(g<<8)+b).toString(16).slice(1).split('.')[0];} %>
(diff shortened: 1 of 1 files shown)


cheekiatng:
I think @jawa9000 should take a look at this, and if it's ok you can go ahead to merge it.

apidoc/templates/parity.ejs
```diff
@@ -1,115 +1,215 @@
...
+	.p { display: none; } /* pseudo */
+	/* .n == normal/non-pseudo */
+	.child { display: none; }
+	.e { color: #FFF; text-align: center; font-weight: bold; } /* entry */
+	th.stats_description { font-weight: bold; }
+	.s { background-color: <%= green %>; } /* supported */
+	.u { background-color: <%= red %>; } /* unsupported */
+
+	/* Create classes for the gradients of support */
+<% for (var i = 0; i < platformCount; i++) { -%>
```
garymathews:
```Javascript
for (var i = 0; i <= platformCount; i++) {
```

garymathews:
@sgtcoolguy Nice!! This is a lot faster than before!

apidoc/templates/parity.ejs
sgtcoolguy: Good catch, fixed.

jawa9000:
Looks good! Merging.

PR #7806, opened 2016-03-07
[TIMOB-20519] If --platform is not explicitly set, then automatically…

cb1kenobi:
https://jira.appcelerator.org/browse/TIMOB-20519

[TIMOB-20519] If --platform is not explicitly set, then automatically set it to the platform in the module's manifest. Moved module manifest validation to validate(). Resolved module's platform prior to validation.


feons:
FR passed. *Merging...*

PR #7812, opened 2016-03-07
[TIMOB-20526] iOS Module: metadata.json file is written outside ios module folder

feons:
https://jira.appcelerator.org/browse/TIMOB-20526


cb1kenobi:
**APPROVED**

PR #7825, opened 2016-03-10
Revert "[TIMOB-20464] iOS: Picker cant be set as a percentage"

pec1985:
Reverts appcelerator/titanium_mobile#7781


iphone/Classes/TiViewProxy.m
```diff
@@ -2330,15 +2330,7 @@ -(void)relayout
         }
 
         if (layoutChanged && [self _hasListeners:@"postlayout" checkParent:NO]) {
-
```
AngelkPetkov: We should probably leave this in, as it was for the issue when calling the postLayout we couldn't get the properties on on main thread

AngelkPetkov:
The issue that the customer has reported was prior to the merger of the picker PR. If we merge this it will be exactly the same effect the user reported. Just tested , you just get a black screen

PR #7828, opened 2016-03-10
[TIMOB-16355] iOS: separatorInsets should not be used for header/sect…

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-16355)

### DIFF
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -849,6 +849,9 @@ properties:
 
   - name: separatorInsets
     summary: The insets for list view separators (applies to all cells). This property is applicable on iOS 7 and greater.
+    deprecated:
+        since: "5.2.0"
+        notes: Use tableSeparatorInsets instead
     description: |
         In iOS 7 and later, cell separators do not extend all the way to the edge of the list view. 
         This property sets the default inset for all cells in the table. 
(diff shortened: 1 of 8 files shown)


apidoc/Titanium/UI/ListView.yml
```diff
@@ -849,6 +849,9 @@ properties:
 
   - name: separatorInsets
     summary: The insets for list view separators (applies to all cells). This property is applicable on iOS 7 and greater.
+    deprecated:
+        since: "5.2.0"
```
hansemannn: The [Ti.UI.TableView version of this PR](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/TableView.yml#L1553) says deprecated since 5.4.0, this one since 5.2.0. They should be the same

apidoc/Titanium/UI/ListView.yml
```diff
@@ -875,6 +878,46 @@ properties:
     platforms: [iphone, ipad]
     availability: creation
 
+  - name: tableSeparatorInsets
+    summary: The insets for the table view header and footer. This property is applicable on iOS 7 and greater.
+    description: |
+        In iOS 7 and later, cell separators do not extend all the way to the edge of the list view.Set this to a
```
hansemannn: Missing space after `view.Set`. Also check the TableView version

iphone/Classes/TiUIListView.m
```diff
@@ -544,6 +546,13 @@ -(NSInteger)sectionForSearchSection:(NSInteger)section
 -(void)setSeparatorInsets_:(id)arg
 {
     [self tableView];
+    DEPRECATED_REPLACED(@"UI.ListView.separatorInsets", @"5.2.0", @"UI.ListView.tableSeparatorInsets");
+    [self setTableSeparatorInsets_:arg];
+}
+-(void)setTableSeparatorInsets_:(id)arg
+{
+    [self tableView];
```
hansemannn: We should not call it here, since it just creates the instance of not available already. Change `_tableView` to `[self tableView]` below and in the other cases you used it.

apidoc/Titanium/UI/ListView.yml
AngelkPetkov: There should both be 5.2.0. Ill change tableView as well. As the original PR was merged in 5.2.0

PR #7829, opened 2016-03-10
[TIMOB-16355](5_2_X) iOS: separatorInsets should not be used for header/sect…

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-16355)


hansemannn:
CR + FT passed, PR approved!

PR #7835, opened 2016-03-11
[TIMOB-20534](5_2_X) Replacing with updated titanium-verify.jar and libtiverify.so

ashcoding:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20534

[TIMOB-20534] Android: Services enabled app crashes with error "Didn't find class ti.modules.titanium.network.NonValidatingSSLSocketFactory" when built with 3.5.2 & 4.0.1 SDK

Notes:- Libs were built against Titanium Mobile SDK 5.2.0.GA.

See ticket for testing steps.


hieupham007:
We don't need this now, will reopen if we do in the future.

PR #7839, opened 2016-03-13
Docs: Change "since" version form 6.0.0 to 5.4.0

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20559

Fixing the docs issue from the above ticket along with other occurrences in the docs.


FokkeZB:
I didn't check if all these are actually 5.4 instead of 6.0 (but since master is 5.4 that's probably the case right?) looks good.

**APPROVED**

PR #7844, opened 2016-03-14
Seated various misplaced methods

jawa9000:
https://jira.appcelerator.org/browse/TIDOC-2489

The following methods were not properly seated in the YAML heirarchy:Point, Gradient, GradientColorRef, Dimension, and AnimationOption. With these methods seated under the properties tree, the local build properly shows these methods as well as the entire document.


hansemannn:
Hey Brian, we have some merge conflicts with latest master here, can you check?

apidoc/Titanium/Media/Media.yml
```diff
@@ -1526,7 +1526,9 @@ properties:
     default: false
   - name: mediaTypes 
     summary: | 
-        Array of media type constants to allow.
+        Array of media type constants to allow. If you want to allow live photos with
+        Titanium.Media.MEDIA_TYPE_LIVEPHOTO on iOS 9.1 and later, you also need to specify
+        at least Titanium.Media.MEDIA_TYPE_PHOTO as a fallback.
```
hansemannn: Incorrect, because the camera doesn't support live photos (yet). That's why I didn't add it there :-)

apidoc/Titanium/UI/iOS/LivePhotoView.yml
```diff
@@ -60,6 +60,14 @@ events:
         summary: Returns the `playbackStyle` that was provided to start the playback.
         type: Number
 
+properties:
+  - name: livePhoto
```
hansemannn: Not sure why this doesn't throw merge conflicts, since the master already contains [this property](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/iOS/LivePhotoView.yml#L22). Can you check? Thanks man!

apidoc/Titanium/UI/iPhone/AlertDialogStyle.yml
```diff
@@ -9,7 +9,7 @@ platforms: [iphone, ipad]
 createable: false
 deprecated:
     since: "6.0.0"
-    notes: Use [Titanium.UI.iOS.AlertDialogStyle](Titanium.UI.iOS.AlertDialogStyle) instead.
+    notes: Use Titanium.UI.iOS.AlertDialogStyle instead.
```
hansemannn: Why not use `<Ti.UI.iOS.AlertDialogStyle>` which should auto-link it?

apidoc/Titanium/UI/iPhone/iPhone.yml
```diff
@@ -181,7 +181,7 @@ properties:
     type: Number
     deprecated:
         since: "6.0.0"
-        notes: Use [Ti.UI.iOS.appBadge](Ti.UI.iOS.appBadge) instead.
+        notes: Use Titanium.UI.iOS.appBadge instead.
```
hansemannn: Same as above

apidoc/Titanium/UI/iPhone/AlertDialogStyle.yml
jawa9000: We are having problems crossing linking if people use the short-hand of Titanium (Ti). Right now, our document system doesn't treat Titanium.whatever and Ti.whatever the same. Some links are broken, in particular those using the incorrect module prefix. Ultimately, we need to unify which one we are going to use.

hansemannn: Makes sense! But still, wouldn't the Autolinking make sense (using <>)?

apidoc/Titanium/UI/iPhone/iPhone.yml
```diff
@@ -216,42 +216,26 @@ properties:
...
-    
-  - name: animationStyle
-    summary: Style of the animation.
-    description: |
-        One of the status bar animation style constants 
-        <Titanium.UI.iPhone.StatusBar.ANIMATION_STYLE_NONE>, 
-        <Titanium.UI.iPhone.StatusBar.ANIMATION_STYLE_SLIDE>, 
-        or <Titanium.UI.iPhone.StatusBar.ANIMATION_STYLE_FADE>.
-    type: Number
+  - name: hideStatusBarParams
```
hansemannn: Why do we do this? The `----` is meant to be used  for dictionaries that need to be passed to other API's. They are no properties / methods.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -851,7 +851,11 @@ properties:
     summary: The insets for list view separators (applies to all cells). This property is applicable on iOS 7 and greater.
     deprecated:
         since: "5.2.0"
+<<<<<<< HEAD
+        notes: Use tableSeparatorInsets instead
+=======
```
hansemannn: Merge problems

hansemannn:
Hey @jawa9000, there are lots of merge issues resulting on diff-ing the master. Can you submit a new one that includes the changes you focussed on initially?

PR #7847, opened 2016-03-15
[TIMOB-20396] Fix main-thread issues in callbacks

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20396


jhaynie:
curious why we aren't using GCD and are using threads instead here? wouldn't GCD be cleaner code and more performant?

pec1985:
The `[NSThread detachNewThreadSelector...` is legacy code that needs to be removed once Kroll Thread is removed. If we're running on the main thread, we do not need to call this method on a new thread or async, since the callback and event system takes care of that in a different place.
We're placing an `ifdef` and introducing new code _if_ on main thread.

hansemannn:
Yes, everything in `#ifdef TI_USE_KROLL_THREAD` is legacy for people who don't want to use the main-thread in Ti.SDK >= 5.4.0. The delay in `[self dispatchCallback:..]` is necessary to fire the event after the modal transition finishes.

pec1985:
@hansemannn  there are some conflicts here. I will merge it once those conflicts are resolved

hansemannn:
Updated yesterday

pec1985:
PR reviewed. **APPROVED!**

PR #7850, opened 2016-03-15
[TIMOB-20570] (5_2_X) Added support for the Resources dir and --build-only to…

cb1kenobi:
https://jira.appcelerator.org/browse/TIMOB-20570

[TIMOB-20570] Added support for the Resources dir and --build-only to the Android module build. Handful of other updates including how docs are generated.


feons:
Looks good! **Approved.**

PR #7852, opened 2016-03-16
[TIMOB-20548] (5_3_X) Fixed bug where Android emulator was not be detected as…

cb1kenobi:
https://jira.appcelerator.org/browse/TIMOB-20548

[TIMOB-20548] Fixed bug where Android emulator was not be detected as an emulator and was being treated as a device.


feons:
Looks great! **Approved.**

PR #7854, opened 2016-03-16
[TIMOB-20397] iOS: Proxies become nil if not created with pageContext (Main-Thead)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20397


pec1985:
**APPROVED!**

PR #7671, opened 2016-02-03
[TIMOB-20251]: Enabling request for external storage.

hieupham007:
Testing steps in JIRA: https://jira.appcelerator.org/browse/TIMOB-20251

PR #7858, opened 2016-03-16
[TIMOB-20440]: Remove unnecessary check on write method.

hieupham007:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20440
Need to test https://github.com/appcelerator/titanium_mobile/pull/7671 as well, since the change was originally introduced there.

### DIFF
--- a/ti_mocha_tests/ti.filesystem.test.js
+++ b/ti_mocha_tests/ti.filesystem.test.js
@@ -432,9 +432,9 @@ describe("filesystem", function() {
 		finish();
 	});
 	//TIMOB-20440
-	it(“applicationDataDirectory”, function(finish) {
+	it("applicationDataDirectory", function(finish) {
 		var file = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'filter.txt');
-		var text = “testing a file”;
+		var text = "testing a file";
 		file.write(text);
 		should(file.exists()).be.true;
 		var blob = file.read();
--- a/ti_mocha_tests/ti.filesystem.test.js
+++ b/ti_mocha_tests/ti.filesystem.test.js
@@ -431,6 +431,23 @@ describe("filesystem", function() {
 		}
 		finish();
 	});
+	//TIMOB-20440
+	it(“applicationDataDirectory”, function(finish) {
+		var file = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'filter.txt');
+		var text = “testing a file”;
+		file.write(text);
+		should(file.exists()).be.true;
+		var blob = file.read();
+		should(blob).not.be.null;
+		var readText = blob.text;
+		should(readText).not.be.null;
+		should(readText).not.be.type("undefined");
+		should(readText).be.a.String;
+		should(readText.length).eql(text.length);
+		should(readText).eql(text);
+		file.deleteFile();
+		finish();
+	}
 	//TIMOB-4469
 	it("existsMethod", function(finish) {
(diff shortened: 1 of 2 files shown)


ashcoding:
Could you also include tests in this folder as well? https://github.com/appcelerator/titanium_mobile/tree/master/ti_mocha_tests

Also with regards to this using the test case:-
```javascript
    backgroundColor : '#ffffff'
});
    top : 10,
    width : 200,
    height : 50,
    title : "write file"
});
win.add(button);
    var f = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'fileter.txt');
    f.write('writing to the file would be enough to create it');
    alert(f.read());
});
 
win.open();
```

This is fine.

If I changed the directory to 
```javascript
var f = Ti.Filesystem.getFile(Ti.Filesystem.externalStorageDirectory, 'fileter.txt');
```
I will get the error (On Android 6.0 and above with no permission granted)
```
[ERROR] :  TiFileProxy: (KrollRuntimeThread) [8580,8580] IOException encountered
[ERROR] :  TiFileProxy: java.io.FileNotFoundException: /storage/emulated/0/sg.com.appcelerator/fileter.txt: open failed: EACCES (Permission denied)
```
Is this okay? Should the developer know that externalStorageDirectory requires permission in order to do this or should there be a way to handle this too?

hieupham007:
Well that's expected. If they're opening a file that requires a permission, they would need permission. However, depending the location of that file, they may not need permissions to do it, hence this change. Since this is a Kroll method, we can't stop devs from calling it, so we would let them handle the logic.

ashcoding:
Sounds good to me. :+1:
Code reviewed and functionally tested. Just needs tests in the ti_mocha_tests folder.

ti_mocha_tests/ti.filesystem.test.js
```diff
@@ -431,6 +431,23 @@ describe("filesystem", function() {
 		}
 		finish();
 	});
+	//TIMOB-20440
+	it(“applicationDataDirectory”, function(finish) {
```
ashcoding: [ERROR] Unexpected character '“' [line 435, column 4]

I think you used the wrong double inverted commas here. Wrong character.

ti_mocha_tests/ti.filesystem.test.js
```diff
@@ -431,6 +431,23 @@ describe("filesystem", function() {
...
+		var file = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'filter.txt');
+		var text = “testing a file”;
```
ashcoding: Same here. Wrong character.

ashcoding:
`npm test` failing due to wrong character used.

ti_mocha_tests/ti.filesystem.test.js
```diff
@@ -431,6 +431,23 @@ describe("filesystem", function() {
...
+		should(blob).not.be.null;
+		var readText = blob.text;
+		should(readText).not.be.null;
+		should(readText).not.be.type("undefined");
+		should(readText).be.a.String;
+		should(readText.length).eql(text.length);
+		should(readText).eql(text);
+		file.deleteFile();
+		finish();
+	}
```
ashcoding: Missing `);`

ashcoding:
Code reviewed and functionally tested to be working.
**Accepted**
Test code has some typos. Will be doing a PR for test case after this.

ashcoding:
Fix for Test case is here: https://github.com/appcelerator/titanium_mobile/pull/7871

PR #7865, opened 2016-03-18
Update for outdated example

falkolab:
`resume` event deprecated since 3.4.0


hansemannn:
FT passed, PR approved!

PR #7868, opened 2016-03-18
[TIMOB-20595] Module Build: Update pre and post compile hooks event name 5_3_X

feons:
https://jira.appcelerator.org/browse/TIMOB-20595


cb1kenobi:
**APPROVED**

PR #7876, opened 2016-03-23
Fix for DM-12

jawa9000:
https://jira.appcelerator.org/browse/DM-12

Several properties were sitting outside the property tree.

Fixed several misplaced properties, removed unnecessary line returns,
and fix a few indentation issues. All this was causing this API doc to
fail to render during the doc build process.

### DIFF
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -567,7 +567,6 @@ events:
         type: Boolean
 
 properties:
-
   - name: allowsSelection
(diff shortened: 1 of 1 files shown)


hansemannn:
@jawa9000 Just for interest, do the generated docs work with these changes? I'm concerned that the dictionary-params are treated like properties which there aren't. The docs may failed rendering because of a missing blank line before the params separator (`----------`).

jawa9000:
From my Yamllint and localized tests, param separators (missing or present) doesn't cause any problems.

PR #7881, opened 2016-03-24
[TIMOB-6305] iOS: Add Stepper UI element

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-6305) 
Adding Support for UIStepper


apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
+---
+name: Titanium.UI.iOS.Stepper
+summary: |
+    A widget used to increment and decrement a value
+description: |
+    <table id="IOS Stepper">
```
hansemannn: An `id` is not allowed to have spaces. And why is a HTML id even required here?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+      <tr>
+        <td><img src="images/stepper/stepper.png" height="50" /></td>
```
hansemannn: You calling the directory "stepper", but the below folder is called "Stepper"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        <td><img src="images/stepper/stepper_customImages" height="51" /></td>
```
hansemannn: Missing ".png" suffix. Did you test the generated docs before?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+      </tr>
+      <tr><th colspan="2">iOS</th></tr>
+    </table>
+
+    A Stepper displays its selected state while it is being pressed.
+    You can specify background images for each state on iOS.
+    Use the <Titanium.UI.createstepper> method or a **`<stepper>`** Alloy element to create a stepper
```
hansemannn: It's `createStepper`, not `createstepper` (It will cause JS runtime errors). Also `<Stepper>` instead of `<stepper>`. And did you implement the Alloy-part of this PR? I can't see any Alloy-related code+docs.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        <td><img src="images/stepper/stepper.png" height="50" /></td>
+        <td><img src="images/stepper/stepper_customImages" height="51" /></td>
+      </tr>
+      <tr><th colspan="2">iOS</th></tr>
+    </table>
+
+    A Stepper displays its selected state while it is being pressed.
+    You can specify background images for each state on iOS.
+    Use the <Titanium.UI.createstepper> method or a **`<stepper>`** Alloy element to create a stepper 
+    (see [Examples](#!/api/Titanium.UI.iOS.stepper-examples)) .
```
hansemannn: Link won't work, because there is no `Titanium.UI.iOS.stepper`.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+    * When assigning a custom image to the `backgroundImage` property, the image must be
+    partially or wholly transparent for the background color or background gradient to be visible.
+    * Similarly, for an assigned background gradient to show through, the `backgroundColor` may need to be
+    set to 'transparent'.
+
+    *Note that the stepper cannot be re-sized.
```
hansemannn: Missing space between the work and the `*`. It will cause the list to break in generation

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+extends: Titanium.UI.View
+since: "5.4.0"
+excludes:
+    events: [dblclick, doubletap, focus, keypressed, longclick, longpress, pinch, postlayout,
+             singletap, swipe, touchcancel, touchend, touchmove, touchstart, twofingertap]
```
hansemannn: Did you verify, that all these events don't work with the stepper? I would expect that at least `doubleclick` and the `touch*` events work like with other UIControls.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+properties:
+  - name: value
+    summary: The current value of the stepper.
+    type: Number
+
+  - name: continous
```
hansemannn: It's called `continuous`

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+  - name: autorepeat
+    summary: |
+        If YES, the user pressing and holding on the stepper repeatedly alters value.
+        The default value is YES.
+    type: Boolean
+
+  - name: wraps
+    summary: |
+        If YES, incrementing beyond maximumValue sets value to minimumValue; likewise, decrementing below
```
hansemannn: Code-highlight `maximumValue ` and `minimumValue `. Use A point instead of a semicolon.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        minimumValue sets value to maximumValue. If NO, the stepper does not increment beyond maximumValue nor
+        does it decrement below minimumValue but rather holds at those values.
+        The default value is NO.
+    type: Boolean
+
+  - name: minimumValue
+    summary: The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
+    type: Number
+
+  - name: minimumValue
```
hansemannn: Copy-pasted. It's `maximumValue`

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+    summary: The maximum value the stepper will be set to, the value must be greater than the miniumumValue.
+    type: Number
+
+  - name: minimumValue
```
hansemannn: Not sure what this should be, but it's not `minimumValue` again.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        does it decrement below minimumValue but rather holds at those values.
+        The default value is NO.
+    type: Boolean
+
+  - name: minimumValue
+    summary: The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
+    type: Number
+
+  - name: minimumValue
+    summary: The maximum value the stepper will be set to, the value must be greater than the miniumumValue.
```
hansemannn: What happens if the value is equal to the `minimumValue`?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+    type: Number
+
+  - name: minimumValue
+    summary: The value the stepper will decrement and increment by.
+    type: Number
+
+  - name: tintColor
+    summary: Sets the color for the widget, any backgroundImages added will be set to the same color.
+    description: |
+        Default color is blue
```
hansemannn: Missing point. Also think of removing this sentence, because the default tintColor depends on the global tint-configuration set. See https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/TransitionGuide/AppearanceCustomization.html

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+    summary: |
+        Background image for the stepper in its normal state, specified as a local file path or URL.
+    description: |
+        Also sets the background image for other stepper states (focused, selected),
+        unless they've been specified explicitly. On iOS, if backgroundDisabledImage is unset, the
+        image will appear faded to indicate being disabled. Also on iOS, if backgroundSelectedImage
+        is unset, the image will be darkened to indicate being selected.
+    type: String
+
+  - name: backgroundDisabledImage
```
hansemannn: Did you test this? Cannot see a `enabled` property that would cause this state. Please add tests for all states to the FT-example.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+    summary: |
+        Background image for the stepper in its disabled state, specified as a local file path
+        or URL. 
+    description: |
+        Also sets the background image for the other stepper states (normal, focused, selected), 
+        unless they've been specified explicitly, except for iOS, which requires a backgroundImage.
+    type: String
+    platforms: [android, iphone, ipad, mobileweb]
```
hansemannn: Why are these platforms here? Isn't it iOS-only?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+  - name: backgroundSelectedImage
+    summary: |
+        Background image for the stepper in its selected state, specified as a local file 
+        path or URL.
+    type: String
+    platforms: [iphone, ipad]
```
hansemannn: Similar as above: As the API is iOS-only, we don't need to explicitly declare the platform support

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+  - name: decrementImage
+    summary: |
+        Background image for the stepper decrement button in its normal state, specified as a local
+        file path or URL.
+    description: |
+        Also sets the background image for the other stepper states (disabled, focused, selected),
+        unless they've been specified explicitly, except for iOS, which requires a backgroundImage.
+    type: String
+    platforms: [iphone, ipad]
```
hansemannn: Same as above. Please remove the platforms in all other below occurrences as well.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+
+  - name: backgroundImage
+    summary: |
+        Background image for the stepper in its normal state, specified as a local file path or URL.
+    description: |
+        Also sets the background image for other stepper states (focused, selected),
+        unless they've been specified explicitly. On iOS, if backgroundDisabledImage is unset, the
+        image will appear faded to indicate being disabled. Also on iOS, if backgroundSelectedImage
+        is unset, the image will be darkened to indicate being selected.
+    type: String
```
hansemannn: What happens if I use a TiBlob/TiFile?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        file path or URL.
+    description: |
+        Also sets the background image for the other stepper states (disabled, focused, selected),
+        unless they've been specified explicitly, except for iOS, which requires a backgroundImage.
+    type: String
+    platforms: [iphone, ipad]
+
+  - name: incrementDisabledImage
+    summary: |
+        Background image for the stepper increment button in its dissabled state, specified as a local
```
hansemannn: Typo: "dissabled" -> "disabled"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        file path or URL.
+    description: |
+        Also sets the background image for the other stepper states (disabled, focused, selected),
+        unless they've been specified explicitly, except for iOS, which requires a backgroundImage.
+    type: String
+    platforms: [iphone, ipad]
+
+  - name: decrementDisabledImage
+    summary: |
+        Background image for the stepper decrement button in its dissabled state, specified as a local
```
hansemannn: Typo: "dissabled" -> "disabled"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        Background image for the stepper increment button in its selected state, specified as a local
+        file path or URL.
+    type: String
+    platforms: [iphone, ipad]
+
+events:
+  - name: change
+    summary: Fired every time the stepper value changes.
+    properties:
+      - name: Value
```
hansemannn: Lowercase: "Value" -> value

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+    summary: |
+        Background image for the stepper increment button in its selected state, specified as a local
+        file path or URL.
+    type: String
+    platforms: [iphone, ipad]
+
+events:
+  - name: change
+    summary: Fired every time the stepper value changes.
+    properties:
```
hansemannn: Any other properties in that event that would make sense?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,180 @@
...
+        summary: Current value of the stepper.
+        type: Number
+    platforms: [iphone,ipad] 
+
+examples:
+  - title: Simple stepper Example
+    example: |
+
+            var win = Ti.UI.createWindow({
+            backgroundColor: '#fff'
```
hansemannn: Indentation. Please also fix the other lines below

iphone/Classes/TiUIiOSProxy.m
```diff
@@ -715,6 +719,13 @@ -(id)createBlurView:(id)args
 }
 #endif
 
+#ifdef USE_TI_UIIOSSTEPPER
+-(id)createStepper:(id)args
+{
+    return [[[TiUIiOSStepperProxy alloc] _initWithPageContext:[self executionContext] args:args] autorelease];
```
hansemannn: Why not `[self pageContext]`? I remember we should use that for newer API's, but they should work the same.

iphone/Classes/TiUIiOSStepper.h
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+#ifdef USE_TI_UIIOSSTEPPER
+#import "TiUIView.h"
+
+@interface TiUIiOSStepper : TiUIView {
+@private
+    UIStepper *stepper;
+    UIImage *backgroundImageCache;
```
hansemannn: So here are my thoughts on this: Since we subclass the `TiUIView` in every UI element, I think we could do all the backgroundImage stuff on that view instead of adding another view that slows down the UI creation. Doing that, we wouldn't even need all the custom setter and logic to add the image, because the `TiUIView` already does all that by default. 

Please test this and verify it's necessary to do all this.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+#ifdef USE_TI_UIIOSSTEPPER
+#import "TiUIiOSStepper.h"
+#import "TiUIiOSStepperProxy.h"
+#import "TiUtils.h"
+
+@implementation TiUIiOSStepper
+
+-(UIStepper*)stepper
+{
+    if (stepper ==nil)
```
hansemannn: Missing space in `==nil`.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    {
+        stepper = [[[UIStepper alloc]init] retain];
```
hansemannn: Missing space in `alloc]init]`. Also, why do we need to manaully retain the UI element? Don't know of a case where it would be nil here.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+        [stepper setMaximumValue:100];
```
hansemannn: Use the proxy value from `maximumValue` here instead of static values.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+        [stepper setMinimumValue:0];
```
hansemannn: Use the proxy value from `minimumValue ` here instead of static values.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+{
+    [stepper removeTarget:self action:@selector(stepperChanged:) forControlEvents:UIControlEventValueChanged];
+    [stepper removeTarget:self action:@selector(stepperTouch:forEvent:) forControlEvents:UIControlEventTouchDown];
+    RELEASE_TO_NIL(stepper);
+    RELEASE_TO_NIL(backgroundImageCache);
+    [super dealloc];
+}
+-(void)setContinuous_:(id)value
+{
+    [[self stepper] setContinuous:[TiUtils boolValue:value]];
```
hansemannn: Validate the user input before: `ENSURE_TYPE(value, NSNumber);`

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+
+-(void)setAutorepeat_:(id)value
+{
+    [[self stepper] setAutorepeat:[TiUtils boolValue:value]];
```
hansemannn: Same as above, missing input validation.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+
+-(void)setWraps_:(id)value
+{
+   [[self stepper] setWraps:[TiUtils boolValue:value]];
```
hansemannn: Same as above, missing input validation.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+
+-(void)setMaximumValue_:(id)value
+{
+    ENSURE_TYPE(value, NSNumber)
```
hansemannn: I like to do a closing semicolon after macros to unify the style with other methods in Obj-C, but that's up to you :blush:

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+-(void)setMinimumValue_:(id)value
+{
+    ENSURE_TYPE(value, NSNumber)
+    [[self stepper] setMinimumValue:[TiUtils doubleValue:value]];
+}
+
+-(void)setStepValue_:(id)value
+{
+    ENSURE_TYPE(value, NSNumber)
+    NSNumber *stepValue = (NSNumber*)value;
```
hansemannn: Use `NSNumber *stepValue = [TiUtils doubleValue: value]` instead.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    if (stepValue > 0) {
+        [[self stepper] setStepValue:[TiUtils doubleValue:value]];
+    } else {
+        NSLog(@"[WARN]: The stepValue must be bigger than 0");
```
hansemannn: Remove the `:` and don't forget the point at the end.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+-(void)setTintColor_:(id)value
+{
+    [[self stepper] setTintColor:[[TiUtils colorValue:value] color]];
```
hansemannn: Missing input validation for `TiColor` values.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#pragma mark backgroundImage
+-(void)setBackgroundImage_:(id)value
```
hansemannn: Missing blank-line after the `#pragma`

hansemannn: As discussed above, think of doing all the image-stuff using the superclass instead of re-implementing it.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+
+-(void)setBackgroundDisabledImage_:(id)value
+{
+    backgroundImageCache = [self imageWithImage:[self loadImage:value] fullScale:YES];
+    [[self stepper] setBackgroundImage:backgroundImageCache forState:UIControlStateDisabled];
+}
+
+#pragma mark Decrement backgroundImage
+-(void)setDecrementImage_:(id)value
```
hansemannn: Missing blank-line.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    backgroundImageCache = [self imageWithImage:[self loadImage:value] fullScale:NO];
+    [[self stepper] setDecrementImage:[self loadImage:value] forState:UIControlStateSelected];
+}
+
+-(void)setDecrementDisabledImage_:(id)value
+{
+    [[self stepper] setDecrementImage:[self loadImage:value] forState:UIControlStateDisabled];
+}
+
+#pragma mark Increment backgroundImage
```
hansemannn: Missing blank-line.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+}
+-(void)setTintColor_:(id)value
+{
+    [[self stepper] setTintColor:[[TiUtils colorValue:value] color]];
+}
+
+#pragma mark backgroundImage
+-(void)setBackgroundImage_:(id)value
+{
+    backgroundImageCache = [self imageWithImage:[self loadImage:value] fullScale:YES];
```
hansemannn: Validate the input before (if you decide to use this solution)

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+{
+    // since this guy only works with touch events, we always want them
+    // just always return YES no matter what listeners we have registered
+    return YES;
+}
+
+- (IBAction)stepperChanged:(id)sender
+{
+    NSNumber *newValue = [NSNumber numberWithDouble:[[self stepper] value]];
+    [self.proxy replaceValue:newValue forKey:@"value" notification:YES];
```
hansemannn: Make sure you also call this method on every other proxy-setter like `mininumValue` etc. Otherwise, the properties won't be accessible after creating the proxy e.g. `stepper.mininumValue` etc.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    
+    if ([self.proxy _hasListeners:@"change"]) {
+        [self.proxy fireEvent:@"change" withObject:[NSDictionary dictionaryWithObject:newValue forKey:@"value"]];
+    }
+}
+
+- (IBAction)stepperTouch:(id)sender forEvent:(UIEvent *)event
```
hansemannn: Isn't the click event handled using the super-class (`TiUIView`)?

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+{
+    UITouch *touch = [[event allTouches] anyObject];
+    if([self.proxy _hasListeners:@"click"]) {
+        [self.proxy fireEvent:@"click" withObject:[NSDictionary
+                                                   dictionaryWithDictionary:[TiUtils touchPropertiesToDictionary:touch
```
hansemannn: The format looks messed-up

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+                                                                                andPoint:[touch locationInView:self]]]];
+    }
+}
+
+
+-(UIImage *)imageWithImage:(UIImage *)image fullScale:(bool)full {
+    //UIGraphicsBeginImageContext(newSize);
+    // In next line, pass 0.0 to use the current device's pixel scaling factor (and thus account for Retina resolution).
+    // Pass 1.0 to force exact pixel size.
+    [self stepper];
```
hansemannn: We talked about this during the fixes for ListView: Don't call the getter before you use it. Or: Think about a case where the stepper could be nil in this case and initialize it before this place.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -0,0 +1,204 @@
...
+    
+    if (image.size.height > bounds.size.height && image.size.width > bounds.size.width) {
+        UIImageView* theView = [[UIImageView alloc] initWithFrame:bounds];
+        [theView setImage:image];
+        UIGraphicsBeginImageContextWithOptions(bounds.size, [theView.layer isOpaque], 0.0);
+        [theView.layer renderInContext:UIGraphicsGetCurrentContext()];
+        UIImage *image = UIGraphicsGetImageFromCurrentImageContext();
+        UIGraphicsEndImageContext();
+        [theView release];
+        return [image retain];
```
hansemannn: Is this logic from somewhere else? Where do you release the `image`?

iphone/Classes/TiUIiOSStepperProxy.m
```diff
@@ -0,0 +1,49 @@
...
+
+-(NSString*)apiName
+{
+    return @"Ti.UI.iOS.Stepper";
+}
+
+-(void)_initWithProperties:(NSDictionary *)properties
+{
+    [super _initWithProperties:properties];
+}
```
hansemannn: Not necessary, if you don't do any custom logic inside the proxy constructor.

iphone/Classes/TiUIiOSStepperProxy.m
```diff
@@ -0,0 +1,49 @@
...
+-(TiDimension)defaultAutoHeightBehavior:(id)unused
+{
+    return TiDimensionAutoSize;
+}
+#endif
+
+-(UIView *)parentViewForChild:(TiViewProxy *)child
+{
+    return [[(TiUIiOSStepper*)[self view] stepper] superview];
```
hansemannn: Necessary? What does it do?

iphone/Classes/TiUIiOSStepperProxy.m
```diff
@@ -0,0 +1,49 @@
...
+}
+
+-(NSNumber*)value
+{
+    NSNumber *value = [NSNumber numberWithDouble:[[(TiUIiOSStepper*)[self view] stepper] value]];
+    return value;
```
hansemannn: Create a getter for your `TiUIiOSStepper` class and then call it with:
```objc
return NUMDOUBLE([[[self stepper] stepper] value]);
```

hansemannn:
Left some comments, will do the FT after the comments are addressed.

iphone/Classes/TiUIiOSStepper.h
AngelkPetkov: if i don't store the image inside a variable before setting it , it doesn't seem to want to get displayed.  Im following the same logic we do inside the button class.

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: Ive added this so if the user doesn't set a value this will be the default one. If a user sets a value it will be set and the value set here will be overwritten.

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: You cant initialize an NSNumber with a simple C type such as double , which is why I'm casting it :P. Thats what i actually tried first but the compiler didn't like it.

hansemannn: Pretty sure you can, since it calls the double-initializer. We do that in many places.

AngelkPetkov: No you definitely cant do it by doing NSNumber *stepValue = [TiUtils doubleValue: value]. I tried it before and just tried it again you get an error.

hansemannn: I missed the constructor I talked about above:
```
NSNumber *stepValue = [NSNumber numberWithDouble:[TiUtils doubleValue:value]];
```

hansemannn: So you can call `[[self stepper] setStepValue:[NSNumber numberWithDouble:[TiUtils doubleValue:value]]];` directly

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: This is called every time you decrement or increment , I'm not sure what you mean ? Changing the maximum value or minimum shouldn't effect the current value unless the maximum value is smaller than the current value or the minimum is larger. In which case it will change internally and then this will be called. Or if not soon as you increment/decrement it will be updated.

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: Thats there uncase this somehow gets called while the stepper is nil , which did happen and caused crash, this way if nil it will be initialized.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -5,35 +5,40 @@ summary: |
...
     set to 'transparent'.
 
-    *Note that the stepper cannot be re-sized.
+    * Note that the stepper cannot be re-sized.
 
 extends: Titanium.UI.View
 since: "5.4.0"
 excludes:
-    events: [dblclick, doubletap, focus, keypressed, longclick, longpress, pinch, postlayout,
-             singletap, swipe, touchcancel, touchend, touchmove, touchstart, twofingertap]
```
hansemannn: Did you validate the events work?

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -48,28 +53,29 @@ properties:
...
-    summary: The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
-    type: Number
-
-  - name: minimumValue
-    summary: The maximum value the stepper will be set to, the value must be greater than the miniumumValue.
+  - name: minimum
+    summary: |
+        The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
+        If you attempt to set a value equal to or greater than maximumValue, the system will throw
+        an NSInvalidArgumentException exception.
```
hansemannn: `maximumValue`, also the `NSInvalidArgumentException` is not thrown and should be caught on the native side. Same with the property below.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -28,107 +28,115 @@ -(UIStepper*)stepper
...
 }
 
-#pragma mark backgroundImage
--(void)setBackgroundImage_:(id)value
+-(void)setEnabled_:(id)value
 {
-    backgroundImageCache = [self imageWithImage:[self loadImage:value] fullScale:YES];
-    [[self stepper] setBackgroundImage:backgroundImageCache forState:UIControlStateNormal];
+    ENSURE_TYPE(value, NSNumber);
+    [stepper setEnabled:[TiUtils boolValue:value]];
```
hansemannn: Use `[self stepper]`

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -28,107 +28,115 @@ -(UIStepper*)stepper
...
 }
 
--(void)setBackgroundSelectedImage_:(id)value
+-(double)getValue
```
hansemannn: Objective-C getter don't have the `get`-Prefix, so just `- (double)value`. Why do we need this?

apidoc/Titanium/UI/iOS/Stepper.yml
AngelkPetkov: Yeah ,  all the ones i didn't take out work. dblcick and all the touchEvents. Added an new method to check events, same one as button class. As its just 2 buttons inside the view for the stepper.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -160,21 +168,54 @@ - (IBAction)stepperChanged:(id)sender
     [self.proxy replaceValue:newValue forKey:@"value" notification:YES];
     
     if ([self.proxy _hasListeners:@"change"]) {
-        [self.proxy fireEvent:@"change" withObject:[NSDictionary dictionaryWithObject:newValue forKey:@"value"]];
+        [self.proxy fireEvent:@"change" withObject:[NSDictionary dictionaryWithObjectsAndKeys:
+                                                    [NSNumber numberWithDouble:stepper.value],@"value",
```
hansemannn: You can use `NUMDOUBLE()` here, makes it easier to read.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -160,21 +168,54 @@ - (IBAction)stepperChanged:(id)sender
...
+                fireActionEvent = [touch tapCount] == 1 ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
+            }
+            break;
+        case UITouchPhaseCancelled:
+            touchStarted = NO;
+            fireEvent = @"touchcancel";
+            break;
+        default:
+            return;
+    }
```
hansemannn: Do we need to do this manually? Want to make sure we don't use business-logic multiple times

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -160,21 +168,54 @@ - (IBAction)stepperChanged:(id)sender
...
+    //[self setHighlighting:button.highlighted];
+    NSMutableDictionary *evt = [NSMutableDictionary dictionaryWithDictionary:[TiUtils touchPropertiesToDictionary:touch andPoint:[touch locationInView:self]]];
```
hansemannn: Just as a hint: I love to use `NSMutableDictionary *evt = [[TiUtils touchPropertiesToDictionary:touch andPoint:[touch locationInView:self]] mutableCopy];` here, but doesn't make any difference.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -160,21 +168,54 @@ - (IBAction)stepperChanged:(id)sender
...
-- (IBAction)stepperTouch:(id)sender forEvent:(UIEvent *)event
+- (void)controlAction:(id)sender forEvent:(UIEvent *)event
 {
     UITouch *touch = [[event allTouches] anyObject];
-    if([self.proxy _hasListeners:@"click"]) {
-        [self.proxy fireEvent:@"click" withObject:[NSDictionary
-                                                   dictionaryWithDictionary:[TiUtils touchPropertiesToDictionary:touch
-                                                                                andPoint:[touch locationInView:self]]]];
+    NSString *fireEvent;
+    NSString * fireActionEvent = nil;
```
hansemannn: Typo

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: Yeah we do , other wise the events don't get called. Its the same used inside the button class and the stepper are just 2 buttons inside a view as i mentioned above.

iphone/Classes/TiUIiOSStepper.m
```diff
@@ -160,21 +168,54 @@ - (IBAction)stepperChanged:(id)sender
...
+            fireEvent = @"touchstart";
+            break;
+        case UITouchPhaseMoved:
+            fireEvent = @"touchmove";
+            break;
+        case UITouchPhaseEnded:
+            touchStarted = NO;
+            fireEvent = @"touchend";
+            if (stepper.highlighted) {
+                fireActionEvent = [touch tapCount] == 1 ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
```
hansemannn: Difference between `fireEvent` and `fireActionEvent` here? Both could be fired with `touchend` and `click`

iphone/Classes/TiUIiOSStepperProxy.m
```diff
@@ -35,15 +31,10 @@ -(TiDimension)defaultAutoHeightBehavior:(id)unused
...
--(UIView *)parentViewForChild:(TiViewProxy *)child
-{
-    return [[(TiUIiOSStepper*)[self view] stepper] superview];
-}
 
 -(NSNumber*)value
 {
-    NSNumber *value = [NSNumber numberWithDouble:[[(TiUIiOSStepper*)[self view] stepper] value]];
-    return value;
+    return NUMDOUBLE([(TiUIiOSStepper*)[self view] getValue]);
```
hansemannn: Ok, here we use it. You could just return `NUMDOUBLE([[self stepper] value])`, or am I missing something?

AngelkPetkov: The proxy doesnt have a refrence to the stepper you cant do [self stepper]

iphone/Classes/TiUIiOSStepper.m
AngelkPetkov: We need need both of the string property for touchend to work along side click and dblclick. If only one property is used the fireActionEvent will get set to click and dblclick always as the button will be highlighted.

iphone/Classes/TiUIiOSProxy.m
AngelkPetkov: I was just making it the same one as all the classes above including the blurView, will change it back to [self pageContext].

hansemannn:
- `stepValue` is not documented, also think of renaming to `steps`
- `value` setter is missing (currently creation only)
- `continuous` / `autorepeat` / `wraps` missing default values (check native docs)

AngelkPetkov:
Addressed Comments , i did not add a setValue method as it will be called every time you increment/decrement, we will need to check if the value passed in is greater than the min and larger than max. Which will cause an issue when incrementing because once you reach the max it will be printed. IF you're incrementing by 3 for example and the max is 29.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -48,28 +53,35 @@ properties:
 
   - name: wraps
     summary: |
-        If YES, incrementing beyond maximumValue sets value to minimumValue; likewise, decrementing below
-        minimumValue sets value to maximumValue. If NO, the stepper does not increment beyond maximumValue nor
-        does it decrement below minimumValue but rather holds at those values.
+        If YES, incrementing beyond <Ti.UI.iOS.Stepper.maximum> sets value to <Ti.UI.iOS.Stepper.minimum>. likewise, decrementing below
+        <Ti.UI.iOS.Stepper.minimum> sets value to <Ti.UI.iOS.Stepper.maximum>. If NO, the stepper does not increment beyond <Ti.UI.iOS.Stepper.maximum> nor
```
hansemannn: Shouldnt it set the value to minimum of we set it below the minimum? Same with the maximum, make sure the bounds are correct.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -48,28 +53,35 @@ properties:
...
+        does it decrement below <Ti.UI.iOS.Stepper.minimum> but rather holds at those values.
         The default value is NO.
     type: Boolean
 
-  - name: minimumValue
-    summary: The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
+  - name: minimum
+    summary: |
+        The miniumum value the stepper will be set to, the value must be smaller than the maximumValue.
```
hansemannn: Typo in "miniumum"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -48,28 +53,35 @@ properties:
...
+        If you attempt to set a value equal to or greater than maximumValue, the system will default the 
+        value to 0.
     type: Number
 
-  - name: minimumValue
-    summary: The maximum value the stepper will be set to, the value must be greater than the miniumumValue.
+  - name: maximum
+    summary: |
+        The maximum value the stepper will be set to, the value must be greater than the miniumumValue.
```
hansemannn: Typo in "miniumum", also "minimum", not "minimumVaue"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -48,28 +53,35 @@ properties:
...
+        If you attempt to set a value equal to or lower than minimumValue, the system will default the 
+        value to 100.
     type: Number
 
-  - name: minimumValue
-    summary: The value the stepper will decrement and increment by.
+  - name: steps
+    summary: |
+      The value the stepper will increment and decrement by, default value for this property is 1.When setting
```
hansemannn: Typo, missing space.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -108,21 +103,14 @@ properties:
...
         unless they've been specified explicitly, except for iOS, which requires a backgroundImage.
     type: String
-    platforms: [iphone, ipad]
 
   - name: decrementDisabledImage
     summary: |
-        Background image for the stepper decrement button in its dissabled state, specified as a local
-        file path or URL.
+        Background image for the stepper decrement button in its disabled state, specified as a local
+        file path or URL. The decrement button enteres a disabled state ones the value is equal to the
```
hansemannn: Typo, it's "enters"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -132,39 +120,67 @@ properties:
...
+      - name: minimum
+        summary: The minimum value of the stepper.
+        type: Number
+      - name: maximum
+        summary: The maximum value of the stepper.
+        type: Number
 
 examples:
-  - title: Simple stepper Example
+  - title: Simple Stepper Example
```
hansemannn: Typo: "Simple stepper example"

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -132,39 +120,67 @@ properties:
...
-        win.open();
\ No newline at end of file
+        Previous example as an Alloy view.
+
+        stepper.xml:
+
+            <Alloy>
+                <Window class="container">
+                    <Stepper id="stepper" onClick="doClick" 
+                         top="30" tintColor="green" steps = "3"  />
```
hansemannn: Attributes in one line, no space in `steps = "3"`

hansemannn:
Left some comments. Regarding the `setValue`: It is important to be exposed because it's **very** common to set the value manually after creating the control. We need to take care of that and do the bounds check accordinately.

apidoc/Titanium/UI/iOS/Stepper.yml
AngelkPetkov: This is correct, think of it like a cycle if you keep going up and go past the maximum value. You start from the beginning. Its not a validation where if ,for example the value is less than 0 we set it to 0.

AngelkPetkov:
Addressed Comments , also added the setValue. Disregard what i said before, forgot to turn notification to NO, which fixes the issue of setValue being called upon increment/decrement.

apidoc/Titanium/UI/iOS/Stepper.yml
```diff
@@ -0,0 +1,186 @@
...
+      <tr>
+        <td><img src="images/stepper/Stepper.png" height="50" /></td>
+        <td><img src="images/stepper/Stepper_customImages.png" height="51" /></td>
+      </tr>
+      <tr><th colspan="2">iOS</th></tr>
+    </table>
+
+    A Stepper displays its selected state while it is being pressed.
+    You can specify background images for each state on iOS.
+    Use the <Titanium.UI.createStepper> method or a **`<stepper>`** Alloy element to create a stepper
```
hansemannn: It's `<stepper>` (capitalized), Alloy wouldn't be able to handle it if not.

hansemannn:
Approved!

PR #7885, opened 2016-03-25
[TIMOB-20513] iOS: Fix Ti.UI.ActivityIndicator "message" in AutoLayout

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20513


pec1985:
We're going to have to look for a different solution. With this fix, the text is bigger than the label and the parent `backgroundView`. Change the background color of the activity indicator and you'll see what I man.
PR Rejected.

PR #7886, opened 2016-03-25
[TIMOB-20406] iOS: Make Ti.UI.ProgressBar width default to FILL instead of SIZE

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20406


AngelkPetkov:
CR and FT approved. Merging!

PR #7887, opened 2016-03-26
[TIMOB-20403] iOS: Fix Ti.UI.ProgressBar default values

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20403

### DIFF
--- a/iphone/Classes/TiUIProgressBar.m
+++ b/iphone/Classes/TiUIProgressBar.m
@@ -21,13 +21,13 @@
 }
 #endif
 
--(id)initWithStyle:(UIProgressViewStyle)style_
+-(id)initWithStyle:(UIProgressViewStyle)_style andMinimumValue:(CGFloat)_min maximumValue:(CGFloat)_max;
 {
 	if (self = [super initWithFrame:CGRectZero])
 	{
-		style = style_;
-		min = 0;
-		max = 1;
+		style = _style;
+		min = _min;
+		max = _max;
 		[self setHidden:YES];
         
 #ifdef TI_USE_AUTOLAYOUT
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIProgressBar.m
```diff
@@ -21,13 +21,13 @@ -(void)initializeTiLayoutView
...
+-(id)initWithStyle:(UIProgressViewStyle)_style andMinimumValue:(CGFloat)_min maximumValue:(CGFloat)_max;
 {
 	if (self = [super initWithFrame:CGRectZero])
 	{
-		style = style_;
-		min = 0;
-		max = 1;
+		style = _style;
+		min = _min;
+		max = _max;
```
AngelkPetkov: I see why your using _min as the parameter cannot be named same as the property.However we could do [self setMax:max] or something else.  I don't like how  _variable for parameter looks, don't have to change to just a suggestion :P.

hansemannn: We don't do that, because they are just variables, no properties (with no getter and setter methods). That's why I used the underscore-assignment style.

AngelkPetkov: I know but we have setter methods for min and max. Not for style , i guess doing that much for it would be redundant however so never mind.

hansemannn: Yeah, but that are proxy-methods which expect values from the type `id`. Downcasting that just to use a setter didn't make sense to me. For me it's common to use this kind of setters in this case.

AngelkPetkov:
FT and CR approved , merging !

PR #7894, opened 2016-03-30
[TIMOB-20614]iOS: Ti proxies are not always released from memory

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-20614)
Releasing CF type properly and adding autorealse to krollObject for registerProxy.


AngelkPetkov:
Closing PR, as after further investigation the object were actually being cleaned out of the memory, they just appeared to stay in memory as instruments wasn't refreshed.

PR #7896, opened 2016-03-31
[TIMOB-19908] [TIMOB-20493]Fix scrollable view to work with autolayout and run-on-main-thread enabled

cheekiatng:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-19908
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20493

### DIFF
--- a/iphone/Classes/TiUIScrollableView.m
+++ b/iphone/Classes/TiUIScrollableView.m
@@ -128,6 +128,7 @@
         [_scrollView addConstraints:TI_CONSTR(@"H:|[_contentView(>=_scrollView)]|", views)];
         NSString *dotsViewConstraints = [NSString stringWithFormat:@"V:[_dotsView(%f)]",pageControlHeight];
         [self addConstraints:TI_CONSTR(dotsViewConstraints, views)];
+        [NSLayoutConstraint deactivateConstraints:[_dotsView constraints]];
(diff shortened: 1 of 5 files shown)


iphone/Classes/TiUIScrollableView.m
```diff
@@ -169,6 +173,12 @@ -(void)layoutSubviews
 [wrapperView setViewName: TI_STRING(@"scrollable.wrapper.view%lu", (unsigned long)[[self subviews] count])]; \
 [wrapperView addSubview:view]; \
 
+
+-(void)removeSubview:(nonnull UIView *)view
+{
+    [[view superview] removeFromSuperview];
```
hansemannn: If we remove the superview of the supplied view from the superview..wouldn't that remove the scrollableview from it's parent? I'm sure it works well, just trying to understand.

cheekiatng: the superview in this regard, is NOT scrollableview. It's actually a TiLayoutView, that is an extra layer BETWEEN the scrollableview and said subview, when auto layout is enabled. :)

cheekiatng:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/7914

PR #7900, opened 2016-03-31
[TIMOB-19325] Android: activity indicator causing crash if exiting ap…

falkolab:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-19325

Tested with test case from ticket on device Lenovo A630.


sgtcoolguy:
Awesome, thanks for the PR. Looks good to me. Merging.

PR #7907, opened 2016-04-02
[DOCFIX] Fix docgen compiler warnings, abstract params, ListView insert event, dead links

hansemannn:
This PR fixes 26 docs validation errors occurring after the abstract params have been moved to properties. The problem they didn't work before was, that every abstract param need `---` before, not only once per section.

Also adds the missing docs for the Ti.UI.ListView `insert` event and resolves all compiler warnings when using the docgen script and therefore also fixes all API deadlinks.

Validated everything again and it seems to work fine now. @jawa9000 Please confirm by generating the docs, thanks!


jawa9000:
I'm gonna merge these for now but I'll do a review locally to confirm there are no doc pub issues.

PR #5492, opened 2014-03-19
[TIMOB-16512] Android: added left and right padding to TextField

bijupmb:
https://jira.appcelerator.org/browse/TIMOB-16512

PR #7908, opened 2016-04-02
[TIMOB-16512] Android: added padding to TextField

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16512

Updating the https://github.com/appcelerator/titanium_mobile/pull/5492 PR so can be used with the current branch again. All the credit goes to @bijupmb .


hansemannn:
Please add docs, thanks!

hansemannn:
Thank you. Please also add the `since: "5.4.0"` key so people now since when it's available. For the properties that are already available on iOS, you can check [this example](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/Label.yml#L68) on how the syntax works.

hansemannn:
Will FT now and merge, looks all good! Maybe we should think of moving the padding to a more clean way (on both iOS and Android) with something like this:
```js
var field = Ti.UI.createTextField({
    padding: {
        left: 10,
        right: 10,
        top: 10,
        bottom: 10
    }
});
```
Does that make sense to you? So we would handle it like we do with listSeparatorInsets and other "bounds-like" API's. But that should be covered in a separate ticket.

m1ga:
@hansemannn sounds like a good idea and looks better. Should I change it for this PR or leave it like this for the moment so it's the same on iOS?

hansemannn:
Let's leave like it is for now and I'll file a ticket for it the next days. Thanks!

hansemannn:
You know what, let's do that change in this PR, I will setup one for iOS and we merge both for 5.4.0. Otherwise, we would change and deprecate the whole thing in 5.5.0 / 6.0.0 again and that produces even more work.

m1ga:
Ok, I'll get the android part ready

m1ga:
Changed it ;) I've added `ios,ipad` in the documentation already

apidoc/Titanium/UI/TextField.yml
```diff
@@ -695,3 +691,24 @@ properties:
...
+    type: Number
+    summary: Right padding
+    
+  - name: top
+    type: Number
+    summary: Top padding
+    
+  - name: bottom
+    type: Number
+    summary: Bottom padding
```
hansemannn: iOS only allows left and right padding because an input should always be vertically centered. We should consider doing the same with Android, so we don't need to handle different behaviors across platforms. Also please deprecate the old properties.

m1ga: Done!
* syntex: `padding: {left:10,right:10}`
* updated [example on jira](https://jira.appcelerator.org/browse/TIMOB-16512?focusedCommentId=381355&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-381355)

m1ga: TextArea will still be TOP/LEFT, TextField is always vertically aligned

hansemannn:
Alright! Now you decide: Are you fine with me cherry-picking your commits and setup a new PR including iOS and Android to have the platforms combinded in one PR, or do you want me to merge yours first and setup another one? 😊

m1ga:
See ti-slack: if its less work with cherry-picking feel free to create a new PR with iOS and Android!

ashcoding:
Tested and reviewed this with app.js
```javascript
var win =  Ti.UI.createWindow({
    layout: "vertical",
    backgroundColor: "#fff"
});
 
var t1 =  Ti.UI.createTextField({
    value: "lllllllllmmmmmmmmmmmmmmlllllllllll",
    padding: {
        left: 20,
        right: 0
    },
    width: 100,
    height: Ti.UI.SIZE,
    borderColor: "#000",
    color: "#000",
    borderWidth: 1,
    top: 10
});

 
var t2 = Ti.UI.createTextField({
    value: "lllllllllmmmmmmmmmmmmmmlllllllllll",
    padding: {
        right: 20,
        left: 0
    },
    width: 100,
    height: Ti.UI.SIZE,
    borderColor: "#000",
    color: "#000",
    borderWidth: 1,
    top: 10
});
 
var t3 = Ti.UI.createTextField({
    value: "lllllllllmmmmmmmmmmmmmmlllllllllll",
    padding: {
        left: 20,
        right: 20
    },
    width: 100,
    height: Ti.UI.SIZE,
    borderColor: "#000",
    color: "#000",
    borderWidth: 1,
    top: 10
});
 
var t4 = Ti.UI.createTextField({
    value: "lllllllllmmmmmmmmmmmmmmlllllllllll",

    width: 100,
    height: 100,
    borderColor: "#000",
    color: "#000",
    borderWidth: 1,
    top: 10
});

 
win.add(t1);
win.add(t2);
win.add(t3);
win.add(t4);
 
win.open();

t4.padding = {left: 50, right: 50};
```

Code working fine and looks good.
@hansemannn I assume that you are preparing iOS PR for parity?
We need to do either:-
a) Cherry-pick this and put it in your iOS PR and update the docs for 6.0.0 ( Cause we are close to freezing for 5.4.0 ) and merge this for 6.0.0
b) Merge this with iOS PR soon.

Is there a link for iOS PR?

ashcoding:
I think we should push this for 6.0.0. That way, we won't need to rush this. @hansemannn 
We can update the docs later.

ashcoding:
Code reviewed and tested. **Approved**
Will leave it to @hansemannn to merge when he gets the iOS portion in.

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -45,6 +45,28 @@ methods:
...
+    
+  - name: paddingLeft
+    summary: Left padding of this text field.
+    type: Number
+    platforms: [iphone, ipad]
+    deprecated:
+        since: "5.4.0"
+        notes: Use padding instead    
+    
+  - name: paddingRight
```
hansemannn: So `paddingLeft` and `paddingRight` haven't been documented for `Ti.UI.TextArea` before?

hansemannn: So `paddingLeft` and `paddingRight` haven't been documented for `Ti.UI.TextArea` before?

hansemannn: So `paddingLeft` and `paddingRight` haven't been documented for `Ti.UI.TextArea` before?

ashcoding:
Just a note, this will be for 6.0.0.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -214,8 +214,34 @@ public void processProperties(KrollDict d)
...
+		
 	}
 
+	private void setTextPadding(HashMap<String, Object> d)
+	{
+		int paddingLeft = 0;
+		int paddingRight = 0;
+		
+		if (d.containsKey(TiC.PROPERTY_LEFT)) {
+			paddingLeft = TiConvert.toInt(d, TiC.PROPERTY_LEFT);
```
hieupham007: Please use TiConvert methods that with a fall back default in case the input is invalid like null and such. Something like this TiConvert.toInt(d, TiC.PROPERTY_LEFT, someDefaultValue).

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -214,8 +214,34 @@ public void processProperties(KrollDict d)
...
+		} else {
+			paddingLeft = tv.getPaddingLeft();
+		}
+		if (d.containsKey(TiC.PROPERTY_RIGHT)) {
+			paddingRight = TiConvert.toInt(d, TiC.PROPERTY_RIGHT);
```
hieupham007: same here

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -45,6 +45,28 @@ methods:
...
+    summary: Right padding of this text field.
+    type: Number
+    platforms: [iphone, ipad]
+    deprecated:
+        since: "5.4.0"
```
hieupham007: This should be 6.0.0

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -519,3 +541,16 @@ properties:
   - name: length
     summary: Number of characters selected.
     type: Number
+---    
+name: Padding
+summary: Dictionary object of parameters for the <Titanium.UI.TextArea.padding>  that describes the padding
+since: "5.4.0"
```
hieupham007: Here as well, and a couple instances in TextField.yml

ashcoding:
@hieupham007 This looks like it's good to go.

ashcoding:
**Merging**

m1ga:
@hansemannn was the iOS part merged into 6.0.0? I've tried to set `padding: {left:10,right:10}` to a TextField today and it wasn't working. If its not in the documentation needs to be changed

hansemannn:
@m1ga Not, yet. Updated in https://github.com/appcelerator/titanium_mobile/commit/3f0bb974c0cedbc781c383b46d4ed0f200dbd4aa

m1ga:
:+1:

jpriebe:
It looks like this never got merged.   Is it ever going to get merged?  

Is there any reason that top/bottom padding isn't included in this PR, the way it is in https://github.com/appcelerator/titanium_mobile/pull/5492 ?

BTW -- there's a typo in TextField.yml (https://github.com/appcelerator/titanium_mobile/commit/3f0bb974c0cedbc781c383b46d4ed0f200dbd4aa) - "descrioption" instead of "description"

hansemannn:
@jpriebe This PR was merged on June 26 2016 (see above). For the typo: It is not actually an issue as later commits made the API cross-platform, making the description obsolet. For the top/bottom: Maybe @m1ga knows more about that.

m1ga:
@jpriebe see this comment: https://github.com/appcelerator/titanium_mobile/pull/7908#discussion_r60872477
it is because of parity reasons. I've added top/bottom to but removed since it is not part of iOS.

jpriebe:
Thanks, guys.  I was mistakenly looking at TiUIEditText instead of TiUIText.  

After some experimentation, I have been able to get padding to work on android, but I have found that vertical padding of 0 doesn't really mean 0 from a visual standpoint.  It looks like there is about 12dp of padding on top and bottom that can't be removed.  

When you try to set the size of a text field to something slightly smaller than the default (e.g. 30dp), the padding remains, and the text is cut off.

I guess the bottom line is that there's no way to create a compact TextField on android.  (I know, it violates all sorts of usability guidelines to create something that small, but I think there are some legitimate applications.

PR #7912, opened 2016-04-04
[TIMOB-20374] Android: Longpress event on ScrollView doesn't fire.

sriks:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20374
TIMOB-20374 Android: Longpress event on ScrollView doesn't fire.


ashcoding:
Code reviewed and functionally tested to be working.
**Accepted**

PR #7915, opened 2016-04-04
[TIMOB-20465] iOS: Added a keyboardPress Event

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-20465)


apidoc/Titanium/UI/TextField.yml
```diff
@@ -573,6 +573,14 @@ events:
         summary: Value of this text area.
         type: String
 
+  - name: keyPressed
```
hansemannn: It is `keypressed`: http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.TextField-event-keypressed

Also, in Android the event should already by there in every API, so we just need to add the platforms to the related docs.

iphone/Classes/TiUIView.h
```diff
@@ -279,6 +279,7 @@ void ModifyScrollViewForKeyboardHeightAndContentHeightWithResponderRect(UIScroll
 - (void)processTouchesMoved:(NSSet *)touches withEvent:(UIEvent *)event;
 - (void)processTouchesEnded:(NSSet *)touches withEvent:(UIEvent *)event;
 - (void)processTouchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event;
+-(void)processKeyPressed:(id)key;
```
hansemannn: Missing space, also you can use an `NSString` type here, because it's not on proxy-level and only can be a String as far as I know

iphone/Classes/TiUIView.m
```diff
@@ -1375,6 +1375,14 @@ - (void)processTouchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
 	}
 }
 
+- (void)processKeyPressed:(id)key
+{
+    if ([self.proxy _hasListeners:@"keyPressed"]) {
+        NSDictionary *evt = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:key],@"keyCode", nil];
+        [self.proxy fireEvent:@"keyPressed" withObject:evt];
```
hansemannn: Looks fine! In the last months, we started doing the short-hand implementation of events like this:
```objc
[self.proxy fireEvent:@"keyPressed" withObject:@{@"keyCode": [TiUtils stringValue:key]}];
```

hansemannn:
PR looks good! Only left a few comments we could improve, but that should work out fine.

ericpolkadoc:
Hi -- this will be a great addition, thanks for getting this together.  Will it also work on the keyboard to get the keycode?

AngelkPetkov:
Hi, yes the keyCode will be returned regardless if you're using the device keyboard or if you are using the mac keyboard on the simulator.

ericpolkadoc:
sounds great!

hansemannn:
Approved!

PR #7916, opened 2016-04-05
[TIMOB-23124] Fix build fail with auto layout enabled 

cheekiatng:
**JIRA:**https://jira.appcelerator.org/browse/TIMOB-23124


hansemannn:
CR + FT passed, works like a charm! Updated AL-tests to incorporate with these changes. Thank you @cheekiatng! :rocket:

PR #7923, opened 2016-04-06
[TIMOB-23150] iOS: EventListener issues in ScrollableView for IOS.

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23150?filter=-1)

Fixed issue with JS object  being garbage collected causing the views to have no bridge with the objective-C objects.

### DIFF
--- a/iphone/Classes/TiUIScrollableViewProxy.m
+++ b/iphone/Classes/TiUIScrollableViewProxy.m
@@ -73,9 +73,6 @@
 
 -(void)setViews:(id)args
 {
-#if defined(TI_USE_AUTOLAYOUT) || defined(TI_USE_KROLL_THREAD)
-	ENSURE_UI_THREAD(setViews, args)
-#endif
 	ENSURE_ARRAY(args);
 	for (id newViewProxy in args)
 	{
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIScrollableViewProxy.m
```diff
@@ -73,9 +73,6 @@ -(NSUInteger)viewCount
 
 -(void)setViews:(id)args
 {
-#if defined(TI_USE_AUTOLAYOUT) || defined(TI_USE_KROLL_THREAD)
```
cheekiatng: This line was a critical issue for the following tickets. can you make sure that your change didn't affect these tickets when RUN-ON-MAIN-THREAD = FALSE and AUTO-LAYOUT = FALSE.
https://jira.appcelerator.org/browse/TIMOB-20493
https://jira.appcelerator.org/browse/TIMOB-20218
https://jira.appcelerator.org/browse/TIMOB-20199
https://jira.appcelerator.org/browse/TIMOB-20196
https://jira.appcelerator.org/browse/TIMOB-20138
https://jira.appcelerator.org/browse/TIMOB-19986
https://jira.appcelerator.org/browse/TIMOB-19908

AngelkPetkov: Just tested those issues most of which do relate directly here and none of them have reverted to their broken state. We only need the removeFromSuperview to run on the main thread the rest it doesn't matter if they are running on the kroll or mainThread. The reason this fixes the current issue is because once a js object is out the scope in this case the for loop it will get garbage collected as you know. We create a bridge between the objective C object and the js so it doesn't get retained however when all of this was running on the main-thread while we're running the application on the krill-thread, the kroll thread would have finished executing first. Collecting the JS object before the main-thread had a chance to tell the program to retain it.

cheekiatng:
Well. Go ahead to merge and resolve the ticket if you guys think nothing is going to break then.

pec1985:
PR **APPROVED!**

PR #7917, opened 2016-04-05
[TIDOC-2487]: API refrence linking not working correctly

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIDOC-2487)

PR #7927, opened 2016-04-09
Fix deprecation links

FokkeZB:
JIRA: https://jira.appcelerator.org/browse/AC-3476

### DIFF
--- a/apidoc/Titanium/UI/iPhone/iPhone.yml
+++ b/apidoc/Titanium/UI/iPhone/iPhone.yml
@@ -181,14 +181,14 @@ properties:
     type: Number
     deprecated:
         since: "5.4.0"
-        notes: Use [Ti.UI.iOS.appBadge](Titanium.UI.iOS.appBadge) instead.
+        notes: Use Titanium.UI.iOS.appBadge instead.
 
   - name: appSupportsShakeToEdit
     summary: Determines whether the shake to edit system-wide capability is enabled.
     type: Boolean
     deprecated:
         since: "5.4.0"
-        notes: Use [Titanium.UI.iOS.appSupportsShakeToEdit](Titanium.UI.iOS.appSupportsShakeToEdit) instead.
+        notes: Use Titanium.UI.iOS.appSupportsShakeToEdit instead.
 
   - name: statusBarHidden
     summary: Determines whether the status bar is hidden.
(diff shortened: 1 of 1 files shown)


hansemannn:
We need the links. https://github.com/appcelerator/titanium_mobile/pull/7917/files fixes the links like it's done currently. On long-term, we need to adjust the generator to compile `[name](link)` correctly. @AngelkPetkov please check if you cover all links specified here in your PR.

FokkeZB:
The way I fixed it will generate links because it recognizes it as API.

hansemannn:
Did you test the generated html? If that works, please check if you have covered all links Angel added and we merge yours. Thanks! 🚀

FokkeZB:
See http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.iPhone.NavigationGroup where the link works but displays the `< .. >` from the source at https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/iPhone/NavigationGroup.yml#L7.

This is how JSDuck works, just mention the API and it wil link it.

hansemannn:
Ok great. So let's write a regex to index all deprecated API's and their `notes` key to fix possibly broken links. @AngelkPetkov or @FokkeZB anyone interested?

FokkeZB:
And/or fix the tooling. As [jsduck_generator.py](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/generators/jsduck_generator.py#L204) does run `deprecated:notes` through `markdown_to_html`.

jawa9000:
Do we know where to find an alternative image to this one: http://img.skitch.com/20100406-rwe44533tkd94fdnbqqudxynpe.png on http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.iPhone.NavigationGroup

We need to stop using EverNote (Skitch) and only reference images from our wiki.

FokkeZB:
I have no idea. I think only Ben used Skitch for images?

You could just create a NavigationGroup and screenshot it?

hansemannn:
Where are we here? If that one image is the only thing preventing us from merging this, anybody can make one quickly, let me know if you need help / example code. This PR should focus on the deprecation links.

hansemannn:
Approved for now, better than broken links. Please keep an eye on possible API's that might cause problems in the future.

PR #7931, opened 2016-04-11
[TIMOB-23175] Android: Replaced condition for build tools to accept 23.x

ashcoding:
https://jira.appcelerator.org/browse/TIMOB-23175

Warning "Android Build Tools X are too new" should not replace < with <=


cb1kenobi:
**APPROVED**

PR #7938, opened 2016-04-13
[TIMOB-23181, TIMOB-19673] Bump windowslib to 0.4.6 (5_3_X)

infosia:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23181
- TIMOB-23181 Install of app to Windows 10 Mobile device fails on second build of application 
- TIMOB-19673 windowslib: visualstudio.detect failed for VS 2015

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "temp": "0.8.3",
     "uglify-js": "2.5.0",
     "unorm": "^1.4.1",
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
@infosia we not only need to update the package.json but actually include the new module (for titanium_mobile we pre-package dependencies). I'm going to close this and do it...

PR #7956, opened 2016-04-20
[TIMOB-20250] iOS: Allow Hyperloop views to be removed

pec1985:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20250

Provide a clear PR title prefixed with `TIMOB-20250`


hansemannn:
The PR does not work for the example code provided in the ticket, output:
```
[DEBUG] Application booted in 628.706992 ms
[DEBUG] [HYPERLOOP] didStartNewContext <KrollContext: 0x7f88f35a1b10>
[DEBUG] 2016-04-21 09:38:16.909 Hyperloop_Sample[4256:43114] [ERROR] Invalid type passed to function. expected: TiViewProxy, was: HyperloopPointer -[TiViewProxy remove:] (TiViewProxy.m:319)
[TRACE] [Hyperloop_Sample] [ERROR] Invalid type passed to function. expected: TiViewProxy, was: HyperloopPointer -[TiViewProxy remove:] (TiViewProxy.m:319)
```

iphone/Classes/TiViewProxy.m
```diff
@@ -285,9 +290,33 @@ -(void)replaceAt:(id)args
...
 
+	
+#ifdef HYPERLOOP
+	// ofbuscate this selector
+	static SEL nativeObjSel = nil;
+	if (nativeObjSel == nil) nativeObjSel = NSSelectorFromString([NSString stringWithFormat:@"%@%@", @"native", @"Object"]);
+	// obfuscate this class name
+	static Class hlClassName = nil;
+	if (hlClassName == nil) hlClassName = NSClassFromString([NSString stringWithFormat:@"%@%@", @"Hyperloop",@"Class"]);
+	if ([arg isKindOfClass:[UIView class]] || (hlClassName != nil && [arg isKindOfClass:hlClassName])) {
```
hansemannn: Shouldn't we use `HyperloopPointer` here? And the `hlClassName` can't be nil, because the static variable is set in 302 if it wasn't not before. So the `hlClassName != nil` can be removed.

pec1985:
Fixed

hansemannn:
Works! One thing is noticed is, that if you remove a view and then try to remove it again, the above warning is thrown again. That behaves different then when removing a normal TiUIView. Can we achieve a similar behavior here?

hansemannn:
Approved, thank you Pedro! 🚀

PR #7957, opened 2016-04-21
[TIMOB-23145] Enable run-on-main-thread by default for new projects

cheekiatng:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23145


hansemannn:
As we are talking about default values. Do we enabled autolayout in 5.4.0 now or do we need more tim?

cheekiatng:
i think we need more time. Is auto layout necessary to be set on for hyperloop?

hansemannn:
Nope, and currently produces problems in that combination. Let's do further discussions in JIRA.

sgtcoolguy:
LGTM

PR #7961, opened 2016-04-22
[TIMOB-20451]: Window.close should use new activity*Transition instead of activity*Animation or default

sriks:
JIRA: https://jira.appcelerator.org/browse/TIMOB-20451


ashcoding:
Code reviewed and functionally tested.
**Accepted**

PR #7966, opened 2016-04-26
[TIMOB 16879] iOS/Android: ScrollableView: Add "insertViewsAt()"

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16879


AngelkPetkov:
FT and CR passed. **Approved**!

PR #7969, opened 2016-04-27
[TIMOB-23280] iOS: Use updated APS SDK with TIMOB

sriks:
JIRA: https://jira.appcelerator.org/browse/TIMOB-23280


cheekiatng:
APPROVED!

PR #7970, opened 2016-04-27
[TIMOB-23281] iOS: Set BackgroundColor to the search result list view

TiBaharroth:
JIRA: https://jira.appcelerator.org/browse/TIMOB-23281

Optional Description:

Fix bug : 
When the search result tableview appears, it has a white background by default. Now the search result table view has the same color of his tableview by default.

Add two properties :
-searchResultBackgroundColor To set the backgroundColor of the search result.
-searchResultSeparatorColor To set the separatorColor of the search result.

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -1936,6 +1936,16 @@ static TiViewProxy * FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoin
     [searchButton setTitle:[TiUtils stringValue:searchButtonTitle]];
 }
 
+- (BOOL) searchDisplayController:(UISearchDisplayController *)controller shouldReloadTableForSearchString:(NSString *)searchString
+{
+    TiColor *resultBgColor = [TiUtils colorValue:[self.proxy valueForKey:@"searchResultBackgroundColor"]];
+    
+    resultBgColor? [[self class] setBackgroundColor:resultBgColor onTable:controller.searchResultsTableView]: [[self class] setBackgroundColorWithUIColor:_tableView.backgroundColor onTable:controller.searchResultsTableView];
+    [controller.searchResultsTableView setSeparatorColor:[[TiUtils colorValue:[self.proxy valueForKey:@"searchResultSeparatorColor"]] _color]?:_tableView.separatorColor];
+    
+    return YES;
+}
+
 - (void)searchDisplayControllerDidEndSearch:(UISearchDisplayController *)controller
 {
     self.searchString = @"";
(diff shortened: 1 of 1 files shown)


hansemannn:
Before we review, some questions:
- Wouldn't it be better to apply the exact item style / template to the search results instead?
- Does iOS natively use the same style / template in the results?
- What's about something like backgroundImage and selectionStyle, how would we handle that?

We also need to ensure it does not break or changes the current default behavior.

TiBaharroth:
You can add some questions as :
- If we set the background so we have to set the color of the noResult message
- So we have to expose the color, and why not, the content of the noResult message
- In case of no result, can we expose the backgroundView to let developer set a custom view?

hansemannn:
Yeah, thats another styling improvement. We should file a different ticket for that one (if not already existing). 

Regarding your PR: I would come up with a different solution, what do you think about the following, it adds support for background-colors, separator-colors, separator-insets and separator-styles:

```objc

-(BOOL)searchDisplayController:(UISearchDisplayController *)controller shouldReloadTableForSearchString:(NSString *)searchString
{
    TiColor *resultsBackgroundColor = [TiUtils colorValue:[[self proxy] valueForKey:@"resultsBackgroundColor"]];
    TiColor * resultsSeparatorColor = [TiUtils colorValue:[[self proxy] valueForKey:@"resultsSeparatorColor"]];
    id resultsSeparatorInsets = [[self proxy] valueForKey:@"resultsSeparatorInsets"];
    id resultsSeparatorStyle = [[self proxy] valueForKey:@"resultsSeparatorStyle"];
    
    ENSURE_TYPE_OR_NIL(resultsSeparatorInsets, NSDictionary);
    ENSURE_TYPE_OR_NIL(resultsSeparatorStyle, NSNumber);
    
    if (resultsBackgroundColor) {
        [controller.searchResultsTableView setBackgroundColor:[resultsBackgroundColor _color]];
    }
    
    if (resultsBackgroundColor) {
        [controller.searchResultsTableView setSeparatorColor:[resultsSeparatorColor _color]];
    }
    
    if (resultsSeparatorInsets) {
        [controller.searchResultsTableView setSeparatorInset:[TiUtils contentInsets:resultsSeparatorInsets]];
    }
    
    if (resultsSeparatorStyle) {
        [controller.searchResultsTableView setSeparatorStyle:[TiUtils intValue:resultsSeparatorInsets def:UITableViewCellSeparatorStyleSingleLine]];
    }
    
    return YES;
}
```

TiBaharroth:
I'am agree with your last PR. 
I wanted write this but i wasn't sure about the use of '[self proxy] valueForKey' was not heavy.
And i was limited by my obj-c knowledge and TI sdk :)

Thanks for your help

hansemannn:
Are you OK when I close this PR and open a new one containing the changes? I'm still thankful for you digging out that issue! 😊

TiBaharroth:
Yes of course, close this PR and open a new one with changes :)

hansemannn:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/8006

PR #7971, opened 2016-04-27
[TIMOB-16471] iOS: Add Ti.UI.SearchBar.style property

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16471


AngelkPetkov:
CR and FT looks good! **Approved**

PR #7974, opened 2016-04-27
[TIMOB-16102] iOS: AudioPlayer progress event not firing

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-16102)

### DIFF
--- a/iphone/Classes/TiMediaAudioPlayerProxy.m
+++ b/iphone/Classes/TiMediaAudioPlayerProxy.m
@@ -50,7 +50,9 @@
 {
 	if (count == 1 && [type isEqualToString:@"progress"])
 	{
-		timer = [[NSTimer scheduledTimerWithTimeInterval:1 target:self selector:@selector(updateProgress:) userInfo:nil repeats:YES] retain];
+		if (![timer isValid]) {
+			timer = [[NSTimer scheduledTimerWithTimeInterval:1 target:self selector:@selector(updateProgress:) userInfo:nil repeats:YES] retain];
+		}
 	}
 }
 
--- a/iphone/Classes/TiMediaAudioPlayerProxy.m
+++ b/iphone/Classes/TiMediaAudioPlayerProxy.m
@@ -50,7 +50,7 @@
 {
 	if (count == 1 && [type isEqualToString:@"progress"])
 	{
-		progress = YES;
+		timer = [[NSTimer scheduledTimerWithTimeInterval:1 target:self selector:@selector(updateProgress:) userInfo:nil repeats:YES] retain];
 	}
 }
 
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -50,15 +50,19 @@ -(void)_listenerAdded:(NSString *)type count:(int)count
...
 -(void)_listenerRemoved:(NSString *)type count:(int)count
 {
 	if (count == 0 && [type isEqualToString:@"progress"])
 	{
-		progress = NO;
+        if (timer!=nil)
+        {
+            [timer invalidate];
+            RELEASE_TO_NIL(timer);
+        }
```
hansemannn: Indentation

iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -50,15 +50,19 @@ -(void)_listenerAdded:(NSString *)type count:(int)count
 {
 	if (count == 1 && [type isEqualToString:@"progress"])
 	{
-		progress = YES;
+		timer = [[NSTimer scheduledTimerWithTimeInterval:1 target:self selector:@selector(updateProgress:) userInfo:nil repeats:YES] retain];
```
hansemannn: This will cause problems when the user creates different `progress` event listener. That's why it was in the initial creation only. Please test different event listeners and try to come up with a solution that can handle different event listeners as well (like before).

AngelkPetkov:
Added Validation so the timer is only initialized once, this will allow you to have multiple progress events like so : 
Titanium.UI.setBackgroundColor('#fff');
	 
	var audioPlayer = Ti.Media.createAudioPlayer({
		url : '/Users/Angel/Music/iTunes/iTunes\ Media/Music/Trap\ Nation/Unknown\ Album/SNBRN\ -\ Gangsta\ Walk\ \(feat.\ Nate\ Dogg\)\ \(JayKode\ Remix\).mp3' //Add local or remote file
	});
	 
	   title: 'Start/Stop',
	   top: 10,
	   width: 100,
	   height: 50
	});
	var playing = false;
	button.addEventListener('click',function(e)
	{
		if (playing != true) { 
	   		audioPlayer.start();
	   	} else {
	   		audioPlayer.pause();
	   	}
	   	playing = !playing;
	});
	 
	 
	audioPlayer.start();
	 
	 
	audioPlayer.addEventListener('progress', function(a) {
		Ti.API.info("Cool beans");
	});


	audioPlayer.addEventListener('progress', function(e) {
		Ti.API.info("Hello Mundo");
		// alert('EAC LOGGING: Progress: ' + e.progress);
	});

	 
	win.add(button);
	win.open();

hansemannn:
PR rejected. It only doesn't work in your example because the event listener is placed incorrectly. Placing it before starting the player fixes the issue (which is none, because that's the common way to implement the observer pattern). More info inside the ticket.

PR #7975, opened 2016-04-27
TIMOB-20469 fix to allow us to hook into the toolchain compile before/after

jhaynie:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20469

**Optional Description:**

This change adds 2 new hook events that are emitted just before and after the native toolchain build which is required for hyperloop to intercept the build in the right location.


hansemannn:
Discussed and approved!

PR #7976, opened 2016-04-27
[TIMOB-9111] iOS: base64encode adding a new line too early

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-9111)

### DIFF
--- a/iphone/Classes/Base64Transcoder.c
+++ b/iphone/Classes/Base64Transcoder.c
@@ -129,7 +129,7 @@ bool Base64EncodeData(const void *inInputData, size_t inInputDataSize, char *out
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex] & kBits_00000011) << 4 | (theInPtr[theInIndex + 1] & kBits_11110000) >> 4];
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex + 1] & kBits_00001111) << 2 | (theInPtr[theInIndex + 2] & kBits_11000000) >> 6];
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex + 2] & kBits_00111111) >> 0];
-		if (theOutIndex % 74 == 72)
+		if (theOutIndex % 76 == 74)
 		{
 			outOutputData[theOutIndex++] = '\r';
 			outOutputData[theOutIndex++] = '\n';
@@ -142,7 +142,7 @@ bool Base64EncodeData(const void *inInputData, size_t inInputDataSize, char *out
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex] & kBits_00000011) << 4 | (0 & kBits_11110000) >> 4];
 		outOutputData[theOutIndex++] = '=';
 		outOutputData[theOutIndex++] = '=';
-		if (theOutIndex % 74 == 72)
+		if (theOutIndex % 76 == 74)
 		{
 			outOutputData[theOutIndex++] = '\r';
 			outOutputData[theOutIndex++] = '\n';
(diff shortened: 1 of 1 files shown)


iphone/Classes/Base64Transcoder.c
```diff
@@ -129,7 +129,7 @@ bool Base64EncodeData(const void *inInputData, size_t inInputDataSize, char *out
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex] & kBits_00000011) << 4 | (theInPtr[theInIndex + 1] & kBits_11110000) >> 4];
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex + 1] & kBits_00001111) << 2 | (theInPtr[theInIndex + 2] & kBits_11000000) >> 6];
 		outOutputData[theOutIndex++] = kBase64EncodeTable[(theInPtr[theInIndex + 2] & kBits_00111111) >> 0];
-		if (theOutIndex % 74 == 72)
+		if (theOutIndex % 76 == 74)
```
hansemannn: Where do these values come from? Reference?

hansemannn:
Another issue: We need unit tests to compare the generated hashed. Base64-encoding is very critical to be correct in applications, so it needs to be ensured to work properly.

The following test does not work:

```
var win = Ti.UI.createWindow({backgroundColor:'white'});
var shortString = 'ABCDEFGHIJ1234567890ABCDEFGHIJ12|psndemo2|abcd:1';
var longString  = 'ABCDEFGHIJ1234567890ABCDEFGHIJ12|psndemo2|abcd:12345678901234567890';
var tiBase64ShortResult = Ti.Utils.base64encode(shortString);
var tiBase64LongResult  = Ti.Utils.base64encode(longString);
 
tiBase64ShortResult = String(tiBase64ShortResult);
tiBase64LongResult = String(tiBase64LongResult);
 
var testShort = "QUJDREVGR0hJSjEyMzQ1Njc4OTBBQkNERUZHSElKMTJ8cHNuZGVtbzJ8YWJjZDox";
var testLong = "QUJDREVGR0hJSjEyMzQ1Njc4OTBBQkNERUZHSElKMTJ8cHNuZGVtbzJ8YWJjZDoxMjM0NTY3ODkwMTIzNDU2Nzg5MA==";

if (testShort != tiBase64ShortResult) {
	Ti.API.error("tiBase64ShortResult incorrect!!");
} else {
	Ti.API.warn("tiBase64ShortResult correct")
}

if (tiBase64LongResult != testLong) {
	Ti.API.error("tiBase64LongResult incorrect!!");
} else {
	Ti.API.warn("tiBase64LongResult correct")
}

var label2 = Ti.UI.createLabel({
  color:'blue',
  text: 'long: ' + tiBase64LongResult,
  textAlign: Ti.UI.TEXT_ALIGNMENT_LEFT,
  top: 30,
  width: 300, height: 200
});
win.add(label2);
win.open()
```

The `testLong` fails.

AngelkPetkov:
If you run the same tests without my changes the "tiBase64LongResult incorrect!" is still appearing, the changes i made is simply  where the line break is added it shouldn't effect the actual encryption

AngelkPetkov:
@hansemannn  The text is identical if you use this [online tool](http://text-compare.com) to compare the outputted text and the one you converted using an online converter. Not sure why it says the text is different.

iphone/Classes/Base64Transcoder.c
sgtcoolguy: I'm going to go out on a limb here and assume the number 76 came from this: https://lists.w3.org/Archives/Public/w3c-ietf-xmldsig/2001AprJun/0183.html

Sounds like the spec requires lines no longer than 76 characters for Base64 encoding?

Looking further, that is *sometimes* true: 
- https://tools.ietf.org/html/rfc4648#page-3
- https://en.wikipedia.org/wiki/Base64

Given that we don't match between Android and iOS/Windows, we should come up with what we think is a reasonable default imply regard line newlines/line length. Windows does what iOS does right now, BTW.

I'd argue that our base64 encoder method should actually take an options argument to control this behavior. Sort of like Android's API does: https://developer.android.com/reference/android/util/Base64.html

hansemannn:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/8058.

PR #7988, opened 2016-05-05
[TIMOB] Updating Packaged Android Module

AngelkPetkov:


hansemannn:
We need both iOS and Android for the pre-packaged modules. Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/7991

PR #7992, opened 2016-05-06
[TIMOB-23341] Exception on iPhone: no row found for index. 

TiBaharroth:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23341

Fix the bug : insert in the insertRowBefore and insertRowAfter

``` objc
if (![NSThread isMainThread]) {
        TiThreadPerformOnMainThread(^{[self insertRowBefore:args];},YES); // YES = wait for the operation to complete on the main thread !!!
        return;
}
```

Explaination : 
'The problem here is actually very simple, but it took me few hours to find this out. JS engine itself runs in async context, which means that if one is inserting multiple rows in JavaScript (e.g. in a loop) using insertRowBefore then the obj-c code behind this does the work in async (non-UI) thread. I think the reason the ENSURE_UI_THREAD(insertRowBefore,args); line was commented is becase the actual action happens in UI thread anyway - see how [table dispatchAction:action] works (TiUITableViewProxy.m around line 535).

The code that checks the row indexes and verifies against existing list still runs asynchronously and thus if you run inserting rows in loop this can have unpredictable results because the list might not yet be updated, etc.. So the solution here is to synchronize the insertion and wait for it’s completion. Finally the commented code mentioned by Dave Cadwallader should be replaced with this:'

From https://archive.appcelerator.com/question/129708/obj-c-fix-for-error-on-tableview-insertrowbefore--insertrowafter

### DIFF
--- a/iphone/Classes/TiUITableViewProxy.m
+++ b/iphone/Classes/TiUITableViewProxy.m
@@ -502,6 +502,11 @@ USE_VIEW_FOR_CONTENT_HEIGHT
 {
 //	ENSURE_UI_THREAD(insertRowBefore,args);
 	
+	if (![NSThread isMainThread]) {
+        TiThreadPerformOnMainThread(^{[self insertRowBefore:args];},YES); // YES = wait for the operation to complete on the main thread !!!
+        return;
+}
+	
 	int index = [TiUtils intValue:[args objectAtIndex:0]];
 	NSDictionary *data = [args objectAtIndex:1];
 	NSDictionary *anim = [args count] > 2 ? [args objectAtIndex:2] : nil;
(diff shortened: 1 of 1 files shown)


hansemannn:
Should be obsolete in 5.4.0 and greater, since we use the main thread execution by default. Can you test the issue with 5.4.0 and verify it works without the fix?

cheekiatng:
in 5.4.0, although the main thread is set by default for new projects, devs can still enable it on/off, so this error must not happen in both scenarios. (main thread on / off)

TiBaharroth:
Sorry, i didn't try with SDK 5.4.0, i will try asap but i think it will fix with this version

hansemannn:
Thank you @TiBaharroth, looking forward to that. We will schedule it for 6.0.0.

hansemannn:
@TiBaharroth Can you confirm that it works on 5.4.0 and 6.0.0? I would appreciate!

TiBaharroth:
Can't reproduce the issue with 5.4.0

So i think it's OK :)

hansemannn:
Thanks! Closing for now.

PR #7995, opened 2016-05-10
[TIMOB-19253] iOS: Clicking a list item while editing the list does not fire itemclick event

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-19253)


hansemannn:
Approved!

PR #8002, opened 2016-05-17
[TIMOB-23186] Added improvement to get an event

ashcoding:
JIRA: https://jira.appcelerator.org/browse/TIMOB-23186

Test case:-

```
    backgroundColor : 'black',
    layout : 'vertical'
});

var webview = Titanium.UI.createWebView({
    blacklistedURLs : [ 'https://' ],
    url : 'https://en.m.wikipedia.org/wiki/Main_Page'
});

webview.addEventListener('load', function(e) {
        // This should not be called for blacklistedUrls
    Ti.API.info("load = e.url=" + e.url);
});

webview.addEventListener('onStopBlacklistedUrl', function(e) {
        Ti.API.info("This URL was stopped: " + e.url);
        var result = Ti.Platform.openURL(e.url);
});

win.add(webview);
win.open();
```

### DIFF
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -388,6 +388,16 @@ events:
       - name: code
         summary: SSL error code.
         type: Number
+
+  - name: onStopBlacklistedUrl
+    summary: Fired when a blacklisted Url is stopped
+    description: |
+        This event is fired when a blacklisted Url is stopped
+    platforms: [android]
+    since: "5.4.0"
+    properties:
+      - name: url
+        summary: URL of the web document that is stopped.
         
 properties:
 
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -424,7 +424,8 @@ properties:
         An array of url strings to blacklist.
     description: |
         An array of url strings to blacklist. This will stop the webview from going to urls listed in
-        the blacklist.
+        the blacklist. Note, this only applies in the links clicked inside the webview. The first website
+        that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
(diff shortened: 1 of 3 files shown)


ashcoding:
Blacklisted urls can also be the protocol used eg "http://" or "https://".

This will allow you to open them in an external browser that is not handled by the Titanium App.

hieupham007:
Tested with Android 6.0, nexus 6p, website still loaded for me.

ashcoding:
The website that it goes to should load. What shouldn't load is the links.

hieupham007:
I'm seeing this only: TiAPI   :  load = e.url=https://en.m.wikipedia.org/wiki/Main_Page
Not the other event that should've fired.

ashcoding:
The example is not the best example. In this scenario, the main page https://en.m.wikipedia.org/wiki/Main_Page would load. Then clicking any other links on the page which are https wont load and the event will fire to open an external browser.

ashcoding:
Let me know if it doesn't work that way.

hieupham007:
Ok, that works. The docs need to be modified because if blacklisted urls has "https" and url has "https" it should be blocked. If it doesn't we need to document it.

ashcoding:
Ok. I'll update the docs.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -388,6 +388,16 @@ events:
       - name: code
         summary: SSL error code.
         type: Number
+
+  - name: onStopBlacklistedUrl
```
hansemannn: Don't like the `on` prefix for events. On Alloy, it produces something like `<WebView onOnstopblacklistedurl="" />`. Can we migrate this and `onLoadResource ` to `stopblacklistedurl` and `loadresource` ?

PR #8005, opened 2016-05-18
[TIMOB-23406] Adding 'scrolling' event to ListView to track scrolling…

rlustemberg:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23406

At the moment is not possible to determine the scroll direction of a ListView natively and as the view is scrolling. Very common effects such as the hiding of the navigation bar when scrolling (like on the Facebook app) become impossible to achieve without cumbersome workarounds.
I will send a PR with a modification to the SDK core.

### DIFF
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -598,29 +598,30 @@ events:
         summary: false. This event does not bubble.
         type: Boolean
 
-    - name: scrolling
-	    summary: Fires when the list view is scrolling. Calling the `scrollTo` methods will not fire this event
-	    description: |
-	        This event will fire when the list view is scrolling. Use  `direction` to determine the scroll direction, 
-	         `velocity` to determine scroll speed, targetContentOffset to determine where the scrolling will end (in dpi)
-	    platforms: [iphone, ipad]
-	    since: 5.4.0
-	    properties: 
-	      - name: direction
-	        summary: Direction of the scroll--either 'up', or 'down'.
-	        type: String
+  - name: scrolling
+    summary: Fires when the list view is scrolling. Calling the `scrollTo` methods will not fire this event
+    description: |
+        This event will fire when the list view is scrolling. Use  `direction` to determine the scroll direction,
+        `velocity` to determine scroll speed, targetContentOffset to determine where the scrolling will end (in dpi)
+    platforms: [iphone, ipad]
+    since: 5.4.0
+    properties:
+      - name: direction
+        summary: Direction of the scroll--either 'up', or 'down'.
+        type: String
 
-	      - name: velocity
-	        summary: The velocity of the scroll in scale factor per second
-	        type: number
+      - name: velocity
+        summary: The velocity of the scroll in scale factor per second
+        type: number
 
-	      - name: targetContentOffset
-	        summary: The expected offset when the scrolling action decelerates to a stop.
-	        type: Object
+      - name: targetContentOffset
+        summary: The expected offset when the scrolling action decelerates to a stop.
+        type: Object
 
-	      - name: bubbles
-	        summary: false. This event does not bubble.
-	        type: Boolean
+      - name: bubbles
+        summary: false. This event does not bubble.
+        type: Boolean
+
 
 properties:
   - name: allowsSelection
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -601,13 +601,14 @@ events:
   - name: scrolling
     summary: Fires when the list view is scrolling. Calling the `scrollTo` methods will not fire this event
     description: |
-        This event will fire when the list view is scrolling. Use  `direction` to determine the scroll direction,
-        `velocity` to determine scroll speed, targetContentOffset to determine where the scrolling will end (in dpi)
+        This event will fire while the list view is scrolling and the user releases the finger. Use  `direction` to determine the scroll direction,
+        `velocity` to determine scroll speed, targetContentOffset.y to determine where the scrolling will end (in dpi).The key targetContentOffset.x is always 0.
+         When `direction` is `down`, `velocity` is negative and viceversa
     platforms: [iphone, ipad]
     since: 5.4.0
     properties:
       - name: direction
-        summary: Direction of the scroll--either 'up', or 'down'.
+        summary: Direction of the scroll either 'up', or 'down'.
         type: String
 
       - name: velocity
@@ -615,7 +616,9 @@ events:
         type: number
 
       - name: targetContentOffset
-        summary: The expected offset as an object consisting of x and y keys when the scrolling action decelerates to a stop. 
+        summary: |
+           The expected offset as an object consisting of x and y keys when
+           the scrolling action decelerates to a stop. 
         type: Object
 
 
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -598,6 +598,30 @@ events:
         summary: false. This event does not bubble.
         type: Boolean
 
+    - name: scrolling
+	    summary: Fires when the list view is scrolling. Calling the `scrollTo` methods will not fire this event
+	    description: |
+	        This event will fire when the list view is scrolling. Use  `direction` to determine the scroll direction, 
+	         `velocity` to determine scroll speed, targetContentOffset to determine where the scrolling will end (in dpi)
+	    platforms: [iphone, ipad]
+	    since: 5.4.0
+	    properties: 
+	      - name: direction
+	        summary: Direction of the scroll--either 'up', or 'down'.
+	        type: String
+
+	      - name: velocity
+	        summary: The velocity of the scroll in scale factor per second
+	        type: number
+
+	      - name: targetContentOffset
+	        summary: The expected offset when the scrolling action decelerates to a stop.
+	        type: Object
+
+	      - name: bubbles
+	        summary: false. This event does not bubble.
+	        type: Boolean
+
 properties:
   - name: allowsSelection
     summary: Determines whether this item can be selected.
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -615,12 +615,12 @@ events:
 
       - name: velocity
         summary: The velocity of the scroll in scale factor per second
-        type: number
+        type: Number
 
       - name: targetContentOffset
         summary: |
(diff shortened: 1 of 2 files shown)


nuno:
+1

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,24 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
         [self.proxy fireEvent:@"dragstart" withObject:nil withSource:self.proxy propagate:NO reportSuccess:NO errorCode:0 message:nil];
     }
 }
-
+-(void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset{
+    
+    if ([self.proxy _hasListeners:@"scrolling"]) {
+        NSMutableDictionary *eventArgs = [NSMutableDictionary dictionary];
+        NSString* swipeString = @"none";
+        if (velocity.y > 0){
```
hansemannn: Missing space in `) {`.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,24 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+            swipeString = @"up";
+        }
+        if (velocity.y < 0){
```
hansemannn: Missing space in `) {`.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,24 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+            swipeString = @"down";
+        }
+        if(swipeString !=(id)[NSNull null])
```
hansemannn: It can never be null, because you set the default value to `none`. So please leave the default value to nil and only check `if (swipeString != nil) {`. Also note that the bracket goes into the same line.

hansemannn:
Left some comments. Also please add docs for the new event.

iphone/Classes/TiUIListView.m
rlustemberg: Yes, I'll take it out. It's a leftover from when I was not setting a default (none)

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,24 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
         [self.proxy fireEvent:@"dragstart" withObject:nil withSource:self.proxy propagate:NO reportSuccess:NO errorCode:0 message:nil];
     }
 }
-
+-(void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset{
```
hansemannn: Think about including the velocity and contentOffset as well, so the developer can receive as much information as possible.

rlustemberg: Are the comments for me or for you guys? ;)

hansemannn: For you, since you are the contributor and we are the reviewer 😊

rlustemberg: LOL.
Will do my best to get this ready before Monday. Greetings from Amsterdam, Hans!

iphone/Classes/TiUIListView.m
rlustemberg: If I initialise the swipeString to nil and do a nil check before firing the event, eventually I get a bad access exception. I left if initialised as 'none' and I do if (swipeString != @"none") instead. In that way it doesn't crash.

rlustemberg:
I've added velocity (y axis) and the full x,y coordinates for targetContentOffset as I couldn't make up my mind of a name for only the y point (x is always 0).
I've renamed swipeString to directionString and kept it initialised because for some reason unknown to me , initialised as nil and doing a nil check before firing the event will still through an bad access exception.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,31 @@ events:
...
+
+      - name: velocity
+        summary: The velocity of the scroll in scale factor per second
+        type: number
+
+      - name: targetContentOffset
+        summary: The expected offset when the scrolling action decelerates to a stop.
+        type: Object
+
+      - name: bubbles
```
hansemannn: If it doesn't bubble, remove the property.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,31 @@ events:
...
+      - name: direction
+        summary: Direction of the scroll--either 'up', or 'down'.
+        type: String
+
+      - name: velocity
+        summary: The velocity of the scroll in scale factor per second
+        type: number
+
+      - name: targetContentOffset
+        summary: The expected offset when the scrolling action decelerates to a stop.
```
hansemannn: Mention that the object consists of an `x` and `y` key.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,32 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+
+        if (velocity.y > 0) {
+            directionString = @"up";
+        }
 
+        if (velocity.y < 0) {
+            directionString = @"down";
+        }
+        
+        if(directionString != @"none") {
```
hansemannn: This is bad code-style, we need to do a proper-nil check here.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,32 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+            NSMutableDictionary *eventArgs = [NSMutableDictionary dictionary];
+            NSDictionary *pointObject = [NSDictionary dictionaryWithObjectsAndKeys:
+             [NSNumber numberWithDouble:targetContentOffset->x],@"x",
+             [NSNumber numberWithDouble:targetContentOffset->y],@"y",
```
hansemannn: Don't access the properties like this. I will provide a gist on a proper usage.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,32 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+                                         nil];
+            [eventArgs setValue:directionString forKey:@"direction"];
+            [eventArgs setValue:NUMDOUBLE(velocity.y) forKey:@"velocity"];
+            [eventArgs setObject:pointObject forKey:@"targetContentOffset"];
+            [self.proxy fireEvent:@"scrolling" withObject:eventArgs withSource:self.proxy propagate:NO reportSuccess:NO errorCode:0 message:nil];
```
hansemannn: You need to check if the event-listener is added beforehand. Will also mention it in my gist.

iphone/Classes/TiUIListView.m
rlustemberg: I know, but just checking for nil was still not working. Please remember I'm no Objective C dev

hansemannn:
Here is what I would do: https://gist.github.com/hansemannn/604ad5060bd69710d0326580ebf9ac90

Also what I noticed:
- The direction is always then false, when the user drags without releasing the finger. Also the velocity is0 then.
- When the direction is `down`, the velocity is negative.
- The `targetContentOffset` key `x` is always 0 (of course, since we can only scroll vertically).

That should also be added to the docs to clarify. Please test the gist and update your code and docs if it makes sense to you.

iphone/Classes/TiUIListView.m
```diff
@@ -1815,7 +1815,35 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
...
+            direction = @"down";
+        }
+        
+        NSMutableDictionary *event = [NSMutableDictionary dictionaryWithDictionary:@{
+                                                                                     @"targetContentOffset": @{
+                                                                                             @"x": NUMFLOAT(targetContentOffset->x),
+                                                                                             @"y": NUMFLOAT(targetContentOffset->y),
+                                                                                             },
+                                                                                     @"velocity": NUMFLOAT(velocity.y)
+                                                                                     }];
```
hansemannn: The copy-and-paste caused formatting issues here, please refer to the gist and format manually.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,27 @@ events:
...
+  - name: scrolling
+    summary: Fires when the list view is scrolling. Calling the `scrollTo` methods will not fire this event
+    description: |
+        This event will fire when the list view is scrolling. Use  `direction` to determine the scroll direction,
+        `velocity` to determine scroll speed, targetContentOffset to determine where the scrolling will end (in dpi)
+    platforms: [iphone, ipad]
+    since: 5.4.0
+    properties:
+      - name: direction
+        summary: Direction of the scroll--either 'up', or 'down'.
```
hansemannn: Change to: "Direction of the scroll, either `up`, or `down`."

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,27 @@ events:
...
+        type: String
+
+      - name: velocity
+        summary: The velocity of the scroll in scale factor per second
+        type: number
+
+      - name: targetContentOffset
+        summary: The expected offset as an object consisting of x and y keys when the scrolling action decelerates to a stop.
```
hansemannn: Break this line to
```yml
summary: |
    The expected offset as an object consisting of x and y keys when 
    the scrolling action decelerates to a stop. 
```

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,31 @@ events:
...
+    platforms: [iphone, ipad]
+    since: 5.4.0
+    properties:
+      - name: direction
+        summary: Direction of the scroll either 'up', or 'down'.
+        type: String
+
+      - name: velocity
+        summary: The velocity of the scroll in scale factor per second
+        type: number
```
hansemannn: This has to be `Number`, otherwise the doc validation fails.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -598,6 +598,31 @@ events:
...
+
+      - name: targetContentOffset
+        summary: |
+           The expected y axis offset when the scrolling action decelerates to a stop. 
+        type: number
```
hansemannn: Same here: This has to be `Number`, otherwise the doc validation fails.

hansemannn:
Approved!

PR #8008, opened 2016-05-20
Wrap autolayout-initializer in proper #ifdef

hansemannn:
There was a missing `#ifdef` statement inside the `TiUIView` which is required to resolve certain selector-warnings.


hansemannn:
Approved!

PR #8010, opened 2016-05-21
[TIMOB-23423] Android: Add rotation, rotationX, rotationY, scaleX, scaleY to view

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23423

**Optional Description:**

Allow to set rotation (x,y,z) and scaling (x,y) to a view without the need of 2dmatrix. Similar to the existing translationX(), translationY() (with a small bugfix allowing to use views with borders, too).

With these settings it is possible to do smooth rotation/scaling in a drag event.
GIF comparing 2dmatrix vs setters: http://migaweb.de/matrix.gif

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -407,6 +407,12 @@ public abstract class TiViewProxy extends KrollProxy implements Handler.Callback
 		setPropertyAndFire(TiC.PROPERTY_HEIGHT, height);
 	}
 	
+	/*@Kroll.setProperty(retain=false) @Kroll.method
+	public void setRotation(Float rotation)
+	{
+		setPropertyAndFire(TiC.PROPERTY_ROTATION, rotation);
+	}
+	
 	@Kroll.setProperty(retain=false) @Kroll.method
 	public void setRotationX(Float rotation)
 	{
@@ -429,7 +435,7 @@ public abstract class TiViewProxy extends KrollProxy implements Handler.Callback
 	public void setScaleY(Float scale)
 	{
 		setPropertyAndFire(TiC.PROPERTY_SCALE_Y, scale);
-	}
+	}*/
 
 	@Kroll.getProperty @Kroll.method
 	public Object getCenter()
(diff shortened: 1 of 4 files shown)


android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -406,6 +406,36 @@ public void setHeight(Object height)
...
+	public void setScaleX(Float scale)
+	{
+		setPropertyAndFire(TiC.PROPERTY_SCALE_X, scale);
+	}
+	
+	@Kroll.setProperty(retain=false) @Kroll.method
+	public void setScaleY(Float scale)
+	{
+		setPropertyAndFire(TiC.PROPERTY_SCALE_Y, scale);
+	}*/
```
hansemannn: Commented out? Please either uncomment or remove it.

m1ga: of course! Forgot to push the commit after testing it :)

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -895,21 +895,41 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
 				ViewCompat.setElevation(nativeView, TiConvert.toFloat(newValue));
 			}
 		} else if (key.equals(TiC.PROPERTY_TRANSLATION_X)) {
-			if (nativeView != null) {
-				ViewCompat.setTranslationX(nativeView, TiConvert.toFloat(newValue));
+			if (getOuterView() != null) {
```
hansemannn: Why exactly do we change the `nativeView` in various properties to 'getOuterView()', but leave properties like `elevation` to use the "old" way?

m1ga: In my PR I just tested the transformation and didn't test the elevation part. But in my test I found out that nativeView was moving the view inside the borderView complex. I found the function https://github.com/m1ga/titanium_mobile/blob/8d5e2d9e28fdd7c57b097e240d8a76e92906d258/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java#L1430 that will take the borderView or nativeView so I thought it would be good to use in my cases so it'll use the proper view. I can test the other cases too with a border view

hieupham007:
Code reviewed and functionally tested. Looks good, will merge when master becomes 6.0.0

PR #8011, opened 2016-05-22
[TIMOB-20076] Android: Allow elevation with borderRadius/borderWidth

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20076

Changing nativeView to getOutlerView() inside TiUIView and setting a setOutlineProvider in TiBorderWrapperView

image

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
@@ -26,7 +26,7 @@ import android.os.Build;
 import android.view.View;
 import android.widget.FrameLayout;
 import android.view.ViewOutlineProvider;
-import android.graphics.Outline ;
+import android.graphics.Outline;
 
 /**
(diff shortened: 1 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
```diff
@@ -121,6 +123,14 @@ private void updateBorderPath()
 			borderPath.addRect(innerRect, Direction.CCW);
 			borderPath.setFillType(FillType.EVEN_ODD);
 		}
+		// set the outline for the view in order to use elevation
+		ViewOutlineProvider viewOutlineProvider = new ViewOutlineProvider() {
```
hieupham007: ViewOutlineProvider is an API 21+ only, please add an API check here. Also, this won't work on API 20 and below.

hieupham007:
Code reviewed and tested with 6.0 device. Works fine, please address comments for API 20 and below.

hansemannn:
@hieupham007 Can we get this into 5.4.0, since it's no feature?

PR #8012, opened 2016-05-22
iOS: Support the camera as a view

hansemannn:
No JIRA-ticket, yet (is there one?). Just a quick proof of concept for now, supporting the camera as a view instead of instead a modal picker using the native `AVCaptureSession` class. 

**Example**:

``` js

var camera = Ti.UI.iOS.createCameraView({
    width: 200,
    height: 300
});
win.add(camera);

win.open();
```

Docs, tests and more public API's to come sometime.


rlustemberg:
+1

hansemannn:
Closing in favor of https://github.com/k0sukey/TiCameraView, didn't know there is already an implementation.

rlustemberg:
Still it would be great to extend that module to add the zoom controls it actually lacks.

PR #8013, opened 2016-05-23
[TIMOB-12412]: Fix intent className and packageName

hieupham007:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-12412


ashcoding:
Code reviewed. Functionally tested with test code in JIRA ticket.
Also tested for the bug in the other ticket:
```Bug occurs when you start and app and hit the back button. Then going back into the app.``` 
No such bug occurs with this.

**Approved**
Merging. :+1:

PR #8019, opened 2016-05-25
[TIMOB-23327] iOS:  Introducing onStopBlackListUrl

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23327?filter=-1)

### DIFF
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -393,7 +393,7 @@ events:
     summary: Fired when a blacklisted Url is stopped
     description: |
         This event is fired when a blacklisted Url is stopped
-    platforms: [android]
+    platforms: [android,iphone,ipad]
     since: "5.4.0"
     properties:
       - name: url
@@ -428,7 +428,7 @@ properties:
         that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
     since: "5.4.0"
-    platforms: [android]
+    platforms: [android,iphone,ipad]
     availability: creation
 
   - name: data
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -99,6 +99,7 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 	RELEASE_TO_NIL(reloadData);
 	RELEASE_TO_NIL(reloadDataProperties);
 	RELEASE_TO_NIL(lastValidLoad);
+	RELEASE_TO_NIL(blacklistedURLs);
 	[super dealloc];
 }
 
@@ -151,7 +152,7 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 		webview.contentMode = UIViewContentModeRedraw;
 		webview.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
 		[self addSubview:webview];
-
+        
 		BOOL hideLoadIndicator = [TiUtils boolValue:[self.proxy valueForKey:@"hideLoadIndicator"] def:NO];
 		
 		// only show the loading indicator if it's a remote URL and 'hideLoadIndicator' property is not set.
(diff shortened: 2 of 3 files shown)


apidoc/Titanium/UI/WebView.yml
```diff
@@ -452,6 +452,13 @@ properties:
     default: false
     platforms: [iphone, ipad]
 
+  - name: blacklistFirstLoad
+    summary: Determines if the initial webview url will be checked against the blacklist.
+    platforms: [iphone,iphone]
+    type: Boolean
+    default: false
+    since: "5.4.0"
```
hansemannn: As discussed, I don't like that properly. As it also causes parity problems, we should remove it and check every URL that is in the blacklist. If they want to exclude certain domains, they should remove them from the blacklist or set the `blacklistedURLs` after its creation, which should be possible.

iphone/Classes/TiUIWebView.m
```diff
@@ -385,6 +386,38 @@ -(BOOL)canGoForward
 	return [webview canGoForward];
 }
 
+-(void)setBlacklistedURLs_:(id)value
+{
+    if (blacklistedURLs == nil) {
+        blacklistedURLs = [[[NSMutableArray alloc] init] retain];
+    }
```
hansemannn: We need to release and re-create the array every-time we call this function (as discussed), so make sure it gets releases properly and set it again.

iphone/Classes/TiUIWebView.m
```diff
@@ -672,21 +698,33 @@ -(CGFloat)contentWidthForWidth:(CGFloat)value
...
-		NSDictionary *event = newUrl == nil ? nil : [NSDictionary dictionaryWithObjectsAndKeys:[newUrl absoluteString], @"url", NUMINT(navigationType), @"navigationType", nil];
-		[self.proxy fireEvent:@"beforeload" withObject:event];
-	}
-
-	if (navigationType != UIWebViewNavigationTypeOther) {
-		RELEASE_TO_NIL(lastValidLoad);
-	}
-
+    
+    if ([self.proxy _hasListeners:@"beforeload"])
```
hansemannn: Indentation is messed up through the whole new code-block.

iphone/Classes/TiUIWebView.m
```diff
@@ -672,21 +698,33 @@ -(CGFloat)contentWidthForWidth:(CGFloat)value
...
+    if (navigationType != UIWebViewNavigationTypeOther) {
+        RELEASE_TO_NIL(lastValidLoad);
+    }
+    
+    if (blacklistedURLs) {
+        for (NSString*check in blacklistedURLs)
+        {
+            if ([[newUrl absoluteString] rangeOfString:check options:NSCaseInsensitiveSearch].location != NSNotFound) {
+                if([self.proxy _hasListeners:@"onStopBlacklistedUrl"]) {
+                    NSDictionary* evt = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:request], @"url", @"Webview did not load blacklisted url.", @"messsage", nil];
```
hansemannn: Did you test that you can cast a `request` of the type `NSURLRequest` to a string?

AngelkPetkov: Yeah i tested it to see if it works, updated the demo-code in the ticket as well with the use-case i used. Also newURL is an NSURL not an NSURLRequest

hansemannn: Check it again:
```objc
NSDictionary* evt = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:request], @"url", @"Webview did not load blacklisted url.", @"messsage", nil];
```

uses the variable `request` which is no `NSURL` (which should also be received correctly).

AngelkPetkov: Oh that sorry, thought you was talking about the if statement. Yeah it works, i cant pass in the URL as it causes a weird bug which we disguised, where the page will finish loading before reaching the return statement.

iphone/Classes/TiUIWebView.m
```diff
@@ -672,21 +698,33 @@ -(CGFloat)contentWidthForWidth:(CGFloat)value
...
 	if (navigationType != UIWebViewNavigationTypeOther) {
 		RELEASE_TO_NIL(lastValidLoad);
 	}
-
+    
+	if (blacklistedURLs) {
+           for (NSString*check in blacklistedURLs)
+           {
+               if ([[newUrl absoluteString] rangeOfString:check options:NSCaseInsensitiveSearch].location != NSNotFound) {
+                   if([self.proxy _hasListeners:@"onStopBlacklistedUrl"]) {
```
yomybaby: This event name should be `stopBlacklistedUrl` without `on`, right?

PR #8022, opened 2016-05-26
[TIMOB-23443] Android CLI: Log messages sometimes get truncated arbitrarily

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23443

**Description:**
When running a mocha unit test suite locally, I have test results being spit out as JSON in the logs and the wrapping process looks for them, parses them out and cobbles them together. However, some JSOn logs were getting truncated apparently at random. The root cause appears to be getting event callbacks on stdout 'data' events from spawn mid-line from our logical process. Then our CLI hook assumes it's always sets of full line and regexps fail, so some logs actually got dropped.


android/cli/hooks/run.js
```diff
@@ -233,44 +233,56 @@ exports.init = function (logger, config, cli) {
...
-									break;
-								case 'e':
-									logger.error(line);
-									break;
-								case 'i':
-								default:
-									logger.info(line);
+						} else {
+							// if it begins with something like "E/SQLiteLog( 1659):" it's not a contination, don't log it.
+							if (/^\w\/\w+\s*\(\s*\d+\):/.test(line)) {
```
cheekiatng: So this is where the main solution lies?

sgtcoolguy: Yes and no. Main solution is in adb.js, using stream splitter to pump out events. That library basically enforces we spit an event per-line and not mid-line. This is an additional catch that if a log message gets spread across multiple lines we'll still log it (unless it looks like it's the start of another log message that isn't TiAPI).

cb1kenobi: You should cache this regex up near `tiapiRegExp` since this function is hot.

PR #8026, opened 2016-05-27
[AC-3673] Android: Optimize removeAllChildren

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-3673

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -33,6 +33,7 @@ import org.appcelerator.titanium.util.TiUrl;
 import org.appcelerator.titanium.util.TiUIHelper;
 import org.appcelerator.titanium.view.TiAnimation;
 import org.appcelerator.titanium.view.TiUIView;
+import org.appcelerator.titanium.view.TiCompositeLayout;
 
 import android.animation.Animator;
 import android.animation.AnimatorListenerAdapter;
@@ -45,8 +46,7 @@ import android.os.Message;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewAnimationUtils;
-import org.appcelerator.titanium.view.TiCompositeLayout;
-
+import android.widget.ListView;
 /**
  * The parent class of view proxies.
  */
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
@m1ga Is this now ready for review? cc @ashcoding

sgtcoolguy:
ok to test

ashcoding:
@m1ga if it's good to test, I'll give it a look.

m1ga:
Yes,please test it. I left some comments in the jira ticket about the listview

ashcoding:
Functionally tested it out.
Before:
4026ms
4293ms
3910ms
4207ms
3979ms

After:
264ms
342ms
338ms

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -718,13 +730,20 @@ public void remove(TiViewProxy child)
...
+		if (children != null && view != null) {
+			View nv = view.getNativeView();
+			if (!(nv instanceof TiCompositeLayout) && nv.getClass().getSimpleName()!="ListViewWrapper") {
+				nv = ((ViewGroup) nv).getChildAt(0);
+			}
+				
+			if (nv instanceof ViewGroup) {
+				ViewGroup vg = (ViewGroup) nv;
+				if (TiApplication.isUIThread()) {
+					vg.removeAllViews();
```
ashcoding: The original code uses the method ```remove(child)``` to remove the TiViewProxy child view. This in turns call ```handleRemove()```.

In the handleRemove, the TiViewProxy is released by the method ```releaseViews()```. This is not done here. Would there be a leak caused from here?

m1ga: I didn't do any memory checks but according to http://stackoverflow.com/a/29942391/5193915 it will null all child views but it won't call release() on all children, which is the other part of handleRemove(). I'll try to see if I can do some memory testing.

m1ga: First test looks pretty good: 
https://jira.appcelerator.org/browse/TIMOB-23447?focusedCommentId=389156&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-389156

ashcoding:
I tried this with 500 views.

Prior to this, I get 77,300~ objects after GC (before GC, it was around 121,000). Although it takes a lot of time. 

With this PR, I get 121,404~ objects after GC. Although it is very fast.

In short, the objects are not being released. 😭

m1ga:
:sad: ok, perhaps there is another way of releasing the objects and still keep the removeAllViews(). Just strange that in my test run the objects went down again. But I'm not that familiar with memory testing so I might have done something wrong

ashcoding:
I'll give this a look again when I have time if I can.

hansemannn:
@m1ga any progress here?

m1ga:
@hansemannn sorry, no. It looks like `handleRemove()` has to be called on all objects before removing them and then you have to do the loop over all objects again (like it is now). Since there is a JIRA ticket you might close this PR. Perhaps somebody else comes up with a clever idea to speed up removing of objects

PR #8027, opened 2016-05-27
[TIMOB-23445] iOS: Allow way to avoid the red error controller modal with build setting

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23445

**Description:** This lets you specify a CLI arg of `--hide-error-controller` to force avoiding showing the error controller (like we did in production before). By default we'll hide the error controller in production, but not development or test. This is effectively to allow "skipping" the error controller for simulator/device builds for unit testing. If we pope the error controller that effectively stops iOS from trying to open any new windows/dialogs - so any test failure ultimately makes every UI test that does those things also fail (incorrectly).


hansemannn:
@sgtcoolguy you want that in 5.4.0 or 6.0.0?

cheekiatng:
As said in JIRA, it's for 6.0.0

hansemannn:
Approved!

PR #8031, opened 2016-05-28
[TIMOB-23406] Android: add 'scrolling' event to listView

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23406

Partiy PR

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
@@ -372,9 +372,9 @@ public class TiListView extends TiUIView implements OnSearchChangeListener {
 				
 				String scrollingDir = null;
 				if (_firstVisibleItem > _lastFirstVisibleItem) {
-					scrollingDir = "down";
-				} else if (_firstVisibleItem < _lastFirstVisibleItem) {
 					scrollingDir = "up";
+				} else if (_firstVisibleItem < _lastFirstVisibleItem) {
+					scrollingDir = "down";
 				}
 				_lastFirstVisibleItem = _firstVisibleItem;
 
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
@@ -317,7 +317,7 @@ public class TiListView extends TiUIView implements OnSearchChangeListener {
 		{
 			private int _firstVisibleItem = 0;
 			private int _visibleItemCount = 0;
-			private int _lastFirstVisibleItem = 0;
+			private int lastY = 0;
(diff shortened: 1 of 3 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -367,6 +368,23 @@ public void onScroll(AbsListView view, int firstVisibleItem, int visibleItemCoun
 			{
 				_firstVisibleItem = firstVisibleItem;
 				_visibleItemCount = visibleItemCount;
+				
+				if (listView.getChildAt(0) != null){
```
hieupham007: you can do Child child = listView.getChildAt(0), so you dont have to call it twice

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -367,6 +368,23 @@ public void onScroll(AbsListView view, int firstVisibleItem, int visibleItemCoun
...
+					if (currentY < lastY) {
+						scrollingDir = "up";
+					} else if (currentY > lastY) {
+						scrollingDir = "down";
+					}
+					if (scrollingDir != null){
+						KrollDict eventArgs = new KrollDict();
+						eventArgs.put("direction", scrollingDir);
+						eventArgs.put("velocity", (lastY-currentY));
+						fProxy.fireEvent(TiC.EVENT_SCROLLING, eventArgs, false);
```
hieupham007: onScroll is called many times when the listView is scrolling, this may not be the best way to implement this method. onScrollStateChange may be a better option

m1ga: onScrollStateChange is only called when you start or end scrolling, so you won't get any direction changes, you have to lift your finger and start scrolling again to get the new direction. 

But I can change the event so it will fire only if it gets a new direction, that would reduce the events

If you prefer the onScrollStateChange event the direction parameter could be added to the scrollstart/scrollend callback.

ashcoding:
@m1ga Any movement on this?
The first comment ```View child = listView.getChildAt(0)``` is pretty valid. As well as the second comment.

m1ga:
@ashcoding I'll have a look at it again. There wasn't any feedback on my comment, thats why I didn't go any further with it :) But I'll have a look at it again and change the parts. I think there was a "bug" with checking the position of child(0) because it will jump from time to time (it looks like its removing the top item when its too far up).

hansemannn:
@m1ga Can you give me a timeframe in which you can take a look? I scheduled it for 6.1.0 now.

m1ga:
@hansemannn  Found a way to compile under linux again! So I can continue to work on it at the weekend!

m1ga:
Closing this in favor of: https://github.com/appcelerator/titanium_mobile/pull/8216

PR #8038, opened 2016-05-31
[TIMOB-23459]: Null check in getFile()

hieupham007:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23459

### DIFF
--- a/android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java
+++ b/android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java
@@ -81,6 +81,10 @@ public class FilesystemModule extends KrollModule
 	@Kroll.method
 	public FileProxy getFile(KrollInvocation invocation, Object[] parts)
 	{
+		//If directory doesn't exist, return
+		if (parts[0] == null) {
+			return null;
+		}
 		String[] sparts = TiConvert.toStringArray(parts);
 		return new FileProxy(invocation.getSourceUrl(), sparts);
 	}


android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java
```diff
@@ -81,6 +81,10 @@ public boolean isExternalStoragePresent()
 	@Kroll.method
 	public FileProxy getFile(KrollInvocation invocation, Object[] parts)
 	{
+		//If directory doesn't exist, return
+		if (parts[0] == null) {
```
sgtcoolguy: What if any of the parts are null? Does it cause an NPE or end up with a nonsense "null" string as a path segment?

ashcoding:
Functionally tested this. Crash does not occur any more.
Will wait for comment by @sgtcoolguy to be addressed first before further actions.

android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java
hansemannn: Just out of curiosity: Maybe `Arrays.asList(parts).subList(0, parts.length).contains(null)` or a manual loop?

sgtcoolguy: Also, I'd say we should log a warning or error to the user to let them know they passed in a null argument (maybe spit out the args received?) and we're returning null. Or throw a JS Error or something so they can pinpoint the bad call and know why they get back a null reference (or get the error thrown).

hieupham007: If the file name is null, it won't make it to Java. I think the check should be efficient

ashcoding:
I'm going to merge and will have another PR to log out a warning.

PR #8042, opened 2016-06-05
[TIMOB-23474] Android: Support titleid, messageid & okid in Ti.UI.AlertDialog (parity)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23474


android/modules/ui/src/java/ti/modules/titanium/ui/AlertDialogProxy.java
```diff
@@ -23,8 +23,11 @@
 		TiC.PROPERTY_BUTTON_NAMES,
 		TiC.PROPERTY_CANCEL,
 		TiC.PROPERTY_MESSAGE,
+		TiC.PROPERTY_MESSAGEID,
 		TiC.PROPERTY_TITLE,
+		TiC.PROPERTY_TITLEID,
 		TiC.PROPERTY_OK,
+		TiC.PROPERTY_OKID,
```
hieupham007: you may want to update the "since" for these properties

hansemannn: Thought about that, but the properties themselves have been available before, just the getter/setter haven't. So how would we document that particular case?

sgtcoolguy:
LGTM 👍

android/modules/ui/src/java/ti/modules/titanium/ui/AlertDialogProxy.java
hieupham007: probably easiest to assume they didn't work before, but works now.

hansemannn:
@sgtcoolguy Feel free if you are good with it 😊

PR #8058, opened 2016-06-13
[TIMOB-23508] iOS: Use native NSData base64 encode API

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23508)

Currently the constants are placed inside Ti.Utils however since the android implementation will be using a different API the names of the constants will differ. So the iOS constants could be placed here or moved to Ti.App.iOS.


sgtcoolguy:
Given that we want cross-platform parity, you may want to consider what options the other platforms could do as well. The CRLF vs LF options also applies easily to Android's impl. The length values could be hacked (Android doesn't have explicit options for that, but you could easily insert your own newlines at different lengths as a post-process in our proxy there). 

I assume you'd also want a "no newlines" option as well?

And Android has an URL_SAFE option which I see NSData doesn't, but may make sense to try and "hack" as a post-process to convert `-` and `_` to `+` and `/`. Not sure how useful people may find that variant, maybe don't need it in a first pass?

sgtcoolguy:
I know Windows could easily be changed to handle different line lengths and padding characters. I'm less sure about changing the type of line breaks; and I think it _always_ does the URL SAFE variant (using `+/` instead of `-_`)

hansemannn:
Putting the review on hold for now, we should ensure that the available cross-platform options are covered.

AngelkPetkov:
Closing this PR to create a new clean solution..

PR #8062, opened 2016-06-14
[TIMOB-23514] iOS10: Deprecate iAds

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23514


AngelkPetkov:
FT and CR approved!

PR #8074, opened 2016-06-18
iOS: Fix all memory leaks, Xcode-8 warnings & analyzer comments

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23542


iphone/Classes/TiContactsGroup.m
```diff
@@ -276,14 +276,15 @@ -(void)add:(id)arg
...
 			DebugLog(@"[ERROR] Unable to add");
+
 			return;
 		}
 		if (![ourContactStore executeSaveRequest:saveRequest error:&error]) {
 			[self throwException:[NSString stringWithFormat:@"Unable to add member to group: %@",[TiUtils messageFromError:error]]
 					   subreason:nil
 						location:CODELOCATION];
 		};
-		RELEASE_TO_NIL(saveRequest)
```
cheekiatng: these changes to TiContactsGroup cannot be done. See
https://jira.appcelerator.org/browse/TIMOB-20010
It was done before and I had to revert it because this causes issues.
see this commit:
https://github.com/appcelerator/titanium_mobile/commit/a0d4a50d51f1afe85f92cf9e0d2ce8cca08fcf2f

iphone/Classes/TiContactsPerson.m
```diff
@@ -948,28 +948,28 @@ -(void)setValue:(id)value forUndefinedKey:(NSString*)key
 //For iOS9 deleting contact
 -(CNSaveRequest*)getSaveRequestForDeletion
 {
-	CNSaveRequest *saveRequest = [[CNSaveRequest alloc] init];
+	CNSaveRequest *saveRequest = [[[CNSaveRequest alloc] init] autorelease];
```
cheekiatng: same thing here.
https://github.com/appcelerator/titanium_mobile/pull/7464/files
We cannot do these autorelease.

hansemannn:
If so, we would need to at least make a new PR for https://github.com/appcelerator/titanium_mobile/pull/8074/commits/92f08caea753beff70660e925387723eba4c1884, maybe together with the updated default-entitlements that cause problems in Xcode 8.

PR #8004, opened 2016-05-17
[TIMOB-23382] Parity: Cannot require directory as done in NodeJS

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23382
https://jira.appcelerator.org/browse/TIMOB-23381
https://jira.appcelerator.org/browse/TIMOB-23138

This is implemented on **Android** and **iOS**, and adds some unit tests around the behavior. This likely needs more testing to make sure I didn't break requires of native modules containing CommonJS extensions. It'd be helpful to have some actual examples to test with.

This implements the [NodeJS require algorithm](https://nodejs.org/api/modules.html#modules_all_together) with some tweaks:
- We don't attempt to load *.node files
- We don't currently do lookups via node_modules directories (yet, easy to add)
- If the require string has no ., /, or .. prefix and it's not a native module we fall back to legacy Titanium behavior of assuming the require is meant as "absolute" inside the app (aka, starting at "Resources/")

So what does this do? Well it:
- lets us load JSON files
- lets us load directories
  - look at package.json's main property, resolve it and try to load it
  - look for index.js
  - look for index.json

So to sum up, our algorithm in this PR is:

```
require(X) from module at path Y
1. If X is a core module,
   a. return the core module
   b. STOP
2. If X begins with './' or '../'
   a. LOAD_AS_FILE(Y + X)
   b. LOAD_AS_DIRECTORY(Y + X)
3. If X begins with '/'
   a. LOAD_AS_FILE(X)
   b. LOAD_AS_DIRECTORY(X)
4. If X **does not** contain '/' # Assume we should try and load CommonJS module first...
   a. LOAD_AS_FILE(X/X.js) # Try to load "legacy" CommonJS file named module.id/module.id.js
   b. LOAD_AS_DIRECTORY(X) # try to load CommonJS module as a directory
5. WARN user about possibly bad require being treated as absolute 
   a. LOAD_AS_FILE(X)
   b. LOAD_AS_DIRECTORY(X)
6. THROW "not found"

LOAD_AS_FILE(X)
1. If X is a file, load X as JavaScript text or JavaScript Object (if extension is json).  STOP
2. If X.js is a file, load X.js as JavaScript text.  STOP
3. If X.json is a file, parse X.json to a JavaScript Object.  STOP

LOAD_AS_DIRECTORY(X)
1. If X/package.json is a file,
   a. Parse X/package.json, and look for "main" field.
   b. let M = X + (json main field)
[...]

PR #8075, opened 2016-06-20
[TIMOB-16078] Implement full NodeJS require specs so we can use NPM for modules

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16078

**Description:**
This adds searching node_modules directories up the directory tree for modules when doing a require without any /, ./, or ../ prefix.

Continuation of #8004 work.

TODO: What's the priority between node modules, CommonJS modules, and legacy Titanium requires? Currently this will try to load module.id/module.id.js for CommonJS first, then node_modules, then legacy/CommonJS loading.

So with this PR, the require algorithm looks like this:

```
require(X) from module at path Y
1. If X is a core module,
   a. return the core module
   b. STOP
2. If X begins with './' or '../'
   a. LOAD_AS_FILE(Y + X)
   b. LOAD_AS_DIRECTORY(Y + X)
3. If X begins with '/'
   a. LOAD_AS_FILE(X)
   b. LOAD_AS_DIRECTORY(X)
4. If X **does not** contain '/' # Assume we should try and load CommonJS module first...
   a. LOAD_AS_FILE(X/X.js) # Try to load "legacy" CommonJS file named module.id/module.id.js
   b. LOAD_AS_DIRECTORY(X) # try to load CommonJS module as a directory
5. LOAD_NODE_MODULES(X, dirname(Y))
6. WARN user about possibly bad require being treated as absolute 
   a. LOAD_AS_FILE(X)
   b. LOAD_AS_DIRECTORY(X)
7. THROW "not found"

LOAD_AS_FILE(X)
1. If X is a file, load X as JavaScript text or JavaScript Object (if extension is json).  STOP
2. If X.js is a file, load X.js as JavaScript text.  STOP
3. If X.json is a file, parse X.json to a JavaScript Object.  STOP

LOAD_AS_DIRECTORY(X)
1. If X/package.json is a file,
   a. Parse X/package.json, and look for "main" field.
   b. let M = X + (json main field)
   c. LOAD_AS_FILE(M)
2. If X/index.js is a file, load X/index.js as JavaScript text.  STOP
3. If X/index.json is a file, parse X/index.json to a JavaScript object. STOP

LOAD_NODE_MODULES(X, START)
1. let DIRS=NODE_MODULES_PATHS(START)
2. for each DIR in DIRS:
   a. LOAD_AS_FILE(DIR/X)
   b. LOAD_AS_DIRECTORY(DIR/X)

NODE_MODULES_PATHS(START)
1. let PARTS = path split(START)
2. let I = count of PARTS - 1
3. let DIRS = []
4. while I >= 0,
   a. if PARTS[I] = "node_modules" CONTINUE
   c. DIR = path join(PARTS[0 .. I] + "node_modules")
   b. DIRS = DIRS + DIR
   c. let I = I - 1
5. return DIRS
```


sgtcoolguy:
cc @jawa9000 This PR will extend the behavior changed in #8004 slightly, but should probably be folded into whatever docs you may have prepared for that. The addition here is the search for node_modules (step 5)

iphone/Classes/KrollBridge.m
```diff
@@ -770,7 +770,7 @@ -(id)loadCommonJSModule:(NSString*)code withSourceURL:(NSURL *)sourceURL
 	 *      "/Resources/views/login/window.js"  __dirname = "views/login"
 	 */
 
-	NSString *filename = [sourceURL lastPathComponent];
+	NSString *filename = [sourceURL lastPathComponent]; // FIXME This should be the absolute path to the file!
```
cheekiatng: @sgtcoolguy isn't the absolute file path the following line? 
```
NSString *js = [[NSString alloc] initWithFormat:TitaniumModuleRequireFormat, dirname, filename,code];
```
What do you need changed here?

sgtcoolguy: I opened a separate ticket for parity about this, and didn't really investigate it in this one. Basically __filename should be the full absolute path to the resolved file on all platforms. On iOS and Android we report just the filename (i.e "test.js"). On Windows we report the base name of the file (i.e. "test")

See https://jira.appcelerator.org/browse/TIMOB-23543

sgtcoolguy: And in this case, it's some variation of the sourceURL without the preceding scheme and maybe some of the first path segments we don't "expose" (like maybe a "Resources/" prefix)?

cheekiatng: @sgtcoolguy so now with this fixed here: https://github.com/appcelerator/titanium_mobile/pull/8098
do we still need to address this? @hansemannn fyi.

hansemannn: Should be fixed with the mentioned PR. The line is only used to expose the `__filename` and that's done.

cheekiatng: k. @sgtcoolguy can you fix the merge conflicts? Thanks.

sgtcoolguy:
@cheekiatng Ok, rebased on master and fixed the conflict.

iphone/Classes/KrollBridge.m
```diff
@@ -1124,6 +1173,14 @@ - (id)require:(KrollContext *)kroll path:(NSString *)path
 					return module;
 				}
 			}
+
+			// Need base path to work from for determining the node_modules search paths.
+            NSString *workingPath = [oldURL relativePath];
```
cheekiatng: correct this indentation

cheekiatng:
FT passed. Can you fix that minor indention and I'll be able to merge this.

FokkeZB:
👏🙌👏🙌👏🙌

sgtcoolguy:
Merged this manually after massaging the commits a little to combine the iOS changes into one, Android into one, unit test as one.

PR #8077, opened 2016-06-21
[TIMOB-19482] Android: Add multiple TiUIViews to the view hierarchy in one call

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-19482


ashcoding:
This PR needs more work.

hansemannn:
As discussed with Ash on Flowdock, it needs some adjustments regarding Kroll-Annotations. Will take a closer look to that this week.

ashcoding:
I don't think it's done that way. In any case, I'll take a look at it and hopefully I can help you solve this issue. :)

ashcoding:
https://github.com/appcelerator/titanium_mobile/blob/415bd6c66dcc55b1a59a59574f3babd3c3a84ede/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java#L554-L584

This is where the add method is being called. Not the one that was changed in your PR.

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -552,6 +552,25 @@ public void releaseViews()
...
 	public abstract TiUIView createView(Activity activity);
 
+	
+	/**
+	 * Adds multiple child to this view proxy.
+	 * @param childs The child view proxies to add.
+	 * @module.api
+	 */
+	@Kroll.method
+	public void add(TiViewProxy[] childs)
```
ashcoding: We should be using the same add method as below in line 580 (originally 561). Instead, we change the params to allow for an Object class, and in the method, we check the type if it is of type TiViewProxy child or if it's some form of array, cast it correctly and handle the code appropriately.

ashcoding:
You're moving in the right step forward. :+1:

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
hansemannn: You mean all the checks? I thought it would be a good idea to leave them where they are, since the method is called for every child anyway. What else do I need to do? It seems that it still does not work.

hansemannn:
@ashcoding So I think I finally made it by handling it inside the main `add` method. I needed to do some adjustments on the classes that `@Override` the method, but it looks like it works pretty well now. Demo-code also updated. We can schedule it for 6.1.0.

garymathews:
CR: **PASS**
FR: **PASS**

Looks good 👍 

I won't merge just yet since @hansemannn `I needed to do some adjustments on the classes that @Override the method` ?

hansemannn:
@garymathews thanks man! Can you edit it directly from the PR? I just enabled maintainer-edits.

garymathews:
@hansemannn I was referring to the changes you wanted to make :)

I don't mind taking over, I can address the changes in another PR if I can't push directly to this one

hansemannn:
I would appreciate that. Since it's a very affective change, I don't want to break something :-) Feel free to cherry-pick and close this one.

garymathews:
Continued by https://github.com/appcelerator/titanium_mobile/pull/8452

PR #8079, opened 2016-06-22
[TIMOB-23459] Android: Adding a warning when null dir is passed to getFile()

ashcoding:
https://jira.appcelerator.org/browse/TIMOB-23459


ashcoding:
**Merging**

PR #8086, opened 2016-06-24
[TIMOB-23501] iOS: Label.ellipsize should not return undefined

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23501?filter=-1)


iphone/Classes/TiUILabel.m
```diff
@@ -436,11 +436,6 @@ -(void)setColor_:(id)color
 -(void)setEllipsize_:(id)value
 {
 	ENSURE_SINGLE_ARG(value, NSNumber);
-	//for bool case and parity with android
-	if ([TiUtils intValue:value] == 1) {
-		[[self label] setLineBreakMode:NSLineBreakByTruncatingTail];
-		return;
-	}
```
hansemannn: Cannot remove the old implementation, we need to throw at least a deprecation warning for `boolean` inputs.

apidoc/Titanium/UI/Label.yml
```diff
@@ -114,8 +114,8 @@ properties:
     summary: |
         Causes words in the text that are longer than the view is wide to be ellipsized instead of broken in the middle.
     constants: Titanium.UI.TEXT_ELLIPSIZE_TRUNCATE_*
-    type: [Number, Boolean]
-    default: false
+    type: Number
+    default: Titanium.UI.TEXT_ELLIPSIZE_TRUNCATE_END
```
hansemannn: Is Android behaving the same?

apidoc/Titanium/UI/Label.yml
```diff
@@ -114,8 +114,8 @@ properties:
     summary: |
         Causes words in the text that are longer than the view is wide to be ellipsized instead of broken in the middle.
     constants: Titanium.UI.TEXT_ELLIPSIZE_TRUNCATE_*
-    type: [Number, Boolean]
-    default: false
```
hansemannn: So Android does not return a boolean and does not need to be changed?

AngelkPetkov: Android has the same issue and returns undefined

m1ga:
I've added the android part and set the default value to `false` because you can turn it off that way (and its off when you just create a label). So the return value is a boolean in my case. Is there a way to turn it off on iOS, too?

AngelkPetkov:
@m1ga  Hey, i just dropped a comment on the PR, instead of setting a bool for the property we should actually be returning the constant which is currently set for the property.

iphone/Classes/TiUILabel.m
```diff
@@ -436,9 +436,8 @@ -(void)setColor_:(id)color
 -(void)setEllipsize_:(id)value
 {
 	ENSURE_SINGLE_ARG(value, NSNumber);
-	//for bool case and parity with android
 	if ([TiUtils intValue:value] == 1) {
-		[[self label] setLineBreakMode:NSLineBreakByTruncatingTail];
+		[[self label] setLineBreakMode:NSLineBreakByCharWrapping];
```
hansemannn: Instead of this, please check if the input is a bool and leave the if-block as it is, if so.

hansemannn:
@AngelkPetkov please cherry-pick his Android-changes into this PR and I will FT it.

AngelkPetkov:
Updated Boolean check and added https://github.com/appcelerator/titanium_mobile/pull/8094. 

Thank you @m1ga !

hansemannn:
There is an issue here @AngelkPetkov @m1ga:
```js

Ti.API.warn(label.getEllipsize()); // Should be "4" (Integer for Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_END - default)
Ti.API.warn(label.getEllipsize() == Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_END); // Should be "true"

label.setEllipsize(Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_MIDDLE);

Ti.API.warn(label.getEllipsize()); // Should be "5" (Integer for Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_MIDDLE)
Ti.API.warn(label.getEllipsize() == Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_MIDDLE); // Should be "true"
Ti.API.warn(label.getEllipsize() == Ti.UI.TEXT_ELLIPSIZE_TRUNCATE_END); // Should be "false"
```

As the constants only represent a name for a number, the above logs output numbers. On iOS, it is 4 and 5, on Android it is 2 and 1. I think, as long as the user checks agains the constants and not against the numbers, it's fine. We cannot ensure that all constants across all platforms represent the same number, so this is the way to go. Do you guys agree?

apidoc/Titanium/UI/UI.yml
```diff
@@ -1243,6 +1243,30 @@ properties:
     permission: read-only
     platforms: [iphone, ipad]
   
+  - name: TEXT_ELLIPSIZE_TRUNCATE_WORD_WRAP
+    summary: Add ellipses at word boundaries, unless the word itself doesnt fit on a single line
```
hansemannn: - `doesnt` -> `doesn't`
- Missing period at the end of the sentence

apidoc/Titanium/UI/UI.yml
```diff
@@ -1243,6 +1243,30 @@ properties:
...
+        One of the group of constants for the <Titanium.UI.Label.ellipsize> property.
```
hansemannn: But into `description` field.

apidoc/Titanium/UI/UI.yml
```diff
@@ -1243,6 +1243,30 @@ properties:
...
+    summary: Add ellipses before the first character that doesnt fit.
+        One of the group of constants for the <Titanium.UI.Label.ellipsize> property.
+    type: Number
+    permission: read-only
+    platforms: [iphone, ipad]
+    since: "6.0.0"
+
+  - name: TEXT_ELLIPSIZE_TRUNCATE_CLIP
+    summary: Lines are simply not drawn past the edge of the text container.
+        One of the group of constants for the <Titanium.UI.Label.ellipsize> property.
```
hansemannn: But into `description` field.

apidoc/Titanium/UI/UI.yml
```diff
@@ -1243,6 +1243,30 @@ properties:
...
+    summary: Add ellipses at word boundaries, unless the word itself doesnt fit on a single line
+        One of the group of constants for the <Titanium.UI.Label.ellipsize> property.
+    type: Number
+    permission: read-only
+    platforms: [iphone, ipad]
+    since: "6.0.0"
+
+  - name: TEXT_ELLIPSIZE_TRUNCATE_CHAR_WRAP
+    summary: Add ellipses before the first character that doesnt fit.
+        One of the group of constants for the <Titanium.UI.Label.ellipsize> property.
```
hansemannn: But into `description` field.

apidoc/Titanium/UI/UI.yml
AngelkPetkov: Yeah i had it as doesn't originally however node validate doesn't like ' as its a non ASCII character, so i removed it

hansemannn:
FT passed, please address the documentation comments before merging.

apidoc/Titanium/UI/UI.yml
hansemannn: Use the correct sign there as we do in related docs. Search `UI.yml` for other usages and copy it from there.

hansemannn:
CR + FT passed, PR approved.

PR #8089, opened 2016-06-27
[TIMOB-23566] iOS: Support global tintColor that inherits to child-views

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23566

### DIFF
--- a/apidoc/Titanium/UI/UI.yml
+++ b/apidoc/Titanium/UI/UI.yml
@@ -2413,7 +2413,7 @@ properties:
 
   - name: tintColor
     summary: |
-        Sets the global tint color of the application. It is inherited to the child views and can be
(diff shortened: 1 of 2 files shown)


apidoc/Titanium/UI/UI.yml
```diff
@@ -2407,6 +2411,14 @@ properties:
     type: Number
     platforms: [android, iphone, ipad]
 
+  - name: tintColor
+    summary: |
+        Sets the global tint color of the application. It is inherited to the child views and can be
```
AngelkPetkov: Change to :"It is inherited by the child views"

AngelkPetkov:
FT and CR Approved!

PR #8094, opened 2016-06-28
[TIMOB-23500] Android: default value for ellipsize

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23500

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
@@ -44,7 +44,7 @@ public class TiUILabel extends TiUIView
 
 	private int defaultColor;
 	private boolean wordWrap = true;
-	private TruncateAt ellipsize;
+	private TruncateAt ellipsize = TruncateAt.END;
 	private float shadowRadius = DEFAULT_SHADOW_RADIUS;
 	private float shadowX = 0f;
 	private float shadowY = 0f;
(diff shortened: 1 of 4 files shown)


AngelkPetkov:
Hello, thank you for the contribution! However the value for the ellipsize should not actually be a boolean, but it should return the constant. An Android fix is in progress https://github.com/appcelerator/titanium_mobile/pull/8086 for parity reasons.

m1ga:
I've added a comment in your PR :) Sorry for crossposting

AngelkPetkov:
FT Approved!

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -135,6 +135,7 @@ public boolean onTouchEvent(MotionEvent event) {
 		tv.setKeyListener(null);
 		tv.setFocusable(false);
 		tv.setSingleLine(false);
+		tv.setEllipsize(ellipsize);
```
hansemannn: How does that effect current labels that do not have any `ellipsize` property defined? We should try to ensure that nothing breaks after this change (although 6.0 would be the right time to do this).

m1ga: I think it will just be a "good" change:
ell

it will just affect the fields where you have a fixed width (no wordWrap) and the text would be out of the label anyway. That's mostly the point where you are searching on how to switch on ellipsize :)
But since Android can switch of the ellipsizing it might have a different default value. I would prefer `truncateAt.END` because it will make the label look the same on both platforms from the beginning. You decide ;)

hansemannn: Like it, thanks for that great example!

hansemannn:
Cherry-picked into #8086 to be able to test both together. Thank you @m1ga, good work! 🚀

PR #8098, opened 2016-07-01
[TIMOB-23543] iOS: Update require-implementation of __filename

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23543


sgtcoolguy:
LGTM :+1:

sgtcoolguy:
Would be good to include the fixed test for this from ti.require.test.js under tests/Resources. 
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.require.test.js#L111

hansemannn:
Updated the PR with the above-mentioned unit-test, left the Android-skipping for now until the Android-PR arrives. Thanks!

PR #8041, opened 2016-06-03
[TIMOB-23310] Update v8 engine to 5.0.71.33

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23310
https://jira.appcelerator.org/browse/TIMOB-23476

**Description:**

This is a **MAJOR** update of the V8 engine we use for Android. Our current engine is 3.9.24.29, which dates back to May/June 2012, so this brings the engine forward a full 4 years!

Note that right now even this version of V8 is behind. Once we get this merged we should regularly update. Right now Android stable version is: 5.1.281.57 - see https://omahaproxy.appspot.com
Looks like they "release" pretty much weekly. We should consider having regular updates of V8 ourselves so that next time we don't have a major overhaul of all the native code and can handle API changes as they come.

This is a major breaking change and will mean all compiled/native modules for Android will need to be rebuilt.

This currently builds fine against Android NDK r11c for me, and I was able to run a classic two-tab app without problems.

PR #8101, opened 2016-07-01
[TIMOB-20221] Android: Remove deprecated properties and methods pre-5.0.0

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20221

**Description:**
Removes nearly all the deprecated code in Android. I left some Geolocation constants, because I was unsure if we really want them removed yet.

One notable change is the removal of TiContext which was used in every single proxy class as an additional constructor arg for extra constructors. I believe removing this may also be a breaking change for native Android modules. Given that we already need to recompile all of them for 6.0.0 due to the V8 update, this seems fine to me.


ashcoding:
Generally looks okay at a first glance. Are you able to direct me to perhaps a test app, code or test suite to test this?

Test case with emphasis on the classes that has TiContext removed to ensure no regressions.

sgtcoolguy:
_Every_ proxy had a constructor that took TiContext.

ashcoding:
I guess Kitchen Sink would be a good test app. :D

ashcoding:
Functionally tested. Working as per normal. :)

ashcoding:
Found some issues while testing Kitchen Sink but commented it here https://github.com/appcelerator/titanium_mobile/pull/8041#issuecomment-230379452

For the purpose of deprecation, this PR is good to go.

sgtcoolguy:
@ashcoding It seems to work locally for me as well. I'm concerned that the PR build keeps failing with issues where the ti.map recompiled module complains about being unable to find a changed method signature on ProxyFactory class in native code in the SDK. Does Kitchen Sink use the ti.map module? Maybe you can quickly confirm that using some of these recompiled native modules work locally for you as they do for me (and we can chalk it up to some weird build agent setup issue)?

PR #8104, opened 2016-07-02
[TIMOB-23311] iOS simulator inabilty to send mail not restricted to a specific simulator version

stephenfeather:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23311

**Optional Description:**
The message indicates that its a problem with sending mail on just the iOS 8 simulator, when in reality, no simulator up through iOS 9.x has supported sending mail.  Removed reference to '8' to clarify.


hansemannn:
Approved!

PR #8105, opened 2016-07-03
[TIMOB-23297] iOS: Fix Ti.App.getArguments() call on cold app-launch

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23297


cheekiatng:
CR and FT passed. Very good!.

PR #8111, opened 2016-07-07
[TIMOB-23518] iOS: Cannot build with Xcode 8.0

cheekiatng:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23518

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -5674,7 +5674,7 @@ iOSBuilder.prototype.processTiSymbols = function processTiSymbols() {
 };
 
 iOSBuilder.prototype.removeFiles = function removeFiles(next) {
-	this.unmarkBuildDirFiles(path.join(this.buildDir, 'ModuleCache'));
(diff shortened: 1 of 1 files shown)


cb1kenobi:
I thought the `Intermediates` had to go inside the derived data path?

cheekiatng:
@cb1kenobi I considered that, and tried it, but surprisingly they don't sit well together. The build fails. 
everything in `DerivedData` has to be separate from the Intermediate, Products, PrecompiledHeaders directories. Through further investigation, it seems derivedData is supposed to be separate. Native Xcode projects have their DerivedData here:
`/Library/Developer/Xcode/DerivedData`
So it is ok for https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L5939 and the following 2 lines to remain where they are. In fact, if you comment out these 3 lines, by default the build folder is still generated in the same place. But let's keep them for now.

cb1kenobi:
Tested and works! Love it! **APPROVED**!!!

PR #8114, opened 2016-07-07
[TIMOB-20453] (5_3_X) Remove hardcoded launchscreen-define

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20453

**Optional Description:**
Required to fix the remaining launchscreen issues on 5_3_X. It is already removed in 5_4_X and master. We only need this patch if we do a 5.3.2.

### DIFF
--- a/iphone/Classes/defines.h
+++ b/iphone/Classes/defines.h
@@ -6,7 +6,6 @@
 //
 
 // MODULES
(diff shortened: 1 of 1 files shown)


cheekiatng:
APPROVED.

PR #8120, opened 2016-07-11
[TIMOB-23614] Adopt Swift 3 syntax changes in Apple Watch template

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23614

**Optional Description:**
Renames a few methods to comply with Swift 3 naming changes and adds a delegate method that is required since WatchOS 2.2

### DIFF
--- a/templates/applewatch/watchos2-swift/template/{{WatchkitExtName}}/InterfaceController.swift.ejs
+++ b/templates/applewatch/watchos2-swift/template/{{WatchkitExtName}}/InterfaceController.swift.ejs
@@ -10,21 +10,20 @@ import WatchKit
 import WatchConnectivity
 import Foundation
 
-
 class InterfaceController: WKInterfaceController, WCSessionDelegate {
 
     override init() {
         super.init()
 
         if WCSession.isSupported() {
-            let session = WCSession.defaultSession()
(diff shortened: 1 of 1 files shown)


templates/applewatch/watchos2-swift/template/{{WatchkitExtName}}/InterfaceController.swift.ejs
```diff
@@ -39,4 +38,9 @@ class InterfaceController: WKInterfaceController, WCSessionDelegate {
         super.didDeactivate()
     }
 
+    @available(watchOS 2.2, *)
+    public func session(_ session: WCSession, activationDidCompleteWith activationState: WCSessionActivationState, error: NSError?) {
+        // This will be called when the activation of a session finishes.
+    }
+
```
hansemannn: Do we need this block?

janvennemann:
This method is required since iOS 9.3/watchOS 2.2 when using the `WCSessionDelegate` but somehow it didn't threw an error when building under Xcode 7. Building it now with iOS 10/watchOS 3/Xcode 8 requires you to implement this method and to add the `@available` attribute.

hansemannn:
Approved!

PR #8121, opened 2016-07-11
Fix Xcode-8 detection issues

hansemannn:
Since there was a problem with `__IPHONE_10_0`, using this method succeeds on Xcode 7.3.1 as well.


cheekiatng:
CR and FT passed. APPROVED.

PR #8123, opened 2016-07-11
[TIMOB-20083] iOS build output error and exit when a WatchOS1 extension is detected.

feons:
https://jira.appcelerator.org/browse/TIMOB-20083


cheekiatng:
APPROVED!

PR #8128, opened 2016-07-12
Android: Fix shared property issue with recent V8 update

sgtcoolguy:
**Description:**
Fix issue where all instances of a given proxy type shared the same _properties object to hold any JS properties not overridden with getter/setters. Tweak how we dereference Proxy pointers.


cheekiatng:
CR and FT passed! APPROVED!

PR #8129, opened 2016-07-12
Add a new scons command 'cleanbuild'

sgtcoolguy:
**Description:**

Add a new scons command 'cleanbuild' to clean, build, package and install a given platform for dev purposes.

```
node scons.js cleanbuild android
```

is equivalent to:

```
node scons.js clean
node scons.js build android
node scons.js package android
node scons.js install
```

Note that this basically just runs these serially and doesn't optimize to avoid zipping and unzipping the end SDK (we could avoid zipping and just copy the files into the destination as we go to improve performance quite a bit).

### DIFF
--- a/build/scons-cleanbuild.js
+++ b/build/scons-cleanbuild.js
@@ -1,0 +1,133 @@
+#!/usr/bin/env node
+
+var exec = require('child_process').exec,
+	os = require('os'),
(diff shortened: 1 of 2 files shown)


cheekiatng:
CR and FT passed. APPROVED.

PR #8140, opened 2016-07-18
[TIMOB-23649] Use per-platform moduleAPIVersions in the SDK manifest.json

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23649

**Description:**
This change the moduleAPIVersion in the SDK's manifest.json from a single string value (of "2") to an object with string values for each platform. Right now we're at "2" for every platform except android, which is now "3".

We should likely wait until the Studio PR to enable support for this and the older model is merged: 
appcelerator/titanium_studio#788

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,12 @@
   "name": "titanium-mobile",
   "description": "Appcelerator Titanium Mobile",
   "version": "6.0.0",
-  "moduleApiVersion": 2,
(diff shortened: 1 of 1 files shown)


ashcoding:
Will wait for Studio PR.

sgtcoolguy:
@ashcoding The Studio PR has been merged.

PR #8142, opened 2016-07-19
TIMOB-23458 - build command should include -d flag

skypanther:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23458

Adds the `-d` (downgrade permitted) flag to the `adb install` command generated when you build to an Android device. This prevents the `[ERROR] Error: INSTALL_FAILED_VERSION_DOWNGRADE` failure if the app exists on the device with a newer version. 

It seems like the titanium-sdk module must be maintained in a separate repository. But, I could not find it. So, I've made the modification here. I'll be glad to submit the PR elsewhere if needed.


cb1kenobi:
Looks good! **APPROVED**

PR #8152, opened 2016-07-24
[TIMOB-20453] (5_4_X) iOS: Fix LaunchScreen from being RootView

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20453


cheekiatng:
CR and FT passed.

PR #8160, opened 2016-07-26
[TIMOB-23684] iOS: Expose properties to hide the navbar

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23684

Exposed the methods described above: 

https://developer.apple.com/library/ios/documentation/UIKit/Reference/UINavigationController_Class/#//apple_ref/doc/uid/TP40006934-CH3-SW36

### DIFF
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -466,6 +466,38 @@ properties:
     default: false
     since: 3.2.0
 
+  - name: hidesBarsOnSwipe
+    summary: Set this to true to hide the navigation bar on swipe.
+    description: When this property is set to true, an upward swipe hides the navigation bar and toolbar. A downward swipe shows both bars again. If the toolbar does not have any items, it remains visible even after a swipe.
+    platforms: [iphone, ipad]
+    type: Boolean
+    default: false
+    since: 6.0.0
+
+  - name: hidesBarsOnTap
+    summary: Set this to true to hide the navigation bar on tap.
+    description: When the value of this property is true, the navigation controller toggles the hiding and showing of its navigation bar and toolbar in response to an otherwise unhandled tap in the content area.
+    platforms: [iphone, ipad]
+    type: Boolean
+    default: false
+    since: 6.0.0
+
+  - name: hidesBarsWhenVerticallyCompact
+    summary: Set this to true to hide the navigation bar in a vertically compact environment.
+    description: When the value of this property is true, the navigation controller hides its navigation bar and toolbar when it transitions to a vertically compact environment. Upon returning to a vertically regular environment, the navigation controller automatically shows both bars again. In addition, unhandled taps in the content area cause the navigation controller to show both bars again.
+    platforms: [iphone, ipad]
+    type: Boolean
+    default: false
+    since: 6.0.0
+
+  - name: hidesBarsWhenKeyboardAppears
+    summary: Set this to true to hide the navigation bar when the keyboard appears..
+    description: When this property is set to true, the appearance of the keyboard causes the navigation controller to hide its navigation bar and toolbar.
+    platforms: [iphone, ipad]
+    type: Boolean
+    default: false
+    since: 6.0.0
+
   - name: left
     summary: Window's left position, in platform-specific units.
     description: |
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -476,7 +476,7 @@ properties:
     type: Boolean
     default: false
     since: 6.0.0
-    osver: {ios: "8.0"}
+    osver: {ios: {min: "8.0"}}
 
   - name: hidesBarsOnTap
     summary: Set this to true to hide the navigation bar on tap.
@@ -488,7 +488,7 @@ properties:
     type: Boolean
     default: false
     since: 6.0.0
-    osver: {ios: "8.0"}
+    osver: {ios: {min: "8.0"}}
 
   - name: hidesBarsWhenVerticallyCompact
     summary: Set this to true to hide the navigation bar in a vertically compact environment.
@@ -501,7 +501,7 @@ properties:
     type: Boolean
     default: false
     since: 6.0.0
-    osver: {ios: "8.0"}
+    osver: {ios: {min: "8.0"}}
 
   - name: hidesBarsWhenKeyboardAppears
     summary: Set this to true to hide the navigation bar when the keyboard appears.
@@ -512,7 +512,7 @@ properties:
     type: Boolean
     default: false
     since: 6.0.0
-    osver: {ios: "8.0"}
+    osver: {ios: {min: "8.0"}}
 
   - name: left
     summary: Window's left position, in platform-specific units.
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -504,7 +504,7 @@ properties:
     osver: {ios: "8.0"}
 
   - name: hidesBarsWhenKeyboardAppears
-    summary: Set this to true to hide the navigation bar when the keyboard appears..
+    summary: Set this to true to hide the navigation bar when the keyboard appears.
     description: |
         When this property is set to true, the appearance of the keyboard causes the 
         navigation controller to hide its navigation bar and toolbar.
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -473,6 +473,7 @@ properties:
     type: Boolean
     default: false
     since: 6.0.0
+    osver: {ios: "8.0"}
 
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiUIWindowProxy.m
```diff
@@ -304,6 +304,7 @@ - (void)viewWillAppear:(BOOL)animated;    // Called when the view is about to ma
 - (void)viewDidAppear:(BOOL)animated;     // Called when the view has been fully transitioned onto the screen. Default does nothing
 {
 	[self updateTitleView];
+    [self updateHidesBars];
```
AngelkPetkov: Please fix indentation

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -920,6 +921,45 @@ -(void)setToolbar:(id)items withObject:(id)properties
 	
 }
 
+-(void)setHidesBarsOnSwipe:(id)value
+{
```
AngelkPetkov: We need validation, please add  

`ENSURE_TYPE(value, NSNumber);`

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -920,6 +921,45 @@ -(void)setToolbar:(id)items withObject:(id)properties
...
+    [self replaceValue:value forKey:@"hidesBarsOnSwipe" notification:NO];
+    [self updateHidesBars];
+}
+
+-(void)setHidesBarsOnTap:(id)value
+{
+    [self replaceValue:value forKey:@"hidesBarsOnTap" notification:NO];
```
AngelkPetkov: Same here, add it for all please

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -920,6 +921,45 @@ -(void)setToolbar:(id)items withObject:(id)properties
...
+}
+
+-(void)setHidesBarsWhenKeyboardAppears:(id)value
+{
+    [self replaceValue:value forKey:@"hidesBarsWhenKeyboardAppears" notification:NO];
+    [self updateHidesBars];
+}
+
+-(void)updateHidesBars
+{
```
AngelkPetkov: Use `ENSURE_UI_THREAD `

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -920,6 +921,45 @@ -(void)setToolbar:(id)items withObject:(id)properties
...
+    [self updateHidesBars];
+}
+
+-(void)setHidesBarsWhenKeyboardAppears:(id)value
+{
+    [self replaceValue:value forKey:@"hidesBarsWhenKeyboardAppears" notification:NO];
+    [self updateHidesBars];
+}
+
+-(void)updateHidesBars
```
AngelkPetkov: Why do we call all of them together as oppose as each to their own setter ?

AngelkPetkov:
Did a quick review, looks good just some minor changes to be addressed. 

Also could you please provide a  javaScript demo-code for the FT.

iphone/Classes/TiUIWindowProxy.m
kopiro: Uhm, it's a bool!

iphone/Classes/TiUIWindowProxy.m
kopiro: For a better code organization, nope?

iphone/Classes/TiUIWindowProxy.m
kopiro: image

I think someone broke the indentation with tabs.

iphone/Classes/TiUIWindowProxy.m
hansemannn: They come to the proxies as NSNumbers (because of the JavaScriptCore below it). So usually this is what you do in boolean-setters:
```objc
- (void)setHidesBarsOnSwipe:(id)value
{
    ENSURE_TYPE(value, NSNumber);
 
    [myView setWhatever:[TiUtils boolValue:value def:YES]];
    [self replaceValue:value forKey:@"hidesBarsOnSwipe" notification:NO];
}
```

kopiro: Perfect! Thank you!

iphone/Classes/TiUIWindowProxy.m
hansemannn: It's more about performance. Imagine, you use all provided properties. Then 5 setters will be called and 25 native calls will be made, since it calls the `[self updateHideBars]` every time. The only reason this way would make sense is, when one property depends on another one and they need to be updated as soon as another one is. If that is not the case, updating it in the setter is the way to go. 

If they require main-thread, use `ENSURE_MAIN_THREAD(methodName, argName)` to ensure that the method runs on the main-thread.

iphone/Classes/TiUIWindowProxy.m
hansemannn: tl;dr:
It's a legacy part of the code, where tabs have been used. 

The way to go here is to use the pattern that is already there. All new sources have spaces by default, but some old one have tabs. The only reason we don't run a script to fix it to spaces everywhere is, to not loose the ability to git-blame the source, because it sometimes can be very important to see why a certain change was made back in the days.

iphone/Classes/TiUIWindowProxy.m
kopiro: Should I keep `TiThreadPerformOnMainThread` ?

iphone/Classes/TiUIWindowProxy.m
kopiro: Ok, so tab ;)

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -304,6 +304,10 @@ - (void)viewWillAppear:(BOOL)animated;    // Called when the view is about to ma
 - (void)viewDidAppear:(BOOL)animated;     // Called when the view has been fully transitioned onto the screen. Default does nothing
 {
 	[self updateTitleView];
+    [self updateHidesBarsOnSwipe];
+    [self updateHidesBarsOnTap];
+    [self updateHidesBarsWhenVerticallyCompact];
+    [self updateHidesBarsWhenKeyboardAppears];
```
hansemannn: Indentation again. And do we need to call these on appear as well? I would expect that is already gets set by the setters used and the default behavior for the remaining ones.

iphone/Classes/TiUIWindowProxy.m
```diff
@@ -920,6 +924,86 @@ -(void)setToolbar:(id)items withObject:(id)properties
...
+{
+    if ([TiUtils isIOS8OrGreater]) {
+        TiThreadPerformOnMainThread(^{
+            if ((controller != nil) && ([controller navigationController] != nil)) {
+                UINavigationController *ourNC = [controller navigationController];
+                ourNC.hidesBarsOnSwipe = [TiUtils boolValue:[self valueForUndefinedKey:@"hidesBarsOnSwipe"] def:NO];
+            }
+        }, NO);
+    }
+}
```
hansemannn: I would do that in the setters directly to keep the number of methods small and the code semantically together.

iphone/Classes/TiUIWindowProxy.m
kopiro: We need to call because you can pass on constructor, not?

kopiro: Ok, you're right, no need to do that.

kopiro:
@hansemannn and @AngelkPetkov - fixed and added doc in YML.

apidoc/Titanium/UI/Window.yml
```diff
@@ -466,6 +466,54 @@ properties:
...
+    summary: Set this to true to hide the navigation bar on swipe.
+    description: |
+        When this property is set to true, an upward swipe hides the navigation bar and toolbar. 
+        A downward swipe shows both bars again.  If the toolbar does not have any items, it remains
+        visible even after a swipe.
+    platforms: [iphone, ipad]
+    type: Boolean
+    default: false
+    since: 6.0.0
+    osver: {ios: "8.0"}
```
hansemannn: This is invalid. Check my comment again, it's
```yml
    osver: {ios: {min: "8.0"}}
```

hansemannn:
We can merge as soon the demo-code is provided in JIRA

hansemannn:
Approved!

kopiro:
ERROR] Invalid type passed to function. expected: Number, was: (null) -[TiUIWindowProxy setHidesBarsOnTap:] (TiUIWindowProxy.m:943)

kopiro:
I commited with `ENSURE_TYPE_OR_NIL(value, NSNumber);`

hansemannn:
Well, it is correct as it is. All methods expect a boolean value like natively.

PR #8161, opened 2016-07-27
[TIMOB-16985] Android: Removed Ti.UI.Window.url

ashcoding:
https://jira.appcelerator.org/browse/TIMOB-16985


ashcoding:
I believe this is the only part for the ```url``` property to be removed.

hansemannn:
Will wait for the log-warning to be added as discussed.

ashcoding:
I do not see the usage of ```TiC.PROPERTY_URL``` which is the string ```"url"``` for the WindowProxy in Android. For it to have a log-warning, the property still needs to be in. Since it's not in use, I think removing it is fine.

With the documentation for 6.0.0 and all the information informing people about breaking changes, I think this should suffice.

hansemannn:
Alrighty! Approved!

PR #8162, opened 2016-07-27
[TIMOB-23691] iOS: Expose statusBarBackgroundColor

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23691


iphone/Classes/TiUIWindowProxy.m
```diff
@@ -363,6 +363,23 @@ -(void)setBarColor:(id)colorString
...
+    if (controller != nil && [controller navigationController] != nil)
+    {
+        if (statusbarBg == nil) {
+            statusbarBg = [[UIView alloc] initWithFrame:[UIApplication sharedApplication].statusBarFrame];
+            [[controller navigationController].view addSubview:statusbarBg];
+        }
+        TiColor* newColor = [TiUtils colorValue:color];
+        statusbarBg.backgroundColor = [newColor _color];
+    }
+}
```
hansemannn: Not sure if that's the best way to go. I like the style proposed here: http://stackoverflow.com/a/34658477/5537752

Also, `statusbarBg` should be released from memory every time it is set again, as well as in the `dealloc` method to clean it from memory after the gets GC'd.

Please try the above-mentioned way to do it.

kopiro:
@hansemannn Wow, thank you!
It works like a charm.

Pushed ;)

hansemannn:
Happy to hear. One thing that concerns me is, that we can set the `statusBarBackgroundColor` in every `Ti.UI.Window` context but it is set globally. So either we place it in `Ti.UI` instead or find a way to make it window-depending.

hansemannn:
And remove `NSString *color = [TiUtils stringValue:colorString];` + replace it with `ENSURE_SINGLE_ARG(value, NSString)`. Name it `value` as well.

kopiro:
Mmm, theoretically we can just call the `setBackground` on every Window open.

hansemannn:
Right. Then please move it to `Ti.UI` and let the user decide how to handle it. Guess that keeps it the most flexible for both use-cases (global- and context-based).

kopiro:
Hi @hansemannn, I thought about, could be better to move to `Ti.UI.iOS` instead?

hansemannn:
Excellent idea! Now please restore the TiUIWindowProxy.m with the original one and place the docs in `apidoc/Titanium/UI/iOS/iOS.yml` (should be the one).

hansemannn:
Can you please pull from master? Cloning your PR locally includes old stuff like the manual scons-scripts.

hansemannn:
Approved!

PR #8170, opened 2016-07-29
[TIMOB-23701] (5_4_X) iOS: Fix storyboard-regression

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23701


cheekiatng:
CR and FT passed. Good.

PR #8177, opened 2016-07-31
iOS: Fix unbalanced USE_TI_UIIOSSYSTEMBUTTON macro

hansemannn:
The `SystemButton` is declared in the header-file, if the `USE_TI_UIIOSSYSTEMBUTTON` precompilation macro is defined. In this case, it was trying to compile it when `USE_TI_UIIOSSYSTEMBUTTONSTYLE` is defined, which is an unbalanced define-condition causing the build to fail for device- and dist-builds.


cheekiatng:
APPROVED!

PR #8185, opened 2016-08-01
[TIMOB-23638] Android: Unable to debug an application with newer V8

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23638

**Description:**

The should fix some of the issues seen with the Android V8 debugger on master SDK builds. There's still a fix that needs to be applied to Studio to have breakpoints actually get hit.

This fix is related to threading issues leading to hard crashes. Specifically we would very rarely get Android complaining about a JNIEnv\* mismatch that caused a hard crash (on emulator only). This should also fix crashes seen much more frequently around asking V8 to process the incoming debug messages (where it would typically crash doing GC).

### DIFF
--- a/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
+++ b/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
@@ -25,6 +25,7 @@ import android.os.Handler;
 import android.os.Looper;
 
 import org.appcelerator.kroll.common.Log;
+import org.appcelerator.kroll.common.TiMessenger;
 
 public final class JSDebugger
 {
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
Likely also fixes https://jira.appcelerator.org/browse/TIMOB-23628

ashcoding:
Code change looks good. I assume I have to wait for a recent Studio (after the PR) to see this working?

ashcoding:
Or can this be merged?

kolipakakondal:
LGTM. Verified with the SDK changes from this PR.

ashcoding:
Compiled the SDK for @kolipakakondal and tested it with the Studio PR as well.
**Merging**

sgtcoolguy:
Awesome, thanks @kolipakakondal and @ashcoding !

PR #8186, opened 2016-08-02
[TIMOB-23712] iOS: Stepper add getValue

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23712)

### DIFF
--- a/iphone/Classes/TiUIiOSStepperProxy.m
+++ b/iphone/Classes/TiUIiOSStepperProxy.m
@@ -20,7 +20,7 @@ USE_VIEW_FOR_CONTENT_HEIGHT
 
 -(NSNumber*)value
 {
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIiOSStepperProxy.m
```diff
@@ -18,7 +18,10 @@ -(NSString*)apiName
     return @"Ti.UI.iOS.Stepper";
 }
 
-
+-(NSNumber*)value
+{
+    return NUMINT([[(TiUIiOSStepper*)[self view]stepper] value]);
```
hansemannn: The view-method is double. Using this breaks the double-precision ... Use `NUMDOUBLE` instead. Also missing space between `]stepper`.

hansemannn:
Approved!

PR #8187, opened 2016-08-02
ioslib downgrade to 0.10.4

cb1kenobi:
Downgraded ioslib so that we can get back to node-ios-device 0.8.x.


cb1kenobi:
Wrong branch.

PR #8195, opened 2016-08-08
[TIMOB-23744] Android: Crash using Ti.Android.R values

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23744

**Description:**
Proxies using "interceptors" on the latest V8 were broken (i.e. The Ti.Android.R proxy). When internally adding our _properties object to the object it would trigger some callbacks to happen that we tried to get all the properties on the object before it was all initialized. I modified the ordering of when we wrap the JS object in C++ and when we define _properties. I also added in a null check to the interceptor C++ code, so that if we try to get a property handled by the interceptor and we haven't initialized the Java object in the JVM, then we'll return undefined for the time being.

This bug fix eliminates the crash seen before when simply adding a reference to Ti.Android.R.anim to your app.js file.

It'd be good to do more thorough testing to be sure that proxy works as intended, i.e. doesn't always give back undefined for values and gives us back values we expect...

### DIFF
--- a/android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
+++ b/android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
@@ -336,6 +336,12 @@ void ${className}::interceptor(Local<String> property, const v8::PropertyCallbac
 	}
 
 	jobject javaProxy = proxy->getJavaObject();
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Android.R

farfromrefug:
this fixes it for me

sgtcoolguy:
Awesome, thanks for testing it @farfromrefug !

PR #8196, opened 2016-08-08
[TIMOB-23745] Android: Crash using Math.random

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23745

**Description:**
Calling Math.random() in JS code, like so:

``` js
console.log(Math.random());
```

Would crash against Titanium SDK 6.0+ on Android.

The culprit was a local reference/instantiation of a V8 ArrayBufferAllocator. I needed to move the definition of the allocator up into the global scope so it wouldn't get GC'd in C++ when the V8Runtime.nativeInit finished.


farfromrefug:
awesome job can't wait to test it!

farfromrefug:
i can confirm this fixes it

sgtcoolguy:
Thanks for testing, @farfromrefug !

PR #8198, opened 2016-08-08
[TIMOB-23774] Team id is now required in the tiapp.xml when building …

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23774
- Team id is now required in the tiapp.xml when building an app with a watch extension and Xcode 8.
- Added build configuration settings for swift files in watch apps.
- Removed old watchos2 template in favor of existing watchos2-objc template.
- Set default min iOS SDK version to 8.0 if not explicitly set in the tiapp.xml and compiling using Xcode 8.
- Removed old maintainers from package.jsons.


cheekiatng:
APPROVED!

PR #8203, opened 2016-08-09
[TIMOB-23710] Fix aps-analytics.jar for Android

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23710
**Description:**
Use locally built APS analytics 2.0.5 jar from aps_sdk repo. Size looks correct. Should fix re-opened TIMOB-23710.

This isn't the same issue as TIMOB-23710, though that got tagged as the culprit. The root cause is a bad JAR checked in for the NATIVESDK-348 fix. I rebuilt the aps-analytics.jar locally (with 2.0.5 version) and checked it in to fix the issue.


ashcoding:
Tested it out with "com.appcelerator.apm" version 2.0.0. Works well. No issues.
**Approved**

PR #8208, opened 2016-08-10
[TIMOB-23753] Android: Crash when adding Ti.Network event listener

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23753

**Description:**
This JS code caused a hard crash:

``` javascript

Ti.Network.addEventListener('change', function() {
      console.log('test');
  })
```

I modified proxy method callbacks in C++ to be able to look up the prototype chain on "this" to get the JS object wrapping the Java proxy. In some cases we do some wrapping of APIs to be able to "lazily" load sub-types/methods, and this causes issues in C++ if we assume "this" always wraps the proxy. In some cases "this" is that JS lazy wrapper so we need to search up it's prototype chain to the real JS proxy object holding our Java object. I had already done the same thing for generated proxies in ProxyBindingV8.cpp.fm, but needed to add it manually for titanium::Proxy.cpp.

### DIFF
--- a/android/runtime/v8/src/native/Proxy.cpp
+++ b/android/runtime/v8/src/native/Proxy.cpp
@@ -246,7 +246,12 @@ void Proxy::hasListenersForEventType(const v8::FunctionCallbackInfo<v8::Value>& 
 		return;
 	}
 
-	Proxy* proxy = NativeObject::Unwrap<Proxy>(args.Holder());
+	Local<Object> holder = args.Holder();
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
cc @farfromrefug This PR should hopefully fix the issue you reported...

farfromrefug:
@sgtcoolguy will test it as soon as possible!
BTW shouldn't we make those 3 lines of code a function? not to duplicate it everywhere...

farfromrefug:
@sgtcoolguy it's fixed!

PR #8209, opened 2016-08-12
[TIMOB-23756] Improved app name and resource name blacklist so you ca…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23756

Improved app name and resource name blacklist so you can't name your app _CodeSignature.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -76,7 +76,11 @@ function iOSBuilder() {
 		'_codesignature',
 		'embedded.mobileprovision',
 		'info.plist',
-		'pkginfo'
+		'pkginfo',
+		'assets.car',
+		'modules',
+		'LaunchScreen.storyboardc',
+		'hyperloop'
 	];
 
 	this.graylistDirectories = [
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -1265,7 +1269,13 @@ iOSBuilder.prototype.validate = function (logger, config, cli) {
 
 		// make sure the app doesn't have any blacklisted directories or files in the Resources directory and warn about graylisted names
 		var platformsRegExp = /^(android|ios|iphone|ipad|mobileweb|blackberry|windows|tizen)$/;
-		this.blacklistDirectories.push(this.tiapp.name);
+
+		if (this.blacklistDirectories.indexOf(this.tiapp.name.toLowerCase()) !== -1 || this.tiapp.name.toLowerCase() === 'frameworks') {
```
hansemannn: Just out of interest, can't we just add `frameworks` to the blacklist as well?

cb1kenobi: I was unable to find evidence that `frameworks` was forbidden by Apple, so I graylist it and show a warning. We could blacklist it, but that would be a breaking change.

cheekiatng:
APPROVED!

PR #8230, opened 2016-08-15
[TIMOB-17467] Android: Add setRequestHeaders() to WebView

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17467

**Optional Description:**

```
var win = Ti.UI.createWindow();
var webView = Ti.UI.createWebView({
    requestHeaders: {
        "User-Foo": "FooBar",
        "foo": "bar",
        "other": "string"
    },
    url: "https://httpbin.org/headers"
});
var btn = Ti.UI.createButton({
    title: "Reload",
    bottom: 50
});

btn.addEventListener("click", function() {
    webView.reload();
});

var btnChange = Ti.UI.createButton({
    title: "Change Header",
    bottom: 0
});

btnChange.addEventListener("click", function() {
    var date = new Date().getTime();
    webView.setRequestHeaders({
        "User-Foo": "FooBar_" + date,
        "foo": "bar_" + date,
        "other": "string_" + date
    });
});

win.add(webView);
win.add(btnChange);
win.add(btn);
win.open();
```

PR #8239, opened 2016-08-17
[TIMOB-17467] iOS: Add setRequestHeaders() to WebView

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17467

Parity for #8230

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -675,6 +675,30 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 
 - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)request navigationType:(UIWebViewNavigationType)navigationType
 {
+	NSDictionary *newHeaders = [[self proxy] valueForKey:@"requestHeaders"];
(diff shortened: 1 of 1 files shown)


AngelkPetkov:
CR and FT **Approved!**

chreck:
Hi

I had several issues with this implementation. 

1. **A second POST request to the same address**: Because of the second loadUrlRequest call it could not send a second POST request. I mean the call `[self loadURLRequest:newRequest];` in the implementation. I made a fix for that see https://jira.appcelerator.org/browse/TIMOB-24329
But still this fix does not solve all request cycles propberly which can happen in a WebView itself.

2. **Own Request-Headers and Basic-Authentication-Headers**: I had a strange behaviour with including Basic Authentication and my own Request-Headers. It was difficult to determine the error but in the end I removed the second `[self loadURLRequest:newRequest];` by setting the headers and Basic Authentication in the loadURLRequest itself. See at https://github.com/chreck/titanium_mobile/commit/4d67118bebe74e9c9f008df1e40327ffb44feb8e (I hope the link works for you)

Best regards

Christoph

hansemannn:
Hey Christoph, can you file an own ticket with exactly those findings? That makes it easier to check the issue.

PR #8247, opened 2016-08-22
[TIMOB-23797] iOS10: Support Ti.UI.Clipboard.setItems including options

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23797


iphone/Classes/UIModule.m
```diff
@@ -624,9 +624,32 @@ -(NSNumber*)ATTRIBUTE_LINE_BREAK_BY_TRUNCATING_MIDDLE
 }
 #endif
 
+#ifdef USE_TI_UICLIPBOARD
+-(NSString*)PASTEBOARD_OPTION_LOCAL_ONLY
+{
+    if ([TiUtils isIOS10OrGreater]) {
+#if IS_XCODE_8
```
AngelkPetkov: There is a #if IS_XCODE_8 inside the isiOS10OrGreater method, so this is redundant

iphone/Classes/UIModule.m
```diff
@@ -624,9 +624,32 @@ -(NSNumber*)ATTRIBUTE_LINE_BREAK_BY_TRUNCATING_MIDDLE
...
+        return UIPasteboardOptionLocalOnly;
+#endif
+    } else {
+        return @"";
+    }
+}
+-(NSString*)PASTEBOARD_OPTION_EXPIRATION_DATE
+{
+    if ([TiUtils isIOS10OrGreater]) {
+#if IS_XCODE_8
```
AngelkPetkov: Same as above

iphone/Classes/UIModule.m
hansemannn: It still needs to be wrapped for Xcode < 8, otherwise the build fails.

PR #8251, opened 2016-08-22
[TIMOB-7406] iOS: Introduce userAgent setter for webView

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-7406?filter=-1)

**Note:** If testing through xcode you wont get the sdk version in the logs, it will work through studio.

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -488,6 +488,13 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 	[[self scrollview] setScrollsToTop:scrollsToTop];
 }
 
+- (void)setUserAgent_:(id)args
+{
+    ENSURE_TYPE(args, NSString);
+    NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:args], @"UserAgent", nil];
+    [[NSUserDefaults standardUserDefaults] registerDefaults:dict];
+}
+
 - (void)setUrl_:(id)args
 {
 	ignoreNextRequest = YES;
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -490,9 +490,16 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 
 - (void)setUserAgent_:(id)args
 {
-    ENSURE_TYPE(args, NSString);
-    NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:args], @"UserAgent", nil];
+    ENSURE_TYPE_OR_NIL(args, NSString);
+    NSString *userAgent = [TiUtils stringValue:args];
+    
+    if (!userAgent) {
+        userAgent = [[TiApp app] systemUserAgent];
+    }
+    
+    NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:userAgent, @"UserAgent", nil];
     [[NSUserDefaults standardUserDefaults] registerDefaults:dict];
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIWebView.m
```diff
@@ -488,6 +488,13 @@ -(void)setScrollsToTop_:(id)value
 	[[self scrollview] setScrollsToTop:scrollsToTop];
 }
 
+- (void)setUserAgent_:(id)args
+{
+    ENSURE_TYPE(args, NSString);
+    NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils stringValue:args], @"UserAgent", nil];
+    [[NSUserDefaults standardUserDefaults] registerDefaults:dict];
```
hansemannn: The user must be able to reset it by passing `null` to get back the old native behavior. That's what prevented us from doing it before.

iphone/Classes/TopTiModule.m
```diff
@@ -44,8 +44,15 @@ -(id)userAgent
...
-    [[TiApp app] setUserAgent:[TiUtils stringValue:value]];
+    ENSURE_TYPE_OR_NIL(value, NSString);
+    NSString *userAgent = [TiUtils stringValue:value];
+    
+    if (!userAgent) {
+        [[TiApp app] setUserAgent:[[TiApp app] systemUserAgent]];
+        return;
+    }
+    
+    [[TiApp app] setUserAgent:userAgent];
```
hansemannn: Replace line 50-55 with:
```objc
[[TiApp app] setUserAgent:(userAgent ?: [[TiApp app] systemUserAgent])];
```
That should do the same.

iphone/Classes/TiUIWebView.m
```diff
@@ -488,6 +488,20 @@ -(void)setScrollsToTop_:(id)value
 	[[self scrollview] setScrollsToTop:scrollsToTop];
 }
 
+- (void)setUserAgent_:(id)args
+{
+    ENSURE_TYPE_OR_NIL(args, NSString);
+    NSString *userAgent = [TiUtils stringValue:args];
```
hansemannn: Rename `args` to `value`, since the value is no dict or array.

iphone/Classes/TiUIWebView.m
```diff
@@ -488,6 +488,20 @@ -(void)setScrollsToTop_:(id)value
...
+    
+    if (!userAgent) {
+        userAgent = [[TiApp app] systemUserAgent];
+    }
+    
+    NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:userAgent, @"UserAgent", nil];
+    [[NSUserDefaults standardUserDefaults] registerDefaults:dict];
```
hansemannn: I'm still not sure if we can do that, because it might mess-up different webviews. Let's say you set the property in webview 1 and don't set it in webview 2, webview 2 will most likely have the same user-agent as webview 1, because it is referenced globally. Correct?

AngelkPetkov: Hmm, yeah that's correct. So I propose we move userAgent as a property of when setting the URL. So in other words when setting  the URL we can pass in a single URL or a dict with URL and userAgent. The reason behind this is because to set the userAgent we have to reload the request.  Setting the userAgent would look something like this: 

```
NSURL *url = [NSURL URLWithString:@"someAdress.com"];
NSMutableURLRequest *requestObj = [NSMutableURLRequest requestWithURL:url];
[requestObj setValue:@"Foobar/1.0" forHTTPHeaderField:@"User_Agent"];
[webView loadRequest: requestObj]'; 
```

AngelkPetkov:
Closing PR, after discussing with @hansemannn  it would be best if we add the userAgent as a global property to be set through the TopTI as oppose as independent user-Agents for each webView

PR #8252, opened 2016-08-23
[TIMOB-23798] iOS: Fix clearData nil-value argument

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23798

Additional infos:
- Extracts the argument from the arguments-array properly
- Sets the correct mime-type for resetting the data to `NSData` (which is `application/octet-stream`)


AngelkPetkov:
**Approved!**

PR #8259, opened 2016-08-24
(5.5.X)Fix wrong ifdef statmenet

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-23824)

### DIFF
--- a/iphone/Classes/TiUIiOSProxy.m
+++ b/iphone/Classes/TiUIiOSProxy.m
@@ -366,7 +366,7 @@ RELEASE_TO_NIL(x); \
 }
 #endif
 
-#ifdef USE_TI_UIIOSANIMATIONSTYLE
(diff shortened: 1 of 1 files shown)


hansemannn:
Needs approval from @eric34  and @cheekiatng before. It basically fixes an issue where the ProgressBarStyle constants could not found unless you define an AnimationStyle constant. @AngelkPetkov please also file a ticket for that asap and link it here.

AngelkPetkov:
Will file a ticket, but its not the only issue it will cause the app to crash too, if you use animationStyle as it would define the progressBar in the m file but the actuall file will not be defined and will crash with null pointer.

cheekiatng:
we want this in 5.5.0. we want a ticket too to track. do asap.

hansemannn:
Approved!

PR #8268, opened 2016-08-25
[TIMOB-23827] Fixed module id validation to allow underscores when cr…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23827

Fixed module id validation to allow underscores when creating an Android-only module. Also fixed issue where ioslib.detect() was erroring out when it didn't find a compatible version of node-ios-device, but the error wasn't properly handled.


feons:
LGTM!

PR #8273, opened 2016-08-26
[TIMOB-23830] Fixed Android modules that have name that begins with a…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23830

Fixed Android modules that have name that begins with a number.


hansemannn:
PR rejected by @cb1kenobi via JIRA.

PR #8287, opened 2016-08-30
[TIMOB-20206] Android: TiCompositeLayout's viewSorter fix

ashcoding:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20206

As mentioned in the ticket, the updates done here https://android-review.googlesource.com/#/c/257511/1/ojluni/src/main/java/java/util/TreeMap.java will cause an issue with our current code beyond Android N.


garymathews:
:+1:

PR #8292, opened 2016-08-31
[TIMOB-23798] (6_0_X) iOS: Fix clearData nil-value argument

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23798


hansemannn:
Approved!

PR #8296, opened 2016-08-31
[TIMOB-23851] Fixed iOS module lipo call and arch verification.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23851


feons:
LGTM!

PR #8309, opened 2016-09-03
[TIMOB-23834] Windows: Skip mobileweb init() when not targeting mobileweb

garymathews:
- Skip mobileweb `init` when targeting other platforms
  - Prevents `windowslib.detect()` being called when targeting any platform on Windows, which causes a significant delay; especially when a Windows device is connected.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23834)


infosia:
Looks good 👍

PR #8315, opened 2016-09-06
[TIMOB-23871] iOS: Safari/WebInspector Debugger crashes app when hitting breakpoint

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23871

**Description:**
To reproduce:
- launch a Titanium app with use-jscore-framework set to true and run-on-main-thread. (I used the standard template and just added a click listener for the label on each tab's content)
- Open Safari > Develop > Simulator > JSContext.
- Set a breakpoint in a callback function (i.e. on the label click event handler or a setInterval callback).
- When it hits the breakpoint, it crashes.

The way we set up on particular object in the JS world, our code doesn't handle it. Specifically we hang a "Kroll" object off the global and have it wrap a single instance of our KrollContext class in Obj-C. But all our code for handling wrapped JS objects assumes it's always a KrollObject inside. In this case, it's a KrollContext. So our Kroll*Property functions in KrollObject need to check for and handle this special object.

### DIFF
--- a/iphone/Classes/KrollObject.m
+++ b/iphone/Classes/KrollObject.m
@@ -99,7 +99,7 @@ TiValueRef ConvertIdTiValue(KrollContext *context, id obj)
 void KrollFinalizer(TiObjectRef ref)
 {
     waitForMemoryPanicCleared();
-	id o = (KrollObject*)TiObjectGetPrivate(ref);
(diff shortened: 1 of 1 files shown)


cheekiatng:
CR and FT passed. APPROVED.

PR #8317, opened 2016-09-06
[TIMOB-23865] Updated to ioslib@0.13.3.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23865

### DIFF
--- a/node_modules/ioslib/lib/simulator.js
+++ b/node_modules/ioslib/lib/simulator.js
@@ -95,11 +95,11 @@ const devicePairCompatiblity = exports.devicePairCompatiblity = {
 	},
 	'8.x': {             // Xcode 8.x
 		'9.x': {         // iOS 9.x
-			'2.x': true, // watchOS 2.x
+			'2.2': true, // watchOS 2.2
 			'3.x': true  // watchOS 3.x
 		},
 		'10.x': {        // iOS 10.x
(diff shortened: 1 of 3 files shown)


node_modules/ioslib/lib/simulator.js
```diff
@@ -95,11 +95,11 @@ const devicePairCompatiblity = exports.devicePairCompatiblity = {
 	},
 	'8.x': {             // Xcode 8.x
 		'9.x': {         // iOS 9.x
-			'2.x': true, // watchOS 2.x
+			'2.2': true, // watchOS 2.2
```
hansemannn: Does this also catch 2.1 and 2.2.2 (latest watchOS2 version)?

cb1kenobi: I was unsuccessful in pairing a watchOS 2.0 or 2.1 simulator with any iOS simulator with Xcode 8 Beta 6. Furthermore, there is no watchOS 2.2.2 simulator, only 2.2 and since 2.2 is the last 2.x watchOS simulator runtime, that means we can safely say `2.2`.

hansemannn: Then we're good. Thanks for clearing!

hansemannn:
Approved!

PR #8324, opened 2016-09-07
[TIMOB-23864] CLI: Build process does not exit if using an incompatib…

feons:
…le module based on apiversion 6_0_X

https://jira.appcelerator.org/browse/TIMOB-23864


cb1kenobi:
**APPROVED**

PR #8325, opened 2016-09-07
[TIMOB-23864] CLI: Build process does not exit if using an incompatib…

feons:
…le module based on apiversion

https://jira.appcelerator.org/browse/TIMOB-23864
1. Bump to using node-appc@0.2.37
2. Update `/node_modules/titanium-sdk/lib/builder.js`


cb1kenobi:
**APPROVED**

PR #8333, opened 2016-09-09
[TIMOB-23880] Default classic app template does not show tabs on Windows

infosia:
[TIMOB-23880](https://jira.appcelerator.org/browse/TIMOB-23880)

When building the default app template for Windows platform the tabs are not visible, this is because on Windows the default text color is white and the background color is set to white.

From comment of the app template:

``` javascript
 /**
  * Set a global background-color (used for iOS)
  */
```

So if this is only for iOS, we should keep this effective only on iOS.

### DIFF
--- a/templates/app/default/template/Resources/app.js
+++ b/templates/app/default/template/Resources/app.js
@@ -1,7 +1,9 @@
 /**
  * Set a global background-color (used for iOS)
  */
-Ti.UI.setBackgroundColor("#fff");
+if (Ti.Platform.osname === 'iphone' || Ti.Platform.osname === 'ipad') {
+    Ti.UI.setBackgroundColor("#fff");
+}
 
 /**
  * Create a new `Ti.UI.TabGroup`.


hansemannn:
Thanks Kota! What is with Android? I thought it looked pretty good there as well, so we might want to explicitly say something like:
```js
// Global background-color is only used on iOS and Android
if (['windowsphone', 'windowsstore'].indexOf(Ti.Platform.osname) == -1) {
    Ti.UI.setBackgroundColor("#fff");
}
```
Or am I missing something? :-) Thanks!

hansemannn:
@infosia ?

sgtcoolguy:
@hansemannn The code check and the comment seem inconsistent at first glance in your code above. You're assuming that windows is the only other platform (ever), which is what got us this issue in the first place. If we want it to be iOS/Android only, make the check for iOS and Android, please.

```js
/**
  * Set a global background-color of white (used for iOS and Android). Windows retains default background (black).
  */
if (['iphone', 'ipad', 'android'].indexOf(Ti.Platform.osname) != -1) {
    Ti.UI.setBackgroundColor("#fff");
}
```

But, this is very ugly and pretty much shoots a hole in our "cross-platform" story right off the bat on the first line of our default template. Do we **need** to even set a global background color? What would it default to on iOS and Android if we didn't? Why not just use the system/OS defaults?

The issue we have here is that we explicitly set a white bg color and Windows defaults to white fg on black bg, so setting this makes white on white. Wouldn't iOS use white bg by default anyways? Would Android?

sgtcoolguy:
Looks like Android would default to a light grey/off-white using the light theme by default: https://material.google.com/style/color.html#color-themes

hansemannn:
The default view-color on iOS is white, I guess Android is black then? If so, we can just accept this PR and go for it. If we leave it black for iOS, the navigation-bar (top) und tabbar (bottom) will have a black translucency.

sgtcoolguy:
I'll take a quick look and see what it looks like without the global bg set on iOS and Android and take screenshots.

sgtcoolguy:
So I know this is totally off on a tangent now - but @hansemannn do we have APIs for querying the default theme colors for things? We could just eliminate setting the bg and fg colors in this template to use defaults and look like the natural phone OS, which probably is best and would make the code smaller (though wouldn't give the users a quick example of how to set the bg/fg).

But I'm guessing it'd be useful to be able to look up the colors in the active theme to use as baselines? Maybe you want to change a color temporarily but then change back to the default, would setting null for the value do that?

sgtcoolguy:
Ok, so on iOS 10.0, iPhone 7 Plus simulator, setting global bg color or not, it looks the exact same.

screen shot 2016-10-06 at 10 33 55 am

sgtcoolguy:
Same on Android:
screen shot 2016-10-06 at 10 37 17 am

sgtcoolguy:
What if we eliminate setting fg/bg on anything in the sample?

Android:
screen shot 2016-10-06 at 10 38 26 am

iOS: 
No difference from above.

sgtcoolguy:
Long story short - we should just ditch the global bg color setting entirely.

sgtcoolguy:
Closing in favor of  #8479

infosia:
👍

PR #8335, opened 2016-09-09
[TICKET] Correct check if proxy is set

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to a string with the content `'undefined'`


hansemannn:
Hey @mathiasrw, thanks for the PR! Can you please file a [JIRA-ticket](https://jira.appcelerator.org) with some additional information and a test-case to validate your fix? Thanks you! 🚀

mathiasrw:
Comparing the result of `typeof` with a variable named undefined will always give false - no matter what you get the type off. 

I have no idea how to do a propper test, and feel too lazy to make the whole jira signup to add chars to your code. 

I have no idea how to do a propper test of your code, and feel too lazy to make the whole jira signup just to add 2 chars to your code. 

Please feel free to implement the changes

PR #8342, opened 2016-09-10
[TIMOB-23891] iOS: Support iOS 10 Haptic Engine API on the iPhone 7

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23891


cheekiatng:
CR AND FT PASSED. APPROVED!

PR #8346, opened 2016-09-11
[TIMOB-23890] iOS: Post notification in openURL-delegate, use newer delegate for iOS9+

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23890

Required for https://github.com/appcelerator-modules/ti.facebook/pull/63

### DIFF
--- a/apidoc/Titanium/App/iOS/iOS.yml
+++ b/apidoc/Titanium/App/iOS/iOS.yml
@@ -1124,6 +1124,18 @@ events:
     osver: {ios: {min: "9.0"}}
     since: '5.1.0'
 
+  - name: handleurl
+    summary: Fired when a new URL is handled by the application.
+    description: |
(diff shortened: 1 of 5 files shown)


apidoc/Titanium/App/iOS/iOS.yml
```diff
@@ -1124,6 +1124,18 @@ events:
     osver: {ios: {min: "9.0"}}
     since: '5.1.0'
 
+  - name: handleurl
```
hansemannn: @cheekiatng In case you ask why we need this here: It is triggered by https://jira.appcelerator.org/browse/TIMOB-23884 that is caused by an open Apple issue. It is both workaround until Apple fixes it as well as the future-way to receive updates from URL's that get handled by the application.

cheekiatng:
CR AND FT PASSED. APPROVED!

PR #8371, opened 2016-09-13
[TIMOB-20506] iOS: Fix null-value in Ti.App.Properties.setList method

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20506


sgtcoolguy:
It'd be good to have a unit test for this fix. Our current suite for Ti.App.Properties is here: https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.app.properties.test.js

Please clone/pull that repo locally, copy the file into titanium_mobile/tests/Resources/ti.app.properties.test.js and edit it to add a test (or tests) for this case. Then add it to your PR and the build can run and confirm the fix.

hansemannn:
Added a unit-test in https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/4

cheekiatng:
CR and FT passed.

PR #8379, opened 2016-09-14
[TIMOB-14763] iOS: Fix titleid/messageid in Ti.UI.AlertDialog

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-14763


cheekiatng:
CR and FT Passed. APPROVED!

PR #8382, opened 2016-09-14
[TIMOB-23903] Build against V8 5.7.492.71

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-23903

This updates Android's runtime to build against V8 5.7.492.71

V8 introduced yet more breaking API changes and I was unable to keep that from propagating out to our native modules. So this is a BREAKING NATIVE MODULE CHANGE!

Specifically, V8 removed WeakCallbackData and moved to WeakCallbackInfo. That was a minor update and I think could have avoided breaking module compatibility.

The larger change they made was to restrict values you can set as properties on a Template. We were setting actual Function objects on the prototype and instance templates for the types - we needed to change to setting the FunctionTemplate itself, not the function it generated. Secondly, it meant we could set an External wrapping the class (java class in JNI) that the proxy represented. This was used to instantiate instances of the Java class itself when constructing a proxy object in JS-world. To workaround this, instead of explicitly setting that External as the value for **javaClass**, I now set a  "native data property" for that property name with a getter callback. This means it's a property that always goes straight to the native callback (can't be overridden in JS). And the data value we set with it gets passed to every callback invocation, so we use that as a sneaky "hack" to pass along the External wrapping the java class. The callback method simply grabs that value and returns it.

Relates to appcelerator/v8_titanium#12


android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
```diff
@@ -54,13 +54,26 @@ void ${className}::bindProxy(Local<Object> exports, Local<Context> context)
 	Isolate* isolate = context->GetIsolate();
 
 	Local<FunctionTemplate> pt = getProxyTemplate(isolate);
-	Local<Function> proxyConstructor = pt->GetFunction(context).ToLocalChecked();
+
+	v8::TryCatch tryCatch(isolate);
+	Local<Function> constructor;
+	MaybeLocal<Function> maybeConstructor = pt->GetFunction(context);
+	if (!maybeConstructor.ToLocal(&constructor)) {
+		V8Util::fatalException(isolate, tryCatch);
```
garymathews:
```Java
titanium::V8Util::fatalException(isolate, tryCatch);
```

android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
```diff
@@ -54,13 +54,26 @@ void ${className}::bindProxy(Local<Object> exports, Local<Context> context)
...
+		return;
+	}
+
 	Local<String> nameSymbol = NEW_SYMBOL(isolate, "${proxyAttrs.name}"); // use symbol over string for efficiency
 	<#if isModule>
-	Local<Object> moduleInstance = proxyConstructor->NewInstance(context).ToLocalChecked();
+	MaybeLocal<Object> maybeInstance = constructor->NewInstance(context);
+	Local<Object> moduleInstance;
+	if (!maybeInstance.ToLocal(&moduleInstance)) {
+		V8Util::fatalException(isolate, tryCatch);
```
garymathews:
```Java
titanium::V8Util::fatalException(isolate, tryCatch);
```

garymathews (changes requested):

garymathews (approved):
Awesome! 👍 

CR: **PASS**
FT: **PASS**

Although we may need to use `5.7.54.0`.

###### NOTE: These changes are also valid for `5.9`

sgtcoolguy:
@garymathews Given that they've moved away from the older debug protocol, I think e should look into simply updating studio's debugger for Android to work with the new inspector protocol so we can update V8 to latest. It shouldn't be *too* bad since I already had to write the inspector protocol plus extra layers for iOS/JSCore.

Here's the studio ticket for updating the Android debugger (again!): https://jira.appcelerator.org/browse/TISTUD-8750

hansemannn:
@sgtcoolguy So it's no breaking change? Otherwise, we cannot cause all native modules on Github and our marketplace to be recompiled again, especially not in a non-major release version :(.

PR #8384, opened 2016-09-15
[TIDOC-2611] Since 5.5.0.GA iOS platform return iOS not iPhone OS

HazemKhaled:
**JIRA:** https://jira.appcelerator.org/browse/AC-4460

### DIFF
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -208,6 +208,8 @@ properties:
     summary: |
         Name of the platform. Returns `android` for the Android platform, `iPhone OS` for the iOS
         platform (iPhone, iPad, or iPod Touch), `windows` for the Windows platform, and `mobileweb` for the Mobile Web platform.
+        
+        Since 5.5.0.GA return 'iOS' for for the iOS platform (iPhone, iPad, or iPod Touch).
     type: String
     permission: read-only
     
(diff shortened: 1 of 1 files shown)


apidoc/Titanium/Platform/Platform.yml
```diff
@@ -208,6 +208,8 @@ properties:
     summary: |
         Name of the platform. Returns `android` for the Android platform, `iPhone OS` for the iOS
         platform (iPhone, iPad, or iPod Touch), `windows` for the Windows platform, and `mobileweb` for the Mobile Web platform.
+        
+        Since 5.5.0.GA return 'iOS' for for the iOS platform (iPhone, iPad, or iPod Touch).
```
hansemannn: I would rephrase it to:

> Since 5.5.0.GA and Xcode 8, this property returns `iOS` for for the iOS platform (iPhone, iPad, or iPod Touch).

Are you fine with that? And good catch btw!

HazemKhaled: done 👍

HazemKhaled: I just noticed Ti.Platform.name return `iPhone OS` with Xcode 8 and iOS simulator 9.3, so it has to be.

```
On iOS 10+ return 'iOS' for for the iOS platform (iPhone, iPad, or iPod Touch).
```

PR #8408, opened 2016-09-21
[TIMOB-23924] Move internal focus-call to selector

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23924

Temporary solution until Apple fixes their private API-detection, since `focus:` is no private selector but our own `TiKeyboardFocusableView` delegate.


cheekiatng:
APPROVED!

steipete:
This can be reverted; it seems that this has been fixed inside Apple. We faced the same problem and also opened up a rdar about it:

hansemannn:
We also didn't receive any new reports since then. @cheekiatng we can remove it for master again, but let's wait a few more weeks for Apple to confirm this.

PR #8409, opened 2016-09-21
[TIMOB-23924] (6_0_X) Move internal focus-call to selector

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23924

Temporary solution until Apple fixes their private API-detection, since `focus:` is no private selector but our own `TiKeyboardFocusableView` delegate.


cheekiatng:
APPROVED!

PR #8411, opened 2016-09-21
[TIMOB-23901] Fix IdleNotificationDeadline() duration

garymathews:
- This is the root cause of the splash screen issue, the `nativeDispose()` call hangs the `KrollRuntimeThread` which prevents the message handler from processing the `MSG_INIT` to re-initialise `KrollRuntime`
- This is due to an incorrect `IdleNotificationDeadline()` duration
###### TEST CASE
1. Create and launch any Titanium Android application
2. Press the back button to close the app
3. Launch the app again
###### EXPECTED
- Application should resume
###### ACTUAL
- Application hangs on splash screen

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23901)

### DIFF
--- a/android/runtime/v8/src/native/V8Runtime.cpp
+++ b/android/runtime/v8/src/native/V8Runtime.cpp
@@ -419,7 +419,7 @@ JNIEXPORT void JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nativeDi
 	// idle event in V8Runtime.java), we can't count on that running anymore at this point.
 	// So as our last act, run IdleNotification until it returns true so we can clean up all
 	// the stuff we just released references for above.
-	while (!V8Runtime::v8_isolate->IdleNotificationDeadline((V8Runtime::platform->MonotonicallyIncreasingTime() * static_cast<double>(1000)) + 100.0));
+	while (!V8Runtime::v8_isolate->IdleNotificationDeadline(V8Runtime::platform->MonotonicallyIncreasingTime() + 1.0));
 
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
ok to test

sgtcoolguy:
add to whitelist

sgtcoolguy:
looks good to me.

sgtcoolguy:
@garymathews If this one is wrong, is this one too? https://github.com/garymathews/titanium_mobile/blob/aded055ba50c95732f1c19711243ff22c65b0431/android/runtime/v8/src/native/V8Runtime.cpp#L322

sgtcoolguy:
@ashcoding @garymathews I guess maybe the old V8 integration had effectively the same bug?

It seemed unlikely, but maybe it's not? 

These seem relevant: 
- https://github.com/lloyd/node-memwatch/pull/60/files#diff-69f2d078b3741241d21a4c6dc64b61cfR245
- http://stackoverflow.com/questions/18701003/v8idlenotification-never-returns-true

They moved from:
```c++
while(!V8::IdleNotification()) {};
```
to:
```c++
while(!V8::IdleNotification(500)) {};
V8::LowMemoryNotification();
```

Though, I honestly don't think there's much difference there. The first is actually equivalent to:
```cpp
while(!V8::IdleNotification(1000)) {};
```

I mean that code could _possibly_ just loop forever and not return true - but I don't really know.

ashcoding:
@garymathews Tested this PR by building master branch and using Nexus 6 with Android 6.0 running. Still facing the issue. I'll add comments on the JIRA ticket to ask for QE to help test this.

PR #8418, opened 2016-09-22
[TIMOB-23908] (5_5_X) Fixed APS environment entitlements. Removed ExternalAcc…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23908

Fixed APS environment entitlements. Removed ExternalAccessory.framework when building an app that does not contain a watch app.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -3413,7 +3413,9 @@ iOSBuilder.prototype.writeEntitlementsPlist = function writeEntitlementsPlist() 
 		if (!plist.hasOwnProperty('application-identifier')) {
 			plist['application-identifier'] = pp.appPrefix + '.' + this.tiapp.id;
 		}
-		plist['aps-environment'] = this.target === 'dist-appstore' || this.target === 'dist-adhoc' ? 'production' : 'development';
+		if (pp.apsEnvironment) {
+			plist['aps-environment'] = this.target === 'dist-appstore' || this.target === 'dist-adhoc' ? 'production' : 'development';
+		}
 		if (this.target === 'dist-appstore' && !plist.hasOwnProperty('beta-reports-active')) {
 			plist['beta-reports-active'] = true;
 		}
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -3403,18 +3410,16 @@ iOSBuilder.prototype.writeEntitlementsPlist = function writeEntitlementsPlist()
 	// if we have a provisioning profile, make sure some entitlement settings are correct set
 	if (pp) {
 		// attempt to customize it by reading provisioning profile
+		if (!plist.hasOwnProperty('application-identifier')) {
+			plist['application-identifier'] = pp.appPrefix + '.' + this.tiapp.id;
+		}
+		plist['aps-environment'] = this.target === 'dist-appstore' || this.target === 'dist-adhoc' ? 'production' : 'development';
```
hansemannn: Should we really set the `aps-environment` key although mit might not be inside the provisioning profile? I liked the below check.

iphone/cli/commands/_build.js
```diff
@@ -2626,7 +2632,7 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 						child.comment = '"' + appName + '.app"';
 					} else if (child.comment === 'Titanium.entitlements') {
 						child.comment = '"' + appName + '.entitlements"';
-					} else if (child.comment === 'LaunchScreen.storyboard' && appc.version.lt(this.xcodeEnv.version, '7.0.0')) {
+					} else if ((child.comment === 'LaunchScreen.storyboard' && appc.version.lt(this.xcodeEnv.version, '7.0.0')) || child.comment === 'ExternalAccessory.framework') {
```
hansemannn: Not sure if that the `ExternalAccessory` change is related to this ticket. Let's not mix too many changes together if possible.

cb1kenobi: It was causing the Wireless Accessory Configuration capability to be enabled. I'm simply removing it if there's no watch app.

iphone/cli/commands/_build.js
cb1kenobi: OK, so we could do this: 

```javascript
if (pp.apsEnvironment) {
    plist['aps-environment'] = this.target === 'dist-appstore' || this.target === 'dist-adhoc' ? 'production' : 'development';
}
```

If they use APS, then we set it, but we set it to values based on the target, not the value in the provisioning profile.

hansemannn:
Approved!

PR #8419, opened 2016-09-22
[TIMOB-23939] Fixed bug when building an iOS module when the cwd is n…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23939

Fixed bug when building an iOS module when the cwd is not the iOS module directory. Other minor improvements to the iOS module build.


feons:
**APPROVE**

PR #8421, opened 2016-09-22
[TIMOB-23901] Disable run-on-main-thread

garymathews:
- Even with https://github.com/appcelerator/titanium_mobile/pull/8411, `run-on-main-thread` needs to be disabled. There seems to be an issue running everything on the main thread. @sgtcoolguy Is there a reason why `KrollRuntime` needed to be executed on the main thread?
- Fixed another occurrence of `IdleNotificationDeadline()`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23901)


sgtcoolguy:
@garymathews Yes, there was - we need to run on main thread for Android hyperloop to work.

It sounds like we're just doing something wrong on app "shutdown" where it never comes back up. It'd be good to run under a debugger or with a ton of debug logs to see what exactly is happening here.

garymathews:
@sgtcoolguy Okay, I'll figure out what's happening 👍

PR #8428, opened 2016-09-23
[TIMOB-23628] Android: Unable to debug an application with run-on-main-thread set to true

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23628

**Description:**
These are bug fixes for the debugger agent I had to add to our Android SDK wen we updated V8. **Please note that merging this does not mean we should resolve the ticket above**

The notable changes are:
- I fixed up how we parse the incoming messages from Studio (or other client) debuggers. We could end up in a state where we mangled the messages and caused an OutOfMemoryException.
- I forced a debug breakpoint at the very beginning of the app's lifecycle. This is so that we can wait until the debugger connects and does it's setup (like sending over all the breakpoints) before we launch into app.js - otherwise we couldn't ever hit "early" breakpoints in the app's execution. Now we can hit a breakpoint as early as the first line of app.js (or technically even in our own bootstrapping code).

### DIFF
--- a/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
+++ b/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Scanner;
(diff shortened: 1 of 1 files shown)


garymathews:
CR: **PASS**
FR: **PASS**

👍

PR #8432, opened 2016-09-26
[TIMOB-23934] iOS: Fix first-child from not receiving click events of it's parent

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23934


cheekiatng:
CR and FT Passed. APPROVED!

PR #8426, opened 2016-09-23
[TIMOB-23901] Android: Fix KrollRuntime init on main thread

garymathews:
- Fix initialisation of `KrollRuntime` on main thread when resuming application
- Fix `IdleNotificationDeadline()` duration in `nativeIdle()`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23901)

PR #8433, opened 2016-09-26
[TIMOB-20459][TIMOB-20490][TIMOB-20502] Fix Android intent issues

garymathews:
- [TIMOB-20459](https://jira.appcelerator.org/browse/TIMOB-20459): Fire `newintent` event when the app is already active and receives a new intent
- [TIMOB-20490](https://jira.appcelerator.org/browse/TIMOB-20490): Respond to intents with both `onIntent` and `newintent`
- [TIMOB-20502](https://jira.appcelerator.org/browse/TIMOB-20502): Allow alloy applications to initialise global variables etc... before launching the intent activity
###### TEST CASE
- Project: [com.alloy.zip](https://jira.appcelerator.org/secure/attachment/60328/com.alloy.zip)
  - **must retain `tiapp.xml` use `appc new --import`**
1. Create two shortcuts one `WITH TEXT` checked and one without
   - You should see two shortcuts `ALLOY` and `ALLOY TEXT` on your homescreen
2. Press the home button, keeping the application in the background
3. Press the `ALLOY` shortcut, you should see a **RED** screen with `GLOBAL TEXT` in the centre
   - This means the `sample.js` activity has launched and accessed the `Alloy.Globals.text` property
4. Press the back button, you should be at the `CREATE SHORTCUT` screen
5. Press the home button again, keeping the application in the background
6. Click the `ALLOY TEXT` shortcut, you should see an **ORANGE** screen with `GLOBAL TEXT` in the centre
   - This means the `newintent` event has been fired
7. Close the application so it is not running in the background
8. Press the `ALLOY` shortcut, again you should see a red screen with `GLOBAL TEXT` in the centre
   - This means the `sample.js` activity has launched as the accessed the `Alloy.Globals.text` property even though the root activity was not available in the background like before
9. Close the application so it is not running in the background
10. Press `ALLOY TEXT`, you should see a **GREEN** background with `GLOBAL TEXT` in the centre
    - This means the `onIntent` event has been fired
###### NOTES
- **Requires https://github.com/appcelerator/titanium_mobile/pull/8426**
- Logs will provide `EXTRA_TEXT` string

```
onIntent: received intent
TEXT: INTENT EXTRA_TEXT
```

### DIFF
--- a/android/modules/android/src/java/ti/modules/titanium/android/TiJSActivity.java
+++ b/android/modules/android/src/java/ti/modules/titanium/android/TiJSActivity.java
@@ -37,6 +37,18 @@ public abstract class TiJSActivity extends TiLaunchActivity
 	}
 
 	@Override
+ 	protected void onResume()
+ 	{
+ 		super.onResume();
+ 		
+ 		// launched alloy activity from intent
+ 		// finish to prevent redundant activity
+ 		if (this.alloyIntent) {
+ 			this.finish();
+ 		}
+ 	}
+
+	@Override
 	public String getUrl()
 	{
 		if (url == null) {
--- a/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
+++ b/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
@@ -105,6 +105,22 @@ public class KrollAssetHelper
 		return null;
 	}
 
+	public static boolean assetExists(String path) {
+		if (assetCrypt != null) {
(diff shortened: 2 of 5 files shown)


sgtcoolguy:
I tried the steps above on a recent master SDK build and they failed in a few ways:
- At step 4, pressing back went back to an essentially empty window in the app, rather than the create shortcut view.
- Step 10 gave me red background (I think)?

I then re-tried with the changes in the PR and all the steps above worked as expected.

Given the difference, I'm going to assume this PR is OK to merge.

PR #8438, opened 2016-09-28
Update deprecation msg with ListView one

HazemKhaled:
**JIRA:** Soon

This msg doesn't make sense

```
[WARN]  Ti.UI.iPhone.ListViewSeparatorStyle.NONE DEPRECATED in 5.2.0, in favor of Ti.UI.TABLE_VIEW_SEPARATOR_STYLE_NONE
```

Updated with

```
[WARN]  Ti.UI.iPhone.ListViewSeparatorStyle.NONE DEPRECATED in 5.2.0, in favor of Ti.UI.iOS.ListViewSeparatorStyle.NONE
```

But i think Ti.UI.iOS.ListViewSeparatorStyle missing

### DIFF
--- a/iphone/Classes/TiUIiPhoneListViewSeparatorStyleProxy.m
+++ b/iphone/Classes/TiUIiPhoneListViewSeparatorStyleProxy.m
@@ -18,13 +18,13 @@
 
 -(NSNumber*)SINGLE_LINE
 {
-    DEPRECATED_REPLACED(@"UI.iPhone.ListViewSeparatorStyle.SINGLE_LINE", @"5.2.0", @"UI.TABLE_VIEW_SEPARATOR_STYLE_SINGLE_LINE");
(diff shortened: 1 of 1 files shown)


nuno:
Nice @HazemKhaled !

hansemannn:
I think it was moved intentionally, because Android (we added it for parity) does not have the namespace-based constants. And since they internally work the same for iOS and Android, it was decided to name it as it is.

HazemKhaled:
@hansemannn Not bug, it just feel it's more organized to use Ti.UI.iOS.ListView* instead of TableView one, I know internally same value .. just more organized

hansemannn:
Same here. But this PR does not add this behavior, so we can't merge it. It may be an improvement-ticket, but then we have the android issue again and have to deprecate the API once again. Guess the developers are fine with just having an API that works consistently, instead of jumping between he names every second release. Ok?

PR #8458, opened 2016-10-02
[TIMOB-23974] Android: Add property to release space when hiding an object

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23974

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/UIModule.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/UIModule.java
@@ -178,6 +178,9 @@ public class UIModule extends KrollModule implements Handler.Callback
 	@Kroll.constant public static final int INPUT_TYPE_CLASS_NUMBER = InputType.TYPE_CLASS_NUMBER;
 	@Kroll.constant public static final int INPUT_TYPE_CLASS_TEXT = InputType.TYPE_CLASS_TEXT;
 
+	@Kroll.constant public static final int VIEW_GONE = View.GONE;
+	@Kroll.constant public static final int VIEW_INVISIBLE = View.INVISIBLE;
+
 	protected static final int MSG_SET_BACKGROUND_COLOR = KrollProxy.MSG_LAST_ID + 100;
 	protected static final int MSG_SET_BACKGROUND_IMAGE = KrollProxy.MSG_LAST_ID + 101;
 	protected static final int MSG_LAST_ID = MSG_SET_BACKGROUND_IMAGE;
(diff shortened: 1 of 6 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -1447,6 +1447,11 @@
 	/**
 	 * @module.api
 	 */
+	public static final String PROPERTY_FREE_SPACE = "freeSpace";
```
hansemannn: I don't like the property-name. What about a more "boolean-like" property, e.g. `showSpaceWhenHidden` or `truncateSpace`? Or even expose the used constant directly and introduce a `hiddenBahavior` property. That keeps it more flexible, allows more configuration and sounds more native. Thanks!

apidoc/Titanium/UI/View.yml
```diff
@@ -1292,6 +1292,17 @@ properties:
...
+  - name: freeSpace
+    summary: Releases the space when hiding an object
+    description: |
+        If setting `freeSpace` to true it will automatically release the space the view occupied.
+        For example: in a vertical layout the views below the object will move up when you hide
+        an object with `freeSpace:true`.
+    type: Boolean
+    default: false
+    platforms: [android]
+    since: 6.0.0
```
hansemannn: 6.0.0 is already in QE-testing, please set it to 6.1.0 and we'll try to review it in that timeframe. Thx!

hansemannn (changes requested):

android/titanium/src/java/org/appcelerator/titanium/TiC.java
m1ga: Good idea! I'll go for `hiddenBehaviour: Ti.UI.VIEW_GONE | Ti.UI.VIEW_INVISIBLE`
Then it looks almost like the native View.GONE and View.INVISIBLE

hansemannn (approved):

hansemannn:
Looks great now! Putting into further review for 6.1.0.

android/modules/ui/src/java/ti/modules/titanium/ui/UIModule.java
```diff
@@ -178,6 +178,9 @@
 	@Kroll.constant public static final int INPUT_TYPE_CLASS_NUMBER = InputType.TYPE_CLASS_NUMBER;
 	@Kroll.constant public static final int INPUT_TYPE_CLASS_TEXT = InputType.TYPE_CLASS_TEXT;
 
+	@Kroll.constant public static final int VIEW_GONE = View.GONE;
+	@Kroll.constant public static final int VIEW_INVISIBLE = View.INVISIBLE;
```
hansemannn: Thinking about it, we might want to rename them to `Ti.UI.HIDDEN_BEHAVIOR_GONE` and `Ti.UI.HIDDEN_BEHAVIOR_INVISIBLE`. What do you think? And please add + validate the docs for the constants, they are not documented, yet.

m1ga: Sure, changed the property and updated the constants inside the doc! `HIDDEN_BEHAVIOR_INVISIBLE` is fine, too. So its more connected to the property name

hansemannn:
CR + FT passed, PR approved! Great work! 🚀

PR #8459, opened 2016-10-03
[TIMOB-23966] Fix kroll-v8 TypeConverter

garymathews:
- Fix `jsValueToJavaObject` for numeric values
- Always parse JSON properties key as `String` for both `jsValueToJavaObject` and `jsObjectToJavaKrollDict`
###### TEST CASE

``` Javascript
// should not throw java.lang.ClassCastException
Ti.Analytics.featureEvent('test', {'1':'data'});
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23966)

### DIFF
--- a/android/runtime/v8/src/native/TypeConverter.cpp
+++ b/android/runtime/v8/src/native/TypeConverter.cpp
@@ -663,20 +663,18 @@ jobject TypeConverter::jsValueToJavaObject(v8::Isolate* isolate, JNIEnv *env, v8
 
 			for (int i = 0; i < numKeys; i++) {
 				v8::Local<v8::Value> jsObjectPropertyKey = objectKeys->Get((uint32_t) i);
-				bool keyIsNew, valueIsNew;
-				jobject javaObjectPropertyKey = TypeConverter::jsValueToJavaObject(isolate, env, jsObjectPropertyKey, &keyIsNew);
+				bool valueIsNew;
+				jstring javaStringPropertyKey = TypeConverter::jsValueToJavaString(isolate, env, jsObjectPropertyKey);
 				v8::Local<v8::Value> jsObjectPropertyValue = jsObject->Get(jsObjectPropertyKey);
 				jobject javaObjectPropertyValue = TypeConverter::jsValueToJavaObject(isolate, env, jsObjectPropertyValue, &valueIsNew);
 
 				jobject result = env->CallObjectMethod(javaHashMap,
 				                                       JNIUtil::hashMapPutMethod,
-				                                       javaObjectPropertyKey,
+				                                       javaStringPropertyKey,
 				                                       javaObjectPropertyValue);
 				env->DeleteLocalRef(result);
+				env->DeleteLocalRef(javaStringPropertyKey);
 
-				if (keyIsNew) {
-					env->DeleteLocalRef(javaObjectPropertyKey);
-				}
 				if (valueIsNew) {
 					env->DeleteLocalRef(javaObjectPropertyValue);
 				}
@@ -715,20 +713,18 @@ jobject TypeConverter::jsObjectToJavaKrollDict(v8::Isolate* isolate, JNIEnv *env
 
 		for (int i = 0; i < numKeys; i++) {
 			v8::Local<v8::Value> jsObjectPropertyKey = objectKeys->Get((uint32_t) i);
-			bool keyIsNew, valueIsNew;
-			jobject javaObjectPropertyKey = TypeConverter::jsValueToJavaObject(isolate, env, jsObjectPropertyKey, &keyIsNew);
+			bool valueIsNew;
+			jstring javaStringPropertyKey = TypeConverter::jsValueToJavaString(isolate, env, jsObjectPropertyKey);
 			v8::Local<v8::Value> jsObjectPropertyValue = jsObject->Get(jsObjectPropertyKey);
 			jobject javaObjectPropertyValue = TypeConverter::jsValueToJavaObject(isolate, env, jsObjectPropertyValue, &valueIsNew);
 
 			jobject result = env->CallObjectMethod(javaKrollDict,
 				                                   JNIUtil::krollDictPutMethod,
-				                                   javaObjectPropertyKey,
+				                                   javaStringPropertyKey,
 				                                   javaObjectPropertyValue);
 			env->DeleteLocalRef(result);
+			env->DeleteLocalRef(javaStringPropertyKey);
 
-			if (keyIsNew) {
-				env->DeleteLocalRef(javaObjectPropertyKey);
-			}
 			if (valueIsNew) {
 				env->DeleteLocalRef(javaObjectPropertyValue);
 			}


sgtcoolguy:
This fix looks like it should work for this particular case. But the disconcerting thing for me is how the string '1' is ever getting converted into an integer in the first place (when the HashMap is created). Did you track down where that is happening? Maybe in the native c++ code like TypeConverter.cpp?

I'd think we'd want it fixed there (wherever it's converting it), but don't know if that could cause lots of unintended bugs as a result.

sgtcoolguy:
For reference, the native code for Analytics module's featureEvent does this for the data arg:
{code:cpp}
Local<Value> arg_1 = args[1];
jArguments[1].l = titanium::TypeConverter::jsObjectToJavaKrollDict(isolate, env, arg_1, &isNew_1);
{code}

Which points to https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/native/TypeConverter.cpp#L708

I assume that ultimately, it calls https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/native/TypeConverter.cpp#L609 and maybe IsNumber() returns true for '1'? It's difficult to tell if that would return true or not without adding some logging to confirm and doing a one-off local build. It may need to be changed to IsNumberObject() explicitly? Maybe IsNumber() can do implicit conversion/coercion?

garymathews:
@sgtcoolguy Shouldn't the property key always be a string? https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/native/TypeConverter.cpp#L667

sgtcoolguy:
My suspicions have been confirmed locally (finally!):

```
[INFO]  passing in data object to featureEvent!
[WARN]  TypeConverter: jsObjectToJavaKrollDict converting object...
[WARN]  TypeConverter: isNumber == true!
[WARN]  TypeConverter: isString == true!
[ERROR] JNIUtil: jsObjectToJavaKrollDict key: java.lang.Integer
[WARN]  W/System.err: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
[WARN]  W/System.err: 	at org.appcelerator.titanium.util.TiConvert.toJSON(TiConvert.java:723)
[WARN]  W/System.err: 	at ti.modules.titanium.analytics.AnalyticsModule.featureEvent(AnalyticsModule.java:98)
[WARN]  W/System.err: 	at org.appcelerator.kroll.runtime.v8.V8Object.nativeFireEvent(Native Method)
[WARN]  W/System.err: 	at org.appcelerator.kroll.runtime.v8.V8Object.fireEvent(V8Object.java:62)
[WARN]  W/System.err: 	at org.appcelerator.kroll.KrollProxy.doFireEvent(KrollProxy.java:872)
[WARN]  W/System.err: 	at org.appcelerator.kroll.KrollProxy.handleMessage(KrollProxy.java:1095)
[WARN]  W/System.err: 	at org.appcelerator.titanium.proxy.TiViewProxy.handleMessage(TiViewProxy.java:354)
[WARN]  W/System.err: 	at android.os.Handler.dispatchMessage(Handler.java:98)
[WARN]  W/System.err: 	at android.os.Looper.loop(Looper.java:148)
[WARN]  W/System.err: 	at android.app.ActivityThread.main(ActivityThread.java:5417)
[WARN]  W/System.err: 	at java.lang.reflect.Method.invoke(Native Method)
[WARN]  W/System.err: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
[WARN]  W/System.err: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
[ERROR] TiExceptionHandler: (main) [43784,43784] ----- Titanium Javascript Runtime Error -----
[ERROR] TiExceptionHandler: (main) [0,43784] - In /alloy/controllers/index.js:16,35
[ERROR] TiExceptionHandler: (main) [0,43784] - Message: Uncaught java.lang.Integer cannot be cast to java.lang.String
[ERROR] TiExceptionHandler: (main) [0,43784] - Source:         var result = Ti.Analytics.featureEvent("levels", data);
[ERROR] V8Exception: Exception occurred at /alloy/controllers/index.js:16: Uncaught java.lang.Integer cannot be cast to java.lang.String
[ERROR] V8Exception: java.lang.Integer cannot be cast to java.lang.String
```

garymathews:
@sgtcoolguy Yep, I just pushed a fix 👍

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -663,20 +663,18 @@ jobject TypeConverter::jsValueToJavaObject(v8::Isolate* isolate, JNIEnv *env, v8
 
 			for (int i = 0; i < numKeys; i++) {
```
sgtcoolguy: This code looks nearly the same as jsObjectToJavaKrollDict, but in one case we're generating a HashMap, and in the other we're generating a KrollDict. I don't see why that should be the case, and we should consider refactoring this code out to a common method (or maybe we can just call jsObjectToJavaKrollDict here?)

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -606,7 +606,7 @@ jobject TypeConverter::jsValueToJavaObject(v8::Isolate* isolate, v8::Local<v8::V
 
 jobject TypeConverter::jsValueToJavaObject(v8::Isolate* isolate, JNIEnv *env, v8::Local<v8::Value> jsValue, bool *isNew)
 {
-	if (jsValue->IsNumber()) {
+	if (jsValue->IsNumberObject()) {
```
sgtcoolguy: So clearly this is the correct thing to do, otherwise '1' returns true for IsNumber() leading to the bug we saw. However, are we certain this fix won't actually break other code? I'm guessing it shouldn't, but our Java code isn't consistent about enforcing HashMap\<String, Object\>(and besides, as seen here, generics are more a compile time suggestion and the mix of native code can get around it).

android/runtime/v8/src/native/TypeConverter.cpp
sgtcoolguy: Again, this change at the heart of our runtime scares me, but I think since KrollDict extends HashMap\<String, Object\> that always generating a KrollDict would be ok...

And on a real tangent, I don't see why we actually use HashMap in our APIs everywhere and not just Map to be more generic.

sgtcoolguy (approved):
LGTM 👍 

I think this fixes the root cause and is the right thing to do, but obviously with our not so great unit test coverage any change in some core runtime functionality like this scares me...

sgtcoolguy:
@garymathews Feel free to merge so long as the builds are happy

sgtcoolguy:
@garymathews Yikes, the build is **not** very happy. Looks like changing the IsNumber() to IsNumberObject() breaks a lot of functionality. Maybe we can get away with leaving it as IsNumber() check there, and when we convert a JS object to HashMap/KrollDict we force the key to call jsValueToJavaString - as you did? That might be enough to fix this issue for now. It is disconcerting that the IsNumber check is so important...

sgtcoolguy:
so clearly I think our existing behavior is basically flukey and happened by chance/luck more than anything. We should definitely revisit the behavior for jsValueToJavaObject, since order of comparisons clearly matters. The test failures on this past PR build should be pretty informative to see exactly what case broke by moving to a more explicit Number check. But I'd argue we should define the actual conversion order in a more formal way, likely trying explicit checks like IsNumberObject(), IsBooleanObject(), etc first, and if all of those fail, then moving down to the more forgiving checks (but again, what order we try those will still matter - not sure if there's any spec on that).

Looking at the code, it looks like this conversion is used when:
- we are trying to convert to a generic java.lang.Object
- we are trying to convert to a KrollProxy
- converting elements in varargs array
- the result of invoking a V8Function
- the result of calling a native property on a V8Object
- the result of nativeEvalString in V8Runtime
- converting js args into an Object[] (for each element)
- converting js array to java Object[] (for each element)
- converting keys and values in a JS Object (which this PR would fix to only be used for values)
- in jsObjectIndexPropsToJavaArray (which seems to be used only in some special edge case on creating a proxy object?)

I'm not sure if we can move away from using it so much and use more specific conversions where we know the target types in advance?

On another tangent, since we've updated V8 to ES6 compat, we should actually consider the new JS types that have been introduced and how to convert/handle those:
- Generators
- Promises
- Maps
- Sets
- Typed arrays (uint8, uint8 clamped, int8, uint16, int16, uint32, int32, float32, float64)

garymathews:
I think this will suffice for now. I have created a ticket [TIMOB-23990](https://jira.appcelerator.org/browse/TIMOB-23990) to improve the current implementation.

sgtcoolguy:
@garymathews Did you revert the IsNumber(0 -> IsNumberObject() check? I don't see a commit with that, but I think that should fix this...

garymathews:
@sgtcoolguy Nope, it's the first change here https://github.com/appcelerator/titanium_mobile/pull/8459/files

But the other changes I made prevent the property key from being interpreted by `jsValueToJavaObject` and just interpret it as a string. [JSON Objects](https://tools.ietf.org/html/rfc4627#section-2.2)

sgtcoolguy:
@garymathews Yeah, but I think we need to revert the IsNumberObject() change, because it is breaking a ton of our tests. So yes, let's retain the explicit calls to jsValueToJavaString for keys when generating the KrollDict/Map - that should fix the original ticket. But we need to remove the IsNumberObject() change to avoid breaking all our tests.

garymathews:
Updated both PRs

garymathews:
LGTM 👍

PR #8466, opened 2016-10-04
[TIMOB-23494] Android: Support #createFile() method on Ti.Filesystem.File

frankieandone:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23494


cheekiatng:
@fmerzadyan can you fix the conflict please?

fmerzadyan:
I can't merge. Unresolved conflicts on my side too. I may have a broken my fork. Maybe I should close my PRs, revert back to a stable commit like `8dd8462 Merge pull request #8398 from cb1kenobi/timob-23912`, git fetch and pull upstream master, then re-add my changes? Unless there's a way to remove commits/middle-nodes from master branch without affecting the other nodes, my fork's health or others?

sgtcoolguy:
@fmerzadyan The easiest way to fix this PR is to:
```
git checkout master
git pull upstream master
git checkout -b TIMOB-23494-good
git cherry-pick 00cb1b5
git branch -D TIMOB-23494
git branch -m TIMOB-23494-good TIMOB-23494
git push origin TIMOB-23494 --force
```

What does that do?
- Change to master branch
- Pull the latest changes on master branch from "upstream" remote (I'm assuming that's what you named the remote pointing at our shared repo and not your fork?)
- Create a new branch off master named "TIMOB-23494-good"
- Cherry pick over the one commit we want to our new branch
- force delete the local copy of the bad branch this PR is currently using.
- rename our good branch to the original ticket name (and the name Github is using for this PR)
- forcibly push what you have locally up to Github (to update the PR's commit history with just our one good cherry-pick commit)

android/titanium/src/java/org/appcelerator/titanium/io/TiFile.java
```diff
@@ -113,6 +113,17 @@ public boolean createDirectory(boolean recursive)
 		}
 	}
 
+    @Override
+    public boolean createFile() {
```
sgtcoolguy: I don't think this will compile. You're trying to re-throw the IOException and the method doesn't declare it throws that exception type (and neither does the super implementation we're overriding).

Also, if you're just catching an exception and rethrowing it, you don't really need to do that, you're basically just doing what would have already happened anyways.

So you either need to declare here and in TiBaseFile that #createFile() throws IOException, or you need: catch it, ignore or log it, and return false.

Given that our API doesn't mention throwing errors on failure, but uses a straight boolean, I'd keep the method signature as is and do:

```java
@Override
public boolean createFile() {
	try {
		return file.createNewFile();
	} catch (IOException e) {
		Log.e(TAG, "Error creating new file: ", e);
		return false;
	}
}
```

sgtcoolguy (changes requested):

PR #8474, opened 2016-10-04
Updated Facebook.yml per TIDOC-2625

jawa9000:

**Optional Description:**

Facebook.yml had two descriptions in the permissions property. Content of the two descriptions were merged.


hansemannn:
The `Ti.Facebook` in this docs haven't been touched since years, because our docs-script grabs them from https://github.com/appcelerator-modules/ti.facebook/tree/master/apidoc instead. So this is not fixing the docs-issues as well.

PR #8478, opened 2016-10-06
Fix  6.1.0.v20161006012211-osx iOS App Store distribution build fail

snakewa:
Just a quick fix


sgtcoolguy:
ok to test

hansemannn:
Thanks for the PR, great! I would suggest the following solution that looks a bit cleaner from a syntax-perspective: https://github.com/appcelerator/titanium_mobile/commit/a191bf16cd1ec274b456b4eeb327ce531675b651

Let me know if it works for you! Thanks!

PR #8479, opened 2016-10-06
[TIMOB-23880] Default classic app template does not show tabs on Windows

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23880

**Optional Description:**

Overrides #8333 
- Don't set global bg color at all. Makes no difference on iOS/Android by default, and actually breaks Windows UI.


hansemannn:
It is causing rendering artifacts while switching tabs on iOS, but fine.

PR #8278, opened 2016-08-28
[TIMOB-9007] iOS: Fix default value for Ti.UI.Switch

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-9007

PR #8484, opened 2016-10-07
[TIMOB-23983] (6_0_X) iOS Debugger: Breakpoint not hit for Alloy applications

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23983

**Description:**
- Call TiDebuggerBeginScript and TiDebuggerEndScript before and after loading a JS file via require when not using jscore framework
- This fixes the "legacy" debugger, used when use-jscore-framework is false or not set
- New iOS debugger still has an issue, and I'm looking into it.

### DIFF
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -445,15 +445,16 @@ CFMutableSetRef	krollBridgeRegistry = nil;
 	if (exception == NULL) {
 #ifndef USE_JSCORE_FRAMEWORK
 		if ([[self host] debugMode]) {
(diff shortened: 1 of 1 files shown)


hansemannn:
LGTM. So we need to wrap the JS-evaluation with `TiDebuggerBeginScript` and `TiDebuggerEndScript`? Makes sense.

hansemannn:
And if we backport https://github.com/appcelerator/titanium_mobile/pull/8278, our 6_0_X unit-tests should also succeed. @cheekiatng ?

cheekiatng:
CR and FT passed. APPROVED!

PR #8485, opened 2016-10-07
[TIMOB-23983] iOS Debugger: Breakpoint not hit for Alloy applications

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23983

**Description:**
- Call TiDebuggerBeginScript and TiDebuggerEndScript before and after loading a JS file via require when not using jscore framework
- This fixes the "legacy" debugger, used when use-jscore-framework is false or not set
- New iOS debugger still has an issue, and I'm looking into it.


cheekiatng:
CR and FT passed! APPROVED!

PR #8493, opened 2016-10-10
[TIMOB-24011] iOS: require fails to resolve paths like 'hyperloop/uikit/uilabel'

sgtcoolguy:
**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-24011
- https://jira.appcelerator.org/browse/TIMOB-24012

**Description:**
This bug was found in hyperloop 2.0.0 against SDK 6+. We generate "legacy" style require paths with no leading '/', './', or '../' in them, but instead something like:

``` js
require('hyperloop/uikit/uilabel');
```

This path style is now considered deprecated and paths should conform to Node.JS style.

In any case, the new require behavior should have fallen back to assuming an "absolute" path absent the leading slash and worked anyways. Instead there's a bug where if the first segment of the path matches a native module in use, we will not properly fall back (and will return the native nodule instance instead).

This patch fixes the issue - we first look in the native module for the asset, and if that fails we should fallback to loading the file from the user's app.

### DIFF
--- a/android/runtime/common/src/js/module.js
+++ b/android/runtime/common/src/js/module.js
@@ -311,7 +311,8 @@ Module.prototype.require = function (request, context) {
 Module.prototype.loadCoreModule = function (id, context) {
 	var wrapper = this.wrapperCache[id],
 		parts,
-		externalBinding;
+		externalBinding,
+		externalCommonJsContents;
 	// check if we have a cached copy of the wrapper
 	if (wrapper) {
 		return wrapper;
@@ -330,14 +331,16 @@ Module.prototype.loadCoreModule = function (id, context) {
 
 		// Could be a sub-module (CommonJS) of an external native module.
 		// We allow that since TIMOB-9730.
-		externalCommonJsContents = kroll.getExternalCommonJsModule(id);
-		if (externalCommonJsContents) {
-			// found it
-			// FIXME Re-use loadAsJavaScriptText?
-			var module = new Module(id, this, context);
-			Module.cache[id] = module;
-			module.load(id, externalCommonJsContents);
-			return module.exports;
+		if (kroll.isExternalCommonJsModule(parts[0])) {
+			externalCommonJsContents = kroll.getExternalCommonJsModule(id);
+			if (externalCommonJsContents) {
+				// found it
+				// FIXME Re-use loadAsJavaScriptText?
+				var module = new Module(id, this, context);
+				Module.cache[id] = module;
+				module.load(id, externalCommonJsContents);
+				return module.exports;
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


sgtcoolguy:
Hey @hansemannn can you take a look at this PR?

The test failures from the PR build are unrelated (Picker UI tests?)

hansemannn:
@sgtcoolguy The error message indicates an Android problem:
```
17:21:57 Titanium.UI.Picker
17:21:57   ✓ PlainPicker.removeRow 
17:21:57   1) PlainPicker.add (PickerRow)
17:21:57     {"state":"failed","duration":14562,"suite":"Titanium.UI.Picker","title":"PlainPicker.add (PickerRow)","error":{},"stdout":"[INFO] : \b\b \b Choreographer: Skipped 648 frames!  The application may be doing too much work on its main thread.\n","stderr":"[ERROR] : \b\b \bSurface: getSlotFromBufferLocked: unknown buffer: 0xeecf4100\n[ERROR] : \b\b \bI/GAv4-SVC( 2221): Google Analytics 9.4.52 is starting up.\n[ERROR] : \b\b \bSurface: getSlotFromBufferLocked: unknown buffer: 0xdf788f10\n"}
17:21:57   ✓ PlainPicker.add(multiple PickerColumn) 
17:21:57   2) PlainPicker.add(PickerColumn)
17:21:57     {"state":"failed","duration":11641,"suite":"Titanium.UI.Picker","title":"PlainPicker.add(PickerColumn)","error":{},"stdout":"[INFO] : \b\b \b Choreographer: Skipped 658 frames!  The application may be doing too much work on its main thread.\n[INFO] : \b\b \b Choreographer: Skipped 326 frames!  The application may be doing too much work on its main thread.\n","stderr":"[ERROR] : \b\b \bSurface: getSlotFromBufferLocked: unknown buffer: 0xdf788b90\n[ERROR] : \b\b \bSurface: getSlotFromBufferLocked: unknown buffer: 0xdf788c00\n"}
```
The related unit-test is [this one|https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.ui.picker.test.js#L177]. I will re-trigger it now to ensure it's a sustained error.

sgtcoolguy:
@hansemannn yeah, it should be entirely unrelated, and is likely just an intermittent failure due to timeout or something. (we have a few intermittent UI failures because the tests don't specify a long enough timeout and rapidly executing all these UI tests/actions can cause the system to get "behind").

hansemannn:
@sgtcoolguy succeeds now :-)

hansemannn:
CR looks good, I will do a FT on some of my bigger projects to verify the different scenarios.

sgtcoolguy:
@hansemannn This specific issue regards paths whose first segment is the same as a native module id, has multiple segments, and has no leading '/', './', or '../'. It's not a particularly likely scenario in normal usage, I wouldn't think. 

I'm unsure if the Android fix is even a regression - it may have still occurred if you attempted to load a non-existent JS resource in a native module. I haven't tried in an older SDK, but looking at the code as it existed, it probably had the same issue.

hansemannn:
You are right, then at least we have nothing to be afraid of! 🙂

hansemannn:
The iOS-side looks good (both CR + FT). We can merge it when Android is approved as well.

PR #8499, opened 2016-10-11
[TIMOB-23185] TiAPI: Remove Ti.include and advertise workaround

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23185

**Description:**
This finally removes Ti.include for Android, iOS and Mobileweb. Also removes the entry from docs.

I'm not sure if we want to actually retain the doc entry and instead just say it's been removed and please use require? Seems odd to me that we'd retain a doc entry for a method that no longer exists and has been deprecated since 3.3.0.


hansemannn:
@sgtcoolguy Can you pull from master and do the 6_0_X backport? Thx!

PR #8500, opened 2016-10-11
[TIMOB-23997] Reverted bad CommonJS module logic

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23997

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -4,7 +4,7 @@
  * @module cli/_build
  *
  * @copyright
- * Copyright (c) 2009-2015 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-2016 by Appcelerator, Inc. All Rights Reserved.
  *
  * Copyright (c) 2012-2013 Chris Talkington, contributors.
  * {@link https://github.com/ctalkington/node-archiver}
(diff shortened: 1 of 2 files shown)


android/cli/commands/_build.js
```diff
@@ -1472,24 +1472,32 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
...
-									process.exit(1);
+					// look for legacy module.id.js first
+					var libFile = path.join(module.modulePath, module.id + '.js');
+					module.libFile = fs.existsSync(libFile) ? libFile : null;
+					if (!module.libFile) {
+						var pkgJsonFile = path.join(module.modulePath, 'package.json');
+						if (fs.existsSync(pkgJsonFile)) {
+							try {
+								var pkgJson = require(pkgJsonFile);
+								if (pkgJson && pkgJson.main && fs.existsSync(libFile = path.join(module.modulePath, pkgJson.main + (/\.js$/.test(pkgJson.main) ? '' : '.js')))) {
```
sgtcoolguy: So this clearly is doing more validation that the package.json has a main property and checking to see if the main property file exists, which we'd want.

Because package.json is used by require logic, we can figure out the possible valid values for main property sort of indirectly here: https://nodejs.org/api/modules.html#modules_all_together

Effectively we know it tries to load the value as a file, which means the test for no extension and then appending '.js' wouldn't be right. We need to try, in order:
```js
path.join(module.modulePath, pkgJson.main);
path.join(module.modulePath, pkgJson.main + '.js');
path.join(module.modulePath, pkgJson.main + '.json');
```

sgtcoolguy (changes requested):
If we're going to validate package.json's main property, we need to test a few more scenarios.

android/cli/commands/_build.js
sgtcoolguy: I updated the code as it existed. I was unaware that libFile was unused, but I viewed this block as basic validation of the structure of CommonJS modules. As such, I updated the logic to check for more than the module_id.js file so npm style modules would work.

So yeah, we could nuke the block and just assume that the CommonJS modules are ok, and bomb at runtime instead if they're not. Or we could retain the validation and improve it.

cb1kenobi:
**APPROVED**

sgtcoolguy:
@cb1kenobi This has broken our support for commonJS modules that don't use the legacy naming where there has to be a JS file named the exact same as the module id. I specifically did not approve the PR because of this concern. This has broken our test suite which validates that we can handle nom-style modules. I need to revert this.

PR #8504, opened 2016-10-13
[TIMOB-24017] (6_0_X) Added support for validating the connection to the log server.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24017


iphone/Classes/TiLogServer.m
```diff
@@ -223,6 +242,23 @@ +(void)startServer
...
+			TI_APPLICATION_NAME, @"name",
+			TI_APPLICATION_ID, @"appId",
+			TI_APPLICATION_VERSION, @"version",
+			TI_APPLICATION_DEPLOYTYPE, @"deployType",
+			TI_APPLICATION_GUID, @"guid",
+			version, @"tiSDKVersion",
+			@"__GITHASH__", @"githash",
+			nil];
+
+		headers = [[[[TiUtils jsonStringify:map] stringByAppendingString:@"\r\n"] dataUsingEncoding:NSUTF8StringEncoding] retain];
```
cheekiatng: i wonder if this `headers` has to be released eventually somewhere. @hansemannn thoughts?

iphone/Classes/TiLogServer.m
```diff
@@ -223,6 +242,23 @@ +(void)startServer
 		connections = [[NSMutableArray alloc] init];
 	}
 
+	if (headers == nil) {
+		NSString* version = [NSString stringWithCString:TI_VERSION_STR encoding:NSUTF8StringEncoding];
+		NSDictionary* map = [[NSDictionary alloc] initWithObjectsAndKeys:
```
cheekiatng: there's a release eventually for map, which is fine. 
An alternative way you can do this is with `[NSDictionary dictionaryWithObjectsAndKeys ...]` so that you don't have to call release later.

iphone/Classes/TiLogServer.m
```diff
@@ -283,6 +319,13 @@ +(void)startServer
 		[connections addObject:conn];
 		forcedNSLog(@"[INFO] Log server connections: %d", [connections count]);
 
+		// send the header
+		dispatch_data_t buffer = dispatch_data_create(headers.bytes,
+													  headers.length,
```
cheekiatng: the indentation here is weirded.

iphone/Classes/TiLogServer.m
```diff
@@ -283,6 +319,13 @@ +(void)startServer
...
+													  logDispatchQueue,
+													  DISPATCH_DATA_DESTRUCTOR_DEFAULT);
+		[conn send:&buffer];
```
cheekiatng: Not sure if we should check for nil here for buffer.

iphone/Classes/TiLogServer.m
cb1kenobi: That's the Xcode default. I don't love it. Other than putting everything on a single line, how would you like to see it indented?

iphone/Classes/TiLogServer.m
cb1kenobi: Oh, so `[NSDictionary dictionaryWithObjectsAndKeys ...]` is essentially a static method. That's cool. Want me to update it?

cheekiatng: maybe not yet. I like Hans to take a look too. by the way, I did the functional tests, it didn't work :( they are still happily logging individually, even though i set the same port number for both apps.

iphone/Classes/TiLogServer.m
hansemannn: We should release it in the `-(void)dealloc` delegate.

iphone/Classes/TiLogServer.m
```diff
@@ -368,6 +368,10 @@ +(void)stopServer
 		RELEASE_TO_NIL(connections);
 	}
 
+	if (headers != nil) {
```
hansemannn: The `if` is not necessary since it's checked in `RELEASE_TO_NIL`

cb1kenobi: Fixed.

cheekiatng:
CR and FT passed. APPROVED!

PR #8507, opened 2016-10-13
Fix missing summary- and type-keys

hansemannn:


hansemannn:
Both `node validate` and `python validate.py` are succeeding now. Merging!

PR #8513, opened 2016-10-14
[TIMOB-23493] Android: #append method in Ti.Filesystem.File

frankieandone:
**JIRA:** https://jira.appcelerator.org/browse/timob-23493

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
@@ -140,8 +140,7 @@ public class TiFileProxy extends KrollProxy
 	@Kroll.method
 	public boolean append(Object data)
 	{
-		Object[] objectArray = {data};
-		return write(objectArray, true);
+		return write(new Object[]{data, true});
 	}
 
 	@Kroll.method
@@ -289,13 +288,15 @@ public class TiFileProxy extends KrollProxy
 		return tbf.spaceAvailable();
 	}
 
-	private boolean write(Object[] args, boolean append)
+	@Kroll.method
+	public boolean write(Object[] args)
 	{
 		try {
 
 			if (args != null && args.length > 0) {
+				boolean append = false;
 				if (args.length > 1 && args[1] instanceof Boolean) {
-					append = ((Boolean)args[1]).booleanValue();
+					append = (Boolean) args[1];
 				}
 
 				if (args[0] instanceof TiBlob) {
@@ -317,12 +318,6 @@ public class TiFileProxy extends KrollProxy
 			Log.e(TAG, "IOException encountered", e);
 			return false;
 		}
-	}
-
-	@Kroll.method
-	public boolean write(Object[] args)
-	{
-		return write(args, false);
 	}
 
 	@Kroll.method
--- a/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
@@ -139,6 +139,17 @@ public class TiFileProxy extends KrollProxy
 	}
 
 	@Kroll.method
+	public boolean append(Object data)
+	{
+		try {
+			return ((TiFile) tbf).append(data);
+		} catch (IOException e) {
+			Log.e(TAG, "append failed: ", e);
+		}
+		return false;
+	}
+
+	@Kroll.method
 	public boolean copy (String destination)
 		throws IOException
 	{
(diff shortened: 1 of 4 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -135,6 +134,19 @@ public boolean getWritable()
 	}
 
 	@Kroll.method
+    public boolean append(Object data)
```
garymathews: Use tabs

android/titanium/src/java/org/appcelerator/titanium/io/TiBaseFile.java
```diff
@@ -78,105 +79,136 @@ protected TiBaseFile(int type) {
...
 	 * @return true if the file is a symbolic link, false otherwise.
 	 * @module.api
 	 */
-	public boolean isSymbolicLink() {
+	public boolean isSymbolicLink()
+	{
 		return flagSymbolicLink;
 	}
 
+    public boolean append(Object data) throws IOException
```
garymathews: Use tabs

garymathews:
It's best to not modify the original formatting of the files you are making changes to; it's hard to review as I cannot distinguish the original code from the changes you have made.

sgtcoolguy:
@fmerzadyan My comments from #8471 still apply. Appending using writers is incorrect and wasteful in terms of performance as well as may transform the contents. Please just copy bytes (for File) as you do with TiBlob.

Also, please try to avoid closing and re-opening new PRs for the same code/ticket. We lose history of the reviews and it becomes difficult to track between ticket and correct PR.

sgtcoolguy:
@fmerzadyan There's a merge conflict that needs to be resolved; looks like somehow one of Gary's commits made it into your branch; and lastly, we need to update the yaml docs to indicate that append is now supported on android (apidoc/Titanium/Filesystem/File.yml)

android/titanium/src/java/org/appcelerator/titanium/io/TiFile.java
```diff
@@ -95,6 +97,71 @@ public boolean isWriteable()
 		return file.canWrite();
 	}
 
+	@Override
```
garymathews: Why not just make use of the `write` methods that already exist with an `append` flag?

```Java
@Override
public boolean append(Object data) throws IOException
{
	try {
		if (data instanceof String) {
			write((String) data, true);
		} else if (data instanceof TiFileProxy) {
			// write((TiFileProxy) data, true); // implement an append flag into this
		} else if (data instanceof TiBlob) {
			write((TiBlob) data, true);
		}
		return true;
	} catch (IOException e) {
		Log.e(TAG, "Error appending: ", e);
	}
	return false;
}
```

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/io/TiFile.java
fmerzadyan: If you're going that way then it would just be better to edit TiFileProxy instead because less changes would need to be made.

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -139,6 +138,13 @@ public boolean getWritable()
 	}
 
 	@Kroll.method
+	public boolean append(Object data)
```
garymathews: You can just do this, there's no need for any other changes.

```Java
@Kroll.method
 public boolean append(Object data)
 {
	return write(new Object[]{data, true});
}
```

fmerzadyan: There was only 1 write method. The original write method had only one argument (Object[] args). I changed it so that it could take in append argument which is necessary. It keeps the original write exposed, the new write method not exposed and append is exposed too. But I'll be sure to change this one line.

garymathews: [Titanium.Filesystem.write](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Filesystem.File-method-write) should be exposed with the optional `append` parameter, so you can just make use of that 👍

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -283,23 +288,21 @@ public double spaceAvailable()
...
-	@Kroll.method
-	public boolean write(Object[] args)
+	private boolean write(Object[] args, boolean append)
 	{
 		try {
 
 			if (args != null && args.length > 0) {
-				boolean append = false;
 				if (args.length > 1 && args[1] instanceof Boolean) {
 					append = ((Boolean)args[1]).booleanValue();
```
fmerzadyan: I thought I removed that, my mistake.

garymathews:
CR: **PASS**
FR: **PASS**

👍

garymathews:
test this please

PR #8514, opened 2016-10-14
[timob-23494] Android: Support #createFile() method on Ti.Filesystem.File

frankieandone:
**JIRA:** https://jira.appcelerator.org/browse/timob-23494

**Android did not support createFile**

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
@@ -6,6 +6,7 @@
  */
 package org.appcelerator.titanium;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
@@ -13,10 +14,12 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 
+import android.os.Environment;
 import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.kroll.common.Log;
 import org.appcelerator.titanium.io.TiBaseFile;
+import org.appcelerator.titanium.io.TiFile;
 import org.appcelerator.titanium.io.TiFileFactory;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.util.TiFileHelper2;
@@ -72,7 +75,6 @@ public class TiFileProxy extends KrollProxy
 		} else {
 			path = TiFileHelper2.joinSegments(parts);
 		}
-		
 		if (resolve) {
 			path = resolveUrl(scheme, path);
 		}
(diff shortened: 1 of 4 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -152,6 +152,12 @@ public boolean createDirectory(@Kroll.argument(optional=true) Object arg)
 		return tbf.createDirectory(recursive);
 	}
 
+    @Kroll.method
+    public boolean createFile()
```
garymathews: Formatting, use tabs

sgtcoolguy:
@fmerzadyan Please update the unit test for Ti.Filesystem.File.createFile to no longer skip on Android and confirm that this fix works.

Easiest way to do so is to clone appcelerator/titanium-mobile-mocha-suite locally, and copy Resources/ti.filesystem.file.test.js into appcelerator/titanium_mobile/tests/Resources. Then modify the test to no longer skip (around line 204):
```javascript
it('#createFile() and #deleteFile()', function () {
    var newFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, 'myfile');
```

Add the updated file into your PR.

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -153,6 +157,15 @@ public boolean createDirectory(@Kroll.argument(optional=true) Object arg)
 	}
 
 	@Kroll.method
+	public boolean createFile()
+	{
+		Context context = TiApplication.getInstance().getApplicationContext();
```
sgtcoolguy: Is all of this necessary? I think this is hard-coding the location of the file when it shouldn't be. Don't we already create the correct TiBaseFile up in the constructor? You should be able to just do `return tbf.createFile();` here.

sgtcoolguy (changes requested):
The test is passing, so great job there. I think just one more small change is necessary - you appear to be recreating a TiBaseFile using a hard-coded location unnecessarily. Just use the TiBaseFile that's already generated in the constructor and wrapped by the proxy.

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
fmerzadyan: `return tbf.createFile()` would mean `TiBaseFile`'s `createFile` method which would simply return false everytime (see [tbf.createFile).](https://github.com/appcelerator/titanium_mobile/pull/8514/files#diff-0c93a7953c3607b38b61844d828ba1a4R193) To start from the beginning, when the url is passed to the `TiFileProxy`, `tbf` is instantiated to a child class `TiResourceFile` which is a sibling to `TiFile`. In order to use `TiFile.createFile` I needed to convert `tbf` to `TiFile` and to do this I needed a path. `tbf.getNativePath` would throw errors stating that directory and file did not exist. Even using `mkdirs` or `mkdir` didn't fix this so I used `ContextWrapper` instead. It doesn't affect other url's and also `applicationDataDirectory`, `resourcesDirectory`, etc would return the same url in end when it is parsed in `TiFileFactory`. (see [storage locations](http://docs.appcelerator.com/platform/latest/#!/guide/Filesystem_Access_and_Storage)

sgtcoolguy (approved):

PR #8515, opened 2016-10-14
[TIMOB-24034] Fixed differential builds by no longer selecting a rand…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24034

Fixed differential builds by no longer selecting a random log server port.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -1469,8 +1469,9 @@ iOSBuilder.prototype.validate = function (logger, config, cli) {
 
 		// validate the log server port
 		var logServerPort = this.tiapp.ios['log-server-port'];
-		if (!/^dist-(appstore|adhoc)$/.test(this.target) && logServerPort && (typeof logServerPort !== 'number' || logServerPort < 1 || logServerPort > 65535)) {
-			logger.error(__('Invalid <log-server-port> found in the tiapp.xml: %s', logServerPort) + '\n');
+		if (!/^dist-(appstore|adhoc)$/.test(this.target) && logServerPort && (typeof logServerPort !== 'number' || logServerPort < 1024 || logServerPort > 65535)) {
+			logger.error(__('Invalid <log-server-port> found in the tiapp.xml'));
+			logger.error(__('Port must be a positive integer between 1024 and 65535') + '\n');
 			process.exit(1);
 		}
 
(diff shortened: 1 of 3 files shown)


iphone/cli/commands/_build.js
```diff
@@ -2288,45 +2289,81 @@ iOSBuilder.prototype.determineLogServerPort = function determineLogServerPort(ne
...
+
+			// connect to the port and see if it's a Titanium app...
+			//  - if the port is bound by a Titanium app with the same appid, then assume
+			//    that when we install new build, the old process will be terminated
+			//  - if the port is bound by another process, such as MySQL on port 3306,
+			//    then we will fail out
+			//  - if the port is bound by another process that expects data before the
+			//    response is returned, then we will just timeout and fail out
+			client = net.connect({
+					host: 'localhost',
```
hansemannn: Can we move the `host` to an own variable instead of hardcoding `localhost` two times?

iphone/cli/commands/_build.js
```diff
@@ -2288,45 +2289,81 @@ iOSBuilder.prototype.determineLogServerPort = function determineLogServerPort(ne
...
+						var headers = JSON.parse(data.toString().split('\n').shift());
+						if (headers.appId !== _t.tiapp.id) {
+							_t.logger.error(__('Another Titanium app "%s" is currently running and using the log server port %d', headers.appId, _t.tiLogServerPort));
+							_t.logger.error(__('Stop the running Titanium app, then rebuild this app'));
+							_t.logger.error(__('-or-'));
+							_t.logger.error(__('Set a unique <log-server-port> between 1024 and 65535 in the <ios> section of the tiapp.xml') + '\n');
+							process.exit(1);
+						}
+					} catch (e) {
+						die();
```
hansemannn: We should log something here, although the server died.

hansemannn (changes requested):

iphone/cli/commands/_build.js
cb1kenobi: No. It's fine.

hansemannn (approved):

iphone/cli/commands/_build.js
cb1kenobi: Why? If it was a variable, that would imply that the host could change, but it never will. It will always be `'localhost'`. I set `host` to `'localhost'` for readability despite not setting a host defaults to `'localhost'` anyways: https://nodejs.org/dist/latest-v6.x/docs/api/net.html#net_socket_connect_options_connectlistener.

cheekiatng:
CR and FT Passed. APPROVED

PR #8520, opened 2016-10-18
[TIMOB-23684] (6_0_X) iOS: Remove unsupported method, update docs

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23684


vijaysingh-axway (approved):
FT and CR approved.

PR #8526, opened 2016-10-19
[TIMOB-16800] iOS: Update default properties in rowData (parity)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16800


sgtcoolguy:
This looks perfect for our unit tests. We already have plenty of tests that verify the default values of properties for various proxies, we should add a test (or re-enable an existing that) that verifies these defaults.

vijaysingh-axway (approved):
FT and CR passed.

PR #8530, opened 2016-10-21
[TIMOB-24053] Android: Fix Ti.UI.TextField passwordMask when not editable

garymathews:
- Allow `passwordMask` to be set correctly when `editable` is set to `false`
###### TEST CASE

``` Javascript
var win = Ti.UI.createWindow({backgroundColor: 'grey', layout: 'vertical'}),
    txt = Ti.UI.createTextField({value: 'hello', editable: false, passwordMask: true}),
    btn = Ti.UI.createButton({title: 'SHOW/HIDE'});

    txt.passwordMask = !txt.passwordMask;
});

win.add(txt);
win.add(btn);
win.open();
```

[JIRA TIcket](https://jira.appcelerator.org/browse/TIMOB-24053)


sgtcoolguy:
I think the description is backwards, this fix is for when `editable` is `false`. In any case, I verified the broken behavior on an old SDK, and the fix with this PR. LGTM 👍

PR #8534, opened 2016-10-24
[TIMOB-16100] iOS9: TextField with passwordMask & fontFamily changes its fontSize

vijaysingh-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16100

**Optional Description:**

[TIMOB-16100] Position of cursor was not resetting while unmasking as end of document calculation goes wrong in iOS below 10 . That made fix by resetting textfield value to blank character and then setting to its actual value.

### DIFF
--- a/iphone/Classes/TiUITextField.m
+++ b/iphone/Classes/TiUITextField.m
@@ -305,6 +305,13 @@
 		NSNotificationCenter * theNC = [NSNotificationCenter defaultCenter];
 		[theNC addObserver:self selector:@selector(textFieldDidChange:) name:UITextFieldTextDidChangeNotification object:textWidgetView];
 	}
+  
+    if (![TiUtils isIOS10OrGreater]) {
+        NSString *str = ((UITextField *)textWidgetView).text;
+        ((UITextField *)textWidgetView).text = @"";
+        ((UITextField *)textWidgetView).text = str;
+    }
+    
 	return textWidgetView;
 }
 
--- a/iphone/Classes/TiUITextField.m
+++ b/iphone/Classes/TiUITextField.m
@@ -306,16 +306,16 @@
 		[theNC addObserver:self selector:@selector(textFieldDidChange:) name:UITextFieldTextDidChangeNotification object:textWidgetView];
 	}
   
-    // TIMOB-16100: Native issue that prevents the textfield to mutate the font-config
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUITextField.m
```diff
@@ -305,6 +305,13 @@ - (void) dealloc
 		NSNotificationCenter * theNC = [NSNotificationCenter defaultCenter];
 		[theNC addObserver:self selector:@selector(textFieldDidChange:) name:UITextFieldTextDidChangeNotification object:textWidgetView];
 	}
+  
+    if (![TiUtils isIOS10OrGreater]) {
+        NSString *str = ((UITextField *)textWidgetView).text;
+        ((UITextField *)textWidgetView).text = @"";
+        ((UITextField *)textWidgetView).text = str;
+    }
```
hansemannn: Please update the code to the snippet we talked about today and migrate the indentation to tabs in this example, because it uses legacy-code. For new classes, we always use spaces, but in this case it's for consistency.

hansemannn (changes requested):
Left one command. Please also rename this PR to match the title of the JIRA ticket. "Issue fix" is not very describing. Thanks! 🚀

iphone/Classes/TiUITextField.m
```diff
@@ -305,6 +305,17 @@ - (void) dealloc
...
+  
+    // TIMOB-16100: Native issue that prevents the textfield to mutate the font-config
+    BOOL needsAdjustment = (![TiUtils isIOS10OrGreater] && ((UITextField *)textWidgetView).secureTextEntry);
+    
+    if (needsAdjustment)
+    {
+        NSString *str = ((UITextField *)textWidgetView).text;
+        ((UITextField *)textWidgetView).text = @"";
+        ((UITextField *)textWidgetView).text = str;
+    }
```
hansemannn: Please adjust the indentation to the one used in this source (it looks like the legacy tabs have been used). The rule it that we use tabs for old code that uses it and spaces for new code that already uses that. Thx!

hansemannn (changes requested):

sgtcoolguy:
Can we expose this bug in a unit test? i.e. if we create a TextField with a password mask and a specific font size/family, can we programmatically expose the incorrect font size to prove this bug fix? Or is this just a visual thing we can test for?

hansemannn (approved):

hansemannn:
PR **approved**!

PR #8537, opened 2016-10-24
[TIMOB-24057] Properly set SDK for Simulator builds

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24057

**Optional Description:**
Sets the SDK to iphonesimulator so the build setting EFFECTIVE_PLATOFRM_NAME will be set correctly on Sim builds (used by CocoaPods in Hyperloop enabled projects)

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -6225,6 +6225,9 @@ iOSBuilder.prototype.invokeXcodeBuild = function invokeXcodeBuild(next) {
 		// so that it can compile all targets (phone and watch targets) for the simulator
 		args.push('-destination', "platform=iOS Simulator,id=" + this.simHandle.udid + ",OS=" + appc.version.format(this.simHandle.version, 2, 2));
 		args.push('ONLY_ACTIVE_ARCH=1');
+		// We also need to set the SDK to iphonesimulator so the build setting EFFECTIVE_PLATOFRM_NAME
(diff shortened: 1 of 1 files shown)


hansemannn:
I'd rather have this one reviewed by @cb1kenobi to ensure it's integrity. Thx!

cb1kenobi:
I'm not certain why we need this, but it doesn't appear to hurt anything. In fact, we used to specify `-sdk` back in the old Python build system.

**APPROVED**

janvennemann:
Just a heads up on why we need this: CocaPods uses shell script build phases that require `EFFECTIVE_PLATOFRM_NAME` to be set properly. Without specifying `-sdk` that build setting always has the value `iphone` and the build phases would fail for simulator builds (due to wrong Paths that are resolved based on `EFFECTIVE_PLATOFRM_NAME`).

cb1kenobi:
@janvennemann Thanks for reminding me about this. I've reverted this PR as it breaks building watch apps. We cannot use `-sdk`. When this is present in the xcodebuild command, we get:

    [ERROR] Error details: target specifies product type 'com.apple.product-type.watchkit2-extension', but there's no such product type for the 'iphonesimulator' platform

The correct SDK value when building the watch extension should be `watchsimulator`, but then that would break building the ios app. This is why I had to introduce a scheme and `-destination`.

You'll either need to fix the xcodebuild command to continue to support watch apps, or figure out another hack to make cocoapods happy.

PR #8539, opened 2016-10-24
Include Titanium.Android.Activity onIntent event documentation

garymathews:
- Include documentation for `Titanium.Android.Activity` `onIntent` event

### DIFF
--- a/apidoc/Titanium/Android/Activity.yml
+++ b/apidoc/Titanium/Android/Activity.yml
@@ -294,6 +294,14 @@ events:
         type: Titanium.Android.Intent
     since: '3.0.0'
 
+  - name: onIntent
+    summary: Fired when the activity is launched.
+    properties:
+      - name: intent
+        summary: The `Intent` used to launch the Activity.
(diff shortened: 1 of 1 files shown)


apidoc/Titanium/Android/Activity.yml
```diff
@@ -294,6 +294,14 @@ events:
         type: Titanium.Android.Intent
     since: '3.0.0'
 
+  - name: onIntent
```
hansemannn: Is the event called `onIntent`? This will cause problems with Alloy, since it will be `onOnintent=""` there. Can we consider a better name here?

hansemannn: @sgtcoolguy @cheekiatng thoughts please. I will merge it for now so we have it in the docs, but we should file a ticket to migrate all "on<event-name>" events to proper names so they don't conflict with Alloy and our usual naming conventions regarding events.

hansemannn:
Approved!

PR #8538, opened 2016-10-24
[TIMOB-24057] Properly set SDK for Simulator builds (6_0_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24057

**Optional Description:**
Sets the SDK to iphonesimulator so the build setting EFFECTIVE_PLATOFRM_NAME will be set correctly on Sim builds (used by CocoaPods in Hyperloop enabled projects)

PR #8544, opened 2016-10-25
Revert "[TIMOB-24057] Properly set SDK for Simulator builds (6_0_X)"

cb1kenobi:
Reverts appcelerator/titanium_mobile#8538


cb1kenobi:
Need to revert because when building apps with watch apps, we get:

    [ERROR] Error details: target specifies product type 'com.apple.product-type.watchkit2-extension', but there's no such product type for the 'iphonesimulator' platform

PR #8547, opened 2016-10-26
[TIMOB-23969] Fix exitOnClose default value

garymathews:
- Fix default behaviour of `exitOnClose`
- Documentation states `Default: true if this is the first window launched else false`
- `exitOnClose` must be manually set to `false` for the app to resume to its previous state
###### TEST CASE
1. Open app
2. Press back
3. Open app, should re-launch to root activity and **NOT** resume from previous state

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -862,7 +862,7 @@ public abstract class TiBaseActivity extends AppCompatActivity
 			// override back press to background the activity
 			// note: 2 since there should always be TiLaunchActivity and TiActivity
 			if (TiApplication.activityStack.size() <= 2) {
-				if (topWindow != null && !TiConvert.toBoolean(topWindow.getProperty(TiC.PROPERTY_EXIT_ON_CLOSE), false)) {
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
Tested with the following app.js:
```javascript
	backgroundColor:'white'
});
 
var lbl = Ti.UI.createLabel({
	text: "See me disappear",
	color:'black',
	font:{
		fontSize:20
	}
});
 
setTimeout(function(){
    lbl.hide();
}, 3000);
 
 
win.add(lbl);
win.open();
```

Opening app, label disappears after 3 seconds, hitting android back button leaves app. Re-opening app, the label is back and then disappears. Explicitly setting `exitOnClose: false` on the window: when re-opening the app the label is always gone. Looks good to me.

Then tested with second window:
```js
	backgroundColor:'white'
});

var lbl = Ti.UI.createLabel({
	text: "See me disappear",
	color:'black',
	font:{
		fontSize:20
	}
});

win.add(lbl);
win.open();

	backgroundColor:'green',
	exitOnClose: true
});

	text: "Hi there",
	color:'black',
	font:{
		fontSize:20
	}
});

win2.add(lbl2);

setTimeout(function(){
    lbl.hide();
    win2.open();
}, 3000);
```

Here I expected that pressing back after the second green window opened would close the app (it does). Re-opening the app, I should see the label and then the green window should re-appear (it does).

Toggling `exitOnClose` to `false` for the second window (or removing the property altogether), I expected hitting back on green window to take me to the white window with no label.

*Approved* 👍

PR #8549, opened 2016-10-26
[TIMOB-23634] IOS: Toolbar on TextField doesn't hide smoothly when calling blur

cheekiatng:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23634

A better approach should be to use NSConstraints in auto layout, or using UIInputView. 
But this approach is basically to make the toolbar animation asynchronous to the keyboard dismissal and appearance.


hansemannn:
Cannot accept at this time.

PR #8557, opened 2016-10-27
[TIMOB-14285] Fix kroll-v8::nativeSetProperty() to set accessor

garymathews:
- `nativeSetProperty()` does not set an accessor for custom properties, making them unavailable in JavaScript

###### TEST CASE

``` Javascript
win.open();
win.applyProperties({customString: 'TEST'});
Ti.API.info('customString: ' + win.customString);
// customString should output 'TEST'
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-14285)


sgtcoolguy:
Can we add a unit test for this? Do other platforms support this? I think Windows does...

garymathews:
I'll add test, but this does not only affect `applyProperties`. Any internal calls to `setProperty` would fail if it is to set properties that are not defined by a Titanium proxy.

garymathews:
Test case: https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/11

sgtcoolguy:
test this please

garymathews:
Tests are passing 👍

PR #8559, opened 2016-10-27
[TIMOB-24079] Android Debugger: Not all breakpoints are hit during debug of android app with SDK 6.0.0

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24079

**Description:**

In Ti SDK 6+, I fixed require behavior and internally we use paths with a leading '/' in them. This works great for require, but I never touched Ti.include. As a result, debug breakpoints don't behave properly when Ti.include is used. This fix uses the leading slash for the filepath it builds internal when executing Ti.include.


garymathews:
CR: **PASS**
FT: **PASS**

👍 

Jenkins is failing due to an unrelated test case, merging.

PR #8562, opened 2016-10-29
[TIMOB-3408] iOS: Support justified text-align in Labels, Buttons and TextFields

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-3408


iphone/Classes/TiUIButton.m
```diff
@@ -412,26 +412,9 @@ -(void)setVerticalAlign_:(id)align
 -(void)setTextAlign_:(id)align
 {
     button = [self button];
-    NSTextAlignment alignment = NSTextAlignmentNatural;
```
sgtcoolguy: Note that this actually does change behavior here. If the user specified null/undefined/ or a bad int value, we'd default to NSTextAlignmentNatural, while we can no longer specify any value to get that setting now. We may want to expose a "natural" constant value as well?

sgtcoolguy (changes requested):
Also, it'd be good to add/edit a unit test to validate that the new constant is a valid value to assign to these properties. (i.e. assign it and check the getter? Validate the constant lists and is actually a constant (not writable)/has the value we expect?)

iphone/Classes/TiUIButton.m
hansemannn: Agree. I'd rather change [this constant](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiUtils.m#L1354) to `NSTextAlignmentNatural`  though. They look the same, but `NSTextAlignmentNatural` is the native way to go and default since iOS 6.

vijaysingh-axway (approved):
FT and CR passed .

sgtcoolguy (approved):

PR #8563, opened 2016-10-29
[TIMOB-23185] Docs: Add Ti.include removal to docs

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23185


sgtcoolguy:
@cheekiatng So, as part of my PR to remove Ti.include, I actually removed the doc entry for it entirely. Do we want to add it back in and have this note that @hansemannn wrote? I understand we want people to know how to migrate, but it's been deprecated for a very long time and it feels odd to me we'd have a doc entry for API that no longer exists. This feels like a blog post/release note issue to me.

PR #8565, opened 2016-10-30
[TIMOB-24085] iOS: Add ListItem subtitle colors

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24085

### DIFF
--- a/apidoc/Titanium/UI/ListItem.yml
+++ b/apidoc/Titanium/UI/ListItem.yml
@@ -226,13 +226,38 @@ properties:
     type: String
     accessors: false
     
+  - name: subtitleColor
+    summary: Default text color of the subtitle, as a color name or hex triplet.
+    description: |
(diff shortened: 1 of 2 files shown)


hansemannn:
@sgtcoolguy I thought about unit-tests but couldn't think of one that makes sense here, since we "just" fill a dictionary of template values that could be anything else as well without crashing - since the only result can be seen in the UI. Hard to describe.

sgtcoolguy:
I mean in this case the unit test would basically just be ensuring that the API exists- that the property exists on an instance of a ListItem for iOS, and that you can set String color values on it. Presumably you could also add some validation testing that it doesn't crash if you pass in null/undefined/integer/non-color string.

vijaysingh-axway (approved):
FT and CR passed.

PR #8568, opened 2016-10-31
[TIMOB-24086] iOS: Label doesn't fire link event correctly

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24086

### DIFF
--- a/iphone/Classes/TiUILabel.m
+++ b/iphone/Classes/TiUILabel.m
@@ -225,96 +225,86 @@
     }
 }
 
+// Code taken from https://github.com/AliSoftware/OHAttributedStringAdditions
+
+- (NSTextContainer*)currentTextContainer
+{
+    NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:self.label.bounds.size];
+    textContainer.lineFragmentPadding  = 0;
+    textContainer.maximumNumberOfLines = (NSUInteger)self.label.numberOfLines;
+    textContainer.lineBreakMode = self.label.lineBreakMode;
+    return [textContainer autorelease];
+}
+
+- (NSUInteger)characterIndexAtPoint:(NSMutableAttributedString*)theString atPoint:(CGPoint)point
+{
+    NSTextContainer* textContainer = self.currentTextContainer;
+    NSTextStorage *textStorage = [[NSTextStorage alloc] initWithAttributedString:theString];
+    
+    NSLayoutManager *layoutManager = [NSLayoutManager new];
+    [textStorage addLayoutManager:layoutManager];
+    [layoutManager addTextContainer:textContainer];
+    
+    // UILabel centers its text vertically, so adjust the point coordinates accordingly
+    NSRange glyphRange = [layoutManager glyphRangeForTextContainer:textContainer];
+    CGRect wholeTextRect = [layoutManager boundingRectForGlyphRange:glyphRange
+                                                    inTextContainer:textContainer];
+    point.y -= (CGRectGetHeight(self.bounds)-CGRectGetHeight(wholeTextRect))/2;
+    
+    // Bail early if point outside the whole text bounding rect
+    if (!CGRectContainsPoint(wholeTextRect, point)) {
+        RELEASE_TO_NIL(textStorage);
+        RELEASE_TO_NIL(layoutManager);
+        return NSNotFound;
+    }
+    
+    // ask the layoutManager which glyph is under this tapped point
+    NSUInteger glyphIdx = [layoutManager glyphIndexForPoint:point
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
ok to test

iphone/Classes/TiUILabel.m
```diff
@@ -225,96 +225,77 @@ -(void)handleListenerRemovedWithEvent:(NSString *)event
...
     
-    if (idx != NSNotFound) {
-        if(idx > theString.string.length) {
-            return NO;
-        }
-        NSRange theRange = NSMakeRange(0, 0);
-        NSString *url = nil;
+    NSRange theRange = NSMakeRange(0, 0);
+    NSString *url = nil;
+    NSString *url2 = nil;
```
hansemannn: Is this used somewhere?

iphone/Classes/TiUILabel.m
```diff
@@ -225,96 +225,77 @@ -(void)handleListenerRemovedWithEvent:(NSString *)event
...
+    if (url == nil || url.length == 0) {
+        return NO;
     }
-    return NO;
+        
+    NSDictionary *eventDict = [NSDictionary dictionaryWithObjectsAndKeys:
+                                url, @"url",
+                                [NSArray arrayWithObjects:NUMUINTEGER(theRange.location), NUMUINTEGER(theRange.length),nil],@"range",
+                                    nil];
+    [[self proxy] fireEvent:@"link" withObject:eventDict propagate:NO reportSuccess:NO errorCode:0 message:nil];
```
hansemannn: Add a check to see if the event-listener is added, e.g. `[[self proxy] _hasListeners:@"link"]`. It may be checked even before to prevent unnecessary memory usage. Maybe in the method call inside `recognizedTap:`. 

And finally, I noticed the method returns a bool but it's callee does not use the return-value. I never worked on that API before, can you think of we that was done?

iphone/Classes/TiUILabel.m
```diff
@@ -225,96 +225,77 @@ -(void)handleListenerRemovedWithEvent:(NSString *)event
...
-    thePoint.y = (drawRect.size.height - thePoint.y);
-    CTFramesetterRef theRef = CTFramesetterCreateWithAttributedString((__bridge CFAttributedStringRef)theString);
-    if (theRef == NULL) {
-        return;
-    }
+    NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:self.label.bounds.size];
+    textContainer.lineFragmentPadding  = 0;
+    textContainer.maximumNumberOfLines = (NSUInteger)self.label.numberOfLines;
+    textContainer.lineBreakMode = self.label.lineBreakMode;
+    return textContainer;
```
hansemannn: Still will probably leak, did you run the Analyser already?

hansemannn (changes requested):
Added some comments. Please also provide the test-case from the video in the JIRA. 

Please check this leaks as well:

screen shot 2016-11-01 at 12 15 21

**EDIT**: Proposal for fixing the leaks (untested so far): https://gist.github.com/hansemannn/eb81b09926a64b208268741da4cf3711

iphone/Classes/TiUILabel.m
kopiro: Analyser?

iphone/Classes/TiUILabel.m
kopiro: Nope, sorry, typo.

iphone/Classes/TiUILabel.m
kopiro: I'll see that.

kopiro: @hansemannn Is already checked on the first line of `recognizedTap`:

```
    BOOL testLink = (label != nil) &&([(TiViewProxy*)[self proxy] _hasListeners:@"link" checkParent:NO]);
```

kopiro: @hansemannn Yes, its boolean value is never used (NOW).
But I think that we should keep the method signature as a bool just... *because we can*.

iphone/Classes/TiUILabel.m
hansemannn: Xcode Analyzer, great tool to find memory leaks in your code without running the app:
ev3wb

hansemannn:
@kopiro Let me know when you applied the changes, the FT works fine!

kopiro:
@hansemannn Implemented but there was a `[textStorage release]` that bricked thinks.

Implemented as a `RELEASE_TO_NIL` before return.

Tested.

hansemannn:
Ohh, nice! PR approved 🚀

PR #8572, opened 2016-11-01
TIMOB-19629 Android: user defined letter case for button text

frankieandone:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-19629

### DIFF
--- a/android/templates/AndroidManifest.xml
+++ b/android/templates/AndroidManifest.xml
@@ -13,7 +13,7 @@
 		<!-- TI_APPLICATION -->
 
 		<activity android:name=".${config['classname']}Activity"
-			android:label="${config['appname']}" android:theme="@style/Theme.Titanium"
+			android:label="${config['appname']}"
 			android:configChanges="keyboardHidden|orientation">
 			<intent-filter>
 				<action android:name="android.intent.action.MAIN" />


apidoc/Titanium/UI/Button.yml
```diff
@@ -290,6 +290,13 @@ properties:
     constants: Titanium.UI.TEXT_ALIGNMENT_*
     default: <Titanium.UI.TEXT_ALIGNMENT_CENTER>
 
+  - name: textAllCaps
+    summary: |
+      All text to upper case letters, specified using true or false.
+    type: [boolean]
```
hansemannn: Use `type: Boolean`

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
```diff
@@ -130,6 +131,12 @@ public void processProperties(KrollDict d)
...
 			btn.setShadowLayer(shadowRadius, shadowX, shadowY, shadowColor);
 		}
+		if (d.containsKey(TiC.PROPERTY_TEXT_ALL_CAPS)){
+			boolean caps = TiConvert.toBoolean(d.get(TiC.PROPERTY_TEXT_ALL_CAPS));
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
+				btn.setAllCaps(caps);
+			}
+		}
 		btn.invalidate();
 	}
```
garymathews: You should also add a similar setter to `propertyChanged`

```Java
...
} else if (key.equals(TiC.PROPERTY_TEXT_ALL_CAPS)) {
    boolean caps = TiConvert.toBoolean(d.get(TiC.PROPERTY_TEXT_ALL_CAPS), false);
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
        btn.setAllCaps(caps);
    }
...
```

garymathews:
So looking at the ticket, the issue is to do with `android:textAllCaps` being ignored. Exposing `textAllCap` as a Titanium API does not fix this issue.

fmerzadyan:
There are 3 fixes available which this is one #1.
1. custom theme set in application tag in tiapp.xml was overridden by default theme set in activity tag. removed default theme attribute in activity in the template.  
2. user explicitly changes their tiapp.xml so that their application has custom theme set and ALSO the activity theme set.
3. expose textAllCaps and programmatically set caps to false which was already pushed and PR'd but was rejected because it didn't address the xml.

garymathews:
Superseded by https://github.com/appcelerator/titanium_mobile/pull/8592

PR #8576, opened 2016-11-02
[TIMOB-23891] iOS: Place feedback generator correctly

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23891


hansemannn:
Approved!

PR #8583, opened 2016-11-04
[TIMOB-23936](6_0_X) :  Ti.Media.showCamera() with overlay bug on iOS 10

vijaysingh-axway:
JIRA: https://jira.appcelerator.org/browse/TIMOB-23936


hansemannn:
PR approved! Jenkins just fails because of two unit tests that are not in the 6_0_X branch. We may need to think about how we could use branch-based unit-tests like i talked about with @sgtcoolguy.

PR #8584, opened 2016-11-04
[TIMOB-24087] iOS: Ti.Network.registerForPushNotifications does not c…

vijaysingh-axway:
JIRA:  https://jira.appcelerator.org/browse/TIMOB-24087


iphone/Classes/defines.h
```diff
@@ -132,6 +132,7 @@
 #define USE_TI_UIREFRESHCONTROL
 #define USE_TI_UIIOSAPPLICATIONSHORTCUTS
 #define USE_TI_UIIOSBLURVIEW
+#define USE_TI_NETWORKREGISTERFORPUSHNOTIFICATIONS
```
hansemannn: This does not have any effect on the device builds. The constants defined in the `defines.h` are used for simulator builds only. For device builds, they get generated based on the used symbols. So if you use `Ti.Network.registerForPushNotifications()`, it will generate the above constant in the CLI (right [here](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L5779)).

But I'm still wondering how this could have fixed the issue for you.

hansemannn (changes requested):

hansemannn:
Will still merge to ease the debug with this API. Note that this still only effect Xcode debug builds, not builds from the compiled SDK.

hansemannn (approved):

PR #8587, opened 2016-11-07
[TIMOB-23885] Do not disable HW acceleration for WebView

garymathews:
- Do not disable hardware acceleration for `Titanium.UI.WebView`
- Hardware acceleration is needed for HTML5 embedded video to decode

###### TEST CASE
```Javascript
var win = Titanium.UI.createWindow({layout : 'vertical'}),
	webView = Titanium.UI.createWebView({
		url : 'http://edition.cnn.com/videos',
		borderRadius : 1, // this causes HW acceleration to be disabled
		enableZoomControls : false,
		width: Ti.UI.FILL,
		height: Ti.UI.FILL
	});

win.add(webView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23885)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
@@ -958,4 +958,10 @@ public class TiUIWebView extends TiUIView
 	{
 		return chromeClient.interceptOnBackPressed();
 	}
+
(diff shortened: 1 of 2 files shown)


fritzzetik:
Do I understand?
So if I remove borderRadius : 1, (as advised by support) from my code it would work?

garymathews:
@fritzzetik Yes, I believe so

PR #8588, opened 2016-11-07
[TIMOB-24110] iOS: Expose keyboardDisplayRequiresUserAction in Ti.UI.WebView

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24110

When this property is set to true, the user must explicitly tap the elements in the web view to display the keyboard (or other relevant input view) for that element. When set to false, a focus event on an element causes the input view to be displayed and associated with that element automatically.
The default value for this property is true.

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -673,8 +673,10 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 
 - (void)setKeyboardDisplayRequiresUserAction_:(id)value
 {
+    ENSURE_TYPE(value, NSNumber);
     [[self proxy] replaceValue:value forKey:@"keyboardDisplayRequiresUserAction" notification:NO];
-    webview.keyboardDisplayRequiresUserAction = [TiUtils boolValue:value def:YES];
+    
+    [[self webview] setKeyboardDisplayRequiresUserAction:[TiUtils boolValue:value def:YES]];
 }
 
 #pragma mark WebView Delegate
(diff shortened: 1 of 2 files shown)


hansemannn:
ok to test

iphone/Classes/TiUIWebView.m
```diff
@@ -671,6 +671,12 @@ -(CGFloat)contentWidthForWidth:(CGFloat)value
     return ret;
 }
 
+- (void)setKeyboardDisplayRequiresUserAction_:(id)value
+{
+    [[self proxy] replaceValue:value forKey:@"keyboardDisplayRequiresUserAction" notification:NO];
+    webview.keyboardDisplayRequiresUserAction = [TiUtils boolValue:value def:YES];
+}
```
hansemannn: Implement like this:
```objc
- (void)setKeyboardDisplayRequiresUserAction_:(id)value
{
    ENSURE_TYPE(value, NSNumber);
    [[self proxy] replaceValue:value forKey:@"keyboardDisplayRequiresUserAction" notification:NO];

    [webview setKeyboardDisplayRequiresUserAction:[TiUtils boolValue:value def:YES]];
}
```

hansemannn (changes requested):

iphone/Classes/TiUIWebView.m
hansemannn: And please implement the getter in the proxy + add unit tests in [titanium-mobile-mocha-suite](https://github.com/appcelerator/titanium-mobile-mocha-suite)

hansemannn (approved):

hansemannn:
Added a unit-test and updated the parameter validation. CR + FT passed, PR **approved**! 🚀

PR #8592, opened 2016-11-08
[TIMOB-19629] Use application theme as parent for Theme.Titanium

garymathews:
- When an application theme is defined in `tiapp.xml` make sure the parent of `Theme.Titanium` is set as this theme

##### tiapp.xml
```XML
...
    <manifest android:versionCode="1" android:versionName="1.0">
        <application android:theme="@style/Theme.Test"/>
    </manifest>
</android>
...
```

##### platform/android/res/values/test.xml
```XML
<resources xmlns:android="http://schemas.android.com/apk/res/android">
    <style name="Theme.Test" parent="@style/Theme.AppCompat">
        <item name="android:textAllCaps">false</item>
    </style>
</resources>
```

##### app.js
```Javascript
	btn = Ti.UI.createButton({title: 'lower case'}),
	lv = Ti.UI.createListView();

lv.appendSection(Ti.UI.createListSection({
	headerView: btn
}));
win.add(lv);
win.open();
```

For Android 5.0+, the `lower case` title of the button should NOT be uppercase.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19629)

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -3384,6 +3384,12 @@ AndroidBuilder.prototype.generateTheme = function generateTheme(next) {
 		if (this.tiapp.fullscreen || this.tiapp['statusbar-hidden']) {
 			flags += '.Fullscreen';
 		}
+		if (this.tiappAndroidManifest && this.tiappAndroidManifest.application && this.tiappAndroidManifest.application.theme) {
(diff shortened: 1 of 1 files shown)


fmerzadyan:
The fix only works on devices with Android API 20 and up.

PR #8599, opened 2016-11-11
[TIMOB-23139] (6_0_X) iOS: Fix image cache from not being retained

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24139


cheekiatng:
CR and FT Passed! APPROVED!

PR #8601, opened 2016-11-11
[TIMOB-24140] (6_0_X) iOS: Fix audio session category macros

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24140


vijaysingh-axway (approved):
FT and CR Passed .

PR #8608, opened 2016-11-16
[TIMOB-24148] iOS: Add note for CommonJS modules colliding with Ti namespaces

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24148

### DIFF
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -860,6 +860,16 @@ CFMutableSetRef	krollBridgeRegistry = nil;
 	// If no such module exists, bail out!
 	if (moduleClass == nil) {
 		return nil;
+    
+	}
+    
+    // If there is a core module that collides with the given path,
+    // throw an error log
+    NSURL *jsPath = [NSURL URLWithString:[NSString stringWithFormat:@"%@/%@.js", [[NSURL fileURLWithPath:[TiHost resourcePath] isDirectory:YES] path], path]];
+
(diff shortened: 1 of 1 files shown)


cb1kenobi:
+1

sgtcoolguy:
I'd tweak this a little bit. The wording seems to indicate that there was a failure and to always use the './' prefix.

What we want the user to know is that there is a clash between a native module with the given id and a JS file with the module id as it's base name - and that we're giving preference to the native module. If they intended to address the JS file, they should use the prefix to make it explicit.

hansemannn:
> What we want the user to know is that there is a clash between a native module with the given id and a JS file with the module id as it's base name - and that we're giving preference to the native module. If they intended to address the JS file, they should use the prefix to make it explicit.

I have to confess that it's hard for me to find the correct wording on this. As the use-case only effects a very specific kind of local CommonJS modules with *no* sub-directories *and* clashing names with core-modules, the suggested `./` prefix would always make sense for that kind of file. 

> I'd also worry that even if they fix their usage but have a JS file whose base name matches a native module id they'll keep seeing this message - and we don't spell out for them that to avoid the message they need to effectively rename the js file 

Native modules won't have a JS file in the app like CommonJS modules have, so the won't collide. I can setup a separate test-case for that if you want

> (Which may be another reason why we'd want this at a WARNING level instead of ERROR)

Agree, I'll change it to `[WARN]`

sgtcoolguy:
Merged manually by copying the changes and tweaking wording some (and bumping down to WARN).

PR #8611, opened 2016-11-16
[AC-4619] Wording Update View.yml on hiddenBehavior property

TiBaharroth:
Maybe we should fix HiddenBahaviour by HiddenBehaviour ?


hansemannn:
Approved!

PR #8613, opened 2016-11-16
[TIMOB-24152] iOS: Fix Ti.Media.saveToPhotoGallery macro for device-builds

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24152


vijaysingh-axway (approved):
FT and CR passed.

PR #8615, opened 2016-11-16
[TIMOB-24153] iOS: Add parity for Ti.UI.TextField.padding

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24153


vijaysingh-axway (approved):
FT and CR approved .

PR #8617, opened 2016-11-17
[TIMOB-24095] Android: Set Ti.UI.WebView.html as writable

garymathews:
- Allow `html` property to be set

###### TEST CASE
```Javascript
var w = Ti.UI.createWindow(),
    wv = Ti.UI.createWebView({scalesPageToFit: true});

wv.html = '<html><body style="background-color:black"><img src="https://goo.gl/VyjZWo" width="100%"></body></html>';

w.add(wv);
w.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24095)


fmerzadyan:
CR/FR: PASS

PR #8623, opened 2016-11-17
[TIMOB-23997] Reverted bad CommonJS module logic

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23997

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -4,7 +4,7 @@
  * @module cli/_build
  *
  * @copyright
- * Copyright (c) 2009-2015 by Appcelerator, Inc. All Rights Reserved.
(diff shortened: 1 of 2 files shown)


cb1kenobi:
**APPROVED**

sgtcoolguy:
@cb1kenobi This has broken our support for commonJS modules that don't use the legacy naming where there has to be a JS file named the exact same as the module id. I specifically did not approve the PR because of this concern. This has broken our test suite which validates that we can handle nom-style modules. I need to revert this.

cb1kenobi:
@sgtcoolguy Correct. Your PR should have never been merged due to back logic. I submitted a fix, but it was never approved, so I reverted it all back to pre-6.0.0. I'm happy you re-submitted #8648 and #8649 so now you can fix it to be correct.

PR #8624, opened 2016-11-18
Replace Ti.UI.iPhone with Ti.UI.iOS

HazemKhaled:
Just update documentation, still need JIRA ticket and test case?


hansemannn:
Yeah! Wondering that there still are some occurrences of it, although I did some pretty extended search-and-replace operations, haha. Will scan it once again, but merge this before. Thx!

PR #8629, opened 2016-11-18
[TIMOB-23992] iOS: Fix text-selection on devices

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23992


vijaysingh-axway (changes requested):
As I verified using - 
```js
    url: "http://google.com",
    disableContextMenu: true
});
win.add(webView);
win.open();
```

On selection of text , menu popups e.g copy, select all, share etc or getting open . 
Changing value of  disableContextMenu, do not have any effect . 
I think something missing. Please check.

hansemannn:
Applied changes as discussed, thx!

vijaysingh-axway (approved):
FT and CR passed.

PR #8631, opened 2016-11-21
[TIDOC-2697] Docs: AndroidView Property is Creation Only

TiBaharroth:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-2697

IMHO androidView property is creation only :
1) No get/set
2) Set this property after the creation do nothing.


hansemannn:
Makes sense, thx! 🚀

PR #8635, opened 2016-11-25
[TIMOB-24169] : backgroundTopCap + backgroundLeftCap broken on Ti.UI.Button in 6.0.0.GA (Regression)

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24169

### DIFF
--- a/iphone/Classes/TiUIButton.m
+++ b/iphone/Classes/TiUIButton.m
@@ -263,7 +263,7 @@
 {
 	[backgroundImageCache release];
 	RELEASE_TO_NIL(backgroundImageUnstretchedCache);
-	backgroundImageCache = [[TiUtils loadBackgroundImage:value forProxy:[self proxy]] retain];
+    backgroundImageCache = [[self loadImage:value] retain];
     self.backgroundImage = value;
 	[self updateBackgroundImage];
 }
--- a/iphone/Classes/TiUIButton.m
+++ b/iphone/Classes/TiUIButton.m
@@ -263,8 +263,8 @@
 {
 	[backgroundImageCache release];
 	RELEASE_TO_NIL(backgroundImageUnstretchedCache);
-    backgroundImageCache = [[self loadImage:value] retain];
-    self.backgroundImage = value;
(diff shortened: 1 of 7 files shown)


iphone/Classes/TiUIButton.m
```diff
@@ -263,7 +263,7 @@ -(void)setBackgroundImage_:(id)value
 {
 	[backgroundImageCache release];
 	RELEASE_TO_NIL(backgroundImageUnstretchedCache);
-	backgroundImageCache = [[TiUtils loadBackgroundImage:value forProxy:[self proxy]] retain];
+    backgroundImageCache = [[self loadImage:value] retain];
     self.backgroundImage = value;
```
hansemannn: Please fix indentation of both lines 266 and 267 to conform with the one of 265.

iphone/Classes/TiUIView.m
```diff
@@ -330,6 +330,7 @@ -(UIImage*)loadImage:(id)image
 		NSLog(@"[WARN] could not find image: %@",image);
 		return nil;
 	}
+    [TiUtils loadBackgroundImage:image forProxy:[self proxy]];
```
hansemannn: Indentation. And I feel that this will cause caching problems, so I need some more time to think about the effect ob all our use-cases.

hansemannn (changes requested):

hansemannn:
We cannot accept this solution at this time. Issues:
- `[TiUtils loadBackgroundImage:image forProxy:[self proxy]];` is called no matter of what the value is before
- The method has a return value (`UIImage`) but it is used without returning the result. I guess you just call it to trigger methods in that method, but then we should rather call those directly
- As said before, this will cause an overhead for results that already return a valid image (which are all local images) and therefore consumes additional memory without any cache-checks

That's why I would rather tend to revert [TIMOB-23666](https://jira.appcelerator.org/browse/TIMOB-23666) for now, since a) remote images can still be fetched before and used a file-reference of the local image and b) merging this PR would affect many more subclasses of `TiUIView`, as mentioned before.

PR #8641, opened 2016-11-29
[6_0_X] [TIMOB-24157] Install Windows certificate in new window

infosia:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24157

When we install Windows Store app, we need to ensure developer certificate installed first, and then it needs to be executed in a new Window to workaround user-input issue in Node.js (described in the ticket).

To see if you already installed developer certificate, launch “Manage user certificates” from control panel and navigate to “Trusted People > Certificates”. “CMake Test Cert” is the developer cert. You can remove it in order to test this issue.

60735_certmgr

```
appc run -p windows --wp-sdk 10.0 --target ws-local -l trace
```

### DIFF
--- a/node_modules/windowslib/CHANGELOG.md
+++ b/node_modules/windowslib/CHANGELOG.md
@@ -1,3 +1,7 @@
+0.4.20.1 (11/29/2016) - Only for SDK 6.0.1
+-------------------
+  * [TIMOB-24157] Install certificate in new window
+
 0.4.20 (8/26/2016)
 -------------------
   * [TIMOB-23816] Fix 8.1 emulator listing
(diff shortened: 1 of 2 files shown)


garymathews:
@infosia missing a couple of files?

infosia:
@garymathews  Yeah, this PR intends to be a "manual fix" since we can not bump up the version simply. We have pushed TIMOB-24157 in 0.4.24 but it includes fixes that is marked "6.1.0" (in 0.4.21 to 0.4.23), so we want to update certain file that is only related to 0.4.24 without bumping up the version. I did not actually bump up the version because I was not sure if we can set the version that does not exist, and I'm guessing we can't use`npm install --production` in this case.

garymathews:
Looks good 👍

PR #8648, opened 2016-12-01
Revert "[TIMOB-23997] Reverted bad CommonJS module logic"

sgtcoolguy:
Reverts appcelerator/titanium_mobile#8500

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -4,7 +4,7 @@
  * @module cli/_build
  *
  * @copyright
- * Copyright (c) 2009-2016 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-2015 by Appcelerator, Inc. All Rights Reserved.
  *
  * Copyright (c) 2012-2013 Chris Talkington, contributors.
  * {@link https://github.com/ctalkington/node-archiver}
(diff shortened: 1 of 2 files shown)


android/cli/commands/_build.js
```diff
@@ -1472,12 +1472,24 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 				if (module.platform.indexOf('commonjs') != -1) {
 					module.native = false;
 
+					// Look for legacy module.id.js first
 					module.libFile = path.join(module.modulePath, module.id + '.js');
 					if (!fs.existsSync(module.libFile)) {
-						this.logger.error(__('Module %s version %s is missing module file: %s', module.id.cyan, (module.manifest.version || 'latest').cyan, module.libFile.cyan) + '\n');
-						process.exit(1);
+						// then package.json TODO Verify the main property points at reale file under the module!
+						module.libFile = path.join(module.modulePath, 'package.json');
```
cb1kenobi: You must not use `module.libFile` to store the path as you run your checks. Use a local variable and when you have a match, then set `module.libFile`.

Currently, if the `<module id>.js` file doesn't exist, then you proceed to check for a `package.json`, which is fine. However, if the module id file doesn't exist and there's no `package.json`, then the libFile will be the `package.json` which is very wrong. `package.json` should never be the "libFile".

Anyways, read the `package.json` and check for a `main` property, then check if the `main` file exists. If the `main` file doesn't exist, then continue on to the `index.js`.

Having a `.json` file as the main module file should not be supported by Titanium. The CommonJS spec does not describe `.json` files as being valid modules. Please remove the `index.json` checks.

android/cli/commands/_build.js
```diff
@@ -1472,12 +1472,24 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
...
+						if (!fs.existsSync(module.libFile)) {
+							// then index.js
+							module.libFile = path.join(module.modulePath, 'index.js');
+							if (!fs.existsSync(module.libFile)) {
+								// then index.json
+								module.libFile = path.join(module.modulePath, 'index.json');
+								if (!fs.existsSync(module.libFile)) {
+									this.logger.error(__('Module %s version %s is missing module files: %s, package.json, index.js, or index.json', module.id.cyan, (module.manifest.version || 'latest').cyan, path.join(module.modulePath, module.id + '.js').cyan) + '\n');
```
cb1kenobi: Again, `.json` files are not valid CommonJS modules.

cb1kenobi: Where does Node.js's implementation say that it treats `package.json` as a module? It is meant to parsed and use the `main` property to locate the real module.

sgtcoolguy:
The failing unit test appears to be an intermittent one related to a Ti.UI.Webview test on iOS (unrelated to this PR).

cb1kenobi:
Regardless, I very much appreciate you spending the time to fix the PR. I'll carefully review it.

cb1kenobi:
CR'd and FR'd. Looks good! Thank you! **APPROVED**

PR #8650, opened 2016-12-01
[AC-4649] Android: Adding getLocationOnScreen() to retrieve location after translation

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-4649


hansemannn (changes requested):
Please add docs, thx!

garymathews:
Thanks, but I believe this is a bug with the `Titanium.View.rect` property. After an animation the `left right top bottom` properties will stay as originally set. But the `rect` `x` and `y` properties should update to the new position.

Adding a new method should not be necessary, `rect` should be fixed to address this issue 👍

m1ga:
I see,thanks. I'll have a look and close this one. May be the parts I've added could fix the `rect` problem already.

PR #7959, opened 2016-04-22
[TIMOB-18053] Android: expose the text color for the picker

sriks:
JIRA: https://jira.appcelerator.org/browse/TIMOB-18053

PR #8651, opened 2016-12-01
[TIMOB-18053] Android: fix picker row text color

frankieandone:
[Last PR](https://github.com/appcelerator/titanium_mobile/pull/7959) for issue didn't solve all bugs. Text color behaved weird if color not set for all rows.

**TEST CASE**
Comment and uncomment colors in each picker row.
```js
    exitOnClose: true,
    layout: 'vertical',
    backgroundColor: 'black'
});

    top: 50
});

var data = [];
data[0] = Ti.UI.createPickerRow({
    title: 'Bananas',
    color: 'yellow'
});
data[1] = Ti.UI.createPickerRow({
    title: 'Strawberries',
    // color: 'red'
});
data[2] = Ti.UI.createPickerRow({
    title: 'Mangos',
    color: 'orange'
});
data[3] = Ti.UI.createPickerRow({
    title: 'Grapes',
    // color: 'purple'
});

picker.add(data);
picker.selectionIndicator = true;

win.add(picker);
win.open();
```
**EXPECTED**
A picker row's text color should only be set if specified otherwise should be default color.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18053

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
@@ -25,7 +25,6 @@ import ti.modules.titanium.ui.PickerColumnProxy;
 import ti.modules.titanium.ui.PickerProxy;
 import android.app.Activity;
 import android.content.Context;
-import android.graphics.Color;
 import android.os.Build;
 import android.view.MotionEvent;
 import android.view.View;
@@ -42,7 +41,7 @@ public class TiUINativePicker extends TiUIPicker
 {
 	private static final String TAG = "TiUINativePicker";
 	private boolean firstSelectedFired = false;
-	private static int DEFAULT_TEXT_COLOR = 0;
+	private static int defaultTextColor = 0;
 	
 	public static class TiSpinnerAdapter<T> extends ArrayAdapter<T>
 	{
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
```diff
@@ -81,9 +81,11 @@ private void styleTextView(int position, TextView tv) {
 		        fontProperties[TiUIHelper.FONT_STYLE_POSITION]);
 		    }
 		    if (rowProxy.hasProperty(TiC.PROPERTY_COLOR)) {
-		        final int color = TiConvert.toColor(rowProxy.getProperties(), TiC.PROPERTY_COLOR);
-		        tv.setTextColor(color);
-		    }
+				final int color = TiConvert.toColor(rowProxy.getProperties(), TiC.PROPERTY_COLOR);
+				tv.setTextColor(color);
+			} else {
+				tv.setTextColor(Color.WHITE);
```
garymathews: Instead of defining the text color as white, it would be better to obtain the default color and set it back to that.

```Java
private static int defaultTextColor = -1;
...
public static class TiSpinnerAdapter<T> extends ArrayAdapter<T>
{
    ...
    private void styleTextView(int position, TextView tv) {
        ...
        if (rowProxy.hasProperty(TiC.PROPERTY_COLOR)) {
            final int color = TiConvert.toColor(rowProxy.getProperties(), TiC.PROPERTY_COLOR);
            // obtain default text color
            if (defaultTextColor == -1) {
                defaultTextColor = tv.getCurrentTextColor();
            }
            tv.setTextColor(color);
        } else {
            tv.setTextColor(defaultTextColor);
        }
    ...
...

fmerzadyan:
Updated.

garymathews:
Looks good, but since `DEFAULT_TEXT_COLOR` is not a constant it should use camel-case instead of snake-case.

garymathews:
CR: **PASS**
FT: **PASS**

PR #8654, opened 2016-12-02
[TIMOB-24142] Only write the entitlements file if there isn't a custom one

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24142


hansemannn:
CR + FT passed, PR **approved**!

PR #8659, opened 2016-12-05
[TIMOB-24196] Fix 'undefined is not an object' on event listeners

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24196

**Description:**

Fix error seen while debugging Ti apps through Safari debugger with event listeners. addEventListener would show an error moarker stating 'undefined is not an object'. We were always attempting to convert an argument to an object, even if it wasn't supplied (and was therefore undefined). We need to not cast it to a TiObjectRef immediately but verify the value is an object before attempting to convert. If it is not, then we generate an empty object to use.

### DIFF
--- a/iphone/Classes/KrollObject.m
+++ b/iphone/Classes/KrollObject.m
@@ -1336,12 +1336,19 @@ TI_INLINE TiStringRef TiStringCreateWithPointerValue(int value)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


hansemannn:
CR + FT **passed**, PR **approved**! 🚀

PR #8660, opened 2016-12-06
[TIMOB-16369] fix disableContextMenu set to false error

frankieandone:
**CHANGED**
+ show menu when disableContextMenu = false
+ accessors (get/set)

**TEST CASE**
Long click on text to select. Wait to see if floating context menu appears.
```
        url: "http://www.lipsum.com",
        disableContextMenu: true
    });
window.add(webView);
window.open();
```
**EXPECTED**
You should not see floating context menu when disableContextMenu set to false

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16369


garymathews:
CR: **PASS**
FT: **PASS**

Can enable and disable action bar when selecting text. 👍

PR #8661, opened 2016-12-06
[TIMOB-18053] Android: fixed default text color value changing

frankieandone:
**CHANGED**
+ default color value kept getting changed so let default color be set only once

**TEST CASE**
Comment and uncomment colors in each picker row.
```js
    exitOnClose: true,
    layout: 'vertical',
    backgroundColor: 'black'
});

    top: 50
});

var data = [];
data[0] = Ti.UI.createPickerRow({
    title: 'Bananas',
    color: 'yellow'
});
data[1] = Ti.UI.createPickerRow({
    title: 'Strawberries',
    // color: 'red'
});
data[2] = Ti.UI.createPickerRow({
    title: 'Mangos',
    color: 'orange'
});
data[3] = Ti.UI.createPickerRow({
    title: 'Grapes',
    // color: 'purple'
});

picker.add(data);
picker.selectionIndicator = true;

win.add(picker);
win.open();
```
**EXPECTED**
A picker row's text color should only be set if specified otherwise should be default color.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18053

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
@@ -41,7 +41,8 @@ public class TiUINativePicker extends TiUIPicker
 {
 	private static final String TAG = "TiUINativePicker";
 	private boolean firstSelectedFired = false;
-	private static int defaultTextColor = 0;
+	private static int defaultTextColor;
+	private static boolean setDefaultTextColor = false;
 	
 	public static class TiSpinnerAdapter<T> extends ArrayAdapter<T>
 	{
@@ -80,11 +81,12 @@ public class TiUINativePicker extends TiUIPicker
 				fontProperties[TiUIHelper.FONT_SIZE_POSITION], fontProperties[TiUIHelper.FONT_WEIGHT_POSITION],
 				fontProperties[TiUIHelper.FONT_STYLE_POSITION]);
 			}
+			if (!setDefaultTextColor) {
+				defaultTextColor = tv.getCurrentTextColor();
+				setDefaultTextColor = true;
+			}
 			if (rowProxy.hasProperty(TiC.PROPERTY_COLOR)) {
-				final int color = TiConvert.toColor(rowProxy.getProperties(), TiC.PROPERTY_COLOR);
-				if (defaultTextColor != color) {
-					defaultTextColor = tv.getCurrentTextColor();
-				}
+				final int color = TiConvert.toColor((String) rowProxy.getProperty(TiC.PROPERTY_COLOR));
 				tv.setTextColor(color);
 			} else {
 				tv.setTextColor(defaultTextColor);


android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
```diff
@@ -41,7 +41,8 @@
 {
 	private static final String TAG = "TiUINativePicker";
 	private boolean firstSelectedFired = false;
-	private static int defaultTextColor = 0;
+	private static int defaultTextColor = -1;
+	private static boolean setDefaultColor = false;
```
garymathews: the boolean isn't needed

fmerzadyan: That would throw an error: you can't use boolean operators on integer types in Java.

garymathews: You could do this instead

```Java
if (defaultTextColor == 0) {
```

fmerzadyan: If you mean set it to `defaultTextColor` to 0 and using your snippet then it behaves weird initially as text is not visible. Also, this if block would get called x2 for every item with text color set to 0. Seems inefficient?

garymathews: Ah, I see a problem. We should be setting the `defaultColor` regardless if `PROPERTY_COLOR` is defined. Otherwise, the initial text wont be set to the default text color unless an item with `PROPERTY_COLOR` is set first. And yes, good catch keep the Boolean 👍 

See if this works?

```Java
private static int defaultTextColor;
private static boolean setDefaultTextColor = false;
...
private void styleTextView(int position, TextView tv) {
...
if (!setDefaultTextColor) {
    defaultTextColor = tv.getCurrentTextColor();
    setDefaultTextColor = true;
}
if (rowProxy.hasProperty(TiC.PROPERTY_COLOR)) {
    final int color = TiConvert.toColor(rowProxy.getProperties(), TiC.PROPERTY_COLOR);
    tv.setTextColor(color);
} else {
    tv.setTextColor(defaultTextColor);
}
...
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/picker/TiUINativePicker.java
```diff
@@ -41,7 +41,8 @@
 {
 	private static final String TAG = "TiUINativePicker";
 	private boolean firstSelectedFired = false;
-	private static int defaultTextColor = 0;
+	private static int defaultTextColor;
+	private static boolean setDefaultTextColor;
```
garymathews: Looks good! But lets initialise this `setDefaultTextColor = false`

garymathews:
CR: **PASS**
FT: **PASS**

PR #8662, opened 2016-12-06
[TIMOB-23848] iOS: Ti.Media add focus methods and properties

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23848

### DIFF
--- a/apidoc/Titanium/Media/Media.yml
+++ b/apidoc/Titanium/Media/Media.yml
@@ -904,6 +904,24 @@ properties:
     summary: Constant for media device busy error.
     type: Number
     permission: read-only
+  - name: FOCUS_MODE_AUTO
+    summary: Focus mode constant for auto focus mode of camera 
+    type: Number
+    permission: read-only
+    since: "6.1.0"
+    platforms: [android, iphone, ipad]
+  - name: FOCUS_MODE_CONTINUOUS_AUTO
+    summary: Focus mode constant for continuous auto focus mode of camera 
+    type: Number
+    permission: read-only
+    since: "6.1.0"
+    platforms: [iphone, ipad]
+  - name: FOCUS_MODE_FIXED
+    summary: Focus mode constant for fixed/locked focus mode of camera
+    type: Number
+    permission: read-only
+    since: "6.1.0"
+    platforms: [android, iphone, ipad]
   - name: MEDIA_TYPE_PHOTO
     summary: Media type constant for photo media.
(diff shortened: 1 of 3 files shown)


iphone/Classes/MediaModule.m
```diff
@@ -1613,6 +1617,26 @@ -(void)showPicker:(NSDictionary*)args isCamera:(BOOL)isCamera
...
+                    {
+                        if ([device lockForConfiguration:nil])
+                        {
+                            device.focusMode = focus;
+                        }
+                        [device unlockForConfiguration];
+                    }
+                }
+            }
+        }
```
hansemannn: Some issues:
* Put the square brackets in the same line as the if-statement (`if (exp) {`), same in `for (expr) {`
* Put the configuration in the setter of the property, e.g. `- (void)setFlashMode:(id)value` and handle it there
* Provide a getter, e.g. `- (id)flashMode` as well
* I'm curious if setting the `AVCaptureDevice` will work here, because we don't use the `AVFoundation` camera API's but the `UIImagePickerController` API's

apidoc/Titanium/Media/Media.yml
```diff
@@ -1749,6 +1767,13 @@ properties:
     default: true
     since: "3.4.0"
     platforms: [ipad]
+  - name: focusMode
+    summary: Constant to indicate focus mode of camera.
+    constants: Titanium.Media.FOCUS_MODE_*
+    type: Number
+    default: Depends on camera
+    platforms: [android, iphone, ipad]
```
hansemannn: You don't need to specify the platforms if all three are supported. It will use all by default. Same for the other docs as well.

apidoc/Titanium/Media/Media.yml
```diff
@@ -1749,6 +1767,13 @@ properties:
     default: true
     since: "3.4.0"
     platforms: [ipad]
+  - name: focusMode
+    summary: Constant to indicate focus mode of camera.
+    constants: Titanium.Media.FOCUS_MODE_*
+    type: Number
+    default: Depends on camera
```
hansemannn: Can we write something else here? It doesn't really help the developer. So either more infos on how it depends on the camera or leave it out and add a link to the Apple docs in the `description` value.

hansemannn (changes requested):

hansemannn:
Added some comments, but it looks pretty good already!

PR #8671, opened 2016-12-09
[TIMOB-24203] iOS: Fix backgroundColor on rounded Ti.UI.TextField

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24203


vijaysingh-axway (approved):
FT and CR passed.

PR #8676, opened 2016-12-10
[TIMOB-24206] iOS: Fix image-picker memory-leak, fix 0.5 delay when closing the dialog

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24206


vijaysingh-axway (approved):
FT and CR passed.

PR #8680, opened 2016-12-13
[TIMOB-24069] (6_0_X) Fixed pairing of iOS- and watchOS sims

cb1kenobi:
https://jira.appcelerator.org/browse/TIMOB-24069

[TIMOB-24069] Fixed pairing of ios and watch sims that has been around since Xcode 8 where you can't unpair certain ios and watch sims.

https://jira.appcelerator.org/browse/TIMOB-24136

[TIMOB-24136] Removed dependency on 'simctl getenv' which is broken in Xcode 8.2 beta and prevented knowning when an Xcode 8.2 sim has launched.

Updated to ioslib 1.1.0.

Major surgury to the iOS build to move the tiapp.xml validation before the --device-id validation. Also cleaned up min ios version and device family handling.


hansemannn:
Approved!

PR #8683, opened 2016-12-13
[TIMOB-24069] Fixed device family bug.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24069


eric34:
Merged to unblock testing

CR looks good

PR #8691, opened 2016-12-15
[TIMOB-23666]: iOS Button setting remote backgroundImage doesn't app…

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23666

### DIFF
--- a/iphone/Classes/TiUIView.m
+++ b/iphone/Classes/TiUIView.m
@@ -330,7 +330,7 @@ DEFINE_EXCEPTIONS
 		NSLog(@"[WARN] could not find image: %@",image);
 		return nil;
 	}
-	return [[ImageLoader sharedLoader] loadImmediateStretchableImage:url withLeftCap:leftCap topCap:topCap];
+    return [TiUtils loadCappedBackgroundImage:image forProxy:[self proxy] withLeftCap:leftCap withTopCap:topCap];
 }
 
 -(id)transformMatrix
--- a/iphone/Classes/TiUIView.m
+++ b/iphone/Classes/TiUIView.m
@@ -330,7 +330,7 @@ DEFINE_EXCEPTIONS
 		NSLog(@"[WARN] could not find image: %@",image);
 		return nil;
 	}
-    return [TiUtils loadCappedBackgroundImage:image forProxy:[self proxy] withLeftCap:leftCap withTopCap:topCap];
+    return [TiUtils loadCappedBackgroundImage:image forProxy:[self proxy] withLeftCap:leftCap topCap:topCap];
 }
 
 -(id)transformMatrix
--- a/iphone/Classes/TiUIView.m
+++ b/iphone/Classes/TiUIView.m
@@ -330,7 +330,7 @@ DEFINE_EXCEPTIONS
 		NSLog(@"[WARN] could not find image: %@",image);
 		return nil;
 	}
-    return [TiUtils loadCappedBackgroundImage:image forProxy:[self proxy] withLeftCap:leftCap topCap:topCap];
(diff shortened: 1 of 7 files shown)


iphone/Classes/TiUIView.m
```diff
@@ -330,7 +330,7 @@ -(UIImage*)loadImage:(id)image
 		NSLog(@"[WARN] could not find image: %@",image);
 		return nil;
 	}
-	return [[ImageLoader sharedLoader] loadImmediateStretchableImage:url withLeftCap:leftCap topCap:topCap];
+	return [TiUtils loadCappedBackgroundImage:image forProxy:[self proxy] withLeftCap:leftCap withTopCap:topCap];
```
hansemannn: `withTopCap` should be `topCap`, since the `with` prefix is only used for the 1st argument, based on Obj-C coding guidelines.

iphone/Classes/TiUtils.m
```diff
@@ -1982,6 +1982,38 @@ +(NSString*)getResponseHeader:(NSString *)header fromHeaders:(NSDictionary *)res
...
+        if((resultImage != nil) && ([resultImage imageOrientation] != UIImageOrientationUp)) {
+            resultImage = [UIImageResize resizedImage:[resultImage size]
+                                 interpolationQuality:kCGInterpolationNone
+                                                image:resultImage
+                                                hires:NO];
+        }
+    } else if ([image isKindOfClass:[TiBlob class]]) {
+        resultImage = [UIImageResize resizedImageWithLeftCap:leftCap topCap:topCap image:[(TiBlob*)image image]];
+    }
+    return resultImage;
```
hansemannn: Where does this method logic come from? I feel like I saw this before, and if so, we should use this logic for all usages to avoid duplicate code.

iphone/iphone/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -37,6 +47,16 @@
...
+      "idiom" : "ipad",
+      "size" : "20x20",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "ipad",
+      "size" : "20x20",
+      "scale" : "2x"
+    },
+    {
```
hansemannn: How is this change related to the PR? Please restore the initial Contents.json or give me a quick explanation on this. The icons are generated automatically for packaged SDK's anyway. :-)

hansemannn (changes requested):

iphone/Classes/TiUtils.m
vijaysingh-axway: There is a method in TIUtils.m named - +(UIImage*)loadBackgroundImage:(id)image forProxy:(TiProxy*)proxy
. But this method do not use capping . So a new method needed which uses capping.

hansemannn (approved):

htbryant94:
FR Approved! Tested with demo code on iPhone 7 (10.2) Sims & Device and was able to set the BackgroundImage of button elements both with remote and local image paths. Additionally, ran the test case for the following elements:

Ti.UI.iOS.Stepper 
Ti.UI.iOS.Toolbar
Ti.UI.SearchBar
Ti.UI.TabGroup (setActiveTabBackgroundImage)
Ti.UI.TextField

htbryant94 (approved):

hansemannn:
@htbryant94 I did not finish the CR, yet and would tackle that early this week, so please don't merge / resolve, yet, thx!

hansemannn:
CR + FT **passed**, PR **approved**. Great work @vijaysingh-axway !

PR #8692, opened 2016-12-15
[TIMOB-24227] Updated ioslib to v1.1.1 which fixes ios sim launching bug

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24227


hansemannn:
CR + FT passed, PR **approved**! 🚀

PR #8696, opened 2016-12-16
[TIMOB-24069] Updated to ioslib@1.1.2 which fixed bugs with verifying…

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24069

Updated to ioslib@1.1.2 which fixed bugs with verifying sim unpair and sim is running check.


hansemannn:
LGTM, **approved**!

PR #8707, opened 2016-12-21
TIMOB-24249 - Ensuring Ti.UI.iOS.LivePhotoView updates it's size when setting the livePhoto property

rlustemberg:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24249

**Optional Description:**
If a Ti.UI.iOS.LivePhotoView is instantiated without the livePhoto property set, and later it is set (ie, after an http request) , the view size remains at 0x0 pixels instead of updating.


hansemannn:
LGTM, very nice PR. Thank you @rlustemberg, PR **approved**! 🚀

PR #8709, opened 2016-12-21
[TIMOB-23802] Android: expose sustainedPerformancemode for future development

frankieandone:
Added property and updated docs for future reference. Feature is currently only available on Android 7 Nexus 6P + API docs/dev is seems premature, experimental and subject to changes. See [Android docs for info](https://developer.android.com/about/versions/nougat/android-7.0.html#sustained_performance_api)

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23802


apidoc/Titanium/UI/Window.yml
```diff
@@ -6,6 +6,21 @@ since: "0.9"
...
+    platforms: [android]
+
+  - name: setSustainedPerformanceMode
+    summary: sets sustainedPerformanceMode property
+    parameters:
+      - name: mode
+        summary: true to enable sustained performance mode
+        type: Boolean
+    since: "6.1.0"
+    platforms: [android]
```
hansemannn: You don't need the custom getter and setter, they are auto-generated in the docs

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -164,6 +166,10 @@ protected void handleOpen(KrollDict options)
 	            topActivity.startActivity(intent);
 	        }
 	    }
+	    
+	    if (options.containsKey(TiC.PROPERTY_SUSTAINED_PERFORMANCE_MODE)) {
+	        setPropertyAndFire(TiC.PROPERTY_SUSTAINED_PERFORMANCE_MODE, options.getBoolean(TiC.PROPERTY_SUSTAINED_PERFORMANCE_MODE));
```
hansemannn: So where is the property used on the native side? I only see the proxy-methods so far. WiP?

apidoc/Titanium/UI/Window.yml
```diff
@@ -800,6 +815,14 @@ properties:
     since: "3.1.3"
     platforms: [iphone,ipad]
 
+  - name: sustainedPerformanceMode
+    summary: Maintain a sustainable level of performance
```
hansemannn: Missing "." at the end.

apidoc/Titanium/UI/Window.yml
```diff
@@ -800,6 +815,14 @@ properties:
...
+    description: |
+        Performance can fluctuate dramatically for long-running apps, because the system throttles system-on-chip engines as device components reach their temperature limits. This fluctuation presents a moving target for app developers creating high-performance, long-running apps. Setting this feature to true will set sustained performance mode for the corresponding window. If property is undefined then it defaults to false. IMPORTANT NOTE: feature is only available on supported devices and awaiting for further API development by Google. See [Android docs](https://developer.android.com/about/versions/nougat/android-7.0.html#sustained_performance_api) for further info.
```
hansemannn: Break into multiple lines

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
fmerzadyan: I was afraid of adding it currently because that feature is only available in Android API 24 and not 23. We only currently support 23. But if I were to add it, it would be in the thread state methods in `TiBaseActivity`

fmerzadyan:
Also, I suggest adding `isSustainedPerformanceModeSupported()` to check if device supports feature. See: [Android docs: PowerManager](https://developer.android.com/reference/android/os/PowerManager.html#isSustainedPerformanceModeSupported%28%29). 

Footnote: adding [Android docs: Consistent Performance](https://source.android.com/compatibility/7.0/android-7.0-cdd.html#8_5_consistent_performance) for future reference.

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
garymathews: Just encapsulate it

```Java
if (Build.VERSION.SDK_INT == 24) {
    ...
}
```

fmerzadyan: Alright I'll see :+1:

apidoc/Titanium/UI/Window.yml
```diff
@@ -800,6 +800,18 @@ properties:
...
 
+  - name: sustainedPerformanceMode
+    summary: Maintain a sustainable level of performance.
+    description: |
+        Performance can fluctuate dramatically for long-running apps, because the system throttles system-on-chip engines as device components reach their temperature limits. This fluctuation presents a moving target for app developers creating high-performance, long-running apps.
+        
+        Setting this feature to true will set sustained performance mode for the corresponding window. If property is undefined then it defaults to false.
+        
+        Note: feature is only available on supported devices and functionality is experimental and subject to change in future releases. See [Android docs](https://developer.android.com/about/versions/nougat/android-7.0.html#sustained_performance_api) and [Google VR docs](https://developers.google.com/vr/android/ndk/jni/reference/com/google/vr/ndk/base/AndroidCompat.html#setSustainedPerformanceMode(Activity, boolean)) for further info.
+    type: Boolean
```
hansemannn: The lines are still too long. Example of proper line breaks:
```yml
Performance can fluctuate dramatically for long-running apps, because the system throttles 
system-on-chip engines as device components reach their temperature limits. This fluctuation 
presents a moving target for app developers creating high-performance, long-running apps.
     
Setting this feature to true will set sustained performance mode for the corresponding window. 
If the property is undefined it defaults to false.
        
Note: feature is only available on supported devices and functionality is experimental and subject 
to change in future releases. See the [Android docs](https://developer.android.com/about/versions/nougat/android-7.0.html#sustained_performance_api) and [Google VR docs](https://developers.google.com/vr/android/ndk/jni/reference/com/google/vr/ndk/base/AndroidCompat.html#setSustainedPerformanceMode(Activity, boolean)) for further info.
```

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
hansemannn: I'd prefer `>= 24` but yeah :-)

garymathews: @hansemannn nice catch! :)

apidoc/Titanium/UI/Window.yml
```diff
@@ -800,6 +800,22 @@ properties:
...
+        presents a moving target for app developers creating high-performance, long-running apps.
+        
+        Setting this feature to true will set sustained performance mode for the corresponding window.
+        If property is undefined then it defaults to false.
+        
+        Note: feature is only available on supported devices and functionality is experimental and
+        subject to change in future releases. See [Android docs](https://developer.android.com/about/versions/nougat/android-7.0.html#sustained_performance_api) and [Google VR docs](https://developers.google.com/vr/android/ndk/jni/reference/com/google/vr/ndk/base/AndroidCompat.html#setSustainedPerformanceMode(Activity, boolean)) for further info.
+    type: Boolean
+    since: "6.1.0"
+    platforms: [android]
```
hansemannn: Can you add `osver: {android: {min: "7.0"}}` ? Rest looks cool to me!

hansemannn (approved):

garymathews:
Continued by https://github.com/appcelerator/titanium_mobile/pull/9582

PR #8712, opened 2016-12-24
[TIMOB-24259] iOS: Multiple Webviews with same local URL does not inject Titanium XHR bridge properly

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24259

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -41,6 +41,7 @@ static NSString * const kContentTextEncoding = @"kContentTextEncoding";
 static NSString * const kContentMimeType = @"kContentMimeType";
 static NSString * const kContentInjection = @"kContentInjection";
 
+static unsigned long localId = 0;
 
 NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 {
@@ -294,6 +295,8 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 	NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:requestURL];
 	[NSURLProtocol setProperty:textEncodingName forKey:kContentTextEncoding inRequest:request];
 	[NSURLProtocol setProperty:mimeType forKey:kContentMimeType inRequest:request];
+    
+    [request setValue:[NSString stringWithFormat:@"%lu", (localId++)] forHTTPHeaderField:@"X-Titanium-Local-Id"];
 	
 	[self loadURLRequest:request];
 	if (scalingOverride==NO) {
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIWebView.m
```diff
@@ -290,7 +290,10 @@ -(void)loadFile:(NSString*)absolutePath
 		textEncodingName:(NSString*)textEncodingName
 	   mimeType:(NSString*)mimeType
 {
-	NSURL *requestURL = [NSURL fileURLWithPath:absolutePath];
+    static unsigned long uniqueIdentifier = 0;
+    uniqueIdentifier++;
```
hansemannn: Couldn't we use the (existing) static `pageToken` property here? 😊

hansemannn (changes requested):

iphone/Classes/TiUIWebView.m
hansemannn: Update: Ok, so the page-token does not work here. I would still rather fix that one instead of manipulating the URL, because it will also manipulate the `url` in the `load` event that is used by many users to check certain files, so this change would break their apps.

kopiro: @hansemannn I tried so many things to avoid this standard NSURLProtocol behaviour, but with no luck.
I've tried to select a "no-cache" policy in the request - no results.
I've tried to override the LocalProtocolHandler methods, but only one instance is created - so you can't do anything here.
I've checked that there aren't multithreading related issues and trying to avoid them using NSLock - no results.

After ~5 hours of debugging, this is the best solution I've found.
Let me know

hansemannn: Adding it would be a possible breaking change for some users, that is my only problem. Maybe we could we generate a UUID-identifier in a mutable data structure (like NSSet) every time a local URL is set?

kopiro: Yes, already tried to add a key to the *NSURLRequest*, but for iOS they are anyway the same requests.
PS: I'm not a native ObjC/Swift developer, I only try to do my best, so I'm really sure that you'll see what I missed.

kopiro:
@hansemannn I added in the headers, this should be better! :)

hansemannn:
@kopiro Yeah this is great! And works nice. Can you move the static variable to the top of the class to match code-style? And rename it like this to be more descriptive:
```objc
    static unsigned long localId = 0;
    [request setValue:[NSString stringWithFormat:@"%lu", (localId++)] forHTTPHeaderField:@"X-Titanium-Local-Id"];
```

kopiro:
Done! :)

hansemannn (approved):

hansemannn:
CR + FT passed, PR **approved**! Good work @kopiro!

PR #8714, opened 2016-12-27
[TIMOB-24261] iOS: Fix Ti.Media precompile macros regarding camera access

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24261


vijaysingh-axway (approved):
FT and CR passed.

PR #8722, opened 2016-12-30
Cleanup tiapp.xml

hansemannn:
This PR cleans up the following components of the tiapp.xml - used for *new* projects:
- Remove the `<mobileweb>` configuration since it's deprecated and should not be used in new projects
- Remove the `UIRequiresPersistentWiFi`, `UIPrerenderedIcon`, `UIStatusBarHidden` and `UIStatusBarStyle` since this are already the default values used by iOS
- To be discussed: I changed the `UISupportedInterfaceOrientations` to incorporate the behavior of new native iOS projects, which also include landscape-support on iOS (which got more relevant the last years because of the bigger device-screens and splitview-layouts). As it's the default behavior, we could also remove it, but to keep it easy for people to adjust orientations, it should be left there. That's different to the other flags that are more specific.

@cb1kenobi is a good discussion partner for this 🙂

### DIFF
--- a/support/all/tiapp.xml
+++ b/support/all/tiapp.xml
@@ -16,9 +16,11 @@
 	<ios>
 		<plist>
 			<dict>
-				<key>UISupportedInterfaceOrientations~iphone</key>
+				<key>UISupportedInterfaceOrientations</key>
 				<array>
 					<string>UIInterfaceOrientationPortrait</string>
+					<string>UIInterfaceOrientationLandscapeLeft</string>
+					<string>UIInterfaceOrientationLandscapeRight</string>
 				</array>
 				<key>UISupportedInterfaceOrientations~ipad</key>
 				<array>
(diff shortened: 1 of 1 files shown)


cb1kenobi:
We can't merge this until master is officially 7.0.0.

cb1kenobi:
Please restore the Mobile Web configuration lines.

hansemannn:
Thought it's being removed in SDK 7.0.0? Or can we merge it in 6.1.0 if I remove the lines? Thx

cb1kenobi:
We will handle Mobile Web when the time comes.

If you only remove the `<iphone>` section, then it's a gray area. It only affects new apps I believe, so it's not really a breaking change.

hansemannn:
How does it affect new apps? If we use the same base-plist as Xcode does, those keys are already there. And they are default values since iOS 7, so when we don't specify them, they will be used anyway.

cb1kenobi:
Ha, I just realized which `tiapp.xml` you updated. `support/all/tiapp.xml` is the old one used by the by the old Python scripts. So really, it shouldn't even exist. I don't care what you change in this file. :)

hansemannn:
I give up. Let's not change anything.

PR #8723, opened 2016-12-30
[TIMOB-24250] Android: TextField AUTOCAPITALIZATION_ALL property doesn't work with googles new keyboard: GBoard

maggieaxway:
Titanium.UI.TEXT_AUTOCAPITALIZATION_ALL is not working. It doesn't work in native Android as well if user choose shift to enter small letter. Listen to the TextChange to fix the bug

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24250

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
@@ -84,6 +84,8 @@ public class TiUIText extends TiUIView
 	private boolean disableChangeEvent = false;
 
 	protected EditText tv;
+	
+	private KrollDict keyboard;
 
 	public TiUIText(final TiViewProxy proxy, boolean field)
 	{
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -84,6 +84,8 @@
 	private boolean disableChangeEvent = false;
 
 	protected EditText tv;
+	
+	private KrollDict keyboard;
```
garymathews:
```Java
private boolean autoCapAll = false;
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -472,6 +493,9 @@ public void handleKeyboard(KrollDict d)
 		int autocorrect = InputType.TYPE_TEXT_FLAG_AUTO_CORRECT;
 		int autoCapValue = 0;
 
+		// Initialize keyboard
+		keyboard = d;
+		
 		if (d.containsKey(TiC.PROPERTY_AUTOCORRECT) && !TiConvert.toBoolean(d, TiC.PROPERTY_AUTOCORRECT, true)) {
 			autocorrect = 0;
 		}
```
garymathews: ##### LINE 548
```Java
...
case TEXT_AUTOCAPITALIZATION_ALL:
    autoCapValue = InputType.TYPE_TEXT_FLAG_CAP_CHARACTERS;
    autoCapAll = true;
    break;
```

set `autoCapAll = false;` for other AUTOCAPITALIZATION conditions

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -8,6 +8,7 @@
 
 import java.util.HashMap;
 
+import android.text.*;
```
garymathews: Revert import changes, we try not to use wildcards

PR #8726, opened 2017-01-03
[TIMOB-24262] Android: Only fire callback events whilst view is alive

garymathews:
- Make sure the view `proxy` is alive before calling `fireEvent`

###### TEST CASE
```Java
var w = Ti.UI.createWindow(),
    v = Ti.UI.createView({
        backgroundColor:'red',
        width: Ti.UI.FILL,
        height: '50%'
    });

v.addEventListener('click', function() {
    w.remove(v);
});

w.add(v);
w.open();
```

[TIMOB-24262](https://jira.appcelerator.org/browse/TIMOB-24262)


maggieaxway:
CR: PASS
FT: PASS

PR #8727, opened 2017-01-03
[TIMOB-24270] iOS: Fix autorelease-crash when using kroll-thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24270

### DIFF
--- a/iphone/Classes/TiUtils.m
+++ b/iphone/Classes/TiUtils.m
@@ -783,10 +783,10 @@ bool Base64AllocAndEncodeData(const void *inInputData, size_t inInputDataSize, c
     CGContextConcatCTM(context, transform);
     
     CGContextDrawImage(UIGraphicsGetCurrentContext(), CGRectMake(0, 0, width, height), imgRef);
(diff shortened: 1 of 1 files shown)


hansemannn:
@vijaysingh-axway I'm not feeling save with the `retain + autorelease` solution so far, although it seem to work fine in all my use-cases. I'd love your opinion on this. The retain-cycle of the `media` property is a bit complicated (through different threads and contexts).

vijaysingh-axway:
@hansemannn I was thinking something of using autorelease pool and retaining the image inside  caller function, but I am not very sure.  Let me think in more detail on this and will let you know.

vijaysingh-axway (commented):
While verifying with-
//#define TI_USE_KROLL_THREAD
#define USE_JSCORE_FRAMEWORK
I am getting memory leak . 

@hansemannn  Are you sure this solution is working fine .

hansemannn:
I begin to think that the image is not really the problem, but the `TiBlob` is. Otherwise the issue would happen on both JSCore and TiCore.

PR #8728, opened 2017-01-05
[TIMOB-23810]: iOS: Delay in firing click event listener on View when using run-on-main-thread

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23810

### DIFF
--- a/iphone/Classes/TiViewProxy.m
+++ b/iphone/Classes/TiViewProxy.m
@@ -230,27 +230,29 @@ static NSArray* touchEventsArray;
 		pthread_rwlock_unlock(&childrenLock);
         
 		[childView setParent:self];
-
-		// Turn on clipping because I have children
-		[[self view] updateClipping];
-
-		[self contentsWillChange];
-		if(parentVisible && !hidden)
-		{
-			[childView parentWillShow];
-		}
+        
+        if (windowOpened)
+        {
+            // Turn on clipping because I have children
+            [[self view] updateClipping];
+            [self contentsWillChange];
+            if(parentVisible && !hidden)
+            {
+                [childView parentWillShow];
+            }
 		
 #ifndef TI_USE_AUTOLAYOUT
-		//If layout is non absolute push this into the layout queue
-		//else just layout the child with current bounds
-		if (!TiLayoutRuleIsAbsolute(layoutProperties.layoutStyle) ) {
-			[self contentsWillChange];
-		}
-		else
+            //If layout is non absolute push this into the layout queue
+            //else just layout the child with current bounds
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiViewProxy.m
```diff
@@ -230,27 +230,29 @@ -(void)add:(id)arg
...
+        
+        if (windowOpened)
+        {
+            // Turn on clipping because I have children
+            [[self view] updateClipping];
+            [self contentsWillChange];
+            if(parentVisible && !hidden)
+            {
+                [childView parentWillShow];
+            }
```
hansemannn: Very interesting fix, will do a detailed review. But please fix the indentation already and try to use `if () {` instead, thx!

hansemannn:
Fixed the indentation and did a FT, looks all good on both kroll- and mein-thread. Good work! 

One think I noticed though was that we might have another issue there regarding auto-layout. Look at this snippet:

```objc
if (windowOpened) {
	// Turn on clipping because I have children
	[[self view] updateClipping];
	[self contentsWillChange];
	
	if(parentVisible && !hidden) {
		[childView parentWillShow];
	}

#ifndef TI_USE_AUTOLAYOUT
	// If layout is non absolute push this into the layout queue
	// else just layout the child with current bounds
	if (!TiLayoutRuleIsAbsolute(layoutProperties.layoutStyle) ) {
		[self contentsWillChange];
	} else
#endif
	{
		[self layoutChild:childView optimize:NO withMeasuredBounds:[[self view] bounds]];
	}
}
```

If `TI_USE_AUTOLAYOUT` is defined, it would generate a proper if-else statement. But if it's not defined, the snippet would like this - having a closure:
```objc
if (windowOpened) {
	// Turn on clipping because I have children
	[[self view] updateClipping];
	[self contentsWillChange];
	
	if (parentVisible && !hidden) {
		[childView parentWillShow];
	}

        // Check this line, this is not good
	{
		[self layoutChild:childView optimize:NO withMeasuredBounds:[[self view] bounds]];
	}
}
```
So the question would be, if having the closure is still ok or if we should migrate the statement so it properly unwraps it. But that's something different.

vijaysingh-axway:
I don't think there is any problem in having closure. if TI_USE_AUTOLAYOUT is defined , it will be executed on basis of  " if (!TiLayoutRuleIsAbsolute(layoutProperties.layoutStyle) )" otherwise it will execute always. I think readability of code is better in this way.

ewieberappc:
FR passed. Click events fired, without delay, on all views for the test cases. Encountered a related, documented issue but its not a blocker for this.

PR #8731, opened 2017-01-05
[TIMOB-24273] iOS: Overwriting existing event listeners only calls the last one (Regression from 6.0.1)

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24273

**Optional Description:**
Dumb typo referred to an always null variable rather than the correct one, resulting in us always creating a new array to hold listeners for events of any given type (resulting in the last listener being it's only member).


hansemannn (approved):
CR + FT passed, PR approved! Thank you Chris 🙂

PR #8732, opened 2017-01-09
[TIMOB-23327]:Webview request send out beforeload event is excuted on Webview

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23327

### DIFF
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -394,7 +394,7 @@ events:
     description: |
         This event is fired when a blacklisted Url is stopped
     platforms: [android, iphone, ipad]
-    since: [android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"]
+    since: {android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"}
     properties:
       - name: url
         summary: URL of the web document that is stopped.
@@ -427,7 +427,7 @@ properties:
         the blacklist. Note, this only applies in the links clicked inside the webview. The first website
         that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
-    since: [android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"]
+    since: {android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"}
     platforms: [android, iphone, ipad]
     availability: creation
 
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -393,7 +393,7 @@ events:
     summary: Fired when a blacklisted Url is stopped
     description: |
         This event is fired when a blacklisted Url is stopped
-    platforms: [android]
+    platforms: [android, iphone, ipad]
     since: "5.4.0"
     properties:
       - name: url
@@ -428,7 +428,7 @@ properties:
         that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
     since: "5.4.0"
-    platforms: [android]
+    platforms: [android, iphone, ipad]
     availability: creation
 
   - name: data
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -394,7 +394,7 @@ events:
     description: |
         This event is fired when a blacklisted Url is stopped
     platforms: [android, iphone, ipad]
-    since: "5.4.0"
+    since: [android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"]
     properties:
       - name: url
         summary: URL of the web document that is stopped.
@@ -427,7 +427,7 @@ properties:
         the blacklist. Note, this only applies in the links clicked inside the webview. The first website
         that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
-    since: "5.4.0"
+    since: [android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"]
     platforms: [android, iphone, ipad]
     availability: creation
 
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -100,6 +100,7 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 	RELEASE_TO_NIL(reloadData);
 	RELEASE_TO_NIL(reloadDataProperties);
 	RELEASE_TO_NIL(lastValidLoad);
+    RELEASE_TO_NIL(blacklistedURLs);
 	[super dealloc];
 }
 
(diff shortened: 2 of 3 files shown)


apidoc/Titanium/UI/WebView.yml
```diff
@@ -393,7 +393,7 @@ events:
     summary: Fired when a blacklisted Url is stopped
     description: |
         This event is fired when a blacklisted Url is stopped
-    platforms: [android]
+    platforms: [android, iphone, ipad]
     since: "5.4.0"
```
hansemannn: Change to `since: { android: "5.4.0", iphone: "6.10", ipad: "6.1.0"}`

apidoc/Titanium/UI/WebView.yml
```diff
@@ -428,7 +428,7 @@ properties:
         that is loaded will not be stopped even if it matches the blacklist.
     type: Array<String>
     since: "5.4.0"
```
hansemannn: Same as above, include the `since` for iOS

iphone/Classes/TiUIWebView.m
```diff
@@ -100,6 +100,7 @@ -(void)dealloc
 	RELEASE_TO_NIL(reloadData);
 	RELEASE_TO_NIL(reloadDataProperties);
 	RELEASE_TO_NIL(lastValidLoad);
+    RELEASE_TO_NIL(blacklistedURLs);
```
hansemannn: Indentation

iphone/Classes/TiUIWebView.m
```diff
@@ -518,6 +519,15 @@ - (void)setUrl_:(id)args
 	}
 }
 
+-(void)setBlacklistedURLs_:(id)args
+{
+    ENSURE_TYPE(args, NSArray);
+    if (blacklistedURLs) {
+        RELEASE_TO_NIL(blacklistedURLs);
+    }
+    blacklistedURLs = [[NSArray alloc] initWithArray:args];
```
hansemannn: Should we retain this?

iphone/Classes/TiUIWebView.m
```diff
@@ -712,6 +722,20 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
 	}
 
 	NSURL * newUrl = [request URL];
+    
+    if (blacklistedURLs && blacklistedURLs.count > 0) {
+        NSString *urlAbsStr = [newUrl absoluteString];
+        
+        for (NSString *blackListedUrl in blacklistedURLs) {
+            if([urlAbsStr rangeOfString:blackListedUrl options:NSCaseInsensitiveSearch].location !=NSNotFound) {
+                if([self.proxy _hasListeners:@"onStopBlacklistedUrl"]) {
```
hansemannn: Replace `if(` with `if (` in 730 and 731

iphone/Classes/TiUIWebView.m
```diff
@@ -712,6 +722,20 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
 	}
 
 	NSURL * newUrl = [request URL];
+    
+    if (blacklistedURLs && blacklistedURLs.count > 0) {
+        NSString *urlAbsStr = [newUrl absoluteString];
+        
+        for (NSString *blackListedUrl in blacklistedURLs) {
+            if([urlAbsStr rangeOfString:blackListedUrl options:NSCaseInsensitiveSearch].location !=NSNotFound) {
```
hansemannn: Replace `!=NSNotFound` with `!= NSNotFound`

iphone/Classes/TiUIWebView.m
```diff
@@ -712,6 +722,20 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
...
+                if([self.proxy _hasListeners:@"onStopBlacklistedUrl"]) {
+                    NSDictionary *eventDict = [NSDictionary dictionaryWithObjectsAndKeys:urlAbsStr,@"url",@"Webview did not load blacklisted url.", @"messsage", nil];
```
hansemannn: Try to use the shorthand dictionary next time. It looks more solid to read (in my personal opinion). Like this:
```objc
[[self proxy] fireEvent:@"onStopBlacklistedUrl" withObject:@{
    @"url": urlAbsStr,
    @"message": @"Webview did not load blacklisted url."
}];

And is the message the same as used on Android? Maybe we can unify that as well.

File: iphone/Classes/TiUIWebView.m [hunk 8]
```diff
@@ -712,6 +722,20 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
 	}
 
 	NSURL * newUrl = [request URL];
+    
+    if (blacklistedURLs && blacklistedURLs.count > 0) {
+        NSString *urlAbsStr = [newUrl absoluteString];
```
hansemannn: Always use readable variables in Obj-C, e.g. `urlString`

iphone/Classes/TiUIWebView.m
```diff
@@ -712,6 +722,20 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
...
+        
+        for (NSString *blackListedUrl in blacklistedURLs) {
+            if([urlAbsStr rangeOfString:blackListedUrl options:NSCaseInsensitiveSearch].location !=NSNotFound) {
+                if([self.proxy _hasListeners:@"onStopBlacklistedUrl"]) {
+                    NSDictionary *eventDict = [NSDictionary dictionaryWithObjectsAndKeys:urlAbsStr,@"url",@"Webview did not load blacklisted url.", @"messsage", nil];
+                    [self.proxy fireEvent:@"onStopBlacklistedUrl" withObject:eventDict];
```
hansemannn: Try to use `[self proxy]` instead of `self.proxy`, but just out of personal preference :-)

hansemannn (changes requested):

iphone/Classes/TiUIWebView.m
vijaysingh-axway: As this is allocated here and is getting release when dealloc called or new blacklisted urls are getting set. So I think it is right.

hansemannn: The release-cycle is clear, but think we don't work with ARC here, I see cases where the array could be GC'd while still being used. Especially when we have more requests over the time and the variable remains in idle.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -393,8 +393,8 @@ events:
     summary: Fired when a blacklisted Url is stopped
     description: |
         This event is fired when a blacklisted Url is stopped
-    platforms: [android]
-    since: "5.4.0"
+    platforms: [android, iphone, ipad]
+    since: [android: "5.4.0", iphone: "6.1.0", ipad: "6.1.0"]
```
hansemannn: Use curly brackets (like an object) instead of square brackets. Same for the below as well

iphone/Classes/TiUIWebView.m
hansemannn: This is what I would propose:

```objc
-(void)setBlacklistedURLs_:(id)args
{
    ENSURE_TYPE(args, NSArray);
    
    if (blacklistedURLs) {
        RELEASE_TO_NIL(blacklistedURLs);
    }
    
    for (id blacklistedURL in args) {
        ENSURE_TYPE(blacklistedURL, NSString);
    }
    
    blacklistedURLs = [args retain];
}
```

hansemannn:
Another issue: The loading spinner doesn't stop when the URL is blocked. I will add that change, fix the docs and merge then.

hansemannn:
CR + FT passed, PR approved!

hansemannn (approved):

PR #8744, opened 2017-01-11
[TIMOB-19241]: iOS: Keyboard loses focus after 1 character when TextField is in a TableView TableViewRow

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-19241


hansemannn (approved):
Did a minor adjustment in getting the UITableView instance and formatted the comment properly, but it looks great. Good work man!

ewieberappc (approved):
FR Passed. Tested using an app with textAreas in TableViewRows. The keyboard was not dismissed after multiple characters were typed into the TextAreas. Auto-Layout was enabled, disabled, and not present in the tiapp.

PR #8746, opened 2017-01-12
[TIMOB-7406]: iOS: Set a user-agent for a webview

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-7406]

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -43,8 +43,6 @@ static NSString * const kContentInjection = @"kContentInjection";
 
 static unsigned long localId = 0;
 
-static NSString *defaultUserAgent;
-
 NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 {
 	if (encoding == NSUTF8StringEncoding) {
@@ -179,8 +177,10 @@ NSString *HTMLTextEncodingNameForStringEncoding(NSStringEncoding encoding)
 			[spinner sizeToFit];
 			[spinner startAnimating];
 		}
-        if (defaultUserAgent == nil) {
-            defaultUserAgent = [webview stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
+
+        if ([[NSUserDefaults standardUserDefaults] objectForKey:@"DefaultUserAgent"] == nil) {
+            NSString *defaultUserAgent = [webview stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
+            [[NSUserDefaults standardUserDefaults] setObject:defaultUserAgent forKey:@"DefaultUserAgent"];
         }
 	}
 	return webview;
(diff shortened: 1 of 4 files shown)


iphone/Classes/TiUIWebView.m
```diff
@@ -540,6 +540,13 @@ -(void)setHandlePlatformUrl_:(id)arg
     willHandleUrl = [TiUtils boolValue:arg];
 }
 
+-(void)setUserAgent_:(id)args
+{
+    ENSURE_TYPE(args, NSString);
+    NSDictionary *dict = @{@"UserAgent":args};
```
hansemannn: Please allow the ability to pass `null` so the user-agent will be resetted to it's default,

hansemannn: Pass the dictionary directly

iphone/Classes/TiUIWebView.m
```diff
@@ -540,6 +540,13 @@ -(void)setHandlePlatformUrl_:(id)arg
     willHandleUrl = [TiUtils boolValue:arg];
 }
 
+-(void)setUserAgent_:(id)args
+{
+    ENSURE_TYPE(args, NSString);
```
hansemannn: Use `[self proxy] replaceValue:args forKey:@"userAgent" notification: NO]` for retaining the proxy state.

hansemannn (changes requested):
Left some comments (most of them based on our Flowdock discussion). Also please change the docs to add the `since` version and platform for iOS

iphone/Classes/TiUIWebView.m
vijaysingh-axway: By passing null , it will create exception. And it will not reset to default user agent. Is this really required to give facility to reset default one ?

hansemannn: That's I meant that is should be allowed (e.g. by checking the argument type) :-) Check the Android implementation - if the reset functionality is not exposed there, it's ok for now to skip it.

iphone/Classes/TiUIWebView.m
```diff
@@ -177,6 +179,9 @@ -(UIWebView*)webview
 			[spinner sizeToFit];
 			[spinner startAnimating];
 		}
+        if (defaultUserAgent == nil) {
+            defaultUserAgent = [webview stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
```
hansemannn: What happens on the second call, when there is no default user-agent anymore? Then it could never be resetted again?

iphone/Classes/TiUIWebView.m
```diff
@@ -542,7 +547,10 @@ -(void)setHandlePlatformUrl_:(id)arg
 
 -(void)setUserAgent_:(id)value
 {
-    ENSURE_TYPE(value, NSString);
+    ENSURE_TYPE_OR_NIL(value, NSString);
+    if (value == nil || [value isEqualToString:@""]) {
+        value = defaultUserAgent;
```
hansemannn: Here is the problem I had thinking more about this:

- This whole config will set it globally, not only for this webview
- We would need to store the "old" default user agent somewhere so we can restore it

hansemannn (changes requested):

iphone/Classes/TiUIWebView.m
vijaysingh-axway: Ahh. Missed that. Good catch. Updated. Please check.

hansemannn:
@vijaysingh-axway Good solution! Will do a FT today, then we're ready to merge.

hansemannn (approved):

hansemannn:
CR + FT **passed**, PR **approved**! Handing over to @ewieberappc.

hansemannn:
@htbryant94 Ping!

htbryant94 (approved):
FR Approved! setting the userAgent property works both on createWebView() and setting it after creation with .setUserAgent. Tested on iOS 10.2 / 9.3 Devices and Simulators.

PR #8752, opened 2017-01-14
[TIMOB-24310] iOS: Start TiLogServer before finishing launching

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24310

### DIFF
--- a/iphone/Classes/TiApp.m
+++ b/iphone/Classes/TiApp.m
@@ -310,9 +310,7 @@ TI_INLINE void waitForMemoryPanicCleared();   //WARNING: This must never be run 
 - (void)applicationDidFinishLaunching:(UIApplication *)application 
 {
 	[TiExceptionHandler defaultExceptionHandler];
-#ifndef DISABLE_TI_LOG_SERVER
-	[TiLogServer startServer];
-#endif
+
 	[self initController];
 	[self launchToUrl];
 	[self boot];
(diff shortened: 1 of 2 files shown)


cb1kenobi:
I guess I don't understand. The log server was being initialized at the earliest possible point. How does moving it to TiHost initialize the log server sooner?

hansemannn:
The `[TiHost initialize]` is called earlier, so logs between that method and `application:didFinishLaunching:withOptions` are also captured. See the ticket for more infos

cb1kenobi:
I guess I can't reproduce the issue with the iOS Simulator. I didn't test on device.

When I test this, I get the same results with and without your changes.

```
[DEBUG] Reading stylesheet from: /Users/chris/Library/Developer/CoreSimulator/Devices/7A165B3B-9706-40C0-A4CE-5C13287B1A96/data/Containers/Bundle/Application/FF39747D-61BD-4C1F-B611-8D27BDB5B4F1/testapp.app/stylesheet.plist
[INFO]  testapp/1.0 (6.1.0.5b4b0f5)
[DEBUG] Loading: /Users/chris/Library/Developer/CoreSimulator/Devices/7A165B3B-9706-40C0-A4CE-5C13287B1A96/data/Containers/Bundle/Application/FF39747D-61BD-4C1F-B611-8D27BDB5B4F1/testapp.app/app.js, Resource: app_js
```

I'm unable to see the `[INFO]  Application started` log message before or after your changes.

Given that your change has no apparent change and that things seem to work before your PR, then I don't think we should merge this.

PR #8755, opened 2017-01-16
[TIMOB-24313] iOS/Android: Expose Ti.UI.ImageView.aspectFill

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/AC-4734


kopiro:
I'm working to implement it in Android

hansemannn:
Hey @kopiro, after discussing this with the team, we should rather go with a more flexible solution to allow all other possible content-mode properties, see [here](https://developer.apple.com/documentation/uikit/uiviewcontentmode?language=objc) for details.

sgtcoolguy:
@hansemannn @kopiro Is this PR dead? Is there another PR or effort that supersedes this one (and this can be closed)?

hansemannn:
Closing due to merge conflicts and stale discussion. I still think it'd be a good idea to do the change, but we have to do it more flexible as I see problems regarding some existing use-cases already.

PR #8756, opened 2017-01-16
[TIMOB-24314] iOS: Run HapticEngine API from main-thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24314


vijaysingh-axway (approved):
CR and FT passed.

PR #8757, opened 2017-01-17
[TIMOB-24317] Android: Focus area for view pinch event

HazemKhaled:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24317

**EDIT by Hans**: `no tests`, because only testable by pinching with the actual finger, so a manual test is required.


apidoc/Titanium/UI/View.yml
```diff
@@ -345,6 +345,72 @@ events:
...
+        summary: |
+            The Y coordinate of the current gesture's focal point.
+            Note: This property is only available for Android
+        type: Number
+        
+      - name: isInProgress
+        summary: |
+            true if a scale gesture is in progress.
+            Note: This property is only available for Android
+        type: Number
```
hansemannn: For all docs here, please add:
```yml
since: "6.2.0"
platforms: [android]
```

hansemannn:
iOS Parity available for (some of) the properties?

hansemannn:
Another thing I notices aside from this PR: The docs say that the "pinch" gesture is only available on iOS (since 1.8.0), but obviously it's available on Android as well. Do you know since when? I' git-blame-check it otherwiese

HazemKhaled:
As usual you faster, thanks for updating docs, don't have a plan to implement it with iOS.

hansemannn:
Hey @HazemKhaled, I'd like to integrate this today! Can you please do me a huge favor and sign the CLA again? We moved to a unified Axway CLA server recently that requires to re-sign once. Thank you!

HazemKhaled:
From where i can sign the new one?

hansemannn (approved):
CR looks good, PR **approved**!

PR #8760, opened 2017-01-17
[TIMOB-24322] Android: SearchView “change” event is fired on SearchView first initialization.

grazianogrespan:
**Ticket link** : https://jira.appcelerator.org/browse/TIMOB-24322

Because this issue is very similar to this :  https://jira.appcelerator.org/browse/TIMOB-13721 ,
I have applied the fix in the same way.
For additional details read Jira issue.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
@@ -25,7 +25,7 @@ import android.widget.EditText;
 public class TiUISearchView extends TiUIView implements SearchView.OnQueryTextListener, SearchView.OnCloseListener {
 	private SearchView searchView;
 
-	private boolean changeEventEnabled = false;
+	private boolean changeEventEnabled = true;
 
 	public static final String TAG = "SearchView";
 
@@ -51,11 +51,11 @@ public class TiUISearchView extends TiUIView implements SearchView.OnQueryTextLi
 		if (props.containsKey(TiC.PROPERTY_HINT_TEXT)) {
 			searchView.setQueryHint(props.getString(TiC.PROPERTY_HINT_TEXT));
 		}
-		changeEventEnabled = false;
 		if (props.containsKey(TiC.PROPERTY_VALUE)) {
+			changeEventEnabled = false;
 			searchView.setQuery(props.getString(TiC.PROPERTY_VALUE), false);
+			changeEventEnabled = true;
 		}
-		changeEventEnabled = true;
 		if (props.containsKey(TiC.PROPERTY_ICONIFIED)) {
 			searchView.setIconified(props.getBoolean(TiC.PROPERTY_ICONIFIED));
 		}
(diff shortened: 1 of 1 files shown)


ssjsamir:
CR - Passed

android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
```diff
@@ -114,7 +118,8 @@ public boolean onQueryTextChange(String query) {
 		if (searchChangeListener != null) {
 			searchChangeListener.filterBy(query);
 		}
-		fireEvent(TiC.EVENT_CHANGE, null);
+		if (changeEventEnabled)
```
garymathews: Use braces

```Java
if (changeEventEnabled) {
    fireEvent(TiC.EVENT_CHANGE, null);
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
```diff
@@ -47,19 +49,21 @@ public void processProperties(KrollDict props) {
 
 		// Check if the hint text is specified when the view is created.
 		if (props.containsKey(TiC.PROPERTY_HINT_TEXT)) {
-			searchView.setQueryHint(props.getString(TiC.PROPERTY_HINT_TEXT));			
-		} 
+			searchView.setQueryHint(props.getString(TiC.PROPERTY_HINT_TEXT));
+		}
+		changeEventEnabled = false;
```
garymathews: Could place these inside the `if` statement

grazianogrespan: Hi @garymathews, if I put line 54 ( `changeEventEnabled = false;`) inside the `if` , I would cause this bad behaviour: 
on JS side if you try to run this code and then with softKeyboard manually change the text of the SearchView you will not see the alert (which is not the standard behaviour), this is because in native side the changeEventEnabled var is set to `false` and will never change to `true` until the `value` prop is set on method processProperties. That's why I left line 54 outside `if` statement.
`var win = Ti.UI.createWindow();
 
var searchView = Ti.UI.Android.createSearchView({
hintText:"abcde" 
});
 
searchView.addEventListener("change",function(){
    alert("Fired Change Event!");
});
 
win.add(searchView);
win.open();`

lokeshchdhry:
FR Passed.

The change event is not fired on init if `value` property is set rather is now fired after change after initialization.

Appc Studio : 4.8.1.201612050850
SDK Version : Local 6.1.0
Mac OS Version : 10.12.2
Xcode Version : Xcode 8.2.1 Build version 8C1002
Appc CLI AND Appc NPM : {"NPM":"4.2.9-1","CLI":"6.1.0"}
Ti CLI : 5.0.11
Alloy : 1.9.5
Node : v4.6.0
Device: Nexus 6 6.0.1, Emulator 7.1.1, Emulator 5.1.1, Emulator 4.4.2

android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
garymathews: Try set the default value of `changedEventEnabled = true;` and do this:

```Java
if (props.containsKey(TiC.PROPERTY_VALUE)) {
    changedEventEnabled = false;
    searchView.setQuery(props.getString(TiC.PROPERTY_VALUE), false);
    changedEventEnabled = true;
}
```

grazianogrespan: Ok done! Tested and works. Already committed all the changes. 👍

lokeshchdhry:
Changes look good 👍

PR #8778, opened 2017-01-23
[TIMOB-24337] (6_0_X) iOS: Resolve SearchQuery API dependencies

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24337


vijaysingh-axway (approved):
FT and CR passed.

ewieberappc (approved):
FR passed. Verified that, on device, indexed app content can be searched for and the app is properly switched to when selected. Used the test case in TIMOB-23828 and modified versions of it.

PR #8792, opened 2017-01-26
[TIMOB-18565] Add support for Android Libraries

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18565

**Description:**
This allows Android Libraries in the form of Android Archive (AAR)
files to be dropped in as module dependencies.


AppWerft:
nice. But what is with this problem: the third party aar uses more then one R.class. Is it solved with this PR?

janvennemann:
@AppWerft, do you happen to have an example .aar to test against? Currently only the main package defined in AndroidManifest.xml from inside the .aar will be passed as an extra package to AAPT.

AppWerft:
The aar is not open source. Inside are res for library, for v4 and for v7. Currently we have merged all stuff in one res. But in classes are references to 3 *.*.R. I have xinged you.

android/cli/hooks/aar-transform.js
```diff
@@ -0,0 +1,45 @@
...
+var fs = require('fs');
+var path = require('path');
+
+exports.cliVersion = '>=3.2';
+
+exports.init = function (logger, config, cli) {
+	cli.on('build.module.pre.compile', {
+		post: function (builder, callback) {
+			var foundAndroidArchives = [];
+			var aarOutputPath = path.join(builder.buildDir, 'intermediates/exploded-aar');
```
garymathews:
```Javascript
var aarOutputPath = path.join(builder.buildDir, 'intermediates', 'exploded-aar');
var assetsDestinationPath = path.join(builder.buildDir, 'intermediates', 'assets');
var libraryDestinationPath = path.join(builder.buildDir, 'intermediates', 'lib');
var sharedLibraryDestinationPath = path.join(builder.buildGenJniLocalDir, 'jni');
```

janvennemann:
@garymathews, updated the PR with my latest changes to the module build. The app build requires some more work and will follow tomorrow.

garymathews (approved):
CR: **PASS** after minor change
FT: **PASS**

###### TEST CASE
- Compiled a number of modules with AAR and JAR libs using:
```
 appc run -p android --build-only -l trace
```
and verified their output and usage.

AWESOME!! 🎆

android/cli/hooks/aar-transform.js
```diff
@@ -0,0 +1,446 @@
...
+			 * we can skip the whole transform process on subsequent builds.
+			 *
+			 * @param {Function} done Function to call once the hash has been computed
+			 */
+			function hashFile(done) {
+				var hash = crypto.createHash('sha1');
+				var fileReadStream = fs.createReadStream(aarPathAndFilename);
+				fileReadStream.on('readable', function() {
+					var data = fileReadStream.read();
+					if (data)
```
garymathews: minor, but we should use brackets here

janvennemann:
@garymathews, addressed the wrong `--no-crunch` option and missing brackets.

garymathews (approved):
CR: **PASS**
FT: **PASS**

👍

hansemannn:
@janvennemann Please resolve the merge conflicts

williamrijksen:
@janvennemann Cool stuff!

hansemannn:
@janvennemann Please resolve the merge conflicts (😁). Should be the last one, but we moved the `titanium` library into an own `node-titanium-sdk` one, so this causing issues when requiring. Thanks!

android/cli/commands/_build.js
```diff
@@ -35,8 +35,8 @@ var ADB = require('node-titanium-sdk/lib/adb'),
 	temp = require('temp'),
 	SymbolLoader = require('appc-aar-tools').SymbolLoader,
 	SymbolWriter = require('appc-aar-tools').SymbolWriter,
-	ti = require('titanium-sdk'),
-	tiappxml = require('titanium-sdk/lib/tiappxml'),
+	ti = require('node-titanium-sdk'),
+	tiappxml = require('node-titanium-sdk/lib/tiappxml'),
```
hansemannn: Can we please scan **all** usages for this require-issue? This is the 3rd commit where we need to manually adjust the path after initially changing it. @cb1kenobi ?

cb1kenobi: Yeah, I think someone should just grep for `titanium-sdk` and replace with `node-titanium-sdk`. There shouldn't be 100's of instances.

hansemannn: @janvennemann Can you do that? I'll be out until Wednesday and we should really get that in before we forget it. Maybe in an own PR which we can merge before this one gets queued. Thanks guys!

lokeshchdhry:
FR done for backport PR.
Merging.

PR #8795, opened 2017-01-26
[TIMOB-15275] iOS: Use ARC in new native iOS modules

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-15275

- [x] Enable ARC in the project build-settings
- [x] Remove `dealloc:` call from module due to ARC
- [x] Use iOS 8 as the base iOS SDK for new modules
- [x] Recommend `appc ti build -p ios --build-only` in the `build.py`
- [x] Set Xcode compatibility version to 7.0 (from 3.2, lol)

**Note**: I used the `-Wno-arc-performSelector-leaks` in the SDK project, to supress the selector warnings for ARC-based selectors in UI-Thread macros (`ENSURE_UI_THREAD_1_ARG` and `ENSURE_UI_THREAD_0_ARGS`). We use those macros to ensure that the method is executed on the main-thread and always have left it to the developer to know how to use the macro. Even chcking if the instance response to the selector (`respondsToSelector:`) won't mute this warning, so it's recommend ([here](http://stackoverflow.com/a/7933931/5537752), [here](http://stackoverflow.com/a/10797893/5537752), [here](http://stackoverflow.com/a/11895530/5537752), ...) to mute the warning in this case.

This is **no** breaking change, ARC- and non-ARC modules / SDK can play together without problems.

### DIFF
--- a/iphone/Classes/TiThreading.h
+++ b/iphone/Classes/TiThreading.h
@@ -10,10 +10,12 @@
 #define ENSURE_UI_THREAD_1_ARG(x)	\
 if (![NSThread isMainThread]) { \
 	SEL callback = _cmd;\
-	TiThreadPerformOnMainThread(^{[self performSelector:callback withObject:x];}, NO);\
+	_Pragma("clang diagnostic push") \
+	_Pragma("clang diagnostic ignored \"-Warc-performSelector-leaks\"") \
+	TiThreadPerformOnMainThread(^{[self performSelector:callback withObject:x];}, NO); \
+	_Pragma("clang diagnostic pop") \
 	return; \
 } \
-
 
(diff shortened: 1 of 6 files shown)


eric34:
Perhaps 'appc ti build -p ios --build-only' should be 'appc run -p iOS --build-only'

cb1kenobi:
FYI, on any files that you modify, if there's a copyright notice, then be sure to bump the year. The copyright year(s) is per file. We should not mass update all dates for every file.

hansemannn:
Spoke to @ingo about copyright notices in general recently. The suggestion was to either only keep the first year the copyright was made valid or to say "2009-Present" like Facebook does in their open source apps.

cb1kenobi:
CR'd and FR'd. Looks good. **APPROVED**

PR #8798, opened 2017-01-27
[6_0_X] [TIMOB-24183] Windows: Failed to install WP 8.1 app

infosia:
Cherry-pick https://github.com/appcelerator/windowslib/pull/64 for 6_0_X [TIMOB-24183](https://jira.appcelerator.org/browse/TIMOB-24183) manually because it's critical.

`appc run -p windows --target wp-device` fails to deploy apps to Windows Phone 8.1 device.

It turns out that deployment failed because outputs from Windows Phone deployment tools have been changed at some point of the installation, now it could output nothing but returns error code other than "0". As far as I can see it means deployment went successful and we can ignore the error code when output from tools are empty.

This PR intends to be a "manual fix" since we can not bump up the version simply. I'm pushing https://github.com/appcelerator/windowslib/pull/64 but it includes some fixes that is marked "6.1.0" (in 0.4.21 to 0.4.27), so we want to update certain file that is only related to SDK 6.0.2 without bumping up windowslib version. I did not actually bump up the version just like we did in https://github.com/appcelerator/titanium_mobile/pull/8641 .

### DIFF
--- a/node_modules/windowslib/CHANGELOG.md
+++ b/node_modules/windowslib/CHANGELOG.md
@@ -1,3 +1,7 @@
+0.4.20.2 (01/27/2017) - Only for SDK 6.0.2
+-------------------
+  * [TIMOB-24183] Failed to install WP 8.1 app
+
 0.4.20.1 (11/29/2016) - Only for SDK 6.0.1
 -------------------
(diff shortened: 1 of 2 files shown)


garymathews:
LGTM 👍

PR #8805, opened 2017-01-31
[TIMOB-23391] : iOS View is not getting horizontally wrapped properly

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23391

### DIFF
--- a/iphone/Classes/TiUILabel.m
+++ b/iphone/Classes/TiUILabel.m
@@ -57,7 +57,7 @@
 -(CGSize)sizeForFont:(CGFloat)suggestedWidth
 {
 	NSAttributedString *value = [label attributedText];
-    CGSize maxSize = CGSizeMake(suggestedWidth<=0 ? 480 : suggestedWidth, 1000);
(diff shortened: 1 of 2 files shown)


hansemannn:
The CR & FT works, but we need to discuss a few more items:
- Did you test the changes with other horizontal layouts / sizings already?
- I guess you've chosen this three conditions because of the `offsetH` that effects the left / right properties? If so, please provide additional examples to catch the other two conditions to be triggered, so we can ensure those also still work.

@ewieberappc We need to test this PR **very** detailed to avoid layout regressions. It looks goo to me, but you guys might have more test-cases to check with.

hansemannn (approved):

ewieberappc (approved):
FR Passed. Tested using:
MacOS 10.12 (16A323)
Studio 4.8.1.201701192045
Ti SDK 6.1.0
Appc NPM 4.2.8
Appc CLI 6.1.0
Alloy 1.9.6
Xcode 8.3 (8W120l)

Tested the provided sample code, ran our layout test suites, KitchenSink tests around layout, and any local projects I had that modified layouts in any way. No errors encountered

PR #8807, opened 2017-01-31
[TIMOB-24358] (6_0_X) iOS: Fix Ti.Media audio-players on device

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24358


vijaysingh-axway (approved):
FT and CR passed.

htbryant94 (approved):
FR Approved, Ti.Media.openMusicLibrary() no longer causes an app to fail on device builds. Tested with both iOS9.3 & iOS10.2.

PR #8809, opened 2017-01-31
[TIMOB-24361] iOS/Android: Fix rgb-warning, allow 2-digit float values

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24361


vijaysingh-axway (approved):
iOS FT and CR passed.

vijaysingh-axway:
@garymathews Please review android changes.

garymathews (commented):
@hansemannn The Android changes are good, but the test case you posted on JIRA should not be valid. The float values should be from `0.0 - 1.0`, `255.0` should be invalid (which it is on Android) is this invalid on iOS? http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI

Also, looking at the iOS code, it seems to manually parse the RGB string. Why not use `NSRegularExpression` and the same expressions from Android?

hansemannn:
@garymathews No idea why it was manually parsed, will update that in the future. But iOS definitely uses 255.

ewieberappc (approved):
FR Passed, using:
MacOS 10.12 (16A323)
Studio 4.9.0.201704061825
Ti SDK 6.1.0.v20170503095848
Appc NPM 4.2.9
Appc CLI 6.2.0
Alloy 1.9.11
Xcode 8.3.2 (8E2002)

2-digit float values are accepted on both iOS and Android platforms, for rgb() and rgba().

PR #8814, opened 2017-02-02
[TIMOB-24362] Changed iOS adhoc builds to use `xcodebuild archive`

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24362


hansemannn (approved):
CR + FT passed, PR approved.

PR #8816, opened 2017-02-03
Android: Only use touch-sound when view is supposed to receive touches

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16080

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
@@ -1720,11 +1720,11 @@ public abstract class TiUIView
 			return;
 		}
 
+		boolean enabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_ENABLED), true);
+
 		if (proxy.hasProperty(TiC.PROPERTY_TOUCH_ENABLED)) {
-			boolean enabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_ENABLED), true);
-			if (!enabled) {
-				touchable.setEnabled(false);
-			}
+			touchable.setSoundEffectsEnabled(enabled);
+			touchable.setEnabled(enabled);
 		}
 		registerTouchEvents(touchable);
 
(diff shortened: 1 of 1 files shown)


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1720,11 +1720,11 @@ protected void registerForTouch(final View touchable)
 			return;
 		}
 
+		boolean enabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_ENABLED), true);
```
garymathews: Should keep this inside

```Java
if (proxy.hasProperty(TiC.PROPERTY_TOUCH_ENABLED)) {
```

hansemannn: restored

fmerzadyan:
Sound effects are still present after change. `enabled` is set to false initially but changes to true which is probably why sound is still enabled.

hansemannn:
This solution will retain the default behavior. Do you have a better idea to solve it more dynamically? The PR was taken from [this commit](https://github.com/gstreetmedia/titanium_mobile/commit/86839826cad0c4e4eeb8f491c3922004161b9545), so we might want to restore the `else` case again, which should disable it by default. Thoughts?

hansemannn:
@fmerzadyan Changed it to disable by default. But I can see that this could cause problems for buttons / list items that do not have the "touchEnabled" property defined manually. Or do we have default values set for this?

antw12:
Hi Hans, I had a look at this ticket and found a potential solution... 

`protected void registerForTouch(final View touchable)
	{
		if (touchable == null) {
			return;
		}

		if (proxy.hasProperty(TiC.PROPERTY_TOUCH_ENABLED)) {
			boolean enabled = TiConvert.toBoolean(proxy.getProperty("soundEffectsEnabled"), true);
			touchable.setSoundEffectsEnabled(enabled);
			touchable.setEnabled(enabled);
		}
		
		registerTouchEvents(touchable);`

Seems that because of the boolean conversion was using the constant PROPERTY_TOUCH_ENABLED which is always true, i think is the reason why it always had sounds 
please let me know if you want me to do anything. thanks

hansemannn:
touchEnabled is not always true, so the latest solution would handle both cases. @garymathews thoughts?

garymathews:
@hansemannn @antw12 I think we should implement a separate property for sound (e.g `PROPERTY_TOUCH_SOUND_ENABLED`) that's independent to `PROPERTY_TOUCH_ENABLED`.

PR #8817, opened 2017-02-04
[TIMOB-24372] iOS: Expose all missing media item properties

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24372


hansemannn:
@vijaysingh-axway Please review this next week if possible, we just got another request from the customer, thx!

vijaysingh-axway (commented):
1. MPMediaItemPropertyHasProtectedAsset  is available with iOS 9_2 and above so macro for same should be there.
2. Any particular reason for deleting  MPMediaItemPropertyPodcastTitle, @"podcastTitle",
and 
 MPMediaItemPropertyPlayCount, @"playCount"
from itemProperties method.

3. In
 +  - name: playCount
+    summary: The  to filter on. Value should be a Number.
should be 

 +  - name: playCount
+    summary: The  playCount to filter on. Value should be a Number.

hansemannn:
1. Added the version check. Will return `NO` on unsupported devices now. Using a mutable dict for this would be overhead.
2. Because they are filterable. They can still be fetched the same way, but now people can also use them to filter queries (e.g. in `Ti.Media.queryMusicLibrary`).
3. Fixed

vijaysingh-axway (commented):
Implementation of 
++(BOOL)isIOSVersionOrGreater:(NSString *)version;
missing in TIUtils.m :)

vijaysingh-axway (commented):
FT and CR passed.

htbryant94 (approved):
PR approved! tested with demo code provided in JIRA ticket, including additional exposed media item properties. iOS 10.2 & 9.3.

PR #8821, opened 2017-02-09
[TIMOB-23799] : iOS Support Named Pasteboards

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23799

### DIFF
--- a/apidoc/Titanium/UI/Clipboard/Clipboard.yml
+++ b/apidoc/Titanium/UI/Clipboard/Clipboard.yml
@@ -176,7 +176,7 @@ properties:
     availability: creation
 
   - name: allowCreation
-    summary: Create a clipboard identified by name if it doesn,t exist.
+    summary: Create a clipboard identified by name if it doesn't exist.
     description: |
         If "create" is false, it will return the existing named clipboard but will 
         not create new one.
--- a/apidoc/Titanium/UI/Clipboard/Clipboard.yml
+++ b/apidoc/Titanium/UI/Clipboard/Clipboard.yml
@@ -151,27 +151,13 @@ methods:
         - type: Array<Dictionary>
     since: "5.5.0"
     osver: {ios: {min: "10.0"}}
-    platforms: [iphone, ipad]  
+    platforms: [iphone, ipad]
 
   - name: remove
     summary: Removes the clipboard.
     since: "6.1.0"
     platforms: [iphone, ipad]
 
-  - name: persistent
-    summary: A Boolean value that indicates whether the clipboard is persistent.
-    description: |
-        When a clipboard is persistent, it continues to exist past app terminations and across system 
-        reboots. Clipboard that are not persistent only last until the owning (creating) app quits.
-    parameters:
-      - name: persistent
-        summary: value of persistent.
-        type: Boolean
-        default: false in case of named clipboard otherwise true.
-    since: "6.1.0"
-    osver: {ios: {max: "9.x.x"}}
-    platforms: [iphone, ipad]
-    
 properties:
   - name: name
     summary: Create a new named clipboard.
@@ -209,6 +195,20 @@ properties:
     since: "6.1.0"
     platforms: [iphone, ipad]
     availability: creation
+
+  - name: persistent
+    summary: A Boolean value that indicates whether the clipboard is persistent.
+    description: |
+        When a clipboard is persistent, it continues to exist past app terminations and across system 
+        reboots. Clipboard that are not persistent only last until the owning (creating) app quits.
+    parameters:
+      - name: persistent
+        summary: value of persistent.
+        type: Boolean
+        default: false in case of named clipboard otherwise true.
+    since: "6.1.0"
+    osver: {ios: {max: "9.x.x"}}
+    platforms: [iphone, ipad]
 
 examples:
   - title: Copy Text to the Clipboard
--- a/apidoc/Titanium/UI/Clipboard/Clipboard.yml
+++ b/apidoc/Titanium/UI/Clipboard/Clipboard.yml
@@ -162,22 +162,15 @@ properties:
   - name: name
     summary: Create a new named clipboard.
     description: Create a new clipboard with a given name. See the example on how to use this.
-    parameters:
-      - name: name
-        summary: name of clipboard.
-        type: String
     since: "6.1.0"
     platforms: [iphone, ipad]
     availability: creation
 
   - name: unique
     summary: Create a new clipboard identified by a unique system-generated name.
-    description: Create a new clipboard identified by a unique system-generated name. See the example 
-    on how to use this.
-    parameters:
-      - name: unique
-        summary: value of unique.
-        type: Boolean
+    description: |
+        Create a new clipboard identified by a unique system-generated name. 
+        See the example on how to use this.
     since: "6.1.0"
     platforms: [iphone, ipad]
     availability: creation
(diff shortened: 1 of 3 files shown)


hansemannn (changes requested):

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -151,7 +151,57 @@ methods:
...
     since: "5.5.0"
     osver: {ios: {min: "10.0"}}
+    platforms: [iphone, ipad]  
+
+  - name: remove
+    summary: Removes the clipboard.
+    since: "6.1.0"
+    platforms: [iphone, ipad]
+
+  - name: name
```
hansemannn: This is no method

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -151,7 +151,57 @@ methods:
...
+  - name: create
+    summary: Returns a clipboard identified by name, optionally creating it if it doesn,t exist.
+    description: If create is false, it will return the existing named clipboard but will not create new one. If create is true, it will return the existing clipboard (if exists) otherwise will create new clipboard with given name.
+    parameters:
+      - name: create
+        summary: value of create.
+        type: Boolean
+    since: "6.1.0"
+    platforms: [iphone, ipad]
+    availability: creation
```
hansemannn: What is this supposed to be? :-) Maybe we should rather call it different? Like `override` or `allowOverride`. And it doesn't return anything, so the docs should be updated to match the Apple docs. And check the line-break for `description`, you need to break it into multiple lines here.

iphone/Classes/TiUIClipboardProxy.m
```diff
@@ -65,18 +65,110 @@ static ClipboardType mimeTypeToDataType(NSString *mimeType)
...
 @implementation TiUIClipboardProxy
 
+-(id) init
+{
+    if (self = [super init])
+    {
+        shouldCreatePasteboard = true;
+        persistent = false;
+        isNamedPasteBoard = false;
+        isUnique = false;
```
hansemannn: We should try to avoid all these helper flags. I need to go through the code in detail to check what's required. Thinkng about it, we should move all named-pasteboard-related properties to `[self pasteboard]` and check if they exist from there. If they do, we create a named one, otherwise not. Then we also don't need the manual setters and are able to configure the named pasteboard more centralized.

hansemannn (changes requested):

apidoc/Titanium/UI/Clipboard/Clipboard.yml
vijaysingh-axway: Put these under properties section.

apidoc/Titanium/UI/Clipboard/Clipboard.yml
vijaysingh-axway: made this allowCreation.

iphone/Classes/TiUIClipboardProxy.m
vijaysingh-axway: We need these variables as pasteboard will be created on combination of these. If we create  pasteboard using [self pasteboard] , all calculation will happen inside pasteboard function and every time we use pasteboard (e.g read name ) [self pasteboard] will get called and all calculation will happen again. So I think current implementation is fine.

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -151,7 +151,64 @@ methods:
...
     since: "5.5.0"
     osver: {ios: {min: "10.0"}}
+    platforms: [iphone, ipad]  
+
+  - name: remove
+    summary: Removes the clipboard.
+    since: "6.1.0"
+    platforms: [iphone, ipad]
+
+  - name: persistent
```
hansemannn: This is a property, no method.

hansemannn (changes requested):

apidoc/Titanium/UI/Clipboard/Clipboard.yml
vijaysingh-axway: Done.

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -153,6 +153,63 @@ methods:
...
+    description: |
+        When a clipboard is persistent, it continues to exist past app terminations and across system 
+        reboots. Clipboard that are not persistent only last until the owning (creating) app quits.
+    parameters:
+      - name: persistent
+        summary: value of persistent.
+        type: Boolean
+        default: false in case of named clipboard otherwise true.
+    since: "6.1.0"
+    osver: {ios: {max: "9.x.x"}}
```
hansemannn: Actually, we should leave this whole property out, as we should not include deprecated iOS-API's in new Ti-API's:
> Persistent named pasteboards are deprecated in iOS 10. The system sets the persistence of all pasteboards automatically, with named pasteboards marked as nonpersistent and the systemwide general pasteboard marked as persistent. If you try to set the setPersistent(_:) property on a pasteboard, Xcode issues a deprecation warning.
Instead of persistent named pasteboards, used shared containers. For information on shared containers, read about the containerURL(forSecurityApplicationGroupIdentifier:) method in FileManager, and read about the com.apple.security.application-groups entitlement key in Adding an App to an App Group in Entitlement Key Reference.

Thoughts?

vijaysingh-axway: Agreed. Default behavior will be same in below iOS10 and iOS10 above as well. Will remove this property.

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -153,6 +153,49 @@ methods:
...
+    parameters:
+      - name: unique
+        summary: value of unique.
+        type: Boolean
+    since: "6.1.0"
+    platforms: [iphone, ipad]
+    availability: creation
+
+  - name: allowCreation
+    summary: Create a clipboard identified by name if it doesn,t exist.
```
hansemannn: `doesn,t` -> `doesn't`

apidoc/Titanium/UI/Clipboard/Clipboard.yml
```diff
@@ -153,6 +153,49 @@ methods:
...
+    description: |
+        If create is false, it will return the existing named clipboard but will not create new one.
```
hansemannn: Please state that "create" is a property. Either by using `<Titanium.UI.Clipboard.create>` or just using `"create"`. Same in the sentence afterwards.

hansemannn (changes requested):

hansemannn (approved):

ssjsamir (approved):
FR Passed, was able to see the named pasteboard using the application code provided in https://jira.appcelerator.org/browse/TIMOB-23799.

**Test Steps** 
+ Built the SDK from this PR
+ Copied the test case from https://jira.appcelerator.org/browse/TIMOB-23799 
+ Pasted the test case into a titanium project 
+ Ran the titanium project 
+ Pressed the trigger button on the screen 
+ Saw the following output in the console:
```
[INFO] :   clipboard name is: test
[INFO] :   clipboard text is: Hello
[INFO] :   Unique clipboard name is: 8E393EE2-2163-461A-B885-D3F078945669
[INFO] :   Unique clipboard text is:  Hello
```
Previously the console output would be: (6.0.3.GA):
```
[INFO] :   clipboard name is: test
[INFO] :   clipboard text is: Hello
[INFO] :   Unique clipboard name is: undefined
[INFO] :   Unique clipboard text is:  Hello
```

**Environement**
Appcelerator Command-Line Interface, version 6.1.0
iPhone 6s+ (iOS 10.0.1)
Node.js Version: 4.6.0
npm: 4.2.8
Xcode: 8.2
Appcelerator Studio: 4.8.1.201612050850

PR #8823, opened 2017-02-09
[TIMOB-23496] TiAPI Fix: Converge on same Object properties/types for…

maggieaxway:
… GeocodedAddress

### DIFF
--- a/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
+++ b/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
@@ -335,13 +335,12 @@ public class TiLocation implements Handler.Callback
 		address.put(TiC.PROPERTY_CITY, place.optString(TiC.PROPERTY_CITY, ""));
 		address.put(TiC.PROPERTY_REGION1, ""); // AdminArea
 		address.put(TiC.PROPERTY_REGION2, ""); // SubAdminArea
-		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString("zipcode", ""));
+		address.put(TiC.PROPERTY_ZIPCODE, place.optString(TiC.PROPERTY_ZIPCODE, ""));
 		address.put(TiC.PROPERTY_COUNTRY, place.optString(TiC.PROPERTY_COUNTRY, ""));
 		address.put(TiC.PROPERTY_STATE, place.optString(TiC.PROPERTY_STATE, ""));
-		address.put("countryCode", place.optString(TiC.PROPERTY_COUNTRY_CODE, "")); // TIMOB-4478, remove this later, was old android name
-		address.put(TiC.PROPERTY_COUNTRY_CODE, place.optString(TiC.PROPERTY_COUNTRY_CODE, ""));		
-		address.put(TiC.PROPERTY_LONGITUDE, place.optString(TiC.PROPERTY_LONGITUDE, ""));
-		address.put(TiC.PROPERTY_LATITUDE, place.optString(TiC.PROPERTY_LATITUDE, ""));
+		address.put(TiC.PROPERTY_COUNTRY_CODE, place.optString(TiC.PROPERTY_COUNTRY_CODE, ""));
+		address.put(TiC.PROPERTY_LONGITUDE, place.optDouble(TiC.PROPERTY_LONGITUDE));
+		address.put(TiC.PROPERTY_LATITUDE, place.optDouble(TiC.PROPERTY_LATITUDE));
 		address.put(TiC.PROPERTY_DISPLAY_ADDRESS, place.optString(TiC.PROPERTY_ADDRESS));
 		address.put(TiC.PROPERTY_ADDRESS, place.optString(TiC.PROPERTY_ADDRESS));
 
--- a/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
+++ b/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
@@ -344,6 +344,8 @@ public class TiLocation implements Handler.Callback
 		address.put(TiC.PROPERTY_DISPLAY_ADDRESS, place.optString(TiC.PROPERTY_ADDRESS));
 		address.put(TiC.PROPERTY_ADDRESS, place.optString(TiC.PROPERTY_ADDRESS));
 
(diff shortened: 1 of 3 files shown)


android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
```diff
@@ -335,13 +335,12 @@ private KrollDict buildAddress(JSONObject place)
 		address.put(TiC.PROPERTY_CITY, place.optString(TiC.PROPERTY_CITY, ""));
 		address.put(TiC.PROPERTY_REGION1, ""); // AdminArea
 		address.put(TiC.PROPERTY_REGION2, ""); // SubAdminArea
-		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString("zipcode", ""));
+		address.put(TiC.PROPERTY_ZIPCODE, place.optString(TiC.PROPERTY_ZIPCODE, ""));
```
hansemannn: This is a breaking change. We need to retain the old values until the next major release.

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
```diff
@@ -335,13 +335,12 @@ private KrollDict buildAddress(JSONObject place)
...
 		address.put(TiC.PROPERTY_COUNTRY, place.optString(TiC.PROPERTY_COUNTRY, ""));
 		address.put(TiC.PROPERTY_STATE, place.optString(TiC.PROPERTY_STATE, ""));
-		address.put("countryCode", place.optString(TiC.PROPERTY_COUNTRY_CODE, "")); // TIMOB-4478, remove this later, was old android name
```
hansemannn: Breaking change as well. `countryCode` != `country_code`.

apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -1008,13 +1008,13 @@ properties:
...
     summary: Longitude of the geocoded address.
-    type: String
+    type: Number
     since: { mobileweb: "3.1.1" }
     platforms: [android, iphone, ipad, mobileweb]
 
   - name: latitude
     summary: Latitude of the geocoded address.
-    type: String
+    type: Number
```
hansemannn: Need to leave a note in the description that is was unified in 6.1.0.

hansemannn (changes requested):
Left some noted to the breaking changes this PR causes. We should retain all old values and remove the duplicate ones in a follow-up 7.0.0 ticket. 

Also, I removed the 6.0.2 version. This is no 6.0.0 regression, so it should go into the next feature release (6.1.0) instead.

maggieaxway:
Test case:

  backgroundColor: 'white',
  exitOnClose: true,
  fullscreen: false,
  layout: 'vertical',
  title: 'TIMOB-23492'
});

  color: '#900',
  font: { fontSize:18 },
  textAlign: Ti.UI.TEXT_ALIGNMENT_LEFT,
  width: Ti.UI.SIZE, height: Ti.UI.SIZE
});

Ti.Geolocation.reverseGeocoder(1.278509, 103.848656, function (response) {

    var location = JSON.stringify(response, null, ' ');
    Ti.API.info(location);
    label1.setText(response.places[0].country_code);

});

win.add(label1);
win.open();

//====== Log output ======//

02-16 20:38:29.928 28645-28674/com.titanium.test I/TiAPI:  {
                                                           "source": {
                                                            "locationServicesEnabled": true,
                                                            "hasCompass": true,
                                                            "lastGeolocation": null,
                                                            "apiName": "Ti.Geolocation",
                                                            "bubbleParent": true,
                                                            "__propertiesDefined__": true
                                                           },
                                                           "code": 0,
                                                           "places": [
                                                            {
                                                             "address": "71 Robinson Road, 068895 Singapore, Singapore",
                                                             "street": "Robinson Road",
                                                             "country_code": "SG",
                                                             "country": "Singapore",
                                                             "street1": "Robinson Road",
                                                             "longitude": 103.848796,
                                                             "region2": "",
                                                             "zipcode": "068895",
                                                             "city": "Singapore",
                                                             "latitude": 1.2783746,
                                                             "region1": "",
                                                             "state": "",
                                                             "displayAddress": "71 Robinson Road, 068895 Singapore, Singapore"
                                                            }
                                                           ],
                                                           "success": true
                                                          }


//+++++++++++++++++++//

Related ticket: https://jira.appcelerator.org/browse/TIMOB-23492

Related docs: http://docs.appcelerator.com/platform/latest/#!/api/GeocodedAddress


Concerns: 

country_code (on all platforms), and countryCode (only on Android) (we should likely have the same property, and given our naming scheme, it should be called "countryCode")

If we change the property name to countryCode will affect most current users. The JSONObject name is country_code, that's why the TiC.java line 1195 is this:

	public static final String PROPERTY_COUNTRY_CODE = "country_code";

garymathews:
Update the `TiC.PROPERTY_COUNTRY_CODE` constant to be `countryCode`. Lets also throw a warning about the removal of `postalCode` and `country_code`.

```Java
Log.w(TAG, "postalCode has been removed please use zipCode instead!");
Log.w(TAG, "country_code has been removed please use countryCode instead!");
```

mukherjee2:
@maggieaxway can you please look at the merge conflicts? Is this ready for QE testing? It looks like it needs CR approval. @hansemannn , @garymathews are there still changes requested, or does CR pass?

maggieaxway:
@hansemannn , @garymathews please don't review it yet. Couldn't solve the conflict due to delete repo accidentally in the past. Will clone to a new PR

mukherjee2:
@maggieaxway can this be moved to 7.1.0? I don't think there's enough time for to resolve conflicts+CR+QE testing to make it in for release tomorrow.

hansemannn:
@mukherjee2 Yes, but likely 8.0.0 then, because it looks like a breaking change.

mukherjee2:
@hansemannn ok thanks. @eric34 can this be moved to 8.0? Seems that conflicts are not straightforward to resolve, and there are breaking changes which affects multiple platforms.

sgtcoolguy:
I think this PR and ticket need to be revisited badly. We should try and get any deprecations into 7.x ASAP if we know we'll be changing the API here. I also suspect that we should be able to retain the existing properties as deprecated for now and introduce the new properties without necessarily causing breaking changes.

hansemannn:
@sgtcoolguy Agree, let's try that for 7.0.0 GA. Scheduling for tomorrow.

hansemannn:
Superseded by https://github.com/appcelerator/titanium_mobile/pull/9647.

PR #8824, opened 2017-02-09
[TIMOB-24394] Removed call to PackageApplication from iOS device builds.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24394

### DIFF
--- a/iphone/cli/hooks/install.js
+++ b/iphone/cli/hooks/install.js
@@ -21,43 +21,19 @@ exports.init = function (logger, config, cli) {
 		post: function (builder, finished) {
 			if (cli.argv.target !== 'device') return finished();
 
-			var devices = {};
+			if (cli.argv['build-only']) {
+				logger.info(__('Performed build only, skipping installing of the application'));
+				return finished();
+			}
 
-			async.parallel([
-				function (next) {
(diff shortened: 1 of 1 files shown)


iphone/cli/hooks/install.js
```diff
@@ -21,44 +21,20 @@ exports.init = function (logger, config, cli) {
 		post: function (builder, finished) {
 			if (cli.argv.target !== 'device') return finished();
 
-			var devices = {};
-
-			async.parallel([
```
hansemannn: So we don't need to perform the method in the `parallel` queue?

cb1kenobi: No. The old code was calling `PackageApplication` at the same time as re-detecting all connected iOS devices. Since we aren't calling `PackageApplication` anymore, there is only the device re-detection, so no need to parallelize anything.

hansemannn: thx for clarifying

ewieberappc:
FR Passed. Tested using:
MacOS 10.12 (16A323)
Studio 4.8.1.201701192045
Ti SDK 6.0.2 (checked PR on 6.0.X branch)
Appc NPM 4.2.8
Appc CLI 6.1.0
Alloy 1.9.5
Xcode 8.3b2 and 8.2.1 (8C1002)

Built multiple projects for device and did not witness the error/warning. Tested using Xcode 8.3 and 8.2.1, with Alloy and Classic apps.

PR #8827, opened 2017-02-09
[TIMOB-24395] iOS: Fixed differential dist builds.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24395

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -2588,6 +2588,11 @@ iOSBuilder.prototype.checkIfNeedToRecompile = function checkIfNeedToRecompile() 
 			return true;
 		}
 
+		if (this.target === 'dist-adhoc' || this.target === 'dist-appstore') {
+			this.logger.info(__('Forcing rebuild: distribution builds require \'xcodebuild\' to be run so that resources are copied into the archive'));
+			return true;
+		}
+
 		if (fs.existsSync(this.xcodeProjectConfigFile)) {
 			// we have a previous build, see if the app id changed
 			var conf = fs.readFileSync(this.xcodeProjectConfigFile).toString(),
@@ -6083,6 +6088,16 @@ iOSBuilder.prototype.removeFiles = function removeFiles(next) {
 	this.unmarkBuildDirFile(path.join(this.xcodeAppDir, 'Info.plist'));
 	this.unmarkBuildDirFile(path.join(this.xcodeAppDir, 'PkgInfo'));
 	this.unmarkBuildDirFile(path.join(this.xcodeAppDir, 'embedded.mobileprovision'));
+
+	this.unmarkBuildDirFiles(path.join(this.buildDir, 'export_options.plist'));
+	this.unmarkBuildDirFiles(path.join(this.buildDir, this.tiapp.name + '.xcarchive'));
+
+	try {
+		var releaseDir = path.join(this.buildDir, 'build', 'Products', 'Release-iphoneos');
+		if (fs.lstatSync(path.join(releaseDir, this.tiapp.name + '.app')).isSymbolicLink()) {
+			this.unmarkBuildDirFiles(releaseDir);
+		}
+	} catch (e) {}
 
 	this.logger.info(__('Removing files'));
 
--- a/iphone/cli/hooks/package.js
+++ b/iphone/cli/hooks/package.js
@@ -112,31 +112,6 @@ exports.init = function (logger, config, cli) {
 						}
 					});
 
-					var now = new Date;
-					var destInfoPlist = path.join(stagingArchiveDir, 'Info.plist');
-					if (!fs.existsSync(destInfoPlist)) {
-						var origPlist = new appc.plist(path.join(builder.buildDir, 'Info.plist'));
-						var newPlist = new appc.plist();
-						var appBundle = 'Applications/' + name + '.app';
-
-						appc.util.mix(newPlist, {
-							ApplicationProperties: {
-								ApplicationPath: appBundle,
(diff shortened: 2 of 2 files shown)


iphone/cli/commands/_build.js
```diff
@@ -2588,6 +2588,11 @@ iOSBuilder.prototype.checkIfNeedToRecompile = function checkIfNeedToRecompile()
 			return true;
 		}
 
+		if (this.target === 'dist-adhoc' || this.target === 'dist-appstore') {
+			this.logger.info(__('Forcing rebuild: distribution builds require \'xcodebuild\' to be run so that resources are copied into the archive'));
```
hansemannn: Capitalize `distribution`. We have a lot of logs that look "unformatted", so we should at least avoid it for new ones.

iphone/cli/hooks/package.js
```diff
@@ -112,31 +112,6 @@ exports.init = function (logger, config, cli) {
...
-								IconPaths: [
-									appBundle + '/' + builder.tiapp.icon
-								]
-							},
-							ArchiveVersion: newPlist.type('integer', 2),
-							CreationDate: now,
-							Name: name,
-							SchemeName: name
-						}).save(destInfoPlist);
-					}
```
hansemannn: Thank god we don't need that anymore. Can we do the same with [this snippet](https://github.com/cb1kenobi/titanium_mobile/blob/3ea52a187c5b83086941edef6a24aa75bc5b7077/iphone/cli/hooks/package.js#L57-L80)?

iphone/cli/hooks/package.js
```diff
@@ -150,7 +125,9 @@ exports.init = function (logger, config, cli) {
 
 					// move the finished archive directory into the correct location
 					fs.existsSync(archivesDir) || wrench.mkdirSyncRecursive(archivesDir);
-					fs.renameSync(stagingArchiveDir, dest);
+					appc.fs.copyDirSyncRecursive(stagingArchiveDir, dest, {
```
hansemannn: Any reason we use the appc-fs vs the fs of this scope?

iphone/cli/commands/_build.js
```diff
@@ -6084,6 +6089,16 @@ iOSBuilder.prototype.removeFiles = function removeFiles(next) {
...
 
+	this.unmarkBuildDirFiles(path.join(this.buildDir, 'export_options.plist'));
+	this.unmarkBuildDirFiles(path.join(this.buildDir, this.tiapp.name + '.xcarchive'));
+
+	try {
+		var releaseDir = path.join(this.buildDir, 'build', 'Products', 'Release-iphoneos');
+		if (fs.lstatSync(path.join(releaseDir, this.tiapp.name + '.app')).isSymbolicLink()) {
+			this.unmarkBuildDirFiles(releaseDir);
+		}
+	} catch (e) {}
```
hansemannn: Question on this whole method: When we do a new (production) build, why don't we just nuke the whole build directory? I think people rather wait 10s more instead of having the risks of missing something from build to build, especially when Apple introduces more and more options / files.

hansemannn (changes requested):

iphone/cli/commands/_build.js
cb1kenobi: No other "Forcing rebuild:" message has the following sentence capitalized. It may be consistently wrong, but at least it's consistent.

iphone/cli/commands/_build.js
cb1kenobi: Because it saves way more than 10 seconds. It saves upwards of a minute for people that don't have an SSD.

iphone/cli/hooks/package.js
cb1kenobi: Um, no, we need the Info.plist. I had erroneously had the same code twice and thus removed the duplicate. It should run for both dist-appstore and dist-adhoc builds.

iphone/cli/hooks/package.js
cb1kenobi: Node does not have any facilities to copy a directory. Wrench does, but Wrench doesn't do logging of which files are being copied. So, I added a `copyDirSyncRecursive()` to node-appc which does the copying and logs which files are being copied.

iphone/cli/commands/_build.js
hansemannn: Not a good reason...

iphone/cli/commands/_build.js
hansemannn: And on next release Apple adds another 3 files making it even worse to keep track of the manual files.

hansemannn:
TBH, I don't like the whole differential build handling for dist builds, it's causing more possible issues for the developer than it helps. Devs usually don't do multiple differential dist builds, and if they do, they always clean the project to ensure it's cleaned anyway (I can also speak about me, having done about 100+ AppStore uploads with production apps. But if we want this, it's fine for QE-testing.

hansemannn (approved):

cb1kenobi:
You can not like the differential builds all day, but built time is a huge issue and it's been over 6 years and the iOS platform team still hasn't figured out how to pre-compile the almost 300 source files into a single static library which would shave off a significant amount of build time.

ewieberappc (approved):
FR Passed. Tested using:
MacOS 10.12 (16A323)
Studio 4.8.1.201701192045
Ti SDK 6.0.2 (tested on 6_0_X branch)
Appc NPM 4.2.8
Appc CLI 6.1.0
Alloy 1.9.5
Xcode 8.3 (8W120l)

Performed subsequent builds on Alloy and Classic projects and did not encounter any errors. Then tested differential builds by packaging projects multiple times for Ad Hoc and App Store, with minor, trackable changes. Package contents were as expected on each build.

PR #8847, opened 2017-02-22
TIMOB-20498 Android: Added deprecated tags for the deprecated properties animation

antw12:
**JIRA:** 
https://jira.appcelerator.org/browse/TIMOB-20498

**Description:**
Added Deprecated tags for the deprecated properties for android 6.0

**Test Case:**
```
var win2 = Ti.UI.createWindow({fullscreen:false});
win2.open({
    activityEnterAnimation: Ti.Android.R.anim.slide_in_left,
    activityExitAnimation: Ti.Android.R.anim.slide_out_right
});
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
@@ -132,40 +132,40 @@ public class WindowProxy extends TiWindowProxy implements TiActivityWindow
 		super.close(arg);
 	}
 
-	@Override
-	protected void handleOpen(KrollDict options)
-	{
-		Activity topActivity = TiApplication.getAppCurrentActivity();
-		// Don't open if app is closing or closed
-		if (topActivity == null || topActivity.isFinishing()) {
-			return;
-		}
-		Intent intent = new Intent(topActivity, TiActivity.class);
-		fillIntent(topActivity, intent);
+    @Override
+    protected void handleOpen(KrollDict options)
+    {
+        Activity topActivity = TiApplication.getAppCurrentActivity();
+        // Don't open if app is closing or closed
+        if (topActivity == null || topActivity.isFinishing()) {
+            return;
+        }
+        Intent intent = new Intent(topActivity, TiActivity.class);
+        fillIntent(topActivity, intent);
 
-		int windowId = TiActivityWindows.addWindow(this);
-		intent.putExtra(TiC.INTENT_PROPERTY_USE_ACTIVITY_WINDOW, true);
-		intent.putExtra(TiC.INTENT_PROPERTY_WINDOW_ID, windowId);
+        int windowId = TiActivityWindows.addWindow(this);
+        intent.putExtra(TiC.INTENT_PROPERTY_USE_ACTIVITY_WINDOW, true);
+        intent.putExtra(TiC.INTENT_PROPERTY_WINDOW_ID, windowId);
 
-		boolean animated = TiConvert.toBoolean(options, TiC.PROPERTY_ANIMATED, true);
-		if (!animated) {
-			intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
-			topActivity.startActivity(intent);
-			topActivity.overridePendingTransition(0, 0);
-		} else if (options.containsKey(TiC.PROPERTY_ACTIVITY_ENTER_ANIMATION) || options.containsKey(TiC.PROPERTY_ACTIVITY_EXIT_ANIMATION)) {
-			topActivity.startActivity(intent);
-			logDeprecatedProperties();
-			int enterAnimation = TiConvert.toInt(options.get(TiC.PROPERTY_ACTIVITY_ENTER_ANIMATION), 0);
-			int exitAnimation = TiConvert.toInt(options.get(TiC.PROPERTY_ACTIVITY_EXIT_ANIMATION), 0);
-			topActivity.overridePendingTransition(enterAnimation, exitAnimation);
-		} else {
-			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-				topActivity.startActivity(intent, createActivityOptionsBundle(topActivity));
-			} else {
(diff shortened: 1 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -674,12 +674,18 @@
 
 	/**
 	 * @module.api
+	 * @deprecated as of Android 6.0 this property will be replaced by PROPERTY_ENTER_TRANSITION
+	 * and will be removed in android 7.0
```
hansemannn: We talked about SDK 6.0.0 and 7.0.0, not Android 6.0.0 and 7.0.0. We can deprecate it in 6.1.0 and remove in 8.0.0, 7.0.0 (one major release) is too close.

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -674,12 +674,18 @@
...
 	 */
+	@Deprecated
```
hansemannn: Also add the docs so people actually now about the deprecation. Adding a `@deprecated` annotation does not automatically throw a warning, so they should have the docs to lookup.

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -153,6 +153,7 @@ protected void handleOpen(KrollDict options)
 	        topActivity.startActivity(intent);
 	        topActivity.overridePendingTransition(0, 0);
 	    } else if (options.containsKey(TiC.PROPERTY_ACTIVITY_ENTER_ANIMATION) || options.containsKey(TiC.PROPERTY_ACTIVITY_EXIT_ANIMATION)) {
+			logDeprecatedPropertiesWarning();
 	        topActivity.startActivity(intent);
```
hansemannn: Indentation

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -594,4 +596,11 @@ public String getApiName()
 	{
 		return "Ti.UI.Window";
 	}
+
+	private void logDeprecatedPropertiesWarning() {
+		Log.w("Deprecated properties", "These properties are deprecated as of SDK 6.1.0: " + TiC.PROPERTY_ACTIVITY_ENTER_ANIMATION + " and " + TiC.PROPERTY_ACTIVITY_EXIT_ANIMATION);
```
hansemannn: How does a log look in the compiled version? And please add a note to what to use instead.

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -594,4 +595,11 @@ public String getApiName()
 	{
 		return "Ti.UI.Window";
 	}
+
+	private void logDeprecatedPropertiesWarning() {
```
garymathews: You should just call `Log.w` where you need to, no need to create another method.

antw12: the "please use" i believe wont work as the properties exist, however are not being use in the method call and so  would error so for now i will just print a warning about the depreciation of the animation properties

antw12: the properties that they should now use aka the transition properties, are not uses in the method and so it will fail and so i will leave that part out of the warning

garymathews: I don't understand?

```Java
import org.appcelerator.titanium.TiC;
```

Allows you to access any constant in TiC, regardless if it's used in the method or not.

antw12: logDeprecatedPropertiesWarning();
 +			topActivity.startActivity(intent);
 +			int enterAnimation = TiConvert.toInt(options.get(TiC.PROPERTY_ACTIVITY_ENTER_ANIMATION), 0);
 +			int exitAnimation = TiConvert.toInt(options.get(TiC.PROPERTY_ACTIVITY_EXIT_ANIMATION), 0);
 +			topActivity.overridePendingTransition(enterAnimation, exitAnimation);


the convert methods use the old variables as they have not been changed to the new ones yet and so if the user uses the new ones it will fail wont it ?

hansemannn (changes requested):

antw12:
screen shot 2017-02-24 at 9 40 28 am

fmerzadyan:
**FT PASSED**. For CR: may require indention fixes as mentioned by @hansemannn, also not sure what the tag should be for the [Log.w tag](https://github.com/appcelerator/titanium_mobile/pull/8847/files?diff=unified#diff-02d5a32acd7c17340f094404dd82341cR600). Do we have a standard tag for deprecated properties or should we just use the tag from the corresponding file e.g. `WindowProxy`? Also, not sure if we need [function](https://github.com/appcelerator/titanium_mobile/pull/8847/files?diff=unified#diff-02d5a32acd7c17340f094404dd82341cR599) for logging deprecated properties since it makes sense if we intend to deprecate more stuff in the future otherwise it may add to the clutter.

jquick-axway:
We don't want to deprecate this API.  At least not yet.  The reason is because the activity transition APIs are only supported on Android 5.0 and newer OS versions, which is a problem on Android 4.x since Titanium still supports those older OS versions.  This means that the animation APIs (that we want to deprecate) are still needed in order for developer to set the enter/exit animations on all Android OS versions, including Android 4.x.

We can't deprecate the animation APIs until we drop support for Android 4.x (which won't happen in the near future) or until we've replicated the "activityEnterTransition" and "activityExitTransition" APIs on Android 4.x ourselves (which is what I recommend).  Please see the following JIRA case.
https://jira.appcelerator.org/browse/TIMOB-24650

Closing this PR for now.

PR #8848, opened 2017-02-23
[TIMOB-20448] Android: Adding text color to a singular row adds the color to multiple down the list

frankieandone:
Test Case:
Toggle color and background color values. Expect sane behaviour => White text color if background is dark else black. Only row 0 has color green and other rows do not change colors when scrolling.
```
var array = [];
for (var i = 0; i < 20; i++) {
	var obj = {
		title : "Row "+ i,
		backgroundColor: "black"
		// backgroundColor: "white"
	};
	array.push(obj);
}
array[0].color =  "green";
var tableView = Titanium.UI.createTableView({
	data: array
});
win.add(tableView);
win.open();
```
**JIRA:** https://jira.appcelerator.org/browse/timob-20448

### DIFF
--- a/apidoc/Titanium/UI/TableViewRow.yml
+++ b/apidoc/Titanium/UI/TableViewRow.yml
@@ -245,6 +245,8 @@ properties:
   - name: color
     summary: Default text color of the row when not selected, as a color name or hex triplet.
     description: |
+        Default text color is black if background color is closer to white otherwise default text color is white.
+        
         For information about color values, see the "Colors" section of <Titanium.UI>. 
         
         On Android, `selectedColor` is not supported, so the text is always displayed in this color.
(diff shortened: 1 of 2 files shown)


maggieaxway:
CR FT pass

apidoc/Titanium/UI/TableViewRow.yml
```diff
@@ -245,6 +245,8 @@ properties:
   - name: color
     summary: Default text color of the row when not selected, as a color name or hex triplet.
     description: |
+        Default text color is black if background color is closer to white otherwise default text color is white.
```
hansemannn: This is Android-only, move to the Android notes paragraph in 252.

hansemannn (changes requested):

longton95 (approved):
FR passed.

Tested using the test case above and the demo code in the ticket, the bug is no longer present.

ENV
```
SDK Version : Local 6.1.0
Mac OS Version : 10.12.3
Appc CLI : 6.1.0
Appc NPM : 4.2.9-1
Node : v4.6.0
Device: Nexus 5x(6.0.1) Pixel XL(7.1.1)
Emulator: (6.0), (4.4.2)
```

PR #8849, opened 2017-02-24
[TIMOB-24389] iOS: Fix leak when selecting/taking images

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24389


mukherjee2:
I still see the issue after this PR. This is my environment:
NPM Version: 2.15.9
Node Version: 4.5.0
Mac OS: 10.12.3
Appc CLI: 6.1.0
Appc CLI NPM: 4.2.8
iOS Device: 10

I tested this on main thread, and with jscore, and in both case, the app crashed after taking picture 25-30 times.

PR #8852, opened 2017-02-28
TIMOB-16080: Added New sound property for views

antw12:
**JIRA:** 
https://jira.appcelerator.org/browse/TIMOB-16080

**Description:**
Added a new property PROPERTY_TOUCH_SOUND_ENABLED allowing a user to set sound effects for views on or off.

Default for property is true

**Test Case**
```
    backgroundColor: 'black',
});

var view = Ti.UI.createView({
    soundEffectsEnabled : false,
    backgroundColor : "Black"
});
win.add(view);
win.open();
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiC.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiC.java
@@ -2746,6 +2746,11 @@ public class TiC
 	/**
 	 * @module.api
 	 */
+	public static final String PROPERTY_SOUND_EFFECTS_ENABLED = "soundEffectsEnabled";
+
+	/**
+	 * @module.api
+	 */
 	public static final String PROPERTY_TRANSFORM = "transform";
 
 	/**
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -116,6 +116,7 @@ public abstract class TiViewProxy extends KrollProxy implements Handler.Callback
 		pendingAnimationLock = new Object();
 
 		defaultValues.put(TiC.PROPERTY_TOUCH_ENABLED, true);
+		defaultValues.put(TiC.PROPERTY_SOUND_EFFECTS_ENABLED, true);
 		defaultValues.put(TiC.PROPERTY_BACKGROUND_REPEAT, false);
 		defaultValues.put(TiC.PROPERTY_VISIBLE, true);
 		defaultValues.put(TiC.PROPERTY_ENABLED, true);
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
@@ -1719,9 +1719,10 @@ public abstract class TiUIView
 		if (touchable == null) {
 			return;
 		}
+
 		if (proxy.hasProperty(TiC.PROPERTY_TOUCH_ENABLED)) {
 			boolean enabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_ENABLED), true);
-				touchable.setEnabled(enabled);
+			touchable.setEnabled(enabled);
 		}
 		//Checking and setting touch sound for view
 		if (proxy.hasProperty(TiC.PROPERTY_SOUND_EFFECTS_ENABLED)) {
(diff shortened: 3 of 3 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -2746,6 +2746,11 @@
 	/**
 	 * @module.api
 	 */
+	public static final String PROPERTY_TOUCH_SOUND_ENABLED = "soundEnabled";
```
hansemannn: We should name it the same (`soundEffectsEnabled`), which would be mapped to `PROPERTY_SOUND_EFFECTS_ENABLED`.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1726,8 +1726,14 @@ protected void registerForTouch(final View touchable)
 				touchable.setEnabled(false);
 			}
 		}
+		//Checking and setting touch sound for view
+		if (proxy.hasProperty(TiC.PROPERTY_TOUCH_SOUND_ENABLED)) {
+		 boolean soundEnabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_SOUND_ENABLED), true);
```
hansemannn: Indentation incorrect.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1726,8 +1726,14 @@ protected void registerForTouch(final View touchable)
...
+			if (!soundEnabled) {
```
hansemannn: If this is `false` first, setting it to `true` will skip every time. Rather call the setter directly.

hansemannn (changes requested):

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
antw12: It is set to true as default as that is the native behavior, I tested it and it works so i am not sure, if I am still wrong please let me know

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1719,15 +1719,20 @@ protected void registerForTouch(final View touchable)
...
 		}
-
 		if (proxy.hasProperty(TiC.PROPERTY_TOUCH_ENABLED)) {
 			boolean enabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_TOUCH_ENABLED), true);
 			if (!enabled) {
 				touchable.setEnabled(false);
 			}
 		}
+		//Checking and setting touch sound for view
+		if (proxy.hasProperty(TiC.PROPERTY_SOUND_EFFECTS_ENABLED)) {
```
garymathews:
```Java
if (proxy.hasProperty(TiC.PROPERTY_SOUND_EFFECTS_ENABLED)) {
	boolean soundEnabled = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_SOUND_EFFECTS_ENABLED), true);
	touchable.setSoundEffectsEnabled(soundEnabled);
}
```

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -116,6 +116,7 @@ public TiViewProxy()
 		pendingAnimationLock = new Object();
 
 		defaultValues.put(TiC.PROPERTY_TOUCH_ENABLED, true);
+		defaultValues.put(TiC.PROPERTY_SOUND_EFFECTS_ENABLED, true);
```
garymathews: This is necessarily needed since it's already enabled by default

antw12:
for the PROPERTY_TOUCH_ENABLED should i change it to match, as it is set up the same with default as true?

garymathews:
@antw12 Yes, as that will not allow you to set `touchEnabled` back to `true`. Nice catch 👍

garymathews:
CR: **PASS** (note: try not to change the formatting of code you are not changing)
FT: **PASS**

ssjsamir:
FR Passed, was able to set `soundEffectsEnabled` to `true` or `false` and can confirm if it was set to to `true` sound effects can be heard and if set to `false` sound effects can not be heard. 

Steps I took to test:
+ Copy the code provided in the description
+ Ran the program
+ Was not able to hear the sound effects 
+ Edited the code and set `soundEffectsEnabled` to `true` 
+ Was able to hear the sound effects 

Environment

Appcelerator Command-Line Interface, version 6.1.0
Google Pixel XL (Android 7.1.1)
Node.js Version: 4.6.0
npm: 4.2.8
Xcode: 8.2
Appcelerator Studio: 4.8.1.201612050850

PR #8853, opened 2017-03-01
[TIMOB-24389] : iOS Appcelerator Titanium App Crashes after some usage time

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24389

### DIFF
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -1441,7 +1441,7 @@ MAKE_SYSTEM_PROP(VIDEO_TIME_OPTION_EXACT,MPMovieTimeOptionExact);
 		[NSThread detachNewThreadSelector:@selector(dispatchCallback:) toTarget:self withObject:[NSArray arrayWithObjects:@"success",event,listener,nil]];
 #else
 		[self dispatchCallback:@[@"success",event,listener]];
-        	//TIMOB-24389 : Heap was not getting release early. This will cause to free memory than usual time.
+        	//TIMOB-24389 : As memory flushing was taking more time, so memory was growing continuously causing crash of app.
         	KrollContext *krollContext = [self.pageContext krollContext];
         	[krollContext forceGarbageCollectNow];
 #endif
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -1394,6 +1394,12 @@ MAKE_SYSTEM_PROP(VIDEO_TIME_OPTION_EXACT,MPMovieTimeOptionExact);
     [NSThread sleepForTimeInterval:0.5];
     [self _fireEventToListener:type withObject:object listener:listener thisObject:nil];
     [pool release];
+    
+#ifndef TI_USE_KROLL_THREAD
+    //TIMOB-24389: Force the heap to be GC'd to avoid Ti.Blob references to be dumped.
+    KrollContext *krollContext = [self.pageContext krollContext];
+    [krollContext forceGarbageCollectNow];
+#endif
 }
 
 -(void)sendPickerError:(int)code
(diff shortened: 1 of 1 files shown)


iphone/Classes/MediaModule.m
```diff
@@ -2095,8 +2095,14 @@ - (void)imagePickerController:(UIImagePickerController *)picker_ didFinishPickin
     if (cropRect != nil) {
         [dictionary setObject:cropRect forKey:@"cropRect"];
     }
-	
+    
     [self sendPickerSuccess:dictionary];
+    
+#ifndef TI_USE_KROLL_THREAD
+    //TIMOB-24389 : iOS: Appcelerator Titanium App Crashes after some usage time
```
hansemannn: Describe the issue rather then paste the ticket title. That will help others to get the details more quickly

iphone/Classes/MediaModule.m
```diff
@@ -2095,8 +2095,14 @@ - (void)imagePickerController:(UIImagePickerController *)picker_ didFinishPickin
...
+    KrollContext *krollContext = [self.pageContext krollContext];
+    [krollContext forceGarbageCollectNow];
+#endif
```
hansemannn: Can we move this to `dispatchCallback:` instead, so it's ensured to be called for video capturing & editing as well? Or do you see down-sides there?

hansemannn (changes requested):
My use-cases worked, although I'm still curious why the app would cause because of a memory issue without calling the `didReceiveMemoryWarning:` before, which would be a much better place for this method.

iphone/Classes/MediaModule.m
vijaysingh-axway: No side effect. Moved to dispatchCallback.

iphone/Classes/MediaModule.m
vijaysingh-axway: Wrote small note about problem. In ticket I think most of the things are explained.

iphone/Classes/MediaModule.m
```diff
@@ -1441,6 +1441,9 @@ -(void)sendPickerSuccess:(id)event
 		[NSThread detachNewThreadSelector:@selector(dispatchCallback:) toTarget:self withObject:[NSArray arrayWithObjects:@"success",event,listener,nil]];
 #else
 		[self dispatchCallback:@[@"success",event,listener]];
+        //TIMOB-24389 : Heap was not getting release early. This will cause to free memory than usual time.
+        KrollContext *krollContext = [self.pageContext krollContext];
+        [krollContext forceGarbageCollectNow];
```
hansemannn: This is still only executed in `sendPickerSuccess` and has wrong indentation. `saveCompletedForImage:` creates a blob as well, are there no leaks there? And change the comment to this:
```objc
// TIMOB-24389: Force the heap to be GC'd to avoid Ti.Blob references to be dumped.
```

iphone/Classes/MediaModule.m
```diff
@@ -1441,6 +1441,9 @@ -(void)sendPickerSuccess:(id)event
 		[NSThread detachNewThreadSelector:@selector(dispatchCallback:) toTarget:self withObject:[NSArray arrayWithObjects:@"success",event,listener,nil]];
 #else
 		[self dispatchCallback:@[@"success",event,listener]];
+        	//TIMOB-24389 : As memory flushing was taking more time, so memory was growing continuously causing crash of app.
+        	KrollContext *krollContext = [self.pageContext krollContext];
+        	[krollContext forceGarbageCollectNow];
```
hansemannn: This is still only executed in `sendPickerSuccess`. The method `saveCompletedForImage:` and others create a blob as well, are there no leaks there? And change the comment to this:
```objc
// TIMOB-24389: Force the heap to be GC'd to avoid Ti.Blob references to be dumped.
```

vijaysingh-axway: Good catch dude. Made changes.

hansemannn:
LGTM, CR **approved**!

hansemannn (approved):

ewieberappc (approved):
FR passed, using:
MacOS 10.12 (16A323)
Studio 4.8.1.201612050850
Ti SDK 6.1.0
Appc NPM 4.2.8
Appc CLI 6.1.0
Alloy 1.9.5
Xcode 8.2.1 (8C1002)

Tested using the provided sample code by taking over 50 pictures in the app, built with main-thread enabled and disabled. No crash was encountered unless JScore was also enabled. This will be addressed by another ticket/PR.

hansemannn:
@vijaysingh-axway Can you do the 6.0.x backport? Will check tomorrow morning Europe time then. Thx!

PR #8863, opened 2017-03-07
[TIMOB-20267] Android: fix slider track showing

frankieandone:
Test Case:
Expect to see track to be visible beneath transparent thumbImage.
screenshot_1496943982
```
var sldr = Ti.UI.createSlider({
	min: 1,
	max: 51,
	value: 51,
	thumbImage: "thumb.png",
	splitTrack: false
});
win.add(sldr);
win.open();
```
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20267


jquick-axway (approved):
CR: Pass

ssjsamir (approved):
FR passed, was able to see the slider track beneath the transparent thumbImage.

**Test Steps:**
+ Downloaded the SDK Build from this PR
+ Copied the the test case from description
+ Added a transparent thumbimage and background
+ Ran the application
+ Was able to see the track bar under the transparent image
screen shot 2017-06-08 at 11 23 13 am

**Test Environment**
Google Nexus 6P (7.1.1)
Nexus 6 (6.0)
Node.js Version: 6.10.1
Xcode: 8.2

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUISlider.java
```diff
@@ -91,11 +91,12 @@ public void processProperties(KrollDict d)
 		} else {
 			maxRange = max;
 		}
-		
 		if (d.containsKey("thumbImage")) {
 			updateThumb(seekBar, d);
 		}
-		
+		if (d.containsKey(TiC.PROPERTY_SPLIT_TRACK)) {
+			seekBar.setSplitTrack(TiConvert.toBoolean(d.get(TiC.PROPERTY_SPLIT_TRACK)));
```
ypbnv: Can you guard this line to apply only for API 21 and above? It causes crashes on lower levels since it was added for Lollipop.

https://developer.android.com/reference/android/widget/AbsSeekBar.html#setSplitTrack(boolean)

ypbnv (commented):
Guarding for API level.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUISlider.java
jquick-axway: Good catch Yordan.

We also have to watch out in case the property value is not a boolean.  So, the code should look something like this...

```java
if (Build.VERSION.SDK_INT  >= 21) {
	ApiLevel21.setSplitTrack(this.seekBar, TiConvert.toBoolean([[value]], false));
}

private static class ApiLevel21
{
	private ApiLevel21() {}

	public void setSplitTrack(SeekBar seekBar, boolean enabled)
	{
		if (seekBar != null) {
			seekBar.setSplitTrack(enabled);
		}
	}
}
```

The static inner class above avoids the opcode warning on the lower API levels.

ypbnv: Also here:
https://github.com/fmerzadyan/titanium_mobile/blob/08f415de7f9d395a015618ea279ef2f586f51e14/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUISlider.java#L272

PR #8886, opened 2017-03-14
[TIMOB-24484] Exclude .bundle images from xcassets

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24484

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -4678,6 +4678,7 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 		appIconRegExp = appIcon && new RegExp('^' + appIcon[1].replace(/\./g, '\\.') + '(.*)\\.png$'),
 		launchImageRegExp = /^(Default(-(Landscape|Portrait))?(-[0-9]+h)?(@[2-9]x)?)\.png$/,
 		launchLogoRegExp = /^LaunchLogo(?:@([23])x)?(?:~(iphone|ipad))?\.(?:png|jpg)$/,
+		bundleFileRegExp = /.+\.bundle\/.+($|\n)/,
 
 		resourcesToCopy = {},
 		jsFiles = {},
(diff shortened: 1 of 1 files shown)


cb1kenobi:
I don't love this solution, but it works. I recognize that `.bundle` paths are iOS resources, but checking the path seems a little lazy.

A more robust way would be to check the actual directory name being walked and passing through the `useAppThinning` flag:

```
var bundleFileRegExp = /^.+\.bundle$/;

function walk(src, dest, useAppThinning, ignore, origSrc, prefix) {
    fs.existsSync(src) && fs.readdirSync(src).forEach(function (name) {
        // <snip>
        if (srcStat.isDirectory()) {
            return walk(from, to, useAppThinning ? !bundleFileRegExp.test(name) : false, null, origSrc || src, prefix);
        }
        // <snip>
    }
}

this.logger.info(__('Analyzing Resources directory'));
walk(path.join(this.projectDir, 'Resources'),           this.xcodeAppDir, useAppThinning, ignorePlatformDirs);
walk(path.join(this.projectDir, 'Resources', 'iphone'), this.xcodeAppDir, useAppThinning);
walk(path.join(this.projectDir, 'Resources', 'ios'), this.xcodeAppDir, useAppThinning);
```

It's gross.

I propose we accept your solution and then we come up with a better way in the unified Titanium build pipeline to ignore app thinning for entire directory structures.

hansemannn:
Agree. I tried to not touch any method signatures. I would still update the regex to yours, I think mine if overkill, by explicitly allowing everything inside the bundle .

cb1kenobi:
Don't use my regex. It is designed for directory names. You could do something as simple as `/.+\.bundle\/.+/`.

hansemannn:
any character before and after the path? Alrighty!

hansemannn:
@cb1kenobi Need this in 6.1.0.

cb1kenobi:
Looks good to me. **APPROVED**

cb1kenobi (approved):

PR #8892, opened 2017-03-17
[TIMOB-20482]:iOS WebView converting spaces into extra lines

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-20482


iphone/Classes/TiUIWebView.m
```diff
@@ -715,8 +715,8 @@ -(CGFloat)contentHeightForWidth:(CGFloat)value
     CGRect oldBounds = [[self webview] bounds];
     BOOL oldVal = webview.scalesPageToFit;
     [webview setScalesPageToFit:NO];
-    [webview setBounds:CGRectMake(0, 0, 10, 1)];
-    CGFloat ret = [webview sizeThatFits:CGSizeMake(10, 1)].height;
+    [webview setBounds:CGRectMake(0, 0, value, 1)];
```
hansemannn: Where did the static 10 come from? Wondering why that particular value was used (and is also still used in `contentWidthForWidth :`).

vijaysingh-axway: @hansemannn  That value has been made 10 for ticket TIMOB-11777 .I verified the rotation as mentioned in TIMOB-117777, it is working fine  with this change as well. I couldn't get exact test case of TIMOB-11777 as this is very old ticket. If you can get exact test case we can verify same.

hansemannn: There is a test-case in TIMOB-14993 that also suggested this solution instead:
```objc


-(CGFloat)contentHeightForWidth:(CGFloat)value
{
     CGRect oldBounds = [[self webview] bounds];
     [webview setBounds:CGRectMake(0, 0, MAX(value,10), 1)];
     CGFloat result = [[webview stringByEvaluatingJavaScriptFromString:@"document.height"] floatValue];
     [webview setBounds:oldBounds];
     return result;
}
```
Thoughts?

vijaysingh-axway: This problem is similar to TIMOB-14993. In comment of TIMOB-14993 it is mentioned that with proper mime type, it will work fine else not. But this PR works fine with proper mime type and without mime type (plain text). So I think we should go ahead with this PR.

hansemannn (approved):

ewieberappc (approved):
FR passed, using:
MacOS 10.12 (16A323)
Studio 4.9.0.201704061825
Ti SDK 6.1.0
Appc NPM 4.2.9
Appc CLI 6.2.0
Alloy 1.9.11
Xcode 8.3.2 (8E2002)

No additional lines were shown when setting the webview to a local file, using setHTLP or setURL. Tested using the provided sample with modifications.

PR #8903, opened 2017-03-22
[TIMOB-23135] Fix Android: Ti.Geolocation.locationServicesEnabled retur…

maggieaxway:
…ns false

Test case:

```javascript
 
  title: 'Click Me'
});
 
  console.log('Ti.Geolocation.locationServicesEnabled ' + Ti.Geolocation.locationServicesEnabled);
 
  console.log('Ti.Geolocation.hasLocationPermissions() ' + Ti.Geolocation.hasLocationPermissions());
 
  if (Ti.Geolocation.hasLocationPermissions()) {
    getCurrentPosition();
 
  } else {
    Ti.Geolocation.requestLocationPermissions(null, getCurrentPosition);
  }
});
 
function getCurrentPosition() {
  Ti.Geolocation.getCurrentPosition(function(e) {
    console.log('Ti.Geolocation.getCurrentPosition() ' + JSON.stringify(e));
  });
}
```

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23135


android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
```diff
@@ -106,14 +106,7 @@ public boolean getLocationServicesEnabled()
...
 
-		// don't count the passive provider
-		for(String name : providerNames) {
-			if (name.equals(LocationManager.NETWORK_PROVIDER) || name.equals(LocationManager.GPS_PROVIDER)) {
-				return true;
-			}
-		}
-
-		return false;
+		return providerNames.size()!=0 & getLastKnownLocation()!=null;
```
garymathews:
```Java
return providerNames.size() > 0 && getLastKnownLocation() != null;
```

Space the conditional operators and use a logical `&&` instead of a bitwise `&` (even though the bitwise & is still valid in this case)

garymathews (changes requested):

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
maggieaxway: providerNames.size() is always >=1 because there is passive provider. 

But having passive provider doesn't guarantee there is a location available, and when there is a lastKnownLocation available doesn't guarantee there's other service provider available. That's why I didn't use short circuit &&. Does it make sense?

lokeshchdhry:
FR Passed.

`Ti.Geolocation.locationServicesEnabled` returns `true` as expected. 

Studio Ver : 4.8.1.201612050850
SDK Ver :    Local build 6.1.0
OS Ver :     10.12.3
Xcode Ver :  Xcode 8.3.1
Appc NPM :   4.2.9-3
Appc CLI :   6.2.0
Ti CLI Ver : 5.0.12
Alloy Ver :  1.9.11
Node Ver :   6.10.1
Java Ver :   1.8.0_101
Devices :    google Nexus 6 --- Android 6.0.1
google Pixel --- Android 7.1.1

hansemannn:
@maggieaxway Little confusion here: You approved your own PR, but @garymathews didn't so far. We should be cautious with that.

maggieaxway:
@hansemannn Noted thanks. Sorry for the confusion, I didn't mean to approve the PR.

PR #8918, opened 2017-03-29
[TIMOB-24364] iOS: Fix HTTP-leak on Kroll-Thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24364

- Instruments-Trace: [http-fix.trace.zip](https://github.com/appcelerator/titanium_mobile/files/879898/http-fix.trace.zip)
- Test-Cases on JIRA (GET and POST)


htbryant94 (approved):
Approved! Monitored memory allocation of both test cases with instruments, and observed that memory is correctly deallocated after operations conclude.

PR #8919, opened 2017-03-29
[TIMOB-24364] (6_0_X) iOS: Fix HTTP-leak on Kroll-Thread

hansemannn:
Backport of https://github.com/appcelerator/titanium_mobile/pull/8918.


htbryant94 (approved):
Approved! Monitored memory allocation of both test cases with instruments, and observed that memory is correctly deallocated after operations conclude.

PR #8921, opened 2017-03-29
[TIMOB-3797] Tooling: Ignore files listed in tiapp.xml

feons:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-3797

Propose to support adding ignore files in tiapp.xml, eg:
```
  <ignore-files>
    <file>*.tmp</file>
    <file>dist/*</file>
  </ignore-files>
```

### DIFF
--- a/node_modules/titanium-sdk/lib/builder.js
+++ b/node_modules/titanium-sdk/lib/builder.js
@@ -158,6 +158,12 @@ Builder.prototype.validate = function validate(logger, config, cli) {
 Builder.prototype.run = function run(logger, config, cli, finished) {
 	// note: this function must be sync!
 
+	var patterns = cli.tiapp['ignore-files'];
+	if (patterns && patterns.length > 0) {
+		var tiappIgnoreFiles = '^(' + patterns.join('|').replace(/\./g, '\\.').replace(/\*/g, '.*') + ')$';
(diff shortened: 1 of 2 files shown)


cb1kenobi:
Any particular reason to add this to the `tiapp.xml`? Several tools these days use an "ignore" file such as `.gitignore` or `.npmignore`.

hansemannn:
@cb1kenobi I think the reason was that we want to exclude those file in the Titanium build (e.g. test-files, dev-related JS files).

cb1kenobi:
I'm fine with the "feature". I'm questioning the solution. The tiapp.xml is already overloaded and inconsistent. TIMOB-3797 was created 6 years ago and today is a different day. Maybe putting it in the tiapp is not a great idea, but maybe it's not bad either.

cb1kenobi:
Here's an idea... keep this in the tiapp.xml, but instead of ignoring files, we whitelist files. This is exactly what Xcode projects and Visual Studio projects do. Then we can add additional information PER FILE and specify how files should be handled: encrypted, minified, optimized, copied, not copied, etc.

hansemannn:
I think this PR needs to be reworked and rebased, based on the feedback from Chris. Let's close it for now and revisit the changes as they get requested.

PR #8925, opened 2017-04-04
[TIMOB-24501] Android: multi-window support

frankieandone:
#### Test Case
Change target SDK to API 24+ in `AndroidManifest.xml/tiapp.xml`. Default `app.js` is sufficient.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24501


fmerzadyan:
Accidentally closed. Reopening.

jquick-axway (changes requested):
I only have 1 issue with this change. Developers should be allowed to set "resizableActivity" to false so that they can opt-out of this feature.

Also, according to Google's docs, this feature will be enabled by default if targeting API Level 24 or higher. So, the "_build.js" changes might not be necessary. Can you double check please?
   https://developer.android.com/guide/topics/ui/multi-window.html#configuring

hansemannn:
Quick interruption: If I read it correctly, it's just injected into the manifest. Why can't people just enable it today by setting `resizeableActivity="true"` in the `<application>` tag?

fmerzadyan:
It doesn't get added to the AndroidManifest.xml. When it does get added (through other means), it keeps crashing. 

But mainly, I think we would want this to be default state without user needing to add it themselves?

jquick-axway:
We should identify where it's crashing then. This is in case the developer's app has display/layout issues in a very small window. Such as split screen on a phone where the app only takes a 3rd of the display.

Also, I remember the crash was happening in an app using TabViews because they were using an Activity within an Activity. That means the TabView activities need the "configChanges" settings you've added to prevent the crash.

android/cli/commands/_build.js
```diff
@@ -3683,8 +3683,7 @@ AndroidBuilder.prototype.generateAndroidManifest = function generateAndroidManif
 		}
 	}, this);
 
-	
-	if (this.realTargetSDK >= 24 && !finalAndroidManifest.application.resizeableActivity) {
+	if (this.realTargetSDK >= 24 && !("resizeableActivity" in finalAndroidManifest.application)) {
```
cb1kenobi: Generally we prefer `hasOwnProperty()` over `in` like this:

    finalAndroidManifest.application.hasOwnProperty('resizeableActivity')

Also, single quotes, not double quotes.

fmerzadyan:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/8970 because of dependency of code.

PR #8934, opened 2017-04-05
Provide 'code' property on Error thrown by bad require call

sgtcoolguy:
When we throw an Error trying to resolve a bad require, set the 'code' property to 'MODULE_NOT_FOUND'

This is specifically to try and match Node's behavior in the same case.

While this is relatively straightforward to do on Android, I'm not sure how to achieve the same on iOS. (Pleas help @hansemannn !)
Looks like more work where I need to do something similar to:
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiBase.m#L149
and
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiBindingTiValue.m#L284


hansemannn:
So the "Cannot find Module" should include a new property `code` that has the value `MODULE_NOT_FOUND`. Checking today!

hansemannn:
@sgtcoolguy Can you merge my PR to your branch and this one?

PR #8937, opened 2017-04-06
[TIMOB-23436]:ListView Control's Height not setting in iOS

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23436


iphone/Classes/TiUIListView.m
```diff
@@ -1762,7 +1762,7 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
     }
     if (TiDimensionIsDip(height)) {
         return height.value;
-    } else if (TiDimensionIsAutoSize(height)) {
+    } else if (TiDimensionIsAutoSize(height) || TiDimensionIsUndefined(height)) {
```
hansemannn: The test-case works for me. But I am wondering if we shouldn't include this as well:
```objc
    } else if (TiDimensionIsAuto(height) || TiDimensionIsAutoSize(height) || TiDimensionIsUndefined(height)) {
```
And I also noticed that the template do has a height and width specified, so it shouldn't go until that line, can you check that before approving? Thx! Good work!

vijaysingh-axway: While setting Ti.UI.auto for row height , it gives undefined. Rather I found that for Ti.UI.Fill check similar condition should applied . Updated check with TiDimensionIsAutoFill(height). 
Thanks.

hansemannn: Mhh, I'd be really curious if checking for size and fill is a good idea. Pretty sure it breaks something else.

vijaysingh-axway: I can not think of any use case where it can break something. Are you ?

vijaysingh-axway: @hansemannn Ping.

hansemannn:
Please add more details why exactly this change fixes this issue. Also, add UI tests for all layout-specific use-cases for all if-statements available in the source. I am trying to avoid a major layout regression like in https://jira.appcelerator.org/browse/TIMOB-24753, so we better check this more detailled.

hansemannn (approved):
LGTM!

ssjsamir (approved):
**FR Passed:** Was able to set the ListView Control's height without any issues. 

**Test Steps**
+ Downloaded the SDK Build from this PR
+ Copied the test case from the comments in https://jira.appcelerator.org/browse/TIMOB-23436
+ Ran the application 
+ Was able to see the ListView height change 
+ Changed the ListView height in the code 
+ Was able to see the new height

**Test Environment**
iPhone 6S (10.0)
iPhone 6S Simulator(10.2)
Node.js Version: 6.10.1
Xcode: 8.2

PR #8942, opened 2017-04-07
[TIMOB-24465] Android: Replacing RGB_565 with ARGB_8888

kopiro:
To avoid funny gradients

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24465

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
@@ -318,7 +318,7 @@ public class TiDrawableReference
 		BitmapFactory.Options opts = new BitmapFactory.Options();
 		opts.inInputShareable = true;
 		opts.inPurgeable = true;
-		opts.inPreferredConfig = Bitmap.Config.RGB_565;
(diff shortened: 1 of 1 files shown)


hansemannn:
How about performance: http://stackoverflow.com/a/15178050/5537752 ?

garymathews:
Yep, `ARGB_8888` should be used. It is the default on Android since 4.0 and the recommended configuration for Bitmaps. Decreasing colour banding with increased bits-per-pixel.

hansemannn:
Thanks Gary! @kopiro can you update those files in this PR as well? Should be added to 6.1.0 as well.

jquick-axway:
I agree with Gary. We should always be loading images with 32-bit color. Especially since the 565 color format doesn't support alpha transparency. The days of the low-end 24 MB heap limited Android device is pretty much over.

The 3 other files source files Gary mentioned need to be changed as well.

Note:
This change will cause a merge conflict (but easy to resolve) with another change of mine where I too have updated TiDrawableReference to decode with 32-bit color.
https://jira.appcelerator.org/browse/TIMOB-24528

garymathews:
Close in favour of https://github.com/appcelerator/titanium_mobile/pull/8978

PR #8944, opened 2017-04-07
[TIMOB-23313] Disabled Swift settings for main Titanium project, but …

cb1kenobi:
[TIMOB-23313] Disabled Swift settings for main Titanium project, but ensure they are enabled in the extension targets.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23313


iphone/cli/commands/_build.js
```diff
@@ -3490,8 +3491,16 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 						});
 					}
 
-					if (hasSwiftFiles && !extBuildSettings.SWIFT_VERSION) {
-						extBuildSettings.SWIFT_VERSION = '2.2';
+					if (hasSwiftFiles) {
+						if (!extBuildSettings.SWIFT_VERSION) {
+							extBuildSettings.SWIFT_VERSION = '2.2';
```
hansemannn: Should be Swift 2.3

iphone/cli/commands/_build.js
```diff
@@ -2990,7 +2990,8 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 			DEAD_CODE_STRIPPING: 'YES',
 			SDKROOT: 'iphoneos',
 			CODE_SIGN_ENTITLEMENTS: '"' + appName + '.entitlements"'
-		};
+		},
+		legacySwift = version.lt(this.xcodeEnv.version, '8.0.0');
```
hansemannn: Please note that the legacy Swift 2.3 can only be used until Xcode 8.2.1. They removed Swift 2.3 from Xcode 8.3 and later, classic Apple move. We should (very soon) only allow Swift 3.0 and later, so we are ready for Swift 4 already - which will be backwards-compatible to Swift 3.1.

hansemannn (changes requested):

iphone/cli/commands/_build.js
cb1kenobi: Sounds like a different ticket.

hansemannn: Oh boi..okay.

cb1kenobi: I'm not opposed to changing it, but that's outside the scope of TIMOB-23313. I'm also all for dropping Swift 1 and 2.

hansemannn (approved):

PR #8947, opened 2017-04-07
[TIMOB-24531] Android: Fix TiBaseActivity windowStack

garymathews:
- Fix unwanted `windowStack` behaviour
- Fix failing `ListView` test cases


sgtcoolguy:
I do not know why this would fix the issue, but it does.

PR #8958, opened 2017-04-12
[TIMOB-24574] iOS: Fix picker-selection

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24574


apidoc/Titanium/UI/Picker.yml
```diff
@@ -353,10 +353,9 @@ properties:
         If `true`, selection indicator is enabled.
 
         On iOS 7 and later, the picker indicator is always shown and you cannot control it.
-
-        On iOS 6 and prior, if enabled, a blue bar is displayed to indicate the current selection.
+        Manually setting the property was deprecated in 6.1.0 and will be ignored by the SDK.
```
vijaysingh-axway: This is deprecated in iOS only. So iOS should be mentioned in above line.

iphone/Classes/TiUIPicker.m
```diff
@@ -190,19 +191,6 @@ -(void)setType_:(id)type_
 	}
 }
 
-// We're order-dependent on type being set first, so we need to make sure that anything that relies
-// on whether or not this is a date picker needs to be set AFTER the initial configuration.
--(void)setSelectionIndicator_:(id)value
```
vijaysingh-axway: Rather deleting the method , we should deprecate it by giving message to developers. What you think?

hansemannn: It was not used since iOS 7 anyway, so I decided to clean up the core directly. Having it or not does not have an effect to the developer.

apidoc/Titanium/UI/Picker.yml
hansemannn: It was written in the iOS-paragraph, so I hoped people will get it's iOS related, but I'll explicitly say it :-).

vijaysingh-axway (approved):
Ft and CR passed.

htbryant94 (approved):
Approved! picker selection indicator is now visible. Tested on iOS 10.2 / 9.3 Devices and Simulators.

hansemannn:
Reverting due to failing unit-tests.

PR #8959, opened 2017-04-12
[TIMOB-24574] (6_1_X) iOS: Fix picker-selection

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24574

### DIFF
--- a/apidoc/Titanium/UI/Picker.yml
+++ b/apidoc/Titanium/UI/Picker.yml
@@ -353,10 +353,9 @@ properties:
         If `true`, selection indicator is enabled.
 
         On iOS 7 and later, the picker indicator is always shown and you cannot control it.
-
-        On iOS 6 and prior, if enabled, a blue bar is displayed to indicate the current selection.
+        Manually setting the property was deprecated in 6.1.0 and will be ignored by the SDK.
     type: Boolean
-    default: true (Android), false (iPhone, iPad)
+    default: true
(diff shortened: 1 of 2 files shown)


vijaysingh-axway (commented):
1.
-// We're order-dependent on type being set first, so we need to make sure that anything that relies
 -// on whether or not this is a date picker needs to be set AFTER the initial configuration.
 --(void)setSelectionIndicator_:(id)value
 -{
 -	if (picker == nil) {
 -		[[self proxy] replaceValue:value forKey:@"selectionIndicator" notification:NO];
 -	}
 -	else if ([self isDatePicker]==NO)
 -	{
 -		[(UIPickerView*)[self picker] setShowsSelectionIndicator:[TiUtils boolValue:value]];
 -	}
 -}
 -

Rather deleting the method , we should deprecate it by giving message to developers. What you think?

2. 
+        Manually setting the property was deprecated in 6.1.0 and will be ignored by the SDK.

This is deprecated in iOS only. So iOS should be mentioned in above line.

hansemannn:
See notes in the master-PR. Also, please try to format code with 
```
\```objc
Code
\```
```
(Without the \\)

htbryant94 (approved):
Approved! picker selection indicator is now visible. Tested on iOS 10.2 / 9.3 Devices and Simulators.

hansemannn:
Reverting due to failing unit-tests.

PR #8970, opened 2017-04-17
[TIMOB-24570] Android: Fix screen density breaking changes + multi-window in Android 7.0

frankieandone:
#### Test Case:
Add this to `tiapp.xml` in your Titanium app directory.
```
<android>
	<manifest xmlns:android="http://schemas.android.com/apk/res/android">
		<uses-sdk android:targetSdkVersion="24"/>
	</manifest>
</android>
```
In `titanium_mobile/build` directory, run: `node scons.js build && node scons.js package && node scons.js install`. In your Titanium app, run and `cat build/android/AndroidManifest.xml`. Watch out for `density` flag in `configChange` attribute. Verify new flags are present. Then put window in split screen mode.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24570


android/cli/commands/_build.js
```diff
@@ -3697,6 +3697,17 @@ AndroidBuilder.prototype.generateAndroidManifest = function generateAndroidManif
 		});
 	}
 
+	if (this.realTargetSDK >= 24) {
+		Object.keys(finalAndroidManifest.application.activity).forEach(function (name) {
+			var activity = finalAndroidManifest.application.activity[name];
+			if (!activity.configChanges) {
+				activity.configChanges = ['density'];
```
hansemannn: Is `density` the only possible value in `configChanges`? Otherwise it would break other configs. And we should warn them to put it there if they don't do so far.

fmerzadyan: No, there other values that go into `configChanges` but it is a requirement for consistency across different Android 7 devices for zooming in/out. For `activity.configChanges = ['density'];` I used the same logic used elsewhere just in case a situation arises where `configChanges` is empty but the fact is `screenSize` will always be added to `configChanges` so it is kind of redundant in a way.

Also, I should make users aware that `density` is added by default?

jquick-axway: The "density" configChange should always be injected into the AndroidManifest.xml for every activity when targeting Android 7.0 or higher.

If we do not add this configChange setting, then a Titanium activity's UI will disappear if the end-user changes the "Display Size" under Android's main "Settings/Display" screen.

The reason this happens is because apps targeting Android 7.0 (or higher) are expected to handle "Display Size" changes dynamically.  Apps targeting an older OS version (ie: API Level) are force-quit and restarted instead when the "Display Size" changes.

jquick-axway:
@fmerzadyan, can you double check if Titanium's JavaScript APIs correctly report the new "Display Size" set under Android "Settings/Display" please?
`
Ti.API.info("@@@ Screen Density: " + Ti.Platform.displayCaps.density);
Ti.API.info("@@@ Screen DPI: " + Ti.Platform.displayCaps.dpi);
Ti.API.info("@@@ Logical Density Factor: " + Ti.Platform.displayCaps.logicalDensityFactor);
`

Setting the "Display Size" to a larger setting will lower the DPI.  For example, "xxxhdpi" to "mdpi" when setting it to the largest setting.  We need to make sure that we're not caching the above properties since they can change dynamically now.

jquick-axway:
@cb1kenobi, just to double check, is this the best way to inject this attribute setting to *all* activities in the AndroidManifest.xml file if the API Level is >= 24?
(The developer should not be able to remove it. It's required.)

android/modules/platform/src/java/ti/modules/titanium/platform/DisplayCapsProxy.java
```diff
@@ -75,10 +77,8 @@ public String getDensity() {
 				return "xxxhigh";
 			case 640:
 				return "xxxhigh";
-			case DisplayMetrics.DENSITY_LOW :
-				return "low";
 			default :
-				return "medium";
+				return "unknown";
```
jquick-axway: I gave you bad advise. It should not return "unknown". Developers use this name to load images of an appropriate size based on the "density" name property.  So, this code needs to be change to use >= checks.  It should also *not* return "medium" for unknown densities either.

The code should be changed to something like this...
`
public String getDensity() {
	synchronized(dm) {
		getDisplay().getMetrics(dm);
		int dpi = dm.densityDpi;
		if (dpi >= 560) {        // DisplayMetrics.DENSITY_560
			return "xxxhigh";
		} else if (dpi >= 400) { // DisplayMetrics.DENSITY_400
			return "xxhigh";
		} else if (dpi >= 280) { // DisplayMetrics.DENSITY_280
			return "xhigh";
		} else if (dpi >= DisplayMetrics.DENSITY_HIGH) {
			return "high";
		} else if (dpi >= DisplayMetrics.DENSITY_TV) {
			return "tvdpi";
		} else if (dpi >= DisplayMetrics.DENSITY_MEDIUM) {
			return "medium";
		}
		return "low";
	}
}
`

android/cli/commands/_build.js
```diff
@@ -3683,16 +3683,17 @@ AndroidBuilder.prototype.generateAndroidManifest = function generateAndroidManif
 		}
 	}, this);
 
-	// if the target sdk is Android 3.2 or newer, then we need to add 'screenSize' to
-	// the default AndroidManifest.xml's 'configChanges' attribute for all <activity>
-	// elements, otherwise changes in orientation will cause the app to restart
-	if (this.realTargetSDK >= 13) {
+	if (this.realTargetSDK >= 24 && !finalAndroidManifest.application.hasOwnProperty('resizeableActivity')) {
+		finalAndroidManifest.application.resizeableActivity = true;
+	}
```
jquick-axway: Let's remove this from "_build.js". We don't want to force developers to use "resizeableActivity" in case they have layout issues when sized too small in a split-screen window... in which case they can opt-out by setting this attribute to false in the "tiapp.xml" file.

jquick-axway (approved):
Approved

lokeshchdhry:
@fmerzadyan , I am seeing an issue where app built with this fix has an invalid `Androidmanifest.xml`.
```
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.app.timob24570" android:versionCode="1" android:versionName="1.0">
	<uses-sdk android:minSdkVersion="14" android:targetSdkVersion="24"/>
	<application android:icon="@drawable/appicon" android:label="TIMOB-24570" android:name="Timob24570Application" android:debuggable="false" android:theme="@style/Theme.AppCompat" android:resizeableActivity="true">
		<activity android:name=".Timob24570Activity" android:label="@string/app_name" android:theme="@style/Theme.Titanium" android:configChanges="keyboardHidden|orientation|fontScale|screenSize|smallestScreenSize|screenLayout|density"/>
		<service android:name="com.appcelerator.aps.APSAnalyticsService" android:exported="false"/>
	</application>
	<uses-permission android:name="android.permission.INTERNET"/>
	<uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
</manifest>
```
This is causing the app to get installed but not launch or show up in the apps list on the device.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.2
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.11
Node Ver:       6.10.1
Java Ver:       1.8.0_101

lokeshchdhry:
FR Passed.

The app switches as expected to split screen to normal mode & vice versa. Also, changing the display size in the settings appropriately reflects the change in the app but, this change is only seen when the app is restarted & not when the app is resumed from paused state or from background.
Confirmed with @jquick-axway about this behavior as we do not have a event like e.g `resize` for the app to change its UI on the fly.

Studio Ver:     4.9.1.201707130742
SDK Ver:        6.2.0.v20170713162837  from PR
OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.2
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.11
Node Ver:       6.10.1
Java Ver:       1.8.0_101
Device:          Pixel Android 7.0
Emulator:      Android 7.0, android 7.1.1

PR #8971, opened 2017-04-18
[TIMOB-24537]TI Gestures not firing even on orientation change on Android

ypbnv:
Trigger 'orientationchange' event from configChanged instead of orientationEvent. That will keep parity between platforms.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24537

**Test Case**

```
var window = Titanium.UI.createWindow();

Ti.Gesture.addEventListener('orientationchange', function(){
	alert('Orientation Changed');
});

window.open();
```

First orientation change does not trigger the listener.

SDK: 6_0_X
Android Emulator: Nexus 5
Android version: 6.0

**Additional info**

I was not able to reproduce the issue on a device.


jquick-axway (commented):
Not approved. (Sorry)

This will not solve the customer's issue.

The issue here is that Titanium's orientation property/event is handled differently between Android and iOS.  On Android, it provides the activity window's orientation (will *not* provide orientations the activity does not support). On iOS, it provides the device's orientation (it *will* return an orientation the app does not support). In fact, we document this here...
   http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Gesture-property-orientation

The reporter wants Android to return the device orientation like iOS, but this would likely be a breaking change.

The thing is, device orientation and app/window orientation are 2 separate concepts.  Both are useful.  Especially for split-screen support on Android 7 and iPad.  The *right* solution may be to deprecate (but not delete) the old orientation property and provide 2 new properties/events providing the device orientation and window orientation.

We should discuss this as a team with Eric.

ypbnv:
I see. I did not know that iOS handled the orientation that way and totally missed that as a disparity.

garymathews:
This needs to be discussed. Closing.

PR #8974, opened 2017-04-18
Migrate titanium-sdk node module to it's own repository

sgtcoolguy:
We currently have one special node module dependency which actually only lives inside the SDK's node_modules folder. This PR "moves out" the dependency to it's own repository at appcelerator/titanium-sdk. It doesn't attempt to grab it from the npm package registry, but simply uses a Git URI to point at master branch of the repo.

This is in preparation for a potential future change to try and use yarn/yarn.lock instead of packaging the full node_modules folder in the SDK repo itself (https://jira.appcelerator.org/browse/TIMOB-24591)


cb1kenobi:
So why don't we just publish {{titanium-sdk}} to NPM? I don't like referencing git repos unless we absolutely need to. I also think we should discuss changing the name to perhaps {{node-tisdk}} or something since it's just a library.

sgtcoolguy:
@cb1kenobi yeah, I agree, I don't like referring git repos and branches either. We should version this and everything. I just wasn't sure if there was any reason we *couldn't* publish it, and I was guessing we wanted to rename it in some way.

cb1kenobi:
After doing some thinking, I feel we should name this `node-tisdk` or `node-titanium-sdk`. This is a Node specific library and thus `node-` prefix seems appropriate. Other than that, I'm fine with moving it out into a separate repo.

sgtcoolguy:
@cb1kenobi I renamed the repo and npm module to be `node-titanium-sdk` and made myself, you and `appcelerator` as owners on the registry. I updated this PR to point to the module on the npm registry rather than the git repo.

cb1kenobi:
@sgtcoolguy Awesome! Thank you for doing this!

I would like to merge this before #8972, however since you modified `node_modules/titanium-sdk/lib/jsanalyze.js` in #8972, you should just update `jsanalyze.js` in `node-titanium-sdk` with the Babel changes. This also means the `babylon`, `babel-core`, `babel-preset-babili`, `babel-traverse`, and `babel-types` as NPM dependencies should go in `node-titanium-sdk`, not in `titanium_mobile`. #8972 effectively becomes a simple update of `node-titanium-sdk` to `v0.2.0`.

sgtcoolguy:
@cb1kenobi Sounds good to me. I've ported the jsanalyze changes to appcelerator/node-titanium-sdk#1

So we can merge this one and then that PR, and then I can update #8972 to simply update the node-titanium-sdk version.

PR #8984, opened 2017-04-21
[TIMOB-24617] Fix SDK build and package of Android on Windows

garymathews:
- Include GnuWin32 binaries `zip unzip gperf` from http://gnuwin32.sourceforge.net/packages.html
- Fix line endings when parsing `libv8.properties`
- Guard against empty path in docgen `mkdirDashP`

###### TEST CASE
```
cd build
node scons.js build android
node scons.js package
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24617)


sgtcoolguy:
Oops, didn't see that you request @ypbnv 's review as well. Oh well, LGTM. Merged to master.

ingo:
I think we should add a note that these are only to be used in packaging, and NOT shipped with the SDK (as they are GPL). This will come up on some security scan at some point.

PR #8986, opened 2017-04-24
[TIMOB-24621] Fixed bug with scons'ing the SDK when timob is extracte…

cb1kenobi:
[TIMOB-24621] Fixed bug with scons'ing the SDK when timob is extracted from a zip instead of git cloned.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24621


sgtcoolguy (commented):
I'd strongly prefer we just detect when there's no .git folder and use a default hash value like 'nogithash' and spit out some warning, rather than totally ignore errors in this callback...

cb1kenobi:
@sgtcoolguy technically the previous code was completely ignoring the error and exiting immediately without message. If it can't find a git hash, it sets it to `n/a` which is basically `nogithash`. I'm all for improving it.

PR #8988, opened 2017-04-25
Update apidoc scripts

sgtcoolguy:
- Move apidoc dependencies to top-level package.json as devDependencies. (So instead of doing `cd apidoc; npm install; node docgen.js` you'd do `npm install; cd apidoc; node docgen.js`)
- Fix up ejs usage to work on ejs 2+.
- Remove old python files for apidoc.
- Update the validate.js to be more strict about required properties as our old validate.py was.
- Validate the docs as part of our build process.
- Fix a doc bug in Ti.UI.Clipboard where 3 properties didn't have a type specified (which the new validate.js found but validate.py did not!)


sgtcoolguy:
cc @jawa9000 I'm not sure if there's other build scripts to update related to this? Basically the big change is that there's no separate package.json inside the apidoc folder with this change, and the `npm install` would run at the root folder of the repo, not inside apidoc.

apidoc/validate.py
```diff
@@ -1,623 +0,0 @@
-#!/usr/bin/env python
```
hansemannn: We should not remove that particular script, yet. It has some syntax-validations that the node-one does not currently have. I usually use both to scan changes, so I could setup a table of missing checks if necessary.

sgtcoolguy: I'd suggest we simply fix the validate.js file to include the additional checks. I don't think it's good for us to have two separate and unequal validation scripts in entirely different language/toolsets.

sgtcoolguy: Do you happen to know off-hand what the additional checks are? Or do we need to compare the two file and try and figure it out from there?

hansemannn: I would just change some docs with typical errors and check both scripts, no big thing to do. I can give you at least a list of differences by tomorrow if it helps. It's only about small things like disallowed the `description` key in function parameters as I remember. The actual docs build script would ignore them anyway.

sgtcoolguy: I'm taking a look now. Looks like a few key differences are that the python script:
- enforced some properties were required on some structures (typically: name, summary)
- enforced a couple properties were valid markdown (not just strings)
- enforce properties whose name were all-caps had permission explicitly set to 'read-only'
- complained the 'notes' fields were '"no longer valid"

Related: should our build process be running the doc validation at some point? Sounds like maybe we should be doing this in an automated way to reject PRs with invalid doc changes...

hansemannn: That would be **awesome**, really.

hansemannn:
Are there more files we clean up in that directory? For example the Readme.md, which includes notes regarding a "tdoc" format, which isn't used since forever.

sgtcoolguy:
I removed the README.md, not sure what else could go. Maybe @jawa9000 has a better idea.

I did see that before (and after) my changes, docgen was broken for `modulehtml` format.

sgtcoolguy:
Since this is a dev/build related change and doesn't affect SDK functionality, I'm gonna merge it.

PR #8994, opened 2017-04-25
[TIMOB-24626] - CLI: Build process errors out with Error: Cannot find module '../node-titanium-sdk/lib/titanium'

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24626

Path didn't traverse far back enough, don't need to traverse the path anyway as the require will find it


cb1kenobi:
LGTM

PR #9002, opened 2017-04-28
Updated Gesture.yml per TIDOC-2848

jawa9000:
https://jira.appcelerator.org/browse/TIDOC-2848

Replaced "screen orientation" with "window orientation" as requested in the Jira ticket.


apidoc/Titanium/Gesture/Gesture.yml
```diff
@@ -56,8 +56,8 @@ properties:
         <Titanium.UI.FACE_DOWN>, 
         or <Titanium.UI.UNKNOWN>. 
         
-        On iOS, returns the current orientation of the *device*, regardless of the screen
-        orientation. To determine the current screen orientation on iOS, use 
+        On iOS, returns the current orientation of the *device*, regardless of the window
+        orientation. To determine the current window orientation on iOS, use 
         <Titanium.UI.orientation>.
```
jquick-axway: This information applies to Windows Phone too. And this paragraph is referencing a deprecated API. How about we change it to the following?

```
On iOS and Windows, returns the current orientation of the *device*, regardless of the window orientation. To determine the current window orientation, use [Window.orientation](Titanium.UI.Window.orientation).
```

jquick-axway (changes requested):

apidoc/Titanium/Gesture/Gesture.yml
hansemannn: @jquick-axway Agree. Updated and merged it!

PR #9007, opened 2017-05-02
[TIMOB-23490][TIMOB-24654] Hybrid CommonJS and Native modules don't export commonjs

sgtcoolguy:
**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-23490
- https://jira.appcelerator.org/browse/TIMOB-24654

**Description:**
- titanium_prep needs JS file listing paths to be relative to assets dir, we sent absolute paths for iOS
- fix merging of native module with encrypted main JS file in iOS require implementation
- titanium_prep argument count changed for android but we never update the module build cli code to send the extra argument
- The android implementation has always relied on a special 'commonjs: true' key/value pair in the manifest file of the packaged android module to hook up the combination of commonjs and native. When we moved from python to node CLI we never added the code to insert this value to the manifest when there are js file inside the assets folder.


sgtcoolguy:
Again, note that this fixes iOS, but *does not fix Android*. We may want to break off another ticket to address Android entirely. I pushed a fix to encrypt the js file into the module, but the merging go the native and commonjs code isn't working/hooked up properly.

iphone/Classes/KrollBridge.m
```diff
@@ -818,7 +818,8 @@ - (TiModule *)loadTopLevelNativeModule:(TiModule *)module withPath:(NSString *)p
 	}
 
 	NSString* contents = [[[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding] autorelease];
-	KrollWrapper* wrapper = (id) [self loadJavascriptText:contents fromFile:path withContext:kroll];
+	NSURL *url_ = [TiHost resourceBasedURL:path baseURL:NULL];
+	KrollWrapper *wrapper = (id) [self loadCommonJSModule:contents withSourceURL:url_];
```
hansemannn: `loadJavascriptText` does some things that `loadCommonJSModule` doesn't, like checking [previously cached modules](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/KrollBridge.m#L970) and [adding them to the cache](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/KrollBridge.m#L1001).

hansemannn: `loadJavascriptText` does some things that `loadCommonJSModule` doesn't, like checking [previously cached modules](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/KrollBridge.m#L970) and [adding them to the cache](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/KrollBridge.m#L1001). I am curious that this could lead to other issues regarding module caching.

sgtcoolguy:
QE - test steps are written up here: https://jira.appcelerator.org/browse/TIMOB-23490?focusedCommentId=419052&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-419052

iphone/Classes/KrollBridge.m
sgtcoolguy: It shouldn't have an impact. Specifically this is a case where we are loading javascript we received as a string and turning it into the exported object, then combining that with the previously loaded 'native' module. The native module with the layered commonjs exports is what should get cached later.

hansemannn: Okay, thanks. I'll do a FT tomorrow morning using multiple require-statements to compare the caching.

sgtcoolguy:
Update the PR to include the Android fix as well. Same steps as iOS, just target android.

mukherjee2 (approved):
Node Version: 6.10.1
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.0 
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.2.0 locally built on branch PR/9007
Xcode 8.3.2

I ran the demo code and saw "hi" in the console rather than "null". This is the expected behvior.
EDIT: The FR was previously done on iOS only. I just tried it with Android, and FR passes on Android as well.

hansemannn (approved):

hansemannn:
CR + FT **passed**, PR therefore **approved**!

mukherjee2:
@sgtcoolguy , I tried to "Update branch", but build fails with this message in Jenkins.

Fatal error: Unable to find local grunt.
If you're seeing this message, grunt hasn't been installed locally to
your project. For more information about installing and configuring grunt,

hansemannn:
@sgtcoolguy 6.1.0 Backport?

mukherjee2:
@hansemannn , I am unable to merge because Update branch failed. I also need this one backported to 6_1_X

PR #9019, opened 2017-05-05
[TIMOB-24659] Android: Add support for the roundIcon attribute

gaspardle:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24659

**Optional Description:**
Android 7.1 allow applications to show a circular icon in the launcher, for devices that supports them (ex. Pixel & Pixel XL).
Titanium should support the android:roundIcon attribute in AndroidManifest.xml.


garymathews (approved):
Approved

PR #9030, opened 2017-05-08
[TIMOB-24642] - Correct require implementation

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24642

*Note* iOS does not like a leading slash on paths, so '/node_modules' in the below is 'node_modules' on iOS

- Decrement i before continuing
- If startDir/path === '/' return early with "/node_modules" to prevent unnecessary loop
- Always appends '/node_modules'
- On iOS correct the path builder to split the array rather than perform a substring on the string

Test case:

1. Download the test app linked in the ticket description
2. Build the app, it should load to a yellow screen and in the console you should see.

```
[INFO]  boz index.js loaded
[INFO]  boo boo.js loaded
[INFO]  baz index.js loaded
[INFO]  bar bar.js loaded
[INFO]  foo lib/foo.js loaded
[INFO]  foo index.js loaded
```


cb1kenobi:
When a path begins with `/`, it implies a resource, not module. `require('/path/to/file.js')` is a resource. `require('path/to/file')` is a module. Module identifiers do not have an extension. Resources require an extension. The initial `require()` implementation handled `/` wrong. If you `require('/foo')`, it should NEVER search `node_modules`. Not even Node does that.

android/runtime/common/src/js/module.js
```diff
@@ -385,6 +385,16 @@ Module.prototype.loadNodeModules = function (moduleId, startDir, context) {
  */
 Module.prototype.nodeModulesPaths = function (startDir) {
 	// 1. let PARTS = path split(START)
```
sgtcoolguy: move this comment back down to the line it's referencing below.

sgtcoolguy:
@ewanharris It'd be useful to add some tests around this in our test suite so we know the issue is fixed, and to verify some of the things we're discussing here.

Copy https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.require.test.js in `titanium_mobile/tests/Resources` and add some unit tests that verify we don't have an infinite loop. You may need to also override `titanium_mobile/tests/Resources/node_modules` to include some additional node modules that trigger the bug (right now there's just the `abbrev` module there: https://github.com/appcelerator/titanium-mobile-mocha-suite/tree/master/Resources/node_modules/abbrev).

As for what @cb1kenobi is talking about, I don't believe we do search for node_modules when the require path is prefixed by `./`, `../` or `/`. This fix is related to gathering the list of possible paths to search for node_modules folders up the directory tree from our current location (in the cases where we should be looking for them).

sgtcoolguy (changes requested):
Requesting unit tests be added to verify the bug is fixed.

ewanharris:
@sgtcoolguy Addressed your comments, added the following tests

1. [this ticket](https://github.com/appcelerator/titanium_mobile/pull/9030/files#diff-686d43a4f6290f22ae8b8cbd8e7854b4R73)
2. [validating the caching of modules and module lookup works as expected i.e closest node_modules first](https://github.com/appcelerator/titanium_mobile/pull/9030/files#diff-686d43a4f6290f22ae8b8cbd8e7854b4R83)
3. [Checking what @cb1kenobi had mentioned](https://github.com/appcelerator/titanium_mobile/pull/9030/files#diff-686d43a4f6290f22ae8b8cbd8e7854b4R101), which exposed a bug on Android, see [TIMOB-24673](https://jira.appcelerator.org/browse/TIMOB-24673)
4. [Unskipped a test that works on android now ](https://github.com/appcelerator/titanium_mobile/pull/9030/files#diff-686d43a4f6290f22ae8b8cbd8e7854b4R66)

mukherjee2 (approved):
Passed FR!
Node Version: 6.10.1
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.0
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.2.0.v20170510080439
Xcode 8.3.2

First tested with 6.0.4.GA. The app hangs. I retried with the fix in this PR, and the app launched and displayed a yellow window. FR passed.

PR #9045, opened 2017-05-10
[TIMOB-24642] (6_1_X) - Correct require implementation

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24642

*Note* iOS does not like a leading slash on paths, so '/node_modules' in the below is 'node_modules' on iOS

- Decrement i before continuing
- If startDir/path === '/' return early with "/node_modules" to prevent unnecessary loop
- Always appends '/node_modules'
- On iOS correct the path builder to split the array rather than perform a substring on the string

Test case:

1. Download the test app linked in the ticket description
2. Build the app, it should load to a yellow screen and in the console you should see.

```
[INFO]  boz index.js loaded
[INFO]  boo boo.js loaded
[INFO]  baz index.js loaded
[INFO]  bar bar.js loaded
[INFO]  foo lib/foo.js loaded
[INFO]  foo index.js loaded
```


mukherjee2 (approved):
Passed FR!
Node Version: 6.10.1
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.0
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.1.0.v20170510103244
Xcode 8.3.2

First tested with 6.0.4.GA. The app hangs. I retried with the fix in this PR, and the app launched and displayed a yellow window. FR passed.

PR #9054, opened 2017-05-11
Tests merge

sgtcoolguy:
**JIRA:** No ticket - development clean up

**Description:**
Remove the unused and confusing ti_mocha_tests folder that was originally to be used for unit tests on Travis but never got used.

Migrate some of the tests over to the `tests` folder to override/include in our mocha test suite that *is* used.


mukherjee2:
@sgtcoolguy @hansemannn Do I need to test anything before merging this PR. I have already done the FR on master and 6.1.0 on the other two PR's. It does not appear to be related to the fix in the ticket. However, I don't want to merge unless I know whether something needs to be tested.

PR #9056, opened 2017-05-11
[TIMOB-24680] (6_1_X) Fixed list of valid provisioning profiles for the given…

cb1kenobi:
[TIMOB-24680] Fixed list of valid provisioning profiles for the given build target which is used to validate watch extension provisioning profile uuids.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24680


iphone/cli/commands/_build.js
```diff
@@ -1434,6 +1434,26 @@ iOSBuilder.prototype.initTiappSettings = function initTiappSettings() {
...
+
+		function getPPbyUUID(ppuuid) {
+			return pps
+				.filter(function (p) {
+					if (!p.expired && !p.managed && p.uuid === ppuuid) {
+						return true;
+					}
+				})
+				.shift();
+		}
```
hansemannn: We use that method too often. Please add this helper to our utils instead of pasting it to every usage.

hansemannn (changes requested):

hansemannn (approved):

iphone/cli/commands/_build.js
cb1kenobi: Great idea for the next overhaul.

htbryant94 (approved):
Approved! Enterprise Adhoc Provisioning Profiles are now validated correctly for Watch Extensions, and build failures no longer occur.

hansemannn:
@htbryant94 Wait a second, should this go into 6.1.1? Because you just merged into 6_1_X, not master. The ticket says 6.2.0, however, I'd be fine for a 6.1.1 if @eric34 approves.

htbryant94:
@hansemannn I asked @eric34 whether to merge the fix into the 6_1_X branch as well as the master, and was confirmed to do so. It would be good to get a second approval just in case, however I do not see any reason why it shouldn't go into 6.1.1

PR #9060, opened 2017-05-12
[TIMOB-24691] (6_1_X) Inject platform back into args just in case the set pla…

cb1kenobi:
[TIMOB-24691] Inject platform back into args just in case the set platform via -p so that the forked build command has the correct platform in the args.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24691


feons:
*LGTM.*

PR #9063, opened 2017-05-16
[TIMOB-24608] Android: respackage value in native module's manifest is ignored

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24608

**Description:**
We never consulted the module's `manifest` file's `respackage` value (in our new node-based module build). We need to look for a non-empty value and if it exists generate a `respackageinfo` file inside the module zip with the value. This is used to generated the R* java/class files for resources on Android.


sgtcoolguy:
**For QE/FR**
Use a native android module like ti.facebook that contains resources (and not just a jar file). The manifest should have a `repackage` value.
Run `appc run -p android --build-only`
The resulting zip file in `android/dist` should have a `respackageinfo` file in the root of it, with the  manifest's `repackage` value.

android/cli/commands/_buildModule.js
```diff
@@ -1474,6 +1474,11 @@ AndroidModuleBuilder.prototype.packageZip = function (next) {
 					}.bind(this));
 				}
 
+				// respackageinfo file
+				if (this.manifest.respackage) {
```
cb1kenobi: Why add a `respackage` key to the manifest instead of just checking if the `respackageinfo` file exists?

sgtcoolguy: I'm not sure I understand. The `respackage` value is not a boolean key, it's an optional manifest value that typically is a java package name - and it's set manually by the user. We look to see if it's set to anything and if so, generate a `respackageinfo` file in the module holding the value.

So we either need to copy it from he manifest to the file expected, or change our app build process that handles modules to not look for this file but instead consult the manifest.

cb1kenobi: Nevermind, I see what it's doing.

cb1kenobi (approved):

lokeshchdhry:
FR Passed.

1. Downloaded repo for facebook module from master branch.
2. Selected the SDK with this fix in the CLI.
3. Ran `appc run -p android --build-only` from the android folder.
4. The resulting zip file in the `dist` folder has the correct `respackageinfo` file at the root with the manifest's `repackage` value.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.2
Appc NPM:       4.2.9
Appc CLI:       6.2.2
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.11
Node Ver:       6.10.1
Java Ver:       1.8.0_101

PR #9066, opened 2017-05-17
[TIMOB-16622] iOS:Replace UISearchDisplayController with UISearchController

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-16622


iphone/Classes/TiUIListView.m
```diff
@@ -39,9 +39,8 @@ @implementation TiUIListView {
     TiUIRefreshControlProxy* _refreshControlProxy;
 #endif
 
-    TiUISearchBarProxy *searchViewProxy;
-    UITableViewController *tableController;
```
hansemannn: This should not be removed.

iphone/Classes/TiUIListView.m
```diff
@@ -528,9 +526,6 @@ - (void)updateSearchResults:(id)unused
         [self buildResultsForSearchText];
     }
     [_tableView reloadData];
-    if ([searchController isActive]) {
-        [[searchController searchResultsTableView] reloadData];
-    }
```
hansemannn: Why did this move away?

iphone/Classes/TiUIListView.m
```diff
@@ -832,11 +827,7 @@ -(void)setCaseInsensitiveSearch_:(id)args
     caseInsensitiveSearch = [TiUtils boolValue:args def:YES];
     if (searchActive) {
         [self buildResultsForSearchText];
-        if ([searchController isActive]) {
-            [[searchController searchResultsTableView] reloadData];
-        } else {
             [_tableView reloadData];
-        }
```
hansemannn: Why did this move away?

iphone/Classes/TiUIListView.m
```diff
@@ -871,6 +861,7 @@ -(void)setSearchView_:(id)args
             [[searchViewProxy searchBar] setPlaceholder:@"Search"];
         }
         self.searchString = [[searchViewProxy searchBar] text];
+        [self initSearchController:self];
```
hansemannn: What happens if the search-view is set multiple times?

iphone/Classes/TiUIListView.m
```diff
@@ -2004,7 +1993,7 @@ - (void)searchBarTextDidBeginEditing:(UISearchBar *)searchBar
     }
     self.searchString = (searchBar.text == nil) ? @"" : searchBar.text;
     [self buildResultsForSearchText];
-    [[searchController searchResultsTableView] reloadData];
+    [_tableView reloadData];
```
hansemannn: `_tableView` is the main table view, but the deleted line reloaded the search-results table view.

iphone/Classes/TiUIListView.m
```diff
@@ -2013,7 +2002,7 @@ - (void)searchBarTextDidEndEditing:(UISearchBar *)searchBar
         self.searchString = @"";
         [self buildResultsForSearchText];
         if ([searchController isActive]) {
-            [searchController setActive:NO animated:YES];
+            [searchController setActive:false];
```
hansemannn: We are in Obj-C, so use `NO`. Also, is there no `animated` parameter?

iphone/Classes/TiUIListView.m
```diff
@@ -2041,74 +2030,52 @@ - (void)searchBarCancelButtonClicked:(UISearchBar *) searchBar
...
+#ifndef TI_USE_AUTOLAYOUT
+            [_searchWrapper layoutProperties]->right = TiDimensionDip(right);
+#endif
+             [_searchWrapper refreshView:nil];
+        }
+    }
+    //IOS7 DP3. TableView seems to be adding the searchView to
+    //tableView. Bug on IOS7?
+    [self clearSearchController:self];
+    [_tableView reloadData];
```
hansemannn: I guess this was copy-pasted over? So you have still been able to reproduce it without this hack?

iphone/Classes/TiUIListView.m
```diff
@@ -2041,74 +2030,52 @@ - (void)searchBarCancelButtonClicked:(UISearchBar *) searchBar
...
+}
 
-- (void)searchDisplayControllerWillBeginSearch:(UISearchDisplayController *)controller
+- (void)presentSearchController:(UISearchController *)controller
 {
+    id dimBgValue = [[self  proxy] valueForKey:@"dimsBackground"];
+    searchController.dimsBackgroundDuringPresentation = [TiUtils boolValue:dimBgValue def:NO];
```
hansemannn: Where does this come from? It's not documented, but may be cool to be exposed & documented.

iphone/Classes/TiUIListView.m
```diff
@@ -2041,74 +2030,52 @@ - (void)searchBarCancelButtonClicked:(UISearchBar *) searchBar
...
-    
-    if (resultsSeparatorInsets) {
-        [controller.searchResultsTableView setSeparatorInset:[TiUtils contentInsets:resultsSeparatorInsets]];
-    }
-    
-    if (resultsSeparatorStyle) {
-        [controller.searchResultsTableView setSeparatorStyle:[TiUtils intValue:resultsSeparatorStyle def:UITableViewCellSeparatorStyleSingleLine]];
-    }
-    
-    return YES;
```
hansemannn: What happened to this whole block? It was introduced as part of [TIMOB-23281](https://jira.appcelerator.org/browse/TIMOB-23281) and has to stay.

iphone/Classes/TiUIListView.m
```diff
@@ -2225,22 +2192,22 @@ -(CGFloat)contentHeightForWidth:(CGFloat)width
...
 {
-    if (sender == self && tableController == nil) {
-        tableController = [[UITableViewController alloc] init];
-        [TiUtils configureController:tableController withObject:nil];
-        tableController.tableView = [self tableView];
-        searchController = [[UISearchDisplayController alloc] initWithSearchBar:[searchViewProxy searchBar] contentsController:tableController];
-        searchController.searchResultsDataSource = self;
-        searchController.searchResultsDelegate = self;
+    if (sender == self && searchController == nil)
+    {
```
hansemannn: We put brackets in the same line

iphone/Classes/TiUIListView.m
```diff
@@ -2225,22 +2192,22 @@ -(CGFloat)contentHeightForWidth:(CGFloat)width
...
+        searchController = [[UISearchController alloc] initWithSearchResultsController:nil];
```
hansemannn: Why are you initializing it with `nil`? I guess we should rather manage an own table-view for it, maybe even in an own helper-class.

iphone/Classes/TiUIListView.m
```diff
@@ -2225,22 +2192,22 @@ -(CGFloat)contentHeightForWidth:(CGFloat)width
...
         searchController.delegate = self;
+        searchController.searchResultsUpdater = self;
+        searchController.hidesNavigationBarDuringPresentation = NO;
+        searchController.searchBar.frame = CGRectMake(searchController.searchBar.frame.origin.x, searchController.searchBar.frame.origin.y, searchController.searchBar.frame.size.width, 44.0);
```
hansemannn: I'm sceptical about giving it a fixed size. Have you tried this on landscape / multiple orientation changes?

iphone/Classes/TiUIListView.m
```diff
@@ -2225,22 +2192,22 @@ -(CGFloat)contentHeightForWidth:(CGFloat)width
...
-        tableController.tableView = [self tableView];
-        searchController = [[UISearchDisplayController alloc] initWithSearchBar:[searchViewProxy searchBar] contentsController:tableController];
-        searchController.searchResultsDataSource = self;
-        searchController.searchResultsDelegate = self;
+    if (sender == self && searchController == nil)
+    {
+        searchController = [[UISearchController alloc] initWithSearchResultsController:nil];
         searchController.delegate = self;
+        searchController.searchResultsUpdater = self;
+        searchController.hidesNavigationBarDuringPresentation = NO;
```
hansemannn: We should stick to the default values, is setting this value necessary?

iphone/Classes/TiUISearchBar.m
```diff
@@ -54,6 +54,19 @@ -(UISearchBar*)searchBar
...
+    if (searchView)
+    {
+        [searchView removeFromSuperview];
+        [searchView setDelegate:nil];
+        RELEASE_TO_NIL(searchView);
+    }
+    searchView = searchBar;
+    [searchView setDelegate:self];
+    [self addSubview:searchView];
+}
```
hansemannn: Not sure what's this used for. It's used in https://github.com/appcelerator/titanium_mobile/pull/9066/files#diff-c25f31a8654c7459f7b97be688ab44efR2210 but using a setter seems to be the wrong way. I'm curious if we should o that way, especially because it will add a new sub-view every time it's called. So if the user changes the search-bar over time (using `searchView` / `searchBar`, multiple search-bars will just be added above each other.

Also brackets miss-formatted again, that needs to be checked in general.

hansemannn (changes requested):
Major changes required, please address those before going into the next review phase.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -1058,40 +1058,6 @@ properties:
...
-    platforms: [iphone, ipad]
-    availability: creation
-
-  - name: resultsSeparatorStyle
-    summary: The separator style of the search results (iOS-only).
-    type: Number
-    constants: Titanium.UI.iPhone.ListViewSeparatorStyle.*
-    since: 6.1.0
-    platforms: [iphone, ipad]
-    availability: creation
```
hansemannn: Removing them is not an option. Can't we just grab the values and style the search-tableview accordingly?

iphone/Classes/TiUIListView.m
```diff
@@ -2046,17 +2047,12 @@ - (void)willDismissSearchController:(UISearchController *)searchController
...
-    //tableView. Bug on IOS7?
-    [self clearSearchController:self];
+    [searchViewProxy ensureSearchBarHeirarchy];
     [_tableView reloadData];
 }
 
 - (void)presentSearchController:(UISearchController *)controller
 {
-    id dimBgValue = [[self  proxy] valueForKey:@"dimsBackground"];
-    searchController.dimsBackgroundDuringPresentation = [TiUtils boolValue:dimBgValue def:NO];
```
hansemannn: Rather document it and set the value in an appropriate place (like init or here if that's more flexible).

hansemannn (changes requested):

iphone/Classes/TiUIListView.m
vijaysingh-axway: In UISearchController there is option to pass nil for result controller. In this case it will use the current tableview for showing the search result. We can not use different controller to show the result because it will create similar problem which is mentioned in TIMOB-16622 and TIMOB-16826. So there is no need of tableViewController.
searchViewProxy is there.

iphone/Classes/TiUIListView.m
vijaysingh-axway: SearchViewController is using current tableView. And it do not have searchResultsTableView.

iphone/Classes/TiUIListView.m
vijaysingh-axway: Not a problem . Every time searchViewController is released and new searchResultController is created.

iphone/Classes/TiUIListView.m
vijaysingh-axway: In this case both tableView is same.

iphone/Classes/TiUIListView.m
vijaysingh-axway: Done. animated is not available with UISearchViewController.

iphone/Classes/TiUIListView.m
vijaysingh-axway: Removed. Only ensureSearchBarHeirarchy called to ensure the hierarchy.

iphone/Classes/TiUIListView.m
vijaysingh-axway: This property is of no use, until searchResultController is passed in UISearchViewController. So removed.

iphone/Classes/TiUIListView.m
vijaysingh-axway: As current table view and result table view is same. So no use of this , as the problem mentioned in ticket will not arise.

iphone/Classes/TiUIListView.m
vijaysingh-axway: Done

iphone/Classes/TiUIListView.m
vijaysingh-axway: In UISearchController there is option to pass nil for result controller. In this case it will use the current tableview for showing the search result. We can not use different controller to show the result because it will create similar problem which is mentioned in TIMOB-16622 and TIMOB-16826.

iphone/Classes/TiUIListView.m
vijaysingh-axway: Default value will hide navigation bar and is misaligning the tableview.

iphone/Classes/TiUIListView.m
vijaysingh-axway: I tried but  don't found. If you face please let me know.

iphone/Classes/TiUISearchBar.m
vijaysingh-axway: First it is removing the search bar then adding it as a subview. So not a problem.

apidoc/Titanium/UI/ListView.yml
vijaysingh-axway: Is it really necessary? Problem mentioned in ticket will not arise.

iphone/Classes/TiUIListView.m
vijaysingh-axway: This property is of no use, until searchResultController is passed in UISearchViewController. So removed it and put a default value while initializing searchViewController.

apidoc/Titanium/UI/ListView.yml
hansemannn: Agree, but we need to state that, by using:

```yml
removed: 
    since: "6.2.0"
    notes: |
        The styling of the search table-view is determined by it's table-view, so there
        is no additional styling required.
```

apidoc/Titanium/UI/ListView.yml
```diff
@@ -1058,6 +1058,52 @@ properties:
...
+    type: String
+    default: undefined (behaves as white) 
+    since: 6.1.0
+    platforms: [iphone, ipad]
+    availability: creation
+    removed: 
+    since: "6.2.0"
+    notes: |
+        The background color of the search table-view is determined by it's table-view, so there is no additional property required.
+
```
hansemannn: This is not the way the docs are written. Please check how other docs handle removal, also remember to break long lines into separate ones.

hansemannn (changes requested):

hansemannn (changes requested):
There is still one (commented in) reference to the `UISearchDisplayController` in the `searchDisplayControllerDidEndSearch:` delegate, please remove that one as well, although it will not throw a warning anyway.

hansemannn:
In my FT right now. I noticed that the old behavior dimmed the background by default, now it doesn't (setting `dimsBackgroundDuringPresentation` to `NO`, which is against the default behavior). Was there a special reason you did it that way? Maybe I am overseeing something, thanks!

vijaysingh-axway:
In our current implementation of UISearchController, we are passing nil for result controller (reason explained in my first comment above). So it is using current table view for showing result. If we use default behaviour of property dimsBackgroundDuringPresentation, we can not be able to select any row from result tableView, as this will be disabled. So dimsBackgroundDuringPresentation is set to NO.

hansemannn:
Well, I tried it with `dimsBackgroundDuringPresentation` and was able to select rows. However, show the edit actions did work with both cases.

vijaysingh-axway:
Are you saying by setting " dimsBackgroundDuringPresentation  = YES" inside TIUIlistView.m, you are able to select row while searching? This property is not exposed to be use with js.

hansemannn:
Yep, inside the Obj-C code :-)

vijaysingh-axway:
@hansemannn As we discussed, I have implemented this using new UITableViewController and a table view in order to show dim background while searching.

vijaysingh-axway:
@hansemannn  Ping!

hansemannn (approved):

vijaysingh-axway:
@hansemannn Can you please review a minor fix in this PR , which I found while working for edit action for search result. Thanks!

iphone/Classes/TiUIListView.m
```diff
@@ -578,10 +578,10 @@ - (void)updateSearchResults:(id)unused
 {
     if (searchActive) {
         [self buildResultsForSearchText];
-        [resultViewController.tableView reloadData];
     }
-    else {
-        [_tableView reloadData];
+    [_tableView reloadData];
+    if ([searchController isActive]) {
+        [resultViewController.tableView reloadData];
```
hansemannn: Indentation. Also, when would `searchActive` be `NO` but `[searchController isActive]` be `YES`?

iphone/Classes/TiUIListView.m
```diff
@@ -898,7 +898,11 @@ -(void)setCaseInsensitiveSearch_:(id)args
     caseInsensitiveSearch = [TiUtils boolValue:args def:YES];
     if (searchActive) {
         [self buildResultsForSearchText];
-        [_tableView reloadData];
+        if ([searchController isActive]) {
+            [[resultViewController tableView] reloadData];
+        } else {
+            [_tableView reloadData];
+        }
```
hansemannn: Same question as above.

iphone/Classes/TiUIListView.m
vijaysingh-axway: there are 2 ways we can use search in list view -
1. add searcbar on window, and show search result in same table view (see last test case in ticket)
2.  make search bar part of list view  (see test case in ticket except last)
In second way, when we click on searchbar [searchController isActive] will be yes.
But for searchActive being to Yes, there should be atleast one character in search bar.
So when second way of search used, user clicked on search bar and not typed any character, [searchController isActive] will be YES and searchActive will be NO.
If you see the old implementation, it is in same way.

Are you sure there is problem in indentation, I don't think so.

ewieberappc (approved):
FR Passed, using:
MacOS 10.12.6 (16G24b)
Studio 4.9.0.201705302345
Ti SDK 7.0.0.v20170802103048
Appc NPM 4.2.9
Appc CLI 6.2.3
Alloy 1.9.13
Xcode 8.3.3 (8E3004b)

Tested searchbar functionality when used with listview and tableview. No issues encountered. Tested using the provided sample code as well as the searchbar suite

PR #9071, opened 2017-05-19
[TIMOB-20037] Android: Wrong Ti.Android.currentActivity in `open` handler

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20037

**Test case:**
```
var window = Ti.UI.createWindow({backgroundColor:'#F00'});

window.addEventListener('open', function (e) {
    Ti.API.debug('Main window opened', e.source.getActivity() === Ti.Android.currentActivity);
});

window.open();
```

**Optional Description:**
Added Kroll property getter in the module.

@jquick-axway I have missed or messed up something here. Building the SDK with the changes failed after a second attempt. I will update you once I find out what exactly happened.

Update: It was a problem with my setup. The code should work fine.


android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -57,9 +18,32 @@
 import android.support.multidex.MultiDex;
 import android.util.DisplayMetrics;
 import android.view.accessibility.AccessibilityManager;
-
 import com.appcelerator.aps.APSAnalytics;
 import com.appcelerator.aps.APSAnalytics.DeployType;
+import org.appcelerator.kroll.*;
```
garymathews: Should revert these imports back to not use wildcards

garymathews (changes requested):
CR: Fix imports

android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
```diff
@@ -325,6 +326,11 @@ public RProxy getR() {
 		return r;
 	}
 
+	@Kroll.method @Kroll.getProperty
+	public ActivityProxy getCurrentActivity() {
+		return TiApplication.getAppCurrentActivity().getActivityProxy();
```
garymathews: It is possible for `getAppCurrentActivity()` to return `null`, best add a guard.

ypbnv:
@garymathews Reverted the imports and added a guard and a warning.

garymathews (approved):
CR: **PASS**

ssjsamir (approved):
FR passed: currentActivity is now the same as the current window activity. This can be seen by the following output in the console: 
```
[DEBUG] :  Window: Checkpoint: postWindowCreated()
[DEBUG] :  Main window opened true
```
**Test Steps**
+ Downloaded the SDK Build form this PR
+ Copied the the test case from description
+ Ran the application 
+ Saw the above output in the console

**Test Environment**
Google Nexus 6P (7.1.1)
Node.js Version: 6.10.1
Xcode: 8.2

PR #9082, opened 2017-05-24
[TIMOB-24723] Android: Support V8 inspector API for debugging to replace "legacy" debugger protocol

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24723

**Description:**
This is to move from the now-deprecated (and in current V8: removed) "legacy" debugger API to the v8-inspector API.

This converts our JSDebugger shim from opening a simple server socket with very basic http header handling and byte payloads; to spinning up a web socket server (thanks to https://github.com/TooTallNate/Java-WebSocket)

**TODO**
- write up docs/notes on using with Chrome Dev Tools to debug


sgtcoolguy:
Ready for review.

To test:
```
npm install --only=dev
cd build
node scons clean build android
cd $HOME
ti create -t app -p android -n inspector --id com.axway.inspector.test -u http://www.axway.com -d .
cd inspector
ti build -p android --debug-host /127.0.0.1:51388
```
Once the app launches, it should stay on the "splash screen" and spit out some logs that it's waiting 60 seconds for the debugger, as well as details on how to connect to it. Specifically it should spit out a URL you can open in Google Chrome browser to debug the app. The debugger should "pause" on the very first line of app.js. You may notice it injected `(function () {` before the file's normal contents. (this is done to be able to "compile" the contents and schedule the debugger pause to happen as soon as we enter/execute them. We wrap all JS files similarly normally, but this wrapping of app.js is only done when debugging out of necessity).

For testing I find it useful to edit the app.js to add a click listener to the labels and have them open an alert, then set a breakpoint there.

You should be able to hit any breakpoint in app.js, even on the first line. (i.e. we don't let the app run until the debugger has had a chance to connect and set up).

You may find the some features offered by Chrome Devtools may not work or may crash the app (i.e. heap profiling). It'd be good to record those in our documentation and open tickets to see if we can eventually support that.

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
```diff
@@ -20,6 +20,20 @@
 import java.util.Scanner;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.net.InetSocketAddress;
```
garymathews: I think these can be cleaned to:

```Java
import java.io.IOException;
import java.net.InetSocketAddress;
import java.util.concurrent.atomic.AtomicBoolean;
import java.util.concurrent.LinkedBlockingQueue;
import java.net.UnknownHostException;
import java.nio.ByteBuffer;
import java.util.UUID;

import org.java_websocket.WebSocket;
import org.java_websocket.framing.Framedata;
import org.java_websocket.handshake.ClientHandshake;
import org.java_websocket.server.WebSocketServer;

import org.appcelerator.kroll.common.Log;
import org.appcelerator.kroll.common.TiMessenger;
```

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
```diff
@@ -76,323 +94,188 @@ public JSDebugger(int port, String sdkVersion) {
...
-			JSDebugger.this.v8Messages.add(STOP_MESSAGE);
+			JSDebugger.this.v8Messages.offer(STOP_MESSAGE);
 		}
 
 		@Override
 		public void run()
 		{
-			this.sendHandshake();
 			while (!stop.get())
 			{
```
garymathews: Lets fix this formatting to be consistent with the rest of the file?

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
```diff
@@ -76,323 +94,188 @@ public JSDebugger(int port, String sdkVersion) {
...
-							if (socket != null) {
-								socket.close();
-							}
-						} catch (Throwable t) {
-							// ignore
-						}
-
-						// Wipe the messages from V8
-						JSDebugger.this.clearMessages();
+				handler = new V8MessageHandler(conn);
```
garymathews: Maybe add debug log to show a client has connected?

android/build/libv8.properties
```diff
@@ -1,2 +1,2 @@
-libv8.version=5.7.492.71
+libv8.version=5.9.211.30
```
garymathews: This version prevents kroll-v8 from compiling, revert back to `5.8.283.32`

android/runtime/v8/src/native/InspectorFrontend.cpp
```diff
@@ -0,0 +1,44 @@
...
+  InspectorFrontend::InspectorFrontend(v8::Local<v8::Context> context) {
+    isolate_ = context->GetIsolate();
+  }
+
+  void InspectorFrontend::Send(const v8_inspector::StringView& string) {
+    v8::HandleScope scope(isolate_);
+
+    int length = static_cast<int>(string.length());
+    assert(length < v8::String::kMaxLength);
+    {
```
garymathews: this block scope doesn't seem necessary?

android/runtime/v8/src/native/JSDebugger.cpp
```diff
@@ -20,80 +22,119 @@ JSDebugger::JSDebugger()
...
 	assert(debuggerClass__ != nullptr);
 
 	handleMessage__ = env->GetMethodID(debuggerClass__, "handleMessage", "(Ljava/lang/String;)V");
 	assert(handleMessage__ != nullptr);
+
+	waitForMessage__ = env->GetMethodID(debuggerClass__, "waitForMessage", "()Ljava/lang/String;");
+	assert(waitForMessage__ != nullptr);
+
+	if (debugger__ != nullptr)
+	{
```
garymathews: fix inconsistent formatting in this file too

android/runtime/v8/src/ndk-modules/libv8/Android.mk
```diff
@@ -9,15 +9,15 @@ LOCAL_PATH := $(call my-dir)
 
 include $(CLEAR_VARS)
 
-V8_VERSION=5.7.492.71
-LIBV8_DIR := ../../../../../../dist/android/libv8/$(V8_VERSION)
+# Read android/build/libv8.properties to get v8 version and mode (release or debug)
+include $(LOCAL_PATH)/../../../../../build/libv8.properties
+V8_VERSION=$(libv8.version)
+LIBV8_MODE := $(libv8.mode)
+LIBV8_DIR := ../../../../../../dist/android/libv8/$(V8_VERSION)/$(LIBV8_MODE)
```
garymathews:
```
LIBV8_DIR := $(TI_DIST_DIR)/android/libv8/$(V8_VERSION)/$(LIBV8_MODE)
```

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -59,7 +59,7 @@
 	"borderColor", "borderRadius", "borderWidth",
 
 	// layout / dimension (size/width/height have custom accessors)
-	"left", "top", "right", "bottom", "layout", "zIndex", TiC.PROPERTY_CENTER,
+	"left", "top", "right", "bottom", "layout", "zIndex",
```
garymathews: I don't think this change is meant to be here?

garymathews (dismissed):
FT: **PASS**

Debugging with inspector works great! 🎆 

- Should it also handle re-connecting? (refresh inspector whilst debugging should continue where it left?)
- minor but, `debuggerEnabled: true/false` in `deploy.json` has no effect
- can we remove/hide the IIFE from the `app.js`?

android/runtime/v8/src/native/InspectorClient.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+  void InspectorClient::Initialize(v8::Local<v8::Object> target, v8::Local<v8::Context> context) {
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::HandleScope scope(isolate);
+
+    DEFINE_METHOD(isolate, target, "callAndPauseOnStart", CallAndPauseOnStart);
+  }
+
+  void InspectorClient::CallAndPauseOnStart(const v8::FunctionCallbackInfo<v8::Value>& args) {
+    v8::Isolate* isolate = args.GetIsolate();
```
garymathews: Why not compile and run the source here so we don't need the IIFE?
```JS
...
return inspectorWrapper(source, filename);
```
```C++
Local<String> source = args[0].As<String>();
Local<String> filename = args[1].As<String>();
Local<Script> script = Script::Compile(source, filename);

JSDebugger::debugBreak();
Local<Value> result = script->Run();
```

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
sgtcoolguy: It should probably be pulled out to it's own PR. The newer V8 versions are much pickier about double property definitions so they'll basically abort on Ti.UI.View (and all subclasses) with this, so I needed the fix to get a working version.

sgtcoolguy:
I have not yet addressed the IIFE for app.js to see if I can avoid it.

There's a larger issue in that the move to using NDK r14b to build v8 seems to have broken our v8 builds silently. Our build scripts weren't exposing the failure. I'm looking into getting V8 5.9 built properly with NDK r12b (which is what chromium uses for V8) and ninja (rather than gyp).

garymathews:
@sgtcoolguy Can we use the latest V8 build from `v8_titanium` `5.9.211.41` ?

Also, see https://github.com/appcelerator/titanium_mobile/pull/9082#discussion_r119995532

sgtcoolguy:
@garymathews Finally got around to removing the wrapping function.

Should be set for QE to test.

Also, we're now using V8 6.0.286.55 as the base. No breaking API changes since 5.7 so far.

android/runtime/v8/src/native/InspectorClient.cpp
```diff
@@ -107,27 +107,32 @@ void InspectorClient::CallAndPauseOnStart(const v8::FunctionCallbackInfo<v8::Val
...
+	// They wrap the source with a function just like Module.wrap for standard code
+	// Then compile the function and pass it into this method to schedule a pause and then invoke it.
+	// Instead, we pass the app.js source and filename and compile it, schedule a pause and then run it.
+	// This does duplicate some existing logic in ScriptModule.cpp for Script.runInThisContext impl
+	v8::Local<v8::String> source = args[0]->ToString();
+	v8::Local<v8::String> filename = args[1]->ToString();
+	v8::Local<v8::Script> script = v8::Script::Compile(source, filename);
+	if (script.IsEmpty()) {
+		// Hack because I can't get a proper stacktrace on SyntaxError
+		args.GetReturnValue().Set(v8::Undefined(isolate));
```
garymathews: Maybe a `TryCatch` would work here?

```C++
v8::TryCatch tryCatch(isolate);
...
v8::Local<v8::Script> script = v8::Script::Compile(source, filename);
if (script.IsEmpty()) {
	v8::String::AsciiValue syntaxError(tryCatch.Exception()); // obtain exception message
	V8Util::fatalException(isolate, tryCatch); // or throw fatal exception?
        ...
```

sgtcoolguy:
For QE: Please use "rebase and merge" to merge this PR once tested. I have made a point of trying to explicitly clean up my commits and would like to retain them.

kolipakakondal:
@garymathews @sgtcoolguy - Are we shipping this for SDK 7.0 release? If yes, I'll move the related ticket TISTUD-8750 for Studio 5.0 release

lokeshchdhry:
FR Passed.

Followed the steps above to test. Breakpoint are hit successfully.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

lokeshchdhry:
@sgtcoolguy , Can you please merged this PR. I am done with the FR.
Danger is not allowing to merge because of missing unit tests.

sgtcoolguy:
@kolipakakondal Yes, this is slated for SDK 7.0

PR #9083, opened 2017-05-24
[TIMOB-24732] Fix issue where build process would continue before zip…

ewanharris:
… creation had finished

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24732

Verification:

1. Get the [titanium-google-maps](https://github.com/hansemannn/titanium-google-maps) module
2. Build the module using `ti build -p ios -l trace`
    - Build process should not error out, and the example app should launch on device


sgtcoolguy:
LGTM. cc @cb1kenobi look ok?

cb1kenobi (approved):
Looks good to me. I can't believe we didn't catch this sooner.

htbryant94 (approved):
Approved! Modules are now built correctly, and a working zip file was created. Running the example app as a new project with the newly built module to simulators and devices was also successful.

PR #9086, opened 2017-05-25
[TIMOB-24734] Add compatibility with Hyperloop AAR handling

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24734

**Optional Description:**
Hyperloop currently still has its own AAR handling which interferes with the new AAR handling in the SDK. This fix will ensure the compatibility by reverting any changes Hyperloop is trying to make to the builder.


cb1kenobi:
LGTM

sgtcoolguy:
I opened appcelerator/hyperloop.next#179 to make hyperloop build against SDK 6.2.0+ (with the new module api version of 4, due to V8 API changes).

android/cli/hooks/aar-transform.js
```diff
@@ -41,9 +41,10 @@ const LIBRARY_ORIGIN_PORJECT = 'Project';
 
 exports.cliVersion = '>=3.2';
 
-exports.init = function (logger, config, cli) {
+exports.init = function (logger, config, cli, appc) {
```
sgtcoolguy: The varying argument order for the same objects bugs me:
`(logger, config, cli, appc)` vs
`(cli, builder, appc, logger)` vs 
`(builder, callback)` vs
`(builder, logger, callback)`

I prefer to try and retain argument order relatively if possible, so I guess in these cases it would be:
`(logger, config, cli, appc, callback)` - where `builder` falls, I'm not sure beyond having it before `callback`

android/cli/hooks/aar-transform.js
```diff
@@ -444,3 +445,102 @@ class SimpleFileCache {
...
+		if (module.id === 'hyperloop' && appc.version.lt(module.version, '2.2.0')) {
+			hyperloopModule = module;
+			return true;
+		}
+	});
+	if (hyperloopModule === null) {
+		return;
+	}
+
+	var hyperloopBuildPath = path.join(builder.projectDir, 'build/hyperloop/android');
```
sgtcoolguy: does node handle unix slashes properly on Windows? I always tend to avoid using separators just in case:
`var hyperloopBuildPath = path.join(builder.projectDir, 'build', 'hyperloop', 'android');`

sgtcoolguy (approved):
LGTM

android/cli/hooks/aar-transform.js
janvennemann: I sorted the arguments a little bit by their priority. `cli` and `builder` being the main objects i interact with, followed by utility stuff like `appc` and `logger` and the `callback` always being the last argument.

android/cli/hooks/aar-transform.js
janvennemann: Yes, `path.join` will properly take care of the slashes on Windows

PR #9097, opened 2017-05-30
[TIMOB-24695] Android: Fix URL decoding

garymathews:
- Use decoded URL for requests

```JS
    iv = Ti.UI.createImageView({
        image: 'http://scontent.xx.fbcdn.net.rsz.io/v/t1.0-9/s720x720/17795851_419371511758070_7481565181564022529_n.jpg?oh=7772aea44d2f40fdbd42285ca61d7aac&oe=59771D60?mode=crop&width=333&height=250'
    });

win.add(iv);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24695)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
@@ -9,6 +9,7 @@ package org.appcelerator.titanium.util;
 import java.io.InputStream;
 import java.lang.ref.SoftReference;
 import java.net.URI;
+import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.concurrent.ExecutorService;
@@ -54,6 +55,10 @@ public class TiDownloadManager implements Handler.Callback
 
 	public void download(URI uri, TiDownloadListener listener)
 	{
+		try {
+			uri = new URI(URLDecoder.decode(uri.toString(), "utf8"));
+		} catch (Exception e) {
+		}
 		if (TiResponseCache.peek(uri)) {
 			sendMessage(uri, MSG_FIRE_DOWNLOAD_FINISHED);
 		} else {
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiFileHelper.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiFileHelper.java
@@ -20,6 +20,7 @@ import java.lang.ref.SoftReference;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -229,8 +230,9 @@ public class TiFileHelper implements Handler.Callback
 	private InputStream handleNetworkURL(String path) throws IOException
 	{
 		InputStream is = null;
+		URL u = new URL(path);
 		try {
-			URI uri = new URI(path);
+			URI uri = new URI(URLDecoder.decode(path, "utf8"));
 			if (TiResponseCache.peek(uri)) {
 				InputStream stream = TiResponseCache.openCachedStream(uri);
 				if (stream != null) {
@@ -238,10 +240,10 @@ public class TiFileHelper implements Handler.Callback
 					return stream;
 				}
 			}
+			u = uri.toURL();
 		} catch (URISyntaxException uriException) {
 		}
 
-		URL u = new URL(path);
 		InputStream lis = u.openStream();
 		ByteArrayOutputStream bos = null;
 		try {


jquick-axway:
@garymathews, you shouldn't send an HTTP request with a "decoded" URL. It can create an invalid URL. For example, if the URL contains an encoded space character (ex: "%20" or '+'), then the URLDecoder would turn that into a literal space ' ' character, making the URL invalid.

garymathews:
@jquick-axway Good catch! But `URLDecoder` shouldn't invalidate the URL like that

```JS
    iv = Ti.UI.createImageView({
        image: 'http://garymathews.com/pin%20wheel+.jpg'
        //image: 'http://garymathews.com/pin wheel+.jpg'
        //image: 'http://garymathews.com/pin%20wheel%2B.jpg'
        //image: 'http://garymathews.com/pin wheel%2B.jpg'
    });

win.add(iv);
win.open();
```

Should all still be valid

jquick-axway:
Okay. Thanks for testing it.
Perhaps the URI constructor re-encodes the URL again making this a non-issue.

**Side Note:**
The Axway JIRA site (link below; if you have access) has recently received an issue regarding the Titanium HttpClient double encoding URLs (ex: "%" -> "%%") on Android. I initially blew it off as an Appcelerator Studio logging issue (which is in fact a logging bug), but it now sounds like this may actually be an HTTP request bug similar to what you are fixing.
https://techweb.axway.com/jira/browse/SWAGGEN-133

android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
```diff
@@ -54,6 +55,10 @@ protected TiDownloadManager()
 
 	public void download(URI uri, TiDownloadListener listener)
 	{
+		try {
+			uri = new URI(URLDecoder.decode(uri.toString(), "utf8"));
+		} catch (Exception e) {
```
ypbnv: Can you add logging the exception?

android/titanium/src/java/org/appcelerator/titanium/util/TiFileHelper.java
```diff
@@ -229,19 +230,20 @@ public InputStream openInputStream(String path, boolean report)
...
+			URI uri = new URI(URLDecoder.decode(path, "utf8"));
 			if (TiResponseCache.peek(uri)) {
 				InputStream stream = TiResponseCache.openCachedStream(uri);
 				if (stream != null) {
 					// Fallback to actual download when null
 					return stream;
 				}
 			}
+			u = uri.toURL();
 		} catch (URISyntaxException uriException) {
```
ypbnv: And here?

ypbnv (changes requested):
Looks fine. Just the empty catch block are bugging me.

jquick-axway:
So, I've spent most of the day digging into this.

First, I've confirmed that the URLDecoder.decode() method in your code **can** produce an invalid URL that will cause the URI() constructor to throw a "URISyntaxException". For example, your "http://garymathews.com/pin%20wheel+.jpg" URL when decoded (%20 and '+' turn into spaces) will cause an exception when fed into the URI() constructor because a space character is not valid in a URL. Your current code catches it and ends up using the old URL instead.

Second, the "URLEncoder" and "URLDecoder" Java classes are only intended to encode/decode URL "parameters" (ie: the part to the right of the '?' in the URL). They are not intended to encode/decode the URL host name and path (I just learned this today as well). Query parameters are expected to use the "application/x-www-form-urlencoded" MIME format. The rest of the URL is expected to follow the "RFC 2373" spec which can be encoded via the Java Uri.encode() and Uri.decode() methods. And in my testing, all of these Java classes encode some characters they shouldn't and don't fully support both specs. Unfortunately, I haven't found any "magic" class/method that'll easily fully encode URLs for us without issues either. :(

So, the root cause of this bug (where question marks in the query parameters are being percent encoded when they shouldn't) lies in our TiUrl.getCleanUri() method. This is the method that gets called by our ImageView and HttpClient to encode URLs. A simple solution to this particular problem would be to modify the line here...
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/util/TiUrl.java#L327

...to the following (I've added '?' and '/' to the do-not-encode list)...

```java
builder.encodedQuery(Uri.encode(Uri.decode(base.getQuery()), "&=?/"));
```

But that said, there are other URL escaping issues in this code as well that's affecting ArrowDB queries in the following JIRA case assigned to me.
https://techweb.axway.com/jira/browse/SWAGGEN-133

I'm already re-working TiUrl.getCleanrUri() now. Do you mind if I take over fixing this? Otherwise our code changes are going to collide.

garymathews:
@jquick-axway Thanks for looking into this, you should take over this issue. I'll close this PR and assign TIMOB-24695 to you 👍

PR #8558, opened 2016-10-27
[TIMOB-24080] Implementing ParagraphStyle Ti.UI.AttributedString

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24080

PR #9100, opened 2017-05-31
[TIMOB-23367]: Support of Paragraph style

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23367


hansemannn:
Cannot take this PR inside the core, here are some reasons:
- The API is not cross-platform, therefore has no place on the `Ti.UI` namespace
- There is an open (community) PR that adds all this properties as attribute-constants, which is more lightweight, since we do not cross the kroll-bridge for every attribute (performance, easier usage)
- The ticket of the mentioned PR does also include all properties mentioned in the ticket of this PR, so we would likely resolve that ticket as a Duplicate and move the communication to the first one

sgtcoolguy:
@hansemannn Are you referring to #8558? Should we close this one in favor of that one?

PR #9106, opened 2017-06-04
[TIMOB-24769] iOS: Remove SBJSON from SDK-core

hansemannn:
This is **no** breaking change!

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24769


vijaysingh-axway (approved):
CR and FT Passed.

htbryant94 (approved):
Approved, using the provided test case, did not experience any errors during compile and run time, information returned to the console log was also verified to be working as intended.

PR #9116, opened 2017-06-06
[TIMOB-24654] Android native module build fails with multiple commonjs assets

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24654

**Description:**
This is a follow-on fix to TIMOB-24654, reported by a user and seen when trying to compile syncserver native module for Android. When there are multiple JS files in the native module's assets folder, the CLI build would choke because it tried to call the same callback for each file rather than after all of them.

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -1153,7 +1153,8 @@ AndroidModuleBuilder.prototype.compileJsClosure = function (next) {
 		}),
 		closureJarFile = path.join(this.platformPath, 'lib', 'closure-compiler.jar');
 
-	async.each(jsFilesToEncrypt, function(file, callback) {
+	// Limit to 5 instances of Java in parallel at max, to be careful/conservative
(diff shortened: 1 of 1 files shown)


android/cli/commands/_buildModule.js
```diff
@@ -1153,7 +1153,7 @@ AndroidModuleBuilder.prototype.compileJsClosure = function (next) {
 		}),
 		closureJarFile = path.join(this.platformPath, 'lib', 'closure-compiler.jar');
 
-	jsFilesToEncrypt.forEach(function (file) {
+	async.each(jsFilesToEncrypt, function(file, callback) {
```
cb1kenobi: This should probably be an `async.eachLimit(jsFilesToEncrypt, 5, function (file, callback) {` so that it doesn't spin up hundreds of Java instances.

cb1kenobi (approved):

cb1kenobi:
**APPROVED**

PR #9117, opened 2017-06-06
(6_1_X) [TIMOB-24654] Android native module build fails with multiple commonjs assets

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24654

**Description:**
This is a follow-on fix to TIMOB-24654, reported by a user and seen when trying to compile syncserver native module for Android. When there are multiple JS files in the native module's assets folder, the CLI build would choke because it tried to call the same callback for each file rather than after all of them.

The master branch fix for same issue is #9116

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -1179,7 +1179,7 @@ AndroidModuleBuilder.prototype.compileJsClosure = function (next) {
 		}),
 		closureJarFile = path.join(this.platformPath, 'lib', 'closure-compiler.jar');
 
-	jsFilesToEncrypt.forEach(function (file) {
+	async.each(jsFilesToEncrypt, function(file, callback) {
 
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
To test/QE:
Try to build the android native module for https://github.com/appcelerator-modules/appcelerator.syncserver.client
Prior to this PR, it will fail due to a JS callback being fired multiple times.

These steps should fail with 6.1.0.GA, but if you select the 6.1.1 SDK version built from this PR, it should work.
```
git clone git@github.com:appcelerator-modules/appcelerator.syncserver.client.git
cd appcelerator.syncserver.client
cd android
ti sdk select 6.1.0.GA # or the version of the SDK from this PR
ti build -p android --build-only
```

cb1kenobi (approved):

cb1kenobi:
**APPROVED**

PR #9120, opened 2017-06-07
[TIMOB-24753]: iOS View sizes are incorrect in 6.1.0.GA (regression)

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24753


hansemannn (approved):
As the ticket that implemented the bounding-rect usage works with this solution. But please note that the `maxSize` property was there for another reason, see the `480px` check in it's declaration. I am still validating that case, but QE can already proceed with their tests. Required label-tests are:
- horizontal / vertical layout
- auto-width (FILL and SIZE) / percentages / static values
- label container that size different than the children / label
- left / right usages

We may want to create an own layout-focussed QE-app if we don't have so far. This kind of changes are always too sensitive to test with only 1-2 test-cases.

htbryant94 (approved):
Approved! View sizes are now calculated correctly when using percentages. Additionally, performed tests related to the following parameters:

- horizontal / vertical layout
- auto-width (FILL and SIZE) / percentages / static values
- label container that size different than the children / label
- left / right usages
- Portrait / Landscape Orientations


Tested on Device and Simulator.

PR #9121, opened 2017-06-07
[TIMOB- 24753](6_1_X):iOS View sizes are incorrect in 6.1.0.GA (regression)

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB- 24753


htbryant94 (approved):
Approved! View sizes are now calculated correctly when using percentages. Additionally, performed tests related to the following parameters:

horizontal / vertical layout
auto-width (FILL and SIZE) / percentages / static values
label container that size different than the children / label
left / right usages
Portrait / Landscape Orientations
Tested on Device and Simulator.

PR #9127, opened 2017-06-07
Increasing unit test timeout to 40 minutes. 

wluu:
The tests are timing out when there's a huge load on the Jenkins build. Plus, throttling doesn't help.


wluu:
Closing PR as invalid. The test suites were trying to load a file that didn't exist and the app failed to proceed. And, Williams is on top of it.

PR #9130, opened 2017-06-07
[TIMOB-24794] Updated to ioslib@1.4.3 which updates to node-ios-devic…

cb1kenobi:
[TIMOB-24794] Updated to ioslib@1.4.3 which updates to node-ios-device@1.3.3 and adds support for Node.js 8.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24794


hansemannn:
node-ios-device supporting Node 8 does not mean that Titanium supports Node 8 right? How is the status of supporting node 6/7 right now? I think I lost track about it.

cb1kenobi:
@hansemannn I have been using Node 8 for a while now with Titanium and it seems to work just fine. Even `node-ios-device` worked, but that's because it detected the binary was missing and rebuilt it for Node 8 on-the-fly. We got lucky. Sometimes a new Node version will break native addons.

sgtcoolguy:
@hansemannn @ewieberappc What can we do to move this PR on? I'd like to get this in and merged so that I can update #9048 with it and then merge that in (moving from checked-in node_modules to using npm5's package-lock.json).

hansemannn (approved):
Talked to Chris about some lines - LGTM, but would like to request a CR from @janvennemann just to be sure.

janvennemann (approved):
LGTM! 👍

ewieberappc (approved):
FR Passed. Able to build for sim and device using recent Xcodes (except 8.3.3 because of a known bug). No issues encountered and built apps ran without error.

PR #9131, opened 2017-06-08
[TIMOB-24796] iOS: Hybrid CommonJS and Native modules don't handle multiple js files

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24796

**Description:**
Fixes require resolution for js files living in the assets folder of a "native" module. Specifically, simply avoid mangling the filenames.

We hand titanium_prep a list of JS files to encrypt into a class, that uses a map of file paths to indices that then look up a block of bytes. Our require behavior for encrypted assets was changing '.' to '_' in the file paths first. This is because our old python-based module build for iOS would mangle filenames in the same way for js files in the assets folder. (In fact we still do mangle filenames like this in normal apps' JS files on device when encrypting, and we convert to the mangled name in TiUtils.m)

This fix looks for the non-mangled name first (because that's what the Node-js based module builds produce, and because name mangling doesn't seem to gain us anything); then falls back to trying the mangled name next.


sgtcoolguy:
QE/FR steps: https://jira.appcelerator.org/browse/TIMOB-24796?focusedCommentId=421562&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-421562

hansemannn (approved):

mukherjee2:
FR Passed. Details are in PR/9131.

PR #9138, opened 2017-06-12
[TIMOB-24731] : iOS Resolve iOS 8 deprecations for willAnimateRotationToInterfaceOrientation

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24731


iphone/Classes/TiUIScrollableViewProxy.m
```diff
@@ -336,13 +336,28 @@ -(CGFloat)autoHeightForSize:(CGSize)size
...
+
+- (void)willTransitionToTraitCollection:(UITraitCollection *)newCollection withTransitionCoordinator:(id <UIViewControllerTransitionCoordinator>)coordinator
+{
+    
+}
+
+- (void)preferredContentSizeDidChangeForChildContentContainer:(id <UIContentContainer>)container
+{
+    
+}
```
hansemannn: Super calls?

hansemannn (changes requested):
Great job @vijaysingh-axway! Left one note regarding `super` calls (or did you leave them blank intentionally), the rest looks fine. 

We need extensive test-cases for orientation tests of all involved elements (especially top-level-views like tab-group, navigation-window and container-views like tableview and scrollable-view). 

QE will have proper ones I guess!

iphone/Classes/TiUIScrollableViewProxy.m
vijaysingh-axway: This is intentional. No need to call super. These are end calls . These protocols  are implemented here  to play safe. If it get call from somewhere without check, app will ensure to not crash.  Required changes for scrollview is implemented as per previous implementation , in method -
- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id <UIViewControllerTransitionCoordinator>)coordinator

hansemannn (approved):

mukherjee2 (approved):
Node Version: 6.10.3
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.2
Appc CLI NPM: 4.2.9
Titanium SDK version: 7.0.0.v20170801101611
Appcelerator Studio, build: 4.9.1.201707200100
Xcode 8.3.2
iOS 9 and 10 Devices, iPhone 6S Plus, iPad Pro 12.9"

I tried all four test cases in the original Jira ticket, and rotating the windows appeared to work correctly. Split window worked correctly on the iPad when rotated.

PR #9027, opened 2017-05-08
[TIMOB-24664] Update to Android support library 25.3.1

garymathews:
- Update Android support libraries to `25.3.1`
- Set build target API from `23` -> `25`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24664)

PR #9142, opened 2017-06-14
[TIMOB-24504] Android: Implement bundled notifications

garymathews:
- Implement bundled notifications for Android N

###### TEST CASE
```JS
var win = Ti.UI.createWindow({title: 'TIMOB-24504', layout: 'vertical'}),
    btn_normal = Ti.UI.createButton({title: 'CREATE NOTIFICATIONS', height: 50, width: Ti.UI.FILL}),
    btn_group = Ti.UI.createButton({title: 'CREATE GROUP NOTIFICATIONS', height: 50, width: Ti.UI.FILL}),
    id = 100;

function createNotifications(num, group) {
    for (var i = num; i > 0; i--) {
        var n = Ti.Android.createNotification({
            icon: Ti.Android.R.drawable.ic_dialog_info,
            contentTitle: 'NOTIFICATION #' + i,
            contentText : 'Lorem ipsum dolor sit amet, facer eruditi omittantur cu pri, nibh nonumy putant eam eu.'
        });
        if (group) {
            n.setGroup(group);
        }
        Ti.Android.NotificationManager.notify(id++, n);
    }
    if (group) {
        var n = Ti.Android.createNotification({
            icon: Ti.Android.R.drawable.ic_dialog_info,
            contentTitle: 'NOTIFICATIONS',
            contentText : 'You have ' + num + ' notifications.',
            group: group,
            groupSummary: true
        });
        Ti.Android.NotificationManager.notify(id++, n);
    }
}

btn_normal.addEventListener('click', function() {
    createNotifications(3);
});

btn_group.addEventListener('click', function() {
    createNotifications(3, 'group_key');
});

win.add([btn_normal, btn_group]);
win.open();
```

##### NOTE: this requires https://github.com/appcelerator/titanium_mobile/pull/9027

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24504)


jquick-axway:
@garymathews, your doc changes state that this feature is only supported on Android 7 and higher, but since we're using Google's NotificationCompat, the Java setGroup() and setGroupSummary() methods are (according to Google; link below) supposed to be supported on older Android OS versions as well.
   https://android-developers.googleblog.com/2016/06/notifications-in-android-n.html

You may want to double check. Although, a stackoverflow post I've seen suggests that notification bundles via this compat library doesn't work on Android 4.0.
   https://stackoverflow.com/questions/31483772/android-grouped-notifications-and-summary-still-shown-separately-on-4-4-and-bel

jquick-axway:
Also, I think the "group" property should be renamed to "groupKey".

This makes it clear to the reader that this property must be given a string key. Versus a name like "group" implies that it stores/provides a collection of items. I know what I'm suggesting doesn't match Google's equivalent Java method, but it does match how they document it.

garymathews:
@jquick-axway Updated PR.

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -1504,6 +1504,21 @@
 	/**
 	 * @module.api
 	 */
+	public static final String PROPERTY_GROUP_KEY = "groupKey";
+
+	/**
+	 * @module.api
+	 */
+	public static final String PROPERTY_GROUP_ALERT_BEHAVIOUR = "groupAlertBehaviour";
```
jquick-axway: Our APIs use American English spelling rules. So, we should rename "Behaviour" to "Behavior", without the 'u'. (Sorry)

This would match our spelling with iOS' notification behavior property...
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.App.iOS.UserNotificationAction-property-behavior

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationProxy.java
```diff
@@ -342,7 +374,13 @@ public void setProgress(int max, int progress, boolean indeterminate)
 	public Notification buildNotification()
 	{
 		Notification notification = notificationBuilder.build();
-		notification.flags = this.flags;
+		
+		if (hasProperty(TiC.PROPERTY_GROUP_KEY)) {
+			// remove FLAG_AUTO_CANCEL as this will prevent group notifications
+			this.flags &= ~Notification.FLAG_AUTO_CANCEL;
+		}
+		notification.flags |= this.flags;
```
jquick-axway: Regarding `notification.flags |= this.flags`, is there a reason why you are changing it from an assignment to a bitwise-or operation? (Just curious in case there is something I'm not seeing here.)

I ask because the NotificationBuilder might set bits that can no longer be removed and the developer would have no means of clearing them. Versus the developer used to be able to set the exact flags he/she wanted before.

jquick-axway (changes requested):
Looks good. And definitely a great feature to add.

Just 1 question and 1 minor change request left.

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationProxy.java
jquick-axway: Never mind. The "Notification.Builder" sets flags (such as group summary) that we don't want to clear with an assignment.

garymathews:
@jquick-axway Updated

lokeshchdhry:
FR Passed.

The `groupkey` & the `groupSummary` properties works as expected. Also, notifications with this property group/bundle into a single notification.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.3-21
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.13
Node Ver:       6.10.1
Java Ver:       1.8.0_101
Devices:        Android 7.0

PR #9148, opened 2017-06-16
[TIMOB-24710] Android: Only dispose valid runtime

garymathews:
- Prevent re-disposal of `KrollRuntime` which would crash the app

##### NOTE: This issue is very hard to reproduce, I have no test-case

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24710)


android/runtime/common/src/java/org/appcelerator/kroll/KrollRuntime.java
```diff
@@ -213,14 +213,16 @@ protected void doInit()
...
 	{
-
-		Log.d(TAG, "Disposing runtime.", Log.DEBUG_MODE);
-
 		// Set state to released when since we have not fully disposed of it yet
 		synchronized (runtimeState) {
+			if (runtimeState == State.DISPOSED) {
+				return;
+			}
 			runtimeState = State.RELEASED;
```
jquick-axway: We should probably check if the runtime state is set to RELEASED as well to prevent the `internalDispose()` method from being called twice within the `dispose()` method.

```Java
if ((runtimeState == State.RELEASED) || (runtimeState == State.DISPOSED)) {
   return;
}
```

And we should add a guard to the `internalDispose()` method too...

```Java
synchronized (runtimeState) {
	if (runtimeState == State.RELAUNCHED) {
		// Abort the dispose if the application has been re-launched since we scheduled this dispose during the
		// last exit. Then set it back to the initialized state.
		runtimeState = State.INITIALIZED;
		return;
	}
	if (runtimeState == State.DISPOSED) { // <- Add this.
		return;                           // <- Add this.
	}                                     // <- Add this.
	runtimeState = State.DISPOSED;
}
```

What do you think?

jquick-axway (changes requested):

garymathews:
@jquick-axway Updated PR, I didn't keep the `runtimeState == State.RELEASED` guard as that will prevent full disposal of the runtime if that condition were to be met

jquick-axway (approved):
CR: Pass

mukherjee2:
Completed FR with this environment:
Node Version: 6.10.3
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.2
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.1.X PR-9200
Xcode 8.3.2

FR completed. See testing notes in PR-9200

PR #9150, opened 2017-06-16
[TIMOB-24841] iOS: Add Ti.UI.AlertDialog.tintColor

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24841


sgtcoolguy:
Please add a test for this property on iOS.

Copy https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.ui.alertdialog.test.js into `tests/Resources` and add a test that runs on iOS that creates an alert dialog with this property set and changed (basically copy the 'title' test but have it set/change/validate the tintColor)

hansemannn:
@sgtcoolguy There is no unit-test that could properly test it, only pseudo-tests like setting the kroll-property. PR's like this one would rather need a UI-test, but I can still add a test, sure.

sgtcoolguy:
You'd be surprised, but just some simple tests to set/change/get simple properties like this have actually turned up bugs. I understand that really this needs some sort of function UI test to verify the proper portion of the dialog gets colored appropriately, but that's beyond what we can do automated right now.

We could also long-term improve the test suite to validate properties handling specific types (i.e. that it handles all the same possible values for colors as other color properties do: hex, names, etc; or that it can handle being given an invalid type such as a boolean or number in an expected way) to make this test (and others like it) more useful.

hansemannn:
Did I get you correctly that I can copy the whole file into `tests/` and the build will automatically move it to the titanium-mobile-mocha-suite repo? Basically, I'd add this:

```js
	// Skip on Android, since it's an iOS-only property
	(utilities.isAndroid() ? it.skip : it)('cancel', function () {
		var bar = Ti.UI.createAlertDialog({
			tintColor: 'red'
		});
		
		// Check getter
		should(bar.tintColor).be.a.String;
		should(bar.getTintColor).be.a.Function;
		should(bar.tintColor).eql('red');
		should(bar.getTintColor()).eql('red');
		
		// Set new value
		bar.tintColor = '#f00';
		
		// Check getter again
		should(bar.tintColor).be.a.String;
		should(bar.getTintColor).be.a.Function;
		should(bar.tintColor).eql('#f00');
		should(bar.getTintColor()).eql('#f00');
	});
```

sgtcoolguy:
Yes, copy the whole file and make the edits to add the new test into it. We take the `tests/` folder as an override and copy it over top the titanium-mobile-mocha-suite when generating the unit test project.

hansemannn:
@sgtcoolguy thank you, added in 073b293.

tests/Resources/ti.ui.alertdialog.test.js
```diff
@@ -0,0 +1,104 @@
...
+		var bar = Ti.UI.createAlertDialog({});
+		should(bar.cancel).be.a.Number; // undefined on iOS and Android
+		should(bar.getCancel).be.a.Function;
+		bar.cancel = 1;
+		should(bar.cancel).eql(1);
+		should(bar.getCancel()).eql(1);
+	});
+	
+// Skip on Android, since it's an iOS-only property
+(utilities.isAndroid() ? it.skip : it)('tintColor', function () {
```
sgtcoolguy: We want to skip for any non-iOS platforms (mobilweb, windows).
`(utilities.isIOS() ? it : it.skip)('tintColor', function () {`

tests/Resources/ti.ui.alertdialog.test.js
```diff
@@ -0,0 +1,104 @@
...
+	should(bar.getTintColor).be.a.Function;
+	should(bar.tintColor).eql('red');
+	should(bar.getTintColor()).eql('red');
+	
+	// Set new value
+	bar.tintColor = '#f00';
+	
+	// Check getter again
+	should(bar.tintColor).be.a.String;
+	should(bar.getTintColor).be.a.Function;
```
sgtcoolguy: Shouldn't need to check this...

tests/Resources/ti.ui.alertdialog.test.js
```diff
@@ -0,0 +1,104 @@
...
+	should(bar.tintColor).eql('#f00');
+	should(bar.getTintColor()).eql('#f00');
```
sgtcoolguy: Is there a setter method we should check exists, and can be called?

It sounds simple/dumb, but I swear on Android I saw a case where the property setter and the actual getter method were "disconnected" where you could set different values for each and they didn't update the same underlying property.

tests/Resources/ti.ui.alertdialog.test.js
hansemannn: I thought it may be possible that the generated getter could be manipulated so the test would fail. Adjusting.

vijaysingh-axway (commented):
CR and FT passed.

htbryant94 (approved):
Approved! tintColor property for AlertDialog is now supported. Tested with both string & hex color values, as well as checking for errors if incorrect argument values are specified.

PR #9151, opened 2017-06-16
[TIMOB-24830] iOS: Attempt to fix require-caching

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24830


iphone/Classes/KrollBridge.m
```diff
@@ -1149,6 +1149,26 @@ - (TiModule *)loadNodeModules:(NSString *)path withDir:(NSString *)start withCon
...
+- (TiModule * _Nullable)cachedRequireModuleFromPath:(NSString *)path
+{
+	// See if we cached this path already
+	TiModule *module = [modules objectForKey:path];
+	if (module != nil) {
+		return module;
+	}
+
+	// Load the path if it does not exist in the path-cache so far
+	module = [self loadAsFileOrDirectory:[[path substringFromIndex:1] stringByStandardizingPath] withContext:context];
```
sgtcoolguy: You're dropping the first character here everytime, even though the places you replace with this method call isn't always do so.
I'd just take the path as-is. (or still do the stringByStandardizingPath call)
```
module = [self loadAsFileOrDirectory:[path stringByStandardizingPath] withContext:context];
```

iphone/Classes/KrollBridge.m
```diff
@@ -1163,18 +1183,25 @@ - (id)require:(KrollContext *)kroll path:(NSString *)path
 
 		// 2. If X begins with './' or '/' or '../'
 		if ([path hasPrefix:@"./"] || [path hasPrefix:@"../"]) {
+
 			// Need base path to work from for relative modules...
 			NSString *workingPath = [oldURL relativePath];
 			NSString *relativePath = (workingPath == nil) ? path : [workingPath stringByAppendingPathComponent:path];
-			module = [self loadAsFileOrDirectory:[relativePath stringByStandardizingPath] withContext:context];
-			if (module) {
+
```
sgtcoolguy: so for example, here we didn't drop the first character.

iphone/Classes/KrollBridge.m
```diff
@@ -1149,6 +1149,26 @@ - (TiModule *)loadNodeModules:(NSString *)path withDir:(NSString *)start withCon
 	return nil;
 }
 
+- (TiModule * _Nullable)cachedRequireModuleFromPath:(NSString *)path
```
sgtcoolguy: Might want a better name for this, like "loadWithCaching" or something.

iphone/Classes/KrollBridge.m
```diff
@@ -1163,18 +1183,25 @@ - (id)require:(KrollContext *)kroll path:(NSString *)path
...
+
 			// Treat '/' special as absolute, drop the leading '/'
-		}
-		else if ([path hasPrefix:@"/"]) {
-			module = [self loadAsFileOrDirectory:[[path substringFromIndex:1] stringByStandardizingPath] withContext:context];
-			if (module) {
+		} else if ([path hasPrefix:@"/"]) {
+
+			// Load cached require-module if exists
+			module = [self cachedRequireModuleFromPath:path];
```
sgtcoolguy:
```
module = [self loadWithCaching:[path substringFromIndex:1]];
```

iphone/Classes/KrollBridge.m
sgtcoolguy:
```
module = [self loadWithCaching: relativePath];
```

sgtcoolguy (changes requested):
This doesn't address "legacy" requires or requires of node_modules.

It also likely needs to remove the caching I did do with the fully resolved path (including file extension).

iphone/Classes/KrollBridge.m
hansemannn: Addressed comments, I didn't respect the different path-structures. One thing I am still wondering: Should `./file` be cached as `file` instead of `./file`? In the following test-case, the first and third require-call will write a new cache (one for `file`, one for `./file`):
```js
var test1 = require('/file');
var test2 = require('/file');
var test3 = require('./file');
var test4 = require('./file');

    backgroundColor: '#fff'
});

    title: 'Log Exports'
});

	console.log('#1 ' + test1.hello);
	console.log('#2 ' + test2.hello);
	console.log('#3 ' + test3.hello);
	console.log('#4 ' + test4.hello);
});

win.add(btn);
win.open();
```

sgtcoolguy:
Changes look good for absolute and relative paths, but "legacy" unprefixed paths are still effectively uncached. Also the caching I already had in wasn't modified/removed.

Node's approach is to resolve the filename first and then look at the cache or load.  Contrary to the user's ticket, it's not as simple as just checking the cache with the raw require string first: https://github.com/nodejs/node/blob/master/lib/module.js

Now for a simple relative/absolute path (so any path starting with '.' or '/') we can very quickly resolve the full path and store the module in our cache with that since it should always end up resolving to the same file. But that approach does present cache misses when they shouldn't miss (i.e. `require('file')` and `require('file.js')` both may load `/file.js` but be stored as two cache entries this way - or `require('/dir')` and `require('/dir/index') both resolving to `/dir/index.js`)

What I didn't do that Node does do is probably the right way to handle this instead: They also have a "path cache". So they effectively cache the full file path resolution result from the raw request key plus the set of possible paths to search : https://github.com/nodejs/node/blob/master/lib/module.js#L168

In our case that would mean we'd modify `require:` to generate a cache key using currentURL plus the raw request path, check the path cache for the resolved path, and then try to load the resolved path from he module cache. The problem right now is that It's not easy to stuff the raw request path and currentURL into a cache with the resolved path.

sgtcoolguy:
I'm taking a look now to see if I can add the path caching, etc.

sgtcoolguy:
Closing in favor of #9158

PR #9156, opened 2017-06-19
[TIMOB-18914] Android: Mark splitActionBar as removed

ypbnv:
Remove from source. Mark in documentation as "removed".

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18914


ypbnv:
@hansemannn Reverted the code removal. Added only deprecation tag and some notes. I guess we can link the Ti.UI.Toolbar in the notes once we have it moved on both platforms in the proper module.

lokeshchdhry:
@ypbnv , can QE do a FR for this ? Is CR pending ?

apidoc/Titanium/UI/Window.yml
```diff
@@ -784,6 +784,9 @@ properties:
     type: Boolean
     since: 3.6.0
     availability: creation
+    deprecated:
+        since: 6.2.0
+        notes: Deprecated in AppCompat theme. The same behaviour can be achived by using Toolbar.
```
hansemannn: So it still work? When you deprecate it, it still has to work. In case it does not anymore, it has to go into a major version (7.0.0).

hansemannn (changes requested):

apidoc/Titanium/UI/Window.yml
ypbnv: Yes, it should work. On Titanium SDK less than 4.0.0. Material Theme was added in that version and it does not support this property.

garymathews (approved):
CR: **PASS**

@ypbnv needs backporting to 6_2_X

mukherjee2:
Passes FR.

PR #9158, opened 2017-06-19
[TIMOB-24830] iOS: require cache improvements

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24830

**Description:**
This supersedes #9151 

This improves our require caching. Specifically, we use two caches now:
- module cache (keyed by full file path)
- path cache (keyed by combo of current dir and raw require string, values are fully resolved file paths)

We now have a top-level cache check where we see if we've ever resolved the raw require string (i.e. `./utilities/utilities`) from the current dir. If so, we get the fully resolved file path and then check the module cache for an entry. If we hit the module cache, we return the module. Otherwise we move on (NOTE: we could conceivably then just try to load the fully resolved path here, but we shouldn't ever have a case of a resolved path with no cached module at that path).

If we miss the path resolution cache, we run through our normal algorithm for resolving possible paths and trying to load them. For any given possible file location - before attempting to load the file from disk/encrypted bytes we check the module cache to see if we've already loaded this exact file path and return if we did. (before this PR we'd attempt to load the file contents first, and then check the cache if the file existed.) If we haven't we try to get the file contents. if we got some, try to load it; if not, move on.

One additional thing I tweaked is that when we see a non-prefix require like `require('something')` and we're checking for the oddball legacy CommonJS module, we search `\something\something.js` only before moving on whereas before we actually tried `\something\something.js`, then `\something\something.js.js` and then `\something\something.js.json` before moving on (there was a FIXME about this).

I've included WARN level statements to show the working path and the two cache behaviors (hits and misses) to make it easier to see what's going on here for FR - we should remove the log statements before merging.

So to sum up some changes:
- we should very quickly hit both caches if more than one file in the same directory uses the same exact require string (say two sibling files load `./utility` or `mocha`, the first time we'll load normally, second time we'll hit the cache right away)
- if the string varies but ends up resolving to a file we've already loaded before, we should grab the cached module instead of loading the file contents and then grabbing the cached module. We will still search the possible paths in-order first. But the list off possible paths searched should be very small for prefixed requires (stat with '/', './', or '../').
- Using legacy style un-prefixed requires to load a file absolutely will give us the worst load behavior when uncached: we'll try to load it as a legacy CommonJS file (id/id.js), then try to load it as a newer commonjs directory ('id/package.json, id/index.js, id/index.json), then check node_modules for a number of locations, and finally assume it should be treated as absolute.
We can see this in our own test suite with `require('ti-mocha')` where we try many paths first and hit on the second-to-last path we'd search. We try the following paths in order:
  - `ti-mocha/ti-mocha.js`
  - `ti-mocha/package.json`
  - `ti-mocha/index.js`
  - `ti-mocha/index.json`
  - `node_modules/ti-mocha`
  - `node_modules/ti-mocha.js`
  - `node_modules/ti-mocha.json`
  - `node_modules/ti-mocha/package.json`
  - `node_modules/ti-mocha/index.js`
  - `node_modules/ti-mocha/index.json`
  - `ti-mocha`
  - `ti-mocha.js`
 - `ti-mocha.json`

I added a note that if we can actually determine file/dir existence, we may be able to shortcut a lot of choices if a given path doesn't exist as either file or directory - but I'm not sure if we can reliably do so when we encrypt JS assets.


iphone/Classes/KrollBridge.m
```diff
@@ -1009,37 +988,63 @@ - (TiModule *)loadJavascriptText:(NSString *)data fromFile:(NSString *)filename
...
-	NSString *filename = path;
-	NSString *data = [self loadFile:filename];
+	// check cache first
+	if (modules != nil) {
+		KrollWrapper *module = [modules objectForKey:path];
+		if (module != nil) {
+			NSLog(@"[WARN][MODULE CACHE HIT] : %@", path);
+			return module;
+		}
+		NSLog(@"[WARN][MODULE CACHE MISS] : %@", path);
```
hansemannn: Remove the `[WARN]` logs or make them `[TRACE]` / `[DEBUG]`. Should be done for all occurrences of this PR. Otherwise, der developers will get complain about internal / unrelated logs thrown to them.

iphone/Classes/KrollBridge.m
```diff
@@ -1009,37 +988,63 @@ - (TiModule *)loadJavascriptText:(NSString *)data fromFile:(NSString *)filename
 	return module;
 }
 
-- (TiModule *)loadAsFile:(NSString *)path withContext:(KrollContext *)kroll
+- (KrollWrapper *)cachedLoadAsFile:(NSString *)path asJSON:(bool)json withContext:(KrollContext *)kroll
```
hansemannn: Change `bool` to Objective-C `BOOL`

iphone/Classes/KrollBridge.m
```diff
@@ -1009,37 +988,63 @@ - (TiModule *)loadJavascriptText:(NSString *)data fromFile:(NSString *)filename
...
+}
+
+- (KrollWrapper *)loadAsFile:(NSString *)path withContext:(KrollContext *)kroll
+{
+	NSString *filename = path;
+
+	// 1. If X is a file, load X as JavaScript text.  STOP
+	// Note: I modified the algorithm here to handle .json files as JSON, everything else as JS
+	NSString *ext = [filename pathExtension];
+	bool json = (ext != nil && [ext isEqual:@"json"]);
```
hansemannn: Same here, use `BOOL`.

iphone/Classes/KrollBridge.m
```diff
@@ -1062,25 +1067,35 @@ - (TiModule *)loadAsDirectory:(NSString *)path withContext:(KrollContext *)kroll
...
+	if (module != nil) {
+		return module;
 	}
+
 	// 3. If X/index.json is a file, parse X/index.json to a JavaScript object. STOP
 	filename = [path stringByAppendingPathComponent:@"index.json"];
-	data = [self loadFile:filename];
-	if (data != nil) {
-		return [self loadJavascriptObject:data fromFile:filename withContext:context];
+	module = [self cachedLoadAsFile:filename asJSON:true withContext:context];
```
hansemannn: Use `YES` :-)

iphone/Classes/KrollBridge.m
```diff
@@ -1152,65 +1167,105 @@ - (TiModule *)loadNodeModules:(NSString *)path withDir:(NSString *)start withCon
 - (id)require:(KrollContext *)kroll path:(NSString *)path
 {
 	NSURL *oldURL = [self currentURL];
+	NSString *workingPath = [oldURL relativePath];
+	NSString *pathCacheKey;
```
hansemannn: Should be mutable (`NSMutableString`), since line 1181 can override it's current value.

iphone/Classes/KrollBridge.m
```diff
@@ -1009,37 +988,63 @@ - (TiModule *)loadJavascriptText:(NSString *)data fromFile:(NSString *)filename
...
+	if (module != nil) {
+		return module;
 	}
+
 	// 3. If X.json is a file, parse X.json to a JavaScript Object.  STOP
 	filename = [path stringByAppendingString:@".json"];
-	data = [self loadFile:filename];
-	if (data != nil) {
-		return [self loadJavascriptObject:data fromFile:filename withContext:context];
+	module = [self cachedLoadAsFile:filename asJSON:true withContext:context];
```
hansemannn: Use `YES`.

iphone/Classes/KrollBridge.m
```diff
@@ -1062,25 +1067,35 @@ - (TiModule *)loadAsDirectory:(NSString *)path withContext:(KrollContext *)kroll
 
 	// 2. If X/index.js is a file, load X/index.js as JavaScript text.  STOP
 	filename = [path stringByAppendingPathComponent:@"index.js"];
-	data = [self loadFile:filename];
-	if (data != nil) {
-		return [self loadJavascriptText:data fromFile:filename withContext:context];
+	KrollWrapper *module = [self cachedLoadAsFile:filename asJSON:false withContext:context];
```
hansemannn: Use `NO`.

iphone/Classes/KrollBridge.m
```diff
@@ -1152,65 +1167,105 @@ - (TiModule *)loadNodeModules:(NSString *)path withDir:(NSString *)start withCon
...
+				// TODO Grab the first path segment and see if it's a node module or commonJS module
+				// We should be able to organize the modules in folder to determine if the user is attempting to
+				// load one of them!
+
+				// Look for CommonJS module
+				if (![path containsString:@"/"]) {
+					// For CommonJS we need to look for module.id/module.id.js first...
+					// Only look for this _exact file_. DO NOT APPEND .js or .json to it!
+					NSString *filename = [[path stringByAppendingPathComponent:path] stringByAppendingPathExtension:@"js"];
+					module = [self cachedLoadAsFile:filename asJSON:false withContext:context];
```
hansemannn: Use `NO`.

iphone/Classes/KrollBridge.m
```diff
@@ -1009,37 +988,63 @@ - (TiModule *)loadJavascriptText:(NSString *)data fromFile:(NSString *)filename
...
+	if (module != nil) {
+		return module;
+	}
+
 	// 2. If X.js is a file, load X.js as JavaScript text.  STOP
 	filename = [path stringByAppendingString:@".js"];
-	data = [self loadFile:filename];
-	if (data != nil) {
-		return [self loadJavascriptText:data fromFile:filename withContext:context];
+	module = [self cachedLoadAsFile:filename asJSON:false withContext:context];
```
hansemannn: User `NO`.

hansemannn (changes requested):
Left a few comments regarding Obj-C types (BOOL vs bool, NSMutableString vs. NSString). Will do a FT tomorrow, do our existing unit-tests address all cases already? I'll test with those as well.

iphone/Classes/KrollBridge.m
sgtcoolguy: I left these in intentionally for testing this PR. I do plan to remove them before we merge. Just thought it'd be easier for you'QE to have some transparency about what's going on under the hood to validate behavior.

hansemannn (approved):

hansemannn:
LGTM!

sgtcoolguy:
@ewieberappc So I added some debug logging to give transparency about cache hits/misses to help validate this PR. Once you've given the thumbs up for this PR, **please do not merge it**. Let me know and I'll back out the logging and then we can merge this and #9238 and #9232

PR #9161, opened 2017-06-19
[6_1_X][TIMOB-24801] Android: Ti.UI.Label refactoring/fixes.

jquick-axway:
**Changes:**
- [[TIMOB-24801]](https://jira.appcelerator.org/browse/TIMOB-24801) Fixed bug where height can fluctuate on every "postlayout" when "minimumFontSize" is applied.
- [[TIMOB-24670]](https://jira.appcelerator.org/browse/TIMOB-24670) Fixed bug where "lines" property wasn't working.
  * Now correctly sets the height of the label view based on font line height.
  * Note that it's okay for "lines" to be greater than 1 if the label is in single-line mode.
  * Set to <= 0 or null to clear this property.
- [[TIMOB-13895]](https://jira.appcelerator.org/browse/TIMOB-13895) Fixed bug where setting "lines" property will not apply END ellipsis to multiline truncated text.
  * Will now correctly apply ellipsis even if "wordWrap" is set true.
- Modified label to not auto-size the view width beyond the bounds of the parent if "minimumFontSize" is applied.
  * Now matches iOS' behavior.
- Fixed "minimumFontSize" to not upscale the font to fit the width. Should only downscale.
  * Used to scale larger than default font size.
  * Now matches iOS' behavior.
- Fixed "minimumFontSize" bug where it would not downscale the text enough to fit the width, causing text to be clipped.
- Fixed "minimumFontSize" to respect the "ti.ui.defaultunit" for give font size.
  * Used to be hardcoded to pixels.
  * Now consistent with the units used by the "Label.font" property.
- Fixed "minimumFontSize" bug where label height would not be adjusted to fit its downscaled text.
  * Will only auto-adjust label height if the "height" or "lines" property have not been set.
- Fixed "minimumFontSize" bug where text can sometimes downscale lower than the minimum.
- Fixed bug where "touchEnabled" property would change to true after setting the "html" or "autoLink" properties.
  * Now preserves the original setting.
- Fixed "minimumFontSize" to show text as single-line like iOS.
  * No longer displays line wrapped multiline text. (This was a bug.)
- Fixed "minimumFontSize" to trim all text proceeding a newline character or "html" break.
  * Now matches iOS' behavior.
- Fixed bug where "ellipsize" constants START and MIDDLE would not show ellipsis.
  * Bug was introduced in 2015 due to [TIMOB-16911](https://jira.appcelerator.org/browse/TIMOB-16911) fix.
  * Note: If "autoLink" is true, then label will use END instead of START or MIDDLE ellipsis due to native Android limitation.
- When setting "ellipsize" to START or MIDDLE, the label text will switch to single-line mode, like how it works with MARQUEE.
- Modified "minimumFontSize" to support ellipsize constants START, MIDDLE, END, MARQUEE, and NONE. (Used to only support END.)

**Note to Reviewers:**
The reason there are so many changes listed above is because our Java code needed to be refactored so that we can update the Android "TextView" native UI settings in a controlled manner via our new Java TiUILabel.updateLabelText() method. The order of method calls to Google's "TextView" class is important (ie: a call to 1 method can reset the setting of a prior method call) and we need to ensure that the single-line/multiline, lines/maxLines, and ellipsis settings don't get reset when auto-downscaling the text via our "minimumFontSize" feature. So, by refactoring code in this manner, we ended up fixing other label issues/regressions as well.

**Test Cases:**
TBD


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -51,8 +56,12 @@
 	private float shadowX = 0f;
 	private float shadowY = 0f;
 	private int shadowColor = Color.TRANSPARENT;
-	private String minimumFontSize = null;
-	private String autoshrinkSetFontSize = null;
+	private int autoLinkFlags;
+	private int viewHeightInLines;
+	private int maxLines = Integer.MAX_VALUE;
+	private float minimumFontSizeInPixels = -1.0f;
```
garymathews: pixels should be defined using `int`

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -88,99 +99,176 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
...
 		};
 		tv.setGravity(Gravity.CENTER_VERTICAL | Gravity.LEFT);
 		tv.setPadding(0, 0, 0, 0);
-		tv.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_FLAG_MULTI_LINE);
-		tv.setKeyListener(null);
 		tv.setFocusable(false);
-		tv.setSingleLine(false);
-		tv.setEllipsize(ellipsize);
+		tv.setEllipsize(this.ellipsize);
+		tv.setSingleLine(!this.wordWrap);
```
garymathews: keep this as:
```Java
tv.setSingleLine(false);
```
as this will prevent the use of multi-lined labels (see [TIMOB-24851](https://jira.appcelerator.org/browse/TIMOB-24851)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -88,99 +99,176 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
...
+		// Note: To keep it simple, don't let the scale be below 1x.
+		//       Only obsolete "ldpi" devices use scales lower than this.
+		float densityScale = view.getResources().getDisplayMetrics().density;
+		if (densityScale <= 0) {
+			densityScale = 1.0f;
+		}
+
+		// Calculate the pixel width within the view that the text will be rendered to.
+		float viewContentWidth;
+		{
```
garymathews: unnecessary inner block (we try not to use these)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -88,99 +99,176 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
...
+		if ((text == null) || (text.length() <= 0)) {
+			return;
+		}
+
+		// Restore the view's original font size.
+		float previousFontSize = textView.getTextSize();
+		textView.setTextSize(android.util.TypedValue.COMPLEX_UNIT_PX, this.unscaledFontSizeInPixels);
+
+		// Automatically downscale the font size to fit the width of the view.
+		// Note: Downscaling text width-wise is inaccurate and may take a few attempts.
```
garymathews: Why should it take a few attempts? Can this be re-worked to get it on the first try?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -88,99 +99,176 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
...
+			float textWidth = textPaint.measureText(text);
+
+			// Do not continue if the text fits within the view's bounds. (Don't need to downscale.)
+			if (textWidth <= viewContentWidth) {
+				break;
 			}
+
+			// Text does not fit within the width of view. Downscale the font.
+			float newFontSize = (viewContentWidth / textWidth) * currentFontSize;
+			{
```
garymathews: unnecessary inner blocks

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -193,53 +281,66 @@ public void processProperties(KrollDict d)
...
-		if (d.containsKey(TiC.PROPERTY_HTML)) {
-			String html = TiConvert.toString(d, TiC.PROPERTY_HTML);
-			if (html == null) {
-				//If html is null, set text if applicable
-				if (d.containsKey(TiC.PROPERTY_TEXT)) {
-					tv.setText(TiConvert.toString(d,TiC.PROPERTY_TEXT));
-				} else {
-					tv.setText(Html.fromHtml(""));
+		// Update the label's text, if provided.
+		{
```
garymathews: unnecessary inner block

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -357,18 +461,17 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
...
+			this.unscaledFontSizeInPixels = tv.getTextSize();
 			tv.requestLayout();
 		} else if (key.equals(TiC.PROPERTY_ELLIPSIZE)) {
-			if (newValue instanceof Boolean){
+			boolean wasUpdated = false;
+			if (newValue instanceof Boolean) {
 				ellipsize = (Boolean) newValue ? TruncateAt.END : null;
-			}
-			if (newValue instanceof Integer){
+				wasUpdated = true;
```
garymathews: remove `wasUpdated` and replace `wasUpdated = true;` with `updateLabelText();`

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -426,4 +533,188 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
...
+	 * Can optionally have measurements units appended to this number such as "px", "dp", etc.
+	 * If units are not assigned, then the default units assigned to Titanium will be used.
+	 * <p>
+	 * String can be null or empty, in which case, the min font size setting will be cleared
+	 * and the auto-downscaling feature will be disabled.
+	 */
+	private void setMinimumFontSize(String stringValue)
+	{
+		// Extract the font size and units (if any) from the given string.
+		float newSizeInPixels = -1.0f;
```
garymathews: use `int` for pixels

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -426,4 +533,188 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
...
+	}
+
+	/**
+	 * Determines if the Android "TextView" is to be displayed single-line or multiline.
+	 * @return Returns true if in single-line mode. Returns false if in multiline mode.
+	 */
+	private boolean isSingleLine()
+	{
+		// We're single-line if word wrapping (aka: line wrapping) is disabled.
+		if (this.wordWrap == false) {
```
garymathews: remove this [TIMOB-24851](https://jira.appcelerator.org/browse/TIMOB-24851)

android/titanium/src/java/org/appcelerator/titanium/util/TiConvert.java
```diff
@@ -331,11 +331,12 @@ public static void updateLayoutCenter(Object value, LayoutParams layoutParams)
 	 */
 	public static boolean toBoolean(Object value, boolean def)
 	{
-		try {
-			return toBoolean(value);
-		} catch (IllegalArgumentException e) {
-			return def;
+		boolean result = def;
+		if (value != null) {
+			try { result = toBoolean(value); }
```
garymathews: formatting

```Java
try {
    result = toBoolean(value);
} catch (Exception e) {
}
```

garymathews (changes requested):
A nice refactor, but needs some minor changes. 👍 

Could you also provide test cases (for TIMOB-24801)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
jquick-axway: There are 2 reasons why I made it this way.

1) Font sizes in JavaScript can be set in units other than pixels by the developer. They can be set in dips, inches, millimeters, and centimeters. By default, font units are in dips in Titanium. So, odds are, all font sizes will be stored in fractional form when converted to pixels.

2) Android supports fractional font sizes... even in pixels. So do Apple and Microsoft. It's actually noticeable on low DPI screens because it affects how the edges of the text are alpha-blended to the nearest whole screen/bitmap pixel.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
jquick-axway: I agree. Newlines should be honored when line/word-wrapping is disabled. But this would be a breaking change that we should not do in a patch release. How about we do this in 6.2?

In 6.2, we should also re-evaluate the usefulness of the "wordWrap" property as well. In my opinion, setting the width of the label implies that text should line-wrap at that width. This is how we do it on iOS and we don't even support "wordWrap" on that platform (because we don't need to).

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
jquick-axway: I did this on purpose to improve readability.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
jquick-axway: Unfortunately no.

The issue here is that non-monospaced fonts have variable character width. For example, the characters 'i' and 'w' have radically different widths. There are also character ligatures such as 'Æ'. And to make things even more complicated are cursive language, such as Arabic, where the text renderer will combine characters, making per character pixel width measurement kind of useless.

The issue the old code had before is that it would not always scale the font size down enough, causing text to be clipped. Such as with a string containing many wide characters.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
garymathews: I know, but the purpose of anonymous inner blocks is to restrict the variable scope and should be used for specific use cases instead of formatting.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
garymathews: Yes, lets keep it in 6.2.0

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
garymathews: 1. Good point, but currently minimumFontSize only accepts a `Number`. You cannot specify units like [fontSize](http://docs.appcelerator.com/platform/latest/#!/api/Font-property-fontSize) which we need to fix.

You can make use of these methods, like we do when setting `fontSize`
```
// TiUIHelper.java - org.appcelerator.titanium.util;
getSizeUnits(fontSize)
getSize(fontSize)

tv.setTextSize(getSizeUnits(fontSize), getSize(fontSize))
```

2. Just checked, we do use `float` values for defining `fontSize` but we use `int` for pixels in our LayoutEngine. Since you were converting to raw pixels I assumed using `int` would suffice and prevent rounding errors. But I agree using `float` is the better solution. I feel we shouldn't need to take multiple attempts to set the correct size though.

jquick-axway: Right. The "minimumFontSize" property still accepts a number, but its units is now relative to the app's default units (making it relative to all other font property units) which can be changed via "tiapp.xml" setting "ti.ui.defaultunit". That's one of the bugs I fixed, because it used to be hard-coded to pixels.

Take the below code as an example. In 6.1.0, the "font" property would have been set to "20dp" and the "minimumFontSize" would have been set to "20px". This is confusing and inconsistent. My new changes have them use the same default units.

```java
textField.font = { fontSize: 20 };
textField.minimumFontSize = 20;
```

Or perhaps you're referring to not allowing the "minimumFontSize" property to accept fontSize style strings like "20px"? You're right in that it would accept this the way it is coded now, but I was going to leave it in as an undocumented feature. But I can change it.

jquick-axway:
> Could you also provide test cases (for TIMOB-24801)

@lokeshchdhry has been working on a project to test each fix/change I've listed in the PR.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
garymathews: Yeah, I mean we should allow `minimumFontSize` to accept `Number/String` just like `fontSize`.

Also note, the docs say "_On Android the default font sizing is in pixels_" for `fontSize`, this needs fixing.

lokeshchdhry:
FR Passed.

1. Created an app for each test cases above to test & verified that minimumFontSize works as expected.
2. Verified fix for TIMOB-24670 & lines in label works fine.
3. Verified fix for TIMOB-13895 & setting ellipsize to a big text shows the ellipsize.

SDK Ver:        6.1.1 local build
OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.2
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.11
Node Ver:       6.10.1
Java Ver:       1.8.0_101
                      ⇨ Samsung Galaxy S4 --- Android 4.4.4

PR #9173, opened 2017-06-22
[TIMOB-24798] Fixed Xcode 9 support.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24798


iphone/cli/commands/_build.js
```diff
@@ -2374,6 +2348,30 @@ iOSBuilder.prototype.initialize = function initialize() {
 	}
 };
 
+iOSBuilder.prototype.findProvisioningProfile = function findProvisioningProfile(target, uuid) {
+	var provisioning = this.iosInfo.provisioning;
+
+	function getPP(type, uuid) {
+		var list = provisioning[type];
+		for (var i = 0, l = list.length; i < l; i++) {
```
janvennemann: A for..of loop would be easier to read

iphone/cli/hooks/package.js
```diff
@@ -174,40 +174,44 @@ exports.init = function (logger, config, cli) {
...
+			if (pp.appPrefix) {
+				exportsOptions.teamId = pp.appPrefix;
 			}
 		}
 
+		var keychains = builder.iosInfo.certs.keychains;
+		Object.keys(keychains).some(function (keychain) {
+			return (keychains[keychain].distribution || []).some(function (d) {
+				if (!d.invalid && d.name === builder.certDistributionName) {
+					exportsOptions.signingCertificate = d.fullname;
```
janvennemann: Missing return so `some` won't break early

janvennemann (changes requested):
Only two minor code style notes, rest looks good!

iphone/cli/commands/_build.js
cb1kenobi: We are not introducing ES2015 syntax into the current Titanium build.

iphone/cli/hooks/package.js
cb1kenobi: Good catch. Fixed.

iphone/cli/commands/_build.js
janvennemann: Makes sense, better to avoid mix and match the different ES versions.

janvennemann (approved):
CR: Passed

PR #9180, opened 2017-06-29
[TIMOB-18182] iOS: Add "inBackground" to local push notification

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18182


apidoc/Titanium/App/iOS/iOS.yml
```diff
@@ -1217,6 +1217,10 @@ properties:
     type: Dictionary
     optional: true
     since: "5.4.0"
+    
+  - name: inBackground
```
vijaysingh-axway: This should be written inside  
events : 
    "localnotificationaction" section because it tells the about received data and "inBackground" is property of received local notification data. Currently it written inside " NotificationParams", which is local notification creation object.

hansemannn: Good catch! Updated the PR.

vijaysingh-axway (approved):
CR and FT passed.

PR #9183, opened 2017-06-30
[TIMOB-24911] Android: Filtering R. files when building Android modules does not work on Windows.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24911

Change regular expression to handle platform specific paths.

**Description:**
Windows OS only. Regular expression does not match files' path with backslashes, so they are included in the .jar and duplicated.


android/cli/commands/_buildModule.js
```diff
@@ -1607,7 +1607,7 @@ AndroidModuleBuilder.prototype.packageZip = function (next) {
 			moduleJarArchive.on('error', cb);
 			moduleJarArchive.pipe(moduleJarStream);
 
-			var excludeRegex = /.*\/R\.class$|.*\/R\$(.*)\.class$/i;
+			var excludeRegex = new RegExp('.*\\' + path.sep + 'R\\\.class$|.*\\' + path.sep + 'R\\\$(.*)\\\.class$', 'i');
```
janvennemann: The three backslashes as escape characters are unnecessary here, for example `R\\.class$` already results in in the correct regex `R\.class$`

janvennemann (changes requested):

ypbnv:
@janvennemann Updated.

cb1kenobi (approved):
CR LGTM.

janvennemann (approved):

garymathews:
Merged for parity with `6_1_X` and `6_2_X`

PR #9191, opened 2017-07-03
[TIMOB-24929] iOS: Ti.UI.TabGroup - setTabsTranslucent

dbankier:
Jira: https://jira.appcelerator.org/browse/TIMOB-24929


iphone/Classes/TiUITabGroup.m
```diff
@@ -402,6 +402,12 @@ -(void)setUnselectedItemTintColor_:(id)value
 }
 #endif
 
+-(void)setTabsTranslucent_:(id)value
+{
+    BOOL tabsTranslucent =  [TiUtils boolValue:value def:YES];
+    controller.tabBar.translucent = tabsTranslucent;
+}
```
hansemannn: Rename to `translucent`, no idea why we ever started prefixing it with `tabs` some properties, it should be more unified. So in your case, it would just be `setTranslucent` instead.

hansemannn (changes requested):
Left a quick comment regarding the API design. I will help documenting the new property.

iphone/Classes/TiUITabGroup.m
dbankier: There is already a translucent property that refers to something else.

dbankier: Also badly named.

hansemannn: Oh, for the navigation-bar, I see. Let's keep it as it is then.

hansemannn (approved):

hansemannn:
LGMT, PR **approved**. Moving over to QE.

mukherjee2 (approved):
Node Version: 6.10.3
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.2
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.2.0 locally built with this PR
Xcode 8.3.2

PR #9196, opened 2017-07-06
[TIMOB-24940] Android: Problem with setTimeout and Toolbar on API 16/17

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24940

https://github.com/appcelerator/titanium_mobile/pull/9147 Needs to be merged first.

**Optional Description:**
An edge case where application restarts with 'stack-corrupted' error on Android JellyBean. It was exposed when setTimeout was used together with Toolbar.


garymathews:
##### NOTE
`-fstack-protector` is enabled by default and `--param=ssp-buffer-size` is set to `8` by default. Reducing it to `4` offers greater stack protection and also fixes this weird issue.

I'm not sure if it's a `gcc-4.9-x86` issue or an issue in Android JellyBean (not apparent in 4.3). But this is a safe way of fixing it.

jquick-axway:
Hmm... interesting. It does sound like a GCC stack protector bug. Probably byte packing related, which should be 4 bytes for an x86/32-bit build.

@lokeshchdhry, we should test this on the following devices:
- x86 Android Emulator
- 32-bit ARM device running Android 4.1 or 4.2, such as a Nexus.
- 64-bit ARM device, such as the Google Pixel XL.

jquick-axway (approved):
CR: Pass

mukherjee2 (approved):
Node Version: 6.10.3
NPM Version: 3.10.10
Mac OS: 10.12.4
Appc CLI: 6.2.3
Appc CLI NPM: 4.2.9
Titanium SDK version: 6.2.0.v20170825112753
Appcelerator Studio, build: 4.9.1.201707200100
Xcode 8.3.2

Passed FR. I tried it on an API 17 device. I used the testcase in the ticket. After pushing "Button"and waiting 3 seconds,  the correct messages were printed in the console, and the app did not experience any crash or restarts.

PR #9210, opened 2017-07-10
[TIMOB-24808]: iOS 11 Add support for Password AutoFill

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24808


apidoc/Titanium/UI/TextArea.yml
```diff
@@ -296,6 +296,16 @@ properties:
     constants: Titanium.UI.KEYBOARD_*
     default: <Titanium.UI.KEYBOARD_DEFAULT>
     platforms: [android, iphone, ipad]
+
+  - name: textContentType
+    summary: Provide the keyboard with extra information about the semantic intent of the text 
+        document.
+    type: String
```
hansemannn: Break summary into two lines.

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -296,6 +296,16 @@ properties:
...
+    constants: Titanium.UI.iOS.CONTENT_TYPE_*
+    default: <NULL>
```
hansemannn: Do not specify the default value if none is set.

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -296,6 +296,16 @@ properties:
     constants: Titanium.UI.KEYBOARD_*
     default: <Titanium.UI.KEYBOARD_DEFAULT>
     platforms: [android, iphone, ipad]
+
+  - name: textContentType
+    summary: Provide the keyboard with extra information about the semantic intent of the text 
+        document.
+    type: String
+    constants: Titanium.UI.iOS.CONTENT_TYPE_*
```
hansemannn: Should be `[Titanium.UI.iOS.CONTENT_TYPE_*]`

apidoc/Titanium/UI/TextField.yml
```diff
@@ -380,6 +380,16 @@ properties:
...
+
+  - name: textContentType
+    summary: Provide the keyboard with extra information about the semantic intent of the text 
+        document.
+    type: String
+    constants: Titanium.UI.iOS.CONTENT_TYPE_*
+    default: <NULL>
+    platforms: [iphone, ipad]
+    since: "6.2.0"
+    osver: {ios: {min: "10.0"}}
```
hansemannn: Same things as above (break into two lines, brackets in the `constants` value, no default value.

apidoc/Titanium/UI/iOS/iOS.yml
```diff
@@ -1699,6 +1699,231 @@ properties:
     osver: {ios: {min: "9.1"}}
     since: "5.1.0"
 
+  - name: CONTENT_TYPE_NAME
+    summary: Tell keyboard that content type is, name.
```
hansemannn: That sounds weird. Please use the [Apple phrasing](https://developer.apple.com/documentation/uikit/uitextcontenttype) for all constants.

apidoc/Titanium/UI/iOS/iOS.yml
```diff
@@ -1699,6 +1699,231 @@ properties:
...
+        Use with <Titanium.UI.TextField.textContentType> or <Titanium.UI.TextArea.textContentType>.    
+    type: String
+    permission: read-only
+    osver: {ios: {min: "11.0"}}
+    since: "6.2.0" 
+
+  - name: CONTENT_TYPE_PASSWORD
+    summary: Tell keyboard that content type is, password.
+    description: |
+        Use with <Titanium.UI.TextField.textContentType> or <Titanium.UI.TextArea.textContentType>.
```
hansemannn: This should also have a paragraph about the default behavior of fields having both email-address keyboard appearance and password-mask enabled, because in that case the password-autofill will be enabled automatically.

iphone/Classes/TiUITextWidget.m
```diff
@@ -160,6 +160,70 @@ -(void)setAutocorrect_:(id)value
 	[[self textWidgetView] setAutocorrectionType:[TiUtils boolValue:value] ? UITextAutocorrectionTypeYes : UITextAutocorrectionTypeNo];
 }
 
+-(void)setTextContentType_:(id)value
+{
+    ENSURE_TYPE_OR_NIL(value, NSString);
+    if (@available(iOS 10.0, *)) {
```
hansemannn: This whole method should be rewritten, since it has way too much manual logic and some missing guard inside it. Let try this something like this:
```objc
- (void)setTextContentType_:(id)value
{
    ENSURE_TYPE_OR_NIL(value, NSString);

    if (![TiUtils isiOS10OrGreater]) {
        NSLog(@"Setting a text-content-type is only available on iOS 10 and later");
        return;
    }

    [[self textWidgetView] setTextContentType:[TiUtils stringValue:value]];    
}
```

This has some advantages:
- Constants are mapped to it's values automatically
- There is no need for the Xcode 9-only `@available` syntax
- There is no need for the (currently missing) `#if IS_XCODE_9` macros
- iOS 11 constants do not need to be checked manually.
- Less error-prone due to no manual validation

iphone/Classes/TiUIiOSProxy.m
```diff
@@ -840,5 +840,36 @@ -(id)createFeedbackGenerator:(id)args
 #endif
 #endif
 
+#if defined(USE_TI_UITEXTWIDGET) || defined(USE_TI_UITEXTAREA) || defined(USE_TI_UITEXTFIELD)
```
hansemannn: `USE_TI_UITEXTWIDGET` is always defined when `USE_TI_UITEXTAREA ` or `USE_TI_UITEXTFIELD ` is defined, so I cannot think of a case where it is required to be checked manually.

iphone/Classes/TiUIiOSProxy.m
```diff
@@ -840,5 +840,36 @@ -(id)createFeedbackGenerator:(id)args
...
+
+MAKE_SYSTEM_STR(CONTENT_TYPE_NAME, UITextContentTypeName);
+MAKE_SYSTEM_STR(CONTENT_TYPE_NAME_PREFIX, UITextContentTypeNamePrefix);
+MAKE_SYSTEM_STR(CONTENT_TYPE_GIVEN_NAME, UITextContentTypeGivenName);
+MAKE_SYSTEM_STR(CONTENT_TYPE_MIDDLE_NAME, UITextContentTypeMiddleName);
+MAKE_SYSTEM_STR(CONTENT_TYPE_FAMILY_NAME, UITextContentTypeFamilyName);
+MAKE_SYSTEM_STR(CONTENT_TYPE_NAME_SUFFIX, UITextContentTypeNameSuffix);
+MAKE_SYSTEM_STR(CONTENT_TYPE_NICK_NAME, UITextContentTypeNickname);
```
hansemannn: Should be `CONTENT_TYPE_NICKNAME `, please also update inside the docs.

iphone/Classes/TiUIiOSProxy.m
```diff
@@ -840,5 +840,36 @@ -(id)createFeedbackGenerator:(id)args
...
+MAKE_SYSTEM_STR(CONTENT_TYPE_SUBLOCALITY, UITextContentTypeSublocality);
+MAKE_SYSTEM_STR(CONTENT_TYPE_COUNTRY_NAME, UITextContentTypeCountryName);
+MAKE_SYSTEM_STR(CONTENT_TYPE_POSTAL_CODE, UITextContentTypePostalCode);
+MAKE_SYSTEM_STR(CONTENT_TYPE_TELEPHONE_NUMBER, UITextContentTypeTelephoneNumber);
+MAKE_SYSTEM_STR(CONTENT_TYPE_EMAIL_ADDRESSS, UITextContentTypeEmailAddress);
+MAKE_SYSTEM_STR(CONTENT_TYPE_URL, UITextContentTypeURL);
+MAKE_SYSTEM_STR(CONTENT_TYPE_CREDIT_CARD_NUMBER, UITextContentTypeCreditCardNumber);
+
+#if IS_XCODE_9
+MAKE_SYSTEM_STR(CONTENT_TYPE_USER_NAME, UITextContentTypeUsername);
```
hansemannn: Should be `CONTENT_TYPE_USERNAME`, please also update inside the docs and JIRA example-code.

hansemannn (changes requested):
Needs some more adjustments in order to be approved, but looking forward to get those solved easily. Thx!

apidoc/Titanium/UI/iOS/iOS.yml
vijaysingh-axway: Do you mean it should be like TEXT_CONTENT_TYPE_NAME ?

iphone/Classes/TiUITextWidget.m
vijaysingh-axway: if (![TiUtils isiOS10OrGreater]) {
        NSLog(@"Setting a text-content-type is only available on iOS 10 and later");
        return;
    }  , I agree. 

  With  [[self textWidgetView] setTextContentType:[TiUtils stringValue:value]], I have already tried but it didn't worked. When I saw the class type of [TiUtils stringValue:value], it gives value	"NSTaggedPointerString " and for content type "__NSCFConstantString". This api accept only Constant string.
So we have to make the method large with manual logic.

apidoc/Titanium/UI/TextArea.yml
vijaysingh-axway: I think if the method accepts single constants it should be mentioned like   Titanium.UI.iOS.CONTENT_TYPE_* rather [Titanium.UI.iOS.CONTENT_TYPE_*] . e.g. see "keyboardType" inside TextArea.yml.

apidoc/Titanium/UI/iOS/iOS.yml
vijaysingh-axway: Sorry, I didn't get. Are you talking about to write a note about how  to enable password autofill ?

apidoc/Titanium/UI/iOS/iOS.yml
hansemannn: No, the summary should be proper english.

apidoc/Titanium/UI/iOS/iOS.yml
hansemannn: If people use the keyboard-type "email" and the text-fied property "passwordMask", it will enable auto-fill automatically, even when you did not build for iOS 11. That should be noted as an own paragraph there.

iphone/Classes/TiUITextWidget.m
hansemannn: You can just cast the string, the way it is now is not the best solution.

apidoc/Titanium/UI/iOS/iOS.yml
vijaysingh-axway: Done.

apidoc/Titanium/UI/iOS/iOS.yml
vijaysingh-axway: We are not mentioning anything related password autofill in document. So I don't think we should write above detail here. Better is we should write a wiki for giving detail about autofill password.

iphone/Classes/TiUITextWidget.m
vijaysingh-axway: We can not convert NSstring to __NSCFConstantString. If we do-
NSString *str = @"test string", it get automatically converted. It is done at compile time. Either we can simply write -
textContentType = @"username" 
or we write 
textContentType = UITextContentTypeUsername
In both cases we have to use if else condition for differentiation. First option is not good as we have to get the string value of each UITextContentType and write there.

iphone/Classes/TiUIiOSProxy.m
vijaysingh-axway: Done.

iphone/Classes/TiUIiOSProxy.m
vijaysingh-axway: Done

vijaysingh-axway:
Closing in favor  of https://github.com/appcelerator/titanium_mobile/pull/9368

PR #9179, opened 2017-06-29
[TIMOB-24898] Allow background transparency when borderRadius is set

garymathews:
- Disable canvas hardware-acceleration when a view has a border and background transparency set

###### TEST CASE
- Label should be visible before and after clicking the view
```JS
var win = Ti.UI.createWindow({title: 'TIMOB-24898', backgroundColor: 'gray'}),
    view = Ti.UI.createView({
        width: 300, height: 300,
        backgroundColor: '#00202020',
        borderRadius: 6, borderWidth: 2,
        borderColor: 'red'
    }),
    lbl = Ti.UI.createLabel({text: 'VISIBLE', color: 'white'});

view.addEventListener('click', function() {
    if (view.backgroundColor.startsWith('#00')) {
        view.backgroundColor = '#AA202020';
    } else {
        view.backgroundColor = '#00202020';
    }
});

win.add(lbl);
win.add(view);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24898)

PR #9214, opened 2017-07-11
[TIMOB-24898] Android: Guard against null backgroundColor

garymathews:
- Fix for https://github.com/appcelerator/titanium_mobile/pull/9179
- Guard against `backgroundColor: null/undefined`
- Jenkins tests should [not fail](https://jenkins.appcelerator.org/blue/organizations/jenkins/titanium-sdk%2Ftitanium_mobile/detail/PR-9207/2/pipeline/#step-121-log-301)

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24898)

PR #9215, opened 2017-07-11
Fix crash from calling Integer.intValue() (implicitly) on a null Integer reference for bgColor by doing unguarded bit shift on the value

sgtcoolguy:
**Description:**
An attempt to fix our crashing Android unit tests currently breaking the build. We didn't guard for a null `bicolor` value before doing a bit shift operation on it.

Relates to #9179


sgtcoolguy:
Looks like @garymathews already saw and posted a PR: #9214

PR #9219, opened 2017-07-13
[TIMOB-24973] iOS: Fix background recording issues while UIBackgroundModes set to true

hellllll0world:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-24973

**Edit by Hans**: `no tests`, since it is only testable when the app is in background, which cannot be triggered via unit-tests.

## bug report
This bug is very easy to reproduce, just follow the step below:

1. Add below code into tiapp.xml
```
<key>UIBackgroundModes</key>
                <array>
                    <string>audio</string>
                </array>
```
2. Using the blow code to do the testing

```
    width: Ti.UI.FILL,
    height: Ti.UI.FILL,
    backgroundColor: '#fff'
})

var recorder = Ti.Media.createAudioRecorder({
    format: Titanium.Media.AUDIO_FILEFORMAT_WAVE,
    compression: Ti.Media.AUDIO_FORMAT_ULAW
})

var startBtn = Ti.UI.createButton({
    top: 50,
    title: 'start'
})

var playBtn = Ti.UI.createButton({
    top: 100,
    title: 'play'
})

var stopBtn = Ti.UI.createButton({
    top: 150,
    title: 'stop'
})

var file

startBtn.addEventListener('click', function() {
    Titanium.Media.audioSessionCategory = Ti.Media.AUDIO_SESSION_CATEGORY_PLAY_AND_RECORD
    recorder.start()
})
playBtn.addEventListener('click', function() {
    Titanium.Media.audioSessionCategory = Titanium.Media.AUDIO_SESSION_CATEGORY_PLAYBACK
    var sound = Ti.Media.createSound({
        url: file,
        allowBackground: true
    });
    sound.play()
})
stopBtn.addEventListener('click', function() {
    file = recorder.stop()
})

win.add(startBtn)
win.add(playBtn)
win.add(stopBtn)
win.open()

```

3. Press the start button and lock the screen by press the power button of iPhone, the screen will be locked. Let me tell you what will happen after this
- if you unlock the screen less than 10 seconds after you lock, everthing is fine
- try to waiting more than 10 seconds before you unlocking your device,  for example 30 seconds, you'll got a blank record, when you try to play this record it will never stop even thoung it's a blank file.

## root cause
Everytime we start a new recording, a temp file will be created, however, the "ProtectionKey" is not set properly, so after the screen being locked, we don't have correct rights to write into that file, so we just unprotect this temp file to fix this bug. uh..., how does the magic "10 seconds" comes up? As we are know, iOS will left about 10 seconds to every app runing on it to do the cleanup thing before the system pause it(e.g. home button pressed), so if we unlock our device again shortly after we lock it everthing will be fine cause our App is stilling running in "foreground" now.

Sorry for my broken English, hopes you guys understand it. If you have any question please let'me know, Thanks.


hansemannn:
Hey there! I've created the JIRA-ticket for you. One thing I was wondering though if we need that e xplicit flag or if we can use a less "open" one (see [here](https://stackoverflow.com/a/23208355/5537752)). Any thoughts on that? I didn't dive into the PR so far, so it would be good to clear that before. Thx!

onceyoung:
hey, hansemannn, I think there's no need for that explicit flag, in most cases,  only we want to play the audio will us touch the audio file. Protect a temporary audio file is not that much important  in this situation. 
What's more, I think supporting saving the audio into a specific file created by users could be a better way to solve this problem(users can then control the protection key by using the Ti.Filesystem APIs) and the default behavior of the recorder is creating a temporary and set  NSFileProtectionNone on this file(same as we did it here). I think the whole thing should be a "feature" and what I did here is a bug fix(a subset of this feature). So, in my opinion, no flag is needed for this bug fix.
I hope I make it clear to you, thanks for your time to review the PR and create the JIRA-ticket for me.

hansemannn:
@onceyoung Let's integrate this change! Can you please re-sign the CLA to get this ready? The CLA-server was migrated to the Axway one lately, so I hope this does not block us from getting your contribution in :-).

onceyoung:
@hansemannn hey there, I re-sign the CLA, FYI

PR #9229, opened 2017-07-19
[TIMOB-25004][6_1_X] - Update ioslib to 1.4.9

ewanharris:
https://jira.appcelerator.org/browse/TIMOB-25004
It looks like we went from ioslib@1.4.8-> ioslib@1.4.7 on 6_1_X, this PR corrects that and updates to iosli@1.4.9, it didn't appear to have any code changes so I deemed it safe to go to 1.4.9 over 1.4.8

I believe this is the cause of the 'Cannot find ioslib' failures we have been seeing on Jenkins as when running `npm prune --production` npm would remove ioslib as the package.json states `^1.4.8` but the installed package was 1.4.7 which is invalid for the stated range, if the downgrade of the package was intentional we need to edit the package.json to reflect this

**This PR does not need to be cherrypicked to any other branch**


ewanharris:
Replace by #9252

PR #9233, opened 2017-07-21
Fetch pre-packaged native modules from URLs

sgtcoolguy:
**Description:**
This moves us from checking in the pre-built modules we ship with the SDK to fetching the zips during our SDK build from URLs. Should slim down our working directory size and repo clone size (assuming you do a shallow clone post-merge).


ingo:
LGTM.

PR #9242, opened 2017-07-24
[TIMOB-25010] Fix ioslib dependency for Windows and Linux

garymathews:
- `ioslib` cannot compile under `Windows` and `Linux`
- Move `ioslib` to `optionalDependencies`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25010)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "fields": "0.1.24",
     "glob": "7.1.1",
     "humanize": "0.0.9",
-    "ioslib": "^1.4.9",
     "lodash.defaultsdeep": "4.6.0",
     "markdown": "0.5.0",
     "moment": "2.18.1",
@@ -50,6 +49,9 @@
     "wrench": "1.5.9",
     "xcode": "0.9.2",
     "xmldom": "0.1.22"
+  },
+  "optionalDependencies":{
+    "ioslib": "^1.4.9"
   },
   "devDependencies": {
     "commander": "*",


sgtcoolguy:
So the issue here is that on non-macOS machines when you build the SDK for iOS, it won't have the right npm dependencies with this fix.

It looks like we already wiped the os restriction on ioslib so that you can install it on Windows/Linux. I think the issue here is that ioslib should sniff the os in it's install script/hook and not try to build on Windows/Linux and instead download the pre-built binaries (as it falls back to anyways).

cc @cb1kenobi

cb1kenobi:
I don't recall there was ever an OS restriction on ioslib. The restriction is in node-ios-device: https://github.com/appcelerator/node-ios-device/blob/master/package.json#L57-L59.

One of the earlier versions of node-ios-device only defined the build target in the gyp file if the platform was darwin, but it was removed and now it relies on npm to enforce the OS.

So, I think we need to merge this, but then we also need to add the OS restriction to ioslib. If @sgtcoolguy is cool with that, then I'll get on it right away.

sgtcoolguy:
Talked about this in today's standup. Consensus seems to be:
- Modify node-ios-device to have no OS restriction, and to not attempt to build on non-darwin OSes. Ideally it'd grab the relevant node binary on Windows/Linux, but not really necessary.
- `scons build|package` already runs node-ios-device's download_all script to grab the pre-built binaries for various node versions.

That should be the easiest/best way to try and maintain our current state where you could build an iOS SDk from Linux (and maybe Windows too).

cb1kenobi:
This PR will be remedied by https://jira.appcelerator.org/browse/TIMOB-25016.

PR #9252, opened 2017-07-25
[TIMOB-25004] Updated ioslib to 1.4.9.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25004


ewanharris (approved):
CR/FR Pass!

Pruning the dependencies no longer removes ioslib from the checked in node_modules directory, built SDK works as expected,

Only nit is that I just noticed that node-titanium-sdk is not at a fixed version not sure whether that should be changed?

Other than that LGTM! Thanks for taking it over :)

(Holding off merge as not sure whether this needs to go into in-qe-testing)

cb1kenobi:
I wouldn't worry about node-titanium-sdk. I don't think we need to wait for QE testing since master has 1.4.9 and it's been working.

ewanharris:
Grand, I'll go ahead and merge then

PR #9260, opened 2017-07-26
[TIMOB-25028] Fix issue where node-ios-device is not hoisted

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25028

node-ios-device can be in `node_modules` or `node_modules/ioslib/node_modules` we should consider that when executing the script.

Verification

1. Run node build/scon.js package ios
    - It should not error out
2. Let Jenkins run
     - It should be happy

### DIFF
--- a/build/packager.js
+++ b/build/packager.js
@@ -257,15 +257,31 @@ Packager.prototype.package = function (next) {
 		}.bind(this),
 		// Now include all the pre-built node-ios-device bindings/binaries
 		function (cb) {
-			if (this.targetOS == 'osx') {
-				exec('node bin/download-all.js', {cwd: path.join(this.zipSDKDir, 'node_modules', 'node-ios-device')}, function (err, stdout, stderr) {
-					if (err) {
-						console.log(stdout);
-						console.error(stderr);
-						return cb(err);
-					}
-					cb();
-				});
+			if (this.targetOS === 'osx') {
+				console.log(path.join(this.zipSDKDir, 'node_modules', 'node-ios-device'));
+				const hoistedPath = path.join(this.zipSDKDir, 'node_modules', 'node-ios-device');
+				const normalPath = path.join(this.zipSDKDir, 'node_modules', 'ioslib', 'node_modules', 'node-ios-device');
+				if (fs.existsSync(hoistedPath)) {
+					exec('node bin/download-all.js', {cwd: hoistedPath}, function (err, stdout, stderr) {
+						if (err) {
+							console.log(stdout);
+							console.error(stderr);
+							return cb(err);
+						}
+						cb();
+					});
+				} else if (fs.existsSync(normalPath)) {
+					exec('node bin/download-all.js', {cwd: normalPath}, function (err, stdout, stderr) {
+						if (err) {
+							console.log(stdout);
+							console.error(stderr);
+							return cb(err);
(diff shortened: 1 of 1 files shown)


build/packager.js
```diff
@@ -257,15 +257,31 @@ Packager.prototype.package = function (next) {
...
-				exec('node bin/download-all.js', {cwd: path.join(this.zipSDKDir, 'node_modules', 'node-ios-device')}, function (err, stdout, stderr) {
-					if (err) {
-						console.log(stdout);
-						console.error(stderr);
-						return cb(err);
-					}
-					cb();
-				});
+			if (this.targetOS === 'osx') {
+				console.log(path.join(this.zipSDKDir, 'node_modules', 'node-ios-device'));
```
cb1kenobi: Is the `console.log()` debug? It should probably be removed.

build/packager.js
```diff
@@ -257,15 +257,31 @@ Packager.prototype.package = function (next) {
...
+				if (fs.existsSync(hoistedPath)) {
+					exec('node bin/download-all.js', {cwd: hoistedPath}, function (err, stdout, stderr) {
+						if (err) {
+							console.log(stdout);
+							console.error(stderr);
+							return cb(err);
+						}
+						cb();
+					});
+				} else if (fs.existsSync(normalPath)) {
```
cb1kenobi: You have a lot of duplicate code. It can be simplified:

```js
let dir = path.join(this.zipSDKDir, 'node_modules', 'node-ios-device');

if (!fs.existsSync(dir)) {
    dir = path.join(this.zipSDKDir, 'node_modules', 'ioslib', 'node_modules', 'node-ios-device');
}

if (!fs.existsSync(dir)) {
    return cb(new Error('Unable to find node-ios-device module'));
}

exec('node bin/download-all.js', { cwd: dir, stdio: 'inherit' }, cb);
```

cb1kenobi (changes requested):

build/packager.js
ewanharris: Yeah my bad I was thinking that it could be better as I wrote it, updated.

build/packager.js
ewanharris: Removed

build/packager.js
```diff
@@ -257,15 +257,26 @@ Packager.prototype.package = function (next) {
...
+
+				if (!fs.existsSync(dir)) {
+				    dir = path.join(this.zipSDKDir, 'node_modules', 'ioslib', 'node_modules', 'node-ios-device');
+				}
+
+				if (!fs.existsSync(dir)) {
+				    return cb(new Error('Unable to find node-ios-device module'));
+				}
+
+				exec('node bin/download-all.js', {cwd: dir},  function (err, stdout, stderr) {
```
cb1kenobi: I would just inherit stdio, then you don't need to manually print the stdout/stderr. You can just pass in `cb` like this:

```js
exec('node bin/download-all.js', { cwd: dir, stdio: 'inherit' }, cb);
```

Also note that we put spaces inside braces/brackets around the props/elements.

cb1kenobi (approved):
Couldn't have written it better myself. **APPROVED**!

hansemannn (commented):
FT also works great!

PR #9263, opened 2017-07-27
[TIMOB-25030] Android: Include Windows gperf binary

garymathews:
- Include `gperf` binary for native Android module development on Windows

###### TEST CASE
```
appc new -p android -t timodule -n module --id com.module
cd module\android
appc run -p android --build-only
```
- Should compile without errors

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25030)


ypbnv (approved):
CR: Pass

sgtcoolguy:
Doesn't this have a GPL license? Can we include this legally? Also, it looks like we're including not just the binary, but all the source too.

cb1kenobi:
@sgtcoolguy We can add it to the repo, but we cannot distribute it.

cb1kenobi:
So, question, does scons add it to the distribution zip file?

ypbnv:
Yes it does.

garymathews:
@sgtcoolguy GPLv2 license, requires source to be available.

cb1kenobi:
@garymathews I think that we have/had contracts with customers that forbid us from distributing GPL dependencies with Titanium. We have a ticket to remove gperf because of this.

garymathews:
It would be nice to remove it as a dependency. I found [TIMOB-10724](https://jira.appcelerator.org/browse/TIMOB-10724) but that's because the user couldn't build an Android module on Windows due to not having gperf, which this will fix.

cb1kenobi:
@ingo Is it OK to:

A) commit GPL software to the titanium_mobile repo

and

B) bundle GPL software with the Titanium SDK distribution zip file?

ingo:
I don't believe we can commit it. Can't we just remove the dependency? It seems like it could be replaced by a different piece of software.

garymathews:
Closing in favour of https://github.com/appcelerator/titanium_mobile/pull/9265

PR #9264, opened 2017-07-27
[TIMOB-23958] Removed deprecated Python and unused files

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23958

**Description:**
Update of old PR #8453

Removes a large chunk of the support directory

PR #9265, opened 2017-07-28
[TIMOB-10724] Android: Remove native module gperf dependency

garymathews:
- Remove `gperf` as a dependency when compiling native Android modules
- Replace bindings lookup with `std::unordered_map` hash table utilising the [FNV-1](https://tools.ietf.org/html/draft-eastlake-fnv-03) hashing algorithm

###### TEST CASE
- Build a native Android module
- Add it to a Titanium project
- Use native module APIs successfully

###### NOTE: we still use gperf when building the SDK

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-10724)


cb1kenobi:
Maybe I'm missing something, but if this change was so trivial, then why did it take almost 5 years? Are we sure the generated code is equivalent performance-wise? I have no idea how std::unordered_map stacks up against the "perfect" hash that gperf generated.

garymathews:
gperf is great and will be faster than this implementation (and most other hash lookup implementations). But we're talking milliseconds, on a result that will be cached after the initial lookup.

I'm not sure why it taken five years to get here; why fix what isn't broken?

sgtcoolguy:
So I think this sort-of relates to #9264 - as I see that `support/common/simplejson` is supposedly used by genbootstrap.py in the Android build. This PR seems to eliminate the need for gperf on module builds, but not SDK builds.

First, I'd go all the way and eliminate it on SDK builds too. Then we can remove gperf as a dependency for any type of builds altogether, and we can remove simplejson as well. This takes us one more step closer to removing python as well (looks like at that point we'll just have it left for js2c.py and bootstrap.py in the android SDK build. Though I'm not sure how we'll get rid of/convert js2c.py since that came from V8 itself. I suppose we can try and rewrite it to nodejs)

Secondly, it looks like they recommend using the FNV-1a hash function now in place of FNV-1, as it's a pretty minor change to the algorithm but results in much more random hashes and therefore even less chance of collision: http://www.isthe.com/chongo/tech/comp/fnv/#FNV-1a

garymathews:
I think I'll address the SDK gperf dependency in another PR, since this could potentially go into `6_2_X` and remove our users dependency on gperf for all platforms.

The reason I choose FNV-1 instead of FNV-1a is because it has less collision for text based keys. Here's a test of _216,553_ words from an English dictionary:

| HASH | LOWERCASE COLLISIONS | UPPERCASE COLLISIONS |
|:------:|:--------------------:|:--------------------:|
| FNV-1 | 1 (0.00046%) | 9 (0.00416%) |
| FNV-1a | 4 (0.00185%) | 11 (0.00508%) |

But our keys are going to be larger which might change this outcome in favour of FNV-1a?

sgtcoolguy (approved):
LGTM 👍

mukherjee2:
@garymathews is it possible to get the commit built so I can get the artifacts? I'm getting errors when I try to build locally on a Windows machine.

lokeshchdhry:
FR Passed.

1.Built a module (ti.paint) with an SDK with this fix.
2. Used the module in an app & built it for android device.
3. Worked fine.

SDK Ver:        7.0.0.v20171110161817
OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

PR #9273, opened 2017-08-03
[TIMOB-20404]: iOS ProgressBar gets clipped in parent with height Ti.UI.SIZE

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-20404


iphone/Classes/TiUIProgressBar.m
```diff
@@ -67,7 +67,10 @@ -(CGFloat)contentHeightForWidth:(CGFloat)width
 {
 	CGSize fontSize = [self sizeForFont:width];
 	CGSize progressSize = [progress sizeThatFits:fontSize];
-	return fontSize.height + progressSize.height;
+    if (messageLabel == nil) {
+        return fontSize.height + progressSize.height;
+    }
+	return fontSize.height + progressSize.height + 5;
```
hansemannn: Indentation is incorrect. I also feel like this is not the proper way to fix it (by manually setting a constant value of 5 just to unclip it). I need to spend more time reviewing the progress-bar layout-code this week.

hansemannn (changes requested):

iphone/Classes/TiUIProgressBar.m
vijaysingh-axway: @hansemannn,  If you see function -(void)layoutSubviews in same file, there is code snippet -
	float fittingHeight = barSize.height + messageSize.height + 5;

which sets the gap of 5 pixel between message and progress bar. So while calculating height it should also give +5 pixel, if message is there.

hansemannn (approved):
LGTM! 🚀

PR #9277, opened 2017-08-04
[TIMOB-12145] kroll.writeSnapshot method for heap dump

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-12145

```javascript
var path = Ti.Filesystem.tempDirectory.replace('file://', '') +
  '/' + Date.now() + '.heapsnapshot';
if (kroll.writeSnapshot(path)) {
  console.log('Heap snapshot successfully dumped to', path);
} else {
  console.error('Failed to dump heap snapshot to', path);
}
```

### DIFF
--- a/android/runtime/v8/src/native/JSException.h
+++ b/android/runtime/v8/src/native/JSException.h
@@ -12,7 +12,7 @@
 #include <v8.h>
 
 #define THROW(isolate, msg) \
-	isolate->ThrowException(v8::String::NewFromUtf8(isolate, msg))
+	isolate->ThrowException(v8::Exception::Error(v8::String::NewFromUtf8(isolate, msg)))
 
 namespace titanium {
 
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
I assume this is sort of a port of this node native module? https://github.com/bnoordhuis/node-heapdump/blob/master/src/heapdump.cc

It looks like a nice low-level utility to inspecting state when imported into Chrome's dev tools...

Just thinking out loud here:
Are we sure we want this in the core of the SDK off the kroll object? Maybe it could be an android native module like the package above is a node module?

drauggres:
Obsolete.
Steps for taking js heapsnapshot in Ti SDK 7:

1. run `ti build `with `--debug-host localhost:8989`
2. open  `chrome-devtools://` link in chrome/chromium
3. in `Memory` tab click `Take snapshot`

I think thats enough for debugging purposes.

PR #9285, opened 2017-08-08
[TIMOB-23495] iOS: Add Ti.Filesystem.File "parent" property (parity)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23495


tests/Resources/ti.filesystem.file.test.js
```diff
@@ -0,0 +1,505 @@
...
+	it('symbolicLink', function () {
+		var file = Ti.Filesystem.getFile('app.js');
+		should(file).have.a.readOnlyProperty('symbolicLink').which.is.a.Boolean;
+	});
+
+	it('parent', function () {
+		var file = Ti.Filesystem.getFile('app.js');
+		// parent may be null if at root?
+		//should(file.parent).be.ok; // not null or undefined. should(file).not.be.null causes a stack overflow somehow.
+		should(file).have.a.readOnlyProperty('parent');
```
sgtcoolguy: I'd assume we'd also want to verify the value is another Ti.Filesystem.File. Also that at some level you can't get a parent object and get null back if you try to go up a level above the app root or whatever the directory tree sandbox happens to be.

apidoc/Titanium/Filesystem/File.yml
```diff
@@ -176,19 +176,6 @@ methods:
         if this object doesn't identify a directory.
     returns: 
         type: Array<String>
-  - name: getParent
```
sgtcoolguy: I assume we actually need to deprecate this and not just remove it? Can we have a parent property that holds a Ti.Filesystem.File return value, but *also* have a getParent() method that returns the path of the parent as a String?

This is a breaking change, so we need to deprecate for one major version first... I assume there are apps out there relying on this getter existing and returning a string.

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -29,45 +29,44 @@ -(id)initWithFile:(NSString*)path_
 	return self;
 }
 
--(void)dealloc
+- (void)dealloc
 {
 	RELEASE_TO_NIL(fm);
     RELEASE_TO_NIL(path);
```
vijaysingh-axway: @hansemannn  Can you Fix Indentation. I think you need to lint the file.

apidoc/Titanium/Filesystem/File.yml
hansemannn: I only removed it because the getter is generated automatically by using `- (NSString *)parent`. But I assume we could match Android's behavior here whereby `getParent` returns a file and `parent` a string. @infosia how does Windows behave here? Although I don't like that the getter of a property returns something different (oddly, I'd rather match the most-useful behavior). Maybe deprecate the difference on Android?

iphone/Classes/TiFilesystemFileProxy.m
```diff
@@ -406,11 +406,17 @@ - (id)extension:(id)args
...
-- (id)getParent:(id)args
+- (NSString *)getParent:(id)args
 {
+  DEPRECATED_REPLACED(@"Filesystem.File.getParent()", @"7.0.0", @"Filesystem.File.parent");
   return [path stringByDeletingLastPathComponent];
 }
 
+- (TiFilesystemFileProxy *)parent
+{
+  return [[[TiFilesystemFileProxy alloc] initWithFile:[path stringByDeletingLastPathComponent]] autorelease];
```
hansemannn: @sgtcoolguy Sorry to bug you on this, but to confirm:

1. Deprecate `getParent()` for parity with other platforms
2. Introduce `parent` property that works the same across all platforms
3. In 8.0.0+, remove the manual `getParent` accessor, so `getParent()` will use the computed getter that matches all platforms without a deprecation.

tests/Resources/ti.filesystem.file.test.js
```diff
@@ -0,0 +1,555 @@
...
+	});
+
+	// Check if parent exists and returns File
+	it('parent', function () {
+		var file = Ti.Filesystem.getFile('app.js');
+
+		should(file).have.a.readOnlyProperty('parent');
+		should(file.parent).not.be.undefined;
+		should(file.getParent).not.be.undefined;
+		should(file.getParent.be.a.Function;
```
vijaysingh-axway: @hansemannn  ')' is missing here. Please update.

hansemannn:
@vijaysingh-axway Updated

vijaysingh-axway (approved):
Cr and FT passed.

PR #9286, opened 2017-08-08
Use latest eslint settings

sgtcoolguy:
**Description:**
This ensures we lint our code with the new eslint settings from appcelerator/eslint-config-axway.
This also involved updating our JS code to adhere to the new lint settings, which was a significant number of changes. In moving over the apidoc stuff, I actually broke the code by converting from `==` and `!=` to `===` and `!==`, as a result of assumptions about comparing to null when getting implicit `undefined` return values from functions. Hopefully This won't happen on the scons/cli code.
Additionally, I moved doc validation directly into our Gruntfile as part of our listing, not as a Jenkins-only separate step.


android/cli/commands/_build.js
```diff
@@ -562,7 +577,7 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 
 										// print list of available avds
 										if (results.length && !cli.argv.prompt) {
-											logger.log(__('Available Emulators:'))
+											logger.log(__('Available Emulators:'));
```
cb1kenobi: Nice catch!

cb1kenobi (approved):
There's a ton in the PR to CR, but it looks good to me.

build/scons-cleanbuild.js
```diff
@@ -11,17 +9,13 @@ var exec = require('child_process').exec,
...
 	DIST_DIR = path.join(ROOT_DIR, 'dist'),
-	ALL_OSES = ['win32', 'linux', 'osx'],
-	ALL_PLATFORMS = ['ios', 'android', 'mobileweb', 'windows'],
 	OS_TO_PLATFORMS = {
-		'win32': ['android', 'mobileweb', 'windows'],
-		'osx': ['android', 'ios', 'mobileweb'],
-		'linux': ['android', 'mobileweb']
+		'win32': [ 'android', 'mobileweb', 'windows' ],
+		'osx': [ 'android', 'ios', 'mobileweb' ],
+		'linux': [ 'android', 'mobileweb' ]
```
hansemannn: Are these mobileweb references removed from master? Guess you can just pull?

build/scons-package.js
```diff
@@ -9,17 +8,15 @@ var os = require('os'),
...
-	ALL_PLATFORMS = ['ios', 'android', 'mobileweb', 'windows'],
+	ALL_OSES = [ 'win32', 'linux', 'osx' ],
+	ALL_PLATFORMS = [ 'ios', 'android', 'mobileweb', 'windows' ],
 	OS_TO_PLATFORMS = {
-		'win32': ['android', 'mobileweb', 'windows'],
-		'osx': ['android', 'ios', 'mobileweb'],
-		'linux': ['android', 'mobileweb']
+		'win32': [ 'android', 'mobileweb', 'windows' ],
+		'osx': [ 'android', 'ios', 'mobileweb' ],
+		'linux': [ 'android', 'mobileweb' ]
```
hansemannn: Same here - mobileweb references?

cli/commands/create.js
```diff
@@ -4,26 +4,21 @@
  * copying template files.
  *
  * @copyright
- * Copyright (c) 2012-2015 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2012-2017 by Appcelerator, Inc. All Rights Reserved.
```
hansemannn: Facebook and some other OSS projects tend to use `-Present` these days, makes it easier to curate it. But from what I talked with @ingo a few months ago, from a legal standpoint we would even only need the `2012`.

PR #9287, opened 2017-08-08
[TIMOB-24806] Update .clang-format, lint complete iOS-source

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24806


sgtcoolguy:
So the most important thing to me beyond doing the actual choosing of the format and formatting all our code - is hooking this up to be enforced in our builds. I'd love to hook iOS formatting checks into the Gruntfile.js to run as part of npm test (in our lint task list, we currently just run 'appsJs').

clang-format doesn't make this particularly easy. But you can invoke it in a way that you can output any formatting changes it'd apply as an xml file. So I think we could write a grunt task that would loop over the files and run clang-format in a way that output an xml file. If the changes are basically empty, the file is good, if not, then we fail/mark the file as out of format.

It seems to me that since we already have grunt-clang-format that that package would be the best place to add this functionality. That package right now just can do the formatting, but I'd think we'd want enforcing format as well/more.

sgtcoolguy:
@hansemannn I'm ok with you merging this (once conflicts are fixed), and then we can open a new PR for enforcing the formatting via grunt. I already began work on this as an offshoot branch from my eslint PR #9286 

Here's the changes to enforce formatting: https://github.com/appcelerator/titanium_mobile/commit/7630ad2feb241cc9dc7eb8f6d4cc23261c11ec18

sgtcoolguy (approved):
Please fix merge conflicts and then we can merge

hansemannn:
@sgtcoolguy It seems impossible to properly rebase this, there are too many concurring commits. I will rebase it to the one where the .clang-format was added, pull from master and re-format again.

hansemannn:
@sgtcoolguy No merge conflicts! Go go go!

PR #9288, opened 2017-08-08
[TIMOB-9111] iOS: Remove Base64Transcoder dependency, fix Base64-parity

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-9111


sgtcoolguy:
Looks like this now fails the `base64#decode(Ti.Blob)` unit test on iOS.

hansemannn:
@vijaysingh-axway Save to review. @sgtcoolguy I didn't add new unit-tests, because the old ones should persist as they are.

vijaysingh-axway (approved):
Cr and FT passed.

mukherjee2 (approved):
Node Version: 8.9.1
NPM Version: 5.5.1
Mac OS: 10.13
Appc CLI: 7.0.0-master.13
Appc CLI NPM: 4.2.11-2
Titanium SDK version: 7.0.0
Xcode 9.1/iOS 11.1 iPhone 7 Plus
Passed FR. I first reproduced with an affected version. I then tried it with the fix version and did not see "NEWLINE" in the alert, as expected.

mukherjee2:
@hansemannn can you look into the unit test failures?

PR #9292, opened 2017-08-09
[TIMOB-25003] iOS: Deprecate Ti.UI.iOS.Toolbar in favor of cross-platform Ti.UI.Toolbar

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25003

Also clang-formatted the new source and made the deprecated API reuse the new one by subclassing it. Test-cases in JIRA:


iphone/Classes/TiUIToolbar.m
```diff
@@ -0,0 +1,183 @@
+/**
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2009-2010 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+#ifdef USE_TI_UIIOSTOOLBAR
```
vijaysingh-axway: It should be -
+#ifdef USE_TI_UITOOLBAR

iphone/Classes/TiUIToolbar.h
```diff
@@ -1,24 +1,25 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2009-2010 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-2017 by Appcelerator, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 
 #ifdef USE_TI_UIIOSTOOLBAR
```
vijaysingh-axway: It should be -
+#ifdef USE_TI_UITOOLBAR

apidoc/Titanium/UI/Toolbar.yml
```diff
@@ -32,20 +55,36 @@ properties:
...
+    platforms: [iphone, ipad]
     
   - name: items
     summary: An array of buttons (or other widgets) contained in the toolbar.
     type: Array<Titanium.UI.View>
+    platforms: [iphone, ipad]
     
   - name: borderTop
-    summary: If `true`, a border is drawn on the top of the toolbar.
+    summary: If `true`, a border is drawn on the top of the toolbar. This property is ignored on iOS 7 and above.
```
vijaysingh-axway: As this property do not have effect in iOS 7 and above and we are supporting iOS 8 and above, deprecate it .

apidoc/Titanium/UI/Toolbar.yml
```diff
@@ -32,20 +55,36 @@ properties:
...
     type: Boolean
     default: true
-    
+    platforms: [iphone, ipad]
+   
   - name: borderBottom
-    summary: If `true`, a border is drawn on the bottom of the toolbar.
+    summary: If `true`, a border is drawn on the bottom of the toolbar. This property is ignored on iOS 7 and above.
```
vijaysingh-axway: Same as above.

hansemannn: Actually, we should use `#if defined(USE_TI_UIIOSTOOLBAR) || defined(USE_TI_UITOOLBAR)` all around the place, since Ti.UI.iOS.Toolbar will invoke TiUIToolbarProxy internally as well.

hansemannn:
@vijaysingh-axway Updated PR!

iphone/Classes/TiUIToolbar.h
hansemannn: Actually, we should use `#if defined(USE_TI_UIIOSTOOLBAR) || defined(USE_TI_UITOOLBAR)` all around the place, since Ti.UI.iOS.Toolbar will invoke TiUIToolbarProxy internally as well.

apidoc/Titanium/UI/Toolbar.yml
hansemannn: Removed all border-related docs, since the property does not even exist in the core anymore. Good catch, keeps it clean!

vijaysingh-axway (approved):
CR passed.

PR #9294, opened 2017-08-09
[TIMOB-23114] iOS: Fix semi-bold and bold fonts-weights

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23114

### DIFF
--- a/iphone/Classes/WebFont.m
+++ b/iphone/Classes/WebFont.m
@@ -151,9 +151,9 @@
                 }
             }
             if (font == nil) {
-                //NO valid family specified. Just check for characteristics. Semi bold is ignored here.
+                //NO valid family specified. Just check for characteristics.
                 if (self.isBoldWeight) {
-                    UIFont* theFont = [UIFont boldSystemFontOfSize:self.size];
+                    UIFont* theFont = [UIFont systemFontOfSize:self.size weight:UIFontWeightBold];
                     if (self.isItalicStyle) {
                         NSString* fontFamily = [theFont familyName];
                         NSArray* fontNames = [UIFont fontNamesForFamilyName:fontFamily];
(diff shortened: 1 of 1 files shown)


ewieberappc (approved):
FR Passed, using:
MacOS 10.12.6 (16G24b)
Studio 4.9.0.201705302345
Ti SDK 7.0.0.v20170809111003
Appc NPM 4.2.9
Appc CLI 6.2.3
Alloy 1.9.13
Xcode 8.3.3 (8E3004b)

Font weight now includes semibold. Bold and italics are working as expected. Semibold fonts can also be used by their postscript name.

iphone/Classes/WebFont.m
```diff
@@ -151,9 +151,9 @@ -(UIFont *) font
                 }
             }
             if (font == nil) {
-                //NO valid family specified. Just check for characteristics. Semi bold is ignored here.
+                //NO valid family specified. Just check for characteristics.
                 if (self.isBoldWeight) {
-                    UIFont* theFont = [UIFont boldSystemFontOfSize:self.size];
+                    UIFont* theFont = [UIFont systemFontOfSize:self.size weight:UIFontWeightBold];
```
vijaysingh-axway: This API is available in iOS 8.2 and above only. Need to put check for same.
+ (UIFont *)systemFontOfSize:(CGFloat)fontSize weight:(CGFloat)weight NS_AVAILABLE_IOS(8_2);

iphone/Classes/WebFont.m
```diff
@@ -180,14 +180,16 @@ -(UIFont *) font
                     }
                 } else if (self.isItalicStyle) {
                     font = [[UIFont italicSystemFontOfSize:self.size] retain];
+                } else if (self.isSemiboldWeight) {
+                    font = [[UIFont systemFontOfSize:self.size weight:UIFontWeightSemibold] retain];
```
vijaysingh-axway: Same as above.

iphone/Classes/WebFont.m
hansemannn: I didn't on purpose, since we target 8.2 and later since SDK 6.0.0.

hansemannn: Actually, we do 8.0 I guess. Maybe we can find out the raw-value and pass that one instead - for backwards-compatibility?

eric34: @hansemannn  We could get it in before 7.0.0 but I am ok with pushing it out of 6.2.0 with a decent comment.

hansemannn: Let's do it in 7.0.0, I am trying not to risk anything here.

hansemannn:
Fixed as part of #9308

PR #9228, opened 2017-07-19
[TIMOB-24683] Update to windowslib 0.5.2

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24683

This takes over from #9203

Includes the following
- appcelerator/windowslib/pull/71 - [TIMOB-24683](https://jira.appcelerator.org/browse/TIMOB-24683) Only run winappdeploycmd when it exists in our info detection and  [TIMOB-24956](https://jira.appcelerator.org/browse/TIMOB-24956) Fix Jenkins failure where it was unable to unzip the app package and copy a file when grabbing the apps product guid
- appcelerator/windowslib/pull/71 - [TIMOB-24922](https://jira.appcelerator.org/browse/TIMOB-24922) Fix cert generation on VS2017 
- appcelerator/windowslib/pull/70 - [TIMOB-24668](https://jira.appcelerator.org/browse/TIMOB-24668) Update VS2017 detection logic to prevent error with VS2015
- Try to make wptool.connect poll emulator status - no ticket?
- appcelerator/windowslib/pull/67 - [TIMOB-24422](https://jira.appcelerator.org/browse/TIMOB-24422) Windows: Fails to install dependencies for app on Windows 10 Phone Emulator

PR #9298, opened 2017-08-10
[TIMOB-24683][6_2_X] Update to windowslib 0.5.3

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24683

Update to windowslib@0.5.3 to include bugfix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3883,9 +3883,9 @@
       "integrity": "sha1-VDjNLqk7IC76Ohn+iIeu58lPnJ0="
     },
     "windowslib": {
-      "version": "0.5.2",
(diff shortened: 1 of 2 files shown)


infosia:
@ewanharris  Looking at previous windowslib update #9228, don't we need to push actual updates under `node_modules/windowslib` too?

ewanharris:
@infosia The 6_2_X and master branches no longer have the node_modules dir checked in, so when we do a module update we just need to run `npm update <package>`, ensure that the versions in package-lock.json and package.json are as expected and then all the node_modules are downloaded in CI before publishing the SDK.

Chris Williams wrote a little bit about the reasoning in #9048

infosia:
ok sounds good to me.

PR #9299, opened 2017-08-10
[TIMOB-6392] Android fix: Ti.UI.Button text values with too many dis…

maggieaxway:
…playable characters draw improperly

[TIMOB-6392]- Android fix: Ti.UI.Button text values with too many
displayable characters draw improperly

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-6392

Test case:

````js


window.add(Ti.UI.createButton(
{
	title: "Single Line Button",
	top: "5%",
}));
window.add(Ti.UI.createButton(
{
	title: "Multiline\nButton",
	top: "5%",
}));
window.add(Ti.UI.createButton(
{
	title: "Single Line Truncated Button",
	top: "5%",
	width: "50%",
}));
window.add(Ti.UI.createButton(
{
	title: "Multiline Button\nThis is the next very long line.",
	top: "5%",
	width: "50%",
}));
window.add(Ti.UI.createButton(
{
	title: "Very long multiline button with word wrapping.",
	top: "5%",
	width: "50%",
}));
window.add(Ti.UI.createButton(
{
	title: "Very long Multiline button with word wrapping.",
	top: "5%",
	width: "50%",
	height: "5%",
}));
window.open();

### CODE REVIEW

**Comment by jquick-axway** [comment]
@maggieaxway, I did a quick test on iOS and it turn out we don't support multiline buttons on that platform. That is, there is no line wrapping and it ignores newline '\n' characters. It's always forced to be single-line. iOS also always applies ellipsis in the middle of the text as well.

So, for parity with iOS, let's do the following in the TiUIButton constructor:
- Force to single line via `setMaxLines(1)`.
- Use `TextUtils.TruncateAt.MIDDLE` ellipsis mode instead. (Matches iOS behavior.)

This also means you don't need to do the above dynamically anymore. So, you can get rid of the `onGlobalLayout()` listener. This is also a breaking change and should only go into Titanium 7.0 (don't back-port it), meaning that multiline button support will be lost. But it's better to add new button properties to both Android and iOS to enable such a feature to make it portable.

File: android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java [hunk 1]
```diff
@@ -84,17 +83,8 @@ public void processProperties(KrollDict d)
...
-				@Override
-				public void onGlobalLayout() {
-					if(btn.getLineCount()>1 && btn.getMeasuredHeight() > btn.getLineHeight()){
-						btn.setEllipsize(TextUtils.TruncateAt.END);
-						btn.setMaxLines(1);
-					}
-				}
-			});
+			btn.setEllipsize(TextUtils.TruncateAt.MIDDLE);
+			btn.setMaxLines(1);
```
jquick-axway: Can you move these 2 lines into the TiUIButton constructor please?
(Before the `setNativeView(btn)` call.)

This way these setting will always be applied. This is in case the "title" property is not set upon button creation and is set afterwards instead.

Other than that, yeah, it should be this simple. :)

jquick-axway (changes requested):

jquick-axway (approved):
CR: Pass

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
maggieaxway: Yes you're right! I didn't notice it. Appreciate your caution!

lokeshchdhry:
FR Passed.

For longer button text on android the text will be shortened using ellipses like IOS.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

lokeshchdhry:
@maggieaxway , Jenkins is asking for unit tests before merge. Can you please add it.

jquick-axway:
I don't think there is a good way to automate testing of this particular case. You have to eyeball the text and make sure it's showing ellipsis.

maggieaxway:
@lokeshchdhry There's no crash before or after the changes, it just meant to display properly. So I agree with Josh it doesn't make sense to make auto test.

lokeshchdhry:
@sgtcoolguy , Can you please merge this as according to @maggieaxway no unit tests can be added. Thanks!

PR #9303, opened 2017-08-11
[TIMOB-24351](6_2_X):iOS Support "editActions" in listview search-results

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24351

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -57,7 +57,6 @@ static TiViewProxy * FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoin
     BOOL pullActive;
     CGPoint tapPoint;
     BOOL editing;
-    BOOL editingOnSearchResult;
     BOOL pruneSections;
     
     BOOL caseInsensitiveSearch;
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIListView.m
```diff
@@ -531,22 +531,20 @@ -(void)buildResultsForSearchText
...
-                            id theTitle = [sectionTitles objectAtIndex:[sectionIndices indexOfObject:theIndex]];
-                            if (filteredTitles == nil) {
-                                filteredTitles = [[NSMutableArray alloc] init];
-                            }
-                            if (filteredIndices == nil) {
-                                filteredIndices = [[NSMutableArray alloc] init];
-                            }
-                            [filteredTitles addObject:theTitle];
-                            [filteredIndices addObject:NUMUINTEGER([_searchResults count] -1)];
+                [_searchResults addObject:thisSection];
```
hansemannn: What happened to the `[thisSection count]` check?

hansemannn (changes requested):

iphone/Classes/TiUIListView.m
vijaysingh-axway: In current implementation we are giving option of edit while searching. Suppose there are 2 sections  (section1 , section2) in list view and we started search. Suppose we searched for an item of section2 and want to move it in section1. Suppose the keywords which we searched is only available in section2 and not in section1. So it should show section1 with 0 rows and section2 with relevant search results, so that item from section2 to section1 can be moved. Mentioned condition get removed to add section1 with 0 rows, in _searchResults.

In previous implementation, if no result found for searched keyword in section1, section1 was deleted. As edit option was not supported for search result previously, so it was not affecting anything. 
 if ([thisSection count] > 0) { ...}
Above condition was doing the stuff for removing the section if search result was not found in particular section.

See the 3rd use case ( Search with search bar and listview), which is having move option.

hansemannn: So we can still move rows while searching?

vijaysingh-axway: Yes. Now we can move rows while searching.

hansemannn (approved):

ewieberappc:
I encounter a crash when trying to edit the ListView when search results are displayed. You can see this if you use the third sample code provided in the ticket. Steps are:
1. Perform a search (I search for "c" because it yields multiple results)
2. Click the edit button at the top of the screen
3. Encounter crash

ewieberappc (approved):
FR Passed.
Able to perform edit/delete actions on search results for ListViews and TableViews using searchbars

PR #9314, opened 2017-08-14
[TIMOB-25029] Removed generation of legacy launch images for iOS 6.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25029

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -66,7 +66,7 @@ function iOSBuilder() {
 	// list of available build-targets
 	this.targets = ['simulator', 'device', 'dist-appstore', 'dist-adhoc'];
 
-	// object of device families to map the --device-family parameter to the 
+	// object of device families to map the --device-family parameter to the
 	// native TARGETED_DEVICE_FAMILY build-setting
 	this.deviceFamilies = {
 		iphone: '1',
@@ -78,8 +78,8 @@ function iOSBuilder() {
 	// device-family set by the --device-family parameter
 	this.deviceFamily = null;
 
-	// blacklisted files and directories that throw an error when used and will 
-	// lead to a rejection when submitted 
+	// blacklisted files and directories that throw an error when used and will
+	// lead to a rejection when submitted
 	this.blacklistDirectories = [
 		'contents',
 		'resources',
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -5400,22 +5400,16 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
...
 					'Default-568h@2x.png':           { idiom: 'iphone', extent: 'full-screen', minSysVer: '7.0', orientation: 'portrait', width: 640, height: 1136, scale: 2, subtype: 'retina4' },
-					// iPhone Portrait - iOS 8,9 - Retina HD 4.7 (750x1334) iPhone 6
-					'Default-667h@2x.png':           { idiom: 'iphone', extent: 'full-screen', minSysVer: '8.0', orientation: 'portrait', width: 750, height: 1334, scale: 2, subtype: '667h' },
 
 					// iPad Landscape - iOS 7-9 - 1x (1024x768)
 					'Default-Landscape.png':         { idiom: 'ipad',   extent: 'full-screen', minSysVer: '7.0', orientation: 'landscape', width: 1024, height: 768, scale: 1 },
 					// iPad Landscape - iOS 7-9 - 2x (2048x1536)
 					'Default-Landscape@2x.png':      { idiom: 'ipad',   extent: 'full-screen', minSysVer: '7.0', orientation: 'landscape', width: 2048, height: 1536, scale: 2 },
-					// iPhone Landscape - iOS 8,9 - Retina HD 5.5 (2208x1242)
-					'Default-Landscape-736h@3x.png': { idiom: 'iphone', extent: 'full-screen', minSysVer: '8.0', orientation: 'landscape', width: 2208, height: 1242, scale: 3, subtype: '736h' },
```
hansemannn: These ones are for iOS 8/9 and are used on the iPhone 6 / iPhone 6S / iPhone 7. Did you test the changes? I highly doubt these are the correct ones to remove.

hansemannn (changes requested):

iphone/cli/commands/_build.js
cb1kenobi: Yes, I tested it. I know. Doesn't make sense, but those 3 are the culprit.

hansemannn: Has to be an Xcode issue then. Those are the most recent launch-screens introduced by Apple, all newer devices would not have a proper launch-screen and scale-up the other ones. Need to dive more into this.

hansemannn:
PR **rejected**, it removes required launch-images for the iPhone 6 / 7 and iPhone 6 Plus / 7 Plus. The initial warning from the linked ticket has to be an Xcode false-positive and will be tracked with Apple.

PR #9317, opened 2017-08-15
[CLI-649] direct users to use appc ti config if run from appc

feons:
https://jira.appcelerator.org/browse/CLI-649


sgtcoolguy (approved):
LGTM 👍

PR #9322, opened 2017-08-16
[TIMOB-25176] Fix node 4 compatibility

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-25176

This PR add node 4 compatibility back, I only added 'use strict' to the files we distribute, so files in apidoc/, build/ and Gruntfile.js will still break on node 4. 

If we want that it's just a matter of me running [the relevant jscodemod](https://github.com/cpojer/js-codemod/blob/master/transforms/use-strict.js) again so no biggy, just let me know

### DIFF
--- a/cli/lib/info.js
+++ b/cli/lib/info.js
@@ -5,8 +5,9 @@
  * See the LICENSE file for more information.
  */
 
-const
-	appc = require('node-appc'),
+'use strict';
+
+const appc = require('node-appc'),
 	fs = require('fs'),
 	path = require('path'),
 	genymotion = require('node-titanium-sdk/lib/emulators/genymotion'),
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -14,6 +14,8 @@
  * Please see the LICENSE included with this distribution for details.
  */
 
+'use strict';
+
 const ADB = require('node-titanium-sdk/lib/adb'),
 	AdmZip = require('adm-zip'),
 	android = require('node-titanium-sdk/lib/android'),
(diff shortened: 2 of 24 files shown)


cli/lib/info.js
```diff
@@ -5,8 +5,9 @@
  * See the LICENSE file for more information.
  */
 
-const
-	appc = require('node-appc'),
```
hansemannn: @sgtcoolguy Wondering if that format is expected with eslint.

cb1kenobi: Sorry, this is my super old coding style. It's bad. Nowadays I specify `const` or `let` per variable.

ewanharris: +1 On not declaring each variable deceleration being separate, as always [there's a rule for that](http://eslint.org/docs/rules/one-var) but not sure whether all teams would buy in to having that in the config. There's also a rule for [strict mode](http://eslint.org/docs/rules/strict) which imo if our min ver will be node 4 should be added to the config (again doubt all times would buy in though)

sgtcoolguy: So I think we can still do per-project overrides, or additional rules on top of the common config. I'd agree that the `"strict": "safe"` rule is probably a good one to turn on while Node 4 is our min supported Node.

Personally, I like having multiple `const/`var`/`let` declarations combined.

sgtcoolguy (approved):

sgtcoolguy:
I merged this, added the eslint rule to our .eslintrc - as well as tweaked the `parserOptions` to specify `sourceType: script` and `ecmaVersion: 2015`, then added `'use strict'` to the remaining files that eslint flagged: https://github.com/appcelerator/titanium_mobile/commit/0bef6bfbddee5efc7ee2c5268271a73e15d01610

PR #9325, opened 2017-08-16
[TIMOB-17450] Android: Ti.Calendar.Event should expose the Attendees (Parity with iOS)

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17450

**Description:**
Add AttendeesProxy and add it as a result for getAttendees() method.

**Test case:**

```
var win = Ti.UI.createWindow({backgroundColor:'gray'});
var pickerCalendar = Ti.UI.createPicker({
    top: 50,
});
var pickerEvents = Ti.UI.createPicker({
    top: 100
});
var buttonGetAttendeesForEvent = Ti.UI.createButton({
    top: 150,
    title: 'Get Attendees'
});
var events = [];
var selectedEventIndex;
buttonGetAttendeesForEvent.addEventListener('click', function () {
    Ti.API.info('----------------------');
    Ti.API.info(events[selectedEventIndex].getAttendees());
    Ti.API.info('----------------------');
    events[selectedEventIndex].getAttendees().forEach(function(attendee){
        Ti.API.info('----------------------');
        Ti.API.info(JSON.stringify(attendee));
        Ti.API.info('----------------------');
    });
});

var listEvents = function (calendarSource) {
    pickerEvents = Ti.UI.createPicker({
        top: 100
    });
    var i=0;
    var eventsRows = [];
    events = [];
    Ti.Calendar.getCalendarById(calendarSource).getEventsInYear(2017).forEach(function(event){
        //create row for each calendar
        events[i] = event;
        eventsRows[i++]=Ti.UI.createPickerRow({title:event.getTitle()});
    });
    if (eventsRows.length > 0) {
        pickerEvents.add(eventsRows);
        pickerEvents.addEventListener('change', function(e){
            selectedEventIndex = e.rowIndex;
        })
        win.add(pickerEvents);
        selectedEventIndex = 0;
    }
}

var getCalendars = function() {
    var calendarRows = [];
    var i=0;
    Ti.Calendar.getAllCalendars().forEach(function(calendar){
        //create row for each calendar
        calendarRows[i++]=Ti.UI.createPickerRow({title:calendar.getName(), calendar: calendar});
    });
    if (calendarRows.length >0 ) {
        pickerCalendar.add(calendarRows);
            pickerCalendar.addEventListener('change', function(e){
                win.remove(pickerEvents);
                listEvents(e.rowIndex+1);
            })
            win.add(pickerCalendar);
            listEvents(1);
               win.add(buttonGetAttendeesForEvent);
    }
}

win.addEventListener('open', function(){
    if (Ti.Calendar.hasCalendarPermissions()) {
        getCalendars();
    } else {
        Ti.Calendar.requestCalendarPermissions(function(){
            getCalendars();
        })
    }
});

win.open();

```

**Note:** Device should have a valid calendar with an event with at least one attendee.
   Also these permissions need to be added in tiapp.xml:
```
<uses-permission android:name="android.permission.WRITE_CALENDAR"/>
<uses-permission android:name="android.permission.READ_CALENDAR"/>
```


android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
```diff
@@ -243,6 +244,36 @@ public static EventProxy createEvent(CalendarProxy calendar, KrollDict data)
...
+			AttendeeProxy proxyForRow = new AttendeeProxy(cursor.getString(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_EMAIL)),
+															cursor.getString(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_NAME)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_TYPE)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_STATUS)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_RELATIONSHIP))
+					);
+			//add the proxy to the result array
+			result[index++] = proxyForRow;
+		}
+			return result;
```
jquick-axway: The `return result;` line is indented 1 tab too far.

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
```diff
@@ -243,6 +244,36 @@ public static EventProxy createEvent(CalendarProxy calendar, KrollDict data)
...
+				CalendarContract.Attendees.ATTENDEE_NAME,
+				CalendarContract.Attendees.ATTENDEE_EMAIL,
+				CalendarContract.Attendees.ATTENDEE_TYPE,
+				CalendarContract.Attendees.ATTENDEE_RELATIONSHIP,
+				CalendarContract.Attendees.ATTENDEE_STATUS
+		};
+		final String query = "(" + CalendarContract.Attendees.EVENT_ID + " = ?)";
+		final String[] args = new String[]{id};
+		ContentResolver contentResolver = TiApplication.getInstance().getContentResolver();
+		final Cursor cursor = contentResolver.query(CalendarContract.Attendees.CONTENT_URI, attendeeProjection, query, args, null);
```
jquick-axway: The `ContentResolver.query()` method can return null. We need to do a null check on `cursor` before using it below.

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
```diff
@@ -243,6 +244,36 @@ public static EventProxy createEvent(CalendarProxy calendar, KrollDict data)
...
+		int index = 0;
+		AttendeeProxy[] result = new AttendeeProxy[cursor.getCount()];
+		while (cursor.moveToNext()) {
+			//create a proxy instance
+			AttendeeProxy proxyForRow = new AttendeeProxy(cursor.getString(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_EMAIL)),
+															cursor.getString(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_NAME)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_TYPE)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_STATUS)),
+															cursor.getInt(cursor.getColumnIndex(CalendarContract.Attendees.ATTENDEE_RELATIONSHIP))
```
jquick-axway: The `cursor.getColumnIndex()` call can return `-1` which would trigger an out-of-bounds exception when passed through to the cursor getter methods. Also, the cursor getter methods can throw an exception if the column type does not match the type requested (ie: it stores a string when we're requesting an int).

To be safe, we should make sure the column index is valid before using it. And wrap each cursor getter in separate try/catch blocks. We can default fields not found to constants such as NONE.

Also, should we define our own "UNKNOWN" constant? The idea is to future-proof this code for when Google adds new constants that our internal code doesn't support yet.

jquick-axway: I looked up our docs. We do define UNKNOWN types in iOS. We need to do the same on Android for types we don't recognize as well. Perhaps that should be determined in the `AttendeeProxy`? I'll leave that up to you.
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
jquick-axway: Also, we need to return an empty array if we've failed to find any attendees or if an error occurs. This is to match iOS' behavior as documented here...
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar.Event-property-attendees

jquick-axway (changes requested):

ypbnv:
@jquick-axway Updated.
Can we test the guarding for the columns' index and value type somehow, because I was not able on a few devices here?
Currently I have added the UNKNOWN type, status and relationship with the same value, which should do, but we can change it in case we need it to differ.

jquick-axway:
> Can we test the guarding for the columns' index and value type somehow?

@ypbnv, you would have to use a 3rd party contacts app instead of Google's, because they don't have to follow the rules. This is something we have to worry about if a Titanium app is distributed via a different Android app store such as Amazon. For example, an Amazon device will have an Amazon made contacts app instead of the one made by Google.
(From looking at your code though, it looks like it'll handle it fine.)

jquick-axway:
CR: Pass

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

The attendees info for the event can be successfully fetched like `"relationship`, `type`, `status` & `email`.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.3
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.13
Node Ver:       6.10.1
Java Ver:       1.8.0_101
                      ⇨ MotoG2 --- Android 4.4.4

PR #9327, opened 2017-08-17
Maintain fixed versions for all production deps

ewanharris:
**Making this as a prompt to remind me to bring this up in standup so no ticket right now**

This PR changes the production deps to all be fixed versions. I'm proposing this because since npm@5.1.0 when using semver ranges in the package.json if a new version is published the package-lock file will be updated. So, since 2017-08-15 when node-titanium-sdk@0.4.2 was published to npm we've been shipping it in the SDK, despite the package-lock stating 0.3.2, and we only updated the package-lock to this version yesterday in 8f0fd94f5762a2dc7c207eedeabee238c227ba66

If we decide to stick to semver ranges we should enforce that the package-lock can't change in CI as the package-lock in the repo imo should represent what we're shipping in the SDK, but that just seems like it would cause annoying failures.

If we decide to stick with this PR and fixed versions we should add something to CI to enforce that production deps have a fixed version, I took a quick look and found https://github.com/whitfin/dep-validate that I think could do that task (sidenote I'm know that username from somewhere 😄 )

Discussion around the package-lock behaviour here
 npm/npm/issues/17979


sgtcoolguy:
I think it's ridiculous that the lock file doesn't actually lock dependencies. Grr. So for now this is the best solution.

cb1kenobi:
I believe this is a non-issue if we used Yarn: `yarn install --frozen-lockfile`.

PR #9346, opened 2017-08-23
[TIMOB-20557] Allow modules to use third party dynamic libraries

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20557

**Optional Description:**
This adds a fully automated framework configuration including support for dynamic frameworks

- [x] Scans for available frameworks and collects some metadata information about them which is used to integrate them in the Xcode project
- [x] No more need to manually add linker flags or framework search paths to a modules .xcconfig
- [x] Bundle a default script for stripping unused architectures for proper app store submissions. A user can provide his own script by droping it as `platform/ios/strip-frameworks.sh` in project.
- [x] Metabase collection is cached on subsequent builds and only rescans changed or added frameworks if necessary
- [x] Assure backwards compatibility (avoid duplicate linking due to manual `-framework` linker flags). Either by restricting the automated configuration of frameworks to a subdirectory, e.g. `Frameworks` so the new mechanism does not interfere with existing modules, or by intelligently skipping the automated Xcode project integration for frameworks that have a .xcconfig (checking only for the `-framework` linker flag should probably be sufficient). **EDIT:** This seems to be a non issue. Xcode has no problems with multiple `-framework` flags and the framework search path already has a dupe check.


iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	/**
+	 * Checks all found dyanmic frameworks if any of them include both device and
+	 * simulator architectures
+	 *
+	 * @return {Boolean} True if device and simulator archs were found, false otherwise
+	 */
+	hasFrameworksWithFatBinary() {
+		let deviceArchitectures = new Set([ 'armv7', 'arm64' ]);
+		let simulatorArchitectures = new Set([ 'i386', 'x86_64' ]);
+		let hasDeviceArchitectures = false;
```
cb1kenobi: `hasDeviceArchitectures` and `hasSimulatorArchitectures` must be either defined inside the for loop below or reset at the beginning of the loop.

Say the first framework has a device architecture, but not a simulator architecture, then `hasDeviceArchitectures` will be `true`, but `hasSimulatorArchitectures` will be `false`. Next it would loop and go to the second framework. Say the second framework does not have a device architecture, but has a simulator architecture. Since `hasDeviceArchitectures` was not reset to `false`, it will be `true` and `hasSimulatorArchitectures` will be `true` and this function will incorrectly return `true`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+		this._frameworks.forEach(frameworkInfo => {
+			metadataObject[frameworkInfo.path] = {
+				name: frameworkInfo.name,
+				path: frameworkInfo.path,
+				type: frameworkInfo.type,
+				architectures: Array.from(frameworkInfo.architectures)
+			};
+		});
+		if (!fs.existsSync(this._outputDirectory)) {
+			wrench.mkdirSyncRecursive(this._outputDirectory);
```
cb1kenobi: Don't use `wrench`. Use `fs-extra`'s `fs.mkdirsSync()`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+			return false;
+		}
+	}
+
+	/**
+	 * Saves the internal matadata object to disk for reuse on subsequent builds
+	 */
+	writeFrameworkMetadata() {
+		let metadataObject = {};
+		this._frameworks.forEach(frameworkInfo => {
```
cb1kenobi: Use a `for-of` loop instead of `forEach()`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+				let frameworkPath = pathAndFilename.substring(0, pathAndFilename.indexOf('.framework') + 10);
+				if (!changedFrameworks.has(frameworkPath)) {
+					this.logger.trace(`Framework at ${frameworkPath} changed, regenerating metadata`);
+					changedFrameworks.add(frameworkPath);
+				}
+			}
+		});
+
+		return this.inspectFrameworks(changedFrameworks)
+			.then(this.writeFrameworkMetadata.bind(this));
```
cb1kenobi: Tip: instead of doing a `bind()`, you can just do:

```
.then(() => this.writeFrameworkMetadata());
```

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	/**
+	 * Loads stored metadata from disk and recreates the {@link FrameworkInfo}
+	 * objects
+	 *
+	 * @return {Boolean} True if the metadata was sucessfully loaded, false if not
+	 */
+	loadFrameworkMetadata() {
+		try {
+			let metadata = JSON.parse(fs.readFileSync(this._metadataPathAndFilename));
+			Object.keys(metadata).forEach(frameworkPath => {
```
cb1kenobi: Use a `for-of` instead of `forEach()`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+		}
+	}
+
+	/**
+	 * Saves the internal matadata object to disk for reuse on subsequent builds
+	 */
+	writeFrameworkMetadata() {
+		let metadataObject = {};
+		this._frameworks.forEach(frameworkInfo => {
+			metadataObject[frameworkInfo.path] = {
```
cb1kenobi: Why not just use a `FrameworkInfo` instance?

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	 * Inspects each framework for their type and supported architectures
+	 *
+	 * @param {Set.<String>} frameworkPaths List of framework paths to inspect
+	 * @return {Promise}
+	 */
+	inspectFrameworks(frameworkPaths) {
+		let metadataPromises = [];
+		let frameworkInspector = new FrameworkInspector(this._logger);
+		for (let frameworkPath of frameworkPaths) {
+			let metadataPromise = frameworkInspector.inspect(frameworkPath).then((frameworkInfo) => {
```
cb1kenobi: Unnecessary parens around `frameworkInfo`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+		if (frameworkInfo.type === FRAMEWORK_TYPE_DYNAMIC) {
+			this.addEmbeddFrameworkBuildPhase(frameworkInfo, fileRefUuid);
+		}
+		this.addFrameworkSearchPath(path.dirname(frameworkInfo.path));
+	}
+
+	/**
+	 * Add the framework as a new file reference to the Xcode project
+	 *
+	 * @param {frameworkInfo} frameworkInfo Framework metadata
```
cb1kenobi: I believe the type should be `FrameworkInfo`. There are a couple other places below with the same issue.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	 * @return {string} Uuid of the created file reference
```
cb1kenobi: All built-in types should be capitalized: `String`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	 */
+	addFrameworkFileReference(frameworkInfo) {
+		let frameworkPackageName = frameworkInfo.name + '.framework';
+		let fileRefUuid = this._builder.generateXcodeUuid();
+		this._xobjs.PBXFileReference[fileRefUuid] = {
+			isa: 'PBXFileReference',
+			lastKnownFileType: 'wrapper.framework',
+			name: '"' + frameworkPackageName + '"',
```
cb1kenobi: Consider using ES6 template strings to make the code more readable:

```
name: `"${frameworkPackageName}"`,
```

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+		});
+	}
+
+	/**
+	 * Detects all available frameworks
+	 *
+	 * @return {Promise}
+	 */
+	detectFrameworks() {
+		return this.findFrameworkPaths().then((frameworkPaths) => {
```
cb1kenobi: Generally, I recommend putting the `.then()` on the next line. This makes it a little more readable, especially if there's a `.catch()`.

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+				if (stdout.indexOf('current ar archive') !== -1) {
+					type = FRAMEWORK_TYPE_STATIC;
+				} else if (stdout.indexOf('dynamically linked shared library') !== -1) {
+					type = FRAMEWORK_TYPE_DYNAMIC;
+				} else {
+					return reject(new Error('Unknown framework type:\n' + stdout));
+				}
+
+				resolve({
+					type: type,
```
cb1kenobi: You can use shorthand properties here.

```js
resolve({ type, architectures });
```

cb1kenobi (changes requested):

iphone/cli/hooks/frameworks.js
janvennemann: Nice, i've been looking for an elegant solution to get rid of `.bind` and this looks much better!

iphone/cli/hooks/frameworks.js
janvennemann: I wanted to, but since `fs-extra` is no direct dependency in our package.json yet i resorted to `wrench`.

iphone/cli/hooks/frameworks.js
janvennemann: Mainly just for cosmetic reasons as I didn't want the values to be stored with the underscore private variable names^^

iphone/cli/hooks/frameworks.js
```diff
@@ -0,0 +1,841 @@
...
+	 * @param {string} frameworksPath Path to scan for frameworks
+	 * @return {Promise}
+	 */
+	scanPathForFrameworks(frameworksPath) {
+		if (!fs.existsSync(frameworksPath)) {
+			return Promise.resolve();
+		}
+
+		return new Promise((resolve, reject) => {
+			fs.readdir(frameworksPath, (err, files) => {
```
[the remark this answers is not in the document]
janvennemann: `fs.existsSync()` is not deprecated, only `fs.exists()` is. Nevertheless, you are right that it is unnecessary.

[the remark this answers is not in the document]
cb1kenobi: Well, I'll be... `fs.existsSync()` used to be deprecated: https://nodejs.org/dist/latest-v5.x/docs/api/fs.html#fs_fs_existssync_path.

janvennemann:
Awesome input man, thanks! Addressed everything and updated the PR.

cb1kenobi (approved):
CR'd and FR'd. Works as expected. Really good code! **APPROVED**!

PR #9091, opened 2017-05-26
[TIMOB-15910] Android: Implement Ti.UI.Android.DrawerLayout

garymathews:
- Based on community module https://github.com/manumaticx/Ti.DrawerLayout
- Implement `Titanium.UI.Android.DrawerLayout`

##### TEST CASE
```JS
var win = Ti.UI.createWindow(),
    leftView = Ti.UI.createView({backgroundColor:'red'}),
    centerView = Ti.UI.createView({backgroundColor:'yellow'}),
    rightView = Ti.UI.createView({backgroundColor:'orange'}),
    drawer = Ti.UI.Android.createDrawerLayout({
        leftView: leftView,
        centerView: centerView,
        rightView: rightView
    }),
    btn = Ti.UI.createButton({title: 'RIGHT'});

btn.addEventListener('click', function() {
    drawer.toggleRight();
});

centerView.add(btn);

win.addEventListener('open', function(){
    var activity = win.getActivity(),
        actionbar = activity.getActionBar();

    if (actionbar) {
        actionbar.displayHomeAsUp = true;
        actionbar.onHomeIconItemSelected = function() {
            drawer.toggleLeft();
        };
    }
});

win.add(drawer);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-15910)

PR #9348, opened 2017-08-23
[6_2_X][TIMOB-15910] Android: Implement Ti.UI.Android.DrawerLayout

garymathews:
- Cherry-pick of https://github.com/appcelerator/titanium_mobile/pull/9091

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-15910)

### DIFF
--- a/android/modules/ui/res/layout/titanium_ui_drawer_layout.xml
+++ b/android/modules/ui/res/layout/titanium_ui_drawer_layout.xml
@@ -1,0 +1,28 @@
+<android.support.v4.widget.DrawerLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <RelativeLayout
+        android:id="@+id/drawer_layout_container"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent">
+
+           <android.support.v7.widget.Toolbar
+               xmlns:app="http://schemas.android.com/apk/res-auto"
+               android:id="@+id/drawer_layout_toolbar"
+               android:visibility="gone"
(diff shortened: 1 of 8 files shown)


lokeshchdhry:
FR Passed.

The drawer layout works as expected. The left, middle & the right views are shown with the appropriate way of unhiding them. Also, different views like imageView, Label, TextView, tableview, listview etc can be successfully added to the left, middle & right drawer views.
The properties & methods seem to work as expected.

Studio Ver: 4.9.1.201707200100
SDK Ver: 6.2.0 Local build
OS Ver: 10.12.3
Xcode Ver: Xcode 8.3.3
Appc NPM: 4.2.9
Appc CLI: 6.2.3
Ti CLI Ver: 5.0.14
Alloy Ver: 1.9.13
Node Ver: 6.10.1
Java Ver: 1.8.0_101
Devices: ⇨ google Nexus 5 --- Android 6.0.1
Emulator: Android 4.4.2

m1ga:
The example code at http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.Android.DrawerLayout is not showing all the code in the correct style. Looking at other yml files I think an extra line before the actual code is missing

hansemannn:
@m1ga Good catch! Fixed on master and 6_2_X, see [here](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/Android/DrawerLayout.yml).

alangumer:
The background color was hardcoded so was not possible to change the color via theme; also the property android:fitsSystemWindows="true" was added to avoid that the status bar overlaps the App bar; and the elevation was set to 0 which looks nicer especially when using navigation tabs.

https://github.com/appcelerator/titanium_mobile/pull/9439

PR #9352, opened 2017-08-24
[TIDOC-2398] generate google closure externs

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-2398

Generated file could be used as replacement for files from https://github.com/navinpeiris/jsca2js


hansemannn:
@jawa9000 What to do here?

drauggres:
I use generated file only for code completion in webstorm, not as actual closure externs.
Now when TypeScript definitions file generated this ticket is probably obsolete.

janvennemann:
@drauggres The TypeScript definition are available by now. Could you explain why you think that makes this ticket is obsolete?

hansemannn:
Thanks @drauggres, closing this ticket for now! If it becomes necessary again, we can reopen this PR without issues :-).

PR #9367, opened 2017-08-29
[TIMOB-26068] Android: Improve Ti.UI.Toolbar default behavior

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26068

**Description:**
Set a default  Ti.UI.FILL width for the toolbar.
Center content.

**Test case:**
```

var toolbar = Ti.UI.createToolbar({
    bottom: 0,
    backgroundColor: 'red'
});

var items = [];

for (var i=0; i < 3; i++) {
        title: 'Button',
    });
    items.push(button);
}

toolbar.setItems(items);
window.add(toolbar);
window.open();
```


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
```diff
@@ -73,7 +77,11 @@ public TiToolbar(TiViewProxy proxy) {
 	public void setItems(TiViewProxy[] proxies) {
 		if (proxies != null) {
 			for (int i = 0; i < proxies.length; i++) {
-				toolbar.addView(convertLayoutParamsForView(proxies[i].getOrCreateView()));
+				View tempView = convertLayoutParamsForView(proxies[i].getOrCreateView());
+				Toolbar.LayoutParams lp = new Toolbar.LayoutParams(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT);
+				lp.gravity = Gravity.CENTER;
```
hansemannn: We may want to expose this as a property in the future.

hansemannn (approved):
Thanks for the PR, my tests should pass now. Approved!

P.S.: @garymathews may now if there is a more sexy way of setting the default-width to fill, but this should work.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
```diff
@@ -63,6 +66,7 @@
 	public TiToolbar(TiViewProxy proxy) {
 		super(proxy);
 		toolbar = new Toolbar(proxy.getActivity());
+		toolbar.setContentInsetsAbsolute(0,0);
```
garymathews: space after comma

ypbnv:
@garymathews, @jquick-axway  We have been talking with @hansemannn whether we should by default have the children be centered in the Toolbar or start from the beginning. As far as I tested Toolbar does not respect it's children's position attributes and acts solely as a container on native Android as well. So the proposition to have a property for begin\center gravity makes sense to me. And for further customization we can have setCustomView implemented.
What do you think about that?

jquick-axway:
@ypbnv, regarding the toolbar's child view layout positions/attributes being ignored, try this. Add our Java `TiCompositeLayout` as a view to the Toolbar with "autoFillsWidth" set to true and with a "horizontal" layout. And instead of adding child Titanium views to the Toolbar directly, add them to this `TiCompositeLayout` view which will serve as the container. I say this because our `TiCompositeLayout` is needed to handle our custom position/layout/padding settings in the child views. If you look at our TiCompositeLayout.onMeasure() method, then you'll see what I'm talking about.
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java

The only down side here is that our horizontal `TiCompositeLayout` view does not support right-to-left locales (ie: Arabic) versus I'm sure Android's native Toolbar would properly reverse the buttons/views.

hansemannn:
Please note that iOS does not respect the margins / general styles as well, so Android probably should do the same there. It's simply because we want to follow the native behavior there - if people want to size the child-items, they would rather go with a manual horizontal layout which lays out the child-views in line. For now, I'd only change the width to Ti.UI.FILL by default and leave the native behavior as it is. But as @ypbnv said, maybe you as native Android devs see this different, I can only speak as a  Titanium developer / end-user who would be fine and happy seeing the native behavior.

ypbnv:
@jquick-axway That will do the trick, yes. And it is the reason I thought about setCustomView() implementation. With it we will have the option to use any view (TiCompositeLayout) and get advantage of it, but not forcing it whenever it is not actually needed.

ypbnv:
@hansemannn , @jquick-axway , @garymathews I have removed the view with CENTER gravity in favor of 
https://github.com/appcelerator/titanium_mobile/pull/9425
That leaves us to decide do we keep setting the default width to Ti.UI.FILL. And if we do - how can that be noted in the docs (the property is inherited)? Just override it ?

jquick-axway:
@ypbnv, Android typically left-aligns the Toolbar's content, right? And likely right-aligns for Arabic locales. I don't know if centering it is a good idea since that's not the native behavior.

Now, defaulting the Toolbar's width to Ti.UI.FILL does make sense to me. We do the same for other views such as TableView, ListView, ScrollView, etc. We don't document that they're filled by default, but I think most devs expect this to be the default behavior anyways.

ypbnv:
I have update the linked ticket.

PR #9373, opened 2017-08-30
[TIMOB-24834] Adjust value resources pattern for module compatibilty

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24834

**Optional Description:**
The pattern used to detect value resources was only working for Android libraries extracted by Hyperloop. With this fix, resources from both Hyperloop and module provided Android libraries are properly handled.


hansemannn (approved):
LGTM. We can merge this if @garymathews agrees.

garymathews (approved):
LGTM 👍

PR #9377, opened 2017-08-30
[TIMOB-25210] Added support for the new ios-marketing app icon.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25210


hansemannn (approved):
Approved, love it! ❤️

ewieberappc (approved):
FR Passed. I no longer receive the warning about the ios-marketing app icon when submitting an iOS 11 app built with Xcode 9. There is also an appicon-Marketing.png image present in the build.

jvandijk:
@hans @ewieberappc Adding this logic doesn't work when only supporting the iphone device family by setting ipad to false.

Due to the fact that the build is not equal to universal it will hit a positive expression on line https://github.com/appcelerator/titanium_mobile/pull/9377/files#diff-f430483233aa01af5b10df8390f9635dR4931

and gets deleted from the array. In that case you won't be able to successfully publish to the app store.

cb1kenobi:
@jvandijk Doh, good catch! Reopening ticket and fixing. Thank you!!

PR #9384, opened 2017-08-31
[6_2_X] Android: Fix Windows gperf path

garymathews:
- Always use `gperf` that's available in the `PATH` environment variable


mukherjee2:
FR passed. Tried on Windows; modules packaged with no errors.

PR #9386, opened 2017-09-01
[TIMOB-25217] iOS: Fix parsing of unicode-characters

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25217


vijaysingh-axway (approved):
CR and FT passed.

PR #9390, opened 2017-09-01
[TIMOB-25218] (6_2_X) Docs: Fix Ti.UI.Toolbar example

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25218


ypbnv (approved):
CR: Pass

PR #9393, opened 2017-09-04
Remove non-ASCII character

ewanharris:
Docs validation was failing due to the apostrophe being a non-ASCII character


hansemannn (approved):
👁‍🗨

PR #9396, opened 2017-09-04
[TIMOB-17964] Add custom error message for wrong theme used.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17964

**Description:**
Show a custom error message whenever a theme without disabled ActionBar is used with Toolbar acting as ActionBar.


android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -752,7 +757,13 @@ public void onOrientationChanged(int orientation) {
 
 	private void setCustomActionBar() {
 		if (activityProxy.hasProperty(TiC.PROPERTY_SUPPORT_TOOLBAR)) {
-			this.setSupportActionBar(((Toolbar) ((TiToolbarProxy) activityProxy.getProperty(TiC.PROPERTY_SUPPORT_TOOLBAR)).getToolbarInstance()));
+			try {
+				this.setSupportActionBar(((Toolbar) ((TiToolbarProxy) activityProxy.getProperty(TiC.PROPERTY_SUPPORT_TOOLBAR)).getToolbarInstance()));
+			} catch (RuntimeException e) {
+				Log.e(TAG, customToolbarExceptionMessage);
```
garymathews: Since the error message will only be used once, it's best to just define it here

```Java
Log.e(TAG, "Attempting to use Toolbar as ActionBar without disabling the default ActionBar in the current theme.\n" +
           "You must set 'windowActionBar' to false in your current theme. Or use one of the following themes:\n" +
           "[Theme.Titanium, Theme.AppCompat.Translucent.NoTitleBar, Theme.AppCompat.NoTitleBar]\n" +
           "Which have ActionBar disabled by default.");
```

I have also re-worded it 👍

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
hansemannn: Can we do a list of theme instead of []? :-)

ypbnv:
@garymathews , @hansemannn Updated PR.

sgtcoolguy:
@garymathews @jquick-axway ping

hansemannn (approved):

jquick-axway (approved):
CR: Pass

PR #9397, opened 2017-09-04
[TIMOB-25223] Android: Fix remaining height calculation

garymathews:
- Fix remaining height calculation of child views

###### TEST CASE
```JS
		backgroundColor: 'gray'
	}),
	list = Ti.UI.createListView(),
	section = Ti.UI.createListSection({
		items: [
			{properties: {title: 'Square'}},
			{properties: {title: 'Circle'}},
			{properties: {title: 'Triangle'}}
		]
	}),
	view = Ti.UI.createView({
		width: Ti.UI.FILL,
		height: Ti.UI.FILL,
		backgroundColor: 'green'
	});

list.setSections([section]);
win.add([list, view]);

win.open();
```
- Should see a green square

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25223)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
@@ -286,6 +286,7 @@ public class TiCompositeLayout extends ViewGroup
         int wRemain = w;
         int wMode = MeasureSpec.getMode(widthMeasureSpec);
         int h = Math.max(hFromSpec, hSuggested);
+        int hRemain = h;
         int hMode = MeasureSpec.getMode(heightMeasureSpec);
 
         int maxWidth = 0;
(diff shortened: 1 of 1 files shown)


jquick-axway:
@garymathews, this code change creates new issues with "vertical" layouts when using FILL heights or pinned heights. But this is definitely the right solution for "composite" layouts. We need to calculate a remaining height for both vertical and horizontal layouts, but the remaining height should always be set to the view's height for "composite" layouts since it isn't applicable in that case. Good find (and my bad). It saved me time in isolating this issue.

I just pushed PR #9403 which solves both [TIMOB-25223] and [TIMOB-25221] regressions in one shot.

**Side Note:**
From now on, I'll be pushing QE to go through all of the layout test projects I've written if we make a layout code change. Even if it's a small code change. This is to avoid regressions in the future. So, to save QE's time, I wanted to combine these 2 regression fixes into 1.

garymathews:
@jquick-axway Closing, resolved in https://github.com/appcelerator/titanium_mobile/pull/9403

👍

PR #9399, opened 2017-09-05
[TIMOB-25227] Add default fallback to static framework type

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25227

**Optional Description:**
Only specifically check wether a framework is dynamic and fallback to static otherwise.


hansemannn (approved):
Works as expected, the Mach-O frameworks are recognized properly now. CR + FT passed!

PR #9401, opened 2017-09-05
Hook up new tests, fix up suite

sgtcoolguy:
**Description:**
The new ti.ui.toolbar and ti.ui.android.drawerlayout test suites were not hooked into our app.js unit test suite, so never got run. This adds them in and fixes them so that they pass.

I'm also attempting to start differentiate between skipping tests due to bugs/parity issues (effectively known issues we should address); versus not running or skipping or reporting certain tests on some platforms. If we have platform-specific APIs/properties/etc, the tests should only be run on those platforms and shouldn't even get reported as skipped on the others (as we've decided to explicitly limit the impl to certain platforms).

This should allow us over time to tease out the parity issues we need to address with existing APIs not working the same across platforms versus the platform-specific APIs that should remain as-is, or possibly move out to native modules/namespaces.

This is a work in progress.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,8 +2,8 @@
 library 'pipeline-library'
 currentBuild.result = 'SUCCESS'
 
-// Keep logs/reports/etc of last 5 builds, only keep build artifacts of last 3 builds
-properties([buildDiscarder(logRotator(numToKeepStr: '5', artifactNumToKeepStr: '3'))])
+// Keep logs/reports/etc of last 15 builds, only keep build artifacts of last 3 builds
+properties([buildDiscarder(logRotator(numToKeepStr: '15', artifactNumToKeepStr: '3'))])
 
 // Variables which we assign and share between nodes
 // Don't modify these yourself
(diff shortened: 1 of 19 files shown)


sgtcoolguy:
A potentially better way to filter the tests so we can mark them to be run/skipped or not run is to use https://github.com/nboxhallburnett/mocha-filter/blob/master/index.js

Then we can define filters for platform-specific tests, or for test that are broken on certain platforms. It would make for a much easier search of the APIs to determine our bugs/etc.

sgtcoolguy:
@garymathews I tagged you since, I needed to update the docs and the impl of DrawerLayout some to make these tests work.

@ewanharris I tagged you to see if you had any other suggestions to improve the test suite since I know  you've been working with it on Windows. Beyond adding the two missing test files to the suite, I know we had discussed making app.js just look for *.test.js files to require with a possible blacklist. It looks like our Ti.Filesystem.File APIs can list directories, so I can look into this, I just didn't like the idea of relying too much on our own APIs to do the actual setup of the tests...

ewanharris:
@sgtcoolguy Yeah now that I read that it makes more sense to me, (context switching between mocha in a node env and titanium env is tough..)

I would hope that our filesystem APIs would never be broken to that point but maybe another solution would be to have a template for the app.js that we write out beforehand on app creation in the test.js script (using ejs or whatever) to include the wanted requires?

PR #9405, opened 2017-09-06
[6_2_X][TIMOB-24802] apidoc: Mark ListViewAnimationProperties supported on Windows

infosia:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24802

`Ti.UI.ListView` supports `Ti.UI.ListViewAnimationProperties` on Windows as of 6.2.0.RC.

This requires https://github.com/appcelerator/titanium_mobile_windows/pull/1095 to be merged too.

### DIFF
--- a/apidoc/Titanium/UI/ListView.yml
+++ b/apidoc/Titanium/UI/ListView.yml
@@ -1135,7 +1135,7 @@ methods:
         type: Number
 
       - name: animation
-        summary: Animation properties. (iOS and Android only.)
+        summary: Animation properties. (iOS, Android and Windows)
         type: ListViewAnimationProperties
         optional: true
 
@@ -1715,7 +1715,7 @@ properties:
   - name: position
     summary: Specifies what position to scroll the selected cell to.
     type: Number
-    constants: Titanium.UI.iOS.ListViewScrollPosition.*
+    constants: [Titanium.UI.iOS.ListViewScrollPosition.*, Titanium.UI.Windows.ListViewScrollPosition.*]
     default: Titanium.UI.iOS.ListViewScrollPosition.NONE
 
 ---      


apidoc/Titanium/UI/ListView.yml
```diff
@@ -1715,7 +1715,7 @@ properties:
   - name: position
     summary: Specifies what position to scroll the selected cell to.
     type: Number
-    constants: Titanium.UI.iOS.ListViewScrollPosition.*
+    constants: [Titanium.UI.iOS.ListViewScrollPosition.*, Titanium.UI.Windows.ListViewScrollPosition.*]
```
hansemannn: We should move it to Ti.UI if at least 2 platforms support it

infosia: I think that's a very good point. I would define new constants under `Ti.UI.Windows` in this case because they may have a *same name* (such as **TOP**), but they *act differently*. 

I would think that if we can define *same behavior* across platforms for the constant, we should move it to `Ti.UI` namespace. But that was not the case for ListViewScrollPosition for now.

garymathews (approved):
LGTM

hansemannn:
@infosia Feel free to re-run Jenkines once the Windows PR is merged. Right now, it will fail:
```
ListViewAnimationProperties: found 1 error(s)!
	properties
		position
			constants
				Invalid constant: Titanium.UI.Windows.ListViewScrollPosition.*

```

infosia:
Looks like the yaml docs that I added to `titanium_mobile_windows` is not applied yet. @garymathews  Do you know how to push updates on `titanium_mobile_windows/apidoc/WindowsOnly`? It's https://github.com/appcelerator/titanium_mobile_windows/pull/1094 in particular.

hansemannn:
@infosia I think you cannot use any `Titanium.UI.Windows` references in this repo, it has to be merged from titanium_mobile_windows. Because of that, the fact that we always generate our web-docs from master, we release 6.2.0 today and I need to branch 6_3_X, I will close this PR. Please let me know if it should be reopened or re-created for 6.3.0.

PR #9407, opened 2017-09-06
Android: Fix Android development environment disparity

garymathews:
###### NOTE: The changes here are only for Titanium Android SDK developers working in IntelliJ/Android Studio

- Include `app.json` into *TitaniumTest* to allow invocation APIs to be processed when running from IntelliJ/Android Studio
- Move pre-compiled architecture `libs` from *TitaniumTest* into *v8* where they belong
- Fixed support libraries to prevent compilation errors in IntelliJ/Android Studio
- Enforce `1.6` Java language level to match our release builds
- Prevent circular reference from *design* module and *appcompat* module
- Update `*.iml` project files

###### TEST CASE
- Open `<titanium_mobile>/android` with IntelliJ or Android Studio
- Project should import and run without any additional changes necessary


ypbnv (approved):
CR: Pass

PR #9410, opened 2017-09-07
[TIMOB-25238] Android: Refactor TiBorderWrapperView to improve performance

garymathews:
- Refactor `TiBorderWrapperView` to improve performance of `Titanium.UI.View` with `borderColor` defined

###### TEST CASE \#1
```JS
    list = Ti.UI.createListView({
        templates: {
            template: {
                childTemplates: [
                    {
                        type: 'Ti.UI.View',
                        properties: {
                            left: 5,
                            right: 5,
                            top: 5,
                            backgroundColor: 'blue',
                            height: 90,
                            bottom: 5,
                            borderColor: 'orange',
                            borderRadius: 10,
                            borderWidth: 5
                        },
                        childTemplates: [
                            {
                                type: 'Ti.UI.View',
                                bindId: 'item',
                                properties: {
                                    left: 10,
                                    right: 10,
                                    height: 30,
                                    backgroundColor: 'red',
                                    borderColor: 'purple',
                                    borderRadius: 5,
                                    borderWidth: 3
                                }
                            }
                        ]
                    }
                ]
            }
        },
    }),
    section = Ti.UI.createListSection(),
    items = [];

for (var i = 0; i < 64; i++) {
    items.push({
        item: {
            width: '50%'
        }
    });
}

section.setItems(items);
list.setSections([section]);

win.add(list);
win.open();
```
- Should scroll without UI lag

###### TEST CASE \#2
```JS
var win = Ti.UI.createWindow({backgroundColor: 'white'}),
    a = Ti.UI.createView({
        top: 10,
        left: 10,
        width: 200,
        height: 100,
        elevation: 20,
        backgroundColor: 'red',
        borderColor: 'blue',
        borderWidth: 8
    }),
    b = Ti.UI.createView({
        top: 10,
        right: 10,
        width: 100,
        height: 100,
        elevation: 20,
        opacity: 0.5,
        backgroundColor: 'blue',
        borderColor: 'red',
        borderWidth: 8,
        borderRadius: 100
    });

win.add([a, b]);
win.open();
```
- Rectangle should have a border and elevation
- Circle should have a border with no aliasing

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25238)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
@@ -1,14 +1,10 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2012-2014 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2012-2017 by Axway, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 package org.appcelerator.titanium.view;
-
-import java.util.Arrays;
-
-import org.appcelerator.kroll.common.Log;
 
 import com.nineoldandroids.view.ViewHelper;
 
@@ -18,7 +14,6 @@ import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Path;
 import android.graphics.Path.Direction;
-import android.graphics.Path.FillType;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.graphics.RectF;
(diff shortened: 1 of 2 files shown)


ypbnv (approved):
Looks good to me!

jquick-axway:
I'd like to make a suggestion. (Note that this suggestion does not block this PR.)

It would be nice if our `TiUIView.hasBorder()` method would only return true if the border width or radius is greater than zero. The border color property should be ignored in this check since it only applies if a border width is greater than zero.
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java#L445

As it is now, a border view will always be applied to a view, even if these properties are initialized to zero (like how it is with WebView). So, not only will this get rid of the extra canvas draws/clipping in TiBorderWrapperView that'll do nothing, it'll also help prevent hardware acceleration from being disabled when it doesn't need to be.

lokeshchdhry:
FR Passed.

For the first test - UI does not show any lag & for the second test -- Rectangle shows border and elevation & circle has a border with no aliasing.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.4
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.14
Node Ver:       7.10.1
Java Ver:       1.8.0_101

PR #9412, opened 2017-09-07
[TIMOB-24248] Android: Ti.Calendar Recurring Events are not clearly exposed

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24248

**Description:**
Implements _Ti.Calendar.RecurrenceRule_ for Android. Due to the way Android handles recurrence rules it is not as complex as iOS and does not support all of the conditions, but provides parity for the most common cases.

Requirements: 
 - Device with logged in Google account with valid calendars.
 - Permissions in tiapp.xml
```
<uses-permission android:name="android.permission.WRITE_CALENDAR"/>
<uses-permission android:name="android.permission.READ_CALENDAR"/>
```

**Test case:**
```
var win = Ti.UI.createWindow({backgroundColor:'gray'});
var pickerCalendar = Ti.UI.createPicker({
    top: 50,
});
var pickerEvents = Ti.UI.createPicker({
    top: 100
});
var buttonGetRRule = Ti.UI.createButton({
    top: 150,
    title: 'Log RRule'
});
var buttonAddRRule = Ti.UI.createButton({
    top: 200,
    title: 'Add RRule'
});
var buttonGetEvent = Ti.UI.createButton({
    top: 250,
    title: 'Log Event'
});
var events = [];
var selectedEventIndex;
buttonGetRRule.addEventListener('click', function () {
    events[selectedEventIndex].recurrenceRules.forEach(function(rrule){
    	Ti.API.info('----------RRULE----------');
    	Ti.API.info(JSON.stringify(rrule));
    })
});

buttonAddRRule.addEventListener('click', function () {
     var newRule = events[selectedEventIndex].createRecurrenceRule({
                        frequency: Ti.Calendar.RECURRENCEFREQUENCY_MONTHLY,
                        interval: 1,
                        daysOfTheWeek: [{daysOfWeek:1,week:2},{daysOfWeek:2}],
                        end: {occurrenceCount:10}});
     events[selectedEventIndex].recurrenceRules = [newRule];
     events[selectedEventIndex].save();
});

buttonGetEvent.addEventListener('click', function () {
    	Ti.API.info('----------EVENT----------');
    	Ti.API.info(JSON.stringify(events[selectedEventIndex]));
});

var listEvents = function (calendarSource) {
    pickerEvents = Ti.UI.createPicker({
        top: 100
    });
    var i=0;
    var eventsRows = [];
    events = [];
    Ti.API.info(calendarSource);
    calendarSource.getEventsInYear(2018).forEach(function(event){
        //create row for each calendar
        events[i] = event;
        eventsRows[i++]=Ti.UI.createPickerRow({title:event.getTitle()});
    });
    if (eventsRows.length > 0) {
        pickerEvents.add(eventsRows);
        pickerEvents.addEventListener('change', function(e){
            selectedEventIndex = e.rowIndex;
        })
        win.add(pickerEvents);
        selectedEventIndex = 0;
    }
}

var getCalendars = function() {
    var calendarRows = [];
    Ti.Calendar.getAllCalendars().forEach(function(calendar){
        //create row for each calendar
        calendarRows.push(Ti.UI.createPickerRow({title:calendar.getName(), calendar: calendar}));
    });
    if (calendarRows.length >0 ) {
        pickerCalendar.add(calendarRows);
            pickerCalendar.addEventListener('change', function(e){
                win.remove(pickerEvents);
                listEvents(e.row.calendar);
            })
            win.add(pickerCalendar);
            listEvents(pickerCalendar.getSelectedRow(0).calendar);
               win.add(buttonGetRRule);
               win.add(buttonAddRRule);
               win.add(buttonGetEvent);
    }
}

win.addEventListener('open', function(){
    if (Ti.Calendar.hasCalendarPermissions()) {
        getCalendars();
    } else {
        Ti.Calendar.requestCalendarPermissions(function(){
            getCalendars();
        })
    }
});

win.open();
```


android/modules/calendar/src/java/ti/modules/titanium/calendar/CalendarModule.java
```diff
@@ -42,6 +42,13 @@
 	@Kroll.constant public static final int STATE_FIRED = AlertProxy.STATE_FIRED;
 	@Kroll.constant public static final int STATE_SCHEDULED = AlertProxy.STATE_SCHEDULED;
 
+	//region recurrence frequency
+	@Kroll.constant public static final int RECURRENCEFREQUENCY_DAILY = 0;
```
sgtcoolguy: Personally, I'd prefer `RECURRENCE_FREQUENCY_DAILY` to `RECURRENCEFREQUENCY_DAILY` (same goes for the rest)

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
```diff
@@ -117,7 +120,7 @@ public static String getExtendedPropertiesUri()
 			event.hasAlarm = !eventCursor.getString(7).equals("0");
 			event.status = eventCursor.getInt(8);
 			event.visibility = eventCursor.getInt(9);
-
+			event.setRecurrenceRules(eventCursor.getString(10), eventCursor.getInt(11));
```
sgtcoolguy: could this break when pulling up existing events with no recurrence?

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
```diff
@@ -126,6 +129,16 @@ public static String getExtendedPropertiesUri()
 		return events;
 	}
 
+	@Kroll.method
+	public void save() {
+		// Currently only saving added recurrenceRules.
+		String ruleToSave = ((RecurrenceRuleProxy) ((Object[]) getProperty(TiC.PROPERTY_RECURRENCE_RULES))[0]).generateRRULEString();
+		ContentValues contentValues = new ContentValues();
+		contentValues.put(Events.RRULE, ruleToSave);
```
sgtcoolguy: Not a huge fan of the truncated constant name here: `Events.RRULE` versus something like `Events.RECURRENCE_RULE`

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+		calculateDaysOfTheWeek();
+		calculateDaysOfTheYear();
+		calculateEnd();
+		calculateInterval();
+		calculateMonthsOfTheYear();
+		calculateWeeksOfTheYear();
+	}
+
+	//region Methods for converting native to Kroll values
+	private void calculateDaysOfTheMonth(){
```
sgtcoolguy: `private void calculateDaysOfTheMonth() {`

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+			}
+			return;
+		}
+		daysOfTheWeek = new KrollDict[]{};
+	}
+
+	private void calculateDaysOfTheYear() {
+		String days = matchExpression(".*(BYYEARDAY=[0-9]*).*", 10);
+		if (days != null && frequencyMap.get(frequency).equals(CalendarModule.RECURRENCEFREQUENCY_YEARLY)) {
+			daysOfTheYear = new int[1];
```
sgtcoolguy: `daysOfTheYear = new int[] { Integer.valueOf(days) };`

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+
+	// Field with the Date of the event begin to provide parity with iOS.
+	// Some information exposed can't be get solely from the recurrence rule.
+	private Date eventBegin;
+	private String rRule;
+
+	//region values that can be set from a Creation Dictionary
+	private int frequency = -1;
+	private int interval = -1;
+	private int[] daysOfTheMonth;
```
sgtcoolguy: I'm curious why the choice to use `int[]` for these fields that all seem to be either empty to have one single value. Personally I'd have used `Integer` type, with `null` representing a "no value".

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+		// Set the frequency in constructor, because it is required for other properties.
+		String frequency = matchExpression(".*(FREQ=[A-Z]*).*", 5);
+		if (frequency != null) {
+			this.frequency = frequencyMap.get(frequency);
+		}
+	}
+
+	private String matchExpression(String regEx, int length) {
+		Pattern pattern = Pattern.compile(regEx);
+		Matcher matcher = pattern.matcher(rRule);
```
sgtcoolguy: might want to use `this.rRule` for consistency here.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
sgtcoolguy: Can we use something like `Optional`: https://developer.android.com/reference/java/util/Optional.html or not until Api level 24 is our minimum?

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_CALENDAR_WEEKS_OF_THE_YEAR)) {
+			weeksOfTheYear = creationDictionary.getIntArray(TiC.PROPERTY_CALENDAR_WEEKS_OF_THE_YEAR);
+		}
+	}
+
+	public String generateRRULEString() {
+		StringBuilder finalRRule = new StringBuilder();
+		// Handle frequency.
+		if (frequency > -1) {
```
sgtcoolguy: use `this.frequency` to be consistent about field access?

I assume `-1` is some magical constant to denote "not specified"? Again I'd personally use `Optional<Integer>` or at least `Integer` with a `null` value to mark whether we have a value or not. But even if you don't do that, you may want to at least extract a constant for -1 to state it's "UNSPECIFIED" or something.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+			String frequencyPart = "FREQ=" + frequencyMap.keySet().toArray()[frequency - 1];
```
sgtcoolguy: How in the world is this consistent? The frequency map's key set is presumably not sorted/ordered.

I'm starting to think frequency should be an enum.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+						// Case in which we do not have items in daysOfTheWeek array.
+						monthlyReccurencesString.append(String.valueOf(daysOfTheMonth));
+					}
+					finalRRule.append(monthlyReccurencesString);
+					finalRRule.append(";");
+					break;
+			} // end of switch
+ 		}
+		// Handle interval.
+		if (interval > -1) {
```
sgtcoolguy: Again, use of a magic -1 constant is not particularly obvious.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	private void calculateInterval() {
+		String interval = matchExpression(".*(INTERVAL=[0-9]*).*", 9);
+		if (interval != null) {
+			this.interval = Integer.valueOf(interval);
+		}
+	}
+
+	private void calculateMonthsOfTheYear() {
```
sgtcoolguy: combine with `calculateWeeksOfTheYear` since they both really apply to yearly frequencies (and maybe rename to something like `calculateYearlyFrequency`)

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,360 @@
...
+			daysOfTheMonth = new int[1];
+			daysOfTheMonth[0] = Integer.valueOf(days);
+		} else {
+			daysOfTheMonth = new int[]{};
+		}
+	}
+
+	private void calculateDaysOfTheWeek() {
+		// Create a dictionary in the context of a month.
+		if (frequency == CalendarModule.RECURRENCEFREQUENCY_MONTHLY) {
```
sgtcoolguy: these two if clauses are exclusive. I think it may make more sense to calculate the necessary fields grouped by the frequency value, rather than to calculate grouped by specific fields.

Basically, initialize days of the week, days of the year, months of the year, etc with the default empty array value (or whatever empty value you decide to use based on Optional/Integer usage); and then override to the real value if the frequency is of a given type in a method that just handles that frequency type.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
jquick-axway: Our JavaScript API defines it as an array (link below). So, using an `int[]` is fine.
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar.RecurrenceRule-property-daysOfTheMonth

@ypbnv, we need to double check what iOS returns if this property is not set upon creation. That is, if it returns null or an empty array. We don't document this and these details matter. Once known, we can then initialize the `int[]` member variables appropriately (either null or empty).

android/modules/calendar/src/java/ti/modules/titanium/calendar/CalendarModule.java
ypbnv: This is the way we currently have them for iOS:

android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
ypbnv: Apparently it could. I did not get any exception, but it is documented as "implementation-defined"
https://developer.android.com/reference/android/database/Cursor.html#getString(int)
I will add an exception handling.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
ypbnv: @jquick-axway I agree. I used this as a reference:
https://developer.apple.com/documentation/eventkit/ekrecurrencerule/1507410-daysofthemonth?language=objc
Couldn't find the result there. Maybe try it on a device?

android/modules/calendar/src/java/ti/modules/titanium/calendar/CalendarModule.java
sgtcoolguy: I think we'd want to "fix" them to match our naming conventions. We'd have to retain the existing definitions and mark them as deprecated for this release, then in 8.0 or later we can remove the ones missing the underscores. So not really a breaking change, but an important deprecation to get in before GA.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
jquick-axway: @ypbnv, I just tested it out and confirmed that the following properties never return null on iOS. They return an empty array if not assigned upon creation. Personally, I agree with this behavior, because I find most script developers (unlike C/C++ devs) tend to not bother doing null checks.
- daysOfTheMonth
- daysOfTheWeek
- daysOfTheYear
- monthsOfTheYear
- weeksOfTheYear

So, we'll need to match the above behavior on Android.

Also, when you change the docs, would you mind updating the above properties with this info too please? Perhaps add the following sentence to the end of each property description like this...
`Will be an empty array if no recurrence of this type has been configured.`

Note that the `daysOfTheWeek` property is a dictionary and not an array. It definitely returns an empty dictionary for me on iOS in this case as well. So, update the docs appropriately.

Any case, I hope this helps!

ypbnv: They are currently working this way on Android. I will add the description in the documentation.

ypbnv:
@sgtcoolguy Updated the PR. I haven't marked the frequency constant yet. If we are OK to do that on iOS as well I will mark them once I update the documentation.

@jquick-axway Speaking of documentation - I did not find what are the values for `freqency` (should that even be accessible) and `interval` when they are not specified. For the latter we have 1 written in the docs, but it would be great if we can double check them as before setting them on Android side and update documentation.

jquick-axway:
@ypbnv,

The recurrence frequency constants are defined under our "Ti.Calendar" doc here...
- [RECURRENCEFREQUENCY_DAILY](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar-property-RECURRENCEFREQUENCY_DAILY)
- [RECURRENCEFREQUENCY_MONTHLY](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar-property-RECURRENCEFREQUENCY_MONTHLY)
- [RECURRENCEFREQUENCY_WEEKLY](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar-property-RECURRENCEFREQUENCY_WEEKLY)
- [RECURRENCEFREQUENCY_YEARLY](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar-property-RECURRENCEFREQUENCY_YEARLY)

There are no constants for `interval`. It must be an integer value that is >= 1. iOS will throw an exception stating "Interval must be greater than 0" if you call `createRecurrenceRule()` with an `interval` less than 1 or if the `interval` field is missing. I've confirmed this in the iOS simulator.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,336 @@
...
+	private final String until = "endDate";
+	private final String count = "occurrenceCount";
+
+	// Field with the Date of the event begin to provide parity with iOS.
+	// Some information exposed can't be get solely from the recurrence rule.
+	private Date eventBegin;
+	private String rRule;
+
+	//region values that can be set from a Creation Dictionary
+	private Integer frequency = null;
```
jquick-axway: Ideally, `frequency` should be an enum, not an `int` or `Integer`. This way we can easily constrain the value to known constants and provide easy mapping to know Android identifiers.

For example:

```java
public enum TiRecurrenceFrequencyType
{
	DAILY(CalendarModule.RECURRENCEFREQUENCY_DAILY, "DAILY"),
	WEEKLY(CalendarModule.RECURRENCEFREQUENCY_WEEKLY, "WEEKLY"),
	MONTHLY(CalendarModule.RECURRENCEFREQUENCY_MONTHLY, "MONTHLY"),
	YEARLY(CalendarModule.RECURRENCEFREQUENCY_YEARLY, "YEARLY");

	private final int tiIntId;
	private final String rfcStringId;

	private TiRecurrenceFrequencyType(int tiIntId, String rfcStringId)
	{
		this.tiIntId = tiIntd;
		this.rfcStringId = rfcStringId;
	}

	public int toTiIntId()
	{
		return this.tiIntId;
	}

	public String toRfcStringId()
	{
		return this.rfcStringId;
	}

	public static TiRecurrenceFrequencyType fromTiIntId(int value)
	{
		for (TiRecurrenceFrequencyType nextObject : TiRecurrenceFrequencyType.values()) {
			if ((nextObject != null) && (nextObject.tiIntId == value)) {
				return nextObject;
			}
		}
		return null;
	}

	public static TiRecurrenceFrequencyType fromRfcStringId(String value)
	{
		for (TiRecurrenceFrequencyType nextObject : TiRecurrenceFrequencyType.values()) {
			if ((nextObject != null) && (nextObject.rfcStringId == value)) {
				return nextObject;
			}
		}
		return null;
	}
}
```

With the above, you no longer need your `frequencyMap` static variable since the RFC frequency string ID mapping is already embedded into the enum. You can also easily convert/validate the JavaScript `frequency` field in your `RecurrenceRuleProxy` via the enum's `fromTiIntId()` like this...

```java
if (creationDictionary.containsKey(TiC.PROPERTY_FREQUENCY)) {
	int integerId = TiConvert.toInt(creationDictionary.get(TiC.PROPERTY_FREQUENCY));
	this.frequency = TiRecurrenceFrequencyType.fromTiIntId(integerId);
	if (this.frequency == null) {
		// Given frequency ID from JavaScript is invalid. Log an error.
	}
}
if (this.frequency == null) {
	// Default to daily if not assigned or given an invalid frequency.
	this.frequency = TiRecurrenceFrequencyType.DAILY;
}
```

You don't have to do this now, but Gary and I want to slowly transition our code to use the above pattern since it's a much more maintainable and less error prone solution. I've already started doing so with Titanium's orientation type handling here: [TiDeviceOrientation.java](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/util/TiDeviceOrientation.java).

ypbnv: @jquick-axway Updated.

ypbnv:
@jquick-axway Updated.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,376 @@
...
+			this.daysOfTheWeek = creationDictionary.getKrollDictArray(TiC.PROPERTY_CALENDAR_DAYS_OF_THE_WEEK);
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_CALENDAR_DAYS_OF_THE_YEAR)) {
+			this.daysOfTheYear = creationDictionary.getIntArray(TiC.PROPERTY_CALENDAR_DAYS_OF_THE_YEAR);
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_END)) {
+			this.endDictionary = (KrollDict) creationDictionary.getKrollDict(TiC.PROPERTY_END);
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_FREQUENCY)) {
+			this.frequency = TiRecurrenceFrequencyType.fromTiIntId(TiConvert.toInt(creationDictionary.get(TiC.PROPERTY_FREQUENCY)));
```
jquick-axway: The static `fromTiIntId()` method will return null if the given ID is invalid/unknown. We should either default to DAILY or throw an exception.

I've tested an invalid ID on iOS and it accepts the value as-is and puts the recurrence rule in an invalid state.

Perhaps throwing an exception would be better? It might be better to error out instead of assuming DAILY and creating unintended calendar events on the end-user's device. Note that throwing an exception is only valuable if thrown on the JavaScript runtime thread because we'll log a JavaScript stack trace.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,376 @@
...
+		// Set the recurrence rule for future use.
+		this.rRule = rrule;
+
+		// Set the ID of the calendar.
+		this.calendarID = String.valueOf(calendarID);
+
+		// Set the frequency in constructor, because it is required for other properties.
+		String frequency = matchExpression(".*(FREQ=[A-Z]*).*", 5);
+		if (frequency != null) {
+			this.frequency = TiRecurrenceFrequencyType.fromRfcStringId(frequency);
```
jquick-axway: The static method can return null. Should we throw an exception or default to DAILY?

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,376 @@
...
+	private final String until = "endDate";
+	private final String count = "occurrenceCount";
+
+	// Field with the Date of the event begin to provide parity with iOS.
+	// Some information exposed can't be get solely from the recurrence rule.
+	private Date eventBegin;
+	private String rRule;
+
+	//region values that can be set from a Creation Dictionary
+	TiRecurrenceFrequencyType frequency = TiRecurrenceFrequencyType.DAILY;
```
jquick-axway: The `frequency` member variable should be made `private`.
(The missing accessor keyword flags it "internal" at this package level.)

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,376 @@
...
+	private int[] daysOfTheYear = new int[]{};
+	private int[] monthsOfTheYear = new int[]{};
+	private int[] weeksOfTheYear = new int[]{};
+	private String calendarID;
+	private KrollDict endDictionary = new KrollDict();
+	private KrollDict[] daysOfTheWeek = new KrollDict[]{};
+	//endregion
+
+	// enum for matching Titanium frequency constants with Android recurrence rule words
+	public enum TiRecurrenceFrequencyType
```
jquick-axway: Let's make this enum `private` instead of `public` since it is only used be the parent class. This gives us the freedom to make major changes to it without fearing that an external 3rd party module might be making use of it without our knowing in the future (avoids unintended breaking changes).

Alternatively, I'm okay with this enum being `public` if pulled out into its own separate Java file, but once done, we should avoid breaking changes. (Personally, I would do this. I think the design of it is fine.)

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,376 @@
...
+				}
+				break;
+			case MONTHLY:
+				//daysOfTheMonth
+				days = matchExpression(".*(BYMONTHDAY=[0-9]*).*", 11);
+				if (days != null) {
+					this.daysOfTheMonth = new int[]{Integer.valueOf(days)};
+				}
+				//daysOfTheWeek
+				byDay = matchExpression(".*(BYDAY=[,0-9A-Z]*).*", 6);
```
jquick-axway: **Question:**
Do these regular expressions extract negative days?
We document that negative values are supported.

For example, a frequency type of MONTHLY with a -1 would reference the last day of the month, which is a valid use-case.

jquick-axway (changes requested):

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
ypbnv: The regular expressions are used to convert recurrence rule strings from Android format to Titanium format. I don't think we can get one with a negative value.
But I will add the support for negative values the other way around - Titanium to Android.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
ypbnv: Not sure what would be the better way. I will leave at printing an Error message and not adding the wrong rule.

ypbnv:
Updated.

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,372 @@
...
+		if (creationDictionary.containsKey(TiC.PROPERTY_CALENDAR_DAYS_OF_THE_YEAR)) {
+			this.daysOfTheYear = creationDictionary.getIntArray(TiC.PROPERTY_CALENDAR_DAYS_OF_THE_YEAR);
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_END)) {
+			this.endDictionary = (KrollDict) creationDictionary.getKrollDict(TiC.PROPERTY_END);
+		}
+		if (creationDictionary.containsKey(TiC.PROPERTY_FREQUENCY)) {
+			this.frequency =
+				TiRecurrenceFrequencyType.fromTiIntId(TiConvert.toInt(creationDictionary.get(TiC.PROPERTY_FREQUENCY)));
+		}
```
jquick-axway: iOS defaults to DAILY if the frequency was not set. We should do the same. Something like this...
```java
if (creationDictionary.containsKey(TiC.PROPERTY_FREQUENCY)) {
	this.frequency = TiRecurrenceFrequencyType.fromTiIntId(TiConvert.toInt(creationDictionary.get(TiC.PROPERTY_FREQUENCY)));
}
if (this.frequency == null) {
	this.frequency = TiRecurrenceFrequencyType.DAILY;
}
```

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -0,0 +1,372 @@
...
+			this.interval = Integer.valueOf(interval);
+		}
+	}
+	//endregion
+
+	//region kroll proxy methods
+	@Kroll
+		.getProperty
+		@Kroll.method
+		public String getCalendarID()
```
jquick-axway: Unfortunately, adding multiple annotations to the same member triggers this clang formatting issue. We currently work-around it by adding the following comments around it like the below. (It's a bummer, I know. Sorry.)
```java
// clang-format off
@Kroll.getProperty
@Kroll.method
public String getCalendarID()
// clang-format on
{
	return this.calendarID;
}
```

apidoc/Titanium/Calendar/Event.yml
```diff
@@ -86,7 +86,8 @@ methods:
         type: Number
         constants: Titanium.Calendar.SPAN_*
         default: <Titanium.Calendar.SPAN_THISEVENT>
-    platforms: [iphone, ipad]
+    platforms: [android, iphone, ipad]
+    since: {android: "7.0.0", iphone: "3.1.0", ipad: "3.1.0"}
```
jquick-axway: Change Android version to "7.1.0"

apidoc/Titanium/Calendar/Event.yml
```diff
@@ -249,9 +250,13 @@ properties:
...
   - name: recurrenceRules
     summary: The recurrence rules for the calendar item. (Available in iOS 5.1 and above.)
+    description: |
+        On Android only the first element of the recurrenceRules is take into account 
+        due to the way it handles conditions for recurrence rules.
     type: Array<Titanium.Calendar.RecurrenceRule>
     osver: {ios: {min: "5.0"}}
-    platforms: [iphone, ipad]
+    platforms: [android, iphone, ipad]
+    since: {android: "7.0.0", iphone: "3.1.0", ipad: "3.1.0"}
```
jquick-axway: Change Android version to "7.1.0"

apidoc/Titanium/Calendar/Event.yml
```diff
@@ -249,9 +250,13 @@ properties:
     
   - name: recurrenceRules
     summary: The recurrence rules for the calendar item. (Available in iOS 5.1 and above.)
+    description: |
+        On Android only the first element of the recurrenceRules is take into account 
+        due to the way it handles conditions for recurrence rules.
```
jquick-axway: We should add a comma after "On Android," and the word "take" should be "taken".

```
On Android, only the first element of the recurrenceRules is taken into account
due to the way it handles conditions for recurrence rules.
```

android/titanium/src/java/org/appcelerator/kroll/KrollDict.java
```diff
@@ -182,6 +188,22 @@ public KrollDict getKrollDict(String key)
...
 	}
 
+	public KrollDict[] getKrollDictArray(String key)
+	{
+		String[] value = getStringArray(key);
+		KrollDict[] result = new KrollDict[value.length];
+		int index = 0;
+		for (String record : value) {
+			try {
+				result[index++] = new KrollDict(new JSONObject(record));
```
jquick-axway: From looking at `TiConvert.toStringArray()`, an element in the array can be `null`. So, this will wrongly remove null entries and reduce the size of the array. Perhaps the below would be better?

```java
for (String record : value) {
	KrollDict dictionary = null;
	if (record != null) {
		try {
			dictionary = new KrollDict(new JSONObject(record));
		} catch (Exception e) {
			e.printStackTrace();
			Log.w(TAG, "Unable to parse dictionary.");
		}
	}
	result[index++] = dictionary;
}
```

jquick-axway (changes requested):

ypbnv:
@jquick-axway Updated the PR. Also found few other things that should be updated. 
@lokeshchdhry I updated the test case as well.

lokeshchdhry:
FR Passed.

`Recurrence events` works as expected.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101
Devices:        ⇨ samsung SM-G955U1 --- Android 7.0

jquick-axway (commented):
CR: Pass

PR #9420, opened 2017-09-11
Updated DrawerLayout.yml to fix formatting issue for code sample

jawa9000:
Added a missing line return for the code sample in line 133 to fix the broken sample display when rendering this content to our API docs page. Confirmed locally that this fix works.

https://jira.appcelerator.org/browse/TIDOC-3005

### DIFF
--- a/apidoc/Titanium/UI/Android/DrawerLayout.yml
+++ b/apidoc/Titanium/UI/Android/DrawerLayout.yml
@@ -131,6 +131,7 @@ examples:
     example: |
         The following code shows a simple drawer-layout usage.
 
+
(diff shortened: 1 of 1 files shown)


hansemannn (approved):

hansemannn:
Approved. But is this really the issue? I updated Toolbar-related docs last week with only [one blank line](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/UI/Toolbar.yml#L245) and it works (see [here](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.Toolbar)).

PR #9422, opened 2017-09-12
[TIMOB-25242] Android: Improve validation in filenameExists()

garymathews:
- Prevent `filenameExists()` from crashing when `fileIndex` is `null`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25242)


jpriebe:
Questions: does the fact that JSON.parse("index.json") does not return an object indicate a larger problem?  

If we're trying to load a module that *should* exist and was built into the app, but this filenameExists() call returns null, that could be almost as bad as the V8Exception.

And wouldn't any subsequent calls to filenameExists() for *any* files just return null because you're missing this critical index file?

jpriebe:
We hit the exception in question intermittently in our app.  We'll hit it, then launch the app again, and we don't hit it.

Assuming that index.json is built at compile time, then it clearly exists on the device even when the exception is being thrown.  So it's probably not so much an issue of the index.json not existing, but more of a race condition of sorts.

I don't know if that means there should be some kind of retry mechanism when you can't parse index.json.

At any rate, a casual inspection of the code would suggest that if index.json is not parseable, there's no way to require() a JS module.  So that's pretty much going to kill the app.

android/runtime/common/src/js/module.js
```diff
@@ -617,6 +617,9 @@ Module.prototype.filenameExists = function (filename) {
 	if (!fileIndex) {
 		var json = assets.readAsset('index.json');
 		fileIndex = JSON.parse(json);
+		if (!fileIndex) {
```
sgtcoolguy: I think we want to explicitly return a Boolean in this method. I understand that `null` is "truthy" and evaluates to `false`, but I'd strongly prefer we simply alter the return statement below to be:
```javascript
return filename && (filename in fileIndex);
```

sgtcoolguy (changes requested):

sgtcoolguy:
@jpriebe Yes, being unable to load index.json is indicative of a larger problem. That file is generated at build time with the listing of assets in the app. If that file is missing or corrupt, I don't believe you'd be able to require any file.

sgtcoolguy:
So taking a quick look, my best guess here is that the `assets.readAsset('index.json');` call must be returning `null`, since that's how `JSON.parse(json);` would return `null`. If you had empty string I believe it throws a SyntaxError. So I think the underlying bug is likely to be in the asset manager stuff, and given that it's intermittent, @jpriebe is likely right it's some sort of app/runtime startup race condition.

garymathews:
@sgtcoolguy Thanks for taking a look, updated PR!

gimdongwoo:
Any update?

gimdongwoo:
This commit has seriously bugs.

```
[ERROR] TiExceptionHandler: (main) [194,194] ----- Titanium Javascript Runtime Error -----
[ERROR] TiExceptionHandler: (main) [0,194] - In ti:/module.js:305,2
[ERROR] TiExceptionHandler: (main) [0,194] - Message: Uncaught Error: Requested module not found: av.imageview
[ERROR] TiExceptionHandler: (main) [0,194] - Source: 	throw new Error("Requested module not found: " + request);
...
[INFO]  art: Rejecting re-init on previously-failed class java.lang.Class<com.onesignal.OneSignalChromeTab$OneSignalCustomTabsServiceConnection>: java.lang.NoClassDefFoundError: Failed resolution of: Landroid/support/customtabs/CustomTabsServiceConnection;
```

lokeshchdhry:
FR Passed.

No V8 exceptions seen & module loading works fine, checked with using `ti.paint` module in the app.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

PR #9423, opened 2017-09-13
[TIMOB-25269] Support iPhone X launch-screens

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25269

**Optional Description:**
This will add support for the new iPhone X launch-screen dimensions. Note that the new filename-suffix is not `812h` but `2436h` as Apple uses it internally and in the asset-catalog. 

This PR also replaces all other default launch-screens to have a much smaller file-size to not irritate new developers exporting their first IPA. 

Bonus tip: Use Storyboards for smaller app-sizes and easier migration. When using Storyboards, developers do not even need to update their apps for iPhone X as it will adapt automatically.

6_3_X backport coming as soon as it's branched after 6.2.0.GA release.


cb1kenobi (approved):
CR'd and FR'd. Looks good! **APPROVED**

PR #9425, opened 2017-09-13
[TIMOB-17821] Android: Add Support for Custom View to ActionBar

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17821

**Description:**
Expose setCustomView method for ActionBar.
We can add support in Alloy for that, but it would require some changes there as well.

**Test cases:**

_With default action bar._
```
var window = Ti.UI.createWindow({title: 'Title', backgroundColor: 'gray'});
var customView = Ti.UI.createView({layout: 'horizontal'});
var button1 = Ti.UI.createButton({title: 'Button1'});
var button2 = Ti.UI.createButton({title: 'Button2'});
var picker = Ti.UI.createPicker({});
var slider = Ti.UI.createSlider({width: '30%', backgroundColor: 'teal'});
var data = [];

data[0]=Ti.UI.createPickerRow({title:'Row 1'});
data[1]=Ti.UI.createPickerRow({title:'Row 2'});
data[2]=Ti.UI.createPickerRow({title:'Row 3'});
picker.add(data);

customView.add(button1);
customView.add(slider);
customView.add(picker);

button2.addEventListener('click', function(){
    window.activity.actionBar.customView = customView;
});

window.add(button2);
window.open();
```

_With Toolbar used as ActionBar.
Note that you need to use a theme with disabled ActionBar (eg. Theme.Titanium)_

```
var window = Ti.UI.createWindow({title: 'Title', backgroundColor: 'gray'});
var customView = Ti.UI.createView({layout: 'horizontal'});
var button1 = Ti.UI.createButton({title: 'Button1'});
var button2 = Ti.UI.createButton({title: 'Button2'});
var picker = Ti.UI.createPicker({});
var slider = Ti.UI.createSlider({width: '30%', backgroundColor: 'teal'});
var toolbar = Ti.UI.createToolbar({top:0, barColor: '#222222', width: Ti.UI.FILL});
var data = [];

data[0]=Ti.UI.createPickerRow({title:'Row 1'});
data[1]=Ti.UI.createPickerRow({title:'Row 2'});
data[2]=Ti.UI.createPickerRow({title:'Row 3'});
picker.add(data);

customView.add(button1);
customView.add(slider);
customView.add(picker);

window.activity.supportToolbar = toolbar;

button2.addEventListener('click', function(){
    window.activity.actionBar.customView = customView;
});

window.add(toolbar);
window.add(button2);
window.open();
```

**Preview:**
example


android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -378,6 +391,11 @@ private Drawable getDrawableFromUrl(String url)
 		return tfh.loadDrawable(imageUrl.resolve(), false);
 	}
 
+	private void handleSetCustomView(TiViewProxy view)
+	{
+		actionBar.setCustomView(view.getOrCreateView().getNativeView());
```
jquick-axway: We need to handle the case where the custom view is set to `null`.

Perhaps do one of the following:
1. Ignore null and no-op. (Simple case.)
2. Ideally, remove the last custom view set, but I don't think ActionBar supports this. Calling `setDisplayShowCustomEnabled(false)` would hide it, but still exist in memory.

Also, we may want to explicitly call `setDisplayShowCustomEnabled(true)` just in case a forked Android OS doesn't have this as the default setting.

apidoc/Titanium/Android/ActionBar.yml
```diff
@@ -137,7 +137,19 @@ methods:
...
+  - name: setCustomView
+    summary: Sets a view to be used for a custom navigation mode.
+    parameters:
+      - name: view
+        summary: View to be used.
+        type: Titanium.UI.View
+    description: |
+        Inserts a custom view for navigation in the space
+        between the application's home button and menu actions.
+        The cusom view is trimmed to fit the space available.
```
jquick-axway: Typo: Change "cusom" to "custom".

Also, version number needs to change from "7.0.0" to "7.1.0".

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -427,6 +445,9 @@ public boolean handleMessage(Message msg)
 			case MSG_SET_HOME_BUTTON_ENABLED:
 				handlesetHomeButtonEnabled(msg.getData().getBoolean(HOME_BUTTON_ENABLED));
 				return true;
+			case MSG_SET_CUSTOM_VIEW:
+				handleSetCustomView(((TiViewProxy) msg.obj));
+				return true;
```
jquick-axway: I'm not a fan of setters without getters. Ideally, this should be a `customView` property which we should then call `handleSetCustomView()` when the property changes.

jquick-axway (changes requested):

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
ypbnv: The Android method `setCustomView()` accepts `null` as a parameter and it removes the latest custom view added. I think this is fine. We get the Android native view through the proxy that is passed and it will be released once it is not referenced anymore.
`setDisplayShowCustomEnabled()` is called with true the first time a custom view is used.

ypbnv:
@jquick-axway Updated. I couldn't come up with unit tests for that though.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

`customView` for `actionBar` works as expected in both the cases.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101
Emulator: Android 4.2.2

PR #9437, opened 2017-09-15
[TIMOB-25285] Throw error on duplicate third-party frameworks

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25285

**Optional Description:**
Stops the build by throwing an error when duplicate frameworks are present instead of silently choosing the first framework found, which might result in unexpected behavior.


janvennemann:
@hansemannn, let me know if the CR and phrasing are good in this one. I'll then create the backports for 6_3_X and 6_2_X.

iphone/cli/hooks/frameworks.js
```diff
@@ -166,6 +166,7 @@ class FrameworkManager {
 	 *
 	 * @param {Object} hookData - Data from the Xcode project hook
 	 * @param {Function} callback - Callback function
+	 * @return {undefined}
```
hansemannn: Undefined? Is this the appropriate way to document guarded return statements?

janvennemann: It is actually undefined because the callback should have no return value and if a function has no specified return value it's undefined. Same goes for the "else" branch. There is no explicit return value so it will be undefined

hansemannn (approved):

iphone/cli/hooks/frameworks.js
cb1kenobi: FWIW, I never define a `@return` when the function does not explicitly have a `return` statement.

janvennemann: Yeah usually I do that too. In this case the function has a return statement `return callback()` inside an if-branch to return early at the top of the function to leave out the else branch and our linting rules complained about a missing return JSDoc.

cb1kenobi (approved):
LGTM

janvennemann:
@hansemannn, backports for 6_2_X and 6_3_X are ready.

ewieberappc:
FR Passed.
Tested using frameworks in modules (both local and system) and in the platform folder on both classic and Alloy projects. Error and build failure was correctly shown each time.

PR #9439, opened 2017-09-18
titanium_ui_drawer_layout.xml fixes: background, elevation and fitsSy…

alangumer:
The background color was hardcoded so was not possible to change the color via theme; also the property android:fitsSystemWindows="true" was added to avoid that the status bar overlaps the App bar; and the elevation was set to 0 which looks nicer especially when using navigation tabs.


hansemannn:
Thanks for the fix, great catch! We'll handle it for 6.2.1 to get it to everyone asap.

garymathews (approved):
LGTM 👍

PR #9458, opened 2017-09-20
[TIMOB-25322] iOS: Geolocation should be able to handle iOS 11 permission upgrade 

hansemannn:

Unit-Tests not possible as it require's UI-interaction with the confirmation-dialog. Might be a good candidate for UI-tests in our Appium test-suite.


jvandijk:
@hansemannn I understand that you choose to be backwards compatible, but this is not going to work out if you want to support iOS 11 from a functional perspective. If you go for the permission upgrade strategy, then this will work now in iOS11, but on iOS 10 you don't get the chance to upgrade. You'll have to add quite some checks on the javascript side to be back- and forwards compatible. Question is why you weren't allowed to upgrade before? This is quite odd.... isn't it?

hansemannn:
@jvandijk From my understanding, this was not technically possible from Apple's side before, so we guarded it in the past. If it was, we can remove those guards for a better code-structure. The iOS < 11 checks are correct and seem to be necessary from what I know by now.

iphone/Classes/GeolocationModule.m
```diff
@@ -280,12 +279,20 @@ - (CLLocationManager *)locationManager
     }
     locationManager.headingFilter = heading;
 
-    if ([[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationAlwaysUsageDescription"]) {
-      [locationManager requestAlwaysAuthorization];
-    } else if ([[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSLocationWhenInUseUsageDescription"]) {
-      [locationManager requestWhenInUseAuthorization];
-    } else {
-      NSLog(@"[ERROR] The keys NSLocationAlwaysUsageDescription or NSLocationWhenInUseUsageDescription are not defined in your tiapp.xml. Starting with iOS8 this is required.");
+    if ([CLLocationManager authorizationStatus] != kCLAuthorizationStatusAuthorizedAlways &&
```
jvandijk: Perfect addition! Already ran into issues in the past due to this 'auto' permission request logic.

jvandijk:
I've been testing this way of implementing it and also ran into the fact that the didChangeAuthorizationStatus is firing on showing the 2nd authorization request. The code in that method is also not capable of handling the authorization, and could fire the callback unnecessary.

hansemannn:
@jvandijk You mean upgrading the permissions on iOS < 11? So it's not worth looking into it in the SDK? I really focussed on the iOS 11 changes for this one. For the changes, I can also think of two ones for the client (if you want the improved UX):

Normal geo-request:
1. If iOS < 11: Request Always-Auth
2. Else: Request When in Use

Going background / Always-auth / geofencing:
1. If iOS >= 11: Request Always-Auth
2. No change for iOS < 11 required

jvandijk:
No it is worth looking into the SDK, because I experienced this on iOS 11 indeed. The `didChangeAuthorizationStatus` fired, and resulted in a callback on which I execute new things in the client. This then results in the 2nd upgrade permissions request just to give a visual flash.... and disappear again without the user being able to select a new level. And you can only trigger every permission request once.... resulting in a weird UX.

hansemannn:
Can you shoot me a test-case?

hansemannn:
@jvandijk Ping?

jvandijk:
@hansemannn Sorry, was too busy with other things. In my situation I add an eventlistener to the resume event (coming back from the permission dialog) and in that eventlistener verify if the permission has been set to update the button state. That's the only difference from the described testcase on Jira.

hansemannn:
No worries. Did you test the latest version?

jvandijk:
Tested with 6.2.2

hansemannn:
I mean, did you test the latest PR changes?

jvandijk:
@hansemannn I have just one more concern... and that's the hasLocationPermissions method. When checking `WHEN_IN_USE` but the user has given `ALWAYS` should iOS 11 or greater then return true or false?

hansemannn:
Good question. I think it should / could show a warning but return `false`, because the actual permission does not match the one being requested. Does this behave different for iOS 11 / < 11?

jvandijk:
It does not behave differently, just that `always` now implicitly means that you have `when in use` as well because you're able to upgrade.

hansemannn:
That‘s correct, but if you want to check that permission, you would explicitly ask for it. In other words: I cannot see a use-use where this would fail.

vijaysingh-axway (approved):
LGTM!

hansemannn:
@ewieberappc Just resolved a linting-issue, so this is now definitely ready for QE 🙂.

ewieberappc:
When following the steps from the ticket, I get an error that the NSLocationWhenInUseUsageDescription key must be defined. It seems that NSLocationAlwaysAndWhenInUseUsageDescription is not covering the "just when-in-use" request case.

hansemannn:
@ewieberappc Can you send me your test project?

hansemannn:
Closing this one in favor of the above linked one that handles this one + more improvements regarding the new auth-flow and the requirement of the when-in-use key.

PR #9459, opened 2017-09-20
[TIMOB-25277] Android: Implement onLink callback for Ti.UI.WebView

garymathews:
- Implement `onlink` callback property for `Titanium.UI.WebView`
- Callback must return `true` or `false` as a condition for loading the link
- Fixed issue where events from new windows would not fire

###### TEST CASE
```JS
        url: 'https://india.gov.in/website-ministry-commerce-and-industry',
        onlink: function(e) {
            if (e.url.endsWith('.pdf')) {
                alert('PDF: ' + e.url);
                return false;
            }
            return true;
        }
    });

window.add(webView);
window.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25277)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java
@@ -97,20 +97,6 @@ public class TiWebViewClient extends WebViewClientClassicExt
 		if (proxy == null) {
 			return;
 		}
-		if (proxy.hasProperty(TiC.PROPERTY_ON_LINK)) {
-			Object onLink = proxy.getProperty(TiC.PROPERTY_ON_LINK);
-			if (onLink instanceof KrollFunction) {
-				KrollFunction onLinkFunction = (KrollFunction) onLink;
-				KrollDict args = new KrollDict();
-				args.put(TiC.EVENT_PROPERTY_URL, url);
-				Object result = onLinkFunction.call(proxy.getKrollObject(), args);
-				if (result == null || (result instanceof Boolean && ((Boolean) result) == false)) {
-					webView.stopLoading();
-					return;
-				}
-			}
-		}
-
 		KrollDict data = new KrollDict();
 		data.put("url", url);
 		proxy.fireEvent("beforeload", data);
@@ -139,6 +125,19 @@ public class TiWebViewClient extends WebViewClientClassicExt
 		WebViewProxy proxy = (WebViewProxy) webView.getProxy();
 		if (proxy == null) {
 			return super.shouldOverrideUrlLoading(view, url);
+		}
+		if (proxy.hasProperty(TiC.PROPERTY_ON_LINK)) {
+			Object onLink = proxy.getProperty(TiC.PROPERTY_ON_LINK);
+			if (onLink instanceof KrollFunction) {
+				KrollFunction onLinkFunction = (KrollFunction) onLink;
+				KrollDict args = new KrollDict();
+				args.put(TiC.EVENT_PROPERTY_URL, url);
+				Object result = onLinkFunction.call(proxy.getKrollObject(), args);
+				if (result == null || (result instanceof Boolean && ((Boolean) result) == false)) {
+					webView.stopLoading();
+					return true;
+				}
+			}
 		}
 		if (proxy.hasProperty(TiC.PROPERTY_BLACKLISTED_URLS)) {
 			String[] blacklistedSites =
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java
@@ -7,6 +7,7 @@
 package ti.modules.titanium.ui.widget.webview;
 
 import org.appcelerator.kroll.KrollDict;
+import org.appcelerator.kroll.KrollFunction;
 import org.appcelerator.kroll.common.Log;
 import org.appcelerator.titanium.TiApplication;
 import org.appcelerator.titanium.TiC;
@@ -96,6 +97,20 @@ public class TiWebViewClient extends WebViewClientClassicExt
 		if (proxy == null) {
 			return;
 		}
+		if (proxy.hasProperty(TiC.PROPERTY_ON_LINK)) {
+			Object onLink = proxy.getProperty(TiC.PROPERTY_ON_LINK);
+			if (onLink instanceof KrollFunction) {
+				KrollFunction onLinkFunction = (KrollFunction) onLink;
+				KrollDict args = new KrollDict();
+				args.put(TiC.EVENT_PROPERTY_URL, url);
+				Object result = onLinkFunction.call(proxy.getKrollObject(), args);
+				if (result == null || (result instanceof Boolean && ((Boolean) result) == false)) {
+					webView.stopLoading();
+					return;
+				}
+			}
+		}
+
 		KrollDict data = new KrollDict();
 		data.put("url", url);
 		proxy.fireEvent("beforeload", data);
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -560,10 +560,10 @@ properties:
   - name: onlink
     summary: Fired before navigating to a link.
     description: |
-        Android only. The callback will be called before navigating to the link. The Boolean return value of the
-        callback will determine if the link will be navigated.
-    type: Callback<Boolean>
-    platforms: [android]
+        The callback will be called before navigating to the link. The Boolean return value of the
+        callback will determine if the link will be navigated or discarded.
+    type: Callback<OnLinkURLResponse>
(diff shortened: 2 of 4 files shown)


garymathews:
@hansemannn Maybe we could add this to iOS for parity?

hansemannn:
We should call it "link", so without the `on`. The reason we are doing that is because Alloy would use it as `on="onLink"` which looks pretty bad. Also, when exactly is that fired? We already have `beforeload` which sounds pretty similar to me.

garymathews:
I can't make it an event since it requires an immediate return value (similar to `onCreateWindow`)

hansemannn:
But using the existing API's, they should do that already if I got that correct? iOS uses a delegate-method that returns something as well, so checking for the blacklisted URL's changes the return-value. Exposing this as a property would not be possible on iOS and also wouldn't make sense since the above reasons.

jquick-axway:
It sounds like the ultimate goal here is to provide an opportunity to cancel URL navigation and do something native instead. I'd like to share my experience regarding this.

How this is implemented now, the `onLink()` callback is hooked up to the Android [WebViewClient.onPageStarted()](https://developer.android.com/reference/android/webkit/WebViewClient.html#onPageStarted(android.webkit.WebView,%20java.lang.String,%20android.graphics.Bitmap)), which unfortunately is invoked **after** the HTTP request has been sent. So, while you may be able to cancel the page load in-time on a slow Internet connection, in my experience, you'll never be able to cancel it in time when loading a local HTML file. For example, I've worked with mobile app devs who used WebViews and local HTML files more as a fancy UI front-end where the links were unregistered URL schemes to tell the native code to execute an operation (and the navigation needs to be canceled to prevent a 404).

On iOS, our Titanium `beforeload` event is hooked up to the [UIWebView.shouldStartLoadWithRequest](https://developer.apple.com/documentation/uikit/uiwebviewdelegate/1617945-webview?language=objc) which gets invoked before the HTTP request gets sent. That is, UIWebView is asking the app developer for permission to navigate. This event would allow a developer to set up the above functionality mentioned (ie: cancel the HTTP request and perform a native operation instead) and is what we're really after.

On Android, the equivalent to iOS' `shouldStartLoadWithRequest` method is [WebViewClient.onLoadResource()](https://developer.android.com/reference/android/webkit/WebViewClient.html#onLoadResource(android.webkit.WebView,%20java.lang.String)) which asks permission before sending the HTTP request... **but** it has 1 problem. The Android `onLoadResource()` method does not get invoked when navigating to another URL via JavaScript; it only gets invoked when tapping on a link. That was my experience back when I was trying to do this during the Android 2.x days. Back then, navigating via URLs can only be detected via the `onPageStarted()` when it is already too late (I've never found a work-around unfortunately).

So... I'm wondering if the real solution here is to tweak our "beforeload" event to provide this functionality. On Android, "beforeload" technically happens when the next URL has already started and not before it loads like how it is on iOS. But that said, I think it was made this way on purpose on Android to work-around JavaScript URL navigation detection mentioned above. However, we can still tweak our Android code to trigger a "beforeload" via the Java `onLoadResource()` and flag the event as already fired when we reach the Java `onPageStarted()` to prevent the event from being fired twice.

Anyways, that's my 2 cents.
I'm curious if Android 4.x or newer OS versions has made this any easier.

garymathews:
The issue with `onLoadResource()` is that every resource loaded by the page is passed through it (images, css, scripts). Where we only want the link the user has clicked on.

I think in this scenario `onPageStarted()` will do.

jquick-axway:
You know what, I think I got my Android methods mixed up. I meant [WebViewClient.shouldOverrideUrlLoading()](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)), not `onLoadResource`.

garymathews:
Ahh, I do. But I intercept those before they reach `shouldOverrideUrlLoading()` [here](https://github.com/appcelerator/titanium_mobile/pull/9459/files#diff-91b884f96a542246fb19fd75437fe3e2R124).

hansemannn:
@garymathews I've updated the docs to fix the validation error. Let me know if something else should be revisited.

lokeshchdhry:
@garymathews , @jquick-axway , Can anyone of you please fix the lint errors. Thanks !!

apidoc/Titanium/UI/WebView.yml
```diff
@@ -551,7 +551,7 @@ properties:
     type: Callback<Boolean>
     platforms: [android]
     since: "7.2.0"
```
lokeshchdhry: Needs to be changed to 7.3.0

lokeshchdhry:
FR Passed.

`onLink` callback is called successfully with the url of the pdf.

OS Ver:         10.13.4
Xcode Ver:      Xcode 9.3.1
Appc NPM:       4.2.13
Appc CLI:       7.0.3
Daemon Ver:     1.1.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10

lokeshchdhry:
Waiting for CR to pass.

lokeshchdhry:
@garymathews , @jquick-axway , Are we making any changes to this PR ?

sgtcoolguy:
Bumping to 7.4.0 to be merged with an equivalent iOS implementation per https://jira.appcelerator.org/browse/TIMOB-26063 cc @hansemannn 

Also to be merged in same released as Windows PR: appcelerator/titanium_mobile_windows#1211

hansemannn:
I would still disagree with the interface design here. We have something very similar with the "blacklisturl" event already, which is cross-platform and works today already. Can't we just get the URL from there, cancel the HTTP request and live a happy life?

garymathews:
@jquick-axway Could you write a test case that reproduces the threading issue with the `onlink` callback? I'm having trouble reproducing it

jquick-axway:
Okay. I've verified that the `onlink` callback is thread safe when "run-on-main-thread" is set to `false`. I've verified it by running a brute-force loop in JS while tapping on a link to invoke the`onlink` callback.

For everyone's info, the `KrollFunction.call()` will block the current thread and queue the callback to be invoked on the runtime thread when available. This avoids a race condition and prevents the JavaScript stack from getting clobbered. To walk you through it, `KrollFunction.call()` will call `TiMessenger.sendBlockingRuntimeMessage()`, which calls `TiMessenger.sendBlockingMessage()`, and that method will block forever until the callback has been invoked.
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/V8Function.java#L38
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/java/org/appcelerator/kroll/common/TiMessenger.java#L174
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/java/org/appcelerator/kroll/common/TiMessenger.java#L239

@hansemannn, do we have something similar on iOS?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java
```diff
@@ -97,6 +98,20 @@ public void onPageStarted(WebView view, String url, Bitmap favicon)
...
+				KrollDict args = new KrollDict();
+				args.put(TiC.EVENT_PROPERTY_URL, url);
+				Object result = onLinkFunction.call(proxy.getKrollObject(), args);
+				if (result == null || (result instanceof Boolean && ((Boolean) result) == false)) {
+					webView.stopLoading();
+					return;
+				}
+			}
+		}
+
```
jquick-axway: There's one issue here. The `onPageStarted()` method gets called after the HTTP request was sent. If you set up the URL to use a customer URL scheme that was intended to be overridden by `onlink`, then the WebView will display an "unknown url scheme" error page.

I'm able to reproduce this issue with the following:
```javascript
	html: '<!DOCTYPE html><html><body><a href="myapp:HelloWorld">Tap this link</a></body></html>',
	onlink: function(e) {
		Ti.API.info("@@@ onlink() invoked. url: " + e.url);
		var endTime = new Date();
		endTime.setTime(endTime.getTime() + 1000);
		while (endTime >= new Date()) {}
		return (e.url !== "myapp:HelloWorld");
	},
});
window.add(webView);
window.open();
```

The above issue can be solved by using the [WebViewClient.shouldOverrideUrlLoading()](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)) method instead. But the downside to using this method (based on what I remember) is that it only gets called when tapping/clicking on links. It does **not** get called for URLs invoked in JavaScript within the webpage (and I know iOS' UIWebView does not have this limitation). But this might be okay since we've named the property `onlink`.

hansemannn:
@jquick-axway @garymathews Is this PR ready to merge? iOS was just merged.

jquick-axway:
@hansemannn, there is 1 issue left which I've posted above.

garymathews:
@jquick-axway That test case is already handled, I see:
```
I/TiAPI:  @@@ onlink() invoked. url: myapp:HelloWorld
```

PR #9466, opened 2017-09-22
[TIMOB-25332] (6_3_X) iOS 11: Fix table-view estimated heights

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25332

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -244,6 +244,11 @@ static TiViewProxy * FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoin
         _tableView.autoresizingMask = UIViewAutoresizingFlexibleWidth|UIViewAutoresizingFlexibleHeight;
         _tableView.delegate = self;
         _tableView.dataSource = self;
+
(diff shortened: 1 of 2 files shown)


vijaysingh-axway (approved):
CR passed.

hansemannn:
@sgtcoolguy I tried adding unit-tests as well, but as it's an internal change / behavior fix, there are no modified, added or removed API's. Technically, we could detect those by grep'ing for two things:
- `- (void)set<API>:`
- ` valueForKey:`
- ` objectForKey:`
- `- (id/NSString/NSArray/NSNumber/NSDate/NSDictionary)` <- Regex?

In 90 % of the cases, this will detect any new getter or setter applied to the SDK. Unfortunately I am not familiar with writing regex's so far, so that might be a good task to exercise. Do you think it would make sense to apply that to the Dangerfile and file a ticket for it or are there cases where this could fall through?

ewieberappc (approved):
FR Passed. No odd spacing or UI oddities witnessed when using or modifying a listview. Tested using the provided code as well as the ListView test suite.

PR #9481, opened 2017-09-26
[TIMOB-25353] Android: Allow Titanium modules to process module dependencies

garymathews:
##### ~[appc-aar-tools#16](https://github.com/appcelerator/appc-aar-tools/pull/16) NEEDS TO BE MERGED FIRST!~

In an attempt to minimise conflicting Google Play Services library versions I have implemented the ability for Titanium Android native modules to have dependencies on other Titanium Android native modules.

This is to allow dependency on the `ti.playservices` module for modules such as `ti.map` and `ti.admob` for Google Play Services.

This will prevent the need to update each modules Play Services library and negate the need for module developers to include their own Play Services library which could conflict with existing modules.

###### TEST CASE
- Acquire [ti.map-3.3.2](https://github.com/appcelerator-modules/ti.map/releases/download/android-3.3.2/ti.map-android-3.3.2.zip) pre-release
- Add `ti.playservices` and `ti.map` into your `tiapp.xml`
```XML
  <manifest>  
    <application>
      <meta-data android:name="com.google.android.maps.v2.API_KEY" android:value="<YOUR API KEY>"/>
    </application>
  </manifest>
</android>
<modules>
    <module platform="android" version="3.3.2">ti.map</module>
    <module platform="android" version="11.0.40">ti.playservices</module>
</modules>
```
```JS
    Map = require('ti.map'),
    a = Map.createAnnotation({
        title: 'Axway Appcelerator',
        latitude: 37.38935, longitude: -121.980,
        pincolor: Map.ANNOTATION_GREEN
    }),
    map = Map.createView({
        userLocation: true,
        mapType: Map.NORMAL_TYPE,
        animate: true,
        annotations: [a],
        region: {
            latitude: 37.3680, longitude: -121.9145,
            latitudeDelta: 0.1, longitudeDelta: 0.1
        }
    });

map.addAnnotation(a);

win.add(map);
win.open();
```
- Map should display with an annotation

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25353)

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1625,7 +1625,7 @@ process.exit(1);
 			let unresolvedDependencies = [];
 			for (let module of this.nativeLibModules) {
 				const timoduleXmlFile = path.join(module.modulePath, 'timodule.xml'),
-					  timodule = fs.existsSync(timoduleXmlFile) ? new tiappxml(timoduleXmlFile) : undefined;
+					timodule = fs.existsSync(timoduleXmlFile) ? new tiappxml(timoduleXmlFile) : undefined;
 
 				if (timodule && Array.isArray(timodule.modules)) {
 					for (let dependency of timodule.modules) {
@@ -1645,8 +1645,8 @@ process.exit(1);
 								this.cli.tiapp.modules.push({
 									id: dependency.id,
 									version: dependency.version,
-									platform: ['android'],
-									deployType: [this.deployType]
+									platform: [ 'android' ],
+									deployType: [ this.deployType ]
 								});
 
 								unresolvedDependencies.push(dependency);
@@ -1669,8 +1669,8 @@ process.exit(1);
 		 		process.exit(1);
 		 		*/
 
-		 		// re-validate modules
-		 		return this.validateTiModules('android', this.deployType, validateTiModulesCallback.bind(this));
+				// re-validate modules
+				return this.validateTiModules('android', this.deployType, validateTiModulesCallback.bind(this));
 			}
 
 			// check if we have any conflicting jars
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1490,7 +1490,7 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 	}
 
 	return function (callback) {
-		this.validateTiModules('android', this.deployType, function (err, modules) {
+		this.validateTiModules('android', this.deployType, function validateTiModulesCallback(err, modules) {
 			this.modules = modules.found;
 
 			this.commonJsModules = [];
@@ -1640,6 +1640,15 @@ process.exit(1);
 							}
 							if (missing) {
 								dependency.depended = module;
+
+								// attempt to include missing dependency
+								this.cli.tiapp.modules.push({
+									id: dependency.id,
+									version: dependency.version,
+									platform: ['android'],
+									deployType: [this.deployType]
+								});
+
 								unresolvedDependencies.push(dependency);
 							}
 						}
@@ -1647,6 +1656,7 @@ process.exit(1);
 				}
 			}
 			if (unresolvedDependencies.length) {
+				/*
 				let msg = 'could not find required module dependencies:';
 		 		for (let dependency of unresolvedDependencies) {
 		 			msg += __('\n  id: %s  version: %s  platform: %s  required by %s',
@@ -1657,6 +1667,10 @@ process.exit(1);
 		 		}
 		 		logger.error(msg);
 		 		process.exit(1);
+		 		*/
+
+		 		// re-validate modules
+		 		return this.validateTiModules('android', this.deployType, validateTiModulesCallback.bind(this));
 			}
 
 			// check if we have any conflicting jars
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1621,6 +1621,44 @@ process.exit(1);
 			this.modulesNativeHash = this.hash(nativeHashes.length ? nativeHashes.sort().join(',') : '');
 			this.modulesBindingsHash = this.hash(bindingsHashes.length ? bindingsHashes.sort().join(',') : '');
 
+			// check for any missing module dependencies
+			let unresolvedDependencies = [];
+			for (let module of this.nativeLibModules) {
+				const timoduleXmlFile = path.join(module.modulePath, 'timodule.xml'),
+					  timodule = fs.existsSync(timoduleXmlFile) ? new tiappxml(timoduleXmlFile) : undefined;
+
+				if (timodule && Array.isArray(timodule.modules)) {
+					for (let dependency of timodule.modules) {
+						if (!dependency.platform || /^android$/.test(dependency.platform)) {
+
+							let missing = true;
+							for (let module of this.nativeLibModules) {
+								if (module.id === dependency.id) {
+									missing = false;
+									break;
+								}
+							}
+							if (missing) {
+								dependency.depended = module;
+								unresolvedDependencies.push(dependency);
+							}
+						}
+					}
+				}
+			}
+			if (unresolvedDependencies.length) {
+				let msg = 'could not find required module dependencies:';
+		 		for (let dependency of unresolvedDependencies) {
+		 			msg += __('\n  id: %s  version: %s  platform: %s  required by %s',
+		 				dependency.id,
+		 				dependency.version ? dependency.version : 'latest',
+		 				dependency.platform ? dependency.platform : 'all',
+		 				dependency.depended.id);
+		 		}
+		 		logger.error(msg);
+		 		process.exit(1);
+			}
+
 			// check if we have any conflicting jars
 			const possibleConflicts = Object.keys(jarHashes).filter(function (jar) { return jarHashes[jar].length > 1; }); // eslint-disable-line max-statements-per-line
 			if (possibleConflicts.length) {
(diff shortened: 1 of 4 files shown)


garymathews:
After discussion with @jquick-axway it's probably best this exists as an external module.

benbahrenburg:
@garymathews if implemented as an external module will this still remove the conflicts that module developers have?  Is there anyway we can have a standard which we (module developers) build against?

jquick-axway:
@benbahrenburg, the idea is to wrap the Google Play Services main JAR in its own module and to have other modules which depend on it (ex: ti.map) remove the Google Play Services JAR file from their "lib" directory. It'll be our responsibility to keep the Google Play Services module up-to-date. So, this will only resolve dependency conflicts if other modules "play ball", remove their Google Play Services JAR, and instead document that they require this new module as a dependency.

For example, if an app developer wanted to use the "ti.map" module, the developer would have to add the "ti.map" module and the Google Play Services module in order for it to work. However, doing so with an existing module would be a breaking change, so, we'll look into providing a sensible warning message when attempting to a build while a module dependency is missing. May involve having adding a new feature to "timodule.xml" to reference another module as a dependency.

As an added benefit to making Google Play Services its own module, we'll likely expose the APIs needed to ensure that it's installed and enabled on the device as documented by Google below. My impression is that most modules are not doing this right now, which is why we're holding on to an old version of this library. So, exposing these APIs is needed since the Google Play Services library doesn't contain implementation for most of its functionality and instead does inter-process communications with the Google Play app to request it to perform the needed operations.
- https://developers.google.com/android/guides/setup#ensure_devices_have_the_google_play_services_apk
- https://developers.google.com/android/reference/com/google/android/gms/common/GoogleApiAvailability

What do you think?

benbahrenburg:
@jquick-axway sounds like a good plan.   I had tried to keep the different versions of Play straight using this repo https://github.com/benbahrenburg/Ti-Google-Play-Services but it never really worked out.

Having one place that everyone could get the latest version to compile against I think would be a welcome addition.

garymathews:
@janvennemann Updated PR

jquick-axway (approved):
Looks good to me.

@janvennemann, do you have any feedback you want to give?

android/cli/commands/_buildModule.js
```diff
@@ -293,6 +283,76 @@ AndroidModuleBuilder.prototype.initialize = function initialize(next) {
...
+				});
+			}
+		}, this);
+	}
+
+	// module java archive paths
+	this.jarPaths = [ path.join(this.platformPath, 'lib'), path.join(this.platformPath, 'modules'), this.platformPath ];
+
+	// module dependencies java archive paths
+	for (var module of this.modules) {
```
janvennemann: This should not be necessary as the aar hook already loops over all .jar files here: https://github.com/garymathews/titanium_mobile/blob/ac64609be01f8caeeb5baf68c760fe8350bdcd45/android/cli/hooks/aar-transform.js#L305

android/cli/commands/_buildModule.js
```diff
@@ -293,6 +283,76 @@ AndroidModuleBuilder.prototype.initialize = function initialize(next) {
...
+							m.version = latest;
+							m.path = path.join(modulePath, m.version);
+							return true;
+						}
+					}
+					return false;
+				}.bind(this);
+
+			if ((fs.existsSync(localPath) && getModulePath(localPath)) ||
+				(fs.existsSync(globalPath) && getModulePath(globalPath))) {
```
janvennemann: I like to move longer if checks like this into a simple function with a fitting name so it's easier to understand what it does.

janvennemann (changes requested):
Just a few minor notes regarding code style. Apart from that it looks good to me. Nice feature!

janvennemann:
@garymathews, what about when a modules defines dependencies but they are not present? As i understand it, currently this will just be silently ignored. Maybe we should print a nice error message instead of just continuing the build, knowing that it will probably fail later on anyway.

jquick-axway:
When the build can't find a module in "tiapp.xml", we see the following error messages logged:

```
[WARN] :   Could not find a valid Titanium module id=ti.whatever version=latest platform=android deploy-type=test
[ERROR] :  Could not find all required Titanium Modules:
[ERROR] :     id: ti.whatever	 version: latest	 platform: android	 deploy-type: test
```

If we can't replicate the above for dependency modules as well, then that would be great. (Unless it's doing this already. I haven't tested this code change yet.)

Also, does the build simply log all of the `<module>` element's attributes if the module hasn't been found? If so, then perhaps as a **best-practice** we should add a `url` attribute to it so that developers will know where to download it from.

garymathews:
@janvennemann Updated PR

hansemannn (commented):
@garymathews Can you fix the typo in `LIBRARY_ORIGIN_PORJECT` while you're at it?

janvennemann (approved):
Approved! appc-aar-tools is also updated to 1.1.4

We should probably also make changes to the app build so that it is aware of module dependencies and can print an appropriate error message if a dependency is missing.

garymathews:
@janvennemann Like [this](https://github.com/appcelerator/titanium_mobile/pull/9481/files#diff-e409a7861efa1e711ec49a7ff79beee7R328)?

janvennemann:
@garymathews Yes, but i mean in the app build. Let's say you use a module that has another dependency then you would also require both modules during the app build, right?

janvennemann (changes requested):
@garymathews, can you please check the linting errors reported by Jenkins? Thanks!

android/cli/commands/_build.js
```diff
@@ -1640,13 +1640,23 @@ process.exit(1);
 							}
 							if (missing) {
 								dependency.depended = module;
+
+								// attempt to include missing dependency
+								this.cli.tiapp.modules.push({
+									id: dependency.id,
+									version: dependency.version,
+									platform: ['android'],
```
sgtcoolguy: Do dependencies always have to be explicitly the same platform and native, or could it be commonjs too?
Do we need to specify the apiversion here to be the same as the requirer/SDK, or is that handled by validateTiModules?

android/cli/commands/_build.js
```diff
@@ -1657,6 +1667,10 @@ process.exit(1);
 		 		}
 		 		logger.error(msg);
 		 		process.exit(1);
+		 		*/
+
+		 		// re-validate modules
```
sgtcoolguy: Is there a possible infinite loop here? I see before you'd just give up and report missing dependencies, but now we basically implicitly add them for you and re-try.

android/cli/commands/_build.js
garymathews: - commonjs modules aren't supported as module dependencies
- I tried my best to re-use our current module handling code, so all of the existing module checks will take place. `apiversion` needs to match the SDK
- It will be the same as our current module version handling code (I believe it's exact match only?)

android/cli/commands/_build.js
garymathews: - It looks like that, but our existing checks will prevent that from happening and return before reaching this code again

lokeshchdhry:
FR Passed.

We can now use modules using google `playservices` & module `ti.playservices` together in an app.

SDK Ver:        7.0.0.v20171108114616
OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

benbahrenburg:
@garymathews @jquick-axway is there anything that module developers should be doing to embrace this? For example, how would I add this as a dependency to my module projects. Or is this simply asking the module developers to take the same version of playservices and use it in this modules?

sgtcoolguy:
@benbahrenburg Here's the PR that makes use of ti.playservices as a dependency of ti.map: https://github.com/appcelerator-modules/ti.map/pull/216

I think the goal here is to have ti.playservices be a dependency of any native modules requiring it, and then they'll grab the latest ti.playservices module by default (or if you specify an exact version in your tiapp.xml it'll use that). That way we can avoid having all these modules embed differing versions.

jquick-axway:
@benbahrenburg, we plan on writing up a blog post about this sometime this month to inform 3rd party module developers.

For a more specific example, have a look at our "ti.map" module's "timodule.xml" file (link below) and note the new `<modules>` section. This is where you define your dependencies.
https://github.com/garymathews/ti.map/blob/9b3c4120cfd907468ca3a598a93d3c48ff49e9d2/android/timodule.xml#L11

So, when a developer includes your module in their project, our build system will double check that their "tiapp.xml" file also references your module's module dependency too. If it doesn't, then the developer will get a build error stating that they need to also include said module dependency into their project. You can see an example of what a "tiapp.xml" file will look like with the "ti.map" module and the new Google Play Service module "ti.playservices" at the top of this pull request.
https://github.com/appcelerator/titanium_mobile/pull/9481#issue-260779392

The plan is to always bundle the new "ti.playservices" dependency module with Titanium so that developer's won't have to download/install it themselves.

sgtcoolguy:
@jquick-axway Just to note: I asked Gary to modify the behavior of handling module dependencies so that we will attempt to automatically/implicitly add the dependency as if the user had done so in their tiapp.xml (but if they specify it in the tiapp.xml that works too). This was to avoid causing a breaking change to ti.map usage for existing apps (suddenly they'd stop working and the user would have to explicitly add ti.playservices to the app to get it working again).

PR #9482, opened 2017-09-27
[TIMOB-25287] (6_3_X): iPhone X - Ti.UI.SearchBar in list-view get wrongly positioned

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25287

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -2171,9 +2171,11 @@ static TiViewProxy * FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoin
                                animated:NO
                              completion:^{
                                CGFloat topMargin = self.frame.origin.y;
+#if IS_XCODE_9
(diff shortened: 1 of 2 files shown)


hansemannn (approved):

vijaysingh-axway:
@hansemannn Can you review last commit please. It is simpler solution. As we are setting contentInsetAdjustmentBehavior property of tableView UIScrollViewContentInsetAdjustmentNever, similarly setting searchTableView property will fix the problem . No need to adjusting using safeAreaInsets. Thanks!

hansemannn:
Will do! Please rebase the whole PR to the base-branch and only add the new lines. Right now, the PR still has changes to the indentation caused by the master PR. Ideally, the PR should only have 6 lines changed (2 if-statements with the content-insets change).

vijaysingh-axway:
Rather rebasing I restored indentation, which fixes previously existing spaces etc. Though there are few more lines addition and deletion, I think it will be good.

PR #9484, opened 2017-09-27
[TIMOB-18441] Add `--sigalg` option on cli for overriding keystore signing algorithm.

gimdongwoo:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18441

**Optional Description:**

`ti build -p android -T dist-playstore --sigalg SHA1withRSA`

This is a very simple changes.
But, sometimes this is necessary.
Therefore, this is no reason not to provide.

Thank you.


cb1kenobi:
I'm fine with this feature, but this CLI option needs to be explicitly declared so that it shows up in the help, must have validation, and must have a default value.

gimdongwoo:
@cb1kenobi Ok, I will update soon!

android/cli/commands/_build.js
```diff
@@ -878,6 +878,12 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 						order: 120,
 						required: true,
 						values: _t.targets
+					},
+					'sigalg': {
```
cb1kenobi: Can you please add a `default` property with a value of `'MD5withRSA'` so that it signifies the default value in the help screen.

cb1kenobi (changes requested):
This is looking *really* good, just one small change.

android/cli/commands/_build.js
gimdongwoo: Ok, good!

gimdongwoo: Done! Thank you :)

cb1kenobi: Oh boy, I messed up. Specifying a `default` will properly show the correct default in the help screen, but then `this.sigalg` will always have a value and it never falls back to the auto-detected `this.keystoreAlias.sigalg`.

My apologies. Please remove the `default` value.

cb1kenobi (changes requested):

android/cli/commands/_build.js
gimdongwoo: Oh! It's really easy thing. I have committed. Thank you for your kind review. 👍

cb1kenobi (approved):

cb1kenobi:
Looks great! Thank you very much!

PR #9485, opened 2017-09-28
[TIMOB-25301] (6_3_X):iPhone X Need to be able to control the Insets / Layout Margins

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25301

### DIFF
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -441,20 +441,19 @@ properties:
 
   - name: extendSafeArea
     summary: |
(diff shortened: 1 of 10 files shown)


hansemannn:
Unfortunately one of my tests failed: When opening a modal-window with a button attached with right: 20 and top: 20, it will first use the full-margins and then reposition itself to include the safe-area insets. I'll send you a screen-recording on Slack!

vijaysingh-axway:
@hansemannn  Good catch! Thanks for video. Now fixed.

hansemannn (approved):

PR #9486, opened 2017-09-28
[TIMOB-25355] iOS: Update Ti.Facebook iOS to 5.6.0

hansemannn:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-25355


vijaysingh-axway (commented):
In corresponding 6_3_X PR https://github.com/appcelerator/titanium_mobile/pull/9487, I can see one more change . Just curious, is that required as same is not made here.

hansemannn:
@vijaysingh-axway In 6_3_X, there was an old file remaining in the directory that was now cleaned up to sync with master.

PR #9495, opened 2017-10-02
[TIMOB-25358]:iOS Problem with Searchbar, SDK 6.2.0 and later

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25358


iphone/Classes/TiUITableView.m
```diff
@@ -2667,10 +2667,24 @@ - (void)presentSearchController:(UISearchController *)controller
   [viewController presentViewController:controller
                                animated:NO
                              completion:^{
+                               float leftMargin = 0;
+                               float topMargin = 0;
+                               UIView *tempView = self;
+                               while (tempView != viewController.view) {
+                                 leftMargin += tempView.frame.origin.x;
+                                 topMargin += tempView.frame.origin.y;
+                                 tempView = tempView.superview;
```
hansemannn: If the padding is 100px, you are re-assigning a UIView instance a hundred times. We cannot use a while-loop for this.

hansemannn (changes requested):

iphone/Classes/TiUITableView.m
vijaysingh-axway: The while loop will run as many times as many superview are there till its window containing controller's view are found. So it will not run 100 times for a padding of 100px. Rather it will run for as many super views make it 100px. resultViewController is presented on window holding controller, so we have to calculate this padding.

iphone/Classes/TiUIListView.m
```diff
@@ -2154,10 +2154,17 @@ - (void)presentSearchController:(UISearchController *)controller
   [viewController presentViewController:controller
                                animated:NO
                              completion:^{
+                               CGPoint convertedOrigin = [self.superview convertPoint:self.frame.origin toView:viewController.view];
```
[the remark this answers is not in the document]
vijaysingh-axway: Yes :)

hansemannn (approved):

PR #9496, opened 2017-10-02
[TIMOB-25358](6_3_X):iOS Problem with Searchbar, SDK 6.2.0 and later

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25358

### DIFF
--- a/iphone/Classes/TiUITableView.m
+++ b/iphone/Classes/TiUITableView.m
@@ -1644,8 +1644,8 @@
 
 	if (tableHeaderView == nil)
 	{
-    CGFloat wrapperHeight = [TiUtils isIOS11OrGreater] ? 56.0 : 44.0;
-    CGRect wrapperFrame = CGRectMake(0, 0, [tableview bounds].size.width, wrapperHeight);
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUITableView.m
```diff
@@ -1639,7 +1639,8 @@ -(void)updateSearchView
 
 	if (tableHeaderView == nil)
 	{
-		CGRect wrapperFrame = CGRectMake(0, 0, [tableview bounds].size.width, TI_NAVBAR_HEIGHT);
+    CGFloat wrapperHeight = [TiUtils isIOS11OrGreater] ? 56.0 : 44.0;
```
hansemannn: No hardcoded Values please.

hansemannn (changes requested):

iphone/Classes/TiUITableView.m
vijaysingh-axway: You mean I should use macro for same similar to previous implementation ?

hansemannn: Yep. Or if possible get the height from the actual parent frame. Hardcoded values are the worst, we should try to avoid it at any time.

ewieberappc:
FR Passed. Searchbar search results correctly appear on top of the view and not shifted to the side. Tested using the test cases and the searchbar suite.

PR #9497, opened 2017-10-02
Revert "Update modules.json"

ewanharris:
As both 6_3_X and master began to fail after this merge, just reverting to see if Jenkins runs through using the previous version.


ewanharris:
Looks to still be picking up 2.13.0 on Jenkins `[INFO] :   Found better matching module id=ti.map version=latest platform=ios,iphone deploy-type=development path=/Users/build/Library/Application Support/Titanium/modules/iphone/ti.map/2.13.0` we probably need to clean up modules in the same step as cleaning up the SDKs. 

Was able to reproduce myself and filed https://jira.appcelerator.org/browse/MOD-2365 so closing this

PR #9499, opened 2017-10-02
[TIMOB-25346] (6_3_X) iOS: Restore SBJSON for encoded strings

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25346


vijaysingh-axway (approved):
CR and FT passed.

ewieberappc (approved):
FR Passed. Able to use nested JSON payload in fireEvent without issue. Tested using provided code as well as the webview suite.

PR #9500, opened 2017-10-02
[TIMOB-25363] Android: Use ContentProvider for Intent data

garymathews:
- Use `ContentProvider` for `Intent` `DATA`

###### TEST CASE
```JS
var extFile = Ti.Filesystem.getFile(Ti.Filesystem.externalStorageDirectory, 'test.pdf'),
    openPDF = function(pdf) {
        Ti.Android.currentActivity.startActivity(
            Ti.Android.createIntent({
                action : Ti.Android.ACTION_VIEW,
                type : 'application/pdf',
                data : pdf.nativePath
            })
        );
    };

if (extFile.exists()) {
    openPDF(extFile);
} else {
    var httpClient = Titanium.Network.createHTTPClient({
        onload: function() {
            extFile.createFile();
            extFile.write(this.responseData);
            openPDF(extFile);
        }
    });
    httpClient.open('GET','https://github.com/mozilla/pdf.js/raw/master/examples/helloworld/helloworld.pdf');
    httpClient.send();
}
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25363)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/io/TiFileProvider.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/io/TiFileProvider.java
@@ -24,20 +24,23 @@ public class TiFileProvider extends ContentProvider {
 		return true;
 	}
 
-	public static Uri createUriFrom(File file) {
+	public static Uri createUriFrom(String filePath) {
 		final TiApplication tiApp = TiApplication.getInstance();
 		if (tiApp == null) {
 			return null;
 		}
-
 		try {
-			String path = getUriPrefix() + file.getAbsolutePath();
+			String path = getUriPrefix() + filePath.replaceAll("^file://(/android_asset)?", "");
 			return Uri.parse(path);
 		} catch (Exception e) {
 			// ignore exception
 		}
 
 		return null;
+	}
+
+	public static Uri createUriFrom(File file) {
+		return createUriFrom(file.getAbsolutePath());
 	}
 
 	@Override
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/IntentProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/IntentProxy.java
@@ -17,6 +17,7 @@ import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.kroll.common.Log;
+import org.appcelerator.titanium.io.TiFileProvider;
 import org.appcelerator.titanium.TiApplication;
 import org.appcelerator.titanium.TiBlob;
 import org.appcelerator.titanium.TiC;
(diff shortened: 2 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/proxy/IntentProxy.java
```diff
@@ -191,12 +191,14 @@ public void handleCreationDict(KrollDict dict)
...
-				intent.setDataAndType(Uri.parse(data), type);
+				intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+				intent.setDataAndType(TiFileProvider.createUriFrom(data), type);
 			} else {
 				intent.setType(type);
 			}
 		} else if (data != null) {
-			intent.setData(Uri.parse(data));
+			intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+			intent.setData(TiFileProvider.createUriFrom(data));
```
lokeshchdhry: Missing import:
`import org.appcelerator.titanium.io.TiFileProvider;`

android/titanium/src/java/org/appcelerator/titanium/io/TiFileProvider.java
```diff
@@ -24,14 +24,13 @@ public boolean onCreate() {
...
-	public static Uri createUriFrom(File file) {
+	public static Uri createUriFrom(String filePath) {
 		final TiApplication tiApp = TiApplication.getInstance();
 		if (tiApp == null) {
 			return null;
 		}
-
 		try {
-			String path = getUriPrefix() + file.getAbsolutePath();
+			String path = getUriPrefix() + filePath.replaceAll("^file://(/android_asset)?", "");
```
jquick-axway: When we wrote the `TiContentProvider`, we purposely limited the `createUriFrom()` method to only support `File` objects to limit it to only supporting external files on the filesystem. If we were to add support for reading asset files and resource files in the future (and we should), we would have to override the `ContentProvider.openAssetFile()` method in order to access embedded files within the APK as well change the URI prefix to better represent where the file came from (currently hardcoded to "/filesystem").

The test case for this code change is limited to external files. Can we keep the old way of doing it for now?

lokeshchdhry:
FR Passed

No `FileUriExposedException` is thrown & pdf opens successfully.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.9
Appc CLI:       6.2.4
Ti CLI Ver:     5.0.14
Alloy Ver:      1.9.14
Node Ver:       7.10.1
Java Ver:       1.8.0_101

android/titanium/src/java/org/appcelerator/titanium/io/TiFileProvider.java
garymathews: It is still only limited to external files (even though there's regex that cleans our `/android_asset` identifier). I've just added the ability to pass a file path String in addition to the existing File object. Which is more appropriate for passing the `data` String when creating the `Intent` instead of needing to create a `File` object only to grab it's path again.

alg:
Any chance I could add a commit that fixes the same error in email dialogs to this PR?

PR #9508, opened 2017-10-09
[TIMOB-25373] : iOS - Search result view of List did not close when w idow containing list view get closed

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25373


iphone/Classes/TiUITableView.m
```diff
@@ -312,6 +312,11 @@ - (id)init
 
 - (void)dealloc
 {
+  if ([searchController isActive]) {
+    searchController.view.hidden = YES;
+    [searchController setActive:NO];
```
hansemannn: Why not remove it's delegates and nil it as well (like the others)?

vijaysingh-axway: That is already there. See few lines below in this function :) . You will find -

    searchController.searchResultsUpdater = nil;
   searchController.delegate = nil;
   RELEASE_TO_NIL(searchController);

hansemannn: I see this for the list-view, but not the table-view. Am I missing a snippet?

vijaysingh-axway: I think so. You can see at line 338.

hansemannn: I mean list-view, I cannot see nil'ing `searchResultsUpdater ` in `TiUIListView.m`.

vijaysingh-axway: Ok. You are right. Done.

hansemannn (approved):

hansemannn:
@ewieberappc Build is passing now! Merging here for consistency.

PR #9510, opened 2017-10-09
[TIMOB-25374] Android: borderColor not visible on Android tiSDK 6.2.2

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25374

**Description:**
Set default border width of 1 unit. Change in documentation.

https://github.com/appcelerator/titanium_mobile/pull/9507 needs to be merged first for this change to have a point.

**Test case:**
```
var view = Ti.UI.createView({width: 200, height: 200, backgroundColor: 'white', borderColor: 'red'});
win.add(view);
win.open();
```


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1427,14 +1427,18 @@ private void initializeBorder(KrollDict d, Integer bgColor)
 					borderView.setColor(TiConvert.toColor(d, TiC.PROPERTY_BORDER_COLOR));
 				}
 
+				//Have a default border width of 1
+				Object borderWidth = "1";
```
jquick-axway: I've confirmed that Titanium for iOS defaults the `borderWidth` to "1dp" (Apple's framework takes UI measurements in points/dips by default). This can be seen in the `[TiUIView setBorderColor]` method here...
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiUIView.m#L483

So, I recommend that you change "1" to "1dp" here to match iOS' behavior.

jquick-axway (changes requested):

apidoc/Titanium/UI/View.yml
```diff
@@ -1223,7 +1223,7 @@ properties:
   - name: borderWidth
     summary: Border width of the view.
     type: Number
-    default: 0
+    default: 1
```
hansemannn: Not sure here. It is 0 by default, but 1 if any related border is set. So if no border-related properties are set, I think it does not set 1 on iOS (cc @jquick-axway).

ypbnv: @hansemannn Fair enough. Maybe adding as a description that we set default width to 1 if any border related property is set?

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
ypbnv: @jquick-axway I set it as a String with value '1' to use the default measure unit that is set in tiapp.xml. I think iOS does the same here:
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiDimension.m#L103

apidoc/Titanium/UI/View.yml
garymathews: I thought it only gets set to `1` when `borderColor` is set. It is not set to `1` when `borderRadius` is set on either platforms?

hansemannn: That is correct, sorry for the confusion. Quite simple implementation on iOS:
```objc
- (void)setBorderColor_:(id)color
{
  TiColor *ticolor = [TiUtils colorValue:color];
  self.layer.borderWidth = MAX(self.layer.borderWidth, 1);
  self.layer.borderColor = [ticolor _color].CGColor;
}
```

garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

If no `borderWidth` is set & `borderColor` is specified then `borderwidth` will automatically be `1`.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

PR #9514, opened 2017-10-10
[TIMOB-25322] (6_3_X) iOS: Geolocation should be able to handle iOS 11 permission upgrade 

hansemannn:


hansemannn:
@jvandijk Here is the 6.3.x backport that you may want to use already.

jvandijk:
Reviewed using the  RC release. Working as expected! Nice work @hansemannn

PR #9519, opened 2017-10-11
[6_3_X][TIMOB-25363] Android: Use ContentProvider in EmailDialog

alg:
Fixing the issue with EmailDialog file attachments being sent as direct URI's. This update replaces them with ContentProvider URI's

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25363


hansemannn:
Hey @alg, thank you for your contribution! Can you please push to **`master`** first? We'll then decide whether to backport it to **`6_3_X`** or not. The reason here is that 6.3.0 is already close to the RC-release and we are not planning to take any new tickets into account at this point. Anyway, we are happy to consider it if there is a master-PR, thank you!

**cc** @garymathews @jquick-axway

alg:
Hi @hansemannn. Sure, here: https://github.com/appcelerator/titanium_mobile/pull/9520

PR #9523, opened 2017-10-12
[TIMOB-25300] (6_3_X) iOS 11: Fix title-control in landscape

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25300


ewieberappc:
FR Passed. Able to rotate device and have the title controls centered and without being cut off. Tested using provided sample as well as the title bar suite

PR #9524, opened 2017-10-12
[TIMOB-25384]Android: extendBackground extends the screen below soft system navigation

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25384

**Description:**
Fix window flags for `extendBackground` property to not extend the window content behind the soft system navigation controls.

**Test case:**

A theme with disabled default ActionBar is required. Like 'Theme.AppCompat.NoTitleBar'.

```
var win = Ti.UI.createWindow({backgroundColor: '#ACDC77', theme: 'Theme.AppCompat.NoTitleBar'});
var toolbar = Ti.UI.createToolbar({backgroundImage: '/pattern.png', extendBackground: true, top: 0, width: Ti.UI.FILL});
var viewBot = Ti.UI.createView({height:50, backgroundColor: 'red', bottom: 0});
var viewTop = Ti.UI.createView({height:50, backgroundColor: 'blue', bottom: 50});
  color: '#336699',
  top: 80, left: 10,
  width: 250, height: 60
});
win.add(textField);
win.activity.supportToolbar = toolbar;
win.add(toolbar);
win.add(viewBot);
win.add(viewTop);
win.open();
```
Current behavior:

Intended behavior:
pr

pattern.png
pattern


jquick-axway:
@ypbnv, I don't think I like the `extendBakground` property changing the window flag to fullscreen. While it does change the window to behave more like iOS (top status bar overlays on top of the window), this will introduce a couple of issues.

**Issue 1:** I have seen some phones mishandle the fullscreen flag where it wrongly docks a TextField's clipboard under the top status bar. See the link below for an example screenshot. I'm pretty sure this issue only happens on Android forks (like Samsung devices) where they've implemented their own clipboard bar. Unfortunately, I've never found a work-around for this type of issue other than to avoid using the fullscreen flag. My impression is that the fullscreen flag is under-tested by Android device manufacturers.
https://forums.coronalabs.com/uploads/monthly_03_2015/post-292957-0-66207400-1425274624.png

**Issue 2:** Setting the fullscreen flag also causes the bottom virtual keyboard to be overlaid on top of the app window instead of it resizing the activity window to fit between the top status bar and bottom keyboard (the default and most common Android behavior). The consequence of this is that it can overlay on top of TextFields/TextAreas that are actively being edited. The `adjustSize` soft input mode is no longer relevant and I believe you have to switch to `adjustPan` instead in order for the window to automatically keep the TextField/TextArea in view... but this doesn't work for edit fields in a WebView.

Perhaps a better solution is to do the following:

- Change `extendBackground` to not change the window flags and instead only have it increase in height if it detects the statusbar is overlaid on top. You would have to check the window flags to do this.
- Add a `statusBarColor` property to our `Ti.UI.Android.Activity` class which will give devs the ability to change the color that matches the toolbar. This is how most native Android devs do it anyways, but it doesn't allow for bitmap backgrounds. (iOS does not offer an equivalent API, although, I suppose a hack'ish way to do it is to put a colored rectangle under a translucent style statusbar.)

I'm not aware of any Titanium APIs that sets the fullscreen flag. Our [Window.fullscreen](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.Window-property-fullscreen) property and AppCompat fullscreen theme hide the top statusbar instead. Perhaps we could offer a setting to have the statusbar overlay on top of the window, but I don't like the idea due to Issue1 mentioned above. I'd rather avoid implementing features that we can't make reliable.

Hmm... I feel like this may require further discussion offline.

CC: @hansemannn

jquick-axway:
@ypbnv, I've confirmed that a Motorola "Moto G" device has the clipboard issue I've mentioned above. We have this device in the office here for testing. Titanium 6.3.0 definitely shows the same issue I've posted with my screenshot where the clipboard toolbar appears under the statusbar. However, this is really only an issue on Android 4.x Motorola devices. This issue seems tolerable on Android 5.0 since you set the statusbar to transparent making the clipboard controls fully visible, even though some of the clipboards icons partially collide with the statusbar icons. But if we were to implement an `Activity.statusBarColor` property like I suggested, then the Motorola clipboard bar will be clipped again if use an opaque color. Hmm...

Also, you might want to experiment with using Google's `View.setFitsSystemWindow()` method to when the `extendBackground` property is set true/false. This method is supposed to account for the top statusbar when put into fullscreen mode. (Note that I've never used it before. This is based on the docs I've read.)
https://developer.android.com/reference/android/view/View.html#setFitsSystemWindows(boolean)

jquick-axway:
One more thing, you may want to change your `Ti.UI.createWindow()` test code to do the following. This way, QE does not have to modify the `tiapp.xml` to use new NoTitleBar theme. They can simply copy-and-paste your code with no crashes.  😄 

```javascript
var win = Ti.UI.createWindow(
{
	backgroundColor: '#ACDC77',
	theme: 'Theme.AppCompat.NoTitleBar',
});
```

ypbnv:
@jquick-axway I agree that providing an unreliable feature is not the way to go.
I have tried to come up with a different approach with the UI flags to get the same or similar results, but they all depend on the Fullscreen mode. I will see if I can take use `View.setFitsSystemWindow()` to work around that.

If we can't combine `extendBackground` with `backgroundImage` on iOS, then the trouble we might introduce with it on Android is not worth it. We can keep the parity with your proposal for adding `statusBarColor` support for Activities and changing it according to the current color of the Toolbar - if that makes sense to save the user a line of code. 

In general we might address the support of drawing status bar/system navigation bar over the Window's content.

jquick-axway:
@ypbnv, good news. I've tried your PR's code change and have confirmed that the Motorola's clipboard doesn't collide with the statusbar anymore. I'm guessing the `SYSTEM_UI_FLAG_LAYOUT_STABLE` and `SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN` correctly compensate for this. However, it does shift the toolbar down where it is no longer beneath the statusbar and the height of the statusbar ends up being taller than it needs to be (it doesn't need to compensate for the statusbar in this case.

Here is a screenshot of the with the old issue, before your PR's code change:
- [link](https://axwaysoftware-my.sharepoint.com/personal/jquick_axway_com/_layouts/15/guestaccess.aspx?guestaccesstoken=eon8a8bmlFRY%2fAwxbm7fQb8x5w%2bwPxQL%2bN5dLHpmXFA%3d&docid=2_1670116d0f44542fca69def94409278f4&rev=1)

Here is a screenshot with your PR's code. This looks good:
- [link](https://axwaysoftware-my.sharepoint.com/personal/jquick_axway_com/_layouts/15/guestaccess.aspx?guestaccesstoken=xxvIBVlYOPioxI5dOUbLaUwlMRi5MiPA8TmEbBRkU3w%3d&docid=2_13a201db432b4481ca8fc2aca685d2002&rev=1)

So, the only issue here is with the virtual keyboard covering up the TextField. If you put the TextField at the bottom of the screen and tap on it, the virtual keyboard may cover it up. If you use the `SOFT_INPUT_ADJUST_PAN` softInputMode, then it shifts the screen up to reveal the TextField... but... the top toolbar can no longer be seen since it is panned off of the screen. This appears to be a common issue brought up on stackoverflow. Also, the Motorola's top clipboard is shifted off of the screen as well. I'm curious if the fitSystemWindows option will resolve this.

jquick-axway:
@ypbnv, you know what... setting the `window.fullscreen` property to `true` has the same `adjustPan` issue I mentioned above. Meaning, setting the window `fullscreen` property to `true` and setting the `extendBackground` to `false` will yield the same behavior where the top toolbar will be panned off of the screen. Since this is pre-existing behavior and developers seem to be okay with it (or at least I'm not seeing any complaints), then I'm okay with it too.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
```diff
@@ -120,14 +121,16 @@ public void setToolbarExtendBackground() {
...
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+			Window window = TiApplication.getAppCurrentActivity().getWindow();
+			//Calculate Status bar's height
+			int statusBarHeight = calculateStatusBarHeight();
+			//Add padding to extend the toolbar's background
+			toolbar.setPadding(toolbar.getPaddingLeft(),statusBarHeight + toolbar.getPaddingTop(), toolbar.getPaddingRight(),toolbar.getPaddingBottom());
+			//Set flags for the current window that allow drawing behind status bar
+			window.getDecorView().setSystemUiVisibility(View.SYSTEM_UI_FLAG_LAYOUT_STABLE | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN);
+			window.setStatusBarColor(Color.TRANSPARENT);
+		}
```
jquick-axway: We should remove the `setPadding()` call and instead have the OS automatically adjust for the top statusbar via a call to `View.setFitsSystemWidnows(true)` like this...

```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
	// Set flags for the current window that allow drawing behind status bar
	toolbar.setFitsSystemWindows(true);
	Window window = TiApplication.getAppCurrentActivity().getWindow();
	window.getDecorView().setSystemUiVisibility(View.SYSTEM_UI_FLAG_LAYOUT_STABLE | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN);
	window.setStatusBarColor(Color.TRANSPARENT);
}
```

I've confirmed that the above works on a Motorola "Moto G" (Android 5.0) while the TextField top docked clipboard is shown. See the following screenshots for proof. I'm using a purple toolbar background instead of an image for this test:
- Without TextField clipboard:  [Screenshot](https://axwaysoftware-my.sharepoint.com/personal/jquick_axway_com/_layouts/15/guestaccess.aspx?guestaccesstoken=Raqw73ddGLp5lFce0fKvZ91AXWJusmh%2fyladZHasv4k%3d&docid=2_13834d04a1eab468da630fd51d079a7a1&rev=1)
- With TextField clipboard shown:  [Screenshot](https://axwaysoftware-my.sharepoint.com/personal/jquick_axway_com/_layouts/15/guestaccess.aspx?guestaccesstoken=zGfvxs6l8GpWC7t4HelbzOTsHDPm%2bDQ9Aj6o7Yl3Mmc%3d&docid=2_1da83274cedd34b4e97daa4788ac555b0&rev=1)

I've tested the above code on a Pixel XL (Android 7.1) device as well. The toolbar `extendBackground` correctly covers the top statusbar area on it too.

jquick-axway (changes requested):

ypbnv:
@jquick-axway Updated.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

`extendBackground ` does not extend beyond the soft system navigation.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

jquick-axway:
@lokeshchdhry, we also need to do the following tests as well:
- Test on a Motorola "Moto G", making sure that the TextField's top docked copy/paste clipboard doesn't overlap the status bar and the toolbar is sized correclty.
- Test on an Android 4.x device and make sure that the toolbar size doesn't change. (The top status bar cannot be made transparent on 4.x.)

lokeshchdhry:
@jquick-axway , Thanks, Checked on MotoG & the copy/paste clipboard does not overlap the status bar. Also, the toolbar is sized correctly.
For, Android 4.X top status bar is not transparent & toolbar does not change size.

ypbnv:
@jquick-axway I can't come up with some sort of unit testing here, so I suggest if you are OK with the current state and the FR results that we request a force merge from an admin ?

lokeshchdhry:
@ypbnv , If possible can you please add some unit tests for this PR. Merge is blocked.

ypbnv:
@lokeshchdhry I was not able to come up with something reliable.

PR #9526, opened 2017-10-16
[TIMOB-24266] Expose application-delegates to Hyperloop / native modules

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24266

### DIFF
--- a/iphone/Classes/TiUIWindowProxy.m
+++ b/iphone/Classes/TiUIWindowProxy.m
@@ -942,7 +942,6 @@
     return;
   }
 
-  [[controller navigationController] setToolbarHidden:!hasToolbar animated:YES];
   //Need to clear title for titleAttributes to apply correctly on iOS6.
   [[controller navigationItem] setTitle:nil];
   SETPROP(@"titleAttributes", setTitleAttributes);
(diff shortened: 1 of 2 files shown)


vijaysingh-axway (commented):
It will be good if we mention this capability in document also.

hansemannn:
Good point, will draft a document for it today!

hansemannn:
@vijaysingh-axway The guide is now drafted (link later). Can you do the review already?

iphone/Classes/NetworkModule.h
```diff
@@ -23,7 +23,7 @@ typedef enum {
   TLS_VERSION_1_2
 } TLSVersion;
 
-@interface NetworkModule : TiModule {
+@interface NetworkModule : TiModule <UIApplicationDelegate> {
```
vijaysingh-axway: Every thing looks good and working fine. One question -
If we have to use the UIApplicationDelegate methods in module, we have to include UIApplicationDelegate in module interface like below -

`@interface TestAppceleratorModule : TiModule<UIApplicationDelegate>
@end`

Isn't it better to include UIApplicationDelegate in TiModule ?

hansemannn: The main reason was that I tried to not set the delegate if it's not actually used.

vijaysingh-axway (approved):
CR passed.

PR #9527, opened 2017-10-16
[TIMOB-25397] Ti.Database.execute() should return null with no results

garymathews:
- `Titanium.Database.execute()` should return null when no results are found
- The tests are failing as it is counting the fields and not the rows returned; where default fields exist when the database is created. ([test](https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.database.test.js#L343))

###### TEST CASE
```JS
var db = Ti.Database.install('made.up.sqlite', 'category'),
    rows = db.execute('pragma table_info(\'category\');');

// rows should be `null` since there are no results
Ti.API.info('rows: ' + JSON.stringify(rows, null, 2));

db.remove();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25397)

### DIFF
--- a/iphone/Classes/TiDatabaseProxy.m
+++ b/iphone/Classes/TiDatabaseProxy.m
@@ -194,7 +194,7 @@
 
   PLSqliteResultSet *result = (PLSqliteResultSet *)[statement executeQuery];
 
-  if ([[result fieldNames] count] == 0) {
+  if ([result fullCount] == 0) {
     [result next]; // we need to do this to make sure lastInsertRowId and rowsAffected work
     [result close];
     return [NSNull null];
--- a/android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
+++ b/android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
@@ -111,7 +111,7 @@ public class TiDatabaseProxy extends KrollProxy
 					// PRAGMA do. If there are no results, just return null.
 					// Thanks to brion for working through the logic, based off of commit
 					// https://github.com/brion/titanium_mobile/commit/8d3251fca69e10df6a96a2a9ae513159494d17c3
-					if (c.getColumnCount() > 0) {
+					if (c.getCount() > 0) {
 						rs = new TiResultSetProxy(c);
 						if (rs.isValidRow()) {
 							rs.next(); // Position on first row if we have data.


iphone/Classes/TiDatabaseProxy.m
```diff
@@ -194,7 +194,7 @@ - (id)execute:(id)args
 
   PLSqliteResultSet *result = (PLSqliteResultSet *)[statement executeQuery];
 
-  if ([[result fieldNames] count] == 0) {
+  if ([result fullCount] == 0) {
```
hansemannn: Although the change looks ok, there will a reason the field-names count has been used in the past. Also this is a breaking change, so we cannot take it before 7.0.0. And does Alloy / Backbone use Ti.Database internally as well? If so, this needs to be checked as well.

garymathews: The previous comments in the `.java` file suggest returning `null` when no rows are returned. But our documentation does not mention this, maybe the best solution is to always return a `ResultSet` and invalidate the test?

hansemannn: I'd rather want to change the test I think. Or we take this one of 7.0.0, don't know.

sgtcoolguy: Just to note, this was brought up due to a failing unit test that checked we returned an empty result (null) in a specific case that broke Alloy usage on Windows Phone.

Specifically, here's the behavior alloy is relying on: https://github.com/appcelerator/alloy/blob/1e881735f86320299ff7642c97afbc7343c908f1/Alloy/lib/alloy/sync/sql.js#L411

Personally, I think we should mimic whatever the database does here: does it throw an error when you ask for info on a table that doesn't exist? Or does it return an empty result set?

My guess is that we don't mimic the exact behavior and have varying behavior per-platform here.

But in any case, we need to keep Alloy's usage in mind and fix it all together so we don't break alloy SQL usage.

garymathews: Looking at [SQLite#pragma](http://www.sqlite.org/pragma.html#pragma_table_info), the result will always contain columns `cid, name, type, notnull, dflt_value, pk` which means

```JS
db.execute('pragma table_info(\'category\');');
```

will always return an empty `ResultSet` containing no rows, but with those six columns. I'm not sure why the test has not failed previously. Running it on Android produces an empty `ResultSet` instead of `null` too. The changes here will amend the test and allow the current alloy code to work as intended. Or we could leave this as-is (I think we should return an empty `ResultSet` instead of `null`) and fix the alloy code.

garymathews:
Closing, solved by https://github.com/appcelerator/alloy/pull/855

iphone/Classes/TiDatabaseProxy.m
garymathews: UPDATE: iOS 11 and Android 8.0 are what cause this behaviour change. Looks like they are running a newer SQLite.

PR #9528, opened 2017-10-16
[TIMOB-25393] Handle rpath search paths as string

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25393

**Optional Description:**
Switches the LD_RUNPATH_SEARCH_PATH build setting to be handled as a space separated string of paths instead of an array of paths. Xcode can handle both, but the string representation is the default one for this build setting so we stick to that.


cb1kenobi (approved):
CR'd. Looks good to me. **APPROVED**

PR #9529, opened 2017-10-16
[TIMOB-25393](6_3_X) Handle rpath search paths as string

janvennemann:
6_3_X backport of #9528


cb1kenobi (approved):
CR'd. Looks good to me. **APPROVED**

PR #9532, opened 2017-10-17
Fix format on UI.yml

ewanharris:
lint stage of build was broke because of this


ewanharris:
Passes linting now, merging

PR #9538, opened 2017-10-19
[TIMOB-25322] (6_3_X) iOS: Improve new Ti.Geolocation authentication-flow

hansemannn:

### DIFF
--- a/iphone/Classes/GeolocationModule.m
+++ b/iphone/Classes/GeolocationModule.m
@@ -312,18 +312,22 @@ extern BOOL const TI_APPLICATION_ANALYTICS;
             "or\n\n"
             "Ti.Geolocation.requestLocationPermissions(Ti.Geolocation.AUTHORIZATION_WHEN_IN_USE, function(e) {\n"
             "\t// Handle authorization via e.success\n"
-            "})\n");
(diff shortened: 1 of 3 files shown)


jvandijk:
Reviewed the implementation! Works as described.

iphone/Classes/GeolocationModule.m
```diff
@@ -1093,6 +1117,15 @@ - (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatu
                 code = 1;
                 errorStr = @"The user denied access to use location services.";
         }
+      
+        // This is very important for iOS 11+ because even if the user did an incremental authorization before
+        // (by selecting "when in use", he/she will still get a dialog that includes the "when in use" option.
+        // In case that one is still selected then, the developer should know about that selection and the following
+        // statement allows that.
+        if (!requestedStatusMatchesActualStatus) {
```
jvandijk: @hansemannn this addition has zero effect, because when the authorization status is unchanged and thus this whole method is not fired.

hansemannn: It's for the initial "when in use" instead of "always" selection, which would fire but with `success: true`.

jvandijk: You're right

ewieberappc (approved):
FR Passed. The WhenInUse key is always required. To have "always" permissions, both Always and AlwaysAndWhenInUse keys are required. Appropriate error messages appear if the required keys are not present.

PR #9548, opened 2017-10-24
[TIMOB-25438] Android: Toolbar tests and drawables loading callback

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25438

**Description:**
Add tests for Toolbar and loading resources from density specific folders in Ti.UI.ImageView and Ti.UI.Toolbar.


tests/Resources/ti.ui.imageview.test.js
```diff
@@ -0,0 +1,367 @@
...
+			}
+		});
+		win.add(view);
+		win.open();
+	});
+
+	(utilities.isAndroid() ? it : it.skip)('image from density specific folder on Android', function (finish) {
+		var window = Ti.UI.createWindow({backgroundColor: 'gray'});
+		var imageView = Ti.UI.createImageView({width: '50%', image: '/images/git.png'});
+		imageView.addEventListener('load', function() {
```
sgtcoolguy: I think this will never fire because the image is local and has loaded in the create method. You may need to hook the listener first and then assign the image property after?

tests/Resources/ti.ui.toolbar.test.js
```diff
@@ -0,0 +1,219 @@
...
+			finish();
+		});
+		window.activity.supportToolbar = toolbar;
+		window.add(toolbar);
+		window.open();
+	});
+
+	(utilities.isAndroid() ? it : it.skip)('overflowIcon', function(finish) {
+		var toolbar = Ti.UI.createToolbar({top: 0, width: Ti.UI.FILL, overflowIcon: Ti.Filesystem.resourcesDirectory + 'Logo.png', barColor: 'blue'});
+		toolbar.addEventListener('resourceLoaded', function(e) {
```
sgtcoolguy: Possibly the same as above, this may not fire if the resource is local and the image gets loaded too quickly...

tests/Resources/ti.ui.toolbar.test.js
```diff
@@ -0,0 +1,219 @@
...
+			should(e.target).eql('navigationIcon');
+			should(e.loaded).eql(true);
+			finish();
+		});
+		window.activity.supportToolbar = toolbar;
+		window.add(toolbar);
+		window.open();
+	});
+
+	(utilities.isAndroid() ? it : it.skip)('overflowIcon', function(finish) {
```
sgtcoolguy: If the test is intended to be android specific (it's an API meant to only be on android and isn't an API that's broken/missing on other platforms), just do:
`it.android('overflowIcon', function (finish) {`

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
```diff
@@ -487,8 +515,16 @@ private void handleSetContentInsetsRelative(Integer values[]) {
...
 
+	private boolean checkResourceLoading(TiDrawableReference source, Drawable result) {
+		return (source != null && result != null && source.compareBitmapWith(result));
+	}
+
 	@Override
 	public void processProperties(KrollDict d) {
+		//process internal properties
+		if (d.containsKey("resourceLoadedListener")) {
+			resourceLoadedListener = ((KrollFunction) d.get("resourceLoadedListener"));
```
garymathews: should validate the object

`instanceof KrollFunction`

garymathews (changes requested):
see above

garymathews:
it also looks like these tests are failing

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
```diff
@@ -56,14 +63,32 @@
...
+			// Check the resource loading once the view is drawn
+			@Override
+			protected void onDraw(Canvas canvas) {
+				super.onDraw(canvas);
+				//if we have attached listener send the dictionary
+				if (resourceLoadedListener != null) {
+					KrollDict callbackParam = new KrollDict();
+					callbackParam.put(TiC.PROPERTY_LOGO, checkResourceLoading(drawableMap.get(TiC.PROPERTY_LOGO), this.getLogo()));
+					callbackParam.put(TiC.PROPERTY_OVERFLOW_ICON, checkResourceLoading(drawableMap.get(TiC.PROPERTY_OVERFLOW_ICON), this.getOverflowIcon()));
+					callbackParam.put(TiC.PROPERTY_NAVIGATION_ICON, checkResourceLoading(drawableMap.get(TiC.PROPERTY_NAVIGATION_ICON), this.getNavigationIcon()));
```
jquick-axway: The checkResourceLoading() method compares images via Bitmap.sameAs(), which is a per-pixel comparison. I'm worried about performance since this code is doing this on every onDraw(). So, I'm guessing you just want to know when the image/drawable is done loading, right?

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -1904,6 +1909,11 @@
 	 */
 	public static final String PROPERTY_LOAD_URL = "loadUrl";
 
+	/**
+	 * @module.api
+	 */
+	public static final String PROPERTY_LOADED = "loaded";
```
jquick-axway: Titanium's JavaScript events typically do not use past-tence (ex: 'ed) or future-tense (ex: 'ing) names. I know, I prefer to write them your way too, but we should be consistent with the rest of the framework.

Our 'Ti.UI.ImageView' type has a `load` event to indicate that an image has finished loaded asynchronously. We may want to do the same for the toolbar.
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.ImageView-event-load

jquick-axway (changes requested):

jquick-axway:
@ypbnv, we should probably push this PR out to 7.1.0.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
ypbnv: Not only when it was loaded, but also if it was shown correctly. That was the problem that required these checks - it could not get a reference to the source when it is placed in the density specific folders and was passing `TiDrawableReference` that was 'empty' (if I can put it that way). It was loaded successfully but had not actual content.
I was not able to come up with anything else that could serve this purpose.

android/titanium/src/java/org/appcelerator/titanium/TiC.java
ypbnv: Removing this completely. It is a leftover from older commits.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
jquick-axway: > It was loaded successfully but had not actual content

Just to be clear, images in the toolbar are being loaded fine, right? The issue is that the callback you're implementing was being invoked before the image was loaded or rendered onscreen?

I'm a bit confused by this. `TiDrawableReference.getDrawable()` will return a BitmapDrawable **after** the image was fully decoded and loaded as a bitmap and it's a blocking call (not async). I don't see how this can go wrong. But I can see the BitmapDrawable being rendered after the callback was invoked. Perhaps you're missing an invalidate() or requestLayout() call after adding views or assigning drawables?

ypbnv: They were not loaded if they were placed in a density specific folder. `TiDrawableReference.getDrawable()` returned `null` in that case which was passed to `Toolbar.setLogo()`. And the idea of this callback is to be used only for testing whether an existing resource was properly loaded as a logo/overflowIcon/navigationIcon. I added it in the `onDraw()` method, because I think that way it will catch any problems with the image, not only if the `getDrawable()` is `null`. I don't imagine it has any use outside of testing purposes. Maybe we chat about that if it does not make sense?

jquick-axway: The `TiDrawableReference.getDrawable()` can definitely load density specific images from the res directory (ex: res/drawable-xhdpi, res/drawable-xxhdpi, etc.). I've done it. The problem is that this is difficult to test via our "TitaniumTest" Android project because Titanium expects the "res" images to be posted-fixed with an MD5 hash. It's not obvious and I think it makes it overly complicated, but the idea was to avoid filename collision under the APK's res folder. I find it easier to test it via Appcelerator Studio since our "appc" build system will automatically copy images under the "res" folder and postfix them with the expected MD5 hash.

Try this. Create a "Classic" app in Appcelerator Studio and then put an image file under the following directories...
`./Resources/android/images/res-mdpi/<YourImage.png>`
`./Resources/android/images/res-hdpi/<YourImage.png>`

Our appc build will copy the above image under the appropriate res folder when generating a Android project under the "build/android" folder. You can access the above density scaled image using the following path in Titanium...
`/images/<YourImage.png>`

ypbnv: It is working now, yes. We had a regression in 6.2.0 for that which was fixed. The main point of the PR is to add tests to prevent it from happening again. The regression affected 3 places - Ti.UI.ImageView, Toolbar logo/overflowIcon/navigationIcon and BigStyleNotification image. I don't think we can have mocha tests for the latter, because they are visible out of JavaScript context. The test for ImageView relies on the `load` event, but for Toolbar's properties that require drawables that was the only reliable lifecycle moment I was able to come up with.
I went through the MD5 hash adding and have successfully reproduced it in TitaniumTest and debug it there. Also I have tried the tests with only this change applied on 6.2.0. The purpose of this event there is entirely for testing and is not expected to be used for anything else.

hansemannn:
@jquick-axway @ypbnv Should this go into 7.0.1? We need passing tests and a 7_0_X backport in order to.

jquick-axway:
@hansemannn, this is **not** going into 7.0.1.

sgtcoolguy:
@garymathews and @jquick-axway Please take another look and re-review.

@ypbnv Looks like there's linting issues.

hansemannn:
@ypbnv Attempted to fix the linting here. Still a mess to do this manually.

ypbnv:
Should be OK now. Although we may reconsider if this is the best way to approach the testing. Maybe we can achieve better results with Appium?

PR #9549, opened 2017-10-24
[6_3_X][TIMOB-25437] Remove const/let usage and use strict

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25437

#9535 Was a bad fix and exposed errors due to the code now running in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). This PR removes the `'use strict'` and usage of const/let. If we backport any fixes in the CLI code from master we need to remember to **not** use const/let.

## Verification

1. Exercise general building of with Node 4 
2. Download the test app from https://github.com/ewanharris/TIMOB-25437 and build. The error should not be seen and the app should build and load as expected.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -1,4 +1,3 @@
-'use strict';
 
 /**
  * iOS build command.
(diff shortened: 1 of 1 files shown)


hansemannn (approved):

mukherjee2 (approved):
Node Version: 6.11.5
NPM Version: 3.10.10
Mac OS: 10.13
Appc CLI: 6.3.0-master.16
Appc CLI NPM: 4.2.10-2
Titanium SDK version: 6.3.0.v20171026142134 (artifact from this PR)
Xcode 9.0.1/iOS 11.0.3 iPhone 7 Plus

I first built the test app with an SDK without the fix, and saw the same error that was mentioned in the ticket. I then built the app with the SDK that has the fix, and saw no errors. The app also functioned as expected.

PR #9552, opened 2017-10-24
[TIMOB-24404] Expose global object on iOS

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24404

**Optional Description:**
Exposes the global object as a node like `global` variable. Allows the following code to be run:
```
global.testVar = 1;
console.log(testVar); // prints "1" to the log
```
This already works in Alloy apps in Android, but not in classic apps.

### DIFF
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -552,6 +552,10 @@ CFMutableSetRef krollBridgeRegistry = nil;
   prop = TiStringCreateWithCFString((CFStringRef) @"console");
   TiObjectSetProperty(jsContext, globalRef, prop, [KrollObject toValue:kroll value:console], kTiPropertyAttributeNone, NULL);
 
+  // Make the global object itself available under the name "global"
+  TiStringRef globalPropertyName = TiStringCreateWithCFString((CFStringRef) @"global");
+  TiObjectSetProperty(jsContext, globalRef, globalPropertyName, globalRef, kTiPropertyAttributeDontEnum | kTiPropertyAttributeReadOnly | kTiPropertyAttributeDontDelete, NULL);
+
   //if we have a preload dictionary, register those static key/values into our namespace
   if (preload != nil) {
     for (NSString *name in preload) {


hansemannn:
@sgtcoolguy I'd like to add the unit-tests for Jan - should I create a new test-suite, e.g. "globals.test.js" or integrate into an existing one? Eventually, we could also test other globals like setTimeout, setInterval, __filename and __dirname.

hansemannn:
@janvennemann I am not sure about the performance-impact of this change, because (for now) we hang every Ti.* element on the global namespace now, which can be referenced from everywhere now. This example:
```js
  backgroundColor: '#fff'
});

  title: 'Trigger'
});

  console.log(global);  
});

win.add(btn);
win.open();
```
returns:
``` js
{
    L = "<KrollCallback: 0x60c000266900>";
    alert = "<KrollCallback: 0x60c0002669c0>";
    btn = "[object TiUIButton]";
    clearInterval = "<KrollCallback: 0x60c0002667c0>";
    clearTimeout = "<KrollCallback: 0x60c000266880>";
    console = "[object TiConsole]";
    require = "<KrollCallback: 0x60c000266a00>";
    setInterval = "<KrollCallback: 0x60c000266b00>";
    setTimeout = "<KrollCallback: 0x60c000266bc0>";
    win = "[object TiUIWindow]";
}
```

sgtcoolguy:
@hansemannn We have some tests of globals int he suite, but they're poorly organized. Right now they live in:
- https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/1102d6e1d7a0da2bb41082b350a6f22d08f449fa/Resources/ti.internal.test.js
- https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/1102d6e1d7a0da2bb41082b350a6f22d08f449fa/Resources/ti.builtin.test.js

I'd copy and update one of those, or combine them into a global.test.js.

hansemannn:
Ok, I'll remove the ones from `ti.internal.test.js` as they are in `ti.builtin.test.js` already, remove `ti.internal.test.js`, rename `ti.builtin.test.js` to `ti.globals.test.js` and add more tests around them. @janvennemann I'll likely cherry-pick your PR and provide it in an own PR.

hansemannn (approved):

janvennemann:
@hansemannn, this should not change the scope of any variables, it only makes the global object available under the `global` var. The reason the button and window variable show up in that list is because every variable declared in `app.js` is considered a global variable since this is the entry point for the js context. Variables in other files are scoped to their module and will not be available via `global`.

@sgtcoolguy Support for this on Android seems to only work under Alloy. Any idea why?

sgtcoolguy:
@janvennemann uhhh, I don't know.  Maybe an artifact of the difference in the way we treat app.js versus all other required JS files?

In non-app.js we wrap in a self-invoking function ala Node: https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/js/kroll.js#L120

app.js is treated specially: https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/js/module.js#L580

sgtcoolguy:
So to follow up on this: 
I'm actually needing this to get #9512 working with ti-mocha. There is no "global" variable/property in iOS, which is causing issues there, where it attempts to preserve some global functions. There are platform difference between iOS and Android so I can't even try and do an `if(!global)` guard before it re-defines global.

As part of that and this PR I looked at what Node does, and it is *exactly* what Jan does here. For Android, `global` works on anywhere but app.js, because in other files we wrap the file in a self invoking function that passes in the global object as `global`. In app.js we just execute the script in context without a self-invoking function wrapper to pass in things like `kroll` or `global`.

To unify the platforms, I think we should hang "global" off the global object as Jan does here, then global should be available everywhere in any JS file for both platforms. We could then remove the global argument to the self-invoking wrapper function on Android.

On another related note, I'm seeing some undocumented/unofficial global weirdness with require too. You can do `this.myNewProperty = 1;` to affect the global namespace right now in a required file. With #9512 that behavior breaks as it inlines `undefined` to replace `this`. Not sure if we care or not.

PR #9553, opened 2017-10-25
[TIMOB-25416] Ambiguous certificates trigger development provisioning…

feons:
… profiles validation error

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25416

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -215,7 +215,7 @@ iOSBuilder.prototype.assertIssue = function assertIssue(issues, name) {
  * @returns {Array}
  * @access private
  */
-iOSBuilder.prototype.findCertificate = function findCertificate(name, type) {
+iOSBuilder.prototype.findCertificates = function findCertificates(name, type) {
 	const certs = [];
 	/* eslint-disable max-depth */
 	if (name && this.iosInfo) {
@@ -1049,9 +1049,9 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
 
 			let certs;
 			if (target === 'device') {
-				certs = _t.findCertificate(cli.argv['developer-name'], 'developer');
+				certs = _t.findCertificates(cli.argv['developer-name'], 'developer');
 			} else {
-				certs = _t.findCertificate(cli.argv['distribution-name'], 'distribution');
+				certs = _t.findCertificates(cli.argv['distribution-name'], 'distribution');
 			}
 
 			const pems = certs.map(function (c) {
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -3167,7 +3182,8 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 		};
 		xobjs.PBXShellScriptBuildPhase[buildPhaseUuid + '_comment'] = '"' + name + '"';
 	} else if (this.target === 'device') {
-		buildSettings.CODE_SIGN_IDENTITY = '"iPhone Developer: ' + this.certDeveloperName + '"';
```
hansemannn: Doesn't this break the build? Always thought it has to have a certain structure.

iphone/cli/commands/_build.js
```diff
@@ -208,14 +208,15 @@ iOSBuilder.prototype.assertIssue = function assertIssue(issues, name) {
...
- * Retrieves the certificate information by name.
+ * Retrieves the list of certificate information by name.
  *
  * @param {String} name - The cert name.
  * @param {String} [type] - The type of cert to scan (developer or distribution).
- * @returns {Object|null}
+ * @returns {Array}
  * @access private
  */
 iOSBuilder.prototype.findCertificate = function findCertificate(name, type) {
```
hansemannn: Should be plural since it returns multiple certificates now.

iphone/cli/commands/_build.js
```diff
@@ -1020,9 +1027,9 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
 			let maxAppId = 0;
 			let pp;
 
-			function prep(a, cert) {
+			function prep(a, certs) {
 				return a.filter(function (p) {
-					if (!p.expired && !p.managed && (!cert || p.certs.indexOf(cert) !== -1)) {
+					if (!p.expired && !p.managed && (!certs || intersection(p.certs, certs).length > 0)) {
```
hansemannn: So your idea is to grab all certificates, intersect them with the the one in the provided by the build?

iphone/cli/commands/_build.js
```diff
@@ -1160,15 +1171,19 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
...
+					certs = _t.findCertificate(cli.argv['distribution-name'], 'distribution');
 				}
 
-				if (cert && p.certs.indexOf(cert.pem.replace(pemCertRegExp, '')) === -1) {
-					return callback(new Error(__('Specified provisioning profile UUID "%s" does not include the "%s" certificate', value, cert.name)));
+				const pems = certs.map(function (c) {
+					return c.pem.replace(pemCertRegExp, '');
+				})
+
+				if (certs.length > 0 && intersection(p.certs, pems).length < 0) {
```
hansemannn: If the intersection returns 0, it is okay?

hansemannn (changes requested):
Left a few questions

iphone/cli/commands/_build.js
feons: fixed.

iphone/cli/commands/_build.js
feons: fixed.

iphone/cli/commands/_build.js
feons: yeah, intersection of list of certs assigned to the profile and all certs returned by findCertificates.

iphone/cli/commands/_build.js
feons: Does it?

cb1kenobi: I know for distribution builds, we HAD to remove the certificate name. Perhaps it's OK for device (developer) builds too? Also, older Xcodes may have issue, but I'm not sure.

hansemannn: It always used to be `iPhone Developer: XXXXXX` for device and `iPhone Distribution: XXXXXX` for dist-builds. If that's not required anymore: Cool! But we should research that before to confirm it actually does in all cases (Xcode 8+9 at least).

hansemannn: Just confirmed on Xcode 9: It can be left blank, so `iPhone Developer` for device (aka `Debug`) and `iPhone Distribution` for dist-builds (aka `Release`).

hansemannn (approved):

hansemannn:
@feons Please backport to 6_3_X, thx!

PR #9555, opened 2017-10-26
[TIMOB-25346] iOS: Fix JSON-parse edge-case when sending view-proxy

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25346

### DIFF
--- a/iphone/Classes/AppModule.m
+++ b/iphone/Classes/AppModule.m
@@ -17,6 +17,7 @@
 #import <UIKit/UILocalNotification.h>
 #import <unistd.h>
 #import "TiLayoutQueue.h"
+#import "SBJSON.h"
 
 extern NSString * const TI_APPLICATION_DEPLOYTYPE;
 extern NSString * const TI_APPLICATION_ID;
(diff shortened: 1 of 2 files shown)


mukherjee2 (approved):
Node Version: 6.11.5
NPM Version: 3.10.10
Mac OS: 10.13
Appc CLI: 6.2.4
Appc CLI NPM: 4.2.10-2
Titanium SDK version: 6.3.0.v20171026120035 (artifact from this P/R)
Xcode 9.0.1/iOS 11.0.3 iPhone 7 Plus
Launched app. Clicked "Trigger". Clicked "Close". There was no crash and app functioned as expected. I tried again with older SDK (affected version), and the app gave a run-time error when I clicked "Close".

mukherjee2:
@hansemannn FR passed, but there are two tasks that need to be done (1) Does it need CR? (2) Merge button is grayed out because of "commit cannot be built" error from Jenkins.

janvennemann (approved):
Looks good, CR passed!

PR #8016, opened 2016-05-24
[TIMOB-24042] iOS / Android: Refactor calendar module

kopiro:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24042

Hello.
Me and @Jei totally refactored this module.
There are breaking changes, so maybe we have to discuss upfront.

PR #9558, opened 2017-10-26
[TIMOB-24042] iOS / Android: Refactor calendar module

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24042

Cherry-picked from #8016.


iphone/Classes/TiCalendarRecurrenceRule.m
```diff
@@ -53,8 +53,12 @@ - (id)valueForUndefinedKey:(NSString *)key
...
   } else if ([key isEqualToString:@"end"]) {
     EKRecurrenceEnd *end = currRule.recurrenceEnd;
-    NSDictionary *recurrenceEnd = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils UTCDateForDate:end.endDate], @"endDate",
-                                                NUMUINTEGER(end.occurrenceCount), @"occurrenceCount", nil];
+    NSDictionary *recurrenceEnd = nil;
+    if (end.occurrenceCount > 0) {
+      recurrenceEnd = [NSDictionary dictionaryWithObjectsAndKeys:NUMUINTEGER(end.occurrenceCount), @"occurrenceCount", nil];
+    } else if (end.endDate != nil) {
+      recurrenceEnd = [NSDictionary dictionaryWithObjectsAndKeys:[TiUtils UTCDateForDate:end.endDate], @"endDate", nil];
+    }
```
hansemannn: @kopiro Hey Flavio, quick question on this one: What happens if the `EKRecurrenceEnd` has both `occurrenceCount` and `endDate`? Won't the `endDate` get lost with this change? Maybe add an `if`-statement to check for both to retain the old behavior as well, adding the partial constructions as well?

apidoc/Titanium/Calendar/Event.yml
```diff
@@ -246,13 +250,19 @@ properties:
...
     platforms: [iphone, ipad]
-    
+
   - name: recurrenceRules
     summary: The recurrence rules for the calendar item. (Available in iOS 5.1 and above.)
     type: Array<Titanium.Calendar.RecurrenceRule>
     osver: {ios: {min: "5.0"}}
     platforms: [iphone, ipad]
 
+  - name: rrule
```
jquick-axway: The `rrule` property is not defined in the Android `EventProxy`.

jquick-axway: Also, iOS' `recurrenceRules` property appears to be "RFC 2445" as well. I think we should dump this property and try to make it portable with the existing rules property for portability.

jquick-axway (changes requested):

jquick-axway:
@hansemannn, perhaps we should split this PR into two since the creatCalendar() handling needs more work on the Android side and due to the looming 7.0.0 release date.

We can move the date deprecation handling to a PR for 7.0.0. The rest of the changes are features that we can add later to 7.1.0 (they don't appear to be breaking changes). What do you think?

hansemannn:
Thanks for the review @jquick-axway, really appreciate it! If you want, we can split it up, although I am unsure how to retain the cherry-picked commits for that case.

kopiro:
Maybe we have to squash commits :)

jquick-axway:
@hansemannn, @kopiro, how about we hold on to this PR for 7.1.0. We can make a separate/simpler PR with the deprecations for 7.0.0 and cite the source in code comments.

Oh and I'm only recommending that we split the PR because we're supposed to release 7.0.0 by mid-November and that release date is non-negotiable. We have way to much stuff trying to get added into 7.0.0 at the moment and we want to minimize the risk of bugs/issues.

mukherjee2:
@hansemannn is there any resolution on this? Currently it's not passing CR. Is this being moved to 7.1.0?

jquick-axway:
@mukherjee2, this PR is being pushed out to 7.10.

We created new PR #9589 for 7.0.0 which contains this PR's deprecation warnings.

hansemannn:
@ypbnv I just rebased the PR with master and had to solve a few critical merge conflicts that overlapped with #9558. Would you mind checking if the Android side is still valid? The major differences were in the content resolvers and properties that are set to the native event (using the cursor). I will continue finishing this PR afterwards. Thank you!

PR #9563, opened 2017-10-27
[TIMOB-24000] Module build should error if the apiversion in manifest…

feons:
… does not match the SDKs apiversion

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24000


cb1kenobi:
I thought we discussed that the apiversion should be dictated by the module build for the selected SDK version. No?

feons:
then {{minsdk}} should also be updated, but do we have a record of the exact min sdk version for a given apiversion?

cb1kenobi:
I'm not sure. If we don't, then it might be a good idea to start a wiki page (or section of existing page) documenting the min sdk info.

sgtcoolguy:
We do not have an exact record of the sdk versions that bumped the module api versions, but we should.

The apiversion mismatch is a good time to alert the user to update both that and the minSDK. We could just ignore the user-supplied apiversion and stick the current one into the built manifest, but it means the minSDK has to be updated as well.

We could stick a JSON file into the SDK to map from apiversion to sdk version and insert both (or tell the user what they should be)...

android/cli/commands/_buildModule.js
```diff
@@ -58,6 +58,14 @@ AndroidModuleBuilder.prototype.validate = function validate(logger, config, cli)
 
 		this.manifest = this.cli.manifest;
 
+		const sdkModuleAPIVersion = this.cli.sdk && this.cli.sdk.manifest && this.cli.sdk.manifest.moduleAPIVersion && this.cli.sdk.manifest.moduleAPIVersion['android'];
+		if (this.manifest.apiversion && sdkModuleAPIVersion && this.manifest.apiversion !== sdkModuleAPIVersion) {
+			logger.error(__('The module manifest apiversion is currently set to %s', this.manifest.apiversion));
+			logger.error(__('Titanium SDK %s Android module apiversion is at %s', this.titaniumSdkVersion, sdkModuleAPIVersion));
+			logger.error(__('Please update module manifest apiversion to match Titanium SDK module apiversion.'));
```
sgtcoolguy: It'd be good to also alert them that they likely need to update the minSDK value too.

iphone/cli/commands/_buildModule.js
```diff
@@ -52,6 +52,14 @@ iOSModuleBuilder.prototype.validate = function validate(logger, config, cli) {
 	this.buildOnly     = cli.argv['build-only'];
 	this.xcodeEnv      = null;
 
+	const sdkModuleAPIVersion = cli.sdk.manifest && cli.sdk.manifest.moduleAPIVersion && cli.sdk.manifest.moduleAPIVersion['iphone'];
+	if (this.manifest.apiversion && sdkModuleAPIVersion && this.manifest.apiversion !== sdkModuleAPIVersion) {
+		logger.error(__('The module manifest apiversion is currently set to %s', this.manifest.apiversion));
+		logger.error(__('Titanium SDK %s iOS module apiversion is at %s', this.titaniumSdkVersion, sdkModuleAPIVersion));
+		logger.error(__('Please update module manifest apiversion to match Titanium SDK module apiversion.'));
```
sgtcoolguy: alert about minSDK too.

sgtcoolguy (changes requested):
So at a bare minimum we should alert about minSDK too. Ideally we'd be able to tell them the minSDK version to use by adding some JSON file into our SDK (or maybe add to the package.json files) the mapping of apiversion -> minSDK.

If we had that data, we could potentially just inject the apiversion and minSDK for them - though I would assume we should at least alert them that we overrode the values or something.

mukherjee2 (approved):
Node Version: 6.11.5
NPM Version: 3.10.10
Mac OS: 10.13
Windows OS 10.0.14393
Appc CLI: 7.0.0-master.6
Appc CLI NPM: 4.2.11-2
Ti SDK version 7.0.0

PR #9571, opened 2017-10-30
[TIMOB-25459] Android: toImage method needs parity with iOS

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25459

**Description:**
Result from calling toImage is returned directly to the callback function. Should be in a dictionary with "blob" key to match iOS behavior.

**Test case:**
app.js
```
var view = Ti.UI.createView({backgroundColor: 'green', width: 150, height: 150, top: 100});
var content = Ti.UI.createView({backgroundColor: 'blue', width: 45, height: 45});
var imageView = Ti.UI.createImageView({width: 150, height: 150, top: 300});
view.add(content);
win.add(view);
win.addEventListener('postlayout', function() {
	view.toImage(function(e) {
		imageView.setImage(e.blob);
	});
})
win.add(imageView);
win.open();
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -991,7 +991,9 @@ public abstract class TiViewProxy extends KrollProxy implements Handler.Callback
 			blob  = TiBlob.blobFromImage(Bitmap.createBitmap(1, 1, Config.ARGB_8888));
 			Runnable renderRunnable = new Runnable() {
 				public void run() {
-					callback.callAsync(getKrollObject(), new Object[] {handleToImage()});
+					KrollDict callbackResult = new KrollDict();
+					callbackResult.put("blob", handleToImage());
+					callback.callAsync(getKrollObject(), callbackResult);
 				}
 			};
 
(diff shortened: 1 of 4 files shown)


jquick-axway (commented):
@ypbnv, @hansemannn,

We currently document that the callback argument is `Ti.Blob` here...
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.View-method-toImage

So, Android was matching the documentation. iOS was not. And Windows does not appear to support this API (have not tried it)...
https://github.com/appcelerator/titanium_mobile_windows/blob/649177875b31712d5bceacd489d4595111648b3f/Source/TitaniumKit/src/UI/ViewLayoutDelegate.cpp#L73

So, which way should we go?
We can either modify iOS to match the docs or take Yordan's change, but it'll require a doc change too.

garymathews:
Windows does support this API, it's implemented [here](https://github.com/appcelerator/titanium_mobile_windows/blob/649177875b31712d5bceacd489d4595111648b3f/Source/UI/src/WindowsViewLayoutDelegate.cpp#L242) and returns a Ti.Blob as documented.

ypbnv:
@jquick-axway, @garymathews, @hansemannn I did not find another callback returning result without dictionary in the docs. so I will update them as well. If we decide to keep it as it is now I will revert the change.

jquick-axway:
@hansemannn, agreed. Going forward, all callbacks we create in the future should provide a dictionary. That way, we can easily extend them with new properties/fields if needed.

hansemannn:
What I am wondering is, why this never popped up before. I'll do the iOS PR later this week, closing this one.

PR #9573, opened 2017-10-31
[TIMOB-25269] (6_3_X) iOS: Use final iPhone X launch-image names

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25269


ewieberappc (approved):
Approved. Size ordering is the same as other iOS versions and the correct landscape launch image appears in the build for iPhone X sizes

PR #9576, opened 2017-10-31
[TIMOB-25462] Android: Implement Ti.UI.WebView.userAgent

garymathews:
- Implement `userAgent` property for parity with iOS

###### TEST CASE
```JS
        url: 'http://whatsmyuseragent.org',
        userAgent: 'TEST AGENT'
    });

webView.addEventListener('load', function(e) {
    var exp = /user-agent.+\s+.+\>(.*)\</g.exec(e.source.html),
        userAgent = exp && exp.length > 1 ? exp[1] : undefined;

    alert(userAgent);
});

win.add(webView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25462)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -460,6 +460,10 @@ public void processProperties(KrollDict d)
 		if (d.containsKey(TiC.PROPERTY_DISABLE_CONTEXT_MENU)) {
 			disableContextMenu = TiConvert.toBoolean(d, TiC.PROPERTY_DISABLE_CONTEXT_MENU);
 		}
+
+		if (d.containsKey(TiC.PROPERTY_USER_AGENT)) {
```
ypbnv: `processProperties` is called after the creation of the native WebView and this causes a behavior that is a bit confusing - the WebView fires 'load' event twice. This seems to be an issue in chromium itself:
https://bugs.chromium.org/p/chromium/issues/detail?id=315891
Changing the UserAgent while loading a page provokes a reload. I could not find more recent discussion about that.
The result is that the test may fail because of that. The first 'load' event is delivered with an empty html and the UserAgent will never match the value from the creation dictionary.

Setting a `userAgent` value from the creation dictionary when preparing the WebView settings before initializing it seems to avoid this:
https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java#L301

Something like that maybe:
```
if (proxy.hasPropertyAndNotNull(TiC.PROPERTY_USER_AGENT)) {
    settings.setUserAgentString(TiConvert.toString(proxy.getProperty(TiC.PROPERTY_USER_AGENT)));
}
```

Still that leaves us with that undesired result when changing the UserAgent after the WebView has been created. I am not sure what is the behavior on iOS side:
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.WebView-property-userAgent
But does it make sense to set the property as Creation Only ?

ypbnv (changes requested):

tests/Resources/ti.ui.webview.test.js
```diff
@@ -0,0 +1,345 @@
...
+		});
+
+		w.add(webview);
+		w.open();
+	});
+
+	it('userAgent', function (finish) {
+		win = Ti.UI.createWindow({backgroundColor: 'gray'});
+		var webView = Ti.UI.createWebView({
+		        url: 'http://whatsmyuseragent.org',
```
ypbnv: Indentation is with spaces for some of the next lines. Using tabs would match the rest of the file.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
garymathews: @ypbnv nice catch! I'll update the PR to use the creation dictionary 👍

lokeshchdhry:
@garymathews , when I run the test, I see 2 alerts one with no value before the webview loads & the other with the value `TEST AGENT`.

garymathews:
@lokeshchdhry Updated PR

lokeshchdhry:
FR Passed.

User Agent is set & detected successfully.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

PR #9577, opened 2017-10-31
[6_3_X][TIMOB-25405] Remove const/let usage

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25405

Maintaining Node v4 compatibility in the 6_3_X branch

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -45,7 +45,7 @@ var appc = require('node-appc'),
 	series = appc.async.series,
 	version = appc.version;
 var platformsRegExp = /^(android|ios|iphone|ipad|mobileweb|blackberry|windows|tizen)$/;
-const pemCertRegExp = /(^-----BEGIN CERTIFICATE-----)|(-----END CERTIFICATE-----.*$)|\n/g;
+var pemCertRegExp = /(^-----BEGIN CERTIFICATE-----)|(-----END CERTIFICATE-----.*$)|\n/g;
 
 function iOSBuilder() {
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -45,7 +45,7 @@ var appc = require('node-appc'),
 	series = appc.async.series,
 	version = appc.version;
 var platformsRegExp = /^(android|ios|iphone|ipad|mobileweb|blackberry|windows|tizen)$/;
-const pemCertRegExp = /(^-----BEGIN CERTIFICATE-----)|(-----END CERTIFICATE-----.*$)|\n/g;
+var pemCertRegExp = /(^-----BEGIN CERTIFICATE-----)|(-----END CERTIFICATE-----.*$)|\n/g;
```
ewanharris: The usage here is fine and I had left it in my previous PR, but decided to remove this time to keep the file cleaner.

hansemannn (approved):

mukherjee2 (approved):
Node Version: 4.8.2
NPM Version: 2.15.11
Mac OS: 10.13
Appc CLI: 6.3.0-master.16
Appc CLI NPM: 4.2.10-2
Titanium SDK version: 6.3.0.v20171031154017 (from this PR)
Hyperloop 2.2.2
Android 7.1.2 Pixel
Xcode 9.1/ iOS 11.03 iPhone 7 Plus

Passed FR. I first reproduced the error by using the affectedVersion, and then switch to the SDK with the fix and found that the error was no longer thrown, and the app worked as expected.

PR #9578, opened 2017-11-01
[TIMOB-25059] Add Hyperloop as a pre-packged module

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25059

**Optional Description:**
First step of integrating Hyperloop into our SDK by bundling it as a pre-packaged module.


build/packager.js
```diff
@@ -175,6 +184,11 @@ Packager.prototype.includePackagedModules = function (next) {
 		|| supportedPlatforms.indexOf('ipad') !== -1) {
 		supportedPlatforms = supportedPlatforms.concat([ 'ios', 'iphone', 'ipad' ]);
 	}
+
+	// Hyperloop has no single platform downloads yet, so we use a fake platform
+	// that will download the all-in-one distribution.
+	supportedPlatforms = supportedPlatforms.concat([ 'hyperloop' ]);
```
sgtcoolguy: so you can just hijack the "commonjs" entry to mean "All platforms" too. That's what I did locally when I took a look at this.

sgtcoolguy:
LGTM 👍

support/module/packaged/modules.json
```diff
@@ -15,5 +15,8 @@
 	],
 	"commonjs": [
 		"https://github.com/appcelerator-modules/ti.cloud/releases/download/3.2.11/ti.cloud-commonjs-3.2.11.zip"
+	],
+	"hyperloop": [
+		"https://appcelerator-modules.s3.amazonaws.com/hyperloop-2.2.2.zip"
```
hansemannn: Please change to 3.0.0. Eventually, we should create a `hyperloop-releases` repo to store those, which would also have the benefit to attach changelogs and beta-versions.

hansemannn (changes requested):

support/module/packaged/modules.json
janvennemann: We probably should do that right away. I can't update this link right now because that would break builds since ee don't have a 3.0.0 published yet. That would also trigger the automatic update in SDKs <= 6.3.0 so we shouldn't even publish Hyperloop 3.0.0 the old way?.

hansemannn: Maybe we should. Then people with an old SDK can still download it, but that on the other hand still breaks their app because of the bumped api-version that requires 7.0.0 anyway.

hansemannn (approved):

hansemannn:
PR **approved**. We should merge this one and do another PR once Hyperloop 3.0.0 GA is ready.

PR #9585, opened 2017-11-04
[TIMOB-24537] Android: Refactored orientation handling.

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-25483
- https://jira.appcelerator.org/browse/TIMOB-25467
- https://jira.appcelerator.org/browse/TIMOB-24537
- https://jira.appcelerator.org/browse/TIMOB-19878
- https://jira.appcelerator.org/browse/TIMOB-10070

**Summary:**
- [TIMOB-25483](https://jira.appcelerator.org/browse/TIMOB-25483) Added FACE_UP/FACE_DOWN support to Ti.Gesture.
- [TIMOB-25467](https://jira.appcelerator.org/browse/TIMOB-25467) Modified "Ti.UI.Window.orientation" to return screen orientation in split-screen mode, like iOS.
- [TIMOB-24537](https://jira.appcelerator.org/browse/TIMOB-24537) Modified "Ti.Gesture" to provide device orientation instead of app orientation, like iOS.
- [TIMOB-19878](https://jira.appcelerator.org/browse/TIMOB-19878) Fixed bug where "Ti.Gesture" orientation event is not always fired after a rotation. (Always happens in Android emulator.)
- [TIMOB-10070](https://jira.appcelerator.org/browse/TIMOB-10070) Fixed bug where landscape left/right handling was backwards on phones. (Was okay on large tablets.)

---

**Ti.Gesture Device Orientation Test:**

_(This tests Ti.Gesture breaking change from app window orientation to device orientation.)_

1. Build and run the test code attached to [TIMOB-25483](https://jira.appcelerator.org/browse/TIMOB-25483) on an Android phone. (Cannot be the emulator or a tablet.)
2. Lay the phone flat on a table with the screen facing up towards the sky.
3. Verify that the app displays "Face Up" onscreen.
4. Lift the device above your head, with the screen facing straight down towards the Earth.
5. Verify that the app displays "Face Down" onscreen.
6. Move the device to a portrait orientation.
7. Verify that the app displays "Portrait Upright" onscreen.
8. Rotate the device to landscape form, where the bottom navigation bar is on the **right** side of the screen.
9. Verify that the app displays "Landscape Right" onscreen.
10. Rotate the device to landscape form, where the bottom navigation bar is on the **left** side of the screen.
11. Verify that the app displays "Landscape Left" onscreen.
12. Repeat steps 2-11 on a large tablet, such as a Galaxy Tab or Nexus 10. _(Note that "Landscape Right" will be the tablet's upright position and "Portrait Upright" must be 90 degrees clockwise to "Landscape Right" like how it works on a phone.)_
13. Repeat steps 7-11 in the Android emulator, phone or tablet. _(This tests lack of gyroscope sensor and [TIMOB-19878](https://jira.appcelerator.org/browse/TIMOB-19878) bug fix.)_

---

**Landscape Left/Right Test:**

1. Build and run the "OrientationModeTest.js" attached to [TIMOB-10070](https://jira.appcelerator.org/browse/TIMOB-10070) on an Android phone.
2. On app startup, tap "Landscape Right" in the displayed alert dialog.
3. Verify that the navigation bar is on the **right** side of the landscape window.
4. Restart the app.
5. On app startup, tap "Landscape Left" in the displayed alert dialog.
6. Verify that the navigation bar is on the **left** side of the landscape window.
7. Run the app on a large Android tablet such as a Galaxy Tab or Nexus 10.
8. On app startup, tap "Landscape Right" in the displayed alert dialog.
9. Verify that the landscape window is displayed in the tablet's upright position. (The upright orientation of a large tablet is typically landscape.)
10. Restart the app.
11. On app startup, tap "Landscape Left" in the displayed alert dialog.
12. Verify that the landscape window is displayed upside dow on the tablet.

---

**Split-Screen Orientation Test:**

Run the test code/procedure attached to [TIMOB-25467](https://jira.appcelerator.org/browse/TIMOB-25467).


android/modules/gesture/src/java/ti/modules/titanium/gesture/TiDeviceOrientationMonitor.java
```diff
@@ -0,0 +1,631 @@
...
+		if (isRunning()) {
+			return true;
+		}
+
+		// First, attempt to use the accelerometer and gyroscope sensors.
+		this.isUsingSensorManager = startSensor(Sensor.TYPE_ROTATION_VECTOR);
+
+		// If above is not available, then attempt to use accelerometer and magnetic sensors.
+		if (!this.isUsingSensorManager) {
+			boolean wasSensor1Started = startSensor(Sensor.TYPE_ACCELEROMETER);
```
garymathews: Maybe name these better, `accelerometerStarted` `magnetometerStarted` ?

android/modules/gesture/src/java/ti/modules/titanium/gesture/TiDeviceOrientationMonitor.java
```diff
@@ -0,0 +1,631 @@
...
+		try {
+			if (this.sensorManager != null) {
+				Sensor sensor = this.sensorManager.getDefaultSensor(typeId);
+				if (sensor != null) {
+					wasSuccessful = this.sensorManager.registerListener(
+							this.sensorEventHandler, sensor,
+							SensorManager.SENSOR_DELAY_NORMAL, this.handler);
+				}
+			}
+		} catch (Exception ex) {
```
garymathews: Should we still log any exceptions here as warnings?

android/modules/gesture/src/java/ti/modules/titanium/gesture/TiDeviceOrientationMonitor.java
```diff
@@ -0,0 +1,631 @@
...
+		boolean isLockEnabled = false;
+		if (this.isSystemRotationLockIgnored == false) {
+			try {
+				ContentResolver contentResolver = this.context.getContentResolver();
+				if (contentResolver != null) {
+					int value = Settings.System.getInt(
+							contentResolver, Settings.System.ACCELEROMETER_ROTATION, 1);
+					isLockEnabled = (value == 0);
+				}
+			} catch (Exception ex) {
```
garymathews: Same here

android/modules/gesture/src/java/ti/modules/titanium/gesture/TiDeviceOrientationMonitor.java
```diff
@@ -0,0 +1,631 @@
...
+			}
+
+			// Flag this handler as disabled.
+			this.isEnabled = false;
+		}
+
+		private void updateLastReadOrientation()
+		{
+			// Do not continue if the screen handler has been disabled.
+			if (this.isEnabled == false) {
```
garymathews: Minor, maybe use `!isEnabled()` like previous usages?

garymathews (changes requested):

lokeshchdhry:
FR Passed.

1. https://jira.appcelerator.org/browse/TIMOB-24537 -- `orientatonChange` is fired successfully after every change in orientation.
2. https://jira.appcelerator.org/browse/TIMOB-25483 -- `Face Up` & `Face Down` is fired as expected.
3. https://jira.appcelerator.org/browse/TIMOB-25467 -- Correct window orientation is returned in split screen mode.
4. https://jira.appcelerator.org/browse/TIMOB-19878 -- Ti.Gesture `orientation change` is fired successfully every time.
5. https://jira.appcelerator.org/browse/TIMOB-10070 -- Landscape `left` & `right` is in correct direction.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.10
Appc CLI:       6.3.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.7
Node Ver:       7.10.1
Java Ver:       1.8.0_101

android/modules/gesture/src/java/ti/modules/titanium/gesture/TiDeviceOrientationMonitor.java
jquick-axway: I'm already logging a warning in GestureModule here...
https://github.com/appcelerator/titanium_mobile/pull/9585/files#diff-9f99efc31c1562fe3946b86679b014b1R76

The idea is that it'll only log a warning once if it's unable to start the device orientation monitor on startup. If I logged a warning within the DeviceOrientationMonitor class, then we would see a warning logged every time an activty is paused/resumed and I wanted to reduce the noise.

jquick-axway:
Updated PR based on feedback.

android/titanium/src/java/org/appcelerator/titanium/util/TiDeviceOrientation.java
```diff
@@ -0,0 +1,500 @@
...
+/**
+ * Enum type indicating if a device is portrait-upright, landscape-left, landscape-right, etc.
+ * <p>
+ * Provides methods to acquire its unique Titanium integer ID used in JavaScript
+ * and to determine if the device is portrait or landscape.
+ */
+public enum TiDeviceOrientation
+{
+	/** Indicates that the device orientation is unknown. */
+	UNKNOWN(TiIntegerId.UNKNOWN, "Ti.UI.UNKNOWN"),
```
garymathews: Can we simplify it a bit by using the enum as the sole definition and reducing `TiIntegerId` to just `id` (also make it `final`)

```Java
public enum TiDeviceOrientation
{
  UNKNOWN(0),
  PORTRAIT_UPRIGHT(1),
  PORTRAIT_UPSIDE_DOWN(2),
  LANDSCAPE_LEFT(3),
  LANDSCAPE_RIGHT(4),
  FACE_UP(5),
  FACE_DOWN(6);
  
  private final int id;
  
  private TiDeviceOrientation(int id)
  {
    this.id = id;
  }
  
  public int toInt()
  {
    return this.id;
  }

  ...
}
```

and then in our `UIModule.java` we can do this

```Java
@Kroll.constant public static final int PORTRAIT = TiDeviceOrientation.PORTRAIT_UPRIGHT.toInt();
@Kroll.constant public static final int UPSIDE_PORTRAIT = TiDeviceOrientation.PORTRAIT_UPSIDE_DOWN.toInt();
@Kroll.constant public static final int LANDSCAPE_LEFT = TiDeviceOrientation.LANDSCAPE_LEFT.toInt();
@Kroll.constant public static final int LANDSCAPE_RIGHT = TiDeviceOrientation.LANDSCAPE_RIGHT.toInt();
@Kroll.constant public static final int FACE_UP = TiDeviceOrientation.FACE_UP.toInt();
@Kroll.constant public static final int FACE_DOWN = TiDeviceOrientation.FACE_DOWN.toInt();
@Kroll.constant public static final int UNKNOWN = TiDeviceOrientation.UNKNOWN.toInt();
```

I'm not sure how I feel about storing the namespace as a string with the constant, since it means another thing to update. But if we do want to keep it we could do this instead:

```Java
private final String namespace = "Ti.UI." + this.name();
```

garymathews (changes requested):
see above

also, looks like some related test cases are failing

android/titanium/src/java/org/appcelerator/titanium/util/TiDeviceOrientation.java
jquick-axway: Yes, we can set up our Kroll constant like this...
```java
@Kroll.constant public static final int PORTRAIT = TiDeviceOrientation.PORTRAIT_UPRIGHT.toTiIntegerId();
```

But the issue with this is that the Kroll constant can't be used in a switch statement. This is because the constant is assigned via a static initializer at runtime. The way I'm doing it, the constant is inlined at compile time and _**can**_ be used in a switch statement.

jquick-axway: We should not change `TiDeviceOrientation.toTiIntegerId()` to `toInt()`. The reason is to make the class future proof in case we want to provide other `toX()` style method mappings, such as to a Google Surface class rotation constant or an ActivityInfo screen orientation constant. I'm already doing this with the static `fromX()` methods I have now.

Also, I find that it's a good pattern to pair a `toX()` method and the static `fromX()` method with the same identifier name so that it's obvious to the developer that the 2 can be used in conjunction to serialize/deserialize back to the same object type.

```java
int id = TiDeviceOrientation.PORTRAIT_UPRIGHT.toTiIntegerId();
if (TiDeviceOrientation.fromTiIntegerId(id) == TiDeviceOrientation.PORTRAIT_UPRIGHT) {
	boolean isMatching = true;
}
```

lokeshchdhry:
@jquick-axway , can you please resolve the conflicts.

android/titanium/src/java/org/appcelerator/titanium/util/TiDeviceOrientation.java
garymathews: I just feel `toTiIntegerId()` is rather long. We could call it `toId()`/`fromId()` instead?

PR #9592, opened 2017-11-13
[TIMOB-25504] iOS: Fix default-image selection between iPhone 6/7 Plus and iPhone X

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25504


ewieberappc:
FR Passed. The correct images are now selected on iPhone X when both 736h and 2436h images are present.

PR #9239, opened 2017-07-21
[TIMOB-25005] Remove Ti.UI.iPhone namespace

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25005

PR #9602, opened 2017-11-15
[TIMOB-25005] Remove bad-boy

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25005

Finding the lost child from https://github.com/appcelerator/titanium_mobile/pull/9239


ewieberappc (approved):
FR Passed. File is removed and can still build/run projects using button style constants (which are now Ti.UI.iOS).

PR #9605, opened 2017-11-15
[TIMOB-25518][TIMOB-25506] Improve validation

garymathews:
- [TIMOB-25518](https://jira.appcelerator.org/browse/TIMOB-25518): Prevent race condition of `client.disconnect()` upon garbage collection
- [TIMOB-25506](https://jira.appcelerator.org/browse/TIMOB-25506): Validate `TiDownloadListener` before iterating


jquick-axway (approved):
CR: Pass

I see one other edge case that would cause a crash, but we can do resolve that in a future PR (or now if you want).

An invoked listener can add/remove a listener to the main collection, which would end up breaking the listener list iterator and cause an exception. The `synchronized` block won't guard against this since such an issue can happen on the same thread. How this is typically done is you clone the listener collection and then foreach the cloned collection instead. You then don't have to wrap the foreach loop in a synchronized block, which is preferred to avoid potential deadlock (in case listener performs a long blocking operation). We should also check if each cloned listener has been removed from the main collection before invoking it too.

garymathews:
@jquick-axway nice catch! But cloning arrays is expensive, this scenario is perfect for an `Iterator` which handles concurrent modifications of the array during iteration.

Updated PR

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -1289,11 +1289,13 @@ public void progress(int progress) {
 				}
 
 			} catch(Throwable t) {
-				if (client != null) {
-					Log.d(TAG, "clearing the expired and idle connections", Log.DEBUG_MODE);
-					client.disconnect();
-				} else {
-					Log.d(TAG, "client is not valid, unable to clear expired and idle connections");
+				synchronized (client) {
```
sgtcoolguy: if client *is* null, doesn't the synchronized block break?

jquick-axway: Chis is right. From the looks of it, calling abort() during a long download should cause a NullPointerException.

Also, we're not synchronizing against the "client" object anywhere else in the code, making this synchronization block useless.

The race condition that can happen here is with the `abort()` method since it too is accessing and assigning the `client` member variable. This is the only spot in the code where the `client` member variable is accessed on a different thread. While we could wrap a synchronization block around all `client` assignments, in this case, it might be simpler to ditch the synchronization block and do the following in the `abort()` method instead...

```java
public void abort()
{
	if (readyState > READY_STATE_UNSENT && readyState < READY_STATE_DONE) {
		aborted = true;
//<MyNewCode>
		try {
			HttpURLConnection client = this.client;
			if (client != null) {
				client.disconnect();
			}
		} catch (Exception ex) {
		}
//</MyNewCode>
		// Fire the disposehandle event if the request is aborted.
		// And it will dispose the handle of the httpclient in the JS.
		proxy.fireEvent(TiC.EVENT_DISPOSE_HANDLE, null);
	}
}
```

The idea is that we don't want to set `client` to null in the `abort()` method. We should let the Runnable thread do it in its `finally` block (it's already has `abort` member variable checks). Also, Java reference assignment is atomic and copying the `client` member variable to a local like the above avoids a race condition where the `client` member variable reference might change on the other thread, such as to null or to another client connection when redirecting.

garymathews: Good catch! I'll update the PR

garymathews:
@jquick-axway Updated PR

fahad86:
everything ok here? Want this fix to go out soon as recently seeing this issue occur more frequently.

lokeshchdhry:
FR Passed.

Since this is just a validation change the `http` unit tests are sufficient.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.11
Appc CLI:       7.0.1
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101
Devices:        ⇨ google Pixel XL --- Android 8.0.0

PR #9608, opened 2017-11-16
[TIMOB-25522] iOS: Add "navigationWindow" property to Ti.UI.Window, add unit-tests for all nav-window API's

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25522


tests/Resources/ti.ui.window.test.js
```diff
@@ -0,0 +1,73 @@
...
+	});
+
+	afterEach(function () {
+		if (win) {
+			win.close();
+		}
+		win = null;
+	});
+
+	it.ios('#navigationWindow', function () {
```
sgtcoolguy: I've been trying to use a naming scheme so that `#whatever` is used to denote methods, `.whatever` is used to denote properties.

tests/Resources/ti.ui.window.test.js
```diff
@@ -0,0 +1,73 @@
...
+			should(win.navigationWindow).eql(nav);
+
+			should(nav.openWindow).be.a.function;
+			should(win.navigationWindow.openWindow).be.a.function;
+			nav.close();
+			win = null;
+			finish();
+		});
+		
+		var nav = Ti.UI.iOS.createNavigationWindow({
```
sgtcoolguy: var declaration should always be at the top of the scope (or else eslint will complain, since vars are hoisted anyways)
So `var nav;` should go at the very top, then the assignment stays here.

tests/Resources/ti.ui.window.test.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+
+var should = require('./utilities/assertions'),
+	utilities = require('./utilities/utilities');
+
+describe('Titanium.UI.Window', function () {
```
sgtcoolguy: rename this file to `ti.ui.window.navigationwindow.addontest.js` so it gets "added on" the existing suite rather than replaces the existing ti.ui.window tests.

sgtcoolguy (changes requested):

tests/Resources/ti.ui.window.test.js
sgtcoolguy: Alternatively, you could add a `ti.ui.ios.navigationwindow.addontest.js` for this create/property test and any future tests on that type (I see we have 3 methods that likely should have tests: http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.iOS.NavigationWindow)

hansemannn: How to deal with `before`, `after`, `beforeEach` and `afterEach` methods? Should I include those in the addons?

sgtcoolguy: Looks like to be safe on Android the before/after should stay. You don't check for, so don't need `didFocus` or `before` hook. I'd keep the `after` hook, and drop your explicit setting of `win = null` in the test (is that needed?)

hansemannn: Okay, will do. Copied over the `win = null` statement, don't think it's needed as the proxy should clean up after itself.

tests/Resources/ti.ui.window.test.js
hansemannn: Gotcha, will change!

tests/Resources/ti.ui.window.test.js
hansemannn: Decided to use `ti.ui.ios.navigationwindow.addontest.js` and will add unit-tests for the other ones as well to get some more basic coverage. And as Android / Windows will get parity on the navigation-window, they'll be happy to find existing unit-tests already.

tests/Resources/ti.ui.ios.navigationwindow.addontest.js
```diff
@@ -41,6 +41,72 @@ describe('Titanium.UI.Window', function () {
 		}
 		win = null;
 	});
+	
+	it.ios('#openWindow', function () {
```
sgtcoolguy: FYI, you can just do `describe.ios` at the top to limit all sub-tests to iOS and avoid having to do `it.ios` on them all.

hansemannn: Doing now, thx!

tests/Resources/ti.ui.ios.navigationwindow.addontest.js
```diff
@@ -0,0 +1,132 @@
...
+		subWindow.addEventListener('open', function () {
+			should(nav.closeWindow).be.a.function;
+			nav.closeWindow(subWindow);
+		});
+		
+		subWindow.addEventListener('close', function () {
+			finish();
+		});
+		
+		nav.open();
```
sgtcoolguy: This is causing a crash/error on each test. I assume you meant to do `win.open();`?

hansemannn: Mhh, nope, we need to open the navigation window. I am wondering if the following won't cause issues already, because we provide a null-object and set a reference later:
```js
var nav = Ti.UI.iOS.createNavigationWindow({
  window: win // will be null at this point
});

win = Ti.UI.createWindow(); // now it's getting it's reference?
```
Or we declare the `var nav;` and assign it later?

sgtcoolguy (approved):

ssjsamir (approved):
FR Passed: Able to reference the current navigation-window using the "navigationWindow" instance property. From the following test case from https://jira.appcelerator.org/browse/TIMOB-25522: 

```js
  backgroundColor: '#fff'
});
 
var nav = Ti.UI.iOS.createNavigationWindow({
  window: win
});
 
  title: 'Open next window'
});
 
  win.navigationWindow.openWindow(Ti.UI.createWindow({
    title: 'Next Window!',
    backgroundColor: 'green'
  }));
});
 
win.add(btn);
nav.open();
```
**Test steps**
+ Created a build from this PR
+ Created a titanium app
+ Added the test case above
+ Pressed 'open newt window'
+ New window was opened 

**Test Environment**
APPC Studio: 5.0.0.201712081732
APPC CLI: 7.0.3
iphone 6 plus (10.2)
Node.js Version: 8.9.1
Xcode 9.2

hansemannn:
@sgtcoolguy I cannot figure out why the unit tests here are still failing.

PR #9618, opened 2017-11-18
[TIMOB-25195] iOS: Refactor module template and build-settings

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25195

- [x] Creates new iOS modules in `ios/` instead of `iphone/`
- [x] Uses recommended build-settings like post-processing and assert-removal
- [x] Refactors the default readme attached to a module
- [x] Refactors and lints the default templates to match our clang-format
- [x] Prevents the README.md from the `platform/` directory from being zipped
- [x] Adds a .gitignore for default ignored files and directories
- [x] Updates the default license-headers to match Axway


janvennemann (commented):
Changes are looking good. I think you linked this under the wrong ticket number though, should probably be TIMOB-25195? Do you want to edit your commits or should i just squash and merge under the correct ticket?

hansemannn:
It was triggered by the above, but actually TIMOB-25195 is what I thought before, yeah 😄.

hansemannn:
@sgtcoolguy Any objections from the tooling side? I changed the module's `iphone/` directory to `ios/`, I may need to check if that works out for all builds.

sgtcoolguy (approved):
LGTM 👍

janvennemann (approved):
LGTM!

This ticket is marked for 7.0.0 but since this does not seem to be an important bug fix, do we still take that in?

hansemannn:
Technically, it's a major change, although not necessarily a breaking. I would take it if QE can test it (steps in JIRA), but otherwise we go for 7.1.0. @ewieberappc thoughts?

ewieberappc:
FR Passed. Able to create, package, and run projects with the new module template

PR #9620, opened 2017-11-20
[TIMOB-25540] Android: Fix sizeOf() calculation to prevent IllegalStateException

garymathews:
- `onTrimMemory()` can cause an exception when attempting to release memory
- Fix `sizeOf()` calculation to prevent `IllegalStateException`

###### TEST CASE
- Download an image first
- Run app and select image from Downloads folder
- Press back to exit app, app should not crash
```JS
Ti.Media.openPhotoGallery({
  success:function(event) {
    var image = event.media.imageAsResized(300, 300);
    console.log('resized image');
  }
});
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25540)


jquick-axway (changes requested):
There is a problem here.

The value passed to the `LruCache()` constructor is in kilobytes and the value returned by the `sizeOf()` method is in bytes. The two need to use the same units of measure or else the `LruCache` won't function correctly. In this case, it'll remove bitmaps from the cache too soon.

From looking at Google's `Bitmap` class docs, this method should be written like the below. I also noticed that we were not rounding-up to the nearest kilobyte, which might have been the source of the error?

```java
int byteCount;
if (android.os.Build.VERSION.SDK_INT >= 19) {
    byteCount = bitmap.getAllocationByteCount();
} else {
    byteCount = bitmap.getByteCount();
}
return (int)Math.ceil((double)byteCount / 1024.0);
```

Alternatively, switching from kilobytes to bytes for the max size and sizeOf() might be simpler.

jquick-axway:
**Side Note:**
I actually want to get rid of our LruCache class usage. It's not Titanium's job to assume which images need to be cached. It's the developer's job to do so, which they can do now on all platform by caching loaded images to blobs.

garymathews:
@jquick-axway The crash still occurs with your requested changes, not sure why?

garymathews:
I see now, both `getByteCount()` and `getAllocationByteCount()` will return `0` on recycled bitmaps. We need to manually calculate the minimum size:
```Java
int byteCount = bitmap.getRowBytes() * bitmap.getHeight();
return byteCount / 1024;
```

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
FR Passed application no longer crashed when going back and the log showed the following:
`resized image`

**Test steps**
+ Added the test case from the description above 
+ Built the app using the artefacts from this PR 
+ Ran the program
+ Clicked on a downloaded image
+ Console log message was shown as application went to appc logo
+ Pressed back 
+ Application did not crash and was able to view downloaded images 

**Test Environment**
APPC CLI: 7.0.0
Device: Nexus 6p (8.1.0)
Node.js Version: 8.9.1

PR #9628, opened 2017-11-27
[TIMOB-20332] Android: Closing a Ti.UI.TabGroup that contains a ti.map view crashes the app

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20332

**Description:**
Getting a reference to the wrong FragmentManager when an instance of TIUIFragment is used in TabGroup.
Removed an unused import.

**Test case:**
Add the module in tiapp.xml
`<module platform="android">ti.map</module>`

app.js
```
// this sets the background color of the master UIView (when there are no windows/tab groups on it)
Titanium.UI.setBackgroundColor('#000');

// create tab group
var tabGroup = Titanium.UI.createTabGroup();

var win1 = Titanium.UI.createWindow({
    title:'Tab 1',
    backgroundColor:'#fff'
});

var tab1 = Titanium.UI.createTab({
    icon:'KS_nav_views.png',
    title:'Tab 1',
    window:win1
});

var Map = require('ti.map');

var mapview = Map.createView({top: 0, height: '80%'});

win1.add(mapview);

tabGroup.addTab(tab1);

tabGroup.open();

var button = Ti.UI.createButton({bottom: 10});

button.addEventListener('click', function() {
    tabGroup.close();
})

win1.add(button);
```

Expected result - application should close without any crashes.


sgtcoolguy:
Can we please turn that manual test into an automated mocha test? I assume the only difference would be to instead of using a click listener to close the tab group, use something like setTimeout to close the tab after a short delay. The test would probably best go into `tests/Resources/ti.ui.tabgroup.addontest.js`

ypbnv:
@sgtcoolguy Added the unit test. Two things that I think deserve a notice:
 - I added Google Maps API key in tiapp.xml when running the test locally.
```
  <android 
    xmlns:android="http://schemas.android.com/apk/res/android">
    <manifest>
        <meta-data android:name="com.google.android.geo.API_KEY" android:value="<API_KEY>"/>
      </application>
    </manifest>
  </android>
```
 - Even without delay sometime loading the map took more than 2000 ms (the default delay) for tests, thus I have increased it to 5 seconds and removed any time out from the code. We may want to adjust that for a better performance if that slows the tests too much.

garymathews (approved):
CR: **PASS**

sgtcoolguy:
@ypbnv @garymathews Do we have an API key we can use for the unit test app? Looks like without it the test crashes on Android and fails on iOS.

sgtcoolguy:
Ok, this is now passing the new unit tests. Ready for QE validation. Once merged it should likely be cherry-picked to 7_0_X branch.

lokeshchdhry:
FR Passed.

Closing the `tabgroup` with mapview does not crash the app.

OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.11
Appc CLI:       7.0.0
Daemon Ver:     1.0.0
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
Java Ver:       1.8.0_101

PR #9629, opened 2017-11-27
Updated View.yml per TIDOC-2804

jawa9000:
The since value for AnimationOption needs be in quotes since the API doc system considers this value as a string and not a number. Built the docs locally and confirmed that once this value was placed in quotes, the AnimationOption page renders normally.

**JIRA:** https://jira.appcelerator.org/browse/TIDOC-2804


hansemannn:
You sure this causes issues? We have hundreds of those values and they didn't seem to cause issues in the past.

sgtcoolguy:
So if this really is the underlying issue, we should try and see if we can't update the doc validation we run to catch issues like this.
But as far as I can tell, a value like `5.1.0` in yml would be treated as a string whether it's quoted or not...

@jawa9000 And in local testing, it is indeed parsing the value as a string (without this change) when I run `node docgen.js -f html`. How can I reproduce what you're using to generate docs and see this issue?

PR #9630, opened 2017-11-27
[TIMOB-25496] Update DocumentViewer.yml

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25496


hansemannn:
@garymathews @sgtcoolguy The Android TCP test keeps failing since a while, may this be related to the V8 changes?

sgtcoolguy:
That test is timing out after 5 seconds. The timeout should probably be bumped up higher.

PR #9631, opened 2017-11-28
[TIMOB-25562] Android: Automatically migrate older module projects

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25562

To be discussed:
* Can I use emojis here? Would be cool to have a fallback!
  * Decided to remove it for now to not break any old terminals
* This PR my has lint-issues which need to be resolved after Jenkins completes
* Right now the process triggers automatically, do we want to have some i/o here?

bildschirmfoto 2017-11-28 um 12 38 32


android/cli/commands/_buildModule.js
```diff
@@ -59,12 +59,53 @@ AndroidModuleBuilder.prototype.validate = function validate(logger, config, cli)
...
+				moduleId: this.manifest.moduleid,
+				platform: this.manifest.platform,
+				tisdkVersion: this.manifest.minsdk,
+				guid: this.manifest.guid,
+				author: this.manifest.author,
+				publisher: this.manifest.author // The publisher does not have an own key in the manifest but can be different. Will override below
+			});
+						
+			// Migrate missing keys which don't have a placeholder (version, license, copyright & publisher) 
+			manifestContent = manifestContent.replace(/version.*/, 'version: ' + this.manifest.version);
```
ewanharris: Maybe we could be more explicit in the matching? `/^version: \d+\.\d+\.\d+/m`, incase someone has the word version above the version key? Same for the two below?

ewanharris (commented):
Looks awesome!

Left a few comments, I imagine some prompting to trigger this would be good, just a simple yes/no using [fields.select](https://github.com/cb1kenobi/fields#fieldsselectopts) would do I guess?

This logic is pretty similar for all of the platforms (copy pasting it in to each _buildModule.js file _should_ just work), would it be possible to maybe only do this check in one place?

android/cli/commands/_buildModule.js
```diff
@@ -59,12 +59,53 @@ AndroidModuleBuilder.prototype.validate = function validate(logger, config, cli)
...
-			process.exit(1);
+		const cliModuleAPIVersion = this.cli.sdk && this.cli.sdk.manifest && this.cli.sdk.manifest.moduleAPIVersion && this.cli.sdk.manifest.moduleAPIVersion['android'];
+		const needsMigration = this.manifest.apiversion && cliModuleAPIVersion && this.manifest.apiversion !== cliModuleAPIVersion;
+
+		// Migrate modules built for earlier versions of V8 / Titanium 
+		if (needsMigration) {
+			const cliSDKVersion = this.cli.sdk.manifest.version;
+			const manifestSDKVersion = this.manifest.minsdk;
+			const manifestModuleAPIVersion = this.manifest.apiversion;
+			const newVersion = String(parseInt(this.manifest.version.split('.')[0]) + 1) + '.0.0'; // TODO: Make this less error prone
```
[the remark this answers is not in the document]
hansemannn: I tried to prevent adding new dependencies, otherwise `semver` is definitely the best choice. If that's ok to add, i'll do it.

android/cli/commands/_buildModule.js
```diff
@@ -59,12 +59,53 @@ AndroidModuleBuilder.prototype.validate = function validate(logger, config, cli)
...
+			const manifestTemplateFile = path.join(this.platformPath, 'templates', 'module', 'default', 'template', 'android', 'manifest.ejs');
+			
+			logger.info(__('⚠️  Detected Titanium %s that requires API-level %s, but the module currently only supports %s and API-level %s.', cliSDKVersion, cliModuleAPIVersion, manifestSDKVersion, manifestModuleAPIVersion));
```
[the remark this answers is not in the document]
hansemannn: Maybe use https://www.npmjs.com/package/node-emoji ?

android/cli/commands/_buildModule.js
hansemannn: I tried to prevent  restricting the regex too much. I saw modules that have a version like "4.0" instead of "4.0.0" or other possible glitches. This attempts to be the "rule the all" way of doing it.

ewanharris: That comment made me think, I could be wrong but doing semver.inc on '4.0' might return undefined/null as it wont be classed as valid semver. I think node-appc.version has a format function to pad a number to a certain number of segments that might be needed in that case

android/cli/commands/_buildModule.js
```diff
@@ -59,12 +60,53 @@ AndroidModuleBuilder.prototype.validate = function validate(logger, config, cli)
 
 		this.manifest = this.cli.manifest;
 
-		const sdkModuleAPIVersion = this.cli.sdk && this.cli.sdk.manifest && this.cli.sdk.manifest.moduleAPIVersion && this.cli.sdk.manifest.moduleAPIVersion['android'];
-		if (this.manifest.apiversion && sdkModuleAPIVersion && this.manifest.apiversion !== sdkModuleAPIVersion) {
-			logger.error(__('The module manifest apiversion is currently set to %s', this.manifest.apiversion));
-			logger.error(__('Titanium SDK %s Android module apiversion is at %s', this.titaniumSdkVersion, sdkModuleAPIVersion));
-			logger.error(__('Please update module manifest apiversion to match Titanium SDK module apiversion.'));
-			process.exit(1);
+		const cliModuleAPIVersion = this.cli.sdk && this.cli.sdk.manifest && this.cli.sdk.manifest.moduleAPIVersion && this.cli.sdk.manifest.moduleAPIVersion['android'];
```
sgtcoolguy: why is 'android' the only property using bracket notation?

Man, I wish JS had something like optional/safe navigation chaining. Looks like someday they will: https://github.com/tc39/proposal-optional-chaining

sgtcoolguy (changes requested):
So I like the idea and the code seems fine. But I'm slightly worried about the automatic nature of it.

I think this is something that the user should opt into via prompting, and that if the cli arguments are set to avoid prompting that we should just fail the build. (i.e. for CI builds)

On another note, technically we should probably track in some JSON file the mapping from apilevel to sdk version that "introduced" it so we have the minSDK value that should be used. Certainly using the current apilevel and sdk version is a nice start, though.

hansemannn:
@sgtcoolguy Good idea with the apilevel -> SDK-relation. Actually @ewanharris proposed the same today and we asked ourself if we even need an apilevel if we identify the v8-versions via the sdk-version. Anyway, I'll add some i/o, but as it's asynchronous, I'm not sure:
- if and how I should wrap the code into some kind of async-pipe?
- if we already have some i/o package installed?

Feedback appreciated!

android/cli/commands/_buildModule.js
hansemannn: I love the optionals in Swift, but yeah.. The code was taken over from the previous code, but technically the dot annotation should work the same. I'll try and change it.

hansemannn:
@sgtcoolguy Added prompts and changed the property access. Not sure if the validation may need some more love or should look different, please let me know!

lokeshchdhry:
@hansemannn , Jenkins is failing at lint step.

sgtcoolguy:
so selecting 'n', 'no', or '2' - the build still occurs when I'd expect it not to.
Presumably if we choose no we should exit with an error saying that it won't build correctly or something, right?

sgtcoolguy:
Additionally, if I pass the `--no-prompt` argument it will still prompt me.

sgtcoolguy:
Pushed a commit to address lint issue, handle when yes isn't selected, make it use a more compact prompt, have it handle `--no-prompt`.

hansemannn:
@sgtcoolguy Right now, selecting anything except „yes“ will skip the migration and attempt to build. We could log the error about mismatching api versions like before. For the prompting, I did not find a clean solution for checking the arg and will sort it out tomorrow.

@lokeshchdhry Will check the linting as well, probably it‘s the array spacing.

sgtcoolguy:
@hansemannn Already on it, I know it's late there. Go get rest 😴

hansemannn:
Thanks for taking care of it Chris!

PR #9632, opened 2017-11-28
[TIMOB-25546] Simple Code Crash on Android after installing Titanium SDK 7.0.0 Beta

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25546

**Description:**
Prevent Android from crashing when Ti.API.log() is called with non-String parameters.

**Test case:**
app.js
```
Ti.API.log({key: 'value'});
Ti.API.log('debug', {key: 'value'});
```


android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -24,7 +24,7 @@ Utf8Value::Utf8Value(v8::Local<v8::Value> value)
   if (value.IsEmpty())
     return;
 
-  v8::Local<v8::String> string = value.As<String>();
+  v8::Local<v8::String> string = value->ToString();
```
sgtcoolguy: This variant of ToString is deprecated (and will be removed in V8 6.4), and so is the variant node uses right now that takes the isolate as an argument.

They want you to use the `ToString(Local<Context>)`: https://v8docs.nodesource.com/node-8.0/dc/d0a/classv8_1_1_value.html#a1afa7fc2578fde13194e602513180a65

Unfortunately that isn't easy because we're not passing in a context or the isolate right now.

I believe the easiest way is to actually totally remove our titanium::Utf8Value and titanium::TwoByteValue usage and replace them with v8::String::Utf8Value and v8::String::Value usage (which do exactly what we're attempting to do here, but are now official APIs).

tests/Resources/ti.api.test.js
```diff
@@ -0,0 +1,66 @@
...
+		should(Ti.API.trace).be.a.Function;
+		should(Ti.API.trace('trace')).be.undefined;
+	});
+
+	it('warn()', function () {
+		should(Ti.API.warn).be.a.Function;
+		should(Ti.API.warn('warn')).be.undefined;
+	});
+
+	it('Ti.API.log with one non-String parameter', function() {
```
sgtcoolguy: just create this file as `tests/Resources/ti.api.addontest.js` with only the new/added tests inside the describe block. It's easier to do that to add new tests and handle merging them back to the suite instead of overriding the existing tests (which really should be done if you're editing existing test cases).

sgtcoolguy (changes requested):

sgtcoolguy:
@ypbnv I'll try and make this change, since it's low-level V8 stuff.

lokeshchdhry:
FR Passed.

Android does not crash when `Ti.API.log()` is called with non-String parameters.

Studio Ver:     5.0.0.201711280737
OS Ver:         10.12.3
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.11-2
Appc CLI:       7.0.0-master.49
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.9
Node Ver:       8.9.1
Java Ver:       1.8.0_101
Devices:        ⇨ samsung SM-G955U1 --- Android 7.0

PR #9639, opened 2017-11-30
[TIMOB-25576] Ensuring usage of macos bundled core utilities and shells when adding to iOS buildPhases

rlustemberg:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25576

Make sure we are using bundled Unix utilities on iOS build phases


iphone/cli/commands/_build.js
```diff
@@ -3177,7 +3177,7 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 			outputPaths: [],
 			runOnlyForDeploymentPostprocessing: 0,
 			shellPath: '/bin/sh',
-			shellScript: '"cp -rf \\"$PROJECT_DIR/ArchiveStaging\\"/ \\"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/\\""',
+			shellScript: '"/bin/cp -rf \\"$PROJECT_DIR/ArchiveStaging\\"/ \\"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/\\""',
```
hansemannn: There is a performance issue with standalone commands. Apple even recommends to [not use them](https://developer.apple.com/library/content/documentation/OpenSource/Conceptual/ShellScripting/performance/performance.html#//apple_ref/doc/uid/TP40004268-TP40003520-SW14) and gives some intersting benchmarking-tests. I don't this issue is present for many users, but @janvennemann or @sgtcoolguy may have some more insight here.

rlustemberg:
In my case I've uninstalled the homebrew coreutils package. I've just run some simple benchmarks using 'time' and the difference was negligible, 
>Richards-iMac-2:benchmarks richard$ time cp -rf orig/ArchiveStaging/ dest/
>real	0m0.087s
>user	0m0.003s
>sys	0m0.068s
>Richards-iMac-2:benchmarks richard$ time /bin/cp -rf orig/ArchiveStaging/ dest/
>real	0m0.075s
>user	0m0.003s
>sys	0m0.069s

The results were more or less the same every time, sometimes the standalone command would be faster, sometimes slower. The difference was a few percentile points , definitely not orders of magnitude such as Apple's example of the echo command. It was tested on a computer with AFS and ssd drive. I would expect the difference to be much less on machines with hard drives and HFS (if someone still works on them nowadays)

iphone/cli/commands/_build.js
cb1kenobi: I was unaware of a performance difference. Being that it's called a "shellScript", I assumed Xcode was just spawning the command in a shell in which case I don't see why `/bin/cp` would be significantly slower than `cp`. In the Apple docs, they use the example of the `echo` command which writes to stdout and perhaps that's where the speed issue is. `cp` doesn't output anything unless the verbose flag is set. So, I'm not terribly concerned with performance.

I guess I'm more concerned why `/bin` isn't in your `PATH`.

janvennemann: Agreed, we are talking about one iteration of the command here, so i think we can neglect the performance impact, which seems to be a few ms at most according to @rlustemberg benches.

hansemannn: Then let‘s take it for 7.0.1!

janvennemann (approved):

hansemannn:
Backport: https://github.com/appcelerator/titanium_mobile/pull/9663

mukherjee2 (approved):
Approved. Passed FR.

PR #9641, opened 2017-11-30
[TIMOB-25574] iOS: Prevent Ti.Media.VideoPlayer from crashing when closing a window

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25574

### DIFF
--- a/iphone/Classes/TiMediaVideoPlayerProxy.m
+++ b/iphone/Classes/TiMediaVideoPlayerProxy.m
@@ -91,37 +91,51 @@ NSArray *moviePlayerKeys = nil;
   return @"Ti.Media.VideoPlayer";
 }
 
-- (void)configureNotifications
+- (void)addNotificationObserver
 {
   WARN_IF_BACKGROUND_THREAD; //NSNotificationCenter is not threadsafe!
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -91,7 +91,7 @@ - (NSString *)apiName
   return @"Ti.Media.VideoPlayer";
 }
 
-- (void)configureNotifications
+- (void)addNotificationObserver
```
vijaysingh-axway: Please update "configureNotifications" to "addNotificationObserver" from other places also. 
e.g. inside "- (void)setUrl:(id)url_"  method.

vijaysingh-axway (changes requested):

iphone/Classes/TiMediaVideoPlayerProxy.m
hansemannn: Good catch! But from what I see, the `setUrl:` selector is the only one using it. I also added the removal of the old observer, so we don't observe an URL twice. We definitely need more unit-tests around the whole API.

vijaysingh-axway (approved):

PR #9645, opened 2017-12-01
[TIMOB-18355] Begin to enforce Java formatting via clang-format

sgtcoolguy:
**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-18355
- https://jira.appcelerator.org/browse/TIMOB-18621

**Description:**
This hooks up the ability to check java formatting, and will enforce any modified Java files meet the format.

I formatted a few files to verify the checks.

I also updated to the latest `clang-format` npm package, which required some minor reformatting of some iOS files to pass again.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+[*.{js,java}]
+charset = utf-8
+indent_style = tab
+indent_size = 4
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,6 +6,7 @@ const async = require('async'),
 module.exports = function (grunt) {
 
 	const iosSrc = [ 'iphone/Classes/*.h', 'iphone/Classes/*.m' ];
+	const androidSrc = [ 'android/runtime/v8/src/**/*.java' ];
 
 	// Project configuration.
 	grunt.initConfig({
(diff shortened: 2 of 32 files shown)


sgtcoolguy:
Here's what the danger PR comment shows when files fail formatting:
screen shot 2017-12-04 at 3 54 54 pm

sgtcoolguy:
Note that this PR adds enforcing the new formatting only to modified/created files. It does not enforce it on all files or do a bulk update of the formatting to existing files.

Personally, I'd prefer we did a single bulk formatting of all files and enforce it on all files going forward. That would affect existing PRs/branches with some conflicts/annoyance, but the PRs could be updated with the new formatting pretty easily. (run the command to format the code on the changed files).

Doing an incremental enforcement/update would avoid the merge conflicts but would introduce a potentially significant number of formatting-only changes into PRs going forward (making PRs harder to review).

I chose this as a potential starting point, however, given that there was opposition to doing a bulk format during our Engineering Week discussions.

cc @eric34 @ingo @garymathews @jquick-axway

sgtcoolguy:
Please also note the settings for the .clang-format file. I tried to effectively match up the settings with the existing Eclipse settings here: https://github.com/appcelerator/appc-standards/blob/master/languages/java/eclipse-format-android.xml

However, Eclipse, clang, IDEA, etc all have differing settings so it's impossible to get it exactly matched in all cases (and in some cases clang-format is a little funky, like it's handling of `@Kroll.name` annotations).

Please let me know if there are settings you'd prefer to change. The options are all defined here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

sgtcoolguy:
FYI, I've filed a bug against clang-format for the mishandling of the multiple annotations with periods in them: https://bugs.llvm.org/show_bug.cgi?id=35545

sgtcoolguy:
Manually merged.

PR #9647, opened 2017-12-01
[TIMOB-23496] TiAPI: Deprecate old Ti.Geolocation values

hansemannn:

### DIFF
--- a/apidoc/Titanium/Geolocation/Geolocation.yml
+++ b/apidoc/Titanium/Geolocation/Geolocation.yml
@@ -1014,11 +1014,17 @@ properties:
 
   - name: longitude
     summary: Longitude of the geocoded address.
+    description: |
+        Right now, the longitude will be returned as a String. As of Titanium SDK 8.0.0,
+        it will return a Number instead.
     type: String
     platforms: [android, iphone, ipad]
 
   - name: latitude
     summary: Latitude of the geocoded address.
+    description: |
+        Right now, the latitude will be returned as a String. As of Titanium SDK 8.0.0,
+        it will return a Number instead.
     type: String
     platforms: [android, iphone, ipad]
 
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -1014,11 +1014,17 @@ properties:
 
   - name: longitude
     summary: Longitude of the geocoded address.
+    description: |
```
sgtcoolguy: returns String only on Android. returns Number on iOS/Windows.

sgtcoolguy:
I think the general idea here is right, but two things:
- While as an American, I tend to use `zipcode`, is that the property we want to converge on? `postalCode` does seem better for i18n.
- Why converge on `country_code`? That snakecase naming doesn't match our style of camelCase, like `countryCode`.

sgtcoolguy:
So I started taking a deeper look at this PR because I'd like to get it in for 7.1.0. Turns out to be much more than I bargained for.

First, the big bug/discrepancy I see: Android and iOS do forward geocoding differently altogether. We hit an API endpoint with a param deterring "direction" of geocoding to perform. But Android has that hard-coded to 'r' for either forward or reverse calls. iOS uses 'f' properly for forward. The API endpoint returns different results entirely based on this.

With an 'f' it returns 4 comma-delimited values, the second of which is supposed to be accuracy.
With an 'r' it returns JSON with possibly multiple "places".

So looking at forward geocoding of our old office address: "377 North Bernardo Avenue, Mountain View, CA 94043, United States of America"

'f' returns:
200,1,37.976759,-85.964297

'r' returns:
{"success":true,"places":[{"country":"United States of America","country_code":"US","zipcode":"94043","city":"","street":"","state":"California","address":"CA 94043, United States of America","longitude":-122.0775,"latitude":37.4056},{"country":"United States of America","country_code":"US","zipcode":"","city":"","street":"Mountain View Dr","state":"KY","address":"Mountain View Dr, West Point, KY, United States of America","longitude":-85.964297,"latitude":37.976759}]

Android returns the first result in that array.

So in this particular case, iOS is giving bad results (see https://www.google.com/maps/place/37°58'36.3%22N+85°57'51.5%22W/@37.9760501,-85.9642112,16.47z/data=!4m5!3m4!1s0x0:0x0!8m2!3d37.976759!4d-85.964297)

Android just happens to be picking the better option because it was listed first. It also gives us much more details for the results. But it does not give us an accuracy.

So before even getting into which properties should be deprecated or not, it seems we have a pretty fundamental bug in forward geocoding here.

cc @eric34

PR #9654, opened 2017-12-05
[TIMOB-23496] (7_0_X) TiAPI: Deprecate old Ti.Geolocation values

hansemannn:

### DIFF
--- a/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
+++ b/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
@@ -328,7 +328,9 @@ public class TiLocation implements Handler.Callback
 		address.put(TiC.PROPERTY_CITY, place.optString(TiC.PROPERTY_CITY, ""));
 		address.put(TiC.PROPERTY_REGION1, ""); // AdminArea
 		address.put(TiC.PROPERTY_REGION2, ""); // SubAdminArea
-		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString("zipcode", ""));
+		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString(TiC.PROPERTY_ZIP_CODE, ""));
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -1254,22 +1260,29 @@ properties:
...
     summary: Country name.
     type: String
     platforms: [android, iphone, ipad]
 
   - name: countryCode
-    summary: Country code. On iOS, use `country_code`.
+    summary: Country code.
+    deprecated:
+        since: "7.0.0"
+        notes: Use the `country_code` property for better parity instead.
```
sgtcoolguy: Why point them at "country_code"? While it may be available on iOS/Windows, it's name doesn't match the naming scheme. I would think we'd prefer "countryCode".

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiLocation.java
```diff
@@ -328,7 +328,9 @@ private KrollDict buildAddress(JSONObject place)
 		address.put(TiC.PROPERTY_CITY, place.optString(TiC.PROPERTY_CITY, ""));
 		address.put(TiC.PROPERTY_REGION1, ""); // AdminArea
 		address.put(TiC.PROPERTY_REGION2, ""); // SubAdminArea
-		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString("zipcode", ""));
+		address.put(TiC.PROPERTY_POSTAL_CODE, place.optString(TiC.PROPERTY_ZIP_CODE, ""));
+		address.put(TiC.PROPERTY_ZIP_CODE, place.optString(TiC.PROPERTY_ZIP_CODE, ""));
```
sgtcoolguy: so it looks like zipcode is more "common", but taking a step back, which should be deprecated: `zipcode` or `postalCode`? Thinking in re: to i18n, `postalCode` is probably less US-centric.

hansemannn: Thats a bit late now :( Will try but may needs to go 8.0.0 then. Bit late here

PR #9655, opened 2017-12-05
Try to hook up npm test output/results to danger.js

sgtcoolguy:

**Description:**
We record the output from `npm test` and pipe it to a file that we can consume with danger.js to report it to the author.


sgtcoolguy:
For reference, here's what it looked like when I intentionally broke formatting:
screen shot 2017-12-05 at 4 39 27 pm

PR #9656, opened 2017-12-06
[TIMOB-23634] : iOS Toolbar on TextField doesn't hide smoothly when calling blur

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-23634


iphone/Classes/TiUITextWidget.m
```diff
@@ -202,6 +202,7 @@ - (void)setAutocapitalization_:(id)value
 - (void)textWidget:(UIView<UITextInputTraits> *)tw didFocusWithText:(NSString *)value
 {
   TiUITextWidgetProxy *ourProxy = (TiUITextWidgetProxy *)[self proxy];
+  [[ourProxy keyboardAccessoryView] setBounds:CGRectMake(0, 0, 0, [ourProxy keyboardAccessoryHeight])];
```
hansemannn: Please do not set the frame manually if possible. We have a layout helper to fill a height - although I am missing the name right now. It may be *okay* here, because the keyboard always fills the whole width and has a static position, but I feel this may produce issues with splitted keyboards on the iPad landscape.

iphone/Classes/TiUITextWidgetProxy.m
```diff
@@ -228,7 +225,9 @@ - (void)setKeyboardToolbar:(id)value
     if (keyboardUIToolbar != nil) {
       [self updateUIToolbar];
     }
-    //TODO: If we have focus while this happens, we need to signal an update.
+    [[self keyboardAccessoryView] setBounds:CGRectMake(0, 0, 0, [self keyboardAccessoryHeight])];
```
hansemannn: Same here.

iphone/Classes/TiUITextWidgetProxy.m
```diff
@@ -248,7 +244,9 @@ - (void)setKeyboardToolbar:(id)value
     [keyboardUIToolbar setItems:nil];
 
     keyboardTiView = [valueView retain];
-    //TODO: If we have focus while this happens, we need to signal an update.
+    [keyboardTiView setBounds:CGRectMake(0, 0, 0, [self keyboardAccessoryHeight])];
```
hansemannn: Same here.

iphone/Classes/TiRootViewController.m
```diff
@@ -508,19 +508,6 @@ - (void)extractKeyboardInfo:(NSDictionary *)userInfo
   }
 }
 
-- (void)placeView:(UIView *)targetView nearTopOfRect:(CGRect)targetRect aboveTop:(BOOL)aboveTop
```
hansemannn: This bunch of code aas added as part of #4569. The initial author tested against the following suites:

- KS
- KB tests
- Rotation Tests
- Tab Group Memory Tests
- Email, Contacts, Media (All modal/popover stuff)
- Open Close Focus Blur events
- LiveView restart

Can you check with QE if we still have those tests to run against them?

hansemannn (changes requested):
Left a few questions and proposals. Also there are no unit-tests so far. Some basic focus() / blur() tests should be good, as this is primarily a UI-related change.

iphone/Classes/TiRootViewController.m
vijaysingh-axway: This method was used to manage the toolbar position because it was managed manually. Now we are using native api to manage keyboard toolbar, so no need of this method. The #4569 was a bigger change which has affected many components, so it was needed to test all stuff which you have mentioned. Now testing keyboard related stuffs are sufficient.

iphone/Classes/TiUITextWidget.m
vijaysingh-axway: I think you are talking about method "[TiUtils setView:view positionRect:view.bounds]". But this is giving problem in iOS 10 and below. So there is no problem in using this.

iphone/Classes/TiRootViewController.m
vijaysingh-axway: This method was used to manage the toolbar position because it was managed manually. Now we are using native api to manage keyboard toolbar, so no need of this method. The #4569 was a bigger change which has affected many components, so it was needed to test all stuff which you have mentioned. Now testing keyboard related stuffs are sufficient.

mukherjee2 (approved):
Validated fix using this environment:
Node Version: 8.9.1
NPM Version: 5.5.1
Mac OS: 10.13.1
Appc CLI: 7.0.1
Appc CLI NPM: 4.2.11
Titanium SDK version: 7.1.0.v20171225185044
Appcelerator Studio vers 5.0.0
iOS Device 11.2

mukherjee2:
@hansemannn , does CR pass? may I merge the PR?

PR #9664, opened 2017-12-07
[TIMOB-25574] (7_0_X) iOS: Prevent Ti.Media.VideoPlayer from crashing when closing a window

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25574


ewieberappc:
FR Passed. App no longer crashes when a window with a playing video in it is closed. Can merge after CR.

PR #9665, opened 2017-12-07
[TIMOB-25424] (7_0_X) Wrap music-related API's to prevent the requirement of non-used privacy-keys

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25424


hansemannn:
@sgtcoolguy Interesting that danger complains here but not on master. Are the "no tests" changes master only?

sgtcoolguy:
yeah, they're only on master.

PR #9669, opened 2017-12-08
[TIMOB-25591] Android: Programmatically check whether app has notification permission on Android.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25591

**Description:**
Expose NotificationManagerCompat.areNotificationsEnabled method.

**Note:** Different results from this property are returned after a change has been made outside JS runtime, so maybe only check for the default value in a unit test ( I assume it is true by default )?

**Test case:**
app.js
```
Ti.API.info('Notifications are enabled - ' + Ti.Android.NotificationManager.areNotificationsEnabled());
```


apidoc/Titanium/Android/NotificationManager/NotificationManager.yml
```diff
@@ -48,6 +48,12 @@ methods:
     returns:
         type: Titanium.Android.NotificationChannel
 
+  - name: areNotificationsEnabled
+    summary: Returns whether showing notifications is enabled for the application.
+    returns:
+        type: Boolean
+    since: "7.1.0"
```
hansemannn: We need more parity for push notifications. Why not use [remoteNotificationsEnabled](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Network-property-remoteNotificationsEnabled)?

ypbnv: That option will also block local notifications, so that wording would not describe it's full effect.

garymathews (approved):
CR: **PASS**

@ypbnv But test case needs updating due to lint errors.

lokeshchdhry:
FR Passed.

`areNotificationsEnabled()` works as expected.
1. Notifications are enabled by default & `areNotificationsEnabled()` returns `true`.
2. After turning off notifications in settings for the app `areNotificationsEnabled()` returns `false`.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

PR #9675, opened 2017-12-12
[TIMOB-25599] Android: Fix Java object referencing

garymathews:
- Maintain a soft Java object reference to allow JVM to manage our Java objects

###### TEST CASE
- Select a couple of large images
- Repeatedly view each image until crash
```JS
    tableView = Ti.UI.createTableView(),
    start = Ti.UI.createButton({
        title: 'START',
        left: '10dp',
        bottom: '10dp',
    });

start.addEventListener('click',(e) => {
    Ti.Media.openPhotoGallery({
        allowMultiple: true,
        allowEditing: false,
        autohide: true,
        mediaTypes: Ti.Media.MEDIA_TYPE_PHOTO,
        success:(e) => {
        
            for (let i in e.images) {
                var row = Ti.UI.createTableViewRow({title: e.images[i].media.file.name});

                row.addEventListener('click',() => {
                    var child = Ti.UI.createWindow();
                    child.add(Ti.UI.createImageView({
                        image: e.images[i].media
                    }));

                    child.addEventListener('open',() => {
                        setTimeout(() => { child.close() }, 500);
                    });

                    child.open();
                });

                tableView.appendRow(row);
            }
        }
    });
});

window.add([tableView, start]);
window.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25599)


android/runtime/v8/src/native/JavaObject.cpp
```diff
@@ -172,8 +174,9 @@ void JavaObject::MakeJavaStrong()
 			if (stored == NULL) {
 				// Sanity check. Did we get into a state where it was weak on Java, got GC'd but the C++ proxy didn't get deleted yet?
 				LOGE(TAG, "!!! OH NO! We tried to move a weak Java object back to strong, but it's aleady been GC'd by JVM! We're in a bad state! Key: %d", refTableKey_);
+			} else {
```
sgtcoolguy: So this fix should definitely be applied. I assume if the store ref is NULL, trying to delete the local ref will be very bad.

android/runtime/v8/src/native/JavaObject.cpp
```diff
@@ -123,8 +123,10 @@ void JavaObject::MakeJSWeak()
 	// but this time, we say call us back as a finalizer so we can resurrect the
 	// object (save it from really being GCd by V8) and move it's Java object twin
 	// to a weak reference in the JVM. (where we can track when that gets GC'd by the JVM to call back and kill this)
-	persistent().SetWeak(this, DetachCallback, v8::WeakCallbackType::kFinalizer); // MUST BE kFinalizer or our object cannot be resurrected!
-	persistent().MarkIndependent();
+	if (!isDetached()) {
```
sgtcoolguy: So if you're using this quard, it's specifically checking isWeakRef_ is true, which only happens if #MakeJavaWeak() was called (and you've commented that call out from #detach() now). So I'm not sure this guard will ever be false now.

sgtcoolguy (changes requested):
So I think the last else block is correct and will avoid some crashes, but I still think this points to a memory management issue.

Ideally we should never reach a state where the JS side is "weak" and the Java side has been GC'd but suddenly the JS side is re-referenced.

The memory lifecycle here is that we basically wrap the Java object and tie it to a JS object and then explicitly tell V8 that the JS object is "independent" and to let us know when it thinks it can be GC'd because there's no more references. When we get that notification we basically stop the GC call and make the Java side weak. So at that point V8 thinks it no longer needs it and Java can try and clean it up at any point.

It's here that somehow our C++ code is trying to call getJavaObject() to resurrect the references and the Java side was already GC'd (if you're seeing that log message). So it looks like we have a reference issue on the C++ lifecycle.

sgtcoolguy:
Looking at the ticket, it's actually crashing in MakeJSWeak() so it's likely that the JS object already got GC'd.

So from the logging you noted and the crash location in this ticket, it would appear that both/either Java/JS sides are getting GC'd before we've expected them to, so somehow/where we're not properly tracking references in our C++ code, I believe.

garymathews:
@sgtcoolguy So I added the ability to create soft references to our Java objects. This also fixes the issue and makes GC less inclined to collect objects that are still in use, but still allows them to be collected when under memory constrains.

Also, `isDetached` can still return `true` even if `isWeakRef_` is `false`. I added this condition to prevent the hard crash if the Java object had already been collected.

sgtcoolguy (approved):

sgtcoolguy:
Any chance we can automate this test by just repeatedly opening/closing image views on some large image file in the test app?

garymathews:
I don't think we can create a test case for this, I wasn't able to reproduce this through automation and it's very temperamental

lokeshchdhry:
FR Passed.

Used the test above & opened large images ~4.5MB (7 images) randomly multiple times. & did not see any V8 crash.

OS Ver:         10.13.2
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.11
Appc CLI:       7.0.1
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

lokeshchdhry:
@garymathews , I still see a crash with your latest fixes but this time it took more try's:
https://gist.github.com/lokeshchdhry/0b21e9fbb04b9acfc89ebebcdc49afe5

lokeshchdhry:
Forgot to comment.

FR passed.

No app crash seen.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

PR #9679, opened 2017-12-12
[TIMOB-20363] Fix Android: Buttons inside views appear on top of all other children

maggieaxway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20363

````js

Titanium.UI.setBackgroundColor('#000');

var window = Titanium.UI.createWindow({
backgroundColor:'#fff'
});
 
// create a main view which will contain the button and the view
var mainViewContainer = Ti.UI.createView({
top: 0,
left: 0,
right: 0,
bottom:0,
backgroundColor: 'red'
});
 
// create a view which will obscure the button
var shouldObscureButton = Ti.UI.createView({
top: 0,
left: 0,
width: 200,
bottom: 0,
backgroundColor: 'blue'
});
 
// create a button which is wider than the obscure view
height: 350,
width: 350,
left: 0,
backgroundImage: 'buttonTest.png',
backgroundColor: 'yellow'
});
 
// add the button to the main view container before the obscure view
mainViewContainer.add(button);
 
// add a view that should obscure the button after the button has been added
mainViewContainer.add(shouldObscureButton);
 
// add the main view to the window and open it
window.add(mainViewContainer);
window.open();

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
@@ -8,6 +8,7 @@ package ti.modules.titanium.ui.widget;
 
 import java.util.HashMap;
 
+import android.os.Build;
 import android.text.TextUtils;
 import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollProxy;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jquick-axway** [comment]
I've tested this and noticed this removes the button's drop-shadow effect, which we don't want.

I'm pretty sure setting a Titanium view's `elevation` property to a value greater than zero will have the same z-ordering issue buttons have as well. So, this is a much bigger problem.
http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.View-property-elevation

I'm not sure what the correct solution to this is. We need to use Android elevation feature to show the drop-shadow effect. So, I'm curious if we need to something like the below?
```java
setTranslationZ(-getElevation());
```

This requires some experimentation.

maggieaxway:
@jquick-axway Hi I couldn't reproduce the button's drop-shadow effect even before the code change. I did some experiments, it seems only affecting the Z so that the layers. I have been looking into the source code but couldn't figure out where went wrong. :(

jquick-axway:
Closing since this change will remove the button's drop-shadow on Android 5.0 and higher. We need to find an alternative solution.

PR #9680, opened 2017-12-13
[TIMOB-25586](7_0_X) : iOS Crash when selecting filtered row in Ti.UI.SearchBar

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25586


hansemannn (approved):

ewieberappc:
FR Passed. Crash no longer occurs when selecting a filtered row. Verified on both tableView and listView

PR #9685, opened 2017-12-13
[TIMOB-25606] Android: Apply clang-format formatting to all Java sources

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25606

**Description:**
Bulk re-formatting of Android sources via clang-format.

Some notes:
- Had to manually disable/enable some of the formatting for annotations. clang-format doesn't like the `@Kroll.name` style annotations when multiple are applied to one method.
- Manually formatted the property accessors in annotations.
- Wrapping of lines, particularly in method calla arguments, is a little odd. I don't know exactly how it tries to do it, but it appears almost like the the wrapped contents are right aligned. It's possible we may be able to tweak some config to make it behave more sanely.
- For whatever reason, clang-format doesn't wrap braces after interface declarations or anonymous classes. Again, maybe a different brace wrapping style would "fix" that, or maybe it's a bug.


sgtcoolguy:
Manually merged (squashed)

PR #9687, opened 2017-12-13
[TIMOB-25607] Remove deprecated Ti.map API

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25607

This was deprecated in 3.2.0(!) and moved out tot he native module we ship with the SDK. It's time for it to go.


hansemannn (approved):
Important PR, go for it! I think [the docs](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Map) even quote:
> Support for the Titanium.Map module in iOS and Android was removed in Release 3.2.0.

So I don't think anyone expects it to work anymore. Also, the modules got tons of fixes and new improvements, so there are 0 reasons to keep it.

PR #9694, opened 2017-12-18
[TIMOB-25597] Android: Fixed bug where multidexed apps crash on startup on Android 4.x

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25597

**Notes:**
- Issue caused by "_build.js" creating a main dex class list based on target API Level, when it should have been based on min supported API Level. _(Android 4.x does not natively support multidexed apps. So, we must always create this list if that is the min OS version supported, regardless of the targeted version.)_
- Modified app template to init modules via reflection. This way we don't have to add module classes to main dex file, which significantly reduces main dex class list. Especially for modules that use Google Play Services. _(Avoids requiring over 64K limit in main dex file.)_
- I've verified that activities in the "AndroidManifest.xml" do not need to be in the main dex file on Android 4.x on cold startup when launched via an intent from an external app. This verifies that multidexed classes in the manifest file can be successfully loaded by the OS since our `TiApplication` class' call to `Multidex.install()` gets invoked first. Only `Application` derived class and the classes it used must be in main dex file.
- Modified to run a new'ish version of ProGuard via gradle which supports Java 8 libraries instead of using the old ProGuard included with the Android SDK which support up to Java 7.
- Fixed multidex ProGuard usage on Windows where it would fail to load multiple JAR files. (Must use semicolon ';' as a file/path array separator on that platform instead of a ':' colon.)
- Titanium SDK now contains an Android gradle template. This is needed since the Android SDK removed the gradle template last year (it's bundled with Android Studio instead).

**Test:**
Run the test attached to [TIMOB-25597](https://jira.appcelerator.org/browse/TIMOB-25597) on:
- Mac
- Windows


garymathews (approved):
CR: **PASS**

garymathews (approved):
CR: **PASS**

lokeshchdhry:
@jquick-axway , I get this error when I run with the fix on all android versions:
```
[ERROR] Application Installer abnormal process termination. Process exit value was 1
[ERROR] :  Failed to run dexer:
[ERROR] :  
[ERROR] :  Error: Can't read [/Users/lchoudhary/Desktop/android-sdk-macosx/build-tools/27.0.3/lib/shrinkedAndroid.jar] (Can't process class [android/accessibilityservice/AccessibilityService.class] (Unsupported class version number [52.0] (maximum 51.0, Java 1.7)))
```

SDK Ver:        7.0.1.GA
OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.11
Appc CLI:       7.0.1
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101
SDK Tools:      26.1.1
Platform Tools: 27.0.1
Build Tools:    27.0.3

jquick-axway:
Updated PR.

lokeshchdhry:
@jquick-axway , the build in jenkins is failing at linting. Can you please check.

lokeshchdhry:
FR Passed.

No crash seen after launch on android 4.X.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.11
Appc CLI:       7.0.1
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101
Emulator: Android 4.4.2, Android 4.4.4

mukherjee2 (approved):
Passed FR on Windows. Details are in back port version.

PR #9670, opened 2017-12-09
[TIMOB-25592] Fix module migration error when not using valid semver

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25592

PR #9695, opened 2017-12-21
[TIMOB-25621] Android: Do not delete third-party native libraries

garymathews:
- Do not clean third-party native libraries from Titanium modules
- Caused by the changes in https://github.com/appcelerator/titanium_mobile/pull/9670

###### TEST CASE
- Compile the appcelerator.encrypteddatabase module
- Verify the appropriate *.so libraries are included in the .zip output

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25621)

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -548,7 +548,6 @@ AndroidModuleBuilder.prototype.loginfo = function loginfo(next) {
 
 AndroidModuleBuilder.prototype.cleanup = function cleanup(next) {
 	fs.emptyDirSync(this.buildDir);
-	fs.emptyDirSync(this.libsDir);
 
 	this.requiredArchitectures.forEach(function (architecture) {
 		fs.mkdirsSync(path.join(this.libsDir, architecture));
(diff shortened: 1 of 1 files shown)


hansemannn:
This is not a valid change. The native libraries are recreated during the build by compiling for all architectures specified in the package.json of the Ti-Android project. The encrypted database likely failed because before the change the arm64-v8a was not always created which was caused by not cleaning up the existing libs. Please test this very well in case it should be taken.

garymathews:
@hansemannn This behavior never happened before, the issue is any third-party libraries are deleted during the newly added cleanup phase.

garymathews:
@hansemannn I have updated the PR to only remove the built module `*.so` files and not third-party libraries.

lokeshchdhry:
FR Passed.

The `.so` files are included in the compiled zip in `libs/*` folder.

OS Ver:         10.13.2
Xcode Ver:      Xcode 8.3.3
Appc NPM:       4.2.11
Appc CLI:       7.0.1
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.10.10
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

hansemannn:
@garymathews @lokeshchdhry Sorry to step in again, but this ticket should not have been merged. The first reason is that it haven't been CR-tested in any way. The second reason is that the binary that caused the ticket was obviously compiled before 7.0.1 was released. And the last third reason is that this PR will likely introduce a regression for the case that we remove an old architecture again, like in 6.0.0. In that case, the .so will be deleted, but the directory will still be present, which will cause a module build error when verifying the build architectures. Please have this in mind once it pops up.

PR #9699, opened 2018-01-04
[TIMOB-25640] iOS: Fix "playbackMode" constants in Ti.Media.VideoPlayer

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25640

### DIFF
--- a/iphone/Classes/TiMediaVideoPlayerProxy.m
+++ b/iphone/Classes/TiMediaVideoPlayerProxy.m
@@ -96,25 +96,28 @@ NSArray *moviePlayerKeys = nil;
   WARN_IF_BACKGROUND_THREAD; //NSNotificationCenter is not threadsafe!
   NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
 
-  // For durationavailable
+  // For durationavailable event
   [movie addObserver:self forKeyPath:@"player.currentItem.duration" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:nil];
 
-  // For playbackstate
+  // For playbackstate event
   [movie addObserver:self forKeyPath:@"player.rate" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:nil];
 
+  // For playbackState property
+  [movie addObserver:self forKeyPath:@"player.timeControlStatus" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:self];
+  
   // For playing
   [self addObserver:self forKeyPath:@"url" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:nil];
 
   // For load / loadstate / preload
   [movie addObserver:self forKeyPath:@"player.status" options:0 context:nil];
 
-  // naturalSize
+  // For naturalSize event
   [movie addObserver:self forKeyPath:@"videoBounds" options:NSKeyValueObservingOptionInitial context:nil];
 
-  // For complete
+  // For complete event
   [nc addObserver:self selector:@selector(handlePlayerNotification:) name:AVPlayerItemDidPlayToEndTimeNotification object:[[movie player] currentItem]];
 
-  // For error
+  // For error event
   [nc addObserver:self selector:@selector(handlePlayerErrorNotification:) name:AVPlayerItemFailedToPlayToEndTimeNotification object:[[movie player] currentItem]];
 }
 
@@ -123,6 +126,7 @@ NSArray *moviePlayerKeys = nil;
   NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
 
   [movie removeObserver:self forKeyPath:@"player.currentItem.duration"];
+  [movie removeObserver:self forKeyPath:@"player.timeControlStatus"];
   [movie removeObserver:self forKeyPath:@"player.rate"];
   [self removeObserver:self forKeyPath:@"url"];
   [movie removeObserver:self forKeyPath:@"player.status"];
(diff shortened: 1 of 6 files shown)


iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -96,25 +96,25 @@ - (void)addNotificationObserver
...
   NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
 
-  // For durationavailable
+  // For durationavailable event
   [movie addObserver:self forKeyPath:@"player.currentItem.duration" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:nil];
 
-  // For playbackstate
-  [movie addObserver:self forKeyPath:@"player.rate" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:nil];
+  // For playbackState property / playbackstate event
+  [movie addObserver:self forKeyPath:@"player.timeControlStatus" options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld context:self];
```
vijaysingh-axway: timeControlStatus is not available < iOS10. It will not get observed in < iOS10.

vijaysingh-axway (changes requested):

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -890,40 +907,77 @@ - (void)handleLoadStateChangeNotification:(NSNotification *)note
...
+    if (movie.player.currentItem.duration.value == movie.player.currentItem.currentTime.value) {
+      _playbackState = TiVideoPlayerPlaybackStateStopped;
+    } else {
+      _playbackState = TiVideoPlayerPlaybackStatePaused;
+    }
+  } else if ([TiUtils boolValue:[loadProperties valueForKey:@"autoplay"]]) {
+    _playbackState = TiVideoPlayerPlaybackStatePlaying;
+  }
+
+  if ([self _hasListeners:@"playbackstate"] && oldState != _playbackState) {
```
vijaysingh-axway: In iOS 10+, if we observe 'playbackstate' event listener like -
`
video.addEventListener('playbackstate', function(e){
  Ti.API.info(e.playbackState);
});`
Ti.Media.VIDEO_PLAYBACK_STATE_STOPPED is never fired. 
Reason which I found is _playbackState gets update before callback 'handleTimeControlStatusNotification: ' called.

vijaysingh-axway (changes requested):

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -890,40 +907,77 @@ - (void)handleLoadStateChangeNotification:(NSNotification *)note
...
-             @"height" : NUMFLOAT(movie.videoBounds.size.height)
-           }
-         }];
+  TiVideoPlayerPlaybackState oldState = _playbackState;
+  playing = movie.player.timeControlStatus == AVPlayerTimeControlStatusPlaying;
+
+  if (movie.player.timeControlStatus == AVPlayerTimeControlStatusPlaying) {
+    _playbackState = TiVideoPlayerPlaybackStatePlaying;
+  } else if (movie.player.timeControlStatus == AVPlayerTimeControlStatusPaused) {
+    if (movie.player.currentItem.duration.value == movie.player.currentItem.currentTime.value) {
```
vijaysingh-axway: should this line not be like -
`  if (movie.player.currentItem.duration.value == movie.player.currentItem.currentTime.value || !movie.player.currentItem.canStepBackward) {
` as you have made in 'handlePlaybackStateChangeNotification: '.

hansemannn: It is not necessary because it only handles the already running state that cannot interfere with other states - at least not in my tests.

iphone/Classes/TiMediaVideoPlayerProxy.m
hansemannn: @vijaysingh-axway I addressed this by making it an Optional that is not pre-set during configuration. See https://github.com/appcelerator/titanium_mobile/pull/9699/commits/daa4a9dc04a8298b8b57608f5e5868f8d0f5d72e for details. Good catch! 🙃

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -685,7 +683,11 @@ - (void)setSourceType:(id)type
 
 - (NSNumber *)playbackState
 {
-  return NUMINTEGER(_playbackState);
+  if (_playbackState != nil) {
```
vijaysingh-axway: Can we compare NSInteger  to nil ? And default value of _playbackState will be TiVideoPlayerPlaybackStateStopped.

iphone/Classes/TiMediaVideoPlayerProxy.m
vijaysingh-axway: Problem still not resolved. If we add the 'playbackstate' event listener as I suggested in above comment, when player finished playing movie it should fire "TiVideoPlayerPlaybackStateStopped" in 'playbackstate' event.

vijaysingh-axway (changes requested):
After changes also , Event is not firing.

iphone/Classes/TiMediaVideoPlayerProxy.m
hansemannn: I've commited in the wrong state. Actually, I was thinking of doing an object that could be nil but NSInteger obviously is a primitive type. The solution I came up with is to have `TiVideoPlayerPlaybackStateUnknown` which equals -1 and a) is a good default state and b) enumerates fine to 0 (stopped), 1 (playing) etc.

hansemannn:
Superseded by https://github.com/appcelerator/titanium_mobile/pull/9714

PR #9700, opened 2018-01-08
[TIMOB-17210] Android: Fixed TextField/TextArea to not trigger "change" event when setting properties

maggieaxway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17210

Please test in both Android and iOS. 

Test case:
````js
var textField = Ti.UI.createTextField(
//var textField = Ti.UI.createTextArea(
{
	value: "Hello World!",
	width: "90%",
	top: "5%",
});
textField.addEventListener("change", function(e) {
	var message = "TextField 'change' event received.\n\"" + textField.value + "\"";
	Ti.API.info(message);
	if ((Ti.Platform.name === "android") || (Ti.Platform.name === "windows")) {
		var notification = Ti.UI.createNotification(
		{
			message: message,
		}).show();
	}
});
window.add(textField);
var changeValueButton = Ti.UI.createButton(
{
	title: "Change Value",
	top: "5%",
});
changeValueButton.addEventListener("click", function(e) {
	textField.value = "Hello World";
});
window.add(changeValueButton);
var changeAttributedStringButton = Ti.UI.createButton(
{
	title: "Change Attributed String",
	top: "5%",
});
changeAttributedStringButton.addEventListener("click", function(e) {
	textField.attributedString = Ti.UI.createAttributedString(
	{
		text: "Appcelerator Titanium rocks!",
		attributes:
		[
			{
				range: [0, 2]
			},
			{
				value: "red",
				range: [0, 2]
			},
			{
				value: "blue",
				range: [2, 4]
			},
 		],
	});
});
window.add(changeAttributedStringButton);
window.open();

### CODE REVIEW

**Review by jquick-axway — CHANGES_REQUESTED** [review]
@maggieaxway, can you also fix this for the `attributedString` property too please? I've confirmed that changing the `attributedString` property for TextField and TextArea fires a `change` event as well.

**Side Note:**
I've confirmed that iOS TextFields and TextAreas do not fire a "change" event for `value` and `attributedString` property changes.

File: android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java [hunk 1]
```diff
@@ -825,10 +828,13 @@ public void setAttributedStringText(AttributedStringProxy attrString)
 		Bundle bundleText =
 			AttributedStringProxy.toSpannableInBundle(attrString, TiApplication.getAppCurrentActivity());
 		if (bundleText.containsKey(TiC.PROPERTY_ATTRIBUTED_STRING)) {
+			//TIMOB-17210 Android: A textfield change listener is wrongly triggered also if the value is programmatically set before creation
+			disableChangeEvent = true;
```
jquick-axway: Can you change this code block's `disableChangeEvent ` handling to the following please?
```java
boolean wasDisabled = disableChangeEvent;
disableChangeEvent = true;
/* ... */
disableChangeEvent = wasDisabled;
```

The reason is because the `processProperties()` method already sets the `disableChangeEvent` to `true` and this method call will end up changing it back to `false` too soon. The above code will change it back to its original state. Thanks!

jquick-axway (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
maggieaxway: You're right! I was still thinking why it's using wasDiabaled. Thanks for the explanation :)

jquick-axway (approved):
CR: Pass

longton95 (approved):
FR passed.

Tested using the test case the ticket and above, the event is only triggered when the text is changed by the user and not programmatically set.

It is now the same as iOS

ENV
```
SDK Version : 7.0.2.GA, Local 7.1.0
macOS Sierra 10.13.3
Google Pixel 2 XL (8.1.0)
android emulator (7.1.1)
iOS simulator (11.2)
Appc CLI : 7.0.3-master.3
Appc NPM : 4.2.12
Node : v8.9.1
```

longton95:
@maggieaxway What release is this ticket scheduled for?

jquick-axway:
Let's get it into 7.1.0. (I just updated the ticket.)

longton95:
@jquick-axway Would you be able to create a backport for 7_1_X

jquick-axway:
@longton95, here is the back-port: #9834

PR #9719, opened 2018-01-11
[TIMOB-25663] Android: HttpClient.cache. ImageView.cache

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25663

Add `Ti.Android.HttpResponseCache` module
Add `Ti.Network.HTTPClient` `cache` property
Add `Ti.UI.ImageView` `cache` property


hansemannn:
@drauggres Thanks for all your work on this! We will schedule this for the next release now. Can you please re-sign the [Axway CLA](https://cla.axway.com/)? It was changed lately to migrate the legal part of the swaggy open source community to our legal team.

tests/Resources/ti.android.httpresponsecache.test.js
```diff
@@ -0,0 +1,121 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe('Titanium.Android.HttpResponseCache', function () {
```
sgtcoolguy: you can just do `describe.android('Titanium.Android.HttpResponseCache', function () {` and all the nested tests all be limited to android, so that you don't need to ad date filter to each `it`.

tests/Resources/app.js
```diff
@@ -0,0 +1,235 @@
...
+require('./es6.rest.args.test');
+require('./es6.spread.args.test');
+require('./es6.string.interpolation.test');
+require('./es6.class.test');
+require('./es6.import.test');
+// Titanium APIs
+require('./ti.accelerometer.test');
+require('./ti.analytics.test');
+require('./ti.api.test');
+require('./ti.android.httpresponsecache.test');
```
sgtcoolguy: Once I merge to the full test suite, I'd modify the app.js like this, but for PRs the easier way to avoid it is to simply ensure your additional tests are named `ti.whatever.addontest.js` and they'll get picked up automatically. The `addontest.js` suffix is what gets it picked up.

tests/Resources/ti.network.httpclient.test.js
```diff
@@ -0,0 +1,596 @@
...
+				Ti.API.debug(JSON.stringify(e, null, 2));
+				finish(new Error('failed to "' + url + '": ' + e));
+			}
+		};
+
+		xhr.open('GET', url);
+		xhr.send();
+	});
+
+	it.android('HttpResponseCache xhr.cache custom', function (finish) {
```
sgtcoolguy: Again, you can just add an "add-on test" to add new unit tests to the suite and not have to copy/override the existing API test file. Makes it easier for you and for us to see the new tests added.

Copying and modifying is what's necessary for modifying existing tests (say fixing a test we skip on a given platform and re-enabling it).

So anyways, for next time, you could just create a `ti.network.httpclient.addontest.js` and add these two new tests to it.

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return 0;
+	}
+
+	@Kroll.method
+	public int getNetworkCount()
+	{
+		HttpResponseCache cache = HttpResponseCache.getInstalled();
+		if (cache != null) {
+			cache.getNetworkCount();
```
sgtcoolguy: missing `return` here.

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+	public void setHttpCachePath(String value)
+	// clang-format on
+	{
+		httpCachePath = value;
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.setProperty
+	public void setHttpCacheSize(long value)
```
sgtcoolguy: `setMaxSize(long value)`?

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+		if (cache != null) {
+			return cache.size();
+		}
+		return 0;
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.getProperty
+	public String getHttpCachePath()
```
sgtcoolguy: `getPath()`? Remove the `@Kroll.method` annotation (and then the clang-format comments are unnecessary too)?

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	@Kroll.method
+	public long size()
+	{
+		HttpResponseCache cache = HttpResponseCache.getInstalled();
+		if (cache != null) {
+			return cache.size();
+		}
+		return 0;
```
sgtcoolguy: Should this return `-1`? Or do we want to have differing values for "no cache installed" versus "cache installed, but unable to determine size"?

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
```diff
@@ -163,7 +165,14 @@ public static TiDrawableReference fromUrl(KrollProxy proxy, String url)
 		if (url == null || url.length() == 0 || url.trim().length() == 0) {
 			return new TiDrawableReference(proxy.getActivity(), DrawableReferenceType.NULL);
 		}
-		return fromUrl(proxy.getActivity(), proxy.resolveUrl(null, url));
+		boolean useCaches = false;
```
sgtcoolguy: You can shorten this to:
```java
boolean useCaches = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_CACHE), false);
```

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
```diff
@@ -239,8 +248,15 @@ public static TiDrawableReference fromObject(KrollProxy proxy, Object object)
 			object = proxy.resolveUrl(null, (String) object);
 		}
 
+		boolean useCaches = false;
```
sgtcoolguy: Same as above, you can shrink this down to a one-liner

apidoc/Titanium/Android/HttpResponseCache.yml
```diff
@@ -0,0 +1,149 @@
...
+            if(Ti.Platform.osname == "android") {
+                Ti.Android.HttpResponseCache.setHttpCachePath('_http_');
+                Ti.Android.HttpResponseCache.setHttpCacheSize(30 * 1024 * 1024);  // 30MB
+                Ti.Android.HttpResponseCache.install();
+            }
+
+extends: Titanium.Proxy
+platforms: [android]
+createable: false
+since: "7.1.0"
```
sgtcoolguy: Sorry, we clearly missed this release. This will need to be updated for whatever release we can get this in. master is currently targeting 7.4.0, but I think feature freeze may be happening for that real soon.

apidoc/Titanium/Android/HttpResponseCache.yml
```diff
@@ -0,0 +1,149 @@
...
+    parameters:
+      - name: size
+        summary: Maximum cache size in bytes.
+        type: Number
+
+properties:
+
+  - name: httpCacheSize
+    summary: Maximum  cache size in bytes.
+    description: |
```
sgtcoolguy: Maye want to add some details/notes that this needs to be set before `#install()` and once that's called this can't be changed.

apidoc/Titanium/Android/HttpResponseCache.yml
```diff
@@ -0,0 +1,149 @@
...
+        See also:
+        [maxSize](https://developer.android.com/reference/android/net/http/HttpResponseCache.html#maxSize())
+        in the Android API Reference.
+    type: Number
+
+  - name: httpCachePath
+    summary: Relative path for cache
+    description: |
+        HTTP response cache will be stored on external storage if it exists on moment
```
sgtcoolguy: Same as above, maybe more more explicit that this property is consulted at time `#install()` is called, but changing later won't do anything. (In fact we may even want to spit out some warning logs in the impl if they get called after `#install()`).

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+	@Kroll.method
+	public boolean install() throws IOException
+	{
+		HttpResponseCache cache = HttpResponseCache.getInstalled();
+		if (cache != null) {
+			cache.flush();
+			cache.close();
+		}
+		TiApplication tiApp = TiApplication.getInstance();
+		File dir = tiApp.getApplicationContext().getExternalCacheDir();
```
sgtcoolguy: Why not have the `path` (or `httpCacheDir`) property value determine the directory *and* relative path to store the cache?

I'm worried here because the Android docs point out that if the requests contain private data, the external cache dir is a bad place to stick the cache due to access controls not being enforced on external storage.

I'm assuming it's because we don't actually expose the external cache dir anywhere in our APIs? cc @garymathews @jquick-axway

sgtcoolguy (changes requested):
I think this is a very nice API addition that could be very helpful for app developers. Thanks, @drauggres !

My first concern is that the path set is only a relative path and doesn't offer control whether the cache gets put in the external cache or the app cache dir, which can have security/data privacy issues. I think the property should be the full path so the dev has control to explicitly choose that (though perhaps it could default to app cache directory or something). To enable that we may need to expose the external cache dir in our APIs.

 I'm also concerned that this is very Android-specific, as I believe iOS could also implement something similar as well here.

@hansemannn Can you please take a look and see if we couldn't make this a more general API cross-platform? Maybe hang it off `Ti.Network.Cache` or something?

tests/Resources/ti.android.httpresponsecache.test.js
drauggres: done

tests/Resources/app.js
drauggres: Remove app.js
Probably someone should update [README](https://github.com/appcelerator/titanium_mobile#unit-tests)

tests/Resources/ti.network.httpclient.test.js
drauggres: done

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
drauggres: fixed

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+	public String getHttpCachePath()
+	// clang-format on
+	{
+		return httpCachePath;
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.getProperty
+	public long getHttpCacheSize()
```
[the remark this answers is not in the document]
drauggres: `maxSize` it is

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
```diff
@@ -0,0 +1,170 @@
...
+	// clang-format on
+	{
+		return httpCacheSize;
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.setProperty
+	public void setHttpCachePath(String value)
```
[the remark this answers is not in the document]
drauggres: `httpCachePath` -> `path`, property only

apidoc/Titanium/Android/HttpResponseCache.yml
drauggres: Changed to `7.4.0`

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
drauggres: fixed

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
drauggres: fixed

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
drauggres: >Or do we want to have differing values for "no cache installed" versus "cache installed, but unable to determine size"?

I think, we shouldn't mix it.

android/modules/android/src/java/ti/modules/titanium/android/httpresponsecache/HttpResponseCacheModule.java
drauggres: >I'm assuming it's because we don't actually expose the external cache dir anywhere in our APIs?

Yes. Also this is how currently `cacheDir` calculated for [`TiResponseCache`](https://github.com/appcelerator/titanium_mobile/blob/7_2_X/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java#L470).

apidoc/Titanium/Android/HttpResponseCache.yml
drauggres: Add this note:
>`To apply changes of this property to existing HttpResponseCache instance, you should call         [install](Titanium.Android.HttpResponseCache.install) method.`

apidoc/Titanium/Android/HttpResponseCache.yml
drauggres: Add this note:
>`To apply changes of this property to existing HttpResponseCache instance, you should call         [install](Titanium.Android.HttpResponseCache.install) method.`

drauggres:
@sgtcoolguy Thanks for thorough code review.
>My first concern is that the path set is only a relative path and doesn't offer control whether the cache gets put in the external cache or the app cache dir, which can have security/data privacy issues. I think the property should be the full path so the dev has control to explicitly choose that (though perhaps it could default to app cache directory or something). **To enable that we may need to expose the external cache dir in our APIs.**

What about additional boolean property (smth like `preferExternalCacheDir`) with default value `false`? It will be much easier to implement.

jquick-axway:
I have to agree with @hansemannn comment [here](https://jira.appcelerator.org/browse/TIMOB-25663?focusedCommentId=433185&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-433185) that the API to control the response cache (deletion, size, location, etc.) should be redesigned to so that it can work on both Android and iOS.

I also don't like the idea of replacing our `TiResponseCache` with Google's `HttpResponseCache`. We've added additional APIs to our class to have better control over it and to better handle "redirect" cache handling. Whatever feature we're missing should be added to `TiResponseCache`.

I do agree that we should add "cache" property support to `HTTPClient`. That is best done via a separate simpler PR. I've already written a ticket for it here: [TIMOB-26501](https://jira.appcelerator.org/browse/TIMOB-26501)

I'd prefer to keep our `ImageView` simple. Especially from a portability standpoint. If you want greater control over HTTP handling when loading images, then that is best done via our `HTTPClient` and the downloaded image can then be applied to `ImageView`.

drauggres:
Resume:
- `Ti.Network.HTTPClient` `cache` property on Android: see [TIMOB-26501](https://jira.appcelerator.org/browse/TIMOB-26501)
- image cache control: load remote images via `HTTPClient` and control caching manually (or use [To.ImageCache](/Topener/To.ImageCache))
- general HttpResponseCache control and statistics: need to implement it in `TiResponseCache`

PR #9722, opened 2018-01-11
[TIMOB-25652] iOS: Ad-hoc build fails when packaging with WatchApp

feons:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25652

Specify the provisioning profile to use for each executable in the app, including `WatchKit App` and `WatchKit Extension`.


iphone/cli/hooks/package.js
```diff
@@ -228,6 +228,18 @@ exports.init = function (logger, config, cli) {
 		exportsOptions.provisioningProfiles = {};
 		exportsOptions.provisioningProfiles[builder.tiapp.id] = pp.uuid;
 
+		builder.extensions.forEach(function (ext) {
```
cb1kenobi: Sorry, but this code does not work. It always uses the app's provisioning profile uuid because `item.ppUUIDs` does not have a property called `target`. Second, you code does not work for iOS Extensions because it checks the target name to see if it matches "WatchKit App" or "WatchKit Extension". Furthermore, those names are arbitrary and could be changed.

You need to look up each target's bundle identifier by walking the native targets and their build configurations. If it helps, here's the code:

```
builder.extensions.forEach(function (ext) {
	const nativeTargets = ext.objs.PBXNativeTarget;
	ext.targets.forEach(function (extTarget) {
		if (extTarget.ppUUIDs[target]) {
			const targetUUID = Object.keys(nativeTargets).filter(uuid => typeof nativeTargets[uuid] === 'object' && nativeTargets[uuid].name.replace(/^"/, '').replace(/"$/, '') === extTarget.name)[0];
			const buildConf = targetUUID && ext.objs.XCConfigurationList[nativeTargets[targetUUID].buildConfigurationList].buildConfigurations.filter(c => c.comment === 'Release');
			const confUUID = buildConf && buildConf.length && buildConf[0].value;
			const id = confUUID && ext.objs.XCBuildConfiguration[confUUID].buildSettings.PRODUCT_BUNDLE_IDENTIFIER;
			if (id) {
				exportsOptions.provisioningProfiles[id] = extTarget.ppUUIDs[target];
			}
		}
	});
});
```

cb1kenobi (changes requested):

iphone/cli/hooks/package.js
feons: @cb1kenobi, updated. Thanks!

cb1kenobi (approved):
Being as the code is mine, I approve it. However it would be nice if someone else can test it before merging.

PR #9729, opened 2018-01-13
[AC-5514] Android: SearchView hintTextColor

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-5514

**Example**

```js
  fullscreen: false,
  theme: "Theme.AppCompat.Light"
});
 
// Use action bar search view
var search = Ti.UI.Android.createSearchView({
  hintText: "Table Search",
  color: "#000",
  hintTextColor: "#999",
  iconifiedByDefault: false
});
 
var data = [];
data.push(Ti.UI.createTableViewRow({
  title: 'Apple'
}));
data.push(Ti.UI.createTableViewRow({
  title: 'Banana'
}));
data.push(Ti.UI.createTableViewRow({
  title: 'Orange'
}));
data.push(Ti.UI.createTableViewRow({
  title: 'Raspberry'
}));
 
var tableview = Titanium.UI.createTableView({
  data: data,
  search: search
});
 
win.add(tableview);
win.open();
```


m1ga:
The dangerJS formating hints are a bit hard to understand :)

hansemannn:
Why do we have a Ti.UI.SearchBar AND Ti.UI.Android.SearchView again? This should receive some parity love as well. And then expose this property on iOS as well.

jquick-axway:
@hansemannn, the `Ti.UI.SearchBar` is a custom made view we made on Android. `Ti.UI.Android.SearchView` is a native implementation which Google didn't add until Android 3.0.

My understanding is that we made a custom search bar implementation back when Titanium still needed to support Android OS 2.x. The two views look and function slightly different, which is why we haven't replaced the custom implementation with the native one. But I do think the native implementation looks better. Perhaps we should consider swapping it out in Titanium 8.

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
**FR passed:** Was able to see the hint text when a light theme is used.

**Test steps:** 
+ Built a new SDK from  this PR
+ Created a new titanium project
+ Added the test case mentioned above in to the project 
+ Ran the program 
+ Was able to "Table search" which was set as the hint text

**Test Environment**
Appcelerator Command-Line Interface, version 7.0.1
Nexus 6p (8.1.0)
Node.js Version: 8.9.1
Appcelerator Studio: 5.0.0.201712081732

PR #9731, opened 2018-01-14
[TIMOB-25646] (7_0_X): iOS After Listview Search overlay not appearing

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25646


iphone/Classes/TiUIListView.m
```diff
@@ -2433,6 +2339,61 @@ - (void)handleTap:(UITapGestureRecognizer *)tapGestureRecognizer
...
+}
+
+- (void)dismissSearchController
+{
+  [searchController setActive:NO];
+}
+
+- (void)keyboardWillChangeFrame:(NSNotification *)notification
+{
+  NSDictionary *userInfo = [notification userInfo];
```
hansemannn: Unused property. Either remove or change the below statement to use it. Same for `keyboardDidChangeFrame` and the two implementations for the TiUITableView.

hansemannn (changes requested):
Left a minor note. Also there seems to be a UI-glitching when switching orientations between portrait and landscape as attached [here](https://www.dropbox.com/s/mxoosn5lcws1dkx/search-glitch.mov?dl=0).

iphone/Classes/TiUIListView.m
vijaysingh-axway: Change the statement to use the property. 
We are managing the view of SearchController, due to that there is some glitch in orientation change. In slow animation (debug), it is more visible. In normal orientation it will not be visible that much. I'll create a ticket to work on this issue.

hansemannn (approved):

mukherjee2 (approved):
Node Version: 8.9.1
NPM Version: 5.5.1
Mac OS: 10.13.1
Appc CLI: 7.0.1
Appc CLI NPM: 4.2.11
Titanium SDK version: 7.0.2 (with artifacts from this PR)
iOS 11.1 Device

Was able to reproduce the issue on the affected version. On fixed version, overlay appeared correctly after Listview search.

PR #9732, opened 2018-01-14
[TIMOB-24909] iOS: Move Ti.Media.AudioPlayer to AVPlayer

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24909


apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -231,51 +264,89 @@ events:
...
+        since: "7.2.0"
+        notes: Use <Titanium.Media.AUDIO_STATE_BUFFERING> instead.
 
   - name: STATE_INITIALIZED
     summary: Audio playback is being initialized.
     type: Number
     permission: read-only
+    deprecated:
+        since: "7.2.0"
+        notes: Use <Titanium.Media.AUDIO_STATE_BUFFERING> instead.
```
vijaysingh-axway: notes  should be changed with relevant state instead of  'Titanium.Media.AUDIO_STATE_BUFFERING'.

apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -231,51 +264,89 @@ events:
...
 
   - name: STATE_PAUSED
     summary: Playback is paused.
     type: Number
     permission: read-only
+    deprecated:
+        since: "7.2.0"
+        notes: Use <Titanium.Media.AUDIO_STATE_BUFFERING> instead.
```
vijaysingh-axway: Same as above

apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -231,51 +264,89 @@ events:
...
 
   - name: STATE_PLAYING
     summary: Audio playback is active.
     type: Number
     permission: read-only
+    deprecated:
+        since: "7.2.0"
+        notes: Use <Titanium.Media.AUDIO_STATE_BUFFERING> instead.
```
vijaysingh-axway: Same as above

iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -47,300 +44,501 @@ - (NSString *)apiName
...
-  }
-}
+#pragma mark Deprecated APIs
 
-- (void)restart:(id)args
+- (void)setPaused:(NSNumber *)paused
 {
-  BOOL playing = [player isPlaying] || [player isPaused] || [player isWaiting];
-  [self destroyPlayer];
+  DEPRECATED_REPLACED(@"Media.AudioPlayer.setPaused", @"7.1.0", @"Media.AudioPlayer.pause");
```
vijaysingh-axway: Message should have 7.2.0 instead 7.1.0.

iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -47,300 +44,501 @@ - (NSString *)apiName
...
   } else {
-    // just create it
-    [self player];
+    [[self player] play];
   }
 }
 
+- (void)play:(id)unused
+{
+  DEPRECATED_REPLACED(@"Media.AudioPlayer.play", @"7.1.0", @"Media.AudioPlayer.start");
```
vijaysingh-axway: Same as above

iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -47,300 +44,501 @@ - (NSString *)apiName
...
+
+  switch (_player.status) {
+  case AVPlayerStatusUnknown:
+  case AVPlayerStatusFailed:
+    _state = TiAudioPlayerStateStopped;
+    break;
+  case AVPlayerStatusReadyToPlay:
+    if (_player.rate == 1.0) {
+      _state = TiAudioPlayerStatePlaying;
+    } else if (_player.currentItem.duration.value == _player.currentItem.currentTime.value || oldState == TiAudioPlayerStateStopping) {
```
vijaysingh-axway: Please validate state with similar test case mentioned in ticket TIMOB-25801. I guess you have to adjust a bit.

iphone/Classes/TiMediaAudioPlayerProxy.m
```diff
@@ -47,300 +44,501 @@ - (NSString *)apiName
...
-  if (player != nil && [player isPlaying]) {
-    double value = 0;
+  TiAudioPlayerState oldState = _state;
 
-    if (player.bitRate != 0.0) {
-      value = player.progress;
+  if (_player.timeControlStatus == AVPlayerTimeControlStatusPlaying) {
+    _state = TiAudioPlayerStatePlaying;
+  } else if (_player.timeControlStatus == AVPlayerTimeControlStatusPaused) {
+    if (_player.currentItem.duration.value == _player.currentItem.currentTime.value || oldState == TiAudioPlayerStateStopping) {
```
vijaysingh-axway: Same as above.

vijaysingh-axway (changes requested):
Please address mentioned issues. Thanks!

hansemannn:
@vijaysingh-axway Can you re-review please?

vijaysingh-axway (approved):
CR approved for iOS.

tests/Resources/ti.media.audioplayer.test.js
```diff
@@ -0,0 +1,96 @@
...
+	});
+
+	it('.url', function () {
+		should(audioPlayer.url).be.a.String;
+		should(audioPlayer.getUrl).be.a.Function;
+		should(audioPlayer.setUrl).be.a.Function;
+		should(audioPlayer.url).eql(audioPlayer.getUrl());
+	});
+
+	it('#start', function () {
```
sgtcoolguy: Write some more complex test cases, like say this one:

- make sure the audio's `isPlaying()`, `isPaused()`, `playing`, `paused` return false; `idle` returns true. `progress == 0`.
- Call `#start()`
- make sure the audio's `isPlaying()` and `playing` return true, `idle`, `paused` and `isPaused()` return false. Check `state`
- Call `#pause()`
- check the properties/methods again. Check that `progress` > 0.
- Call `#start()`, check the state properties/methods
- Call `#stop()`, check the state properties/methods

sgtcoolguy (dismissed):
The 7.3.0 references need to be updated to 7.4.0. I'd like to see more a comprehensive test as well

android/modules/media/src/java/ti/modules/titanium/media/AudioPlayerProxy.java
```diff
@@ -243,6 +252,31 @@ public int getAudioSessionId()
...
+		return false;
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.setProperty
+	public void setMuted(boolean muted)
+	// clang-format on
+	{
+		TiSound s = getSound();
```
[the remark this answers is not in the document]
hansemannn: Looks like we could use it below 24 when using the support library described in [the docs](https://developer.android.com/studio/write/annotations) ? Except, using Java 8 to compile should do it already.

android/modules/media/src/java/ti/modules/titanium/media/TiSound.java
```diff
@@ -426,6 +426,21 @@ public int getDuration()
 		return duration;
 	}
 
+	public boolean isMuted()
+	{
+		AudioManager audioManager =
+			(AudioManager) TiApplication.getInstance().getApplicationContext().getSystemService(Context.AUDIO_SERVICE);
```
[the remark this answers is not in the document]
hansemannn: I did not guard it because other usages in that class do not as well (like `requestAudioFocus`), but that's not the best reason :-). Would love to know the nullability of these methods!

apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -372,6 +450,44 @@ properties:
...
+        desire to play at the natural rate of the current item.
+    description: |
+        Setting the value of rate to 0.0 pauses playback, causing the value of 
+        the `state` property change to <Titanium.Media.AUDIO_STATE_PAUSED>.
+        Setting the rate to a non-zero value causes the value of the `state` property 
+        to become either <Titanium.Media.AUDIO_STATE_WAITING_FOR_QUEUE> or 
+        <Titanium.Media.AUDIO_STATE_PLAYING>, depending on whether sufficient 
+        media data has been buffered for playback to occur and whether the player's 
+        default behavior of waiting in order to minimize stalling is permitted.
+    default: 0.0
```
jquick-axway: Shouldn't the default rate be 1.0?

jquick-axway: Also, I would assume setting rate to 2.0 would cause it to play audio twice as fast, right?
We might want to state that to make it extra clear what it does.

apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -426,7 +542,10 @@ properties:
     platforms: [iphone, ipad]
 
   - name: bufferSize
-    summary: Size of the buffer used for streaming, in bytes.
+    summary: Size of the buffer used for streaming, in milliseconds.
```
jquick-axway: I don't think we should change `bufferSize` from bytes to milliseconds. Especially since we already have a `duration` property which provides this.

apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -145,14 +146,36 @@ methods:
...
   - name: release
     summary: Stops buffering audio data and releases audio resources.
     description: |
         On Android, this method should be called when you are done streaming a given
         audio object, to release underlying resources, including buffered data.
-    platforms: [android]
+
+        On iOS, this method can be called for parity to release the current player
+        instance and it's observers. Note: If you are listening to the `progress` event,
+        you should remove and re-add the event listener.
```
jquick-axway: Question:
On iOS, what happens if you call an `AudioPlayer` object's methods/property after it has been released?

apidoc/Titanium/Media/AudioPlayer.yml
hansemannn: By default, the rate is 0.0, indicating that it is **not playing**. Once started, it turns to 1.0. Once paused or stopped, it moves to 1.0 again. For x.0, it would play x as fast **or** slow, yep! I will add that to the docs.

apidoc/Titanium/Media/AudioPlayer.yml
hansemannn: I may missing some info here: Can we convert milliseconds to bytes? Is "bytes" simply "BPM"?

apidoc/Titanium/Media/AudioPlayer.yml
hansemannn: The player will be re-initialized by the `[self player]` convenience method. But if Android behaves differently, I could block an instance from being re-initialized again!

apidoc/Titanium/Media/AudioPlayer.yml
jquick-axway: Does AVPlayer provide the bitrate? If so, then you can use that in combination with duration to calculate the byte buffer size.

Also, while some audio formats such as MP3 have a variable bitrate, an audio player typically decodes it to a fixed bitrate (say 16bit) in PCM form before playing it. This is the buffer size we're after since it indicates how much RAM the decoded audio file is consuming (assuming that we're not streaming).

apidoc/Titanium/Media/AudioPlayer.yml
jquick-axway: Currently, Android does the same via its `AudioPlayerProxy.getSound()`.

hansemannn: Then we are fine for parity at that one? :-)

apidoc/Titanium/Media/AudioPlayer.yml
jquick-axway: So, this doesn't quite match my expectation based on my experience with other audio libraries such as OpenAL and Microsoft's XAudio2. In those libraries, rate is a separate setting and would not be 0.0 if paused.

In XAudio2, the equivalent of rate would be `IXAudio2SourceVoice.SetFrequencyRatio()`.
In OpenAL, the equivalent of rate would be `AL_PITCH`.
(I've never used Android's OpenSL library. So, I'm not sure what it is there.)

Is 0.0 what you're getting from Apple?

apidoc/Titanium/Media/AudioPlayer.yml
jquick-axway: Yes. :)

PR #9735, opened 2018-01-15
[TIMOB-25674] Fix regex to allow folders beginning with a platform name

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25674

This only occurs in the iOS build, Android uses a different method of filtering


cb1kenobi (approved):

cb1kenobi:
Looks good to me.

PR #9736, opened 2018-01-15
[7_0_X][TIMOB-25674] Fix regex to allow folders beginning with a platform name

ewanharris:
7_0_X backport for #9735

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25674

This only occurs in the iOS build, Android uses a different method of filtering


sgtcoolguy (approved):
Good for QE to test

ewieberappc (approved):
FR Passed. Folders beginning with a platform name are present in builds

PR #9737, opened 2018-01-16
[TIMOB-25676] iOS: Fix Ti.Network.HTTPClient states to pass unit-tests

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25676

`no tests` because it fixes an existing unit-test.


vijaysingh-axway (changes requested):
As mentioned in [PR9738](https://github.com/appcelerator/titanium_mobile/pull/9738) issue is still there.

hansemannn:
Does not seem to be an issue anymore

PR #9738, opened 2018-01-16
[TIMOB-25676] (7_0_X) iOS: Fix Ti.Network.HTTPClient states to pass unit-tests

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25676

`no tests` because it fixes an existing unit-test.

### DIFF
--- a/iphone/Classes/TiNetworkHTTPClientProxy.m
+++ b/iphone/Classes/TiNetworkHTTPClientProxy.m
@@ -373,7 +373,12 @@ extern NSString *const TI_APPLICATION_GUID;
 - (void)request:(APSHTTPRequest *)request onReadyStateChange:(APSHTTPResponse *)response
 {
   if (hasOnreadystatechange) {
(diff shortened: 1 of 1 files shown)


vijaysingh-axway (changes requested):
1. I am still getting the sequence - 

2018-01-17 15:57:33.822329+0800 Titanium[96028:1683094] [INFO] 1
2018-01-17 15:57:33.842489+0800 Titanium[96028:1683094] [INFO] 3
2018-01-17 15:57:33.843038+0800 Titanium[96028:1683094] [INFO] 3
2018-01-17 15:57:33.846696+0800 Titanium[96028:1683094] [INFO] 4
2018-01-17 15:57:33.847337+0800 Titanium[96028:1683094] [INFO] DONE

2. Sometimes I am getting error "ERROR: onreadystatechange done fired before 100% progress" also. For this as for as I understand, you should fire "ondatastream" also on main thread. Other wise it is possible to send it later than the events which are fired on main thread. Other events which has relative sequence significance, need to fire on main thread.

hansemannn:
Removing from 7.0.2 again. I cannot reproduce the issues @vijaysingh-axway but let's take a closer look before squashing this into 7.0.2.

PR #9739, opened 2018-01-16
Fix apidocs incorrectly reporting default set of platforms for mobileweb-specific entries that had mobile web removed from platforms array

sgtcoolguy:

**Description:**
- Add validation that the 'platforms' array in our apidocs must not be empty. 
- Fix up empty platforms array entries, removing the properties/events that had them (which was an artifact or removing mobileweb from the array). 

This should prevent accidently falling back to the default platform listing incorrectly. To fall back to default platforms, you must remove the platforms key altogether, not have an empty one.


apidoc/Titanium/UI/Switch.yml
```diff
@@ -35,12 +35,10 @@ methods:
         to appear on top of it.
 
         See also: <Titanium.UI.View.add>.
-    platforms: []
 
   - name: remove
     description: |
         See also: [add](Titanium.UI.Switch.add), <Titanium.UI.View.add>.
-    platforms: []
```
hansemannn: I don't think these methods should stay. Instead, it should be added to the `exclude` array of inherited methods.

jquick-axway: I'm actually okay with this because you can't add/remove child views to a Switch on Android either.

jquick-axway (approved):
Looks good!

I've grep'ed the docs and all of the `platform: []` issues appear to have been addressed.

apidoc/Titanium/UI/Switch.yml
hansemannn: So now the docs will indicate that it works for both iOS and Android because thats the behavior when there is no `platforms` tag. Not sure if I like that.

jquick-axway: `Switch` inherits from `View`. The `View` type already has `add()` and `remove()` methods. What the above documentation indicates is that these methods are not supported on all platforms, which is true. You can call these methods on Android, but they'll no-op.

hansemannn: What I meant above is to do it like [here](https://github.com/sgtcoolguy/titanium_mobile/blob/63d1c4afe2288763d391264174443e890b475520/apidoc/Titanium/UI/AlertDialog.yml#L75), which couldn't be done in the past because Mobileweb did support it. Let's just leave it then.

PR #9741, opened 2018-01-17
[TIMOB-25680] iOS: Workaround the iOS 11.2+ document-viewer issue

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25680

`no tests` because it's a change unit-tests cannot measure.


vijaysingh-axway (approved):
CR and FT passed.

PR #9749, opened 2018-01-18
Update the pre-packaged module JSON listing to use integrity hashes

sgtcoolguy:
**Description:**
This is more secure as we can verify the contents haven't been changed unexpectedly, and it also allows us to retain cached copies of the module zips in the OS temp folder. This should avoid the need to constantly re-download modules when the listing changes so infrequently.

This was driven by a desire to avoid re-downloading module zips constantly, particularly to avoid saturating the office network due to CI builds.

I also added two developer `scons` commands to generate these integrity hashes.
`scons ssri <url-list>` will generate integrity hashes for one or more URLs (space separated cmd line args).
`scone modules-integrity` will read the support/module/packaged/modules.json listing and spit out JSON output with the correct integrity hashes filled in. This is useful to pre-populate the value when a module url has changed. (or much rarer, when the url stayed the same but the contents changed and we're ok with the new contents).

We use the same `ssri` library for integrity hashes as that used by npm.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,8 +15,8 @@ def isMainlineBranch = true // used to determine if we should publish to S3 (and
 def isFirstBuildOnBranch = false // calculated by looking at S3's branches.json
 
 // Variables we can change
-def nodeVersion = '6.10.3' // NOTE that changing this requires we set up the desired version on jenkins master first!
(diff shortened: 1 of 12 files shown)


cb1kenobi:
This is cool!

PR #9750, opened 2018-01-18
[TIMOB-25683] iOS: Refactor Ti.Filesystem, fix broken tests, use ARC

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25683


iphone/Classes/TiFile.m
```diff
@@ -63,7 +61,7 @@ - (unsigned long long)size
 
 - (id)blob
 {
-  return [[[TiBlob alloc] _initWithPageContext:[self pageContext] andFile:path] autorelease];
+  return [[TiBlob alloc] _initWithPageContext:[self pageContext] andFile:_path];
```
vijaysingh-axway: Why autorelease removed? will it not cause memory leak?

hansemannn: I moved the whole class to ARC

vijaysingh-axway: Cool. Can you once check your test case in ticket. It is giving error for testRecursiveDirectories() and for others its giving 0 but you have mentioned it should be Double. Am I missing anything? Thanks!

hansemannn: I think there are some issues with the change in general, some paths are released too early. Will put into progress again for now.

vijaysingh-axway (changes requested):
As per comment in PR, there is some issue.

hansemannn:
Closing in favor of #10019

PR #9758, opened 2018-01-19
[TIMOB-25685] Android: TabGroup 'close' event is never fired

garymathews:
- Implement `close` event for `Titanium.UI.TabGroup`

###### TEST CASE
```JS
	tabGroup = Ti.UI.createTabGroup();
	tab = Ti.UI.createTab({
		title: 'Tab',
		window: win
	});

tabGroup.addEventListener('close', function() {
	console.log('TabGroup.close()');
});

tabGroup.addTab(tab);
tabGroup.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25685)


hansemannn:
@garymathews Can you update the docs? They [currently](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.TabGroup-event-close) indicate to be iOS only. Also it seems like the test fails on iOS.

garymathews:
@hansemannn Looks like it wasn't implemented for iOS either

iphone/Classes/TiUITabGroup.m
```diff
@@ -636,6 +636,10 @@ - (void)open:(id)args
 
 - (void)close:(id)args
 {
+  if ([self.proxy _hasListeners:@"close"]) {
+    [self.proxy fireEvent:@"close" withObject:event];
+  }
```
hansemannn: It should be called event-driven (by delegates that actually execute the closing. Let me find a better place to put it.

ypbnv (approved):
CR: Pass

iphone/Classes/TiUITabGroup.m
garymathews: Isn't this `close` method called by [windowDidClose](https://github.com/garymathews/titanium_mobile/blob/88fb153fb72d4d0fb8a560800c62e3394e92737a/iphone/Classes/TiUITabGroupProxy.m#L157) anyway?

hansemannn: It is (by `windowWillClose`), but if you call it manually, the `close` event may fire before the window actually finishing closing. I'll prepare a test-case tomorrow.

hansemannn: Ok, so let's do this (as the above test-case does not actually close the tab group):
```js
	tabGroup = Ti.UI.createTabGroup();
	tab = Ti.UI.createTab({
		title: 'Tab',
		window: win
	});
  
  setTimeout(function() {
    tabGroup.close();
  }, 2000);
})

tabGroup.addEventListener('close', function() {
	console.log('TabGroup.close()');
});

tabGroup.addTab(tab);
tabGroup.open();
```
It will wait 2 seconds after the application finished launching and close the tab-group. First, `windowWillClose` is called to cleanup the navigation stack. Once done, it will will invoke it's super-class that will clean up all child views (TiViewProxy instances). 

Then, finally, the TiWindowProxy, which the TiUITabGroupProxy inherits from cleans up (dismisses the actual controller, fire the child close events) and sends `windowDidClose` to `TiUITabGroupProxy`.

So after all, the close event in its current state would fire before the tabgroup actually finished closing, which is different from the behavior of other root view controllers (like window or split-window on iOS).

And as you stated above, because the class inherits from TiWindowProxy, is should fire the event already (but the link you referenced was linking to the wrong method).

hansemannn (changes requested):
Remove the iOS change as discussed above.

lokeshchdhry:
FR Passed.

For android, the tabgroup `close` event is fired successfully.

Studio Ver:     5.2.0.201804230823
OS Ver:         10.13.4
Xcode Ver:      Xcode 9.3
Appc NPM:       4.2.13
Appc CLI:       7.0.3
Daemon Ver:     1.1.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_131

lokeshchdhry:
@garymathews , Can you please look at the changes that need to be done on the IOS part. Also it would be great if you can look at the conflicts as well.
Thanks !!

sgtcoolguy:
@hansemannn We'd like to get this PR in for 7.3.0, but it seems there's some discussion around the iOS changes. Can you please take a look and see if you can provide an alternate fix for iOS here?

hansemannn:
@garymathews iOS looks fine (even without this change):
```js
	tabGroup = Ti.UI.createTabGroup();
	tab = Ti.UI.createTab({
		title: 'Tab',
		window: win
	});

tabGroup.addEventListener('close', function() {
	console.log('TabGroup.close()');
});

tabGroup.addEventListener('open', function() {
    setTimeout(function () {
        console.log('CLOSING NOW');
        tabGroup.close();
    }, 1000);
})

tabGroup.addTab(tab);
tabGroup.open();
```

sgtcoolguy:
I'm taking a look at this locally and will try to resolve the (possible) issue. Looks to me like the unit test just was written out-of-order, resulting in the iOS failure.

hansemannn:
Seems to pass now 👏

PR #9759, opened 2018-01-20
[TIMOB-23888] iOS: Use asynchronous openURL API on iOS 10+

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23888


iphone/Classes/PlatformModule.m
```diff
@@ -271,8 +271,34 @@ - (NSNumber *)openURL:(NSArray *)args
   NSString *newUrlString = [args objectAtIndex:0];
   NSURL *newUrl = [TiUtils toURL:newUrlString proxy:self];
   BOOL result = NO;
+
+  // iOS 10+
+  KrollCallback *callback = nil;
+  NSMutableDictionary *options = [NSMutableDictionary dictionary];
+
+  if ([args count] > 0) {
+    if ([[args objectAtIndex:1] isKindOfClass:[NSDictionary class]]) {
```
vijaysingh-axway: In case of count = 1, [args objectAtIndex:1] will fail. Due to this first test case is failing.

apidoc/Titanium/Platform/Platform.yml
```diff
@@ -43,13 +43,38 @@ methods:
...
         summary: The url to open.
         type: String
+      - name: options
+        summary: |
+            The optional options to pass to the URL handling (iOS 10+). Pass a
+            dictionary with one or more of the following string-keys:
+                * UIApplicationOpenURLOptionsSourceApplicationKey
+                * UIApplicationOpenURLOptionsAnnotationKey
+                * UIApplicationOpenURLOptionsOpenInPlaceKey
+                * UIApplicationOpenURLOptionUniversalLinksOnly
```
vijaysingh-axway: We should give the detail about the values of these string-keys. e.g For UIApplicationOpenURLOptionUniversalLinksOnly value should be Bool or give an apple link here.
And normally we map apple keys with Titanium keys. Isn't it good to use titanium keys?

vijaysingh-axway (changes requested):

apidoc/Titanium/Platform/Platform.yml
hansemannn: The reason I've not made this a constant is because we cannot use constants as dictionary-keys, so the dev would need to declare a variable first and then use that one as a key. I'll add more infos about the constants though!

vijaysingh-axway (approved):
CR passed.

PR #9506, opened 2017-10-06
[TIMOB-25202] Update to windowlib@0.6.0

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/[TIMOB-25202

Update to windoslib@0.6.0 which filters emulators provided through the detect function using the options.supportedWindowsPhoneSDKVersions value provided by the SDK

PR #9760, opened 2018-01-22
Update to windowlib@0.6.4

infosia:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25616

Update to windowlib@0.6.4

Windowslib have been uninstalling existing app in order to complete the installation when same version of the app is already installed, but we should not uninstall existing app unless user explicitly specifies `forceUnInstall`option. It should ends up showing error message when `forceUnInstall=false` and there's same version of the app installed, which should be the default behavior of `wpToolInstall`.

@ewanharris I saw that we have been updating `package-lock.json` too according to #9506 but I'm not sure how to get it updated. Do we need to use kind of update command or something for `package-lock.json` ?


sgtcoolguy:
@infosia To update `package-lock.json` at the same time, instead of editing `package.json` manually, use an `npm` command to do so:
`npm install --save-prod windowslib@0.6.4`

That says to install 0.6.4 of windowslib and save it as a production dependency in the package.json file. The package-lock.json file gets updated as a side-effect and contain details about what exact versions of the dependency (and it's dependency tree) to use.

infosia:
@sgtcoolguy Awesome thanks 👍 ..Looks like the build failed because [windowslib 0.6.4 is not released onto npm](https://github.com/appcelerator/windowslib). Do you have any idea on that?

sgtcoolguy:
@infosia windowslib hasn't been getting published because the build was failing due to test failures. I've loosened that build to get marked unstable if tests fail, which means it should now release even if the tests fail.You may want to take a look at the failures though:
https://jenkins.appcelerator.org/job/cli/job/windowslib/job/master/24/testReport/

PR #9767, opened 2018-01-24
[TIMOB-15091] iOS: Allow using module views in Ti.UI.ListView templates

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-15091


vijaysingh-axway (approved):
CR and FT passed.

hansemannn:
Moving out of 7.1.0 due to missing Android parity.

sgtcoolguy:
Pushing to 7.4.0 as we should have an Android equivalent merged at the same time.

sgtcoolguy:
@garymathews @jquick-axway I've opened https://jira.appcelerator.org/browse/TIMOB-26138 to address the Android side of this. We shouldn't merge this PR until we have the Android equivalent...

hansemannn:
@sgtcoolguy Let's close this PR and the related ticket. Modules can be used with #10131 by simply returning the view via `getView()` in a module, like:
```js
var map = require('ti.map');
var mapView = map.createView({ /** ... **/ });

exports.getView = () => {
  return mapView;
};
```
so introducing another `module` property is nonsense. Or am I missing something?

vijaysingh-axway:
@sgtcoolguy Should we close this PR as suggested by Hans in above comment?

lokeshchdhry:
@sgtcoolguy , @vijaysingh-axway , What are we doing with this PR ?

PR #9769, opened 2018-01-24
[TIMOB-4789] Add Worker API

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-4789

Tests to be added later, using `no tests` label for now.

## Demo

**app.js**
```js
  backgroundColor: '#fff'
});

  title: 'Trigger'
});

  const task = new Worker('echo.js');

  task.onerror = (error) => {
    Ti.API.error('Error!');
    Ti.API.eror(error.message); // A string constant
  };

  task.onmessage = (message) => {
    Ti.API.info('New message!');
    alert(message.data);

    task.terminate();
  };

  task.onmessageerror = (error) => {
    Ti.API.error('Message error!');
    Ti.API.eror(error.message); // A string constant
  };

  task.postMessage({
    message: 'Hello'
  });
});

win.add(btn);
win.open();
```

**echo.js**
```js
// work in progress
onmessage = (message) => {
  postMessage(message.data);
};
```


sgtcoolguy:
It failed to compile the iOS side on Jenkins...

Also, here's an example web worker pointed at by the MDN docs: https://github.com/mdn/simple-web-worker/blob/gh-pages/worker.js
```javascript
onmessage = function(e) {
  console.log('Message received from main script');
  var workerResult = 'Result: ' + (e.data[0] * e.data[1]);
  console.log('Posting message back to main script');
  postMessage(workerResult);
}
```

iphone/Classes/KrollContext.h
```diff
@@ -52,6 +52,13 @@
 #endif
 }
 
+// Worker properties
+bool TiWorker_setProperty(TiContextRef jsContext, TiObjectRef object, TiStringRef prop, TiValueRef value, TiValueRef *exception);
```
sgtcoolguy: So clearly if we're going to keep adding new types to our global, it'd be better to not clutter up KrollContext with all the methods. I believe in Windows, we point at the headers of all the proxy types and build the type in it's own definition/file. So it's not super-clean but it limits the exposure of the proxy type's implementation.

So basically we have assign the name "Worker" and the value is a call to something like TiWorker.Class() which sets up the JS type for JSC (it's parent type, the method callbacks for Initialize/Cosntructor, static functions, etc.)

That way TiWorker_setProperty/TiWorker_postMessage/TiWorker_terminate would move over to TiWorker.h/m

iphone/Classes/KrollContext.m
```diff
@@ -1141,6 +1143,109 @@ - (int)forceGarbageCollectNow
   return 0;
 }
 
+// Worker class definition
+TiClassRef TiWorker_class(TiContextRef context)
```
sgtcoolguy: I'd move all the TiWorker_* stuff over to TiWorkerProxy.m/h. It's specific to that proxy/type.

iphone/Classes/KrollContext.m
```diff
@@ -1141,6 +1143,109 @@ - (int)forceGarbageCollectNow
...
+{
+  static TiClassRef jsClass;
+  if (!jsClass) {
+    TiClassDefinition definition = kTiClassDefinitionEmpty;
+    definition.className = "Worker";
+    definition.initialize = TiWorker_initialize;
+    definition.finalize = TiWorker_finalize;
+    definition.staticFunctions = TiWorker_staticFunctions;
+    definition.setProperty = TiWorker_setProperty;
```
sgtcoolguy: add:
`definition.callAsConstructor = TiWorker_construct;`

iphone/Classes/KrollContext.m
```diff
@@ -1189,6 +1294,11 @@ - (void)main
   [self bindCallback:@"L" callback:&LCallback];
   [self bindCallback:@"alert" callback:&AlertCallback];
 
+  // create Worker global, instantiated with "new Worker('path/to/file.js')
+  TiStringRef node = TiStringCreateWithUTF8CString("Worker");
+  TiObjectSetProperty(context, globalRef, node, TiObjectMakeConstructor(context, TiWorker_class(context), TiWorker_construct), kTiPropertyAttributeNone, NULL);
```
sgtcoolguy: I'm not entirely sure this is correct. First, I think `TiObjectMakeConstructor` has no mapping/analog in raw JSC headers, so this is likely what's causing the iOS compile to fail.

Basically, in Windows we hung the type as an object in the global namespace, so that'd be more like:
```cpp
TiObjectSetProperty(context, globalRef, node, TiObjectMake(context, TiWorker_class(context), NULL), kTiPropertyAttributeNone, NULL);
```

Basically what that's saying is make a JS object that wraps the class definition we have defined in `TiWorker_class`, and hang it as `"Worker"` in the global namespace. (Under the hood `TiWorker_initialize` should get called when the object is created as this type).

Then when you do "new Worker()" it should invoke whatever the class definition pointed at for it's `callAsConstructor` property, which should be set to `TiWorker_construct` in this example - as per my comment above.

sgtcoolguy:
So, from the docs for web workers, there's clearly some limitations on what they can do: https://developer.mozilla.org/en-US/docs/Web/API/Worker

 They can't access the full APIs. And there's a note that the message can be "any" Javascript Object. Looks liek there's more details here: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Transferring_data_to_and_from_workers_further_details

Do we plan to limit the APIs/scoping of our workers? Will they have full access to all our APIs? Right now it looks like we're spinning up a whole new KrollBridge which acts as separate instance of the JS engine? Do we need that, or can we spin just some part (like a new KrollContext)? Given our move to main thread, will the two+ engines both content for the main thread, or will they run in parallel?

It looks like we're just using JSON serialization/de-serialization to act as our means of "transferring" the data back and forth? That seems to fit expectations since it deals with copies, not shared instances and presumably we can't share state.

iphone/Classes/KrollContext.m
hansemannn: Should this go here as well? It's referenced in the `TiObjectMakeConstructor` call right now.

hansemannn:
Thanks for the review Chris! We only can provide serializable data, which will be detected and handled if not valid. 

The Mozilla docs have been the major document I followed for parity, but there are also (outdated?) docs claiming to use "addEventListener" which used to be a thing back in the days I guess (note: we should move to `on`, `off` and `once` as well, I think we prototyped something in the past).

Regarding the scope of the worker file (e.g. echo.js): It's an own KrollBridge as noted, so we cannot / should not allow passing objects between those scopes to avoid bad references and retain issues, although that may be fixable. What we can do is instantiate Ti-namespaces objects and work with them, e.g. a `Ti.Network.HTTPClient` instance. But I really need to see how sending a message that includes a `TiBlob`, `TiBuffer` or `TiStream` behaves (I was too afraid so far). 

One major issue I still see (regarding the obvious need of refactoring) is to expose the `onmessage` callback on the worker file, so that it can be both assigned in the file (e.g. echo.js) and invoked by the worker (inside the app.js). This should be another JSCore myth to crack, let me know if you have ideas regarding that. In detail (dreaming of ES6):

```js
onmessage = (e) => {
  postMessage({ hello: 'world' });
};
```
This would put both the `onmessage` callback and the `postMessage` method into the global scope of that special file. Oh, and how about multiple workers? The current architecture plans with only one worker at the time, so we'd need session-ids internally and manage those internally.

I may remove that 7.1.0 milestone again... 😁

hansemannn:
@sgtcoolguy Update the PR, moved the JSCore logic to a `TiWorkerJS` utility class and exposed `postMessage` to the global scope inside the worker (echo.js). The big issue remaining is how to hook the "onmessage" callback into the global space so I can invoke it instead of enqueueing an event like before.

sgtcoolguy:
Looking at this: https://blog.sessionstack.com/how-javascript-works-the-building-blocks-of-web-workers-5-cases-when-you-should-use-them-a547c0757f6a

I think we may need to consider explicitly limiting the APIs available to the web workers to non-UI related APIs, and explicitly spin off the KrollBridge in non-main thread.

Basically, the whole point of web workers is to perform non-UI related work async and then post back the result, so that the main/UI thread isn't choked causing UI performance issues doing the non-UI related work.

sgtcoolguy:
relevant: nodejs/node#20876

hansemannn:
@sgtcoolguy This PR now uses JSCore entirely and targets `next`.

hansemannn:
I assume this cannot be completed so let's close this.

PR #9771, opened 2018-01-26
[TIMOB-25619] Android: Allow non-file TiBlob to be previewed

garymathews:
- Allow non-file `TiBlob` to be previewed

###### TEST CASE
```JS
var win = Ti.UI.createWindow({ backgroundColor: 'green' });

win.addEventListener('postlayout', function () {
    win.toImage(function (blob) {
        if (!blob) return;

        win.backgroundColor = 'red';

        Ti.Media.previewImage({
            image: blob,
            success: function (e) {
                console.log('success, displaying preview');
            },
            error: function (e) {
                console.error('failed to display preview');
            },
        });
    });
});

win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25619)

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
@@ -1214,8 +1214,8 @@ public class MediaModule extends KrollModule implements Handler.Callback
 		if (options.containsKey(TiC.EVENT_ERROR)) {
 			errorCallback = (KrollFunction) options.get(TiC.EVENT_ERROR);
 		}
-		if (options.containsKey("image")) {
-			image = (TiBlob) options.get("image");
+		if (options.containsKey(TiC.PROPERTY_IMAGE)) {
+			image = (TiBlob) options.get(TiC.PROPERTY_IMAGE);
 		}
 
 		if (image == null) {
@@ -1223,8 +1223,6 @@ public class MediaModule extends KrollModule implements Handler.Callback
 				errorCallback.callAsync(getKrollObject(), createErrorResponse(UNKNOWN_ERROR, "Missing image property"));
 			}
 		}
-
-		TiBaseFile f = (TiBaseFile) image.getData();
 
 		final KrollFunction fSuccessCallback = successCallback;
 		final KrollFunction fErrorCallback = errorCallback;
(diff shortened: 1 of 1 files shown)


jquick-axway (approved):
CR: Pass

ypbnv (approved):
CR: Pass

android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
```diff
@@ -1230,7 +1228,14 @@ public void previewImage(KrollDict options)
 		intent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
 		TiIntentWrapper previewIntent = new TiIntentWrapper(intent);
 		String mimeType = image.getMimeType();
-		Uri imageUri = TiFileProvider.createUriFrom(f.getNativeFile());
+		Uri imageUri = null;
+
+		if (image.getNativePath() == null) {
```
sgtcoolguy: may want to consider refactoring this handling with the equivalent code in EmailDialogProxy. Basically in both instances you're trying to get a TiBlob converted to a File.

tests/Resources/ti.media.test.js
```diff
@@ -38,6 +38,37 @@ describe('Titanium.Media', function () {
 		should(Ti.Media.previewImage).be.a.Function;
 	});
 
+	it.android('previewImage of non-file blob', function () {
```
sgtcoolguy: my bad, I moved this test into the suite so there's a conflict now. You can move this to a ti.media.addontest.js or copy the ti/media.test front he suite and add this (or uncomment the variant of this I wrote using takeScreenshot which also exposed the issue).

android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
jquick-axway: For everyone's info, the way EmailDialogProxy is saving blobs to a temp file needs to be refactored too. It always saves blobs to the same filename. So, the last blob saved gets overwritten by the next blob saved to file.
https://jira.appcelerator.org/browse/TIMOB-25568

I recommend that we refactor our handling when addressing [TIMOB-25568](https://jira.appcelerator.org/browse/TIMOB-25568).

ssjsamir (approved):
**FR Passed** Was able to see a a preview of non-file TiBlob 

**Test Steps:** 
+ Created a titanium project with the build form this PR
+ Added the above code in to the project 
+ Ran the project
+ Was able to view the blob without any runtime errors (which were seen on older versions of the SDK
+ Also saw the following in the console

`[INFO] :   success, displaying preview`

**Test Environment** 
APPC CLI: 7.0.1
Device: Nexus 6p (8.1.0)
Node.js Version: 8.9.1

garymathews:
`no tests` as `previewImage()` requires user input to dismiss

PR #9773, opened 2018-01-26
[TIMOB-25727] iOS/Android: Fix Base64 behavior and accepted parameter types

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25727


iphone/Classes/UtilsModule.m
```diff
@@ -58,7 +58,17 @@ - (TiBlob *)base64decode:(id)args
...
 
-  NSString *str = [[self convertToString:args] stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
+  NSMutableString *str = [[NSMutableString alloc] initWithString:[[self convertToString:args]
+                                                                     stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]]];
+  // Fill padding for backwards compatibility
+  while ([str length] % 4 != 0) {
+    [str appendString:@"="];
+  }
+
+  if ([str rangeOfString:@"="].length > 2) {
```
sgtcoolguy: Wouldn't you want to check the string before adding padding?

iphone/Classes/UtilsModule.m
```diff
@@ -58,7 +58,17 @@ - (TiBlob *)base64decode:(id)args
 {
   ENSURE_SINGLE_ARG(args, NSObject);
 
-  NSString *str = [[self convertToString:args] stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
+  NSMutableString *str = [[NSMutableString alloc] initWithString:[[self convertToString:args]
+                                                                     stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]]];
+  // Fill padding for backwards compatibility
+  while ([str length] % 4 != 0) {
```
sgtcoolguy: I don't think this is right. Base64 encoded strings will only have one or two '=' characters as padding, which suggest modulo 3, not 4.

https://stackoverflow.com/questions/4080988/why-does-base64-encoding-require-padding-if-the-input-length-is-not-divisible-by

sgtcoolguy (changes requested):
You already have a manual example in the JIRA ticket, please add it to our Base64 unit tests.

Should be pretty easy to add a few tests for improperly padded encoded strings and verify we handle them and decode them as expected.

sgtcoolguy:
Some possible unit test examples:

Decoded | Properly padded | Improperly Padded
------------ | ------------- | ------------
f | Zg== | Zg
fo | Zm8= | Zm8
foo | Zm9v| *N/A*
foob | Zm9vYg== | Zm9vYg
fooba | Zm9vYmE= | Zm9vYmE
foobar | Zm9vYmFy | *N/A*

iphone/Classes/UtilsModule.m
hansemannn: The RFC sounds helpful:

```
(1) the final quantum of encoding input is an integral multiple of 24
   bits; here, the final unit of encoded output will be an integral
   multiple of 4 characters with no "=" padding,

   (2) the final quantum of encoding input is exactly 8 bits; here, the
   final unit of encoded output will be two characters followed by two
   "=" padding characters, or

   (3) the final quantum of encoding input is exactly 16 bits; here, the
   final unit of encoded output will be three characters followed by one
   "=" padding character.
```

hansemannn:
So it's working..but one of them ('foo' <==> 'Zmb9v') fails for me.

```js
    backgroundColor: '#fff'
});

    title: 'Trigger'
});

  // Should be a JSON-payload
  if (!Ti.Utils.base64decode('eyJzdWIiOiJTc2VsdmFyYWoiLCJlbWFpbCI6IlNzZWx2YXJhaiIsImF1ZCI6ImxhYmFsZXJ0IiwianRpIjoibG9Mczh3ajlpbFNTS1ZyRXFvOFowUyIsImlzcyI6Imh0dHBzOlwvXC9zc28uY2VsZ2VuZS5jb20iLCJpYXQiOjE1MTYxNzY0OTcsImV4cCI6MTUxNjE3Njc5NywicGkuc3JpIjoiTEk0ZjFvNUNqam81ejhQdVJOcWNreU95WVE4In0')) {
    alert('ERROR (1)!');
  }

  // More tests!
  if (Ti.Utils.base64decode('Zg').text !== 'f') {
    alert('ERROR (2)!');
  }
  if (Ti.Utils.base64decode('Zm8').text !== 'fo') {
    alert('ERROR (3)!');
  }
  if (Ti.Utils.base64decode('Zm9v').text != 'foo') {
    alert('ERROR (4)');
  }
  if (Ti.Utils.base64decode('Zm9vYg').text !== 'foob') {
    alert('ERROR (5)!');
  }
  if (Ti.Utils.base64decode('Zm9vYmE').text !== 'fooba') {
    alert('ERROR (6)!');
  }
  if (Ti.Utils.base64decode('Zm9vYmFy').text !== 'foobar') {
    alert('ERROR (7)!');
  }
});

win.add(btn);
win.open();
```

It obviously returns "foo", but debugging it returns a length of 4 with a "NULL" unicode char at the end. I have no idea where that comes from, but it blocks.

hansemannn:
@sgtcoolguy Fixed the iOS padding for all cases, verified the unit-tests and fixed a few more of them (all iOS and Android ones). iOS now accepts the "TiFile" parameter as well and Android does for the decoding as well (only worked for encoding on Android). Test-case is also listed on JIRA.

apidoc/Titanium/Utils/Utils.yml
```diff
@@ -10,18 +10,19 @@ platforms: [android, iphone, ipad]
...
         summary: Data to decode.
-        type: [String, Titanium.Blob]
+        type: [String, Titanium.Blob, Titanium.Filesystem.File]
 
   - name: base64encode
     summary: Returns the specified data encoded to Base64.
     description: |
-        Starting from Titanium 3.3.0, `obj` can be a [File](Titanium.Filesystem.File) object
-        on Android.
+        The parameter can be a String, Ti.Blob or Ti.File (iOS and Android only).
```
sgtcoolguy: May still want to retain note that Ti.Filesystem/File is supported on Android 3.3+, iOS 7.0.3+ (or whatever version this PR and cherry-picks gets into).

tests/Resources/ti.utils.test.js
```diff
@@ -0,0 +1,135 @@
...
+	  should(Ti.Utils.base64decode('Zg').text).eql('f');
+	  should(Ti.Utils.base64decode('Zm8').text).eql('fo');
+	  should(Ti.Utils.base64decode('Zm9v').text).eql('foo');
+	  should(Ti.Utils.base64decode('Zm9vYg').text).eql('foob');
+	  should(Ti.Utils.base64decode('Zm9vYmE').text).eql('fooba');
+	  should(Ti.Utils.base64decode('Zm9vYmFy').text).eql('foobar');
+	});
+
+	// FIXME Windows gives: 'base64decode: attempt to decode a value not in base64 char set'
+	it.windowsBroken('#base64decode(Ti.Blob)', function () {
```
sgtcoolguy: Add tests for `#base64decode(Ti.Filesystem.File)`

tests/Resources/ti.utils.test.js
```diff
@@ -0,0 +1,135 @@
...
+	it('#base64encode(String)', function () {
+		var test;
+		should(Ti.Utils.base64encode).be.a.Function;
+		test = Ti.Utils.base64encode('test');
+		should(test).be.a.Object;
+		should(test.apiName).eql('Ti.Blob');
+		should(test.getText()).be.eql('dGVzdA==');
+	});
+
+	it('#base64encode(Ti.Blob)', function () {
```
sgtcoolguy: Add test for `#base64encode(Ti.Filesystem.File)`

sgtcoolguy (approved):
Looks good. I'd add tests for base64encode/decode with a Ti.Filesystem.File argument as well for completeness (we already have Ti.Blob test variants, so just pass the file in rather than convert the file to a Ti.Blob first)

tests/Resources/ti.utils.test.js
hansemannn: Already in https://github.com/appcelerator/titanium_mobile/pull/9773/files#diff-6384f93cac05d6aea1061d799b0fe88aR67. Other test is added.

ewieberappc:
FR Passed on 7_0_X branch

PR #9778, opened 2018-01-28
[TIMOB-25418] Android CLI: Fix doc-generation for non-markdown files

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25418

There were two issues here:

1) The docs have been generated twice. One time, the result was stored in `this.documentation` which wasn't even used later. The second time, it's actually used for packaging.
2) Meta files like images that can be referenced in the documentation's markdown files would not be copied but instead hang up the build process by trying to read it's string contents.


android/cli/commands/_buildModule.js
```diff
@@ -1907,10 +1886,10 @@ AndroidModuleBuilder.prototype.packageZip = function (next) {
 							return walk(file, path.join(parent, name));
 						}
 
-						let contents = fs.readFileSync(file).toString();
+						let contents = fs.readFileSync(file);
```
cb1kenobi: I don't understand why you removed the `.toString()`. `dest.append()` will accept a string.

On a side note, I've started moving away from `fs.readFileSync(file).toString()` and towards `fs.readFileSync(file, 'utf8')`.

hansemannn: For non-markdown files (like images), we do not want to generate HTML from it and therefore only need the string-representation for valid markdown files. So in this case, we cannot use `readFileSync(file, 'utf8')` either as we still need its raw buffer for other files than markdown.

cb1kenobi: That's a pretty darn good reason!

cb1kenobi (approved):
**APPROVED**

PR #9783, opened 2018-01-30
[TIMOB-24610] Follow-on - support disabling transpile via tiapp.xml tag <transpile>false</transpile>. Move transpilation to node-titanium-sdk.

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24610

**Description:**
Now that node-titanium-sdk supports transpiling (via appcelerator/node-titanium-sdk#20), we can rely on that.

This also allows developers to explicitly turn off transpilation by adding a `<transpile>false</transpile>` tag into your tiapp.xml.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -4,7 +4,7 @@
  * @module cli/_build
  *
  * @copyright
- * Copyright (c) 2009-2017 by Appcelerator, Inc. All Rights Reserved.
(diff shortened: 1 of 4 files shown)


sgtcoolguy:
cc @janvennemann Here's the PR to support turning off transpilation

sgtcoolguy:
Marking a work in progress in combination with local testing with appcelerator/hyperloop.next#270 to verify that the solution here works with hyperloop and ES6 code.

PR #9788, opened 2018-02-01
[7_0_X][TIMOB-25733] Android: Fixed bug where view with border may fail to appear when sized larger than screen.

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25733

**Cherry Pick of:** #9787


garymathews:
Reverted https://github.com/appcelerator/titanium_mobile/pull/9752

PR #9790, opened 2018-02-01
[TIDOC-3108] Move a few examples to ES6 and cleanup format

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-3108

- Move `, function(e)` to `, (e) =>`
- Move some of the "var" variables to "let" / "const"
- Reformat common code-sequences, e.g. for-loops and parameter-separation and
- Began to replace mixed strings/variables to string templates, but there were countless more so I stopped for now.

May be merged in the 7.1.0 timeframe but I'd love to get some feedback before.


sgtcoolguy (approved):
The example changes generally look good to me.

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -545,15 +545,15 @@ examples:
     example: |
         The following code excerpt does a simple GET request and logs the response text.
 
-             var url = "http://www.appcelerator.com";
-             var client = Ti.Network.createHTTPClient({
+             const url = "http://www.appcelerator.com";
+             const client = Ti.Network.createHTTPClient({
                  // function called when the response data is available
-                 onload : function(e) {
+                 onload : (e) => {
```
ewanharris: Using an arrow func here causes babel to convert `this.responseText` to `undefined.responseText` as we're parsing the code as module and not script

sgtcoolguy: Oh snap, good catch. Actually, arrow functions don't have a `this`. So any callback functions that refer to `this.whatever` can't become arrow functions (I'm assuming mostly HTTPClient examples). If you have a reference to the client, you could call `client.responseText`, though - which I think is probably better form, though impossible to do if you're trying to refer to the client from inside a callback function passed into the constructor.

hansemannn: I think its in the event as well. Should be good to track, thanks guys!

hansemannn:
@sgtcoolguy Should this be taken for 7.1.0 or 7.2.0? I don't think QE will have much to test here.

hansemannn:
@sgtcoolguy @ewanharris Can we get this merged?

ewanharris:
I'm on the fence about this, I think having code examples that Studio will flag as warnings is not the developer experience we want.

Topener:
I agree with @ewanharris on this. I think Studio needs to be able to handle ES6 first

hansemannn:
Okay, so lets move it to 8.0.0 or whenever transpiling is handled properly / made default.

apidoc/Titanium/Android/BroadcastReceiver.yml
```diff
@@ -63,10 +63,10 @@ examples:
     example: |
         This example creates a broadcast receiver and registers it to handle the airplane mode broadcast.
 
-            var bc = Ti.Android.createBroadcastReceiver({
-                onReceived: function() {
+            const broadcastReceiver = Ti.Android.createBroadcastReceiver({
+                onReceived: () => {
```
sgtcoolguy: Technically you could shrink this down to a one-liner:
```javascript
onReceived: () => Ti.API.info('Handling broadcast.')
```

apidoc/Titanium/Android/Calendar/Calendar.yml
```diff
@@ -233,104 +233,102 @@ examples:
         have been selected in the native Android calendar application.
         
             function showCalendars(calendars) {
-                for (var i = 0; i < calendars.length; i++) {
-                    Ti.API.info(calendars[i].name);
-                }
+              calendars.forEach(calendar => {
```
sgtcoolguy: Another one-liner candidate:
```javascript
calendars.forEach(c => Ti.API.info(c.name));
```

hansemannn:
@sgtcoolguy I have added a few more adjustments. Are we planning to enable transpiling by default in 8.0.0?

hansemannn:
@sgtcoolguy I just rebased and relinted all sources here. We should consider merging this into next if we plan to do enable ES6+ by default in SDK 8.

hansemannn:
It looks like there is no interest in this change. Closing …

PR #9797, opened 2018-02-03
[TIMOB-25747] iOS: Fix error dialog from not showing up when using main thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25747

`no tests` because it's a UI related change.


iphone/Classes/TiErrorController.m
```diff
@@ -16,7 +16,19 @@ @implementation TiErrorController
 - (id)initWithError:(NSString *)error_
 {
   if (self = [super init]) {
-    error = [error_ retain];
+    error = [NSString stringWithFormat:@"⚠️ %@", [error_ retain]];
```
vijaysingh-axway: error  is autoreleased object and in dealloc it is released again, it will cause crash. No need of retaining error_ object. It should be -
 error = [[NSString alloc] initWithFormat:@"⚠️ %@", error_];

iphone/Classes/TiErrorController.m
```diff
@@ -16,7 +16,19 @@ @implementation TiErrorController
...
+  return self;
+}
+
+- (id)initWithError:(NSString *)error_ reason:(NSString *)reason_
+{
+  if (self = [super init]) {
+    error = [NSString stringWithFormat:@"⚠️ %@", [error_ retain]];
+
+    if (reason_ != nil) {
+      reason = [NSString stringWithFormat:@"%@", [reason_ retain]];
```
vijaysingh-axway: Same as above.

vijaysingh-axway (changes requested):

iphone/Classes/TiErrorController.m
hansemannn: Copy paste error :-) Fixed!

iphone/Classes/TiErrorController.m
```diff
@@ -30,6 +42,7 @@ - (void)dealloc
   RELEASE_TO_NIL(messageLabel);
   RELEASE_TO_NIL(disclosureLabel);
   RELEASE_TO_NIL(error);
+  RELEASE_TO_NIL(reason);
```
vijaysingh-axway: error and reason objects are autoreleased objects. These are released again here. It'll cause crash.

vijaysingh-axway (approved):
CR passed.

PR #9803, opened 2018-02-06
[TIMOB-24349] iOS: Fix both AlertDialog and OptionDialog to fire events properly

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24349

`no tests` because it's a UI related change.


vijaysingh-axway (approved):
CR passed.

ewieberappc:
FR Passed. All alert/click events were fired and heard correctly, including those in the background. Tested using the provided samples, modified versions to check additional variants, and the alert dialog suite.

PR #9804, opened 2018-02-06
[TIMOB-25302]: iPhone X Add new home-indicator related API's

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25302


iphone/Classes/TiWindowProxy.m
```diff
@@ -400,6 +400,37 @@ - (BOOL)isModal
   return isModal;
 }
 
+#if IS_XCODE_9
+- (NSNumber *)homeIndicatorAutoHidden
+{
+  if ([TiUtils isIOS11OrGreater] == NO) {
+    NSLog(@"[WARN] This property is available on iOS 11 and above.");
```
hansemannn: Should be `ERROR`, so developers properly wrap it out.

hansemannn (changes requested):
Very good PR! Just one note regarding the log level.

hansemannn (approved):

sgtcoolguy (changes requested):
@vijaysingh-axway Please add a unit test for this new property.
Add a file under `tests/Resources/ti.ui.window.addontest.js`

With a test like:
```javascript
/*
 * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
 */
/* eslint-env mocha */
/* global Ti */
'use strict';
var should = require('./utilities/assertions'),
	utilities = require('./utilities/utilities');

describe('Titanium.UI.Window', function () {
	var win,
		didFocus = false,
		rootWindow;

	this.timeout(5000);

	// Create and open a root window for the rest of the below child window tests to use as a parent.
	// We're not going to close this window until the end of this test suite.
	// Note: Android needs this so that closing the last window won't back us out of the app.
	before(function (finish) {
		rootWindow = Ti.UI.createWindow();
		rootWindow.addEventListener('open', function () {
			finish();
		});
		rootWindow.open();
	});

	after(function (finish) {
		rootWindow.addEventListener('close', function () {
			finish();
		});
		rootWindow.close();
	});

	beforeEach(function () {
		didFocus = false;
	});

	afterEach(function () {
		if (win) {
			win.close();
		}
		win = null;
	});

	it('homeIndicatorAutoHidden', function () {
		win = Ti.UI.createWindow({
			title: 'this is some text'
		});
		should(win.homeIndicatorAutoHidden).be.a.Boolean;
		should(win.homeIndicatorAutoHidden).should.be.false(); // defaults to false
		win.homeIndicatorAutoHidden = true;
		should(win.homeIndicatorAutoHidden).should.be.true();
	});
});
```

You may need to tweak the test, I don't know the iOS version that the default emulator runs on the CI machines, so maybe you can set the value to true if it's an old iOS version.

However, I can already see that on older iOS versions it looks like the value will be `-1` and not a boolean, which the docs don't describe.

tests/Resources/ti.ui.window.addontest.js
```diff
@@ -0,0 +1,71 @@
...
+			title: 'this is some text'
+		});
+
+		win.addEventListener('open', function () {
+			try {
+				should(win.homeIndicatorAutoHidden).be.a.Boolean;
+				should(win.homeIndicatorAutoHidden).be.false;
+				win.setHomeIndicatorAutoHidden(true);
+				should(win.homeIndicatorAutoHidden).be.true;
+				finish();
```
hansemannn: Do we have guard for this? If so, line 60 should still return `false` because we do not support this pre 11.0.

vijaysingh-axway: On CI machine default simulator is iPhone 7s having iOS 11.2. So there is no need of guard here.
In code if this property is accessed with iOS < 11, we are  logging error and returning -1 (because we have to return something) -

+- (NSNumber *)homeIndicatorAutoHidden
 +{
 +  if ([TiUtils isIOS11OrGreater] == NO) {
 +    NSLog(@"[ERROR] This property is available on iOS 11 and above.");
 +    return NUMINT(-1);
 +  }
 +  return NUMBOOL([self homeIndicatorAutoHide]);
 +}

PR #9807, opened 2018-02-06
[TIMOB-23247] CLI: Implement Native Module clean

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23247

**Description:**
```
 ti clean -p <platform>
```

This hooks up ti clean as a command that can be used for modules as we do with normal apps.
It's implemented in a single-platform at a time way, and isn't meant to yet support multi-platform modules like say hyperloop.

I took a best guess at the files we should be wiping out, but certainly open to suggestions/tweaks here.


iphone/cli/commands/_cleanModule.js
```diff
@@ -0,0 +1,51 @@
...
+		.map(function (s) {
+			return s.substring(0, 1).toUpperCase() + s.substring(1);
+		}).join('');
+	const moduleAssetsFile = path.join('Classes', moduleIdAsIdentifier + 'ModuleAssets.m');
+
+	// Guess the generated zipfile name!
+	// TODO: Use a glob and delete any version zipfile generated?
+	const moduleVersion = cli.manifest.version;
+	const moduleZipName = [ moduleId, '-iphone-', moduleVersion, '.zip' ].join('');
+	const toDelete = [ 'build', moduleAssetsFile, moduleZipName ];
```
hansemannn: `metadata.json`

android/cli/commands/_cleanModule.js
```diff
@@ -0,0 +1,37 @@
...
+const fs = require('fs-extra');
+const appc = require('node-appc');
+const __ = appc.i18n(__dirname).__;
+
+// TODO Do we need a validate function?
+
+exports.run = function run(logger, config, cli, finished) {
+	const projectDir = cli.argv['project-dir'];
+
+	const toDelete = [ 'build', 'dist', 'libs', 'java-sources.txt' ];
```
hansemannn: I think the build command already handles the recreation, but "build" has to exist and "libs" has to include the used architectures, e.g `arm64-v8a`. Also, [many](https://github.com/benbahrenburg/benCoding.AlarmManager/tree/master/dist) devs curate a list of archives in the `dist` directory that will be flushed out now

sgtcoolguy: Android goes a little nutty if `libs` already exists, sometimes. I think it tries to do an architecture sanity check, but I'm not sure why because that's an output directory for the static libraries it generates during he build.

I tested cleaning and building a couple native modules, and it handles recreating the `build` dir.

As far as `dist` goes, I get that some devs keep the generated zips there and check them into SCM, but that's not exactly a great practice. The zips should be published to S3 or Github releases or something, not checked into source.

hansemannn: I'd still rather not want to break this in a minor version and don't see the advantage of deleting them before, as they do not cause any issues right now, different to the other cleaned directories. Maybe move that change to 8.0.0 instead.

hansemannn (changes requested):
Some outstanding review changes.

android/cli/commands/_cleanModule.js
sgtcoolguy: I struggle to see how this is a breaking change. If a developer has a not-so-great practice of checking in generated zips to their SCM and the new module clean command will delete that directory, the file will still be in their SCM. This command will not "checkin" the deletion of the zip. And even if the user did check the deletion into SCM, the file would still be in their SCM history to restore.

The `dist` folder is generated output folder that should be cleaned before a clean build.

hansemannn: I only see the following JIRA ticket incoming:
> "I create super duper nice module packages since 6 years and now I lost all my clients because the module zips are gone. Why do you do that?" 
If we can defend that, let's do it.

hansemannn (approved):

hansemannn:
Don‘t we need QE testing here?

PR #9812, opened 2018-02-09
[TIMOB-25766] (7_0_X) Android: Revert Android related base64 changes

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25766

### DIFF
--- a/android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
+++ b/android/modules/utils/src/java/ti/modules/titanium/utils/UtilsModule.java
@@ -40,13 +40,6 @@ public class UtilsModule extends KrollModule
 			return (String) obj;
 		} else if (obj instanceof TiBlob) {
 			return ((TiBlob) obj).getText();
-		} else if (obj instanceof TiFileProxy) {
-			try {
-				return ((TiFileProxy) obj).read().getText();
-			} catch (IOException e) {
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
The revert and skipped test look right to me.

I'd strongly prefer we actually added a test case for this exact scenario so we can avoid the regression in the future. Also, Ideally we'd not just revert but fix the issue in a way that avoids this regression and supports the test case we're skipping again.

ewanharris:
@sgtcoolguy I have one good to go for Android, I just need to test iOS and Windows to ensure I don't break any existing builds, I'll make a PR to mocha-suite Monday

hansemannn:
@sgtcoolguy The issue was including opening the camera / photo gallery which cannot be handled via unit-tests. Maybe a good case for our appium-suite (cc @SamJHaig)

sgtcoolguy:
You don't need to use the camera/photo gallery to test it out. I'm digging into this now to write up some tests.

hansemannn:
We could also use a large image and hange its encoding to see the mismatch. Okay

jquick-axway:
The issue was caused by the blob's byte array being wrongly transcoded to UTF-16 (which changed the binary) before being base64 encoded. So, a simple test would be to create a binary blob from a stream/buffer and then verify that it base64 encodes it correctly. Doing it this way will make the returned result predictable.

PR #9814, opened 2018-02-09
(7_0_X) [TIMOB-25766] Android: Ti.Utils.base64encode for cannot encode (Image) Ti.Blob anymore. Result = null

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25766

**Description:**
Follow-on to #9812 

### DIFF
--- a/iphone/Classes/UtilsModule.m
+++ b/iphone/Classes/UtilsModule.m
@@ -90,20 +90,32 @@
 - (id)sha1:(id)args
 {
   ENSURE_SINGLE_ARG(args, NSObject);
-  NSString *nstr = [self convertToString:args];
-  const char *cStr = [nstr UTF8String];
+
+  NSData *data;
+  if ([args isKindOfClass:[TiBlob class]]) {
+    data = [(TiBlob *)args data];
+  } else {
+    data = [[self convertToString:args] dataUsingEncoding:NSUTF8StringEncoding];
+  }
+
   unsigned char result[CC_SHA1_DIGEST_LENGTH];
-  CC_SHA1(cStr, (CC_LONG)[nstr lengthOfBytesUsingEncoding:NSUTF8StringEncoding], result);
+  CC_SHA1(data.bytes, data.length, result);
   return [TiUtils convertToHex:(unsigned char *)&result length:CC_SHA1_DIGEST_LENGTH];
 }
 
 - (id)sha256:(id)args
 {
   ENSURE_SINGLE_ARG(args, NSObject);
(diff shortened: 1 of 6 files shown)


sgtcoolguy:
Fixing this up made the STREAM_BASE64 Ti.Blob type unused.

I can however see how we may want to improve Ti.Blob to be lazier when wrapping files (or making use of a Ti.Blob that actually wrapped streams rather than byte[] in many cases) - and the sub-type could come back.

A quick example: if you try to base64 encode a file, what we really do is wrap the Ti.Filesystem.File in a Ti.Blob, then call `toBase64()` on it, which effectively reads the full file into a `byte[]` and then encodes it. Not good at all. Android has Base64 input/output streams for just this use case, and a Ti.Blob that wrapped a stream lazily would be able to handle that without loading everything into memory (twice).
Not to mention we make use of Ti.Blob to wrap `byte[]` read from things like HTTPClient (though I guess we send Ti.Blob chunks via ondatastream callbacks which is capped at 4096 bytes by our read loop, and the full response Data is written to a temp file that we eventually wrap in a Ti.Blob)

Even ignoring actual stream use cases, base64 encode/decode may be more memory-friendly if we used Ti.Blobs that wrapped Base64InputStream and Base64OutputStream to stream out the result.

iphone/Classes/UtilsModule.m
```diff
@@ -90,20 +90,32 @@ - (NSString *)md5HexDigest:(id)args
...
   ENSURE_SINGLE_ARG(args, NSObject);
-  NSString *nstr = [self convertToString:args];
-  const char *cStr = [nstr UTF8String];
+
+  NSData *data;
+  if ([args isKindOfClass:[TiBlob class]]) {
+    data = [(TiBlob *)args data];
+  } else {
+    data = [[self convertToString:args] dataUsingEncoding:NSUTF8StringEncoding];
+  }
```
hansemannn: Move to `convertToData:` ?
```objc
- (NSData *)convertToData:(id)arg
{
  if ([args isKindOfClass:[TiBlob class]]) {
    return [(TiBlob *)args data];
  } else {
    return [[self convertToString:args] dataUsingEncoding:NSUTF8StringEncoding];
  }
}
```
And maybe even handle Ti.File objects here as well, similar to what we do in `convertToString:`.

hansemannn:
@sgtcoolguy Are we doing a 7.0.3? Just asking because 7_0_X now contains more commits than documented

PR #9815, opened 2018-02-09
[TIMOB-9366] Android: Added "radial" gradient support

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-9366

**Notes:**
- This feature is intended as an alternative for devs who are drawing circles/rings via a view's `borderRadius` property, which has known issues on Google's end.
- Supports all gradient properties except `endPoint`, which on iOS is used to smear the circular gradient from the start to end point. Android does not support this.
- Updated API docs to indicate Android's support and to better clarify the offset range for linear and radial gradients.
- The unit test written for PR #9819 covers gradient testing for both Android and iOS.

**Test:**
1. Build and run "GradientRadialTest.js" attached to [TIMOB-9366](https://jira.appcelerator.org/browse/TIMOB-9366) on Android.
2. On startup, verify that you see a circle where the outer edge is blue and the center is red.
3. Drag the "Start Radius" slider to the middle.
4. Verify that the circle becomes a ring. _(ie: The center is transparent)_
5. Tap the "Back Fill End" switch to enable it.
6. Verify that the area outside of the circle is now blue.
7. Tap the "Back Fill Start" switch to enable it.
8. Verify that the center of the circle is now solid red. _(ie: No longer transparent.)_
9. Tap the "Back Fill End" switch to disable it.
10. Drag the "End Radius" slider to the left, lower than the start radius position. _(ie: You want the end radius to be less than the start radius.)_
11. Verify that the inner part of the circle is now blue with a transparent center and the part outside of the circle is solid red.
12. Touch-drag the circle with your finger and verify that the center of the circle moves with your finger.

PR #9819, opened 2018-02-13
[TIMOB-25775] iOS: Fix gradient getters

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25775

Eventually to be merged before #9815


hansemannn:
@sgtcoolguy We now lint the unit-tests as well? Thats cool. But do we have a CLI command to fix it as well?

jquick-axway:
**Question:**
Do you need to do the same for `backfillStart` and `backfillEnd`?

hansemannn:
@jquick-axway You mean the getters? There are supposed to work already, but the PR is still work in progress to investigate that.

hansemannn:
@jquick-axway Updated the tests and cleaned up the commit mess. Your decision if 7.1.0 or not :-).

tests/Resources/ti.ui.view.test.js
```diff
@@ -0,0 +1,804 @@
...
+		});
+
+		gradient = {
+			type: 'radial',
+			startPoint: {
+				x: 50,
+				y: 50
+			},
+			colors: [ 'red', 'blue' ],
+			startRadius: '90%',
```
jquick-axway: Do we officially support percentage based radius settings? We currently document it as "Number" and not "Number/String".
http://docs.appcelerator.com/platform/latest/#!/api/Gradient-property-startRadius

I didn't implement support for this on Android because I assumed we didn't support this on iOS. It would be nice to have though, but the question is, what's it relative to? The width? The height? The shortest dimension? The longest dimension?

iphone/Classes/TiGradient.m
```diff
@@ -129,16 +133,29 @@ - (void)setEndPoint:(id)newEnd
...
+}
+
+- (NSNumber *)endRadius
+{
+  return [self valueForUndefinedKey:@"endRadius"];
+}
+
+- (NSNumber *)startRadius
+{
+  return [self valueForUndefinedKey:@"startRadius"];
```
jquick-axway: We don't document percentage based radius settings, but this will return it as a number between 0-100, right? I think it should return the original percentage string.

If we don't officially support percentage based radius, then I'm okay for making this a task for later.

tests/Resources/ti.ui.view.test.js
hansemannn: We do support it because iOS handles all units in a central class (TiDimension). So we could support it on other platforms as well, but later.

iphone/Classes/TiGradient.m
hansemannn: See comment above. But you are right, I need to change the return type to `id`.

tests/Resources/ti.ui.view.test.js
jquick-axway: Okay, from looking at our iOS code here...
https://github.com/hansemannn/titanium_mobile/blob/8261c1c17d11628c9fe4b0a9452905e378fb254b/iphone/Classes/TiGradient.m#L244

A percentage based radius is relative to the distance from the center of the view to one of its corners (ie: hypotenuse divided by 2). Good to know.

I'm trying to provide an option for devs to draw a circle ring without having to use our "borderRadius" feature. So, we might want to make what the percentage based radius is relative to settable in the future so that the circle won't get clipped.

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
FR Passed: Able to receive a log of the gradient properties.

**Test steps**
+ Created a build with the generated zip
+ Created a titanium app
+ Added the test case from https://jira.appcelerator.org/browse/TIMOB-25775
+ Ran the program
+ Was able to see the console log output gradient properties
`[INFO] :   red, blue
`

**Test Environment**
APPC Studio: 5.0.0.201712081732
APPC CLI: 7.0.3
iphone 6 plus (10.2)
Node.js Version: 8.9.1
Xcode 9.2

hansemannn:
@jquick-axway @ssjsamir I've fixed the unit-test linting manually now. Still wondering if there is a better way.

PR #9825, opened 2018-02-15
Add grunt task to auto-fix JS files

ewanharris:
Requires https://github.com/appcelerator-modules/grunt-appc-js/pull/11

Add a grunt task that automatically fixes JS files for people, ran with `grunt fixJs`, happy to bikeshed over the naming if anyone prefers anything else :) The task only runs the linting (no retire etc.), I was going to hang it off `grunt format` but it seemed pointless to lint all iOS/Android source as well.

Do we need a ticket for this?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -132,6 +132,8 @@ module.exports = function (grunt) {
 	// register tasks
 	grunt.registerTask('format', [ 'clangFormat' ]);
 
+	grunt.registerTask('fixJs', [ 'appcJs:src:lint:fix' ]);
+
 	// register tasks
 	grunt.registerTask('default', [ 'lint' ]);
 };


sgtcoolguy:
Looks like great minds think alike. I had recently pushed some changes to make it easier to format iOS or android source via grunt as well.

Personally, I think we should call it 'format:js' and add it to the 'format' task, and then probably add aliases to the ones I did for ios/android (under clangFormat) so that we could do:
```
npx grunt format         # format everything
npx grunt format:android # format Android Java source
npx grunt format:ios     # format iOS objc source
nix grunt format:js      # format js files
```

sgtcoolguy:
OK, just FYI after this merge I did converge on the task names in my comment above.

PR #9834, opened 2018-02-15
[7_1_X][TIMOB-17210] Android: Fixed TextField/TextArea to not trigger "change" event when setting properties

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17210

**Back-Port of:** PR #9700


longton95 (approved):
FR passed.

Tested using the test case the ticket and above, the event is only triggered when the text is changed by the user and not programmatically set.

It is now the same as iOS

ENV
```
SDK Version : 7.0.2.GA, Local 7.1.0
macOS Sierra 10.13.3
Google Pixel 2 XL (8.1.0)
android emulator (7.1.1)
Appc CLI : 7.0.2
Appc NPM : 4.2.12
Node : v8.9.1
```

PR #9770, opened 2018-01-25
[TIMOB-25703] Android: Added VideoPlayer "showsControls" property

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25703

**Summary:**
- This boolean property shows/hides the overlaid video player controls.
- This is for parity with iOS, which added this property in Titanium 7.0.0.
- Did not remove "mediaControlStyle" property (for backward compatibility reasons) which can show/hide controls as well via different style constants. If the new "showsControls" property is false, then it will always hide the controls, regardless what the media style is set to (takes priority).

**Test:**
1. Build and run the below on an Android device.
2. Wait for the video view to start playing. (Requires an Internet connection to stream video.)
3. Tap on the video.
4. Verify that the video controls are **not** shown when tapped.

```javascript
var window = Ti.UI.createWindow({ fullscreen: true });
var videoView = Ti.Media.createVideoPlayer(
{
	url: "http://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_640x360.m4v",
	autoplay: true,
	showsControls: false,
	scalingMode: Ti.Media.VIDEO_SCALING_ASPECT_FIT,
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
//videoView.showsControls = false;
window.add(videoView);
window.open();
```

PR #9852, opened 2018-02-16
[TIMOB-25566] iOS: Fix setting "showControls" on videoplayer when URL is changed after creation

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25566

**no tests**, because it's a UI related change.


jquick-axway:
**Note to everyone:**
I've already written a "showsControls" unit test in PR #9770, which is merged in. So, the unit test is already covered. 👍

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -134,6 +134,14 @@ - (void)configurePlayer
 {
   [self addNotificationObserver];
   [self setValuesForKeysWithDictionary:loadProperties];
+
+  // TIMOB-25566: Video controls need to be handled manually as they can
+  // cause issues if the URL is set after creation
+  id showControls = [self valueForKey:@"showControls"];
```
vijaysingh-axway: Property name is "showsControls". Please update.

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -469,7 +477,10 @@ - (NSNumber *)showsControls
 
 - (void)setShowsControls:(NSNumber *)value
 {
-  [movie setShowsPlaybackControls:[TiUtils boolValue:value def:YES]];
+  [self replaceValue:value forKey:@"showControls" notification:NO];
```
vijaysingh-axway: Same as above. Property name is wrong.

vijaysingh-axway (changes requested):

vijaysingh-axway (changes requested):
Still issue is there. Controls are not hiding.

hansemannn:
@vijaysingh-axway Not sure why I worked before, but I adjusted it to store it in the `loadProperties` which is even more clean :-).

vijaysingh-axway:
@hansemannn  Changes looks good. I guess you can add unit test for this. This is more to setting property showsControls than UI.

hansemannn:
@vijaysingh-axway Added

vijaysingh-axway (approved):
CR passed.

PR #9854, opened 2018-02-16
[TIMOB-19154] (7_1_X) iOS9 Deprecate NSURLConnection and replace with NSURLSession API

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-19154


ewieberappc:
FR Passed. Tested using all of the sample and referenced tickets in the issue ticket as well as the module examples and http test suite

PR #9865, opened 2018-02-21
[TIMOB-25795] Android: Expose all missing Android.R values to be used in Hyperloop

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25795

`no tests` so far, because we do not have Android resources for the new API's in the test suite so far. I'll add those later this week.


android/titanium/src/java/org/appcelerator/titanium/proxy/RProxy.java
```diff
@@ -21,8 +21,11 @@
 {
 	private static final String TAG = "TiAndroidRProxy";
 
-	private static final String[] RESOURCE_TYPES = { "anim", "array",   "attr",   "color",  "dimen", "drawable",
-													 "id",   "integer", "layout", "string", "style", "styleable" };
+	private static final String[] RESOURCE_TYPES = { "anim",    "animator",     "array",    "attr",     "bool",
+													 "color",   "dimen",        "drawable", "fraction", "id",
+													 "integer", "interpolator", "layout",   "menu",     "mimap",
+													 "plurals", "raw",          "string",   "style",    "styleable",
+													 "xml" };
```
jquick-axway: I think "mimap" is a typo. Should be "mipmap", right?

"mipmap" is intended to be used to store app/launcher icons and is intended to replace the "drawable" folder for this purpose.

hansemannn: Good catch! Thats a typo, will resolve after vacation.

sgtcoolguy:
@hansemannn ping

jquick-axway (approved):
CR: Pass

Only thing we're missing is "res/font" access, but we can punt on this for now if you want. It's an Android 8.0 feature that supposedly Support Library v26 provides backward compatibility back to API Level 16 (our current min supported).

hansemannn:
Thanks @jquick-axway! Moved to QE-testing.

PR #9882, opened 2018-02-24
[TIMOB-25755] Android: Allow WebView to support certificate requests

garymathews:
- Allow `Titanium.UI.WebView` to support certificate requests
  - NOTE: Android 4.4 does not support this and will fallback to previous behaviour

###### TEST CASE
- See JIRA ticket

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25755)


ypbnv (approved):
CR: **Pass**

longton95:
@garymathews Is there any other way I can test this without the customer's cert?

PR #9885, opened 2018-02-26
[TIMOB-25766][TIMOB-25727] Android: Ti.Utils.base64encode fixes

sgtcoolguy:
**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-25766
- https://jira.appcelerator.org/browse/TIMOB-25727

**Description:**

This is a fix for both listed tickets. Our previous "fix" was to revert the changes to avoid a regression caused by the fix to the other ticket. But then we never re-opened the original ticket when it was reverted.

This fix attempts to add more comprehensive Ti.Utils tests and fix both cases properly.

* Remove reference to unused TiBlob type
* Add tests for hashing/encoding binary data. Update Utils module to support that and clean up the code.
* Fix iOS to support sha256/sha1 of Ti.Blob with binary data.
* Refactor code for turning args into NSData for Utils


mukherjee2 (approved):
Passed CR and FR. Details are in backported version of this PR.

PR #9886, opened 2018-02-26
(7_1_X) [TIMOB-25766][TIMOB-25727] Android: Ti.Utils.base64encode fixes

sgtcoolguy:
**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-25766
- https://jira.appcelerator.org/browse/TIMOB-25727

**Description:**
back port of #9885

This is a fix for both listed tickets. Our previous "fix" was to revert the changes to avoid a regression caused by the fix to the other ticket. But then we never re-opened the original ticket when it was reverted.

This fix attempts to add more comprehensive Ti.Utils tests and fix both cases properly.

* Remove reference to unused TiBlob type
* Add tests for hashing/encoding binary data. Update Utils module to support that and clean up the code.
* Fix iOS to support sha256/sha1 of Ti.Blob with binary data.
* Refactor code for turning args into NSData for Utils

### DIFF
--- a/android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
+++ b/android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
@@ -895,14 +895,7 @@ public class TiHTTPClient
 				String mimeType = blob.getMimeType();
 				File tmpFile =
 					File.createTempFile("tixhr", "." + TiMimeTypeHelper.getFileExtensionFromMimeType(mimeType, "txt"));
-				if (blob.getType() == TiBlob.TYPE_STREAM_BASE64) {
-					FileOutputStream fos = new FileOutputStream(tmpFile);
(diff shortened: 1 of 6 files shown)


hansemannn (changes requested):
Please lint the Android source

*EDIT*: Looks okay now.

hansemannn (approved):

mukherjee2 (approved):
Node Version: 8.9.1
NPM Version: 5.5.1
Mac OS: 10.13.1
Appc CLI: 7.0.2
Appc CLI NPM: 4.2.12
Titanium SDK version: 7.1.0.v20180226081834
Appcelerator Studio vers 5.0.0
Android 7.1

I tested the PR against the test cases in both JIRA tickets mentioned in the PR. No errors were seen, and expected outputs were observed.

PR #9888, opened 2018-02-27
Try fixing build

ewanharris:


ewanharris:
All the failures don't actually run in a PR, @sgtcoolguy I think this might work as `npm ci` will throw away node_modules before it runs install, it's probably safer to use `npm ci` here than `npm prune --production`

PR #9889, opened 2018-02-27
[TIMOB-25862] Parity: Support async Ti.UI.Webview.evalJS call on Android & iOS

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25862

**Description:**
- Add an async variant of Ti.UI.WebView.evalJS, add unit test for it. This matches Windows API.
This is on both iOS and Android.


sgtcoolguy:
@garymathews @hansemannn @jquick-axway
So if we're unable to find a sync version of this Api to use when run-on-main-thread is on, I think this means the async variant should be available cross-platform and should be the preferred variant - *and that the sync variant with no callback needs to be deprecated*.

android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -110,9 +119,94 @@ public Object evalJS(String code)
...
+			}
+			return null;
+		}
+
+		if (KrollRuntime.getInstance().getKrollApplication().runOnMainThread()) {
+			Log.w(
+				TAG,
+				"Synchronous evalJS is not available when running on the main thread. Please supply an additional callback function argument to be invoked with the result as it's only parameter, to be called when the result is available.");
+			// FIXME This just times out on main thread! I don't know how else to get this to work...
+			// EvalJSSyncRunnable future = new EvalJSSyncRunnable(view, code);
```
sgtcoolguy: This code hung just as if I didn't spin off a thread, probably because I have to block below on the get call and the Javascript APIs have to use the UI thread to complete. In other words, I don't think we **can** run a sync version on main/ui thread.

tests/Resources/ti.ui.webview.addontest.js
```diff
@@ -0,0 +1,72 @@
...
+
+		webview = Ti.UI.createWebView();
+
+		webview.addEventListener('load', function () {
+			if (hadError) {
+				return;
+			}
+
+			// FIXME: Android is dumb and assumes no trailing semicolon!
+			webview.evalJS('Ti.API.info("Hello, World!");"WebView.evalJS.TEST"', function (result) {
```
sgtcoolguy: The Android implementation of evalJS leaves much to be desired. For one thing, it takes your input and appends:
```js
+"";
```

So if you have a trailing semicolon (as you should!) it doesn't behave properly. Also, I think it may be double-escaping strings here? I'd expect the result to be ```'WebView.evalJS.TEST'``` not ```'"WebView.evalJS.TEST"'```

jquick-axway: Welcome to Android. This is par for the course. Nothing is ever easy.

android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -110,9 +119,94 @@ public Object evalJS(String code)
 			Log.w(TAG, "WebView not available, returning null for evalJS result.");
 			return null;
 		}
+		if (callback != null) {
+			// When on Android 19+ we can use the builtin evalAsync method!
+			if (Build.VERSION.SDK_INT >= 19) {
+				final KrollFunction theCallback = callback;
+				view.getWebView().evaluateJavascript(code, new ValueCallback<String>() {
```
jquick-axway: Google documents that `evaluateJavascript()` must be called on the main UI thread here...
[https://developer.android.com/reference/android/webkit/WebView.html#evaluateJavascript](https://developer.android.com/reference/android/webkit/WebView.html#evaluateJavascript(java.lang.String,%20android.webkit.ValueCallback<java.lang.String>))

I looked at Google's source code on github. It'll throw an exception if invoked on any other thread.
https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/webkit/WebView.java#L1053

PR #9897, opened 2018-03-01
[TIMOB-25826] Android: Error requiring a file by relative path where module.exports is a falsey value

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25826

Return the full module object to the require function to avoid the `if (loaded)` check failing due to falsey values. Pushing tests first to ensure that they fail on Jenkins (everything looks good locally, i.e they fail)


tests/Resources/fixtures/empty-double.js
```diff
@@ -0,0 +1 @@
+module.exports = "";
```
sgtcoolguy: All JS files should have a newline at the end. Not sure what editor you use, but for example Atom enforces this on save for me.

tests/Resources/ti.require.test.js
```diff
@@ -0,0 +1,277 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe.only('requireJS', function () {
```
sgtcoolguy: remove this cheat once you're ready for merge 🔍

tests/Resources/ti.require.test.js
```diff
@@ -0,0 +1,277 @@
+/*
```
sgtcoolguy: Why not use an addontest.js file?

ewanharris: Gah, my bad completely forgot that 🤦‍♂️

tests/Resources/fixtures/empty-double.js
ewanharris: Hmm, will double check my settings in vscode

tests/Resources/ti.require.test.js
```diff
@@ -0,0 +1,277 @@
...
+		];
+		for (var obj in files) {
+			var result;
+			should.doesNotThrow(
+				// eslint-disable-next-line no-loop-func
+				function () {
+					result = require('./fixtures/' + obj.filename);
+				}
+			);
+			should(result).be.eql(obj.expected);
```
sgtcoolguy: why are we always expecting 0 as the result? I assume because of the `eql` method being equivalent to `==`?

I think we should likely use `should(result).be.exactly(obj.expected);` and compare to '', false, null, undefined, 0;

For NaN, we can't compare vis `==` or `===`, but should use `isNaN(result).should.be.true();`

ewanharris:
@sgtcoolguy Fix the <sup>many</sup> issues :) Tests are actually correct now

sgtcoolguy (approved):

PR #9903, opened 2018-03-05
[TIMOB-9366] update radial backgroundGradient example to match implementation

Topener:
JIRA: https://jira.appcelerator.org/browse/TIMOB-9366

The current example in the documentation works only for iOS. I updated it to reflect both platforms

### DIFF
--- a/apidoc/Titanium/UI/View.yml
+++ b/apidoc/Titanium/UI/View.yml
@@ -1119,7 +1119,7 @@ properties:
                         startPoint: { x: 50, y: 50 },
                         endPoint: { x: 50, y: 50 },
                         colors: [ 'red', 'blue'],
-                        startRadius: 90,
+                        startRadius: 50,
                         endRadius: 0,
                         backfillStart: true
                     }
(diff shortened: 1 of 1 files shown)


apidoc/Titanium/UI/View.yml
```diff
@@ -1119,7 +1119,7 @@ properties:
                         startPoint: { x: 50, y: 50 },
                         endPoint: { x: 50, y: 50 },
                         colors: [ 'red', 'blue'],
-                        startRadius: '90%',
+                        startRadius: 90,
```
jquick-axway: We should set startRadius to `50` instead.
This way, a full circle will just-fit within the view.

Note that setting this to `90` would make the radius larger than the old `90%` setting. On iOS, `100%` equals `sqrt(width^2, height^2) / 2`, which is the distance between the center of the view to one of its corners.

jquick-axway (changes requested):

Topener:
@jquick-axway good point! Updated, will update the other PR too

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

Example works properly on both Android & IOS.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

PR #9908, opened 2018-03-06
[TIMOB-25836] : iOS Using "run-on-main-thread", multiple global event  listeners do not work

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25836


hansemannn:
Please add unit tests for this. I think the example from JIRA will geko already, together with a finish()

iphone/Classes/AppModule.m
```diff
@@ -188,7 +188,7 @@ - (void)fireEvent:(NSArray *)args
 
     DebugLog(@"[DEBUG] Firing app event: %@", type);
 
-    NSArray *array = [appListeners objectForKey:type];
+    NSArray *array = [[appListeners objectForKey:type] copy];
```
hansemannn: I wonder if we should rather synchronize the `appListeners` setters, so nothing can be mutated during its write-process.

hansemannn: Ok then, taking your word!

hansemannn (approved):
CR **approved**.

PR #9909, opened 2018-03-06
[TIMOB-13379] Implement console.time/timeEnd

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-13379

Implement console.time/timeEnd according to https://console.spec.whatwg.org/#time, with the optional log on `console.time('duplicate label')` implemented, and a slight deviation to log a warning when calling `console.timeEnd('does not exist')`.

iOS can do a higher precision on the timing due to the implementation being in native but I nerfed for parity, happy to un-nerf though.

#### Test code

```js
console.time('set label'); // Set custom labels
console.time('set label'); // Should log "Label set label already exists" at a warn lebel
setTimeout(() => {
    console.timeEnd('set label'); // Should log "set label: 100ms" (not exactly 100ms)
}, 100);

console.time(); // Defaults to "default" if no value passed in
setTimeout(() => {
    console.timeEnd(); // Should log "set label: 5000ms" (not exactly 5000ms)
}, 5000);
console.timeEnd('No exist') // Should log "Label No exist does not exist" at warn level
```

#### Output

```txt
iOS
[WARN]  Label set label already exists
[WARN]  Label No exist does not exist
[INFO]  set label: 106ms
[INFO]  default: 5003ms

Android
[WARN]  Label set label already exists
[WARN]  Label No exist does not exist
[INFO]  set label: 127ms
[INFO]  default: 5004ms
```


android/runtime/common/src/js/console.js
```diff
@@ -44,3 +45,27 @@ exports.error = function () {
 exports.debug = function () {
 	Titanium.API.debug(join(arguments));
 };
+
+exports.time = function (label) {
+	if (times[label]) {
+		exports.warn('Label ' + label + ' already exists');
```
ewanharris: imo, this would read nicer using template literals. Which are supported by the v8 version I believe. What're our best practices for the JS code on Android. The eslint options specify ES5

```json
  "parserOptions": {
    "ecmaVersion": 5,
    "sourceType": "script"
  }
```

iphone/Classes/TiConsole.m
```diff
@@ -14,4 +14,37 @@ - (void)log:(NSArray *)args
   [self logMessage:args severity:@"info"];
 }
 
+- (void)time:(NSArray *)args
+{
+  if (!_times) {
+    _times = [[NSMutableDictionary alloc] init];
+  }
+  NSString *label = [args componentsJoinedByString:@""] ?: @"default";
```
hansemannn: Should a developer be able to pass an array of argumenzts, e.g. `console.time(['test', 'foo'])`? If not, change the arg-type to `id` and use `ENSURE_SINGLE_ARG(args, NSString *);` to force a string directly.

android/runtime/common/src/js/console.js
garymathews: I'm fine with that since it's supported.

android/runtime/common/src/js/console.js
```diff
@@ -25,6 +25,7 @@ function join(args) {
 	}).join(' ');
 }
 
+var times = {};
```
garymathews: Could you move this to the top of `console.js` ?

apidoc/Global/console/console.yml
```diff
@@ -70,3 +70,26 @@ methods:
...
+  - name: time
+    summary: Start a timer to track duration of an operation.
+    description: |
+          Begin a timer that can be used to track the duration of an operation.
+          If not label is passed to the function it will default to "default",
+          if a label already exists then the existing label will not be overwritten
+          and a warning will be logged to the console.
+    parameters:
+      - name: label
+        summary: The label to track the timer by, defaults to "default".
```
sgtcoolguy: Isn't there an actual `default` property you can set here?

android/runtime/common/src/js/console.js
```diff
@@ -44,3 +45,28 @@ exports.error = function () {
...
+		label = 'default';
+	}
+	times[label] = Date.now();
+};
+
+exports.timeEnd = function (label) {
+	if (!label) {
+		label = 'default';
+	}
+	var startTime = times[label];
```
sgtcoolguy: vars should be declared at top of scope, since they get hoisted up there.

sgtcoolguy: Also, given that we know V8 is newer and can support ES6 natively, is there any reason we can't start making use of es6 syntax in runtime JS code?  (i.e. template strings here, use of const/let)

sgtcoolguy: e.g. https://github.com/nodejs/node/blob/master/lib/console.js#L166-L184

ewanharris: @sgtcoolguy  Updated to use template literals, const and a Map instead of a plain object

apidoc/Global/console/console.yml
ewanharris: Wasn't aware of that, updated to include default and optional in docs. Making a note to familiarise myself with our doc stuff too :)

android/runtime/common/src/js/console.js
ewanharris: FYI I filed https://jira.appcelerator.org/browse/TIMOB-25892 to hook the linting up in this folder as `grunt lint` doesn't include it currently

sgtcoolguy:
Windows PR: appcelerator/titanium_mobile_windows#1209

PR #9914, opened 2018-03-06
Fixed deprecated since values for Geolocation.yml

jawa9000:
At this time, the docgen script cannot handle objects passed as values in deprecated since. Please don't pass {<platform>: <version>} as a value for the since property.

See https://jira.appcelerator.org/browse/TIDOC-3124 for details.

### DIFF
--- a/apidoc/Titanium/Geolocation/Geolocation.yml
+++ b/apidoc/Titanium/Geolocation/Geolocation.yml
@@ -429,7 +429,7 @@ properties:
     permission: read-only
     platforms: [android, iphone, ipad]
     deprecated:
-        since: {android: "2.0.0"}
+        since: "2.0.0"
 
   - name: ACCURACY_HUNDRED_METERS
     summary: |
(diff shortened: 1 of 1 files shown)


hansemannn (changes requested):
Missing explanation about the Android specific deprecation, please address that. Also file against master and from your fork if possible.

jquick-axway:
The descriptions for these Android "deprecated" constants already state that they're not recommended to be used. Perhaps we should just remove the "deprecation" fields since they're not deprecated on the iOS side.

On Android, you can still use these constants. The issue is we don't have equivalent settings in Google's FusedLocation provider which match these constants. So, instead, we end up enabling high power/accuracy mode for short distance constants and low power/accuracy mode for long distance constants.

hansemannn:
@jquick-axway Thats right, but doing this change, we will in fact deprecate it for all platforms. Writing that it's not recommended doesn't mean we deprecate it IMHO. But feel free to merge if you are okay with it.

jquick-axway:
I think we should just remove the deprecation fields and call it a day.

hansemannn:
Agree

hansemannn:
Closing PR as it needs to be reopened against `master` first. Also, see the above discussion for details on the proposed changes.

PR #9223, opened 2017-07-17
[TIMOB-24707] Android: Show images and videos in picker

frankieandone:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24707

**Summary:**
Added `MEDIA_TYPE_VIDEO` support to `Ti.Media.openPhotoGallery()` on Android. Can now select both images and videos with this API, like iOS.
_(Does not support live photos. That is an iOS only feature.)_

**Test 1:**
Re-run test attached to PR #9918.

**Test 2:**
Re-run test attached to PR #9906.
Verify that selecting an image from Google Drive (via the cloud) does not cause a crash.

**Test 3:**
Re-run test attached to PR #9771.
Verify that the screenshot taken via `toImage()` can still be viewed via `previewImage()`.

**Test 4:**
_(You need to take photos and videos using the Android device's camera before starting the below test procedure.)_

1. Build and run the below code on an Android 7.0 or newer device.
2. An alert is displayed upon app launch. Tap the "Photo" button.
3. Select a photo from the gallery.
4. Verify the Titanium app display the selected photo.
5. Back out of the app and re-launch it.
6. An alert is displayed upon app launch. Tap the "Video" button.
7. Select a video from the gallery.
8. Verify the Titanium app plays the selected video.
9. Back out of the app and re-launch it.
10. An alert is displayed upon app launch. Tap the "Both" button.
11. Select a video from the gallery.
12. Verify the Titanium app plays the selected video.
13. Repeat steps 2-12 on an Android 4.1 device.

```javascript
var window = Ti.UI.createWindow();
window.addEventListener("open", function(e) {
	var dialog = Ti.UI.createAlertDialog(
	{
		message: "Which media type do you want to open?",
		buttonNames: ["Photo", "Video", "Both"],
	});
	dialog.addEventListener("click", function(e) {
		var mediaTypes;
		if (e.index === 0) {
			mediaTypes = [Ti.Media.MEDIA_TYPE_PHOTO];
		} else if (e.index === 1) {
			mediaTypes = [Ti.Media.MEDIA_TYPE_VIDEO];
		} else if (e.index === 2) {
			mediaTypes = [Ti.Media.MEDIA_TYPE_PHOTO, Ti.Media.MEDIA_TYPE_VIDEO];
		} else {
			Ti.API.info("@@@ Alert was canceled.");
			return;
		}
		Ti.Media.openPhotoGallery(
		{
			allowMultiple: false,
			allowEditing: false,
			autohide: true,
[...]

PR #9918, opened 2018-03-08
[TIMOB-25849] Android: Always specify image mime-type

garymathews:
- Always specify image mime-type to obtain image information correctly
- Remove annoying warning from `TiBlob`

##### TEST CASE
```JS
    btn = Ti.UI.createButton({
        title: 'OPEN GALLERY',
    });

btn.addEventListener('click', function(){
    Ti.Media.openPhotoGallery({
        mediaTypes: [ Titanium.Media.MEDIA_TYPE_PHOTO ],
        success: function (e) {
            var msg = 'media.width: ' + e.media.width
                + '\nmedia.height: ' + e.media.height
                + '\nmedia.length: ' + e.media.length
                + '\nmedia.mimeType: ' + e.media.mimeType
                + '\nmedia.nativePath: ' + e.media.nativePath;
            alert(msg);
        },
        error: function (e) {
            alert('error opening image: ' + e.error);
        }
    });
});

win.add(btn);
win.open();
```
- Should see `width height` and `mimeType` populated after selecting an image

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25849)

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
@@ -1113,6 +1113,7 @@ public class MediaModule extends KrollModule implements Handler.Callback
 	{
 		String[] parts = { path };
 		TiBlob imageData;
+
 		// Workaround for TIMOB-19910. Image is in the Google Photos cloud and not on device.
 		if (path.startsWith("content://com.google.android.apps.photos.contentprovider")) {
 			ParcelFileDescriptor parcelFileDescriptor;
@@ -1125,14 +1126,14 @@ public class MediaModule extends KrollModule implements Handler.Callback
 				parcelFileDescriptor.close();
 				imageData = TiBlob.blobFromImage(image);
 			} catch (FileNotFoundException e) {
-				imageData = createImageData(parts, MIME_IMAGE);
+				imageData = createImageData(parts, null);
 			} catch (IOException e) {
-				imageData = createImageData(parts, MIME_IMAGE);
+				imageData = createImageData(parts, null);
 			}
 		} else {
-			imageData = createImageData(parts, MIME_IMAGE);
+			imageData = createImageData(parts, null);
 		}
-		return createDictForImage(imageData, MIME_IMAGE);
+		return createDictForImage(imageData, null);
 	}
 
 	public static TiBlob createImageData(String[] parts, String mimeType)
--- a/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
@@ -180,6 +180,8 @@ public class MediaModule extends KrollModule implements Handler.Callback
 	private static String extension = ".jpg";
 	private TiTempFileHelper tempFileHelper;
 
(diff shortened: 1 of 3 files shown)


android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
```diff
@@ -1123,14 +1125,14 @@ protected static KrollDict createDictForImage(String path)
...
 			} catch (IOException e) {
-				imageData = createImageData(parts, null);
+				imageData = createImageData(parts, MIME_IMAGE);
 			}
 		} else {
-			imageData = createImageData(parts, null);
+			imageData = createImageData(parts, MIME_IMAGE);
 		}
-		return createDictForImage(imageData, null);
+		return createDictForImage(imageData, MIME_IMAGE);
```
jquick-axway: We shouldn't hard code the mime-type like this. This is because the end-user is capable of selecting files others than images from the gallery. I've done it on my devices. In this case, a zero width/height makes sense.

Have a look at how my open PR handles it here...
[https://github.com/appcelerator/titanium_mobile/pull/9223/files-diff](https://github.com/appcelerator/titanium_mobile/pull/9223/files#diff-32e2c2f2c365eee27a85d7bd7c08e1fcR1120)

Here is a code snipped from my PR. It'll fetch the mime-type from the gallery app's content provider using the given content URI. I've tested this with photos, videos, and text files and it works. And I think it's okay to leave it `null` if the selected file is unknown since that won't happen for photos.

```java
// Determine the mime type for the given file.
String mimeType = null;
boolean isPhoto = false;
try {
	mimeType = TiApplication.getInstance().getContentResolver().getType(Uri.parse(path));
	if ((mimeType != null) && mimeType.toLowerCase().startsWith("image")) {
		isPhoto = true;
	}
} catch (Exception ex) {
}
```

jquick-axway: Also note that the above will correctly set the mime-type for the selected image format such as "image/png", "image/jpeg", etc.

jquick-axway (changes requested):

garymathews:
@jquick-axway Updated PR, we now attempt to obtain the mime-type from the content resolver. Also note that the user should not be able to select anything other than images due to [this](https://github.com/appcelerator/titanium_mobile/pull/9918/files#diff-32e2c2f2c365eee27a85d7bd7c08e1fcR1007) filter; although that may not be enforced correctly on all devices.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

`Width` & `height` of the selected photo is successfully retrieved.
 If we try to select a video the behavior is not enforced on all devices as said by @garymathews above. On a Nexus 6P we are not allowed to select a video & on a Nexus 5 we can select a video but then its width & height is retrieved as zero.

OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.2
Daemon Ver:     1.0.1
Ti CLI Ver:     5.0.14
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_101

PR #9923, opened 2018-03-09
[TIMOB-16511] : Text in label gets truncated (...) when changing font Weight from normal to bold

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-16511


iphone/Classes/TiUILabel.m
```diff
@@ -448,6 +448,7 @@ - (void)setFont_:(id)font
     CGFloat ratio = minFontSize / label.font.pointSize;
     [label setMinimumScaleFactor:ratio];
   }
+  [self padLabel];
```
hansemannn: This is a pretty major change. Can we validate the following use-cases for this:
- Multi-line labels
- ellipsize labels
- minimum-font-sizes

hansemannn (approved):

iphone/Classes/TiUILabel.m
jquick-axway: Right, changing other label properties causes this issue too. Our label code probably needs to re-udpate/invalidate itself after a property change.
https://jira.appcelerator.org/browse/TIMOB-24864

PR #9926, opened 2018-03-09
[TIMOB-26340] Android: Update V8 runtime to 6.9.x - V8 6.9.427.17 and NDK r16b

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-26340

**Description**
This is a PR/branch to keep us up to date with the latest V8 stable branch. This will get updated over time as we approach 8.0 being target of master branch.

Here's how we can track stable branch: https://omahaproxy.appspot.com

So far the update required:
- Compiling with Android NDK r16b
- bumping the native module api version for android to `5` (we will need to rebuild the bundled native modules against a new SDK now)
- Changing from gyp to ninja/gn to build v8 (which also resulted in an easier to consume single static library)
- tweaks around the fallback sourceURL we pass to Java proxies. In most cases we get the url passed in a context object behind the scenes. But there are/were cases we grabbed the url from the function args `Callee()` method which is now removed (and was long deprecated). I can achieve the same thing using a stack trace hack, and I have done so. I also moved it to happen only when we *don't* get the url passed in. In these cases I'm not even sure we need to hack the url from the stack trace. Every instance I saw was a ti:/whatever.js internal runtime file or a "module.id/bootstrap.js" file. We may be able to keep null, or fall back to app://app.js. Or maybe we need to look up the stack higher?
- `HasOwnProperty()` and `GetRealNamedProperty()` calls now require a `Local<Context>` first argument.
- Some APIs Now return `MaybeLocal<T>` and need to be handled more carefully (i.e. actual try/catch or safe conversions to `Local<T>` as they have been doing over time to almost all APIs)
- Changing a reference to the legacy debug API on our APIModule.cpp to delegate to `JSDebugger::debugBreak()` instead to handle `Ti.API.debugBreak();`
- Move away from deprecated APIs that took an `Isolate` and now take a `Context` (typically coercion calls like `ToString()`, `ToObject`, `Int32Value()`, etc on `Local<Value>`).
- Move to APIs taking a first-argument `Context` now, such as `Get` and `Set` calls on `Local<Object>`.

TODO:
- [ ] Rebuild the pre-packaged android native modules against this SDK (there's a bit of a chicken and egg problem to do so. The builds here fail during unit testing until the modules are rebuilt, bu the modules need to pick up a built SDK to build against). This likely means I'll need to move/change this PR to be a long running feature branch on the original repo that actually get published. Maybe wait until we branch for 8.0 and we get it automatically...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,7 +59,7 @@ def unitTests(os, nodeVersion, npmVersion, testSuiteBranch) {
 					manager.addWarningBadge(msg)
 					throw e
 				}
-			}
+			} // dir
 			// copy over any overridden unit tests into this workspace
 			sh 'rm -rf tests'
 			unstash 'override-tests'
@@ -222,7 +222,7 @@ timestamps {
 					// TODO parallelize the iOS/Android/Mobileweb/Windows portions?
 					dir('build') {
 						timeout(15) {
-							sh "node scons.js build --android-ndk ${env.ANDROID_NDK_R12B} --android-sdk ${env.ANDROID_SDK}"
+							sh "node scons.js build --android-ndk ${env.ANDROID_NDK_R16B} --android-sdk ${env.ANDROID_SDK}"
 						} // timeout
 						ansiColor('xterm') {
 							timeout(15) {
--- a/android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBinding.fm
+++ b/android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/ProxyBinding.fm
@@ -506,26 +506,9 @@ fails, a JS exception is returned.
 	} else {
 	</#if>
 <#t>
-	<#local current_arg = "arg_" + index>
-	<#if info.jsType == "Value">
-		<#local type = "Local<Value>">
-		<#local valueExpr = expr>
-	<#elseif info.jsType == "Number">
-		<#local type = "Local<Number>">
-		<#local valueExpr = expr + "->ToNumber(isolate)">
-		<#local stringExpr = expr + "->ToString(isolate)">
-	<#elseif info?keys?seq_contains("jsHandleCast")>
-		<#local type = "Local<" + info.jsType + ">">
-		<#local valueExpr = expr + ".As<" + info.jsType + ">()">
-	<#else>
-		<#local type = "Local<" + info.jsType + ">">
-		<#local valueExpr = expr + "->To" + info.jsType + "(isolate)">
-	</#if>
-
-<#t>
 	<#if info.jsType == "Number">
-		if ((titanium::V8Util::isNaN(isolate, ${expr}) && !${expr}->IsUndefined()) || ${stringExpr}->Length() == 0) {
-			const char *error = "Invalid value, expected type ${info.jsType}.";
+		if ((titanium::V8Util::isNaN(isolate, ${expr}) && !${expr}->IsUndefined()) || ${expr}->ToString(context).FromMaybe(String::Empty(isolate))->Length() == 0) {
+			const char *error = "Invalid value, expected type Number.";
 			LOGE(TAG, error);
 			<#if !(logOnly!false)>
 			titanium::JSException::Error(isolate, error);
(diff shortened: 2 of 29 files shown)


garymathews:
@sgtcoolguy Isn't the latest version [6.5](https://omahaproxy.appspot.com/) ?

sgtcoolguy:
@garymathews Yes, I had started when 6.4 was stable and ran into the `args.Callee()` removal, so I punted for a few weeks to work on other things. Then I came back to fix it up and start a long-running PR to track v8 stable for our 8.0.0 timeline.

So for now, this is already outdated, but I'll be updating this and our V8 builds over time to keep up.

farfromrefug:
@sgtcoolguy I am testing this PR on my fork and it is working great!
Now it showed something i have actually been seeing for quite some time now: the v8 so lib has increased in size drastically since 5.0.0. From around 5mo to more than 12mo.
Do you see a reason for that?
I am really wondering about that because it  makes apps a lot bigger.

farfromrefug:
@sgtcoolguy also wouldn't it be a good time to split modules into separated so libs? That way you could include only necessary ones depending on Ti symbols founds in projects. That would make apps a lot smaller and faster to load.
I am willing to help in starting this. But i don't think i can do this alone :s 
Let me know

farfromrefug:
So i have been running a lot of tests in my fork. First your last V8 build is awesome. It runs flawlessly and seems even faster!

Now about the v8 so lib size. Today it's around 12Mo(armv7) (in my work goes closer to 13Mo).
At first i thought the size came from the Kroll part and the boostrap. However after running some simple tests this is not the case:
- removing all titanium modules/bootstrap => 11.1 Mo
- trying to improve build flags : no gain

What i dont understand is why the lib is so big with the latest versions of v8
Nativescript  android runtime build is actually a lot smaller. They have a normal and an optimized one (optimized uses visibility=hidden). The optimized one is 8Mo. It needs to be compared to the one i get by removing most Kroll things and bootstrap. The reason is that Nativescript does it all like Hyperloop. So it means a 3Mo difference so 25% difference. Seems too big of a difference to me

Any idea what difference might come from?

Also i should mention that Nativescript still uses v8 5.5.372.32. I need to roll back and see Titanium v8 lib file size with that version (but not that easy to rollback).

sgtcoolguy:
@farfromrefug Regarding v8 library size, I too noticed that it seems much larger now in recent v8 builds. I think that v8 itself has gotten much larger since the 4.x/5.x days. Just looking at the tarballs that we package and publish and consume in our SDK builds:
screen shot 2018-04-16 at 1 25 33 pm


Now some of that size increase is due to packaging ARM 64 libraries. Another large change is that v8 moved from gyp to build to ninja/gn. 
We keep our patches to build v8 here: https://github.com/appcelerator/v8_titanium

It's entirely possible that more flags could be used/tweaked to reduce size. I'm no c/c++ expert at all.

sgtcoolguy:
@farfromrefug The version should come from v8 itself. I think I haven't updated the master branch there yet to point at v8 master tip (which is now 6.8.0.0). The build used here is from the https://github.com/appcelerator/v8_titanium/tree/6.7-lkgr branch.

farfromrefug:
@sgtcoolguy So i worked on the kroll-v8 lib size.
I managed to get the lib to 8Mo instead of 12.4Mo
This is great as i am at the same size as Nativescript. But there is a catch.
It will only work good with app not using native modules or using hyperloop modules. The reason is that to get it smaller i had to hide functions symbols from the lib. This is an issue when you use native modules because their so lib will try to lookup symbols in the kroll-v8 lib and they can't find it.

So i am thinking of actually building 2 versions of the lib when packaging. Then when building an app, depending on the modules used i can choose which one to build with.
4Mo is quite a big difference!

You can see the flags that i use here:
https://github.com/Akylas/titanium_mobile/blob/master/android/runtime/v8/src/native/common.mk
The flags which would create the smallest lib is ```--exclude-libs=ALL``` (which is commented for now)

hansemannn:
@sgtcoolguy @farfromrefug V8 7.0 is out, we could try use that one for the next bump 🚀

PR #9932, opened 2018-03-13
[TIMOB-25850] : iOS SearchBar doesn't show Bookmark button when set on creation in TableView

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25850


hansemannn (changes requested):
Please loop through the proxy keys instead of checking all manually.

iphone/Classes/TiUISearchBarProxy.m
```diff
@@ -89,6 +89,15 @@ - (void)setSearchBar:(UISearchBar *)searchBar
 
   // In UISearchController searchbar is readonly. We have to replace that search bar with existing search bar of proxy.
   [(TiUISearchBar *)[self view] setSearchBar:searchBar];
+
+  // Set search bar properties to new search bar
+  NSDictionary *properties = [self allProperties];
+  for (NSString *key in properties.allKeys) {
```
hansemannn: I assume `[self allProperties]` (`dynprops`) cannot be `nil`, otherwise the app could crash here.

hansemannn (approved):

hansemannn:
@vijaysingh-axway Do we have tests for all properties of the search-bar already? If not, please add them. We should get too easy with setting the `no tests` label, otherwise it might get removed again.

vijaysingh-axway:
@hansemannn In TIUISearchBar, getter is not implemented properly. It simply returns the value which is set in proxy. Ideally it should return value by fetching from searchbar. Currently whatever you set the value of property, it will return same even though that value is not passed to searhbar. So I am thinking to create a ticket for that, in which we can add all these property in unit tests as well. Thats why I have not added unit test in this ticket. Lets discuss on this.

hansemannn:
Like it, please do so! Maybe for the 7.3.0 timeline then.

hansemannn:
Closing in favor of #9947

PR #9936, opened 2018-03-14
[FRAME-5] Generate TypeScript definitions from docs

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/FRAME-5

**Optional Description:**
This is the first version of the TypeScript definition generator. It generates a `index.d.ts` file which can be submitted to https://github.com/DefinitelyTyped/DefinitelyTyped. Currently this needs to be done manually but we could automate this in the future if we are sure the generated typings are stable enough.


hansemannn:
@janvennemann Can you fix the ESLint issues? I did locally but also removed the `hasAllProxyMembers` method which I don't want to commit without your permission. Welcome back!

apidoc/lib/typescript_generator.js
```diff
@@ -0,0 +1,542 @@
+'use strict';
+
+/*
+ * Map of invalid types and their replacement
+ */
+const invalidTypeMap = {
+	'2DMatrix': 'Matrix2D',
+	'3DMatrix': 'Matrix3D',
+	'Dictionary': 'any'
```
sgtcoolguy: in our parlance a "Dictionary" is an Object, I believe.

janvennemann: In TypeScript the `object` type only refers to the Object prototype and its properties and methods. Our `Dictionary` refers to any custom object or a specific set of properties/methods of another type, for which i haven't found a proper way of representing that in TypeScript yet. So i just replace it with `any` for now.

apidoc/lib/typescript_generator.js
```diff
@@ -0,0 +1,528 @@
+'use strict';
+
+/*
+ * Map of invalid types and their replacement
+ */
+const invalidTypeMap = {
+	'2DMatrix': 'Matrix2D',
```
hansemannn: We may want to rename the SDK API's for this, either to `Matrix2D` / `Matrix3D` or `Matrix` with `type: Ti.UI.MATRIX_TYPE_2D`. Thoughts?

janvennemann: +1 to rename them. @garymathews  encountered a similar issue when playing around with type constructors instead of factory functions, since any type name starting with a number is invalid.

hansemannn: Which option do you prefer? How do others name it? Any official specification?

janvennemann: From what i have seen, most other libraries only deal with one Matrix at a time, so its simply called Matrix as there is no need to distinguish between 2D and 3D matrices.  Since we need both I'd prefer to name them `Matrix2D` and `Matrix3D` to keep a type for each as they provide a different set of transformation functions you can apply to them.

janvennemann:
@jawa9000 would you mind taking a look at this as it's mostly doc related? Thanks!

PR #9937, opened 2018-03-15
[TIMOB-25756] Android: Restart application upon locale change

garymathews:
- Restart the application upon locale change to load new language

##### TEST CASE
- Project on JIRA
```JS
Ti.Locale.addEventListener('change', function (e) {
    // locale has changed, soft restart prevented
});
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25756)


Topener:
Yeah I agree that event makes sense! How would a user handle the restart of the app though? Or is opening/closing windows enough?

hansemannn:
On iOS, we could listen to the `NSCurrentLocaleDidChangeNotification` notification and fire a Ti.Locale `localechange` event. Restarting iOS apps is not allowed by Apply though and causes app-rejections if hacked through (private) API's.

hansemannn:
Is this a public method? `Ti.App._restart()`? The underscore looks really odd.

garymathews:
@hansemannn nope, it's intended for LiveView

hansemannn:
What do you think about the following btw?
```js
Ti.Locale.addEventListener('change', function (e) {
    // locale has changed
});
```

jquick-axway:
I like @hansemannn idea of a `Ti.Locale` "change" event.

And should we use this event to override the auto-restart behavior like how we did it with the "androidback" button event?

Personally, I wouldn't want the app to restart since you'll lose content and your current position within the app. Most apps, especially Windows desktop apps, will only handle the locale change the next time the app is re-launch by the end-user.

Topener:
I feel like you should be able to override it, so makes sense! As long as you can listen for it and still let it reboot :)

lokeshchdhry:
FR Passed.

1. The locale `change` event is fired if `change` event listener is defined & locale is changed. 
2. If `change` event listener is not defined the app will soft restart to the appropriate language provided `strings.xml` is defined for particular language.

SDK Ver:        local build
OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_161

lokeshchdhry:
@garymathews , Can you please resolve the conflicts.

PR #9939, opened 2018-03-16
(7_1_X) [TIMOB-25875] Babel rewrites app.js in classic project to generated ES5 code

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25875

**Description:**
This issue was not specific to app.js. The issue here was that if a user built a project and a file had no changes through our build pipeline we'd try to symlink to the original source file as a build performance enhancement. But if subsequent builds trigger changes on a file (like transpilation would) and we write the new contents to the destination file, and it's a symlink, that modifies the original source file!

The solution here is pretty simple, if the destination file exists, remove it first before writing to it.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -69,8 +69,8 @@ function AndroidBuilder() {
 	this.maxSupportedApiLevel = parseInt(version.parseMax(this.packageJson.vendorDependencies['android sdk']));
 
 	this.deployTypes = {
-		'emulator': 'development',
-		'device': 'test',
+		emulator: 'development',
+		device: 'test',
 		'dist-playstore': 'production'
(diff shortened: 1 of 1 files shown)


hansemannn (approved):
CR and FT passed!

lokeshchdhry:
FR passed.

OS Ver: 10.13.2
Xcode Ver: Xcode 9.3
Appc NPM: 4.2.12
Appc CLI: 7.0.3-master.24
Daemon Ver: 1.0.1
Ti CLI Ver: 5.1.0
Alloy Ver: 1.11.0
Node Ver: 8.9.1
NPM Ver: 5.5.1
Java Ver: 9.0.4

PR #9941, opened 2018-03-18
[TIMOB-25879] Package iOS module zip in "dist" directory (Parity)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25879


sgtcoolguy (approved):
CR looks fine. But this is a behavioral change and may affect Studio, cc @kolipakakondal @ppedduri 

Also, do we have any actual docs that may need updating if we change the output directory?

hansemannn:
We need to update the [iOS Module Project](https://docs.appcelerator.com/platform/latest/#!/guide/iOS_Module_Project-section-src-43288810_iOSModuleProject-Buildandpackagethemodule) guide that talks about the output directory. It will be updated once this one is merged to keep in sync. Not sure about the Studio impact, but I don't see issues that this change could cause there.

kolipakakondal:
Added studio necessary changes with this https://github.com/appcelerator/titanium_studio/pull/1067 

Based on the SDK 7.3.0 release plan - studio can include this change

PR #9942, opened 2018-03-18
[TIMOB-25874] Fix iOS 10+ refresh-control when using navigation-window and large titles

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25874


vijaysingh-axway (approved):
CR passed.

hansemannn:
@sgtcoolguy @jquick-axway This was the PR introducing the failing refresh-control tests. But in this PR, it was all green. How can that happen?

jquick-axway:
> This was the PR introducing the failing refresh-control tests. But in this PR, it was all green. How can that happen?

@hansemannn, @sgtcoolguy,

It slipped by because the `it.windowsAndAndroidMissing()` function did not exist, it threw an exception, but was caught by the "addonttest" try/catch block here...
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/app.js#L119

Because of that, all of the proceeding tests did not get executed in the addon test script. I've confirmed it by doing the below.
```javascript
it.badFunctionName('test1', function () {
});
it('test2', function () {
	should(false).be.eql(true);
});
```

So, I suppose the thing to do is to either get rid of the addon try/catch or have the catch block fail the test.

PR #9945, opened 2018-03-19
[TIMOB-25864] Android: Set default notification channel

garymathews:
- Set default notification channel to `NotificationChannel.DEFAULT_CHANNEL_ID` -> `"miscellaneous"`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25864)


jquick-axway (approved):
CR: Pass

jquick-axway:
**Note to QE:**
This needs to be tested on Android 8.x and an older OS version, such as Android 4.x.

jquick-axway:
@lokeshchdhry, did you change "tiapp.xml" to target API Level 26 or higher?

lokeshchdhry:
@jquick-axway , Yes I did set it to 26. 
Also setting it to 25 works fine.

jquick-axway:
After chatting with @lokeshchdhry, this PR fails to post local notifications on all Android OS versions when targeting API Level 26. It works on all Android OS versions when targeting API Level 25.

This is odd because we're using the same solution we used for push notifications. Lokesh is going to re-test push notifications from what's currently on "master" to see if it still works or not.

hansemannn:
@garymathews I've added https://github.com/appcelerator/titanium_mobile/pull/9945/commits/db6d91878ce41a049dd27da7265bbca7fd7c0f51 so Jenkins can build it.

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationManagerModule.java
```diff
@@ -106,7 +136,14 @@ public void cancelAll()
 	@Kroll.method
 	public void notify(int id, NotificationProxy notificationProxy)
 	{
-		getManager().notify(id, notificationProxy.buildNotification());
+		Notification notification = notificationProxy.buildNotification();
+
+		// targeting Android O or above? create default channel
+		if (notification.getChannelId() == DEFAULT_CHANNEL_ID) {
```
jquick-axway: The `notification.getChannelId()` method call is API Level 26 only. We need to do an API Level check or else it'll crash on older Android OS versions.

jquick-axway (changes requested):

jquick-axway (approved):
CR: Pass

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationManagerModule.java
```diff
@@ -36,6 +37,11 @@
 	protected static final int PENDING_INTENT_FOR_BROADCAST = 2;
 	protected static final int PENDING_INTENT_MAX_VALUE = PENDING_INTENT_FOR_SERVICE;
 
+	private static NotificationManager notificationManager = null;
+	private static NotificationChannel defaultChannel = null;
+
+	public static final String DEFAULT_CHANNEL_ID = "defaultChannel";
```
jquick-axway: Do you think maybe we should use a prefixing notation to help avoid name collision?
Something like: `"ti_default_channel"`
(Kind of like how it's done with resources.)

lokeshchdhry:
FR Passed.

Local notifications work fine on API 26 & above.

Studio Ver:     5.1.0.201805300954
OS Ver:         10.13.4
Xcode Ver:      Xcode 9.4
Appc NPM:       4.2.13
Appc CLI:       7.0.4-master.2
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.0
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10
                ⇨ google Nexus 6P --- Android 8.1.0

lokeshchdhry:
Waiting for suggested changes to be done to the PR.

PR #9947, opened 2018-03-20
[TIMOB-25866] : iOS TiUISearchBar getter is returning value from proxy not from searchbar view

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25866


hansemannn (approved):
CR looks fine! Please fix the 3 linting issues and it should be good to go for QE-testing.

PR #9652, opened 2017-12-04
[TIMOB-25580] Android: Prevent EVENT_RETURN firing twice

garymathews:
- Prevent `onEditorAction` firing `EVENT_RETURN`

###### TEST CASE
```Java
var win = Ti.UI.createWindow(),
    text = Ti.UI.createTextArea({
        value: 'TEXT',
        top: 10,
        left: 10,
        right: 10,
        height: 200
    });

text.addEventListener('return', function() {
    alert('return event fired');
});

win.add(text);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25580)

PR #9948, opened 2018-03-20
[TIMOB-25882] Android: return event is received twice from Symbol TC55/TC70 scanner.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25882

**NOTE:** This PR relies on https://github.com/appcelerator/titanium_mobile/pull/9933 to be merged first.

**Description:**
As far as I understood/researched and tested the scanner simulates a keyboard for the input and sends an enter key pressed event. `onActionEditor` is called with actionID equal to `TYPE_NULL` and once for every keyEvent action - ACTION_UP and ACTION_DOWN. If the first call is marked as not consumed (the method call returns false) the second one is not triggered. I am not certain why it is handled that way and most likely it is bubbled up to something that prevents the second call. This seems to be the case regardless of the IME_ACTION for the TextField. I was not able to think of a cleaner solution that would cover the cases.

I don't think should be labeled as a bug or improvement, because it has the same behavior in native Android as well. We are doing some of the filtering of the events under the hood.


jquick-axway:
I'm not sure if I understand.

Do you mean actual scanning software that reads texts via the mobile device's camera and outputs it to the app's TextField? And this issue only happens with Motorola Symbol phones? If so, then I would assume its simulating key events via software.

How were you able to test this?

ypbnv:
I was not able to test the exact scenario as the one described in the ticket's comment.

Apparently these devices have an embedded scanner that can be set up by third party software.
https://developer.zebra.com/thread/31628
I assumed that from the link above.
What I tested was sending keyevents through `adb` and by setting an emulator to take advantage of the hardware keyboard. By doing so I got the reported behavior in versions 6.3.0.GA through 7.1.0.GA. I combined this with the code we had in 6.3.0.GA (reported by the user to work for them)
https://github.com/appcelerator/titanium_mobile/blob/6_3_X/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java#L473
and the following changes which "broke" that. This is the closest I could get to the actual scenario, bearing in mind that I have no clue what the actual events thrown by the scanner are.

I don't like what I did in this PR and IMO this is something that is hard to be universally solved, because different configuration of such scanners' input may result in different events and their handling. A better approach for me would be to propagate whatever the Android triggers to the JS, but that will most likely cause disparity for the event and we don't want that.

jquick-axway:
Thanks for the explanation. It looks like PR #9652 is the one that changed this behavior. Can you do a quick test with that PR's code as well? If it passes, then yeah,  let's go with your change.

ypbnv:
@jquick-axway Tried the test case from https://github.com/appcelerator/titanium_mobile/pull/9652 with software kyeboard and bluetooth keyboard both on landscape and portrait modes - the event is fired only once as expected.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -519,15 +519,25 @@ public boolean onEditorAction(TextView v, int actionId, KeyEvent keyEvent)
 			return true;
 		}
 
-		//This is to prevent 'return' event from being fired twice when return key is hit. In other words, when return key is clicked,
-		//this callback is triggered twice (except for keys that are mapped to EditorInfo.IME_ACTION_NEXT or EditorInfo.IME_ACTION_DONE). The first check is to deal with those keys - filter out
-		//one of the two callbacks, and the next checks deal with 'Next' and 'Done' callbacks, respectively.
-		//Refer to TiUIText.handleReturnKeyType(int) for a list of return keys that are mapped to EditorInfo.IME_ACTION_NEXT and EditorInfo.IME_ACTION_DONE.
-		if (actionId == EditorInfo.IME_ACTION_NEXT || actionId == EditorInfo.IME_ACTION_DONE) {
+		if (tv.getMaxLines() == 1) {
```
jquick-axway: This should use the `this.field` property like in the other PR.
The rest looks fine.

jquick-axway (changes requested):

ypbnv:
@jquick-axway Updated.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

Studio Ver:     5.1.0
OS Ver:         10.13.2
Xcode Ver:      Xcode 9.3
Appc NPM:       4.2.12
Appc CLI:       7.0.3-master.27
Daemon Ver:     1.0.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       9.0.4

PR #9952, opened 2018-03-22
[TIMOB-24266] iOS: Fix case where multiple delegate arguments appear in random order

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24266


vijaysingh-axway (approved):
CR approved.

PR #9955, opened 2018-03-22
[TIMOB-24817] Android: Implement ListView Alloy widget support

garymathews:
###### REQUIRES https://github.com/appcelerator/alloy/pull/892
- Implement ability to use Alloy widgets in `Titanium.UI.ListView`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24817)


jquick-axway (approved):
CR: Pass

sgtcoolguy:
Holding to 7.4.0 to give iOS team time to do equivalent ticket: https://jira.appcelerator.org/browse/TIMOB-24776

cc @hansemannn

android/modules/ui/src/js/listview.js
```diff
@@ -80,13 +80,29 @@ exports.bootstrap = function(Titanium) {
...
+				  parentProxy = eval(parentNamespace);
+	
+			if (parentProxy) {
+				return parentProxy['create' + proxyName];
+			}
+	
+		// handle Alloy widgets
+		} else {
+			const widget = Module.main.require('/alloy/widgets/' + namespace + '/controllers/widget');
+			if (widget) {
```
hansemannn: We should handle plain CommonJS modules here as well? For iOS, we use a [fallback case](https://github.com/appcelerator/titanium_mobile/pull/10131/files#diff-e577c4c73d4af3fdf412c02fff71ed4fR2955) to do this. We require common-js modules to export a view via `getView` as well, which is a suitable requirement to have a formal way of knowing where the main view was exported.

hansemannn: @garymathews Ping! The iOS PR was just merged.

garymathews:
Updated PR.

android/modules/ui/src/js/listview.js
```diff
@@ -80,13 +80,32 @@ exports.bootstrap = function(Titanium) {
...
+	
+			if (parentProxy) {
+				return parentProxy['create' + proxyName];
+			}
+	
+		// handle Alloy widgets
+		} else {
+			let widget = Module.main.require('/alloy/widgets/' + namespace + '/controllers/widget');
+			if (!widget) {
+					Module.main.require(namespace);
```
hansemannn: I think the assignment is missing here? Also, this does not seem to work with native modules.

PR #9957, opened 2018-03-24
[TIMOB-25901] iOS: Refactor background-service handlers

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25901

`no tests` because mainly refactoring, but also dependent on background-services that require active user interaction.


vijaysingh-axway (approved):
CR approved.

PR #9960, opened 2018-03-26
Updated Media.yml (AC-5649)

jawa9000:
Added a note to the showcamera method for the Android platform.

See https://jira.appcelerator.org/browse/AC-5649 for details.


hansemannn (changes requested):
This is not correct. On iOS and Windows, you can include both. Please change the note to be Android specific

hansemannn (approved):
Please discard my previous comment, it was my mistake by not seeing the context! Looks goooooood.

PR #9961, opened 2018-03-26
Correct invalid property and class reference in docs

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-3159

Fix most of the warnings output by the docs generation. Most (probably all, I just didn't check each one), would result in a bad link or weird text where the property would just be missing from the sentence (see `Ti.Geolocation`)

I also included 37f1dfb541b32da19ce2f95ff2bc22daba47d0ff in this, I kept getting caught out by not running the docgen script from in the apidoc folder, so it would always generate in `../dist/` relative to my titanium_mobile repo. I'm not sure whether this might break any publication scripts @jawa9000 so happy to revert that if so.

I've tried to categorise as best I can, so here's what I fixed and the location where the problem was.

- [WARN] Cannot find class: Ti.UI.View (multiple, too many to count)
	- Under borderWidth property on things that extend Ti.UI.View
- [WARN] Cannot find class: Ti.Geolocation
    - In the text `API in order to receive the best usability and permission-control during the app-lifecycle using hasLocationPermissions and . ` on Geolocation on Ti.Geolocation
- [WARN] Cannot find API: Titanium.UI.iOS.SystemButtonStyle.BAR (multiple)
    - Looks to have been removed, so removed references in ButtonBar
- [WARN] Cannot find class: Ti.UI.Window
	- Under onBack property on Ti.UI.Window
- [WARN] Cannot find class: Ti.UI.iOS.Stepper (6 warnings)
    - Under wraps property on Ti.UI.iOS.Stepper
- [WARN] Cannot find class: Titanium.UI.Searchbar (2 warnings)
    - Incorrect casing of SearchBar in `Ti.UI.iOS.SEARCH_BAR_STYLE_PROMINENT` and `Ti.UI.iOS.SEARCH_BAR_STYLE_MINIMAL `
- [WARN] Cannot find class: Titanium.UI.AlertDialogStyle (4 warnings)
    - Under all propertys Ti.UI.iPhone.AlertDialogStyle, recommended `Ti.UI.AlertDialogStyle`, not `Ti.UI.iOS.AlertDialogStyle`
- [WARN] Cannot find class: Ti.WatchSession (12 warnings)
    - Properties refered `Ti.WatchSession` not `Titanium.WatchSession`
- [WARN] Cannot find API: Titanium.UI.ActivityIndicator.indicatorDiameter
    - Under style property of ActivityIndicator, was removed when MW was removed as MW only property
- [WARN] Cannot find API: Titanium.Geolocation.createLocationProvider
    - Under `This means the provider passed into createLocationProvider will
    be ignored, as Google Play Services` text

There are still some warnings, most are because I don't have module docs I think. One I need help with.

- [WARN] Cannot find class: Modules.PlayServices (5 warnings)
    - Probably because I don't have the repo cloned?
- [WARN] Cannot find API: Modules.SafariDialog (2 warnings)
    - Probably because I don't have the repo cloned?
- [WARN] Cannot find API: Modules.URLSession (8 warnings)
    - Probably because I don't have the repo cloned?
- [WARN] Cannot find API: Modules.Map (10 warnings)
    - Probably because I don't have the repo cloned?


jquick-axway:
@ewanharris, regarding the `createLocationProvider()` doc error, I believe it's supposed to be referencing the `Ti.Geolocation.Android.createLocationProvider` module method.
https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Geolocation.Android-method-createLocationProvider

apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -85,7 +85,7 @@ description: |
     "When in Use" to "Always", which is the recommended way for managing location permissions in iOS 11 and later.
     Please also remember to request your desired location-permissions before using any geolocation-related API in 
     order to receive the best usability and permission-control during the app-lifecycle using <Titanium.Geolocation.hasLocationPermissions>
-    and <Ti.Geolocation.requestLocationPermissions>. Also note that you also need to include the `NSLocationWhenInUseUsageDescription` key 
+    and <Titanium.Geolocation.requestLocationPermissions>. Also note that you also need to include the `NSLocationWhenInUseUsageDescription` key
```
hansemannn: This is a real pain when not knowing how the docs generator works. Can't we just allow both and move the general usage to "Ti"? Who uses "Titanium.UI.createWindow" ?

ewanharris: We should be able to do that. From my understanding of all this, you reference by the `name` property in the yaml file that defines that API. 

So, in theory we should be able to replace all of these with `Ti.X` and be done with it. But I imagine we'd probably run into a load of issues regarding, links going dead, search results being bad etc. A better option may just be to rewrite the text to `Ti.X` when we write the result.

ewanharris:
@jquick-axway Ha, of course, no idea how in the world I missed that. Thanks!

jawa9000:
Reviewed the changes but cannot update the branch as there are still two checks are that failing.

PR #9962, opened 2018-03-27
[TIMOB-25903] Android: TableView.updateRow method doesn't work on 7.1 if run-on-main-thread is off

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25903

**Description:**
Update model when the row updated from another thread.

**Test case:**
1. Open the "tiapp.xml" file.
2. Set XML property `run-on-main-thread` to `false`.
3. Run the below code on Android.
4. Tap on the "Update Row" button.
5. Verify that text `ROW #00` changes to `NEW ROW`.

---
```xml
   <property name="run-on-main-thread" type="bool">false</property>
</ti:app>
```

---
```javascript
var rowData = [];
height: 120,
title: 'ROW #00'
});
rowData.push(row);
 
var table = Ti.UI.createTableView({
top: 50,
height: 120,
data: rowData
});
table.addEventListener( 'click', function(e) {
console.log(' ***** click on table ');
});
win.add(table);
 
var bt = Ti.UI.createButton({
top: 200,
title: 'UPDATE ROW'
});
 
bt.addEventListener('click',function(e){
var newRow = Ti.UI.createTableViewRow({
height: 120,
title: 'NEW ROW'
});
table.updateRow(0,newRow);
});
win.add(bt);
 
win.open();
```

---
**Note:** I am not sure if our unit tests should be considering `run-on-ui-thread` property.  It will be significantly more time consuming to handle both set ups, but maybe it is worth thinking in that direction? What do you think?

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
@@ -886,6 +886,7 @@ public class TableViewProxy extends TiViewProxy
 		Object asyncResult = null;
 		switch (msg.what) {
 			case MSG_UPDATE_VIEW:
+				setModelDirtyIfNecessary();
 				result = (AsyncResult) msg.obj;
 				if (tableNativeViewCreated) {
 					tableNativeView.updateView();


ewanharris:
@ypbnv After seeing the ticket I thought the same, I think in a PR we should maybe only cover what the default is. But we should definitely cover both in a nightly/weekly/pre-release test run.

jquick-axway (approved):
CR: Pass

hansemannn:
run-on-main-thread is default since 7.0.0 (alrhough it can be turned off). In the future, it will be the only option. I think I talked with Josh about that recently, so we should probably do at least default testing with main thread enabled.

lokeshchdhry:
FR Passed.
`TableView.updateRow ` method works as expected when `run-on-main-thread` if `false`.

Studio Ver:     5.1.0
OS Ver:         10.13.2
Xcode Ver:      Xcode 9.2
Appc NPM:       4.2.12
Appc CLI:       7.0.3-master.23
Daemon Ver:     1.0.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.11.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_131

lokeshchdhry:
Removed needs test as it was removed in 7.1.1 PR as well.

PR #9966, opened 2018-03-28
[TIMOB-25906] : iOS Native exceptions not thrown when using run-on-main-thread

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25906


hansemannn (approved):
CR **approved** based on the previous findings and changes discussed on JIRA. Thanks @vijaysingh-axway!

PR #9968, opened 2018-03-28
[TIMOB-25912] iOS 11.2: Fix document-viewer URL directory

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25912


vijaysingh-axway (approved):
CR passed.
@hansemannn  Please update test case mentioned in ticket for iOS < 11.2. Currently doc viewer is opened in iOS>=11.2. Ideally you have to remove the condition.

hansemannn:
@vijaysingh-axway Test-case updated to be used on all supported iOS versions.

PR #9970, opened 2018-03-28
[TIDOC-3160] - Remove very outdated Ti.Map documentation 

Topener:
Jira: https://jira.appcelerator.org/browse/TIDOC-3160


hansemannn:
Dont we have an open PR that removes all Ti.Map docs from the SDK? They are all pulled down from ti.map/apidoc/*.yml

Topener:
@hansemannn this PR removes it. Don't see another right now

PR #9964, opened 2018-03-28
[TIMOB-25799] Android: Cannot build titanium project with JDK 1.9

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25799

- Update to node-appc@0.2.45 to include JDK detection fix
- Remove `java.ext.dirs` usage which was removed in JDK 9, see `Removed Extension Mechanism` on the [migration guide](https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-7744EF96-5899-4FB2-B34E-86D49B2E89B6). This is compatible with all JDK versions we support (1.8 and above)
- Also handles JDK 10 which was recently released

**Notes**
- This will not cause `[appc] ti info -t jdk` to detect JDK 9 and above. That fix needs Titanium CLI 5.1.0, which is in Appc CLI 7.0.3-master.23

**Verification**

1. Download and install JDK 9/10 from http://www.oracle.com/technetwork/java/javase/downloads/index.html
2. Build a project using both JDK versions, you can switch using `[appc] ti config java.home <PATH>`

I've ran through module building and iOS/Windows building also as they use Java in places, there were no issues seen there

PR #9975, opened 2018-03-30
[TIMOB-25920] Updated node-appc to 0.2.45 to include the JDK 9 and 10 fix.

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25920

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3246,6 +3246,41 @@
           "requires": {
             "minimist": "0.0.8"
           }
+        },
+        "node-appc": {
+          "version": "0.2.43",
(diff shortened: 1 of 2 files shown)


ewanharris:
@cb1kenobi This is done in https://github.com/appcelerator/titanium_mobile/pull/9964 which also deals with the Java 9 incompatibility when spawning dexer, this ticket looks to be for the case when a non-Java 9 supporting JDK is selected and the build is performed. You're much more well versed in this than me but when there's a mismatch, my understanding is  the selected SDK, (the one forking the tiapp selected SDK build process) does do validation so this is to be expected?

PR #9983, opened 2018-04-05
[TIMOB-25927] Add Swift support for native modules, move SDK-core into "TitaniumKit.framework"

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25927

---

 > ⚠️If accepted, it will break native modules to be recompiled using TitaniumKit.framework instead of referencing single header files. It also removes the kroll thread and custom TiCore completely.

 > ⚠️Highly work in progress, requires major CLI, SDK and internal library (APSAnalytics & APSHTTPClient) changes

🗓 Scheduled for Q3/Q4

---

### Overview

Replaces the Titanium iOS Core with framework based alternative that is required if we want to deal with native modules that use Swift libraries, as we cannot bake Swift frameworks into static libraries and Swift frameworks do not allow non modular imports. Especially not with our (currently) hacky solution that attempts to reference symlinks from the application data directory. The following tasks represent the base of this effort.

### SDK Tasks

- [x] Create universal TitaniumKit.framework
- [x] Move SDK-core to TitaniumKit.framework
- [x] Remove Kroll Thread (use Main Thread)
- [x] Remove TiCore (use native JSCore)
- [x] Reference TitaniumKit as a target dependency of the core
- [x] Integrate TitaniumKit into the build pipeline
- [x] Update the CLI to copy TitaniumKit files into the packaged SDK
- [x] Update the CLI to copy TitaniumKit files into generated project build directory
- [ ] Be able to recompile core-framework if compiler flags change (e.g. `IS_XCODE_*`). Can we use `OTHER_C_FLAG=$(inherited) -DISXCODE_*`?
- [ ] Create .podspec for usage with CococaPods

### Module / CLI Tasks
- [x] Create framework based module template (embedded Swift framework)
- [x] Reference TitaniumKit locally
- [x] Test common Module API's (String, Number, Array, Dictionary, Date, id/Any)
- [ ] Update module build
  - [x] Compile (universal) framework instead of static library
  - [x] Package framework instead of static library
  - [x] Link framework based modules in SDK core (update _build.js `validateModules` step)
  - [x] Keep backwards compatibility for Obj-C based modules (which still need to be recompiled)
  - [x] Update templates to include a Swift module template. Prompt to choose between Swift and ObjC-module
  - [x] Be able to link TitaniumKit into module projects without actually linking the framework?
  - [ ] Update Obj-C module template to use TitaniumKit as well
  - [ ] Link frameworks to the "Embedded Frameworks" section (adjust the framework-manager from @janvennemann  to look for it?)
  - [ ] Write migration script to fix up old project imports -> RegEx replace for imports, link TitaniumKit, eventually symlink TitaniumKit to prevent breaking change?


sgtcoolguy:
Is this a continuation of #9798? Does it include the same changes (plus a whole lot more)?

hansemannn:
Yep, it does! Although I begin to feel that we may need to fix some memory issues around JSCore that haven't even been reported so far.

sgtcoolguy:
@hansemannn I wasn't suggesting we should have closed #9798

In fact, I think it would be more useful/easier to review and track if we had major parts of this PR separated:
- Removal of TiCore/move to JSCore exclusively
- Removal of KrollThread (did we deprecate that?)
- Move to framework
- Miscellaneous stuff like the Cocopods spec

hansemannn:
@sgtcoolguy I thought about this quite long so far: How do we minimize the impact of removing core-classes from modules in favor of a framework? Right now, developers import core-classes via
```objc
#import "TiApp.h"
import "TiUtils.h"
```
etc. This is possible, because we have "ghost" headers inside `~/Library/Application Support/Titanium/mobilesdk/osx/7.3.0/iphone/include` that are generated and copied during the scons build and have been removed as part of this PR already. So what if we would leave them there, but import the framework using the module-syntax (e.g. `@import Titanium;`) instead? It could possible work, since the module-syntax only imports the required classes if not already imported to avoid duplicate imports. 

We could still write a migration script that goes through the module's classes and scans for the "blacklisted" headers that are now part of the core-framework. Would this be a practical way? 

Sorry, I just need someone to talk to about this 😄.

sgtcoolguy:
@hansemannn 
I'm not sure I follow why it'd be hard to break apart the major parts of this PR. 

> Its super hard to remove ticore/krollthread once we moved the core to a framework. I am still unsure we will even be able to do the framework change, since it will be a semi-hard task to migrate old modules properly.

I get that the framework change may cause native module breakage, which we should work on in a PR that handles the move to a framework. But we should be able to remove TiCore via the old PR you closed. And I assume removing KrollThread would be similar.

As far as the framework changes around imports, that's outside my knowledge right now.

hansemannn:
Rebased and reworked from `next` in https://github.com/appcelerator/titanium_mobile/pull/10218

PR #9986, opened 2018-04-06
[TIMOB-25936] iOS: Fix appendSection & appendRow on main-thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25936


ewieberappc (approved):
FR Passed. Sections and rows are appended in the correct order. Tested using the provided code as well as modifications to the project to append in different ways and/or orders.

PR #9987, opened 2018-04-06
[TIMOB-25936] (7_1_X) iOS: Fix appendSection & appendRow on main-thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25936


janvennemann (approved):
LGTM. @hansemannn is it intended that the unit test does not have any assertions?

hansemannn:
@janvennemann It's a UI-related change, so the unit-test only ensures that the app does not crash with this concrete change. Thanks for the review!

ewieberappc:
FR Passed. Sections and rows are appended in the correct order. Tested using the provided code as well as modifications to the project to append in different ways and/or orders.

PR #9988, opened 2018-04-06
[FRAME-3] Add default template for Angular

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/FRAME-3

**Optional Description:**
This adds the default project template for Angular that will be used with the new `--ng` switch (see https://github.com/appcelerator/appc-cli-titanium/pull/258)


templates/app/angular-default/hooks/post-create.js
```diff
@@ -0,0 +1,35 @@
...
+		const packageJsonPath = path.join(projectPath, 'app', 'package.json');
+		if (fs.existsSync(packageJsonPath)) {
+			let packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
+			packageJson.name = projectName;
+			fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 4));
+		}
+
+		logger.info('Installing Angular project dependencies');
+		let errorOutput = '';
+		const child = spawn('npm', [ 'i' ], { cwd: path.dirname(packageJsonPath) });
```
ewanharris: Executing npm (or similar) commands through spawn is problematic, usually ending up with an ENOENT. I think one of the following should suffice (personally I think `shell: true`)

* Pass `shell: true` in the options of the spawn call. Only supported on Node 5.7.0 and up and we do support Node 4, but looking at the code I think it's safe to maybe error if you're using less than Node 8 or 6?
* Use exec instead of spawn (I think). But might run into maxbuffer issues
* Do the same as `shell: true` manually. Something like `spawn('cmd', ['/s', '/c', 'npm', 'i'],{ cwd: path.dirname(packageJsonPath) });`

ewanharris (commented):
Comment around spawning npm on Windows

hansemannn (approved):
Project wide changes are looking fine! Awaiting @ewanharris' approval to proceed. Also being blocked by the [appc-cli PR](https://github.com/appcelerator/appc-cli-titanium/pull/258). @ewanharris Can you review those as well?

ewanharris (approved):
Creating a project works on Windows, building to Android looks good, Windows (as a build target) has the known issue which I will file a ticket for.

PR #9999, opened 2018-04-12
fix(android): 9-patch padding box area ignored for Button backgroundImage

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-20357

**Description:**
Fix support for padding (content area) in 9Patch images. Set custom background directly to native components.
Removed some unused code. Reworked a few methods to avoid code duplication. 
Maybe we can have an `enum ` for the constants I have added?

**This is visual fix, so I don't think we can have unit tests for it.**

_Note:_ I think some testing with TiBackgroundDrawable would be worth it in order to check if we can stop using it. As far as I tested with 9Patch images we may not need it. But that would better be addressed in another ticket (if it is not already).

**Test case:**
_app.js_

```
var win = Ti.UI.createWindow({}),
    scrollView = Ti.UI.createScrollView({layout: 'vertical'}),
    btn = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'normalColors.png',
        title: 'very very looooong normal colors'
    }),
    btnPackaged = Ti.UI.createButton({
            top: 10,
            height: 50,
            backgroundImage: 'images/normalcolors.png',
            title: 'very very packaged normal colors'
        }),
    btn1 = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'allColors.9.png',
        title : 'very very loooooog all colors'
    }),
    btn1Packaged = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'images/allcolors.png',
        title : 'very very packaged all colors'
    }),
    btn2 = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'greenOnly.9.png',
        title : 'very very looooong green only'
    }),
    btn2Packaged = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'images/greenonly.png',
        title : 'very very packaged green only'
    }),
    btn3 = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'noPadding.9.png',
        title : 'very very looooong no padding'
    }),
    btn3Packaged = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'images/nopadding.png',
        title : 'very very packaged no padding'
    }),
    btn4 = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'twoColors.9.png',
        title : 'very very looooong two colors'
    }),
    btn4Packaged = Ti.UI.createButton({
        top: 10,
        height: 50,
        backgroundImage: 'images/twocolors.png',
        title : 'very very packaged two colors'
    });

btn1.addEventListener('click', function() {
    view.touchEnabled = !view.touchEnabled;
    btn4.backgroundImage = 'twoColors.9.png';
});
var label = Ti.UI.createLabel({text: 'Click this button!', top: 10});
var view = Ti.UI.createView({backgroundImage: 'ninecolors.9.png',
                            backgroundSelectedImage: 'invertedninecolors.9.png',
                            backgroundDisabledImage: 'disabledninecolors.9.png',
                            borderColor: 'blue',
                            borderWidth: 3,
                            touchEnabled: false,
                            top: 50,
                            width: 200,
                            height: 200});

scrollView.add([btn, btnPackaged, btn1, btn1Packaged, btn2, btn2Packaged, btn3, btn3Packaged, btn4, btn4Packaged]);

scrollView.add(view);
win.add(scrollView);
win.open();
```

Required files:
allcolors 9
greenonly 9
invertedninecolors 9
ninecolors 9
nopadding 9
normalcolors
twocolors 9

**Note:** I updated the test case to work with images from Resources folder and Android compiled in the /res folder. In order to get the latter to show the above images must be placed in a folder under `Resources/android/images/res-hdpi` 
`Resources/android/images/res-mdpi`
and so on.


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1376,7 +1376,7 @@ private void handleBackgroundImage(KrollDict d)
 					bg, TiConvert.toBoolean(d, TiC.PROPERTY_BACKGROUND_REPEAT, false), bgColor, bgSelected,
 					bgSelectedColor, bgDisabled, bgDisabledColor, bgFocused, bgFocusedColor, gradientDrawable);
 
-				background.setBackgroundDrawable(bgDrawable);
+				nativeView.setBackgroundDrawable(bgDrawable);
```
jquick-axway: I don't think we should make this change. The `applyCustomBackground()` method already calls `nativeView.setBackgroundDrawable(this.background)`. And the biggest issue with this change is the `handleBackgroundImage()` method (which gets called when background image/color/alpha/etc. changes dynamically) only modifies the `background` member variable and assumes its been assigned to `nativeView`. Your code above replaces the `background` member variable with the `bgDrawable` local drawable reference.

_(Note that I do agree that I don't like how background handling is implemented either. It's overly complicated.)_

jquick-axway (changes requested):
The 9-patch processing looks fine to me. Just a bit worried about the TiUIView change.

jquick-axway:
@ypbnv, I know that if you put a 9-patch PNG in the "res/drawable-*" directories then the Android build system will automatically extract the 9-patch pixels from the edges, create a new PNG within the APK's "res" folder, and when fetched via `Resources` it'll be a `NinePatchDrawable`. So, have you compared your 9-patch padding handling compared to Android's default handling of 9-patches from the "res" directory?

For a Titanium classic app, all you have to do is add a "*.9.png" file under the following folder...
`./Resources/android/images/res-mdpi`
`./Resources/android/images/res-hdpi`
`./Resources/android/images/res-xhdpi`
etc.

We should modify the test project so that QE can do a similar test to compare how we load a 9-patch from "assets" compared to how the Android OS loads it from "res/drawable" to ensure its loaded correctly. (Might be a good internal test to keep on hand too.)

ypbnv:
@jquick-axway While working on the padding I compared the 9patch with a similar application in native Android. I created a sample today with 9patch drawables in the Resources folder and in the density specific one. The result was that the stretched area and the padding are OK, but still with some differences. I think it has to do with scaling the bitmaps (or may be compiling the 9patches to the Android format) because the differences vary among devices with different dpi. I was not able to pin-point where that happens. Here is a screenshot comparing the assets and packaged 9patch drawables:
buttons

**BUT** the problem is that both of the options do not work if we assign the drawable through `background.setBackgroundDrawable()` method. I copied our TiBackgroundDrawable in a separate native Android project and it has the same behavior - the image is stretched properly, but the padding/content area is ignored. I think this happens due to the assigning the 9patch image indirectly as a background for a view with content. Maybe somehow layering the background through our TiBackgroundDrawable implementation prevents it from inheriting the content area of the native view. I tried to play around with the `innerRect` property, but didn't get anything (neither in Titanium, nor native).

TBH I ran out of ideas of dealing with that without changing the way we use TiBackgroundDrawable. Do you have any thoughts about what would be the best way to approach this?

sgtcoolguy:
@jquick-axway ping

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
jquick-axway: @ypbnv, I think see why there was a padding issue here.

In the old code, before calling `background.setBackgroundDrawable()`, we call `applyCustomBackground(false)` which in turn calls the main View's `setBackgroundDrawable()` method. If you look at Google's `setBackgroundDrawable()` code in the link below, notice that Google fetches and caches the padding from the background drawable. But the view only fetches the padding when the drawable has been assigned. The view is not aware of any padding changes that are made the drawable after it has been assigned.
https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/master/core/java/android/view/View.java

So, the reason the old code's view ignored the padding is because we were setting the 9-patch drawable **after** the background was assigned to the view. This means if we do the below, it should work-around the issue.
```java
background.setBackgroundDrawable(bgDrawable);
if (nativeView != null) {
	nativeView.setBackgroundDrawable(null);
	nativeView.setBackgroundDrawable(background);
}
```

We need to set it `null` first and then reset the background because Google's `View.setBackgroundDrawable()` checks if the reference has changed as an optimization. So, doing the above should fix other potential bugs when changing a view's background dynamically.

ypbnv: I tried that, but no effect on the padding problem. It gives me an idea though - to play with the order by which the background is built. Let's see if I can come up with something in that direction.

jquick-axway: Bummer. I really thought that would do it.
I'm definitely right about the view caching the background drawable's size though.

hansemannn:
Closing for now due to inactivity and merge conflicts.

PR #10003, opened 2018-04-13
[TIMOB-25963] Android: Improve KrollRuntime error output

garymathews:
- ~Provide access to current stack trace using `KrollRuntime.getInstance().getStackTrace()`~
- Update error output to match [node.js](https://nodejs.org/api/errors.html#errors_error_stack)
- Update exception dialog to show more details including stack trace
- Include Javascript stack trace with Java exceptions

###### TEST CASE \#1
```JS
var win = Ti.UI.createWindow({ backgroundColor: 'gray' });

win.addEventListener('open', function (e) {
   Ti.API.info(e.test.crash);
});

win.open();
```
```
/app.js:12
Ti.API.info(e.test.crash);
                   ^
Uncaught TypeError: Cannot read property 'crash' of undefined
    at (/app.js:12:23)
    at value(ti:/events.js:49:21)
    at value(ti:/events.js:101:19)
```

###### TEST CASE \#2
```JS
var win = Ti.UI.createWindow({ backgroundColor: 'gray' });

win.addEventListener('open', function (e) {
   Ti.Geolocation.accuracy = null;
});

win.open();
```
```
/app.js:4
  Ti.Geolocation.accuracy = null;
                          ^
   at (/app.js:4:28)
   at value(ti:/events.js:49:21)
   at value(ti:/events.js:101:19)

Uncaught Unable to convert null
   org.appcelerator.titanium.util.TiConvert.toInt(TiConvert.java:415)
   ti.modules.titanium.geolocation.GeolocationModule.propertyChangedAccuracy(GeolocationModule.java:414)
   ti.modules.titanium.geolocation.GeolocationModule.propertyChanged(GeolocationModule.java:384)
   org.appcelerator.kroll.KrollProxy.firePropertyChanged(KrollProxy.java:969)
   org.appcelerator.kroll.KrollProxy.onPropertyChanged(KrollProxy.java:1058)
   org.appcelerator.kroll.runtime.v8.V8Object.nativeFireEvent(Native Method)
   org.appcelerator.kroll.runtime.v8.V8Object.fireEvent(V8Object.java:63)
   org.appcelerator.kroll.KrollProxy.doFireEvent(KrollProxy.java:962)
   org.appcelerator.kroll.KrollProxy.handleMessage(KrollProxy.java:1186)
   org.appcelerator.titanium.proxy.TiViewProxy.handleMessage(TiViewProxy.java:394)
```

[TIMOB-25963](https://jira.appcelerator.org/browse/TIMOB-25963)
[TIMOB-25965](https://jira.appcelerator.org/browse/TIMOB-25965)

### DIFF
--- a/android/runtime/v8/src/native/JSException.cpp
+++ b/android/runtime/v8/src/native/JSException.cpp
@@ -15,6 +15,8 @@
 
 #include "JSException.h"
 
+#define MAX_STACK 10
+
 using namespace v8;
 
 namespace titanium {
@@ -29,29 +31,33 @@ Local<Value> JSException::fromJavaException(v8::Isolate* isolate, jthrowable jav
 	bool deleteRef = false;
 	if (!javaException) {
 		javaException = env->ExceptionOccurred();
-		env->ExceptionClear();
 		deleteRef = true;
 	}
-	//env->ExceptionDescribe();
+	env->ExceptionClear();
 
 	jstring javaMessage = (jstring) env->CallObjectMethod(javaException, JNIUtil::throwableGetMessageMethod);
 	if (!javaMessage) {
 		return THROW(isolate, "Java Exception occurred");
 	}
-	std::stringstream message(env->GetStringUTFChars(javaMessage, NULL));
+	const char* messagePtr = env->GetStringUTFChars(javaMessage, NULL);
+	std::stringstream message;
+	message << messagePtr;
 
 	jobjectArray frames = (jobjectArray) env->CallObjectMethod(javaException, JNIUtil::throwableGetStackTraceMethod);
 	jsize frames_length = env->GetArrayLength(frames);
-	for (int i = 0; i < (frames_length > 10 ? 10 : frames_length); i++) {
+	for (int i = 0; i < (frames_length > MAX_STACK ? MAX_STACK : frames_length); i++) {
 		jobject frame = env->GetObjectArrayElement(frames, i);
 		jstring javaStack = (jstring) env->CallObjectMethod(frame, JNIUtil::stackTraceElementToStringMethod);
 
-		message << std::endl << env->GetStringUTFChars(javaStack, NULL);
+		const char* stackPtr = env->GetStringUTFChars(javaStack, NULL);
+		message << std::endl << "    " << stackPtr;
 
+		env->ReleaseStringUTFChars(javaStack, stackPtr);
 		env->DeleteLocalRef(javaStack);
 	}
 	message << std::endl;
 
+	env->ReleaseStringUTFChars(javaMessage, messagePtr);
 	env->DeleteLocalRef(javaMessage);
 	if (deleteRef) {
 		env->DeleteLocalRef(javaException);
--- a/android/runtime/v8/src/native/JSException.cpp
+++ b/android/runtime/v8/src/native/JSException.cpp
@@ -4,6 +4,9 @@
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
+#include <cstring>
+#include <sstream>
+
 #include <jni.h>
 #include <v8.h>
 
@@ -18,12 +21,10 @@ namespace titanium {
 
 Local<Value> JSException::fromJavaException(v8::Isolate* isolate, jthrowable javaException)
 {
-	JNIEnv *env = JNIScope::getEnv();
+	JNIEnv* env = JNIScope::getEnv();
 	if (!env) {
 		return GetJNIEnvironmentError(isolate);
 	}
-
-	env->ExceptionDescribe();
 
 	bool deleteRef = false;
 	if (!javaException) {
@@ -31,22 +32,32 @@ Local<Value> JSException::fromJavaException(v8::Isolate* isolate, jthrowable jav
 		env->ExceptionClear();
 		deleteRef = true;
 	}
-
 	//env->ExceptionDescribe();
 
-	jstring message = (jstring) env->CallObjectMethod(javaException, JNIUtil::throwableGetMessageMethod);
-	if (!message) {
+	jstring javaMessage = (jstring) env->CallObjectMethod(javaException, JNIUtil::throwableGetMessageMethod);
+	if (!javaMessage) {
 		return THROW(isolate, "Java Exception occurred");
 	}
+	std::stringstream message(env->GetStringUTFChars(javaMessage, NULL));
 
-	Local<Value> jsMessage = TypeConverter::javaStringToJsString(isolate, env, message);
-	env->DeleteLocalRef(message);
+	jobjectArray frames = (jobjectArray) env->CallObjectMethod(javaException, JNIUtil::throwableGetStackTraceMethod);
+	jsize frames_length = env->GetArrayLength(frames);
+	for (int i = 0; i < (frames_length > 10 ? 10 : frames_length); i++) {
+		jobject frame = env->GetObjectArrayElement(frames, i);
+		jstring javaStack = (jstring) env->CallObjectMethod(frame, JNIUtil::stackTraceElementToStringMethod);
 
+		message << std::endl << env->GetStringUTFChars(javaStack, NULL);
+
+		env->DeleteLocalRef(javaStack);
+	}
+	message << std::endl;
+
+	env->DeleteLocalRef(javaMessage);
 	if (deleteRef) {
 		env->DeleteLocalRef(javaException);
 	}
 
-	return isolate->ThrowException(jsMessage->ToString(isolate));
+	return isolate->ThrowException(String::NewFromUtf8(isolate, message.str().c_str()));
 }
 
 }
--- a/android/runtime/v8/src/native/JSException.cpp
+++ b/android/runtime/v8/src/native/JSException.cpp
@@ -73,7 +73,7 @@ Local<Value> JSException::fromJavaException(v8::Isolate* isolate, jthrowable jav
 
 	// Now explicitly assign our properly generated stacktrace
 	Local<String> javaStack = String::NewFromUtf8(isolate, stackStream.str().c_str());
-	error->Set(context, STRING_NEW(isolate, "javaStack"), javaStack);
+	error->Set(context, STRING_NEW(isolate, "nativeStack"), javaStack);
 
 	// throw it
 	return isolate->ThrowException(error);
--- a/android/runtime/v8/src/native/JSException.cpp
+++ b/android/runtime/v8/src/native/JSException.cpp
@@ -16,8 +16,6 @@
 
 #include "JSException.h"
 
-#define MAX_STACK 10
-
 using namespace v8;
 
 namespace titanium {
--- a/android/runtime/v8/src/native/JSException.cpp
+++ b/android/runtime/v8/src/native/JSException.cpp
@@ -1,6 +1,6 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2011 by Appcelerator, Inc. All Rights Reserved.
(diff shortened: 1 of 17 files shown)


android/runtime/v8/src/native/JSException.cpp
```diff
@@ -12,41 +15,55 @@
...
+
+		const char* stackPtr = env->GetStringUTFChars(javaStack, NULL);
+		message << std::endl << "    " << stackPtr;
+
+		env->ReleaseStringUTFChars(javaStack, stackPtr);
+		env->DeleteLocalRef(javaStack);
+	}
+	message << std::endl;
 
+	env->ReleaseStringUTFChars(javaMessage, messagePtr);
```
sgtcoolguy: any reason we don't just do this cleanup right away after appending the messagePtr to the stream? (i.e. line 45 or so)

sgtcoolguy (changes requested):
The changes look OK to me. But I think you should be able to generate mocha unit tests from your manual examples here.

We can't test any changes to the dialog, but we should be able to test that accessing a undefined property throws an Error and that the stack generally looks like we expect it to.

sgtcoolguy:
```javascript
it('generates expected stack trace for undefined property', function () {
  try {
    Ti.API.info(e.test.crash);
    fail('Did not throw exception for accessing property off undefined');
  } catch(e) {
    // Now add some tests that verify the message/stacktrace are what you expect them to be. (maybe not exact lines/filenames, but matching a regexp pattern?)
    should(e.toString()).contain('Uncaught TypeError: Cannot read property \'crash\' of undefined');
  }
});

win.open();
```

sgtcoolguy:
I've been tweaking the unit tests locally to try and get this PR merge-ready, but I'm seeing some issues I think need to be addressed as a result of writing them.

For the Java exceptions that are being converted, they don't appear to be structured as I'd expect. I'd expect them to have `message` and `stack` properties holding the top-level message and stack trace we expect and then toString() prints them concatenated (with a preceding error type name).

The appears to throw an actual String which is the equivalent of `toString()`.

android/runtime/v8/src/native/JSException.cpp
```diff
@@ -12,41 +15,55 @@
...
+		env->DeleteLocalRef(javaStack);
+	}
+	message << std::endl;
 
 	if (deleteRef) {
 		env->DeleteLocalRef(javaException);
 	}
 
-	return isolate->ThrowException(jsMessage->ToString(isolate));
+	return isolate->ThrowException(String::NewFromUtf8(isolate, message.str().c_str()));
```
sgtcoolguy: So here we have always done the wrong thing which is to throw a String. We should be throwing an actual Exception (Error).

I think the only wrinkle here is to override the `stack` property to hold the java stack trace we generate here.

```cpp
// Grab the top-level error message
Local<Value> jsMessage = TypeConverter::javaStringToJsString(isolate, env, javaMessage);
env->DeleteLocalRef(javaMessage);
// Create a JS Error holding this message
Local<Value> error = v8::Exception::Error(jsMessage);

// Now loop through the java stack and generate a JS String from the result and assign to Local<String> stack
std::stringstream stackStream;
jobjectArray frames = (jobjectArray) env->CallObjectMethod(javaException, JNIUtil::throwableGetStackTraceMethod);
jsize frames_length = env->GetArrayLength(frames);
for (int i = 0; i < (frames_length > MAX_STACK ? MAX_STACK : frames_length); i++) {
	jobject frame = env->GetObjectArrayElement(frames, i);
	jstring javaStack = (jstring) env->CallObjectMethod(frame, JNIUtil::stackTraceElementToStringMethod);

	const char* stackPtr = env->GetStringUTFChars(javaStack, NULL);
 	stackStream << std::endl << "    " << stackPtr;

	env->ReleaseStringUTFChars(javaStack, stackPtr);
	env->DeleteLocalRef(javaStack);
}
stackStream << std::endl;
Local<Value> stack = String::NewFromUtf8(isolate, stackStream.str().c_str());

// Now explicitly assign our properly generated stacktrace
error->Set(context, STRING_NEW(isolate, "stack"), stack);

// throw it
return isolate->ThrowException(error);
```

sgtcoolguy (approved):

android/runtime/v8/src/native/JSException.cpp
```diff
@@ -1,52 +1,82 @@
...
 	}
+	stackStream << std::endl;
+	Local<String> stack = String::NewFromUtf8(isolate, stackStream.str().c_str());
+
+	// Now explicitly assign our properly generated stacktrace
+	error->Set(context, STRING_NEW(isolate, "stack"), stack);
 
-	return isolate->ThrowException(jsMessage->ToString(isolate));
+	// throw it
+	return isolate->ThrowException(error);
```
garymathews: @sgtcoolguy The Java stack trace is now not shown in the log:
```
/app.js:9
       Ti.Geolocation.accuracy = null;
                               ^
   at (/app.js:9:33)
   at value(ti:/events.js:49:21)
   at value(ti:/events.js:101:19)

Uncaught Error: Unable to convert null
```

I appended it to the message as `ExceptionMessage` does not contain a stack element

android/runtime/v8/src/native/JSException.cpp
```diff
@@ -1,52 +1,82 @@
...
+		env->DeleteLocalRef(javaStack);
+	}
 	if (deleteRef) {
 		env->DeleteLocalRef(javaException);
 	}
+	stackStream << std::endl;
+
+	// Now explicitly assign our properly generated stacktrace
+	Local<String> javaStack = String::NewFromUtf8(isolate, stackStream.str().c_str());
+	error->Set(context, STRING_NEW(isolate, "javaStack"), javaStack);
```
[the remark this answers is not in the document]
sgtcoolguy: Yeah, I'd prefer `nativeStack` since we should aim to achieve equivalent structure for the other platforms.

android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -146,17 +144,41 @@ void V8Util::openJSErrorDialog(Isolate* isolate, TryCatch &tryCatch)
...
+
+	if (exception->IsObject()) {
+		Local<Object> error = exception.As<Object>();
+		jsStack = exception.As<Object>()->Get(STRING_NEW(isolate, "stack"));
+		javaStack = exception.As<Object>()->Get(STRING_NEW(isolate, "javaStack"));
+	}
 
-	V8Runtime::exceptionStackTrace.Reset(isolate, message->GetStackTrace());
+	// obtain javascript stack trace
+	if (jsStack->IsNullOrUndefined() || jsStack.IsEmpty()) {
```
sgtcoolguy: reverse the order here? I was worried if trying to call `IsNullOrUndefined()` on an empty handle would crash.

android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -146,17 +144,41 @@ void V8Util::openJSErrorDialog(Isolate* isolate, TryCatch &tryCatch)
...
 	Local<Message> message = tryCatch.Message();
-	Local<Value> stack = tryCatch.Exception().As<Object>()->Get(context, STRING_NEW(isolate, "stack")).ToLocalChecked();
+	Local<Value> exception = tryCatch.Exception();
+
+	Local<Value> jsStack;
+	Local<Value> javaStack;
+
+	if (exception->IsObject()) {
+		Local<Object> error = exception.As<Object>();
+		jsStack = exception.As<Object>()->Get(STRING_NEW(isolate, "stack"));
```
sgtcoolguy: use the `error` objet reference you assign just above here rather than call .As<Object> again

android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -146,17 +144,41 @@ void V8Util::openJSErrorDialog(Isolate* isolate, TryCatch &tryCatch)
...
+			}
+		}
+	}
 
 	jstring title = env->NewStringUTF("Runtime Error");
 	jstring errorMessage = TypeConverter::jsValueToJavaString(isolate, env, message->Get());
 	jstring resourceName = TypeConverter::jsValueToJavaString(isolate, env, message->GetScriptResourceName());
 	jstring sourceLine = TypeConverter::jsValueToJavaString(isolate, env, message->GetSourceLine());
-	jstring javaStack = TypeConverter::jsValueToJavaString(isolate, env, stack);
+	jstring jsStackString = TypeConverter::jsValueToJavaString(isolate, env, jsStack);
```
sgtcoolguy: I think you have to guard here for empty handle and/or undefined. TypeConverter handles `null` -> `NULL`, but I think it converts `undefined` to `'undefined'`. And I think it'll just crash if the handle is empty.

I assigned to a bool up above for when the stack/nativeStack were not null/undefined/empty and if they were, I assigned NULL here and avoided the `DeleteLocalRef` below.

garymathews: I think we should improve `TypeConverter` to handle this validation?

Updated PR

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -114,7 +114,7 @@ jstring TypeConverter::jsValueToJavaString(v8::Isolate* isolate, v8::Local<v8::V
 
 jstring TypeConverter::jsValueToJavaString(v8::Isolate* isolate, JNIEnv *env, v8::Local<v8::Value> jsValue)
 {
-	if (jsValue->IsNull()) {
+	if (jsValue.IsEmpty() || jsValue->IsNullOrUndefined()) {
```
sgtcoolguy: I *think* this is OK, but beware this is a behavioral change here. Before we'd change an `undefined` to `'undefined'`

android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -146,12 +144,41 @@ void V8Util::openJSErrorDialog(Isolate* isolate, TryCatch &tryCatch)
...
+	Local<Context> context = isolate->GetCurrentContext();
 	Local<Message> message = tryCatch.Message();
+	Local<Value> exception = tryCatch.Exception();
+
+	Local<Value> jsStack;
+	Local<Value> javaStack;
+
+	if (exception->IsObject()) {
+		Local<Object> error = exception.As<Object>();
+		jsStack = error->Get(context, STRING_NEW(isolate, "stack")).ToLocalChecked();
```
sgtcoolguy: This and the next line could cause a crash if the handle is empty. You could do something like this to better handle it:
```cpp
jsStack = error->Get(context, STRING_NEW(isolate, "stack")).FromMaybe(Undefined(isolate).As<Value>());
javaStack = error->Get(context, STRING_NEW(isolate, "nativeStack")).FromMaybe(Undefined(isolate).As<Value>());
```

In that code if the handle is empty due to some error, we'll default to `undefined`.

android/runtime/v8/src/native/V8Util.cpp
```diff
@@ -146,12 +144,41 @@ void V8Util::openJSErrorDialog(Isolate* isolate, TryCatch &tryCatch)
...
+		javaStack = error->Get(context, STRING_NEW(isolate, "nativeStack")).ToLocalChecked();
+	}
+
+	// obtain javascript stack trace
+	if (jsStack->IsNullOrUndefined() || jsStack.IsEmpty()) {
+		Local<StackTrace> frames = message->GetStackTrace();
+		if (frames.IsEmpty() || !frames->GetFrameCount()) {
+			frames = StackTrace::CurrentStackTrace(isolate, 10);
```
sgtcoolguy: should we extract the 10 constant here like you did in JSException?

hansemannn:
QE?

PR #10006, opened 2018-04-19
[TIMOB-25971] : iOS minimumFontSize property in Ti.UI.label lowers text baseline as it decreases

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25971


iphone/Classes/TiUILabel.m
```diff
@@ -463,6 +463,7 @@ - (void)setMinimumFontSize_:(id)size
     [label setAdjustsFontSizeToFitWidth:YES];
     CGFloat ratio = minFontSize / label.font.pointSize;
     [label setMinimumScaleFactor:ratio];
+    [label setBaselineAdjustment:UIBaselineAdjustmentNone];
```
hansemannn: Good fix! I am wondering what the baseline-adjustment is before setting this? Since `UIBaselineAdjustmentNone` is already the default. And should we expose this as a property as well? Are there use-cases in our environment for this? And Android parity? Definitely as an own ticket, but maybe not even relevant.

hansemannn (approved):

iphone/Classes/TiUILabel.m
vijaysingh-axway: Default is UIBaselineAdjustmentAlignBaselines. This property is only relevant with "adjustsFontSizeToFitWidth" = true. I don't think it is more relevant to expose, if we have already expose property minimumFontSize which includes minimumScaleFactor and adjustsFontSizeToFitWidth.

PR #10007, opened 2018-04-19
[TIMOB-16596] Android: Animation anchorPoints not working

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-16596

**Description:**
Use Animation's anchorPoint property instead of Operation's.

_Note:_ Currently no unit tests can be added for this. They will be included in an upcoming PR addressing other issues with animation that prevents this.

**Test case:**
1) In the JIRA ticket.

2)
_app.js_
```
var t = Ti.UI.create2DMatrix({rotate:90, anchorPoint:{x: 0, y:0}});

var a = {
  transform: t,
  duration: 2000,
  // This property can be used to check the priority of anchorPoint used
  //anchorPoint: {x:1, y:1}
};

var view = Ti.UI.createView({
  backgroundColor:'#336699',
  width:100, height:100
});

win.add(view);

  title:'Animate ',
  height: 80,
  width: 300,
  top:30
});

win.add(button);

button.addEventListener('click', function(){
  button.touchEnabled = false;
  view.animate(a, function(){
    button.touchEnabled = true;
  });
});

win.open();
```


android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
```diff
@@ -360,7 +360,7 @@ private AnimatorSet buildPropertyAnimators(int x, int y, int w, int h, int paren
 				for (Operation operation : operations) {
 					if (operation.anchorX != Ti2DMatrix.DEFAULT_ANCHOR_VALUE
 						|| operation.anchorY != Ti2DMatrix.DEFAULT_ANCHOR_VALUE) {
-						setAnchor(w, h, operation.anchorX, operation.anchorY);
+						setAnchor(w, h, anchorX, anchorY);
```
jquick-axway: I think the if-condition should be written like the below.
```java
final float EPSILON = Math.ulp(1.0f);
boolean hasAnchorXChanged = (Math.abs(operation.anchorX - anchorX) >= EPSILON);
boolean hasAnchorYChanged = (Math.abs(operation.anchorY - anchorY) >= EPSILON);
if (hasAnchorXChanged || hasAnchorYChanged) {
	setAnchor(w, h, anchorX, anchorY);
}
```

This way if the first time the operations were using `DEFAULT_ANCHOR_VALUE` and the anchor was changed to something else, then the operations would be updated instead of locking-in to using the previous values. The above effectively doing `operation.anchorX != anchorX`, but with floats.

_(I've never played with this code before. So, hopefully I'm understanding things correctly. Am I right?)_

jquick-axway (changes requested):

android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
ypbnv: I don't think that's what was intended. The test case has an animation with anchor point that is different by the default one, which is then followed by an animation with the default one. And this will result in playing the first animation twice, because the difference will be less than EPSILON.

I also found out that we can have anchorPoint as a parameter in the creation dictionary of Ti.UI.2DMatrix. And my changes totally ignore that, which is not OK. So I will revisit the code. check what is the priority on iOS and fix it.

ypbnv:
@jquick-axway I have updated the PR. Now the animation will use `anchorPoint` in the following order of priority:
 - `anchorPoint` set as a property in the instance of Ti.UI.Animation;
 - `anchorPoint` set in the creation dictionary matrix of the transformation;
This is an Android only case, because `anchorPoint` as a part of the creation dictionary for 2DMatrix is not supported on iOS. To me it makes sense to have that priority. I have added a note for that in the docs as well. I have added another test case too.

lokeshchdhry:
@ypbnv , I see that `anchorPoints` are not working on android 4.1. Works fine on Android 8.0 & Android 6.0.1 with the fix.

jquick-axway (changes requested):

jquick-axway:
@ypbnv, can you also remove all of the `PRE_HONEYCOMB ` parts please?
That's Android 3.x which we do not support anymore. Thanks.

ypbnv:
@jquick-axway , @lokeshchdhry Updated with fix for API 16-19. I don't like the way it is done, but couldn't figure out anything else. I assume this is the reason for it:
API 18: 
https://android.googlesource.com/platform/frameworks/base.git/+/jb-release/core/java/android/view/View.java#9110
API 19:
https://android.googlesource.com/platform/frameworks/base.git/+/kitkat-release/core/java/android/view/View.java#9703

To me it looks like it was not accounting if the pivot point was explicitly set and since 0 matched with the default values in TransformationInfo class it just ignored it.

Also removed unused code and improved the current check for default anchorPoint coordinates.

android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
```diff
@@ -1481,18 +1420,23 @@ private void setAnchor(int width, int height, float thisAnchorX, float thisAncho
...
-			setViewPivot(pivotX, pivotY);
-		} else {
-			setViewPivotHC(pivotX, pivotY);
+		// Passing 0 as both X and Y parameters for a view's pivot point
+		// results in setting the point in the view's center for APIs 16
+		// to 19. This is why we add one for both paramters if this is the
+		// case.
+		if (PRE_LOLLIPOP && thisAnchorX == 0 && thisAnchorY == 0) {
+			pivotX++;
+			pivotY++;
```
jquick-axway: Pivot is a normalized value between 0.0 and 1.0.

So, if given pivot is (0,0) which is the top-left corner, then adding 1 to both will make it pivot at the bottom-right corner, right?

Perhaps adding epsilon would be better?

ypbnv: On our end - yes, it is between 0.0 and 1.0. In the Android code it is passed as pixels.
view.setPivotX() leads to
https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/android/view/RenderNode.java#632

It is multiplied by the view's dimensions here:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java#L1484
and here:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java#L1488

So in the end it will count as the pixel in position (1,1). 
Increasing them with EPSILON will work, too.

jquick-axway: Ahh... thanks man. I didn't know that. :)

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

Animation anchor points works as expected.

OS Ver:         10.13.4
Xcode Ver:      Xcode 9.3
Appc NPM:       4.2.13
Appc CLI:       7.0.3
Daemon Ver:     1.1.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10

PR #10010, opened 2018-04-25
[TIMOB-25985] iOS/Android: Add "attributedString" API to Ti.UI.Button

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25985


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
```diff
@@ -188,4 +201,14 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
...
+
+	private void setAttributedStringText(AttributedStringProxy attrString)
+	{
+		AppCompatButton btn = (AppCompatButton) getNativeView();
+		Bundle bundleText =
+			AttributedStringProxy.toSpannableInBundle(attrString, TiApplication.getAppCurrentActivity());
+		if (bundleText.containsKey(TiC.PROPERTY_ATTRIBUTED_STRING)) {
+			btn.setText((Spannable) bundleText.getCharSequence(TiC.PROPERTY_ATTRIBUTED_STRING));
+		}
+	}
```
jquick-axway: I don't think you need to call `toSpannableInBundle()` do you? We don't need to for labels, which uses Android's `TextView` Java class... andr Android `Button` class inherits from `TextView` as well. Also, the `setText()` method accepts a `CharSequence` type that a `Spannable` type inherits from. So, no casting should be needed. How about something like this?

```java
AppCompatButton btn = (AppCompatButton) getNativeView();
if (btn != null) {
	CharSequence text = AttributedStringProxy.toSpannable(attrString, TiApplication.getAppCurrentActivity());
	if (text == null) {
		text = "";
	}
	btn.setText(text);
}
```

jquick-axway (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
hansemannn: It was basically copied from the label implementation, but I could change it!

jquick-axway: Oh I see. You copied it from "TiUIText.java" which is our TextField/TextArea code (seems to be doing an unnecessary step). The way I'm doing it is based on the "TiUILabel.java" code.

hansemannn: Perfect! Will change now.

jquick-axway (approved):
CR: Pass (Android)

vijaysingh-axway (approved):
CR approved for iOS.

PR #10013, opened 2018-04-26
[TIMOB-17363] Android/iOS: Added "MaskedImage" support to Android. Moved BLEND_MODE constants to "Ti.UI".

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-17363

**Summary:**
- Moved `BLEND_MODE_*` constants from `Ti.UI.iOS` to `Ti.UI`.
  * Constants are still defined under `Ti.UI.iOS` for backward compatibility on iOS, but are documented as deprecated.
- Added `Ti.UI.MaskedImage` support to Android.
  * Intended as alternate solution to using rounded borders around images, which has hardware acceleration issues on Android.
- Supports 9-patch images, but only if loaded from APK's "res" directory.
- Supports base View class' `tintColor` and `opacity` properties. Applied to final result.
- Super optimized masking via GPU shaders on Android for:
  * `mask` and `tint` for all blend modes.
  * `mask` and `image` for modes `SOURCE_IN` and `DESTINATION_IN`.
- All other masking combinations not in above list resort to less optimized rendering on Android.
- Android does not support the following blend modes _(will log a warning)_:
  * `BLEND_MODE_COLOR`
  * `BLEND_MODE_COLOR_BURN`
  * `BLEND_MODE_COLOR_DODGE`
  * `BLEND_MODE_DIFFERENCE`
  * `BLEND_MODE_EXCLUSION`
  * `BLEND_MODE_HARD_LIGHT`
  * `BLEND_MODE_HUE`
  * `BLEND_MODE_LUMINOSITY`
  * `BLEND_MODE_PLUS_DARKER`
  * `BLEND_MODE_SATURATION`
  * `BLEND_MODE_SOFT_LIGHT`
- Known issues _(to be written up later)_:
  * Will not auto-rotate JPEGs based on EXIF orientation.
  * `toImage()` renders bad image on Android 7.0 and higher when masking an image.

**Test:**
Run the below test procedure on...
- Android 4.1
- Android 4.4
- Android 7.0 or higher

Test procedure...
1. Create a classic app and add the contents of "MaskedImageTest.zip" to it from [TIMOB-17363](https://jira.appcelerator.org/browse/TIMOB-17363).
2. On app startup, verify appc logo is masked within a circle like [Screenshot1](https://jira.appcelerator.org/secure/attachment/65099/Screenshot1.png)
3. Tap the "Size" button and select "Ti.UI.FILL".
4. Verify image stretches to fit container like [Screenshot2](https://jira.appcelerator.org/secure/attachment/65100/Screenshot2.png).
5. Rotate device to landscape.
6. Verify image stretches to fit landscape container like [Screenshot3](https://jira.appcelerator.org/secure/attachment/65101/Screenshot3.png).
7. Rotate device back to portrait.
8. Tap the "Mode" button and select "SOURCE_OUT".
9. Verify image is only displayed outside of circle like [Screenshot4](https://jira.appcelerator.org/secure/attachment/65102/Screenshot4.png).
10. Tap the "Mask" button and select "MaskCircleLock".
11. Tap the "Image" button and select "None".
12. Tap the "Tint" button and select "Red".
13. Verify lock is red, within a white circle, and outside of circle is red like [Screenshot5](https://jira.appcelerator.org/secure/attachment/65103/Screenshot5.png).
14. Tap the "Mode" button and select "SOURCE_IN".
15. Verify lock is white with a red circle like [Screenshot6](https://jira.appcelerator.org/secure/attachment/65104/Screenshot6.png).
16. Tap the "Size" button and select "300dp".
17. Verify circle image is no longer stretched like [Screenshot7](https://jira.appcelerator.org/secure/attachment/65105/Screenshot7.png).
18. Tap the "Image" button and select "MaskVerticalGradient".
19. Verify circle has a gradient (light red on top; full red on bottom) like [Screenshot8](https://jira.appcelerator.org/secure/attachment/65106/Screenshot8.png).


apidoc/Titanium/UI/UI.yml
```diff
@@ -1203,312 +1203,250 @@ properties:
...
-        removed: "6.0.0"
     summary: |
         Use with [MaskedImage.mode](Titanium.UI.MaskedImage.mode) to specify a blend mode.
+    description: |
+        Android does not support this blend mode.
     type: Number
     permission: read-only
-    platforms: [iphone, ipad]
+    since: "7.2.0"
+    platforms: [android, iphone, ipad]
```
hansemannn: If it is not supported on Android, I would just leave out the `android` key. If we would be very exact here, we would leave the iOS only constants on the iOS namespace and put the parity ones here.

iphone/Classes/UIModule.m
```diff
@@ -164,6 +164,35 @@ - (TiUIActivityIndicatorStyleProxy *)ActivityIndicatorStyle
 MAKE_SYSTEM_PROP(PICKER_TYPE_TIME, UIDatePickerModeTime);
 MAKE_SYSTEM_PROP(PICKER_TYPE_COUNT_DOWN_TIMER, UIDatePickerModeCountDownTimer);
 
+MAKE_SYSTEM_PROP(BLEND_MODE_NORMAL, kCGBlendModeNormal);
```
hansemannn: Please deprecate the old ones. We have a macro for this: 
```objc
// Definition
MAKE_SYSTEM_PROP_DEPRECATED_REPLACED(name, map, api, in, newapi);

// Example
MAKE_SYSTEM_PROP_DEPRECATED_REPLACED(BLEND_MODE_NORMAL, kCGBlendModeNormal, @"UI.iOS.BLEND_MODE_NORMAL", @"7.3.0", @"UI.BLEND_MODE_NORMAL");
```

Note: The `Ti` / `Titanium` is not included into the log as it is replaced by the app name during project generation. For the log, it's manually constructed to avoid being replaced, so we leave it out in every log.

tests/Resources/ti.ui.maskedimage.addontest.js
```diff
@@ -0,0 +1,77 @@
...
+
+	it('mask-image', function (finish) {
+		this.timeout(5000);
+		win = Ti.UI.createWindow();
+		win.add(Ti.UI.createMaskedImage({
+			mask: '/Logo.png',
+			image: '/SplashScreen.png',
+			mode: Ti.UI.BLEND_MODE_SOURCE_IN,
+			width: Ti.UI.FILL,
+			height: Ti.UI.FILL,
```
hansemannn: It should fill by default. How does this behave on iOS?

hansemannn (changes requested):
Left some review changes, mainly about docs and iOS.

tests/Resources/ti.ui.maskedimage.addontest.js
jquick-axway: Android currently auto-sizes by default. I'll double check the iOS handling.

apidoc/Titanium/UI/UI.yml
jquick-axway: I still define the constant on Android, but log a warning instead. I think it's better than letting the app crash.

It is possible to add the other modes on Android, but we would have to manipulate the bitmap pixels ourselves the slow way. (yuck)

I'll remove the "android" platform keys from the unsupported constants.

iphone/Classes/UIModule.m
jquick-axway: Will do.

tests/Resources/ti.ui.maskedimage.addontest.js
jquick-axway: I just re-tested this on iOS. You're right. An iOS `MaskedImage` does a `Ti.UI.FILL` by default. I'll change Android to make it match.

Also, our iOS `MaskedImage` doesn't appear to support `Ti.UI.SIZE`. I get a 0x0 size view when attempting to auto-size it. On Android, I made this work. I auto-size based on the image if available, otherwise I fallback to auto-sizing based on the mask size. (I'll keep this behavior.)

jquick-axway:
@hansemannn, do you know if we need to do anything on the Alloy side after moving the BLEND_MODE constants?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIMaskedImage.java
```diff
@@ -0,0 +1,1399 @@
...
+					try {
+						// If mask is not an 8-bit grayscale bitmap, then convert it into one.
+						if (bitmap.getConfig() != Bitmap.Config.ALPHA_8) {
+							bitmap = bitmap.extractAlpha();
+						}
+
+						// Android 4.4 has a bug where if bitmap byte width does not fit a 4 byte packing alignment,
+						// then mask will appear skewed. Resizing bitmap doesn't work. So, give up if this happens.
+						if ((Build.VERSION.SDK_INT == 19) || (Build.VERSION.SDK_INT == 20)) {
+							if ((bitmap.getWidth() % 4) != 0) {
```
garymathews: Could this be moved above the `bitmap.extractAlpha()` to fail before reaching `bitmap.extractAlpha()` ?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIMaskedImage.java
```diff
@@ -0,0 +1,1399 @@
...
+
+	/**
+	 * Validates given Titanium "tint" property value and applies it to the view.
+	 * @param value Object representing a color value or name. Set to null to disable tinting.
+	 */
+	private void updateMaskTintWith(Object value)
+	{
+		if (this.maskedDrawable != null) {
+			if (value != null) {
+				int color = TiConvert.toColor(TiConvert.toString(value));
```
garymathews: Maybe validate if `value` is a string?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIMaskedImage.java
```diff
@@ -0,0 +1,1399 @@
...
+
+	/**
+	 * Validates given Titanium "tintColor" property value and applies it to the view.
+	 * @param value Object representing a color value or name. Set to null to disable tinting.
+	 */
+	private void updateTintColorFilterWith(Object value)
+	{
+		if (this.maskedDrawable != null) {
+			if (value != null) {
+				int color = TiConvert.toColor(TiConvert.toString(value));
```
garymathews: Same here

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIMaskedImage.java
jquick-axway: Ahh, good catch. I was trying to "fix" the resulting grayscaled bitmap, but later gave up.

jquick-axway:
**Updated PR:**
- `Ti.UI.iOS` BLEND_MODE constants will now log a warning if used.
- Modified Android's `MaskedImage` to be `Ti.UI.FILL` by default like iOS. (Was `Ti.UI.SIZE`.)
- Now logs a warning if a non-string type was given to `tint` and `tintColor` properties.
- Modified Android to support mask, image, and tint all at the same time like iOS.
  * Used to not support both tint and image at the same time.
- Fixed dynamic changes of `mask`, `tint`, and `mode` properties on Android. _(Would not always update onscreen.)_
- Updated docs to indicate which BLEND_MODE constants are not supported by the Android platform.

hansemannn (approved):

PR #10015, opened 2018-04-26
[TIMOB-25992] Android/iOS: Add clang-format to the default modules template

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25992


hansemannn:
Can you add iOS as well? And add a paragraph on how to use it? I could help with that if unsure.

hansemannn (approved):
Approved!

PR #10017, opened 2018-04-30
[TIMOB-25431] Improve getActivity() validation

garymathews:
- `getActivity()` must always return a valid activity. There are many instances where we assume `getActivity()` to return a valid activity [TiUILabel#L71](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java#L71), [TiTableView.java#L307](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java#L307) etc...
- Add validation to `TiUICardView.getLayout()`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25431)


sgtcoolguy (approved):
CR looks good to me.

jquick-axway (approved):
CR: Pass

jquick-axway:
@drauggres, while I agree with you, the problem is too much code (especially modules) assumes this method never returns null and the stack-trace makes it look like a core problem on our end. This issue is already too wide spread. So, with this change, this method will grab the 1st launched splash-screen activity if no other child activity exists on the UI stack since it's guaranteed to exist.

PR #10019, opened 2018-05-01
[TIMOB-26025] TiAPI: Improve Ti.Filesystem.File parity

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-26025

**Description:**
This stole quite a bit of @hansemannn 's PR from #9750 on the iOS side to better define return types of methods/properties and at least log errors that may happen under the hood. This *did not* incorporate the ARC changes (I was concerned they were the cause of the crashes on that PR).

**iOS:**
- Deprecate `Ti.Filesystem.File #createTimestamp()` and `#modificationTimestamp()` and the property equivalents (log deprecation note when old methods are called). Add `#createdAt()` and `#modifiedAt()` as new replacements that return a Date. 
- Add `Ti.Filesystem.File#copy()` method. 
- Default `Ti.Filesystem.File#createDirectory()` to be recursive like other platforms. This took an undocumented single argument to do this recursively in the past. I retained the ability to pass an explicit false in to the method to not create a directory recursively - not sure what we want to do here. This is a behavioral change.
- Enable the `Ti.Filesystem.File.parent` test

**Android:**
- Deprecate `Ti.Filesystem.File #createTimestamp()` and `#modificationTimestamp()` (log deprecation note when old methods are called). Add `#createdAt()` and `#modifiedAt()` as new replacements that return a Date.
- Add `Ti.Filesystem.File#append()` method. 
- Fix returning true for `Ti.Filesystem.File#isDirectory()` on non-existent Resource sub-dirs.

**JIRA:** 
- https://jira.appcelerator.org/browse/TIMOB-19128
- https://jira.appcelerator.org/browse/TIMOB-23495
- https://jira.appcelerator.org/browse/TIMOB-25683
- https://jira.appcelerator.org/browse/TIMOB-23493


hansemannn (approved):
iOS looks good (😄), **CR approved**!

sgtcoolguy:
So my big concern here is that technically the createDirectory and createTimestamp/modificationTimestamp changes may be breaking API changes. I think createDirectory change is borderline, the behavior is slightly different but shouldn't dramatically affect people assuming it's not recursive who would create each folder themselves.

The other change from Date -> Number on the timestamps however could potentially break clients' apps if they're assuming a Date object. We may need to hold off on that change until 8.0.0 or later? Maybe add a deprecation notice when called that the return type will soon change? cc @mukherjee2 @eric34 

I'm not sure how to best handle a case where the implementation is wrong according to our own docs and other platform behavior, but that would cause a return type to change.

hansemannn:
As discussed during standup, lets deprecate it asap. One additional idea: To have a clean migration, let's introduce the new properties `createdAt` and `modifiedAt` that are 7.2.0+. In 8.0.0, we deprecate `createTimestamp` and `modificationTimestamp` in favor of the new properties and finally remove the old ones in 9.0.0.

sgtcoolguy:
So my concern here is that introducing alternate methods would be a bit of work for all the platforms, and that the new names actually suggest return types of Date mores than the current names. iOS just happens to return Date objects for a method clearly named to suggest a unix timestamp return value. I think it makes sense to make a deprecation note now and not change the return type to Number yet, but I don't see a nice way to avoid the issue of the method suddenly changing return type on iOS in the future.

Moving to a new method name is one way to do it, as you suggested, but then that suggests moving all the other platforms to Date to me and requires more work - and I'd still be reluctant to remove the timestamp variants.

Looking at Node's `fs` API, they actually have both `Date` and `Number` versions (`ctime` vs `ctimeMs`, `time` vs `mtimeMs`)

So maybe we combine our solutions?
- Introduce new `createdAt` and `modifiedAt` methods that return `Date` on all platforms
- Add deprecation note on iOS's `creationTimestamp` and `modificationTimestamp` that they will return a Number in a future release (8.0? 9.0?) and to use `createdAt` or `modifiedAt` variants to get a `Date` object.

Topener:
Date Objects make a lot of sense in JavaScript world. For a lot of purposes you'd actually be making date objects yourself as well. And getting timestamp from data object is easy.

I'm in favour of just having 1 property, based on Hans his suggestion, the `createdAt` and `modifiedAt` properties. Removal in 9.0 makes sense too.

hansemannn:
@sgtcoolguy I can help doing the `createdAt` / `modifiedAt` changes incl. the deprecation in a separate PR. Let me know!

sgtcoolguy:
@infosia Hey Kota - so to fix some parity issues we're looking at deprecating `Ti.Filesystem.File` `#createTimestamp()` and `#modificationTimestamp()` to be replaced by new `#createdAt()` and #modifiedAt()` methods that return a Date.

sgtcoolguy:
@hansemannn @Topener I've pushed the changes to deprecate the timestamp methods on iOS and Android and added the new variants that return a Date.

PR #10022, opened 2018-05-03
[TIMOB-26007] iOS: Refactor crash dialog

garymathews:
- Refactor crash dialog to display stack trace

##### TEST CASE `#1`
```JS
var win = Ti.UI.createWindow({ backgroundColor: 'gray' });

win.addEventListener('open', function (e) {
   Ti.API.info(e.test.crash);
});

win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26007)


iphone/Classes/TiUtils.m
```diff
@@ -2097,4 +2097,24 @@ + (BOOL)isSupportedFragment:(id)fragment
       [fragment isKindOfClass:[NSDate class]] || [fragment isKindOfClass:[NSNull class]] || fragment == nil);
 }
 
++ (NSString*)stackFromException:(TiContextRef)context exception:(TiValueRef)exception
+{
+  TiObjectRef exceptionObj = TiValueToObject(context, exception, NULL);
+  TiStringRef stackProperty = TiStringCreateWithUTF8CString("stack");
+  TiValueRef stackValue = JSObjectGetProperty(context, exceptionObj, stackProperty, NULL);
```
hansemannn: Use `TiObjectGetProperty` to make it usable with TiCore.

iphone/Classes/TiUtils.h
```diff
@@ -730,4 +730,6 @@ typedef enum {
  */
 + (UIImage *)imageWithColor:(UIColor *)color;
 
++ (NSString*)stackFromException:(TiContextRef)context exception:(TiValueRef)exception;
```
hansemannn: Please format the source with `clang-format -style=file -i Classes/TiUtils.*` and for other related files. Only necessary at the end :-).

iphone/Classes/TiExceptionHandler.h
```diff
@@ -24,6 +24,11 @@
  */
 @property (nonatomic, readonly) NSInteger lineNo;
 
+/**
+ * Returns line column where error happened.
+ */
+@property (nonatomic, readonly) NSInteger column;
```
hansemannn: If the column cannot be < 0, you can use `NSUInteger` to be more precise.

iphone/Classes/TiErrorController.m
```diff
@@ -42,243 +35,22 @@ - (void)loadView
...
-  [[self view] addConstraint:centerConstraint4];
-  [[self view] addConstraint:centerConstraint5];
-  [[self view] addConstraint:centerConstraint6];
-  [[self view] addConstraint:disclosureContraint1];
-  [[self view] addConstraint:disclosureContraint2];
-  [[self view] addConstraint:disclosureContraint3];
-
-  [dismissButton addTarget:self action:@selector(dismiss:) forControlEvents:UIControlEventTouchUpInside];
+  
+  UIScrollView* view = [[[NSBundle mainBundle] loadNibNamed:@"ErrorScreen" owner:self options:nil] objectAtIndex:0];
```
hansemannn: Are we using an xib for convenience? Usually we try to add the AutoLayout code programmatically to have more control over it, but I understand that this may be critical for more complex layouts.

hansemannn (changes requested):
Just a few changes for now. Really really great work @garymathews!

hansemannn:
I didn‘t forget you here! 🙂 Pushing the Autolayout changes affer GDPR.

hansemannn (approved):

garymathews:
@hansemannn Awesome! Updated PR 🎉

sgtcoolguy:
Forgive me if I'm missing something here, but it looks like we show the native stack in the modal error dialog, but don't set it on `Error` objects in JS-land now?

I think we should be generating the native stack lower-level where ever it is we surface the `Error` objects to JS-world so that the property's value can be read from JS too.

Specifically, I'd like to see this unit test suite tweaked slightly to support accessing nativeStack on both iOS and Android: https://github.com/appcelerator/titanium_mobile/pull/10003/files#diff-3dc400bf7f83b2691685b65f5a5066d1R13

There we added tests to verify 3 typical scenarios:
- A JS error occurred, should be a typical `Error` object with `message` and `stack` properties
- A native error occurred, should be an `Error` object with `message` and `stack` properties, but also a `nativeStack` property. This particular trigger case for Android doesn't cause an issue on iOS, so we'd likely need to some up with some other example where type coercion doesn't get handled properly and causes a native exception in Obj-C.
- throwing a primitive like a string literal in JS, the caught object is the string, not an `Error`

I think e should be able to enable this test on both platforms and have them each pass. Clearly the nativeStack will differ on Android versus iOS, and maybe the triggering code for causing a native error. But the general type/property checks should apply.

hansemannn:
@sgtcoolguy A `message` and `stack` is already included:
```
2018-05-22 23:15:05.481292+0200 Titanium[2312:16120] [ERROR] {
    column = 17;
    line = 20;
    message = "undefined is not an object (evaluating 'obj.test.crash')";
    sourceURL = "file:///Users/hknoechel/Library/Developer/CoreSimulator/Devices/C02D33D2-92F2-490F-A2A8-D978B36B0626/data/Containers/Bundle/Application/C3F8BB88-5231-4F2B-9603-284BF11BE457/Titanium.app/app.js";
    stack = "file:///Users/hknoechel/Library/Developer/CoreSimulator/Devices/C02D33D2-92F2-490F-A2A8-D978B36B0626/data/Containers/Bundle/Application/C3F8BB88-5231-4F2B-9603-284BF11BE457/Titanium.app/app.js:20:17";
}
```
but they are named differently.

For native errors, we include `nativeLocation` and `nativeReason` as well. Should we remove those for iOS?

hansemannn:
@sgtcoolguy @garymathews [Unit-test](https://github.com/garymathews/titanium_mobile/pull/2/files#diff-3dc400bf7f83b2691685b65f5a5066d1R13) added as well. I have removed the "contains string" checks to be less error prone, especially because Windows (cc @infosia) will come as well.

iphone/Classes/TiExceptionHandler.m
```diff
@@ -56,16 +57,29 @@ - (void)reportScriptError:(TiScriptError *)scriptError
...
+  NSArray<NSString *> *exceptionStackTrace = [NSThread callStackSymbols];
+
+  if (exceptionStackTrace == nil) {
+    [[TiApp app] showModalError:[error description]];
+  } else {
+    NSMutableArray<NSString *> *formattedStackTrace = [[[NSMutableArray alloc] init] autorelease];
+    NSUInteger exceptionStackTraceLength = [exceptionStackTrace count];
+
+    // re-size stack trace and format results. Starting at index = 4 to not include the script-error API's
+    for (NSInteger i = 4; i <= (exceptionStackTraceLength >= 20 ? 20 : exceptionStackTraceLength); i++) {
```
hansemannn: @garymathews This is causing an array out of bound exception:
```js
    backgroundColor: '#fff',
    homeIndicatorAutoHidden : true
});
 
var scrollable = Ti.UI.createScrollableView({
});

win.add(scrollable);
win.open();

setTimeout(function () {
    scrollable.setViews([-1]);
}, 4000);
```
We should probably do `i < xxxxx`?

PR #10028, opened 2018-05-04
[TIMOB-26022] TiAPI: Improve Ti.Buffer parity

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26022

**Description:**
- Cherry-picks the fix from #9966 necessary to have the errors thrown.
- Updates the Ti.Buffer tests to be more comprehensive
- Updates the iOS implementation to be more careful about verifying offsets/lengths of buffers 
  - don't just check that the start offset is past the end of buffer or length is too long, but that the combination is OK
  - Don't allow some methods to be ok with extending past the end of a Buffer and only doing partial operations, whereas others would be rigid in bounds checking. Be consistent in bounds checking.
- When cloning a Buffer retain the byteOrder/type properties. If cloning a Buffer wholesale, retain the value property too.


hansemannn (approved):
The change looks valid to me. For functional testing, I trust on the unit tests, since I did not work with buffers in deep so far. Let me know if you need an additional review in that case!

PR #10030, opened 2018-05-07
[TIMOB-26023] TiAPI: Improve Ti.Blob parity

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26023

**Description:**

**Cross-platform**
- Updates the `Ti.Blob` unit test suite to be more comprehensive

**Android**
- Adds `Ti.Blob.size` property and getter for Android to match other platforms.
- Fixes an NPE when calling `Ti.Blob#imageAsCompressed()` that I assume meant this method never worked before.

**iOS**
- Adds the `Ti.Blob#append(Ti.Blob)` method.
- "Normalizes" input images before applying many of the `imageAs` operations. In particular, with an 8-bit PNG with alpha as input file for the blob, we'd typically get a null CGBitmapContextCreate under the hood and fail to run the image op we wanted. This normalizes to a known good combination of bits per channel/color space/alpha flags on iOS to avoid that issue.
- Updates some of the 3rd-party code used by the image APIs to handle the scale value for images

** TODOs**
- [x] Settle on proper behavior expected for the failing tests. 
  - In particular, iOS' `Ti.Blob#imageAsThumbnail()` returns the given size + the border size (and border size defaults to 1: so passing in 50 as size gives us a 52x52 image)
  - Android's `Ti.Blob#imageWithRoundedCorner()` adds the border to the image size (and again this uses a default border of 1 px), growing the image whereas other platforms do not

This is part of my series of API parity PRs:
- #10028
- #10019

### DIFF
--- a/apidoc/Titanium/Blob.yml
+++ b/apidoc/Titanium/Blob.yml
@@ -90,7 +90,8 @@ properties:
     description: |
         If this blob represents an image, this is the total number of pixels in the image.
         Otherwise it returns the number of bytes in the binary data.
-    platforms: [iphone, ipad]
+    platforms: [android, iphone, ipad]
+    since: {android: "7.2.0"}
     permission: read-only
 
 methods:
--- a/apidoc/Titanium/Blob.yml
+++ b/apidoc/Titanium/Blob.yml
@@ -91,7 +91,7 @@ properties:
         If this blob represents an image, this is the total number of pixels in the image.
         Otherwise it returns the number of bytes in the binary data.
     platforms: [android, iphone, ipad]
-    since: {android: "7.2.0"}
+    since: {android: "7.2.0", iphone: "0.9.0", ipad: "0.9.0"}
     permission: read-only
 
 methods:
--- a/apidoc/Titanium/Blob.yml
+++ b/apidoc/Titanium/Blob.yml
@@ -166,6 +166,9 @@ methods:
         Returns the thumbnail image as a blob.
 
         If this blob doesn't represent an image, returns `null`.
+
+        The final height/width of the image will actually be `size + (2 * borderSize)` as the border is added around the image.
+        By default the `borderSize` is `1`.
     platforms: [android, iphone, ipad]
     since: {android: "3.0.0"}
     returns:
@@ -203,6 +206,8 @@ methods:
     summary: Returns a copy of the underlying image with rounded corners added.
     description: |
         Returns the new image as a blob, or `null` if this blob is not an image.
+        The image will grow in height and width by `(2 * borderSize)` as the border is added around the image to avoid scaling.
+        By default the `borderSize` is `1`.
     platforms: [android, iphone, ipad]
     since: {android: "3.0.0"}
     parameters:
@@ -222,6 +227,7 @@ methods:
     summary: Returns a copy of the underlying image with an added transparent border.
     description: |
         Returns the new image as a blob, or `null` if this blob is not an image.
+        The image will grow in height and width by `(2 * borderSize)` as the border is added around the image to avoid scaling.
     platforms: [android, iphone, ipad]
     since: {android: "3.0.0"}
     parameters:
--- a/tests/Resources/ti.blob.test.js
+++ b/tests/Resources/ti.blob.test.js
@@ -1,0 +1,312 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions'),
+	utilities = require('./utilities/utilities');
+
+describe.only('Titanium.Blob', function () {
+	var win;
+
+	afterEach(function () {
+		if (win) {
+			win.close();
+		}
+		win = null;
+	});
+
+	it('.apiName', function () {
+		// FIXME Should be able to do Ti.Blob.apiName
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob).have.a.readOnlyProperty('apiName').which.is.a.String;
+		should(blob.apiName).be.eql('Ti.Blob');
+	});
+
+	it('constructed from File.read()', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob).be.an.Object;
+		// should(blob).be.an.instanceof(Ti.Blob); // Crashes Windows, throws uncaught error on iOS & Android
+	});
+
+	// Windows crashes on instanceof check TIMOB-25012
+	// Windows also crashes if we uncomment this now, I think closing the window (or failing the test) in the blob callback is causing Desktop crash
+	// Android is sometimes timing out... Trying an open event now...
+	// TODO: Test is tempermental, skipping for now...
+	it('constructed from image', function (finish) {
+		var label;
+		win = Ti.UI.createWindow();
+		label = Ti.UI.createLabel({ text: 'test' });
+		win.add(label);
+		win.addEventListener('open', function () {
+			label.toImage(function (blob) {
+				should(blob).be.an.Object;
+				// should(blob).be.an.instanceof(Ti.Blob); // FIXME Crashes Windows, throws uncaught error on iOS & Android
+				// should(blob.getText()).equal(null); // FIXME 'blob.getText is not a function' on iOS
+				// should(blob.text).equal(null); // FIXME this is undefined on iOS, docs say it should be null
+				should(blob.text).not.exist;
+				Ti.API.info(blob.width);
+				should(blob.width).be.a.Number; // FIXME Undefined on iOS
+				should(blob.width).be.above(0); // 0 on Windows
+				should(blob.height).be.a.Number;
+				should(blob.height).be.above(0);
+				should(blob.length).be.a.Number;
+				// FIXME Parity issue, no size property on Android
+				if (!utilities.isAndroid()) {
+					should(blob.size).be.a.Number;
+					should(blob.size).equal(blob.width * blob.height);
+				}
+				finish();
+			});
+		});
+		win.open();
+	});
+
+	it('.text', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob.text).be.a.String;
+		should(blob.text.length).be.above(0);
+		should(blob.text).equal(blob.toString());
+		// TODO Test that it's read-only
+	});
+
+	// FIXME Add to iOS API
+	it.iosMissing('#append()', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob.append).be.a.Function;
+		// TODO Test actually appending data to it
+	});
+
+	it('.nativePath', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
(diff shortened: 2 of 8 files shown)


sgtcoolguy:
I'm also trying to manually inspect some operations to see how the platforms differ/why:

Here's some results...

Thumbnail of a 150x150 PNG of Spongebob down to 50px:

iOS | Android
------------ | -------------
 ios_thumbnail | android_thumbnail

Clearly there's a quality difference there. iOS is way better. But notably, iOS is actually a 52x52px image, with a 1px transparent border around it. Android has no border, despite the API and code seeming to default to a 1px border.

sgtcoolguy:
I'm also trying to manually inspect some operations to see how the platforms differ/why:

Here's some results...

150x150 PNG of Spongebob with rounded corner of 4px:

Original | iOS | Android
------------ | ------------ | -------------
logo | ios_rounded_corner | android_rounded_corner

Here, Android is actually generating a 152x152 px image (the original is 150x150). Android's rounded corners are not anti-aliased. iOS anti-aliases the corners, making it look much nicer, and keeps the original image size of 150x150 - but appears to be cropping part of the original image to use for the border. I'll have to test with larger corner size / border values to see.

sgtcoolguy:
300x150 PNG as a 50px thumbnail.

I was curious to see how it handled the input value to determine how to crop a non-square image. First with a wider image.

Original | iOS | Android
------------ | ------------ | -------------
300x150 | ios_300x150_thumbnail | android_300x150_thumbnail

Again iOS trounces Android in quality.
They both appear to resize based on height, then crop down.

sgtcoolguy:
150x300 PNG as a 50px thumbnail.

I was curious to see how it handled the input value to determine how to crop a non-square image. Now with a taller image.

Original | iOS | Android
------------ | ------------ | -------------
150x300 |  ios_150x300_thumbnail |  android_150x300_thumbnail

Again iOS trounces Android in quality. In fact, Android's is now illegible.

Interesting to compare versus the wider image above. The text size is much smaller now, which appears to mean they're again resizing based on height and then cropping?

I would think given that we only allow a single input that we should be resizing based upon the smaller of the width/height in the original and then cropping. Certainly it is consistent now on using the height as the resizing factor, but I'm not sure the result is what you'd expect (seems "spatially" inconsistent when comparing the resulting images).

sgtcoolguy:
@jquick-axway @hansemannn @garymathews This PR will need someone to do a CR when I'm done, but I'm tagging you guys to get your feedback on the actual image results in my follow-on comments here.

The platforms differ in how they generate images based on the same input image and values. And are actually logically inconsistent between different methods - i.e. `#imageAsThumbnail()` on iOS adds border around the resized image, but `#imageWithRoundedCorner()` lays the border over top the original image (essentially cropping)

I have some opinions on what's the more correct way to do things, but obviously would like input.

jquick-axway:
Regarding the border handling differences between Android and iOS, I agree with how Android is doing it because it respects the given width/height of the view. That is, the border does not extend outside of the view assigned width/height. The border is inset and shrinks the content are to fit within the border. This works well when the view's width/height is set to `Ti.UI.FILL` within another view.

iOS on the other hand sizes the content area based on the given width/height and applies a border on top. The border is "mostly" centered around the edges of the view's rectangle. I say mostly because the border is typically slightly shifted towards the top-left corner (integer rounding issue?). And the problem with centering the border around the view is that the outer part of the border gets clipped when the view's width/height is set to `Ti.UI.FILL`. And the outer border will extend into other views in a vertical/horizontal layout with no padding between views.

---
Regarding Android's thumbnail image quality issue, let's just write it up as a bug for now. Our Android image loading code needs a major refactoring. In this case, I suspect the view is trying get a density scaled drawable of the blob via our `TiDrawableReference.getDensityScaledDrawable()`... and that method does something bad with non-"res" image references where it creates a new bitmap scaled up (when it doesn't need to; wastes RAM) and then returns a drawable scaled down back to the original size (which achieves nothing in the end). The correct solution is to load the bitmap as-is and return a scaled drawable instead (effectively stretching the image over a larger polygon). This is something I would like to address in the future.

sgtcoolguy:
@jquick-axway Just to clarify here: These images are being generated entirely outside the `Ti.UI.View` code. This is specifically what happens when we load a PNG file from the app via `Ti.Filesystem.getFile()` and `read()` it into a `Ti.Blob` and then try to run the various image manipulation methods on the blob and write the result back to disk (which I'm then pulling out of the emulator).

I think in general I'm leaning towards:
- transparent borders should be "added" to the image regardless of which method we're calling. This avoids needing to resize/resample the interior image down (again) to fit the border without getting it cropped. Depending on which method you call, the platforms behave differently here. The options would be to:
  - get the image to the size we want and then add the border around that, so a 50px image with a 2px border is actually 54px. (my preference, and what Android is doing for the rounded corner)
  - Lay the border over the resized image, so a 50px thumbnail is a resized image down to 50px and then the 2px border would be laid on top (cropping the image by 2px all around, which is effectively what iOS is doing on the rounded corner example above with the larger border)
  - resize the image to the final size minus 2*border, then add border around that (so the image would be resized to 46px and then surrounded by a 2px border to get the desired final dimensions)
- We should eliminate the default 1px borders on image manipulations. Why would we default to behavior where asking for a 50px thumbnail gives you a 52px thumbnail with a 1px border? (or a 50px image with 1px border removed? or a 50px image resized down to 48x with a 1px border around it?)

sgtcoolguy:
So here's an example tweaking how iOS behaves for `#imageWithRoundedCorner(25, 20)`.

Original | Pre-PR behavior | With tweak to be like Android
--------- | ---------------- | -----------------------------
logo | ios_rounded_corner25_border20 | ios_tweak


Here the final image is actually 190x190: The 150x150 original surrounded by a 20px border.

I think that's a much more sane alternative, so I'm gonna push that commit to this PR for now (can always revert if people disagree).

jquick-axway:
@sgtcoolguy, just to be clear, you're saying your JavaScript code is not displaying the generated thumbnail image via an ImageView? The resulting thumbnail images you're posting here are the resulting blob thumbnails saved to file?

If so, then that sounds like a scaling issue with Google's `ThumbnailUtils` class. We may need to create the thumbnail ourselves. From looking at Google's code [here](https://github.com/aosp-mirror/platform_frameworks_base/blob/master/media/java/android/media/ThumbnailUtils.java#L357), I can see that they "scale" the image if the requested thumbnail's width/height is larger than the original bitmap (`Canvas.drawBitmap()` call is not using [FILTER_BITMAP_FLAG](https://developer.android.com/reference/android/graphics/Paint.html#FILTER_BITMAP_FLAG) flag) or if the aspect ratios don't line-up (this last one doesn't make sense to me; might be the problem here even though they enable filtering in this case). The thumbnail should never be scaled unless the given width/height is larger than original image.

Here's a good test for you. What happens on iOS if you set the thumbnail width/height larger than the original image? Is it scaled up or is it centered?

apidoc/Titanium/Blob.yml
```diff
@@ -90,7 +90,8 @@ properties:
     description: |
         If this blob represents an image, this is the total number of pixels in the image.
         Otherwise it returns the number of bytes in the binary data.
-    platforms: [iphone, ipad]
+    platforms: [android, iphone, ipad]
+    since: {android: "7.2.0"}
```
hansemannn: Can the "0.9.0" for iOS be omitted these?

tests/Resources/ti.blob.test.js
```diff
@@ -0,0 +1,337 @@
...
+	it('constructed from File.read()', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob).be.an.Object;
+		// should(blob).be.an.instanceof(Ti.Blob); // Crashes Windows, throws uncaught error on iOS & Android
+	});
+
+	// Windows crashes on instanceof check TIMOB-25012
+	// Windows also crashes if we uncomment this now, I think closing the window (or failing the test) in the blob callback is causing Desktop crash
+	// Android is sometimes timing out... Trying an open event now...
+	// TODO: Test is tempermental, skipping for now...
```
hansemannn: Can be removed as it seems like it's not being skipped (anymore).

tests/Resources/ti.blob.test.js
```diff
@@ -0,0 +1,337 @@
...
+				// should(blob.text).equal(null); // FIXME this is undefined on iOS, docs say it should be null
+				should(blob.text).not.exist;
+				Ti.API.info(blob.width);
+				should(blob.width).be.a.Number; // FIXME Undefined on iOS
+				should(blob.width).be.above(0); // 0 on Windows
+				should(blob.height).be.a.Number;
+				should(blob.height).be.above(0);
+				should(blob.length).be.a.Number;
+				// FIXME Parity issue, no size property on Android
+				if (!utilities.isAndroid()) {
```
hansemannn: Could this be addressed as part of this PR? I think we receive the size internally already, so a simple `@Kroll.getProperty` may be able to clear this.

tests/Resources/ti.blob.test.js
```diff
@@ -0,0 +1,337 @@
...
+	it('.text', function () {
+		var blob = Ti.Filesystem.getFile('app.js').read();
+		should(blob.text).be.a.String;
+		should(blob.text.length).be.above(0);
+		should(blob.text).equal(blob.toString());
+		// TODO Test that it's read-only
+	});
+
+	// FIXME Add to iOS API
+	it.iosMissing('#append()', function () {
```
hansemannn: I could expose that. Maybe an own ticket?

tests/Resources/ti.blob.test.js
hansemannn: Oh, you exposed it already :-). Then the guard can be removed.

sgtcoolguy:
@jquick-axway Correct - this is purely programmatic manipulation using `Ti.Filesystem.File` and `Ti.Blob` APIs - no `ImageView` or screencaps here.

As for iOS and Android scaling images up above their original size (via `Ti.Blob#imageAsThumbnail(400)`), they appear to be the same to me:

iOS | Android
------| --------
ios_logo_to400 | android_logo_to400
ios_150x300_to400 | android_150x300_to400
ios_300x150_to400 |  android_300x150_to400

tests/Resources/ti.blob.test.js
sgtcoolguy: Implemented this and updated the test in a subsequent commit.

tests/Resources/ti.blob.test.js
sgtcoolguy: I uncommented/unskipped all tests to see if I could fix them all. This one still needs to be skipped as Android times out.

apidoc/Titanium/Blob.yml
sgtcoolguy: There's a default value if unspecified. It's defined in the doc generator, and I think it's actually 0.8.0. I assumed you could omit and it would use the default - but maybe the doc generator needs to be tweaked to allow that?

tests/Resources/ti.blob.test.js
jquick-axway: @sgtcoolguy, I suspect the Android timeout is caused by opening a window while a previous windows is closing. Not sure why yet (haven't had time to investigate) but doing the following in the `aftereach()` call resolves it for me. Trouble is you would have to implement this in all unit test files.

```javascript
afterEach(function (finish) {
	if (win) {
		win.addEventListener("close", function () {
			finish();
		});
		win.close();
		win = null;
	} else {
		finish();
	}
});
```

sgtcoolguy:
Bumping to 8.0.0 - while not a breaking api change in terms of JS, the changes to how iOS renders the resulting images is certainly a breaking behavioral change.

PR #10031, opened 2018-05-08
[TIMOB-25231] Android: List encrypted assets

garymathews:
- List encrypted assets when using `getDirectoryListing()`

###### TEST CASE BY @jquick-axway
```JS
function logFileTree (file) {
    if (file) {
        Ti.API.info('- ' + file.nativePath + ' (directory: ' + file.isDirectory() + ')');
        let files = file.getDirectoryListing();
        if (files) {
            for (let f of files) {
                logFileTree(Ti.Filesystem.getFile(file.nativePath, f));
            }
        }
    }
}
logFileTree(Ti.Filesystem.getFile(Ti.Filesystem.getResourcesDirectory()));
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25231)

### DIFF
--- a/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
+++ b/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
@@ -26,6 +26,7 @@ public class KrollAssetHelper
 
 	public interface AssetCrypt {
 		String readAsset(String path);
+		String[] getAssets();
 	}
 
 	public static void setAssetCrypt(AssetCrypt assetCrypt)
@@ -74,6 +75,14 @@ public class KrollAssetHelper
 			Log.e(TAG, "Error while reading asset \"" + path + "\":", e);
 		}
 
+		return null;
+	}
+
+	public static String[] getEncryptedAssets()
+	{
+		if (assetCrypt != null) {
+			return assetCrypt.getAssets();
+		}
 		return null;
 	}
 
--- a/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
+++ b/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
@@ -26,7 +26,7 @@ public class KrollAssetHelper
 
 	public interface AssetCrypt {
 		String readAsset(String path);
-		String[] getAssets();
+		String[] getAssetPaths();
 	}
 
 	public static void setAssetCrypt(AssetCrypt assetCrypt)
@@ -78,10 +78,10 @@ public class KrollAssetHelper
 		return null;
 	}
 
-	public static String[] getEncryptedAssets()
+	public static String[] getEncryptedAssetPaths()
 	{
 		if (assetCrypt != null) {
-			return assetCrypt.getAssets();
+			return assetCrypt.getAssetPaths();
 		}
 		return null;
 	}
--- a/android/templates/build/AssetCryptImpl.java
+++ b/android/templates/build/AssetCryptImpl.java
@@ -48,6 +48,10 @@ public class AssetCryptImpl implements KrollAssetHelper.AssetCrypt
 		return new String(filterDataInRange(assetsBytes, range.offset, range.length));
 	}
 
+	public String[] getAssets() {
+		return assets.keySet().toArray(new String[assets.size()]);
+	}
+
 	private static byte[] filterDataInRange(byte[] data, int offset, int length)
 	{
 		try {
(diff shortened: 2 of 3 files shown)


jquick-axway:
Notes:
- The `File.getDirectoryListing()` returns an array of file names and subdirectory names. This array must not contain absolute paths. They are relative to the `File` object that the `getDirectoryListing()` method was called upon.
- The `File.nativePath` property must return an absolute path. It cannot be relative.
- I've verified that "Live View" builds correctly list JS files from the Resources directory on both Android and iOS. This PR fixes the JS file listing issue for non-LiveView builds on Android.

jquick-axway:
@garymathews, I have one minor nitpick. I think the`getAssets()` and `getEncryptedAssets()` methods should be renamed to either `getAssetPaths()` or `getPaths()` since it returns an array of paths and not the asset data itself. This makes it clear on the caller side what is actually returned.

The rest of the code looks fine. 👍

sgtcoolguy:
I would assume we should have some way of being able to unit test this. It may require changes to the test suite code itself to handle adding assets to encrypt into the test project, but then we should be able to add a test to the Ti.Filesystem.File suite to ensure the listing contains the encrypted assets.

jquick-axway:
@sgtcoolguy,

For your info, our mocha test suite uses the `Ti.Filesystem.getDirectoryListing()` method to find "addon" JS files. See the "app.js" file's `loadAddonTestFiles()` method here...
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/app.js#L117

This will currently fail to find "addon" JS files for Android builds... unless you do a LiveView build. Gary's fix resolves this issue.

jquick-axway:
@garymathews,

I wrote up a quick unit test for you. This should replace the `ti.filesystem.file.test.js` file's "#directoryListing()" unit test. I have verified that the below works/passes on iOS. I have not tested it with your PR.

```javascript
it('#directoryListing()', function () {
	var rootDir = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory),
		rootPath,
		filesFound = {};
	should(rootDir.exists()).be.true;
	should(rootDir.getDirectoryListing).be.a.Function;
	should(rootDir.getDirectoryListing()).be.an.Array;

	// Traverse entire Resources directory tree looking for files/directories in "filesFound".
	rootPath = rootDir.nativePath;
	filesFound[rootPath + 'app.js'] = false;
	filesFound[rootPath + 'ti.ui.webview.test.html'] = false;
	filesFound[rootPath + 'fixtures/'] = false; // Subdirectory containing only JS files.
	filesFound[rootPath + 'fixtures/empty-double.js'] = false;
	filesFound[rootPath + 'txtFiles/'] = false; // Subdirectory containing only assets.
	filesFound[rootPath + 'txtFiles/text.txt'] = false;
	function searchFileTree(file) {
		if (file) {
			var fileList = file.getDirectoryListing();
			if (fileList) {
				for (var index = 0; index < fileList.length; index++) {
					var nextFile = Ti.Filesystem.getFile(file.nativePath, fileList[index]);
					if (nextFile) {
						var absolutePath = nextFile.nativePath;
						if (absolutePath in filesFound) {
							filesFound[absolutePath] = true;
						}
						searchFileTree(nextFile);
					}
				}
			}
		}
	}
	searchFileTree(rootDir);
	for (var key in filesFound) {
		Ti.API.info('Checking if found file: ' + key);
		should(filesFound[key]).be.true;
	}
});
```

Note that we need to double check the trailing slash `/` handling for the `nativePath` returned directories/subdirectories. iOS always returns them with a trailing `/`. We should try to be consistent.

jquick-axway:
Hmm... we must be doing a `--liveview` build when running our unit tests. Otherwise the "addontest" fetching code and the "ti.ui.webview.test.js" `data` unit test would fail on a normal build. I think this explains the differences we're seeing when we run the unit tests ourselves.

jquick-axway:
The only discrepancy we have left is with the `nativePath` returned by directories. On Android, they do not end with a trailing slash `/`, but on iOS they do.

Although, perhaps this is not a major issue. If you do the below, then all platforms will properly append the file to the directory path and inject a slash if needed.
```javascript
var file = Ti.Filesystem.getFile(directory.nativePath, 'MyFile.txt');
```

lokeshchdhry:
FR Passed.

All the encrypted assets are listed as expected for device builds.

OS Ver:         10.13.4
Xcode Ver:      Xcode 9.3.1
Appc NPM:       4.2.13
Appc CLI:       7.0.3
Daemon Ver:     1.1.1
Ti CLI Ver:     5.1.0
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10

PR #10032, opened 2018-05-09
[TIMOB-24080] : iOS - Paragraph style support in attributed string

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24080


iphone/Classes/UIModule.m
```diff
@@ -594,6 +594,26 @@ - (NSNumber *)ATTRIBUTE_LINE_BREAK_BY_TRUNCATING_MIDDLE
 {
   return NUMINTEGER(NSLineBreakByTruncatingMiddle);
 }
+- (NSNumber *)ATTRIBUTE_TEXT_ALIGNMENT_LEFT
+{
+  return NUMINTEGER(NSTextAlignmentLeft);
```
hansemannn: We already support these on the UIModule class via `Ti.UI.TEXT_ALIGNMENT_*`

iphone/Classes/UIModule.m
```diff
@@ -504,7 +504,7 @@ - (NSNumber *)convertUnits:(id)args
 MAKE_SYSTEM_PROP(ATTRIBUTE_STRIKETHROUGH_COLOR, AttributeNameStrikethroughColor);
 MAKE_SYSTEM_PROP(ATTRIBUTE_OBLIQUENESS, AttributeNameObliqueness);
 MAKE_SYSTEM_PROP(ATTRIBUTE_EXPANSION, AttributeNameExpansion);
-MAKE_SYSTEM_PROP(ATTRIBUTE_LINE_BREAK, AttributeNameLineBreak);
+MAKE_SYSTEM_PROP(ATTRIBUTE_LINE_BREAK, AttributeNameLineBreak); // deprecated
```
hansemannn: It looks like we only warn the user via code so far, not via docs. Please add.

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+      (for example, the left margin in left-to-right text). If 0 or negative, 
+      it is the distance from the trailing margin. This is a float value.
+    type: Number
+
+  - name: lineBreakMode
+    summary: The mode that should be used to break lines in the receiver.
+    description: |
+      This property contains the line break mode to be used laying out the text of paragraph.
+    type: Number
+    constants: [ Titanium.UI.ATTRIBUTE_LINE_BREAK_* ]
```
hansemannn: I thought this is deprecated?

hansemannn (changes requested):
Some minor issues.

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+      This value is always nonnegative float value.
+    type: Number
+
+  - name: tailIndent
+    summary: The trailing indentation of the receiver.
+    description: |
+      If positive, this value is the distance from the leading margin 
+      (for example, the left margin in left-to-right text). If 0 or negative, 
+      it is the distance from the trailing margin. This is a float value.
+    type: Number
```
jquick-axway: We need to document what units `tailIndex` uses, such as points.

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+      This property contains the minimum height in points that any line in the receiver will occupy, 
+      regardless of the font size or size of any attached graphic. This value must be nonnegative.
+    type: Number
+
+  - name: lineSpacing
+    summary: The distance in points between the bottom of one line fragment and the top of the next.
+    description: |
+      This value is always nonnegative. This value is included in the line fragment heights 
+      in the layout manager.
+    type: Number
```
jquick-axway: We need to document the `lineSpacing` units, such as points.

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+    summary: |
+      A Boolean value indicating whether the system may tighten inter-character spacing 
+      before truncating text.
+    description: |
+      When this property is set to YES, the system tries to reduce the space between characters 
+      before truncating characters. The system performs this tightening in cases where the text 
+      would not otherwise fit in the available space. The maximum amount of tightening performed 
+      by the system is dependent on the font, line width, and other factors. The default value of 
+      this property is NO.
+    type: Number
```
jquick-axway: The `allowsDefaultTighteningForTruncation` type should be boolean, not number.

And the description should use `true` and `false` instead of the Objective-C `YES` and `NO`.

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+
+  - name: paragraphSpacing
+    summary: The space after the end of the paragraph.
+    description: |
+      This property contains the space (measured in points) added at the end of the paragraph 
+      to separate it from the following paragraph. This value must be nonnegative. The space 
+      between paragraphs is determined by adding the paragraphSpacing of previous paragraph and 
+      the paragraphSpacingBefore of current paragraph.
+    type: Number
```
jquick-axway: Perhaps we should rename property `paragraphSpacing` to `paragraphSpacingAfter`. This makes it clear that it's only applied after the paragraph (assuming this is actually how it works). And it clearly separates it from the `paragraphSpacingBefore` property.

jquick-axway (changes requested):

jquick-axway:
@vijaysingh-axway, @hansemannn, it's kind of a bummer that we're hard-coding properties such as `lineSpacing` to points/dips instead of respecting the default units set for the app. This can make view layout/alignment with text a pain.

To be fair, I've noticed that our existing `ATTRIBUTE_BASELINE_OFFSET` is hard-coded to pixels and `ATTRIBUTE_KERN` is hard-coded to points. So, it appears to be a pre-existing issue with attributed strings, but do we really want to continue this trend? Perhaps allowing these properties to be set to a string where the dev can set the units might be better? (If you feel that it's too late, then fair enough.)

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,119 @@ properties:
...
+      the beginning of its text content. The default value of this property is 0.0.
+    type: Number
+
+  - name: lineHeightMultiple
+    summary: The line height multiple.
+    description: |
+      The natural line height of the receiver is multiplied by this factor (if positive) 
+      before being constrained by minimum and maximum line height. The default value of 
+      this property is 0.0.
+    type: Number
```
jquick-axway: The docs here don't do a good job describing what `lineHeightMultiple`, `maximumLineHeight`, and `minimumLineHeight` do. Nor why you would want to use them.

I assume you would want to use `minimumLineHeight` in case you use mixed font sizes and you want all lines to have the same height as the largest font size. That's the only use-case I can think of.

But why would you want to use `maximumLineHeight`? Would it clip text that exceed this height? If there is answer to this, then we should document it.

Also, I have no clue what `lineHeightMultiple` is supposed to do from reading the description... nor why I would want to use it. If there is a valid use-case, then let's document it.

hansemannn:
I agree! There is a helper for convertig values to dimensions. Vijay, can you address that?

apidoc/Titanium/UI/Attribute.yml
jquick-axway: Also, I assume a value of zero for `maximumLineHeight` means do not apply a max, right? Otherwise text would always be clipped. We should document that.

I assume the same is true for `lineHeightMultiple`? A value of zero means don't multiply?

hansemannn:
@vijaysingh-axway Can you check the review comments? Moving to 7.3.0 for now, due to lower priority.

jquick-axway:
Note that I'm not trying to picky. These details matter when we implement this on Android for parity. Documenting these details not only benefits Titanium devs, but it also serves as a specification when implementing this feature on other platforms. Thanks!

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -115,3 +124,124 @@ properties:
...
+      This property contains the space (measured in points) between the top of paragraph and 
+      the beginning of its text content. The default value of this property is 0.0.
+    type: Number
+
+  - name: lineHeightMultiple
+    summary: The line height multiple.
+    description: |
+      The natural line height of the receiver is multiplied by this factor (if positive) 
+      before being constrained by minimum and maximum line height. The default value of 
+      this property is 0.0, which means nothing multipied.
```
jquick-axway: Typo. "multipied" should be "multiplied". (Missing an 'L'.)

hansemannn:
@vijaysingh-axway Except the comment from Josh, the CR looks fine now. But I am curious if there is really **no** Android parity possible here. Can you please discuss this with @ypbnv and see if we can support at least a few of these properties on Android?

jquick-axway:
@hansemannn, @vijaysingh-axway, I've already written up a ticket for Android parity here: [TIMOB-26042](https://jira.appcelerator.org/browse/TIMOB-26042)

I don't know if we'll have time to do it in 7.4.0, but at least we've got it on our to-do list.

vijaysingh-axway:
@hansemannn Fixed the dimension conversion. Now you can review. Thanks!

hansemannn:
@vijaysingh-axway You still should update the docs to reflect that developers can also use non-Number types.

hansemannn (approved):
CR **passed**!

hansemannn:
@jquick-axway Can you re-review / approve this?

jquick-axway (approved):
CR: Pass

keerthi1032:
FR passed. Attributed string Works fine on IOS.
Test Environment:
  # CPUs                      = 8
  Memory                      = 17179869184
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = 7.5.0 local sdk 
iPhone 6s -ios 12
iPhone 5 -ios 11 simulator

PR #10033, opened 2018-05-09
[TIMOB-26021] iOS: Search Bar- search results tableview background color change

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26021


hansemannn (approved):
CR passed. Can be put in QE-testing once unit-tests are added.

PR #10035, opened 2018-05-09
Updated Blob.yml

jawa9000:
Fixed issues raised in https://jira.appcelerator.org/browse/TIDOC-3188 for imageAsThumbnail method parameter.


hansemannn:
I don‘t think this is a proper solution and should be fixed in the docgen instead. Voting for closing this one, but open for another opinion.

sgtcoolguy:
Close in favor of a proper fix in #10036

PR #10036, opened 2018-05-10
[TIDOC-3188] Generated API docs do not display default values for properties/arguments

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-3188

**Description:**
This changes the template used for js duck output for our `apidoc/docgen.js` file.
- Clean up the template to avoid unnecessary whitespace/newlines
- Don't output @description custom tag (but append the value to summary) as it ended up just getting smushed into the summary anyways by jsduck.
- Extract partial templates for EJS for class/method/event/property/params.
- Do less massaging of the data structure used by EJS when it wasn't necessary.
- Use more standard syntax for denoting params/properties that allows us to inject the default values. Basically we now do:
```javascript
/**
 * @param {String} regularParam
 * @param {String} [optionalParam]
 * @param {Boolean} [withDefault=true]
 * @param {String} [withDefaultTwo=undefined]
 * @param {String} [withDefaultThree='']
 */
```
⚠️ Note that I did a quick manual test with `undefined`, `true`, `false`, and `''` to verify that we now output those properly in the resulting JSduck file. What I did notice, however, was that our yaml docs do not use the non-standard way to denote `undefined` in a way that it'll become an actual `undefined` value in the translation rather than a `'undefined'` string. See https://github.com/nodeca/js-yaml#supported-yaml-types

Some notes:
- I didn't try to fix up the apidocs that specified default values that weren't actual literal values. i.e. many state generically what the default value will be relative to some current state (for say UI sizes/positions), or state things like "No hint text" when the actual literal default value would be `undefined` or `''`. Additionally, some have a mix of literal value followed by explainer text.

This is a good place to look through the docs and square away what the specify with our actual behavior. For example, many properties may specify a default value of `undefined` with explainer text that it behaves like `false`, `true`, or some other value! If that's the case, we should probably default to those instead.


apidoc/lib/jsduck_generator.js
```diff
@@ -236,33 +236,32 @@ function exportType(api) {
...
-			optional += ' (optional)';
-		}
-		str += '{' +  member.type.join('/') + '} ' + platforms + member.name + optional + '\n';
-		str += exportSummary(member);
-		str += exportConstants(member);
-		rv.push(str);
+		// 	if ('platforms' in member) {
+		// 		platforms = ' (' + member.platforms.join(' ') + ') ';
+		// 	}
+		// TODO Append the platform list to the summary!
```
hansemannn: Do you want to address this in this PR?

apidoc/lib/jsduck_generator.js
```diff
@@ -421,7 +404,10 @@ function exportAPIs(api, type) {
 				case 'properties':
 					annotatedMember.availability = member.availability || null;
 					annotatedMember.constants = exportConstants(member);
-					annotatedMember.defaultValue = exportDefault(member);
+					// FIXME How can we handle setting empty string, false, or undefined as default values?
+					if ('default' in member) {
+						annotatedMember['default'] = member['default'];
```
hansemannn: I would do an own PR to fix up all these types by syncing with the native return types and write tickets for its parity issues. But I guess we should be fine with this one to start.

PR #10038, opened 2018-05-11
[TIMOB-26032] iOS/Android: Fix Ti.Contacts parity, deprecate old API's properly

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26032

`no tests` because it only refactors existing code. Also requires UI-related interaction (contacts permissions, selecting values from the modal dialog).


android/modules/contacts/src/java/ti/modules/titanium/contacts/ContactsModule.java
```diff
@@ -136,6 +136,13 @@ public void save(Object people)
 
 	@Kroll.method
 	public PersonProxy getPersonByID(long id)
+	{
+		Log.w(TAG, "getPersonByID has been deprecated in favor of getPersonById (parity)", Log.DEBUG_MODE);
```
sgtcoolguy:
```java
Log.w(TAG, "Ti.Contacts.getPersonByID() has been deprecated in favor of Ti.Contacts.getPersonByIdentifier() for cross-platform parity", Log.DEBUG_MODE);
```

sgtcoolguy (changes requested):
The deprecation note on Android has a type in the correct method name. Added suggested fixed text.

android/modules/contacts/src/java/ti/modules/titanium/contacts/ContactsModule.java
hansemannn: Updated!

vijaysingh-axway (approved):
CR passed.
Note - @hansemannn There seems linting issue in Contacts.java as build is failing. Can you fix that.

lokeshchdhry:
FR Passed.

PR #10040, opened 2018-05-13
[TIMOB-18069] Android: Add itemCount to ListSection

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18069

**example**
```javascript
	fullscreen: true
});
var counter = 0;

function genData() {
	var data = [];
	for (i = 1; i <= 10; i++) {
		data.push({
			properties: {
				title: 'ROW ' + (counter + i),
				searchableText: (i % 2) ? "a" : "b"
			}
		})
	}
	return data;
}
section.setItems(genData());

	sections: [section],
	height: Ti.UI.FILL,
	width: Ti.UI.FILL
});
win.add(listView);
	console.log("All data: " + section.getItemCount());
	_.delay(function() {
		listView.setSearchText("a");
		console.log("Filter count: " + section.itemCount);
	}, 200);
})
win.open();
```

The ticket says that `itemCount` is available on iOS already but I couldn't test it at the moment. So might need to change the apidoc


hansemannn:
We definitely need parity here before merging.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListSectionProxy.java
```diff
@@ -234,6 +234,15 @@ public String getFooterTitle()
 		return footerTitle;
 	}
 
+	// clang-format off
+	@Kroll.method
+	@Kroll.getProperty
+	public int itemCount()
```
hansemannn: Do we need this one as well? It looks incorrect to have no `get` as a prefix. Also the `getItemCount()` method already exists and could be extended by the `@Kroll.getProperty` annotation.

m1ga: let me check

m1ga: works fine! Remove the custom getter

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListSectionProxy.java
```diff
@@ -943,6 +934,7 @@ public int getContentCount()
 	 * @return number of entries within section
 	 */
 	@Kroll.method
+	@Kroll.getProperty
 	public int getItemCount()
```
hansemannn: clang-format specifiers?

hansemannn:
LGTM. @jquick-axway for another quick review. Change looks solid.

hansemannn:
@m1ga Please add an addon-unit-test as well.

jquick-axway:
Here is a unite test I've written up. I've verified that it works on iOS. The below needs to be added to a new file named `ti.ui.listview.addontest.js` under the `tests/Resources` directory.

Also note that Android will currently fail the below unit test since its `itemCount` is filtered when applying a search text. On iOS, it returns an unfiltered item count.

```javascript
/* eslint-env mocha */
/* global Ti */
'use strict';
var should = require('./utilities/assertions');

describe('Titanium.UI.ListView', function () {
	var win;
	this.timeout(20000);

	afterEach(function () {
		if (win) {
			win.close();
		}
		win = null;
	});

	it.windowsMissing('sectionItemCount', function (finish) {
		var listView = null,
			maxItemCount = 4;
		this.timeout(20000);

		function createSection() {
			var rows = [],
				index;
			for (index = 1; index <= maxItemCount; index++) {
				rows.push({
					properties: {
						title: 'Row ' + index,
						searchableText: (index % 2) ? 'a' : 'b'
					}
				});
			}
			return Ti.UI.createListSection({ items: rows });
		}

		function verifySectionItemCount(expectedItemCount) {
			var sections = listView.sections,
				index;
			for (index = 0; index < sections.length; index++) {
				should(sections[index].getItemCount()).be.eql(expectedItemCount);
				should(sections[index].itemCount).be.eql(expectedItemCount);
				should(sections[index].items.length).be.eql(expectedItemCount);
			}
		}

		win = Ti.UI.createWindow();
		listView = Ti.UI.createListView({
			sections: [ createSection(), createSection() ]
		});
		verifySectionItemCount(maxItemCount);
		win.add(listView);
			listView.searchText = 'a';
			setTimeout(function () {
				// Item count should not change when filtering out rows via search field.
				verifySectionItemCount(maxItemCount);
				finish();
			}, 500);
		});
		win.open();
	});
});
```

apidoc/Titanium/UI/ListSection.yml
```diff
@@ -23,13 +23,20 @@ properties:
...
-        Using this property and `footerTitle` together is not supported. Use one or the other. 
+        Using this property and `footerTitle` together is not supported. Use one or the other.
         In Alloy you can specify this property with a `<FooterView>` child element of a `<ListSection>`
         element (see Examples).
     type: Titanium.UI.View
     since: 3.2.0
     platforms: [android, iphone, ipad]
 
+  - name: itemCount
+    summary: Returns the item count of the section, also incorporating the search filter if active.
```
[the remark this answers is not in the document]
m1ga: Getting the filtered count was the point of this PR. Otherwise you have no way to get the real (visible) count. You always can use getItems().length to get the whole item count even if you filter the list.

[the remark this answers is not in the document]
jquick-axway: Well, my point here is that iOS does not return the filtered count. It returns the total items (search text does not change this value). I assume iOS' undocumented `itemCount` API was made as an optimization since a `ListView` can have a lot of rows and it's wasteful to fetch a JavaScript array of items via `getItems()` to simply obtain its length.

Perhaps it would be best to create a new API to clearly indicate what's being fetched. Such as `filteredItemCount`. And also maybe`filteredItems` or `filteredIndexes` to obtain an array of which rows were inclusively filtered-in.

Is knowing the filtered item count by itself useful? I'm curious what your use-case is.

hansemannn:
For the item count of the (non-search items), we can use section.items.count already. I should update iOS to properly return the filtered items.

apidoc/Titanium/UI/ListSection.yml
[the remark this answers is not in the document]
hansemannn: +1 for `filteredItemCount` and `itemCount`. We currently do not have the space to do the iOS parity, so this PR may need more community work.

m1ga:
closing this one since #12029 will change a lot. Will check it with the RecyclerView once it is merged

PR #10041, opened 2018-05-13
Fix all dead links (404, 403 and 30x)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-0000

Ticket to be created. Tested using `blc --recursive --filter-level 3 <LOCALLY_DEPLOYED_DOCS>` Make sure to check [broken-link-checker](https://www.npmjs.com/package/broken-link-checker)!

I will likely address the default value crime mentioned in #10036 here as well.


sgtcoolguy:
Is there a way we can fold in checking broken links as part of our doc validation? Looks like that's an npm package, so can we actually have it run against the yml or some html generated files?

sgtcoolguy (approved):
The changes here LGTM, but I'd like to see if we can enforce checking this as part of the doc validation or something...

hansemannn:
So the test I ran took around an hour so it doesn't really perform well when being ran in every docs-related PR. If we could some how run it only on modified files, it'd be a good one.

sgtcoolguy:
So it sounds like not a great thing to try during an `npm test`, but maybe we could do this as part of the danger file. We have the set of changed files there, so if a yml doc is changed we could check that one. (Though it could lead to some nasty long builds when we touch a lot of doc files and don't change links)

hansemannn:
If we are able to get the actual `+`/`-` diff and regex the links to be checked. Shouldn't take too long in that case.

PR #10043, opened 2018-05-14
[TIMOB-25871] : Add ability to hide back button on NavigationWindow

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25871


iphone/Classes/TiUIWindowProxy.m
```diff
@@ -704,6 +705,12 @@ - (void)setBackButtonTitleImage:(id)proxy
   }
 }
 
+- (void)setHideBackButton:(id)arg
```
hansemannn: You use `hideBackButton` here, but `hidesBackButton` above. Also, `navigationItem` can be `nil` on window instances without navigation controllers. Do we even need this method here?

apidoc/Titanium/UI/Window.yml
```diff
@@ -525,6 +525,15 @@ properties:
     osver: {ios: {min: "8.0"}}
     availability: creation
 
+  - name: hidesBackButton
+    summary: Set this to true to hide the back button of navigation bar.
+    description: |
+        When this property is set to true, the navigation controller hides its back button.
```
hansemannn: > When this property is set to \`true\`, the navigation window hides its back button.

hansemannn (changes requested):

iphone/Classes/TiUIWindowProxy.m
vijaysingh-axway: Its not required. Added for testing but forget to remove. Now removed. Good Catch!

apidoc/Titanium/UI/Window.yml
```diff
@@ -528,7 +528,7 @@ properties:
   - name: hidesBackButton
     summary: Set this to true to hide the back button of navigation bar.
     description: |
-        When this property is set to true, the navigation controller hides its back button.
+        When this property is set to `true`, the navigation controller hides its back button.
```
hansemannn: Still incorrect, see the quote: "navigation window" over "navigation controller". The latter one is not a known word in Titanium.

hansemannn (approved):
CR **approved**! Awaiting unit-test and then can be moved to QE.

jquick-axway:
@hansemannn, @vijaysingh-axway, out of curiosity, what's the purpose of this feature? Not using a `NavigationWindow` will achieve the same effect, right?

The reason I ask because we want to implement `NavigationWindow` on Android in the future, but the bottom "back" button is always present on Android.

vijaysingh-axway:
@jquick-axway,  In iOS, when we push any controller in navigation controller, it comes with default back button. On click of which user will move to previous screen. 
If developer do not want to give  back button but want to show navigation title and right bar button, in that scenario hiding back button feature will be helpful.

PR #10066, opened 2018-05-18
[TIDOC-3190] Updated Ti.Network "online" property documentation

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIDOC-3190

**Summary:**
- Fixed error in `online` boolean property description which stated that it indicates "Internet" access. This is wrong.
- This property will be `true` if connected to a private network without Internet access. I've confirmed this behavior on both Android and iOS. I also think this is the correct behavior. If you want to detect Internet access, then you must do an HTTP request to a public website.

**Note:**
@hansemannn I've also noticed that iOS returns a `Number` instead of a `Boolean`. I haven't updated the doc to reflect this (it would be kind of odd to document it as `Boolean/Number`).

### DIFF
--- a/apidoc/Titanium/Network/Network.yml
+++ b/apidoc/Titanium/Network/Network.yml
@@ -433,7 +433,7 @@ events:
         summary: New network type as a string.
         type: String
       - name: online
-        summary: Boolean to indicate if the device is online.
+        summary: Boolean indicating if the device is connected to the network.
         type: Boolean
       - name: networkType
         summary: New network type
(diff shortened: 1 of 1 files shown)


hansemannn:
We should probably also add an API to actually show if it is connected to the internet.

hansemannn (approved):

jquick-axway:
> We should probably also add an API to actually show if it is connected to the internet.

This isn't so simple. On iOS, you have to use its "Reachability" APIs and give it a URL to poll. I don't recall Android or Windows having an equivalent. But if you think about it, you're effectively just HTTP requesting a URL regularly to detect Internet access, which isn't great.

hansemannn:
Yep, the iOS Reachability class is famous. It uses a C-API to ping the URL synchronously, which isn't too great as well. An HTTP-request probably is the easiest way, although not the cleanest. Do other frameworks solve this differently? In most cases, the Ti.Network.online property probably is "okay" for our devs.

jquick-axway:
I'm not aware of any other solution.

And instead of doing an HTTP request, perhaps it would be slightly better to attempt to establish a TCP socket connection to a public DNS server such as Google's `8.8.8.8:53`. But then we'd have to worry about IPv6 vs IPv4 and implement fallback mechanisms.
https://developers.google.com/speed/public-dns/docs/using

Personally, I've always found network-access APIs to be kind of useless. Even if it did detect Internet access, the network connection can go down at any time. Especially when you have shaky network access (like 1 bar with the cellular service or a bad WIFI connection). So, you can't assume that the device will stay connected in the near future.

PR #10069, opened 2018-05-21
[TIMOB-26038] Investigate moving to Obj-C JavaScriptCore API

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26038

**Description:**
This is PR is meant more as a technology spike/investigation and discussion piece.

In 2013 Apple released an Obj-C API for JavaScriptCore to make it easier to bind Obj-C native code to JS like we do. I thought it would be interesting to see how difficult it would be and what improvements/hurdles we'd see if we attempted to move to using that API.

The relevant commit is the last one, this was done on a fork of @hansemannn 's PR to remove TiCore to make things easier for me (since we cannot do this on TiCore).

Here's what I found:
- It cleans up the code quite a bit, pushing the binding code to the API
- It's likely going to result in a performance increase because each native class bound is turned into a JS class definition that is unique and properties/methods are bound more directly than our current generic KrollObject.getProperty style reflection
  - It'd be good if we had any sort of performance tests to try and gauge this versus our current approach (versus using the C API to define each type manually)
- It does not handle array index accessor/setters such as we use in `Ti.Buffer`. I hacked together a way to wrap Ti.Buffer instances in ES6 Proxy objects and that mostly worked, but `#toString()` broke, and some other method calls failed some internal type checks in JSC as well. I think this needs to stay the existing way we bind generically, or move to the C API directly. (Also given that ES6+ now has native buffer types, it may be time to consider deprecating and moving to those?)
- It doesn't auto-generate getter/setter methods for properties, and adding them is pretty redundant. You basically define the getter/setter methods in the header and then in the implementation delegate to the normal property method. (Again, the getter/setter methods are not idiomatic JavaScript. Should we consider deprecating them? We had this redundancy issue on the Windows side too).
- I think if we did move this way it'd be good to try and come up with a mechanism to set static values on the classes without using the bound property methods. We have a handful of cases where a property is read-only and holds a static value (like some of the `Ti.Platform` constants). It seems like it'd be slower/less direct to declare them as Obj-C properties with a method implementation that returns the static value. We should be able to try and modify the JS class/object returned by the Obj-C API the first time to set these static values on the prototype chain and that would likely cause some performance improvement.


hansemannn:
This is huge! Some early thoughts:
- +1 for removing generated getter & setter. Feels pretty Java-ish
- Does this also remove the Ti global namespace in favor of `new Switch(args)` etc? Probably it's a different change, which can benefit from this one already
- Can you give an example of "array index accessor/setters"? Not sure I follow
- Should the kroll-thread removal be a separate PR as well?

Ideally, this PR should go into the core before #9983 so we minimize the merge conflicts.

hansemannn:
If we properly document the breaking changes incl. migration steps, it should be okay. Especially to move forward to a more modern API design. Would we technically be able to keep the old behavior for one more major version to migrate? Otherwise, I think we still could write a little detection script to check for Ti-related getter/setter and error out during compile time. Something like that.

Also, would the following use-case be affected by this change already (heavily used although super ugly)?:
```
var label = Ti['UI]['createLabel'](args);
```

It is the right way forward to clean all this up. I'd even argument to deprecate the global namespace in 8 already, but it may be too much work.

iphone/Classes/PlatformModule.m
```diff
@@ -220,28 +221,38 @@ - (NSString *)identifierForVendor
 }
 
 #if defined(USE_TI_PLATFORMIDENTIFIERFORADVERTISING) || defined(USE_TI_PLATFORMGETIDENTIFIERFORADVERTISING)
-- (NSNumber *)isAdvertisingTrackingEnabled
+- (BOOL)isAdvertisingTrackingEnabled
```
hansemannn: This is a general question: Is our proxy able to return numeric values like `BOOL`, `int`, `NSInteger` or `CGFloat` as a Number? I noticed some crashes with this before, which required them to be `NSNumber` boxed.

iphone/Classes/PlatformModule.m
```diff
@@ -220,28 +221,38 @@ - (NSString *)identifierForVendor
...
+}
 #endif
 
 - (id)id
 {
   return [TiUtils appIdentifier];
 }
 
-- (NSString *)createUUID:(id)args
+- (NSString *)createUUID
```
hansemannn: Another one: Is our proxy now able to handle methods without the unused args? That'd be lit! But we would also need backwards compatibility to not break literally ALL available modules.

iphone/Classes/PlatformModule.m
```diff
@@ -266,31 +277,33 @@ - (NSNumber *)availableMemory
   return [NSNumber numberWithDouble:((vm_page_size * vmStats.free_count) / 1024.0) / 1024.0];
 }
 
-- (NSNumber *)openURL:(NSArray *)args
+- (BOOL)openURL:(NSString *)newUrlString
```
hansemannn: And another one: Can we unbox the first arg automatically now? What happens with > 1 args?

iphone/Classes/TiBuffer.m
```diff
@@ -17,30 +17,30 @@ @implementation TiBuffer
...
 - (id)init
 {
-  if (self = [super init]) {
-    byteOrder = [NUMLONG(CFByteOrderGetCurrent()) retain];
-  }
-  return self;
+    if (self = [super init]) {
+        byteOrder = [NUMLONG(CFByteOrderGetCurrent()) retain];
+    }
+    return self;
```
hansemannn: Was this clang-formatted? It looks like 4 vs 2 spaces (which we usually use).

iphone/Classes/TiBuffer.m
```diff
@@ -51,292 +51,292 @@ - (NSArray *)keySequence
...
-                location:CODELOCATION];
-  }
-
-  NSMutableData *cloneData = [[[NSMutableData alloc] initWithData:[data subdataWithRange:NSMakeRange(offsetVal, lengthVal)]] autorelease];
-
-  TiBuffer *newBuffer = [[[TiBuffer alloc] _initWithPageContext:[self executionContext]] autorelease];
-  [newBuffer setData:cloneData];
-  [newBuffer setByteOrder:byteOrder];
-  return newBuffer;
+    id offset = nil;
```
hansemannn: Isn't it a numeric value (`NSNumber`) for sure? Same for the length.

iphone/Classes/TiBuffer.m
```diff
@@ -51,292 +51,292 @@ - (NSArray *)keySequence
...
-    RELEASE_TO_NIL(data);
-    return;
-  }
-
-  if (data == nil) {
-    data = [[NSMutableData alloc] initWithLength:len];
-  } else {
-    [data setLength:len];
-  }
+    int len = [TiUtils intValue:length];
```
hansemannn: Unrelated to this PR: We should use `NSInteger` and `NSUInteger` instead of `int` in Obj-C.

iphone/Classes/TiPlatformDisplayCaps.m
```diff
@@ -30,17 +30,19 @@ - (NSString *)apiName
   return @"Ti.Platform.DisplayCaps";
 }
 
-- (id)dpi
+- (NSNumber *)dpi
 {
   return [NSNumber numberWithInt:[TiUtils dpi]];
 }
 
+// TODO Remove? Not in our docs!
```
hansemannn: Yep. Android also does not support it. Probably a left-over? We could git-blame to see the details if required.

iphone/Classes/TopTiModule.h
```diff
@@ -1,13 +1,31 @@
...
+@property (readonly, nonatomic) NSString *buildDate;
+@property (readonly, nonatomic) NSString *buildHash;
+@property (readwrite, nonatomic) NSString *userAgent;
+@property (readonly, nonatomic) NSString *version;
 
+- (JSValue *)createBuffer:(NSDictionary *)arg;
+
+// Accessors from JS. This is pretty redundant, Maybe introduce macros for this common case?
+// (i.e. readonly properties also generate getter methods, readwrite generator getter/setters)
+- (NSString *)getBuildDate;
```
hansemannn: in the current implementation, a `@property` in the header can be written as the following inside the implemention:
```objc
- (NSString *)buildDate
{
  return @"____";
}
```
Important: This allows it to be used both as a get-property and getter-method. So no manual `getX` method declaration in the header required, only if we explicitly want one, like if the getter has arguments.

iphone/Classes/TopTiModule.m
```diff
@@ -17,137 +17,160 @@
...
-- (id)version
+DEFINE_EXCEPTIONS
+
+- (NSString *)version
 {
   return @"__VERSION__";
 }
 
-- (id)buildDate
+- (NSString *)getVersion
```
hansemannn: Do we still need manual getters?

iphone/Classes/TopTiModule.m
```diff
@@ -17,137 +17,160 @@
...
-    } else {
-      switch ([TiUtils encodeString:data toBuffer:buffer charset:charset offset:0 sourceOffset:0 length:encodeLength]) {
-      case BAD_DEST_OFFSET: // Data length == 0 : return our empty buffer
-      case BAD_SRC_OFFSET: { // String length == 0 : return our empty buffer (no string encoded into it)
-        return buffer;
-        break;
-      }
-      case BAD_ENCODING: {
-        [self throwException:[NSString stringWithFormat:@"Invalid string encoding type '%@'", charset]
+    ENSURE_SINGLE_ARG_OR_NIL(arg, NSDictionary);
```
hansemannn: This should be moved to the `_initWithPageContext` method of the proxy itself.

hansemannn (commented):
Some comments and proposed changes. Love this PR!

iphone/Classes/APIModule.m
```diff
@@ -23,44 +19,29 @@ - (NSString *)apiName
...
-        return;
-      }
-    }
-#endif
-    NSLog(@"[%@] %@", [severity uppercaseString], [args componentsJoinedByString:@" "]);
+  else if (!([args isKindOfClass:[NSArray class]])) {
+    args = @[args];
   }
+    // If the arg is an NSNumber wrapping a BOOL we should print the string equivalent for the boolean!
+    NSMutableArray *newArray = [NSMutableArray array];
```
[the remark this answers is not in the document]
sgtcoolguy: I was trying to achieve parity in how we log booleans across platforms. Using 0/1 is very much an Obj-C thing, and not a JS thing. I assume that componentsJoinedByString has to enumerate the array itself. Maybe just fold the boolean transform and the building of the string into our own loop?

iphone/Classes/APIModule.m
```diff
@@ -91,34 +72,38 @@ - (void)error:(NSArray *)args
...
 }
 
-- (void)timestamp:(NSArray *)args
+- (void)timestamp:(id)args
 {
-  NSLog(@"[TIMESTAMP] %f %@", [NSDate timeIntervalSinceReferenceDate], [self transform:[args objectAtIndex:0]]);
+  id firstArg = args;
+  if ([args isKindOfClass:[NSArray class]]) {
+    firstArg = [args objectAtIndex:0];
+  }
```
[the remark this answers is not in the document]
sgtcoolguy: I'm not sure if that is still true now with the new API. I'm pretty sure I *had* to change this as a result of finding that the new API attempts to do it's own argument transforms to target types, and that if a method truly accepts multiple types you need to move to `id` and do the type-checking yourself. It's been a while since I worked on this, but I believe it tried to coerce arguments to the native type declared, and would fail to turn a String or Number into an Array.

iphone/Classes/APIModule.m
```diff
@@ -71,17 +52,17 @@ - (id)transform:(id)arg
   return arg;
 }
 
-- (void)debug:(NSArray *)args
+- (void)debug:(id)args
```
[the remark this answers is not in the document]
sgtcoolguy: See my comment below. I believe the new API binding actually does allow you to use the actual types, but if any given method accepts more than one type (or we used to implicitly coerce it ourselves before) it would fail. My way to hack around that was to move to `id` and handle the cases myself if we accept differing types.

iphone/Classes/KrollBridge.m
```diff
@@ -529,47 +382,130 @@ - (void)didStartNewContext:(KrollContext *)kroll
...
+                                   @"Geolocation",
+                                   @"Gesture",
+                                   @"Locale",
+                                   @"Media",
+                                   @"Network",
+                                   @"Stream",
+                                   @"UI",
+                                   @"Utils",
+                                   @"XML",
+                                   @"Yahoo"];
```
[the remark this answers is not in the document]
sgtcoolguy: This was a hack to allow me to mix and match the converted APIs versus the unconverted. If we want to filter them via usage sniffing from the CLI, we can change this.

iphone/Classes/PlatformModule.m
sgtcoolguy: I believe so? https://developer.apple.com/documentation/javascriptcore/jsvalue?language=objc

CGFloat is a typedef for `double`. Looks like `NSNumber` and primitive numeric types are handled.

iphone/Classes/PlatformModule.m
sgtcoolguy: As in "Do I still need to define arguments for a method even if we don't accept/use any?" Then no, you do not need to add useless arguments now. There's an example right in their very sparse docs here for a description method with no args: https://developer.apple.com/documentation/javascriptcore/jsexport?language=objc

iphone/Classes/TopTiModule.h
sgtcoolguy: Yeah and this difference was what raised the removal of getX/setX accessor methods to my mind when doing this PR. I know that our proxy hack will fall back to mangling `getX` to look for an `x` property (or more specifically just a `x` method with no args). 

With the new API, that doesn't happen. It's not a JS idiom, so they didn't build it in. So with the new binding API we need to make the getter/setter methods explicitly and just have them delegate over. (which would be ok if we could write macros to do it for us, but macros aren't flexible enough to camelcase a property name)

sgtcoolguy:
For more info:
- Here's where the API does the type conversions: https://github.com/WebKit/webkit/blob/9be0bde9221db8c348511fa4bd10cbb2eaa607b7/Source/JavaScriptCore/API/JSValue.mm
- Here's where the magic JSExport proxying setup is done: https://github.com/WebKit/webkit/blob/9be0bde9221db8c348511fa4bd10cbb2eaa607b7/Source/JavaScriptCore/API/JSWrapperMap.mm

sgtcoolguy:
I'm going to close this PR for now as it was meant as a discussion to work with @hansemannn about moving in this direction in the 8-9 timeframe. Seeing as we need to land PRs related to TiCore removal/KrollThread, etc to lay the groundwork, I'm going to close this and push to a different branch name

sgtcoolguy:
For reference, this code is now at https://github.com/sgtcoolguy/titanium_mobile/tree/ios-objc-api

PR #10071, opened 2018-05-22
[TIMOB-26033]: iOS Ti.Media.VideoPlayer shows an error on Window Close on SDK 7.1.1.GA

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26033

### DIFF
--- a/iphone/Classes/TiMediaVideoPlayerProxy.m
+++ b/iphone/Classes/TiMediaVideoPlayerProxy.m
@@ -117,8 +117,10 @@ NSArray *moviePlayerKeys = nil;
 
 - (void)removeNotificationObserver
 {
+  if ([self observationInfo]) {
+    [self removeObserver:self forKeyPath:@"url"];
+  }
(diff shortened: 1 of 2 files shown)


iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -117,8 +117,10 @@ - (void)addNotificationObserver
 
 - (void)removeNotificationObserver
 {
+  if ([self observationInfo]) {
+    [self removeObserver:self forKeyPath:@"url"];
```
hansemannn: Does this only apply for the url KVO or possibly others as well (once triggered)?

vijaysingh-axway: For others also we can apply. But 'movie' object already becomes nil. So no need to apply this check for others.

hansemannn (approved):
CR looks fine. Awaiting unit-tests to proceed.

PR #10072, opened 2018-05-22
[TIMOB-26062] iOS: Fix all Xcode Analyzer warnings & possible leaks

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26062

For most of the changes, we either already had unit-tests or they were IDE-related (clang-analyzer). For the `zoomLevel`, I have added a new unit test that passes.


hansemannn:
@sgtcoolguy I would like to include this in 7.3.0 if you agree.

vijaysingh-axway (approved):
CR passed.

PR #10074, opened 2018-05-24
iOS: (7_2_X) Move Analytics unit-test to the correct directory

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26004


hansemannn:
An unrelated unit-test (Android WebView) still fails randomly. Merging as it's not related to this PR and the actual addon-test passes:
```
[INFO] :   !TEST_END: {"state":"skipped","duration":0,"suite":"Titanium.Analytics","title":".optedOut","message":""}
```

PR #10075, opened 2018-05-25
[TIMOB-26069] Android: Bring TitaniumDev up to speed with recent changes.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26069

**Description:**
Minor updates to TitaniumDev:
 - implement a recently added method to AssetCrypt.
 - add /lib as a dependency for UI module


hansemannn:
I have a dream: Using it without the need of 9349223 configurations before. Is that possible? :D

PR #10076, opened 2018-05-26
[TIMOB-16066] Android: Added "foreground" service support

jquick-axway:
> PR #9945 must be merged first, then this PR must be updated afterwards.
> Otherwise this feature will cause a crash and have unit test failures.

**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-16066
- https://jira.appcelerator.org/browse/TIMOB-25974

**Summary:**
- Added method `Service.foregroundNotify(id, notification)`
  * Changes service from the "background" state to the "foreground" state via the native `startForeground()` Java method.
  * Also posts a notification to the status bar to let end-user know service is doing work in the background.
  * Foreground services are less likely to be terminated while the app is in the background. Also needed to collect location data in the background on Android 8.
- Added method `Service.foregroundCancel()`
  * Changes service back to the "background" state via native `stopForeground()` Java method.
  * Automatically removes the foreground notification from the status bar. (Stopping the service will do this too.)

**Test:**
This tests that location data can be collected with a foreground service.
1. Set up a classic Titanium app with the below.
2. Build and run on a physical Android 8.x device.
3. If the app asks for location permission on app startup, then tap yes.
4. Verify that a "Foreground Service" notification is listed in the status bar.
5. Tap the "Home" button to put the app in the background. (Do **NOT** press back button.)
6. Walk around outside so that the app can receive GPS updates.
7. Verify that the devices vibrates about every 20 seconds. It vibrates when the app gets a GPS update. (This tests [TIMOB-25974](https://jira.appcelerator.org/browse/TIMOB-25974).)
8. Swipe down on the status bar.
9. Tap on the "Foreground Service" notification to put the app back in the foreground.
10. Press the back button to exit.
11. Verify that the "Foreground Service" notification has been removed from the status bar.
12. Repeat steps 4-11 on an Android 4.x device.

---
tiapp.xml
```xml
		<services>
			<service url="Service.js"/>
		</services>
	</android>
</ti:app>
```

---
./Resources/Service.js
```javascript
Ti.API.info("@@@ Service started.");

// Gets an intent used to resume the app if running in the background.
function getAppResumeIntent() {
	var intent = Ti.App.Android.launchIntent;
	if (!intent) {
		intent = Ti.Android.createIntent({});
	}
	return intent;
}

// Put this service into the foreground state.
var notificationId = 123;
Ti.Android.currentService.foregroundNotify(notificationId, Ti.Android.createNotification({
	contentTitle: "Foreground Service",
	contentText: "Content Text",
		// Using the launch intent will resume the app when tapped.
		intent: getAppResumeIntent(),
	}),
}));
```

---
./Resources/app.js
```javascript
function monitorLocation() {
	function start() {
		Ti.Geolocation.accuracy = Ti.Geolocation.ACCURACY_HIGH;
		Ti.Geolocation.addEventListener("location", function(e) {
			Ti.Media.vibrate();
			Ti.API.info("@@@ Location Received: " + JSON.stringify(e.coords));
		});
	}
	var hasPermission = Ti.Geolocation.hasLocationPermissions(Ti.Geolocation.AUTHORIZATION_WHEN_IN_USE);
	if (!hasPermission) {
		Ti.Geolocation.requestLocationPermissions(Ti.Geolocation.AUTHORIZATION_WHEN_IN_USE, function(e) {
			if (e.success) {
				start();
			}
		});
	} else {
		start();
	}
}

// Start the service.
var intent = Ti.Android.createServiceIntent({
	url: "Service.js",
});
Ti.Android.startService(intent);

// Display the main window.
window.addEventListener("open", function(e) {
	monitorLocation();
});
window.addEventListener("close", function(e) {
	// Stop the service on app exit.
	Ti.Android.stopService(intent);
});
window.open();
```


android/titanium/src/java/org/appcelerator/titanium/proxy/ServiceProxy.java
```diff
@@ -107,15 +116,55 @@ public void stop()
...
 
+	@Kroll.method
+	public void foregroundNotify(int notificationId, KrollProxy notificationProxy)
+	{
+		// Validate arguments.
+		if (notificationId == 0) {
+			throw new RuntimeException("Notification ID argument cannot be set to zero.");
+		}
+		if (notificationProxy == null) {
+			throw new RuntimeException("Notification object argument cannot be null.");
```
hansemannn: This seems to be thrown on Sony devices, while it works for others. See the latest comment in the JIRA for more logs. It may be a dev-error, but lets be sure to validate it.

garymathews: Maybe the `Ti.App.Android.launchIntent` is `null` on Sony devices?

https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/android/src/java/ti/modules/titanium/android/PendingIntentProxy.java#L58

jquick-axway: Yeah. It does sound like `launchIntent` is `null`. That should never happen. I'll look into it.

jquick-axway: I've concluded that this is not an issue with this PR. A `launchIntent` can in fact return `null` in rare cases. I've modified the test code to handle this case.

garymathews (approved):
CR: **PASS**

PR #10078, opened 2018-05-28
[TIMOB-26071] TiAPI: Add Ti.WebDialog to core-modules

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26071


sgtcoolguy:
Actually, this is not a breaking change as the safari dialog module wasn't removed here. I think we can merge this for 7.4 and remove `ti.safaridialog` in 8.0.0 (Normally we'd have to deprecate and warn with a longer cycle, but I think given that devs can just grab the module themselves if they still haven't migrated that's it's not quite so bad).

sgtcoolguy (approved):

PR #10079, opened 2018-05-28
[TIMOB-26073] iOS: Implement support for custom TextField backgroundColor in SearchBar

AndreaVitale:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26073


Topener:
@AndreaVitale looks like your code needs some linting, can you fix it?

iphone/Classes/TiUISearchBar.m
```diff
@@ -150,6 +150,20 @@ - (void)setColor_:(id)value
   }];
 }
 
+- (void)setFieldBackgroundColor_:(id)value
+{
+    // TIMOB-10368
+    // Remove this hack again once iOS exposes this as a public API
+    UIView *searchContainerView = [[[self searchBar] subviews] firstObject];
+    
+    [[searchContainerView subviews] enumerateObjectsUsingBlock:^(__kindof UIView *_Nonnull obj, NSUInteger idx, BOOL *_Nonnull stop) {
```
hansemannn: Not sure if we should take this. It may be globally rejected by Apple which is not suitable. Can't this be done via Hyperloop if required, e.g.:
```js
var nativeSearchBar = UISearchbar.cast(tiSearchBar);
var searchContainerView = nativeSearchBar.subViews.firstObject;

for (var i = 0; i < searchContainerView.subviews.count; i++) {
  var subview = searchContainerView.subviews.objectAtIndex(i);
  if (subview.isKindOfClass(UITextField.class) {
    subview.backgroundColor = /* Converted hex color to RGB UIColor */
  }
}
```
I'd request some more feedback from @jquick-axway and @vijaysingh-axway and their thoughts!

AndreaVitale: I know, but I saw the same code for changing the textfield text color that is actually working in 7.1.0.GA: https://github.com/EveryUP/titanium_mobile/blob/b96c1eca233b566d5f33d5127c202501cc6f20df/iphone/Classes/TiUISearchBar.m#L139

If we cannot loop over child views the only public API that we can use is related to backgroundImage. An example can be:

```
- (void)setFieldBackgroundImage_:(id)arg
{
    [[self searchBar] setSearchFieldBackgroundImage:[self loadImage:arg] forState:UIControlStateNormal];
}

- (void)setFieldBackgroundDisabledImage_:(id)arg
{
    [[self searchBar] setSearchFieldBackgroundImage:[self loadImage:arg] forState:UIControlStateDisabled];
}
```

but in my opinion is more uncomfortable and heavy :(

hansemannn: I agree. If it is in only places, it should be okay for now. The other thing is that we pushing hard for parity between platforms, so Android would be required as well, if existing. Do you know how it behaves on Android?

AndreaVitale: Not yet, but when I'll finish to work on the current project feature that involved this iOS fix I will investigate the Android side.
I'll let you know!

hansemannn: From looking at the Android docs, there is only one background-color on the search-bar and that is also exposed in Titanium. So I would agree on `fieldBackgroundImage` and `fieldBackgroundDisabledImage` if you do as well. I could fix up the linting then.

jquick-axway: Yeah, let's avoid private API usage. This may cause an app review rejection.

vijaysingh-axway: Ideally we should avoid private APIs. In my view, here we can wrap this function inside a macro, which will ensure that this will come in picture only when this method is used (Similar to MediaModule).  And we can put a note in document also.

AndreaVitale: Done! I removed the fieldBackgroundColor property and introduced `fieldBackgroundImage` and `fieldBackgroundDisabledImage`.

Here is an example:
```
    $.Search.setFieldBackgroundImage(Ti.UI.createView({
        height: 36,
        width: Ti.Platform.displayCaps.platformWidth - 20,
        backgroundColor: '268E8E93',
        borderRadius: 12
    }).toImage());
```
that produces:

schermata 2018-05-31 alle 12 32 49

iphone/Classes/TiUISearchBar.m
```diff
@@ -150,6 +150,18 @@ - (void)setColor_:(id)value
   }];
 }
 
+- (void)setFieldBackgroundImage_:(id)arg
+{
+  [[self searchBar] setSearchFieldBackgroundImage:[self loadImage:arg] forState:UIControlStateNormal];
+  [self.searchBar setSearchTextPositionAdjustment:UIOffsetMake(8.0, 0.0)];
```
hansemannn: I am not sure the 8px padding should be forced. We could rather expose it as an own property. What do you think?

AndreaVitale: I force it due to [this](http://openradar.appspot.com/radar?id=6667237251350528).
After setting the `fieldBackgroundImage`, the textfield lose its horizontal margin.

Of course can be exposed as an own property, but for this specific case we avoid to force the dev to set it after setting the `fieldBackgroundImage`.

hansemannn: I am fine then!

hansemannn (approved):
CR **approved**, thank you @AndreaVitale! Scheduling for 7.4.0, since 7.3.0 is already feature-frozen.

AndreaVitale:
Great!!!

hansemannn:
@AndreaVitale Can you re-sign your [CLA](https://cla.axway.com/) above? I think we changed the CLA server, sorry for the trouble! But your PR is ready to merge and can be integrated today.

AndreaVitale:
No problem, signed again right now!!!

PR #10080, opened 2018-05-29
[TIMOB-26077] iOS: Ti.UI.convertUnits returns incorrect values for 3x resolution devices

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26077

**Description:**
Add comprehensive unit tests around Ti.UI.convertUnits and fix an iOS bug in the implementation which did not account for conversions with 3x resolutions

- This adds tests for various combinations of unit conversions for Ti.Ui.convertUnits.
- This exposed a bug in iOS where 3x devices gave bad results, so the PR includes the simple fix.
- Added tests for special cases like non-numeric inputs or percentage inputs.

TODO
- Looks like iOS does not consult ti.ui.defaultunit property for converting when input units aren't specified. It assumes "dp", which is the default value placed in tiapp.xml.
- Android appears to not handle the special case of input values as percentages. It should bail early and return 0. The test is skipped there for now.


tests/Resources/ti.ui.test.js
```diff
@@ -0,0 +1,239 @@
...
+
+	// TODO Write tests for Ti.UI.global properties below!
+	it('backgroundColor');
+	it('backgroundImage');
+	it('currentTab');
+
+	// FIXME Get working on iOS
+	// FIXME Get working on Android - Ti.UI.currentWindow is null
+	// Supposedly this property should only exist when using Ti.UI.Window.url to load JS files into own context! But we support elsewhere for Windows
+	it.androidAndIosBroken('.currentWindow', function (finish) {
```
hansemannn: This property only existed before we removed the `Ti.UI.Window.url` property and was forgotten during the docs update, see [here](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI-property-currentWindow) for reference.

hansemannn (changes requested):
One unit test and its outdated docs should be removed. Everything else seems fine.

hansemannn (approved):

hansemannn:
Ready to merge once 7_3_X is branched.

sgtcoolguy:
Created tickets for the outstanding bugs found while working on this:
- https://jira.appcelerator.org/browse/TIMOB-26113
- https://jira.appcelerator.org/browse/TIMOB-26114

PR #10082, opened 2018-06-01
[TIMOB-26004] iOS: Use updated APSAnalytics library with ti.foreground fix

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26004


ewieberappc (approved):
Verified that restarting the app with optedOut = true, then toggling it to false will still allow events to be fired

PR #10084, opened 2018-06-02
 [TIMOB-26079] Continue build on invalid Android library ABI

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26079

**Optional Description:**
This aligns the handling of invalid ABIs for native libraries from Android .aar files to be the same as in the other stages of creating the APK archive.

@cb1kenobi I had to remove the override for `console.error` because that caused all our `debug` and `warn` messages to not show up. I'm sure you had your reason to temporarily override it. Do you remember why and if there is a better way to do this nowadays? The comment said something about to many event listeners but i couldn't see anything when testing this.


cb1kenobi:
As the comment describes, we use the [`archiver`](https://www.npmjs.com/package/archiver) package to create the `.apk` (zip format) files. When the Android build was first written, it used `archiver@0.4.10` and that version would write to stderr without a way to prevent it. Now, the titanium_mobile master branch references `archiver@2.1.0` and it's likely the silencing of `console.error()` is no longer necessary.

cb1kenobi (approved):
CR'd and looks good.

android/cli/commands/_build.js
```diff
@@ -4393,200 +4393,191 @@ AndroidBuilder.prototype.createUnsignedApk = function createUnsignedApk(next) {
...
+			}
+		}, this);
+	}, this);
 
-					dest.append(src.readFile(entry), {
-						name: entry.entryName,
-						store: store
-					});
-					entryNames.push(entry.entryName);
+	this.logger.debug('test3');
```
sgtcoolguy: Remove this straggler debug log, please

sgtcoolguy (approved):
LGTM (except that straggler debug log to be removed)

PR #10088, opened 2018-06-05
[TIMOB-26094] iOS 12: Add new Ti.UI.TextField API's

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26094

> Using `no tests` until our build machine runs Xcode 10.


sgtcoolguy:
@hansemannn TiApp.m fails to compile on Jenkins here as well (like it does on #10098). The machine it built on does have Xcode 9 installed...

sgtcoolguy:
Also this PR looks like it's two separate tickets/issues that aren't particularly related...

hansemannn:
@sgtcoolguy I need to rebase with the latest changes, thanks for merging! The reason these two are together is that all Xcode 10 additions to the SDK require the IS_XCODE_10 macro, which I've put there to unblock development. I will split it up now and we need to rebase once the macro is merged, so Jenkins passes.

vijaysingh-axway (approved):
CR passed.

PR #10089, opened 2018-06-05
iOS: Update ioslib to 1.7.1, adding Node 10- and Xcode 10 support

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26096

> **NOTE**: This can only be merged once 7_3_X is branched and master is at least 7.4.0.

PR #10090, opened 2018-06-05
[TIMOB-26080] Fix ERR_INVALID_CALLBACK error on Android build and misc Node 10 fixes

ewanharris:
This PR contains two tickets as they are closely related, however the second commit has some consequences on our supported Node version, as `Buffer.from` is only available in Node 6+. We can easily ponyfill this using [buffer-from](https://github.com/LinusU/buffer-from) but I chose to drop Node 4 for the following reasons

- It is end of life, as per our [third party software support](https://docs.appcelerator.com/platform/latest/#!/guide/Axway_Appcelerator_Deprecation_Policy) it is not required to support Node 4 now that it is EOL, and I would not consider this removal a semver major for that reason.
- [Node 4 support is currently broken](https://jira.appcelerator.org/browse/TIMOB-25944), I haven't seen any one (but me) complain about this.

That said, if anyone has differing opinions, I'm happy to rework fd207d044face2ccf0a70c56a7c813fdeeeeb835 to use the ponyfill

**Tickets:**

1. **JIRA:** https://jira.appcelerator.org/browse/TIMOB-26080
- Fix the `ERR_INVALID_CALLBACK` error when building a project with .aar files.
	- Steps to reproduce:
	1. Download the [facebook module](https://github.com/appcelerator-modules/ti.facebook)
	2. Build using this SDK and Node 10, no error should be thrown

2. **JIRA:** https://jira.appcelerator.org/browse/TIMOB-26081
- Remove deprecation warnings from `new Buffer` usage
- It's kinda easy to miss these logs in a debug/trace log level, it's easy to spot in info. They look like `(node:19383) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.`
	- Steps to reproduce:
	1. Build an app for iOS where the `DefaultIcon.png` has an alpha channel, [here's one](https://www.dropbox.com/s/5hgafhldtoxo1y4/DefaultIcon.png?dl=0)
	2. Build any Android module

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -26,7 +26,7 @@
 		"android platform tools": "27.x",
 		"android tools": "<=26.x",
 		"android ndk": ">=r11c <=r16c",
-		"node": ">=4.0 <=7.x",
+		"node": ">=6.0 <=8.x",
 		"java": ">=1.8.x"
 	},
 	"repository": {
--- a/android/package.json
+++ b/android/package.json
@@ -26,7 +26,7 @@
 		"android platform tools": "27.x",
 		"android tools": "<=26.x",
 		"android ndk": ">=r11c <=r16c",
-		"node": ">=6.0 <=8.x",
+		"node": ">=4.0 <=8.x",
 		"java": ">=1.8.x"
 	},
 	"repository": {
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -1531,7 +1531,7 @@ AndroidModuleBuilder.prototype.jsToC = function (next) {
 
 	if (fs.existsSync(jsBootstrapFile)) {
 
-		const str = new Buffer(fs.readFileSync(jsBootstrapFile)); // eslint-disable-line security/detect-new-buffer
+		const str = Buffer.from(fs.readFileSync(jsBootstrapFile));
 
 		[].forEach.call(str, function (char) {
 			result.push(char);
--- a/android/cli/hooks/aar-transform.js
+++ b/android/cli/hooks/aar-transform.js
@@ -481,7 +481,7 @@ class SimpleFileCache {
 	 */
 	persist() {
 		var dataToWrite = JSON.stringify(this.data);
-		if (!fs.exists(path.dirname(this.cachePathAndFilename))) {
+		if (!fs.existsSync(path.dirname(this.cachePathAndFilename))) {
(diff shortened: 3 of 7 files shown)


android/package.json
```diff
@@ -26,7 +26,7 @@
 		"android platform tools": "27.x",
 		"android tools": "<=26.x",
 		"android ndk": ">=r11c <=r16c",
-		"node": ">=4.0 <=7.x",
+		"node": ">=6.0 <=8.x",
```
hansemannn: So we don't need `<= 10.x` here?

ewanharris: Left out on purpose as this PR doesn't introduce full Node 10 support. This + #10089 + some further testing and I'd feel comfortable claiming support

hansemannn: Alright. I think it should be added and tested with this ticket, as it's all about Node 10 anyway.

ewanharris: Want me to cherry-pick your commit across to here?

ewanharris:
Hmm it appears I'm wrong about the `Buffer.from` availability, [docs state](https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_class_method_buffer_from_array) it was added in 5.10.0, but it does appear to exist in Node 4

android/package.json
hansemannn: Isn't it a different change? My one only updated ioslib and it's compatibility, but I think the CLI will warn if the package.json in the iphone/android dir of the SDK are not updated as well.

sgtcoolguy:
Related: appcelerator/node-appc#129

sgtcoolguy:
So I think the changes to the node version range should probably be rolled back if this change still does work on Node 4, and likely the top-level package.json should be updated to use node-appc 0.24.6 - then this PR can be merged.

I think if do decide to bump our minimum Node support, it should be done in a separate PR explicitly with some attention to Brian to update the docs/support matrix.

android/package.json
sgtcoolguy: @ewanharris  I've merged Hans' ioslib update PR after fixing ioslib/node-ios-device, so this can now be tested for Node 10 support.

ewanharris:
Update to reinstate the node 4 compatibility with an update to latest node-appc and reverting the package.json changes

PR #10091, opened 2018-06-06
[TIMOB-26103] iOS 12: Expose new UserNotifications API's

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26103


hansemannn:
@sgtcoolguy Is it already possible to build with Xcode 10 Beta as of for now? We are 3 months early this time :-).

vijaysingh-axway (changes requested):
1. I can see that  "threadIdentifier" api is not exposed, which can be used to visually group notification. Please expose this.
2. If you can give a test case would help qe to test it. Thanks!

vijaysingh-axway (approved):
CR passed.

PR #10094, opened 2018-06-07
[TIMOB-26106] Windows: Update module apiversion

infosia:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26106

Update module apiversion for Windows because there's module binary incompatibility in upcoming 7.3.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "moduleApiVersion": {
     "iphone": "2",
     "android": "4",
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
I think this raises the issue that perhaps the apiversion should be specified in the per-platform package.json files, rather than top-level. Obviously that has a CLI impact...

PR #10097, opened 2018-06-08
[TIMOB-26110] iOS: Do not create build directory via CLI

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26110

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -2906,16 +2906,11 @@ iOSBuilder.prototype.initBuildDir = function initBuildDir() {
 	const buildDirExists = fs.existsSync(this.buildDir);
 
 	if (this.forceCleanBuild && buildDirExists) {
-		this.logger.debug(__('Recreating %s', cyan(this.buildDir)));
+		this.logger.debug(__('Cleaning %s', cyan(this.buildDir)));
 		wrench.rmdirSyncRecursive(this.buildDir);
-		wrench.mkdirSyncRecursive(this.buildDir);
 	} else if (!buildDirExists) {
-		this.logger.debug(__('Creating %s', cyan(this.buildDir)));
-		wrench.mkdirSyncRecursive(this.buildDir);
 		this.forceCleanBuild = true;
 	}
-
-	fs.existsSync(this.xcodeAppDir) || wrench.mkdirSyncRecursive(this.xcodeAppDir);
 };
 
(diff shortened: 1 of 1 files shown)


cb1kenobi:
I'm pretty sure we cannot do this. We copy all resources into that build directory BEFORE we invoke xcodebuild. If the directory does not exist, then the writing of the resources will fail. This is fundamental to the build.

The new Titanium build pipeline was going to copy the resources into a staging directory (in the build directory) and those files would be added to the Xcode project, then xcodebuild would copy them into the final output directory. I believe this to also be faster. Simply put, it's a major change and we should not invest in such a major change with the current build system.

hansemannn:
@cb1kenobi It works fine with clean and incremental builds in Xcode 9 and 10. And since the "new Titanium build pipeline" isn't expected happen anytime soon, we need to figure out a way to unblock us. Alternatives are welcome.

cb1kenobi (changes requested):
After testing this PR, I have determined the following:

* This does not fix the problem
* Removing these lines doesn't appear to break anything on the surface
* Not initializing these directories is sloppy

Here's the deal. `xcodebuild build` works just fine. The problem is `xcodebuild clean`:

```
[INFO]  Cleaning Xcode derived data
[DEBUG] Invoking: DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer /Applications/Xcode-beta.app/Contents/Developer/usr/bin/xcodebuild clean
[ERROR] note: Using new build system
[ERROR] error: Could not delete `/Users/chris/appc/workspace/testapp3/build/iphone/build` because it was not created by the build system and it is not a subfolder of derived data.
[ERROR] ** CLEAN FAILED **
```

We started calling `xcodebuild clean` to clean derived data ever since I improved the build by keeping intermediate files between builds. Xcode 10 defaults to the new build system for `clean` operations. The solution is to tell xcodebuild to not use the new build system like this:

```
diff --git a/iphone/cli/commands/_build.js b/iphone/cli/commands/_build.js
index c6dbb45aa6..345d7cbb5a 100644
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -4678,7 +4678,7 @@ iOSBuilder.prototype.cleanXcodeDerivedData = function cleanXcodeDerivedData(next
        }

        const exe = this.xcodeEnv.executables.xcodebuild,
-               args = [ 'clean' ];
+               args = [ 'clean', '-UseNewBuildSystem=NO' ];
        let tries = 0,
                lastErr = null,
                done = false;
```

PR #10098, opened 2018-06-08
[TIMOB-26111] iOS: Fix all static analyzer warnings and possible leaks

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26111

> `no tests` because of the existing tests. See the linked JIRA ticket for details.


sgtcoolguy:
@hansemannn Looks like TiApp.m fails to compile on Jenkins here

hansemannn:
@sgtcoolguy It looks like Xcode 9 does not expose the `` protocol public, so I'd like to wrap it for Xcode 9 and 10. But that requires https://github.com/appcelerator/titanium_mobile/pull/10088/files#diff-d564eaef8d6d4f17f70247e5ec99cde7R20 to be merged first, so I can use conditional compiling. Until then, I restored that part, since it's just a warning.

PR #10100, opened 2018-06-11
[TIMOB-26121] iOS: Prioritize selected xcode when determining default simulator

garymathews:
- Always prioritize the selected xcode when determining the default simulator iOS version

##### TEST CASE
- Install xcode-beta 10.0 along with xcode 9.x
- Use `xcode-select` to select 9.x
- Build a project without specifying a simulator: `appc run -p ios -l trace`
- A compatible iOS 11 simulator will be selected

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26121)


iphone/cli/commands/_build.js
```diff
@@ -1936,7 +1936,11 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 				const xcodeInfo = this.iosInfo.xcode;
 
 				function sortXcodeIds(a, b) {
-					return xcodeInfo[a].selected || appc.version.gt(xcodeInfo[a].version, xcodeInfo[b].version) ? -1 : appc.version.lt(xcodeInfo[a].version, xcodeInfo[b].version) ? 1 : 0;
+					// prioritize selected xcode
+					if (xcodeInfo[b].selected) {
+						return 1;
```
cb1kenobi: You need to be mindful of the original sort order. This change makes the selected Xcode the last element where the previous code had it be the first. This sort function is used in 2 places. The first instance reverses the results and the other doesn't. This code would change that behavior.

`iosSdkVersion` is set if an iOS SDK version was passed in via the command line `--ios-version <ver>` option. When this is set, we want to pick the most recent Xcode or selected Xcode.

If there's no `iosSdkVersion` and it's a device/dist build, then we pick the oldest Xcode. Why? I have no idea. Compatibility? Who knows?

So, I'm thinking you either need to remove the `.reverse()` or change the `-1` to `1` and `1` to `-1` in the sort function. The former is probably the better, cleaner route. We should also figure out why the code chose the oldest Xcode to select an iOS Simulator from.

cb1kenobi (changes requested):

garymathews:
@cb1kenobi Updated PR

iphone/cli/commands/_build.js
```diff
@@ -1954,9 +1958,8 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 						logger.error(__('Unable to find any Xcode installations that support iOS SDK %s.', this.iosSdkVersion) + '\n');
 						process.exit(1);
 					}
-				} else if (cli.argv.target === 'simulator' && !cli.argv['build-only']) {
```
cb1kenobi: You can't do this. This line must remain.

If you're doing a sim build, running the app in a sim, and you didn't specify an iOS SDK version, then do nothing. Why? Because on https://github.com/garymathews/titanium_mobile/blob/74c48667224892db965fcca6b70cbf1fffcabffd/iphone/cli/commands/_build.js#L2012 when it calls `ioslib.simulator.findSimulators()`, it will determine the best iOS sim and Xcode version based on the inputs and return the results. If there isn't a selected iOS SDK version, then we use the latest associated with the Xcode version for the simulator it chose (https://github.com/garymathews/titanium_mobile/blob/74c48667224892db965fcca6b70cbf1fffcabffd/iphone/cli/commands/_build.js#L2043-L2046).

cb1kenobi (changes requested):

iphone/cli/commands/_build.js
garymathews: > If there isn't a selected iOS SDK version, then we use the latest associated with the Xcode version for the simulator it chose.

That is what this change fixes. It should not use the latest Xcode as it may not be supported and it's ignoring my `xcode-select` selection. With this change, it will use a supported Xcode version and select the first listed simulator that is greater than our `minIosVersion`.

cb1kenobi (changes requested):
You still have a problem in _build.js with the removal of line 1957-1959. Those lines are necessary.

iphone/cli/commands/_build.js
cb1kenobi: OK. I'm not concerned with what version of Xcode is used. I'm very concerned if the iOS SDK changes depending on whether or not `--build-only` is specified.

I think part of the complexity is the iOS SDK version is passed in as a command line argument when I believe it should be a setting in the `tiapp.xml`.

I'm fine with this PR, but then I believe lines 2043-2046 become dead code.

I think we also need to fix this in ioslib. The Titanium build shouldn't need to be burdened with all this selecting/sorting logic. Care will need to be taken as to not repeat what got us in this situation.

cb1kenobi (approved):
CR'd and FR'd. **APPROVED**!

PR #10101, opened 2018-06-12
[TIMOB-26113] iOS: Prioritize ti.ui.defaultunit over other dimensions

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26113

> Adding tests once CR approved


hansemannn:
Ok, so this does not really work out. Will go with a different approach.

hansemannn:
Closing as this does not seem to work out so far.

PR #10103, opened 2018-06-12
[TIMOB-26123] iOS: Drop iOS 8 support, use iOS 9 as the minimum target

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26123


iphone/Classes/TiContactsPerson.m
```diff
@@ -586,39 +541,26 @@ - (TiBlob *)image
...
-  }
 
-  ABRecordSetValue([self record], kABPersonBirthdayProperty, (CFDateRef)[TiUtils dateForUTCDate:date], NULL);
+  NSDate *saveDate = [TiUtils dateForUTCDate:date];
+  unsigned unitFlags = NSCalendarUnitYear | NSCalendarUnitMonth | NSCalendarUnitDay;
+  NSCalendar *cal = [NSCalendar currentCalendar];
+  person.birthday = [cal components:unitFlags fromDate:saveDate];
+  [self checkAndNotifyObserver];
+
+  return;
```
vijaysingh-axway: No need of return.

iphone/package.json
```diff
@@ -18,8 +18,8 @@
 	"minIosVersion": "8.0",
 	"minWatchosVersion": "1.0",
 	"vendorDependencies": {
-		"xcode": ">=6.0 <=9.x",
-		"ios sdk": ">=8.0 <=11.x",
+		"xcode": ">=7.0 <=9.x",
```
vijaysingh-axway: should be <= 10.x

iphone/package.json
```diff
@@ -18,8 +18,8 @@
...
+		"ios sdk": ">=9.0 <=11.x",
```
vijaysingh-axway: should be <= 12.x

vijaysingh-axway (changes requested):
We should change Deployment target to 9.0 in project.pbxproj file.
And I have left some comments also . Please fix.

iphone/package.json
hansemannn: Thats handled in an own PR.

iphone/package.json
hansemannn: Same as above.

hansemannn:
@vijaysingh-axway Updated PR

hansemannn:
@vijaysingh-axway Updated the PR, but have also put it back in progress, since I want to fix all deprecation warnings as part of this PR as well.

hansemannn:
@vijaysingh-axway I have migrated / removed all deprecated API's that was warned about in Xcode. The only remaining 4 warnings are regarding `CFReadStream*` API's that are being replaced by #9732 already.

cb1kenobi:
I'm not sure if this PR catches ALL of the Xcode 8 logic in the iOS build, but I approve the current changes to the iOS build.

vijaysingh-axway (changes requested):
@hansemannn  I can see couple of more deprecation warnings other than CFReadStream* -
1. Deprecation in TiUIListView.m (appearanceWhenContainedIn)
2. Deprecation in MediaModule.m (UIPopoverController)
3. Deprecation in ContactsModule.m (ABAddressBookRef)

Can you please resolve them and conflicts in the branch. Thanks!

hansemannn:
Meeh, seems like Xcode cached a few warnings so they didn't show up. Filing a radar for that and updating the PR.

hansemannn:
@vijaysingh-axway Updated all remaining references and created [TIMOB-26129](https://jira.appcelerator.org/browse/TIMOB-26129) for you since you extensively worked on the `NSURLSession` API before.

vijaysingh-axway (approved):
CR approved.

sgtcoolguy:
@hansemannn I updated and rebased this PR on next branch. However, the two http client unit tests on iOS consistently fail on each rebuild.

Any ideas? This gets spit out on stderr for each:
```
[WARN] : failed, attempting to retry request...
[ERROR] : send can only be called if client is disconnected(0). Current state is 4
```

sgtcoolguy:
@hansemannn I pushed a fix for the compile error on TiUiScrollView.m, you missed removing the corresponding #endif for a macro block (you removed the #if portion).

Second, after taking a look at the Network tests that are failing, they are specifically tests which use query params. I think the changes to TiUtils.m is likely the root cause of these failures. Not sure which specific changes there yet. The most obvious place to start would be the change to TiUtils#encodeQueryPart()

sgtcoolguy:
@hansemannn I've pushed a couple changes that *should* fix the url encoding issue those tests were surfacing. The relevant change is here: 
https://github.com/appcelerator/titanium_mobile/pull/10103/commits/a2e74299dcdb285192eba67035a323e546bd0a7d

I'd love if you could come up with some idea of other tests we might want to write to cover the changes there. I think the new code is simpler and should work OK, but my first stab at it *did* uncover some broken behavior when it resolved an image file in a Ti.UI.ImageView test - it was around passing in a Ti.Filesystem.File as the "relative path" where it tries to handle special cases of paths looking like they're absolute but that should be resolved relative to the "app:" root (the Resources dir). It was basically pre-pending the resources path to the front without checking to see if the path really was an absolute path. My fix was to check if the path already had the resources path as a prefix, but maybe we should just test if the "relativePath" is really a valid absolute path and if so not try to turn it relative to the resources dir.

hansemannn:
iOS passed now:
```
1282 Total Tests: 1073 passed, 0 failed, 209 skipped.
```
and Android sometimes errors with random errors, which seem unrelated. I can write more tests around the resource handling and add those to this PR.

PR #10104, opened 2018-06-13
[TIMOB-25942] : iOS 11: Screen orientation not working in navigation window

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25942


iphone/Classes/TiRootViewController.m
```diff
@@ -1178,8 +1178,15 @@ - (void)refreshOrientationWithDuration:(id)unused
 #ifdef FORCE_WITH_MODAL
     [self forceRotateToOrientation:target];
 #else
-    [self rotateHostingViewToOrientation:target
-                         fromOrientation:[[UIApplication sharedApplication] statusBarOrientation]];
+    if ([TiUtils isIOS11OrGreater]) {
+      forcingStatusBarOrientation = YES;
+      [[UIApplication sharedApplication] setStatusBarOrientation:target animated:NO];
```
hansemannn: This method is deprecated in iOS 9+, we should use a wrapper to use the iOS 10+ API for newer devices.

hansemannn (changes requested):
One change required. Also, please make sure to test this on split-windows and different devices as well. Also, test using our kitchensink-v2 sample and larger sample apps.

vijaysingh-axway:
Apple has deprecated setStatusBarOrientation method in iOS9+ and haven't given any direct method to set status bar orientation. Need to check some different way to do this.

iphone/Classes/TiRootViewController.m
hansemannn: @vijaysingh-axway Can you add the `[TiUtils isLowerThan:@"12.0"]` (or how the method was named) check to only add this workaround for iOS 11?

PR #10105, opened 2018-06-13
[TIMOB-26090] iOS: Add Xcode 12 dev-env

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26090

### DIFF
--- a/iphone/package.json
+++ b/iphone/package.json
@@ -18,7 +18,7 @@
 	"minIosVersion": "8.0",
 	"minWatchosVersion": "1.0",
 	"vendorDependencies": {
-		"xcode": ">=6.0 <=10.x",
+		"xcode": ">=8.0 <=10.x",
 		"ios sdk": ">=8.0 <=12.x",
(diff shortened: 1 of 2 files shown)


iphone/package.json
```diff
@@ -18,8 +18,8 @@
 	"minIosVersion": "8.0",
 	"minWatchosVersion": "1.0",
 	"vendorDependencies": {
-		"xcode": ">=6.0 <=9.x",
```
hansemannn: @sgtcoolguy Not bumping Xcode here, but we should require Xcode 8+ in Ti SDK 8.0.0, since we actually do via code already.

cb1kenobi: These ranges are used by `ti info` to detect issues. I'd be fine with bumping this to `>=8.0 <=10.x`.

sgtcoolguy (approved):
Good to merge once that Xcode min is bumped to 8.0 in the package.json as per Chris Barber's comment.

PR #10111, opened 2018-06-15
Updated TabGroup.yml

jawa9000:
Update per https://jira.appcelerator.org/browse/TIDOC-3202: Tabs are not supposed to fire click, singletap and other UI-related events, since they are no Ti.UI.View subclass.


hansemannn (changes requested):
We should exclude them via the `excludes` tag instead of in the description.

hansemannn:
@jawa9000 I've updated the docs to use the `excludes` tag and manually going through the code to ensure all actual events are being listed.

hansemannn (approved):

PR #10112, opened 2018-06-16
[TIMOB-26015] Android/iOS: Refactored "app.js" launch handling

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-26015
- https://jira.appcelerator.org/browse/TIMOB-26144
- https://jira.appcelerator.org/browse/TIMOB-26128
- https://jira.appcelerator.org/browse/TIMOB-26124
- https://jira.appcelerator.org/browse/TIMOB-25899
- https://jira.appcelerator.org/browse/TIMOB-26217

**Summary:**
- Added a new "titanium_mobile/common/Resources" JavaScript folder.
  * Provides JS files/extensions intended to be shared by across all platforms.
- Android/iOS now launch using new `ti.main.js` instead of `app.js`.
  * Used to load JS extension/bootstrap scripts that are shared on all platforms.
  * Loads `app.js` once finished loading everything.
- Now logs app version and Titanium version on startup via `ti.main.js` instead of from native code.
  * New Message: `<AppName> <AppVer> (Powered by Titanium <TiVer>.<TiBuildHash>)`
- [TIMOB-26015](https://jira.appcelerator.org/browse/TIMOB-26015) Added bootstrap script support.
  * All files named `bootstrap.js` or end with name `*.bootstrap.js` under the "Resources" directory will be automatically required-in before `app.js` gets loaded. This is a recursive search.
  * This is the CommonJS equivalent of native Android module's `onAppCreate()` or iOS module's `load()` method.
  * Bootstrap script can support optional `exports.execute(finishedCallback)` method intended to display UI to end-user. Will not load `app.js` until given callback is invoked.
- [TIMOB-26124](https://jira.appcelerator.org/browse/TIMOB-26124) Fixed bug where `Ti.buildHash` always returned `null` on Android as of Titanium 6.0.0.
  * Was a typo in `scons-build.js` and `scons-cleanbuild.js` scripts.
  * Fix needed by new Titanium version logging on app startup. (Build hash needed to differentiate between RC and GA versions.)
- [TIMOB-26128](https://jira.appcelerator.org/browse/TIMOB-26128) Modified Android build system to include asset directory names have a leading underscore '_'.
  * This was an intentional limitation in Google's `aapt` Android command line tool.
  * Solved by using `appt` command line argument `--ignore-assets` and omitting `<dir>_*` from filter list.
- [TIMOB-25899](https://jira.appcelerator.org/browse/TIMOB-25899) Modified JS `Error` object to expose non-enumerable properties via `JSON.stringify()`.
  * Was always returning `"{}"` on Android since none of the V8 `Error` object properties are enumerable.
  * Also modified to reveal `stack` property on iOS. (Not enumerable in JavaScriptCore.)
  * Solved by adding `toJSON()` method to `Error` object prototype. (See: `./common/Resources/ti.internal/extensions/Error.js`)
  * Implementing this was not needed for this PR, but it's a proof of concept on how JS prototype can be extended on all platforms in Titanium's core JS code.
- [TIMOB-26144](https://jira.appcelerator.org/browse/TIMOB-26144) Significantly improved Android's performance in accessing files/directories under the "Resources" directory.
  * The `File.getDirectoryListing()` performance is now about 20x faster. The `isDirectory()`, `isFile()`, and `existence()` calls have been improved too.
  * Now stores a hash table of all file/directory paths under the APK's "assets" folder for quick existence checks. Paths are now obtained by accessing APK as a zip file and fetching its zip entries, which is significantly faster than `AssetManager.list()` calls.
- [TIMOB-26217](https://jira.appcelerator.org/browse/TIMOB-26217) iOS: Added missing `Ti.Filesystem.File` methods `isFile()` and `isDirectory()` when referencing an encrypted asset.
  * Was causing JS runtime errors in bootstrap recursive search.
- [TIMOB-16678](https://jira.appcelerator.org/browse/TIMOB-16678) Modified Android `File.read()` to not return an empty string when referencing an encrypted asset.
  * Needed to access new `bootstrap.json` for device/production builds.

**Notes:**
- I purposely did not document the new `*.bootstrap.js` handling in our API docs. We should dog-food it first to see how much we like it before setting it in stone. Plus we may want it to use `Promises` in the future (this feature is missing on iOS 9 at the moment).
- We plan on having our `ti.playservices` module take advantage of the bootstrapping feature in 7.4.0. It's needed to check/install/update "Google Play Services" before loading the `app.js`.

**Test:**
Perform the below tests on the following:
- Android device without LiveView _(This tests encrypted asset handling.)_
- Android emulator _(This tests APK's "assets" handling.)_
- iOS device
- iOS emulator
- iOS 8 _(This tests oldest JavaScript engine version handling.)_

Test procedure:
1. Build and run the app with the below `app.js` and `ui.bootstrap.js` code. Note that the `ui.bootstrap.js` needs to be next to the `app.js` file under the `Resources` directory.
2. On app startup, verify that an alert is displayed stating that it was launched by a bootstrap.
3. Tap on the "OK" button.
4. Next the `app.js` will have loaded, displaying a JavaScript `Error` object in JSON form. _(This verifies that our core JS extensions have been loaded.)_
5. If running on iOS, verify that the displayed JSON contains a "stack" entry. (This used to be excluded on iOS.)
6. If running on Android, verify that the displayed JSON is not an empty `"{}"`. The JSON should display entries "message" and "stack".
7. If running on Android, press the "Back" button and then restart the app. Verify that the bootstrap alert dialog (mentioned in step 2) is shown again.

---
**app.js**
```javascript
	layout: "vertical",
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
scrollView.add(Ti.UI.createLabel({
	text: "JSON.stringify(new Error('My error message'))",
	width: Ti.UI.FILL,
	top: "10dp",
}));
scrollView.add(Ti.UI.createLabel({
	text: JSON.stringify(new Error('My error message'), null, 4),
	width: "100%",
}));
window.add(scrollView);
window.open();
```

---
**ui.bootstrap.js**
```javascript
exports.execute = function(finished) {
		title: 'Alert Dialog',
		message: 'This dialog was launched by a bootstrap script.',
		buttonNames: ['OK'],
		cancel: 0,
	});
	dialog.addEventListener('click', function(e) {
		finished();
	});
	dialog.show();
};
```


android/titanium/src/java/org/appcelerator/titanium/io/TiResourceFile.java
```diff
@@ -31,34 +31,32 @@
...
+		this.path = (path != null) ? path : "";
+		fetchName();
 	}
 
 	@Override
 	public boolean isDirectory()
 	{
-		if (statsFetched) {
-			return this.exists && this.typeDir;
+		if (this.statsFetched == false) {
```
sgtcoolguy: Any particular reason there's a companions to false rather than typical `!this.statsFetched`?

templates/core/Resources/_ti/bootstrap.loader.js
```diff
@@ -0,0 +1,103 @@
...
+					resourceDirectoryPath.length,
+					(bootstrapPath.length - resourceDirectoryPath.length) - JS_EXTENSION.length);
+				bootstrapScripts.push(bootstrapPath);
+			}
+		}
+	}
+	fetchBootstrapScriptsFrom(resourceDirectory);
+
+	// Sort the bootstrap scripts so that they'll be loaded in a consistent order between platforms.
+	bootstrapScripts.sort();
```
sgtcoolguy: Maybe some docs/help on what exactly the order will be? Obviously different OSes may sort differently naturally, so this is a potential point of confusion.

templates/core/Resources/_ti/bootstrap.loader.js
```diff
@@ -0,0 +1,103 @@
...
+	BOOTSTRAP_SUFFIX = '.' + BOOTSTRAP_FILE_NAME;
+
+exports.loadAsync = function (finished) {
+	var resourceDirectory = Ti.Filesystem.getFile(Ti.Filesystem.getResourcesDirectory()),
+		resourceDirectoryPath = resourceDirectory.nativePath,
+		bootstrapScripts = [];
+
+	// Fetches all "*.bootstrap.js" files under given directory and adds them to the "bootstrapScripts" array.
+	// The names stored in the array can be loaded via require() method.
+	function fetchBootstrapScriptsFrom(file) {
```
sgtcoolguy: Why not move these function declarations out to top-level?

sgtcoolguy (commented):
Some general comments:

- The location in the repo of these common JS files still feels odd to me. I'm not a fan of shoving it under a templates subfolder. If this is truly cross-platform public API, then it should probably be at a top-level folder as a sibling to iphone/android (i.e. `titanium_mobile/common`)
- Also, the name of `ti.shell.js` feels off to me.
- How do we handle errors in the bootstrapping? Looks like we don't really have any error-handling here, so I guess we basically blow up?
- Is there a noticeable performance hit? I know that Node basically writes the js code in "core" into a C array to avoid file I/O to improve perf there (and that's what we do in Android for some situations, I believe?)
- Do we anticipate platform-specific "bootstrap" code, or API extensions? Would they just get placed under platform specific sub-folders as in normal apps? We already have core JS code for Android, do we anticipate trying to incorporate that, or leaving it as-is?

I'm also concerned that we have multiple avenues of extending the app/API via common JS code - obviously there's CommonJS modules, but they wouldn't have the opportunity to be involved in the bootstrapping process right now. Is this expected to be limited to us SDK developers to be contributing these? It appears that a developer could potentially place files in a specific location in their app to play along.

jquick-axway:
@sgtcoolguy, thanks for your feedback.

> The location in the repo of these common JS files still feels odd to me.

I agree. Although after doing a packing/install, a "common" folder already exists and we've put a "simplejson" tool there (I don't agree with this; should be under a "bin" or "tools" folder). How about I put the shared JS files under "titanium_mobile/core"?

> the name of ti.shell.js feels off to me.

The name seemed appropriate since Titanium is a JavaScript wrapper for the app and the OS' native APIs, similar to how a Unix shell provides a script interface to the OS' native APIs. The "ti.shell" wraps the "apps.js", controls how the app launches, and provides our own custom JS API extensions. I kind of like the name. Do you have another name in mind?

> How do we handle errors in the bootstrapping?

We don't. This is intentional and makes it easier to debug the bootstrap script.

The `bootstrap.execute()` method is mostly intended to display UI to the end-user and it might be a very long operation. On Android, uses-cases would be to show UI to download/install Google Play Services and in the future Google Play Expansions files which can be up to 2 GB. Since this can be a long job, we should not impose any timeouts. And the bootstrap should be allowed to back/exit out of the app too without loading the "app.js" in case the end-user refused to do the download or is unable to (no Internet connection).  Bootstraps like these need to guarantee that their job was completed before loading the "app.js" or else you'll run into other problems. And any bugs in a bootstrap script that causes the app to "blow-up" will need to be fixed.

> Is there a noticeable performance hit?

Fair question. I'll do some more measurements/testing and get back to you on this.

As a quick summary for now, I have a test app that has about 160 JavaScript files in the root Resources directory. The amount of files in the directory listing is definitely a factor. The overhead is about 100 ms in the iOS Simulator and 140-1000 ms on various Android device. The 1 second overhead  came from my slowest Android device that takes 10 seconds to start an app without my changes (I'm not happy about this). This can be optimized.

> Do we anticipate platform-specific "bootstrap" code or API extensions?

Platform Specific:
- The "String.js" extension I added to this PR adds the missing ES6 functions for iOS 8. It isn't needed on Android or Windows.
- I plan on adding a bootstrap script to our Android-only "ti.playservices" module. (It's injected by the module, making it a non-issue.)

Shared:
- The "Error.js" extension I added allows `JSON.stringify(Error)` to work in V8 and exposes the "stack" property in JavaScriptCore.
- I plan on adding a bootstrap script to "ti.verify" so that it can do module verification in JavaScript instead of native code. (We can discuss this one offline.)
- In the future, extend some of our core Titanium libraries and prototype in JavaScript by dog-fooding our own JS APIs.

I don't anticipate the "core/Resources" folder to be used for platform specific things except for core JS APIs that may have to do some if-android or if-ios types of things. It wouldn't be any worse than a CommonJS module. Ideally, we should keep platform specific code to a minimum.

I did see our Android specific JavaScript code/extension and was going to leave that be for the moment.

> I'm also concerned that we have multiple avenues of extending the app/API via common JS code

The bootstrap concept is my JavaScript equivalent of our native Android module `onAppCreate()` and native iOS module `load()` method. I mostly anticipate it being used by CommonJS modules or hybrid modules. Such as with the "ti.playservices" module. I don't anticipate us using bootstraps in the core since the "ti.shell.js" can simply load what it wants on app startup before the "app.js". The one exception might be "ti.verify".

The "extensions" folder I created is a separate thing. This is only for us and is meant for us to add APIs to our core libraries/prototypes for all platforms. It's an opportunity for us to be more productive and improve parity. If a new feature can be 100% coded in JavaScript, then we should do it instead of implementing it natively on each platform we support. For example, if you want to extend "console" with new logging/performance APIs, then you can do it here and it'll just-work on all platforms.

templates/core/Resources/_ti/bootstrap.loader.js
infosia: I'm curious about that too. I'm not quite sure if calling JavaScript _Array.sort()_ without any arguments guarantees the order across platforms actually. I see that [Mozilla describes it's sorted according to Unicode code point](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) and Microsoft sorts it in [ASCII character order](https://docs.microsoft.com/en-us/scripting/javascript/reference/sort-method-array-javascript), and ECMAScript spec says it orders [according to _ToString()_ comparison](https://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.11) (Do they describe same thing? I'm not sure). Maybe we might want to place _compareFunction_ explicitly to avoid confusion.

infosia: I was thinking that we want to give scripts a chance to define order of the bootstrap script by defining priority like `export.priority` but for now I'm almost fine with current implementation. I think simply sorting files according to file names would work well for developers. For instance in case developer wants to bootstrap network scripts before UI scripts, just rename bootstrap files like

```
001.something_to_boot_Network.js
002.something_to_boot_UI_theme.js
003.google_play.js
```

infosia:
I tried to integrate **ti.shell.js** on Windows, I made it working and the sample code **ui.bootstrap.js**  works good locally. Looks good 👍

jquick-axway:
Thanks for your feedback @infosia.

Would you mind doing a quick performance test for me on Windows please? Add about 100 files to your project and then use the following "ti.shell.js" code to log how long it takes. Note that if your build script copies the core scripts to the "build" directory before copying over the project's scripts, then you don't have to modify the "ti.shell.js" under "titanium_mobile". You can override it by adding a "ti.shell.js" to your Titanium app project instead since it would overwrite it. I consider this a feature and I use it to quick testing of the core scripts I've made (particularly the extensions).
```javascript
'use strict';

Ti.API.info(Ti.App.name + ' ' + Ti.App.version + ' (Powered by Titanium ' + Ti.version + '.' + Ti.buildHash + ')');

var startTime = new Date();
require('./_ti/extension.loader').load();
Ti.API.info("@@@ Extension Loader Duration: " + (new Date() - startTime));

startTime = new Date();
require('./_ti/bootstrap.loader').loadAsync(function () {
	Ti.API.info("@@@ Bootstrap Loader Duration: " + (new Date() - startTime));
	require('./app');
});
```

Ideally, I'd like to leave my bootstrap loader as-is and hunt down bootstrap files dynamically on startup... provided that the performance is fine on all platforms. Alternative solution is to modify our build tools to hunt down the bootstrap files when we do a build and store it to file for quick lookup on app startup instead, but I'm hoping to avoid that for the sake of simplicity.

templates/core/Resources/_ti/extensions/String.js
```diff
@@ -0,0 +1,80 @@
+/**
```
garymathews: We don't need this anymore due to https://github.com/appcelerator/node-titanium-sdk/pull/32

Unless this is for example purposes?

garymathews (commented):
How will we handle platform specific scripts?

Could we have platform directories?
```
- _ti
    - android
    - ios
    - windows
```

Could we modernize the JS a bit by using `const` and `let` where appropriate?

infosia:
Is there any special reason to use leading underscore `_` for the name of bootstrap folder (`_ti`)? On Windows it causes compiler warnings because Visual Studio tries to auto-generate Resources for the files when directory name has leading underscore. It doesn't stop the app building but I don't see special reason to stick with leading underscore when it tend to be treated specially on Windows as well as Android.

```
GENERATEPROJECTPRIFILE : warning PRI263: 0xdef01051 - No default or neutral resource given for 'Files/extensions/Error.js'. The application may throw an exception for certain user configurations when retrieving the resources.
```

jquick-axway:
@infosia, I'm using an `_ti` directory to help avoid name collision with the app developer's files/folders. Kind of like how C/C++ uses leading underscores to avoid name collisions. And I wanted to avoid using a leading period like `.ti` to avoid trouble on Unix-like systems.

On Android, only the "res" directory has character/name limitations since they're turned into Java constants. The Android "assets" directory (where we copy the app project's files) doesn't have any restrictions other than the filter list that is intended to exclude folders such as ".git", ".hg", ".svn", etc.

I wasn't aware of any special treatment of leading underscores in WinRT. I'm guessing we're not adding these files under its equivalent "assets" directory? If we were to ever add images under the `_ti` directory would we run into problems?

infosia:
FYI Tried 100 bootstrap files on WindowsStore app (Intel Core i7-6700 CPU @ 3.40GHz, 16.0GB memory) and it took 14.5 seconds to load.

```
[INFO]  @@@ Bootstrap Loader Duration: 14672
```

infosia:
How about using fully qualified name like Android packages? Such as `com.axway.appcelerator.titanium.boostrap`? Or, we could just use Titanium package-like name such as `ti.app.bootstrap` because we believe we are casually using this kind of package name to save app preferences globally like 'ti.ui.defaultunit'.

jquick-axway:
@infosia, I'm sorry. I didn't mean for you to test 100 bootstrap files. I meant for you to add 100 other file types such as txt files. This is to test how fast (or slow) the recursive directory traversal is.

For example, my app project contains about 200 files. About 25% of them are image files and most of the rest are JS files (not bootstrap related). I only have 2 bootstrap files in my project. (On Android, the `File.getDirectoryListing()` and existence checks under the APK's `Resources` directory is slower than it is from storage when it shouldn't be and I'm resolving it now.)

Sorry for not being specific.

infosia:
Interestingly `File.getDirectoryListing()` takes most of loading time on Windows (14sec). I also observed similar to Android, getting files under appx takes much time (10x slower) than getting files under normal directory. I'm curious how Android is going to take care of this.

hansemannn:
If we let the developer place the bootstrap-files in an own directory (like we plan for the SDK itself as well), wouldn't that cut down the number of files to search ? Right now, the performance would be a huge blocker.

infosia:
Ah I think I get it, Titanium CLI should be able to generate file lists (in JSON format?) under Resources directory beforehand at compile time. That way we don't have to call `File.getDirectoryListing()` on runtime. Nice idea! 👍

jquick-axway:
I've just updated the PR to improve Android's "Resources" file/directory access performance. Recursive directory traversal under the "Resources" directory is now about 7x faster than before. And the `getDirectoryListing()` method is now about 20x faster than before for "Resources".

Here are the performance differences for startup times with my project having 200 files...
- Pixel XL: `142 ms` -> `84 ms` (1.6x faster)
- Amazon Fire HD 8: `389 ms` -> `89 ms` (4.3x faster)
- Galaxy Nexus: `1187 ms` -> `230 ms` (5.1x faster)
- Android 4.1 Emulator: `964` -> `58 ms` (16.6x faster)

It's a pretty significant improvement and all Android devices (except Galaxy Nexus) now outperform iOS. The Galaxy Nexus is our slowest device in the office and takes about 10 seconds to launch an app normally (without this PR's changes), so, this is still a "win" on that device as well. Also, the emulator got the biggest performance gain because JS files are not encrypted and are instead dumped to the APK's "assets" directory which is what I just super optimized.

@infosia, I'm going to see about adding optional support for a JSON of JavaScript files that the "_build.js" can create and bundle with the app. And if the JSON file is not there, then fallback to my existing `getDirectoryListing()` code... which might need to keep for LiveView.

@hansemannn, we need to be able to find bootstrap files belonging to modules and each module is put under their own subdirectory. Because of this, a recursive lookup is required. My master plan is to add a `ti.playservice.bootstrap.js` file to our `ti.playservices` module.

infosia:
@jquick-axway I did quick trial to store file list of `Resources `directory at compile time in JSON file on Windows, but I ended up finding it does not actually produce same result when compared to files in final application package...It is because Windows package manager stores additional files while creating application package, such as DLL dependencies and Windows metadata files. I would guess similar things could happen on Android so I think your current implementation (unzip package and store lists beforehand) is the better working solution now.

iphone/iphone/Titanium.xcodeproj/project.pbxproj
```diff
@@ -2520,7 +2520,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cp -Rf \"$PROJECT_DIR/../Resources/.\" \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app\"\n\"$PROJECT_DIR/../../support/dev/localecompiler.py\" \"$PROJECT_DIR/..\" ios simulator \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app\"\nmkdir -p \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/modules/ui/images\"\ncp -Rf \"$PROJECT_DIR/../Resources/modules/ui/.\" \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/modules/ui/images\"";
+			shellScript = "cp -Rf \"$PROJECT_DIR../../templates/core/Resources/.\" \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app\"\ncp -Rf \"$PROJECT_DIR/../Resources/.\" \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app\"\n\"$PROJECT_DIR/../../support/dev/localecompiler.py\" \"$PROJECT_DIR/..\" ios simulator \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app\"\nmkdir -p \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/modules/ui/images\"\ncp -Rf \"$PROJECT_DIR/../Resources/modules/ui/.\" \"$TARGET_BUILD_DIR/$PRODUCT_NAME.app/modules/ui/images\"";
```
cb1kenobi: @hansemannn can you check this out? I know the iOS build removes the pre-compile phase (https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L3080-L3094) so this would affect you as a Titanium platform developer working directly from Xcode.

cb1kenobi: I'm also unsure if `$PROJECT_DIR` has a trailing slash.

templates/core/Resources/_ti/bootstrap.loader.js
```diff
@@ -0,0 +1,103 @@
...
+			// Last bootstrap has finished execution. Time to load the next one.
+			// Note: Add a tiny delay so whatever UI the last bootstrap loaded has time to close.
+			bootstrapIndex++;
+			setTimeout(function () {
+				doLoad();
+			}, 1);
+		}
+		doLoad();
+	}
+	loadBootstrapScripts(function () {
```
cb1kenobi: This is way too much code and it's too hard to follow. You have scope declared functions and functions that don't need to exist.

A simpler approach would be to do something like the following:

```js
bootstrapScripts.reduce((promise, script) => {
	return promise.then(() => {
		const bootstrap = require(script);
		if (typeof bootstrap.execute === 'function') {
			return new Promise(resolve => bootstrap.execute(() => setTimeout(resolve, 1)));
		}
	});
}, Promise.resolve());
```

That reduces 42 lines of code down to 8 and that's a win.

templates/core/Resources/_ti/extension.loader.js
```diff
@@ -0,0 +1,39 @@
...
+ * This loader is expected to be ran before executing the "app.js" file. Intended to
+ * share Titanium's core JavaScript code/extensions on all platforms.
+ */
+
+'use strict';
+
+var JS_EXTENSION = '.js',
+	DIRECTORY_PATH = '_ti/extensions/';
+
+exports.load = function () {
```
cb1kenobi: There's a lot of code in this function that can be simplified.

I'm not sure why you are checking if `fileNames` and `fileName` is truthy. Why would it not be? Isn't that a bug if `getDirectoryListing()` returns something falsey?

Use a for-of loop. It's faster and cuts down on all the cruft.

Use a regex to check if the file is a `.js` file. It's way cleaner and faster than doing all that string manipulation.
```js
const DIRECTORY_PATH = '_ti/extensions/';

exports.load = function load() {
	const dir = Ti.Filesystem.getFile(Ti.Filesystem.getResourcesDirectory(), DIRECTORY_PATH);
	const jsRegExp = /^(.*)\.js$/;

	for (const filename of directory.getDirectoryListing().sort()) {
		const m = filename.match(jsRegExp);
		if (m) {
			require(DIRECTORY_PATH + m[1]);
		}
	}
};
```

templates/core/Resources/_ti/extensions/Error.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+'use strict';
+
+// Add a toJSON() method to all Error objects needed to output non-enumerable properties.
+// The JSON.stringify() will automatically call this method if it exists to provide custom output.
+// Notes:
+// - In V8, all Error properties are not enumerable. We need this or else stringify() will return "{}".
+// - In JavaScriptCore, only the "stack" property is not enumerable. We want to reveal this.
+if ((typeof Error.prototype.toJSON) !== 'function') {
```
cb1kenobi: No need for the `(` and `)` around the typeof operator.

templates/core/Resources/_ti/extensions/Error.js
```diff
@@ -0,0 +1,23 @@
...
+	Error.prototype.toJSON = function () {
```
cb1kenobi: You should always name anonymous functions.

templates/core/Resources/_ti/extensions/Error.js
```diff
@@ -0,0 +1,23 @@
...
+		var properties = {};
+		Object.getOwnPropertyNames(this).forEach(function (name) {
```
cb1kenobi: Use a for-of loop. It's cleaner and faster. Functional programming with `forEach()` was nice back in the day, but today for-of loops are superior.

```js
const props = {};
for (const prop of Object.getOwnPropertyNames(this)) {
    props[prop] = this[prop];
}
return props;
```

templates/core/Resources/_ti/extensions/String.js
```diff
@@ -0,0 +1,80 @@
...
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2018 by Axway, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+'use strict';
+
+// Determines if given object is not of type function.
+function isNotFunction(object) {
```
cb1kenobi: I'm not sure a function dedicated to determining if something is a function is beneficial. It's such a simple statement.

From a performance perspective, it's fine in this case, but I found that if you were to do something like this on a greater scale, it would be come an extremely hot function and thus should be avoided.

cb1kenobi (changes requested):

cb1kenobi:
How does this PR affect LiveView? With the introduction of this new bootstrap mechanism, does that mean we can stop injecting LiveView on the first line of the `app.js`?

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -138,4 +250,117 @@ public static String getCacheDir()
...
+		public void addPath(String path)
+		{
+			// Null path references the root directory.
+			if (path == null) {
+				path = "";
+			}
+
+			// Remove path separator(s) from the front of the path. It must be relative down below.
+			int startIndex = 0;
+			for (; (startIndex < path.length()) && (path.charAt(startIndex) == '/'); startIndex++);
```
garymathews: Couldn't we just do:
```Java
int startIndex = path.indexOf('/');
```
?

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -114,21 +184,63 @@ public static String readFile(String path)
...
+			Collection<String> collection = encrpytedDirectoryListingMap.get(normalizeAssetCryptPath(path));
+			if (collection != null) {
+				resultList.addAll(collection);
+			}
+		}
+		{
+			Collection<String> collection = apkDirectoryListingMap.get(path);
+			if (collection != null) {
+				for (String entry : collection) {
+					if (!resultList.contains(entry)) {
```
garymathews: Maybe we could swap the order of this method, so we are doing the `!resultList.contains(entry)` on the most likely smaller collection from `encrpytedDirectoryListingMap` ?

Just a thought I had

jquick-axway:
@cb1kenobi, thanks for your feedback.

Since transpile/polyfill is disabled by default and it can also be disabled via the "tiapp.xml" setting `<transpile/>`, I have to stick to ES5 in our core JS code for the moment. Particularly for iOS 8 and iOS 9 support.

I've written up a ticket here [TIMOB-26146](https://jira.appcelerator.org/browse/TIMOB-26146) requesting that we enable transpile/polyfill by default in Titanium 8.0.0. This would mean that we would need to resolve the remaining issues that's preventing us from doing so. But once done, I'm all for changing our core JS code over to ES6 as you've suggested. And also to switch over our bootstrap `execute()` method to use promises instead. :smile:

> How does this PR affect LiveView? With the introduction of this new bootstrap mechanism, does that mean we can stop injecting LiveView on the first line of the app.js?

Yes. Although bootstrap order might matter, which is something I wasn't planning on addressing yet. We may want to guarantee that the LiveView bootstrap gets loaded before any of the other ones. We could cheat for the moment and have our core code invoke the LiveView bootstrap before any other if found. @infosia suggested that we assign priority numbers to bootstraps, but I worry that everyone will want to be priority 1. I'm thinking a bootstrap should instead define what other bootstraps it depends on by name (such as a bootstrap that depends on our Google Play Services bootstrap to get everything up-and-running), but that doesn't solve having LiveView come in first in the list (special handling might be okay in this case).

templates/core/Resources/_ti/extension.loader.js
jquick-axway: We have a parity issue with `getDirectoryListing()`. If the directory is empty, iOS returns `null` and Android returns an empty array. Personally, I prefer the empty array since the app developer may forget to do a `null` check. CC: @hansemannn 

I benchmarked the regex approach. The performance was about the same, but the regex approach is simpler so I'll roll with it. Thanks man.

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
jquick-axway: I'm doing this in case there are 2 (or more) slashes "//" next to each other. Kind of like what Hans is asking me to do above with `$PROJECT_DIR`. :)

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
jquick-axway: Good idea. Especially since the `ArrayList.contains()` is an `O(n)` lookup.

iphone/iphone/Titanium.xcodeproj/project.pbxproj
jquick-axway: I'll add an extra slash.

sgtcoolguy:
@jquick-axway So there's a relevant difference here - transpiling of *user* code is off by default right now. There's no reason we can't transpile our own core/bootstrap code either during SDK compile time, or on-demand during the user app build process. Obviously if we choose the latter and the user doesn't have transpilation turned on, that makes the cli build logic need to be more complicated. If we choose the former, then we need to update our scans/build logic to do the transpile before copying the transpiled scripts.

templates/core/Resources/_ti/extension.loader.js
sgtcoolguy: Sounds like an easy parity fix for iOS. I'll take a quick look today.

cb1kenobi:
@sgtcoolguy That's what I said to @jquick-axway in a private chat. Webpack would be a great tool for this as we can transpile and bundle the code up into a single file.

sgtcoolguy:
@jquick-axway RE: getDirectoryListing() and a static pre-populated file listing. Don't we already have a mechanism for exactly this? I distinctly recall when working on the `require()` logic that we (at least under certain circumstances, like production builds?) had a JSON file with the listing of files? I think it is there for handling when we do encryption of JS files and the code is not actually in a file anymore?
cc @cb1kenobi

templates/core/Resources/_ti/extension.loader.js
cb1kenobi: I found Ti.Filesystem's API parity is pretty bad when I was building Mobile Web. About 2 years ago Frankie cleaned up Android's Filesystem so I had hoped that would resolve the discrepancies, however I forgot that iOS's implementation isn't perfect either.

cb1kenobi:
@sgtcoolguy The static directory listing is only for encrypted JavaScript files.

What do you mean the code is not there anymore?

templates/core/Resources/_ti/extension.loader.js
sgtcoolguy: I opened up #10128 to address the parity issue and add unit tests here.

jquick-axway:
@sgtcoolguy, @cb1kenobi,

Right. We can transpile our core JS code, but what about the polyfill such as for promises? If we do the polyfill too, then we might as well as enable polyfill support by default for all app builds since they will already be getting this injected JS code. And if we require a bootstrap script to use promises, then CommonJS modules that contain bootstrap scripts will need this too.

Alternatively, we can just bundle the polyfill that we need, like ES6 Promises and Strings functions, and call it a day.

cb1kenobi:
I think we should always polyfill unless the API is already implemented. Don't make it a flag.

Secondly, don't reinvent the polyfills. Drop in https://www.npmjs.com/package/core-js and call it a day.

garymathews (commented):
Here's some of my thoughts for some discussion points:

#### Transpiling Bootstrap
I like the idea of supporting ES2018+ features but I don't think transpiling our bootstrap code is the best idea right now. Having independent transpilation processes for our bootstrap and app code sounds problematic.

The bootstrap code needs to work regardless of transpilation setting, so I'm okay with it being ES5 and not having all the fancy ES6+ syntax sugar. We can always update the code when we have figured out all of our transpilation pain points.

#### Polyfill
I can see why Josh decided to roll his own String prototypes. It's easier than having to include `core-js`. They don't override existing implementations. My concern is that [they need to be spec compliant](https://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype).

Otherwise, we should just define some helper methods for use by the bootstrap only, instead of implementing String prototypes.

As suggested by @cb1kenobi, the better option would be to always include `code-js` regardless of transpilation option. [Which would be pretty easy to do.](https://github.com/appcelerator/node-titanium-sdk/blob/master/lib/jsanalyze.js#L200)

sgtcoolguy:
@garymathews I think the larger issue here is that I had suggested using a Promise-based API for the bootstrap callback. That requires polyfilling Promises (for iOS 8/9), so we'd have to polyfill it in the bootstrap code itself. The pointer you gave for polyfilling in node-titanium-sdk is to copy it over for user code when transpile is turned on. Now we could bump that out to always happen regardless of the transpile flag. 

So there's a few options:
- for now we can simply use a callback-based API if this is proving to be too large of a pain.
- Drop iOS 8/9 support and the problem goes away! (I have raised this before on Teams, but the team seems reluctant to drop 9. The numbers already show it below our historical usage threshold for dropping support, and @eric34 also seems to be OK with the idea).
- Always polyfill, regardless of the transpile flag/value - then have our bootstrap API load the Promises polyfill.

garymathews:
+1 for the always polyfill

However, considering we don't use Promise anywhere else. Why should we start using it in this scenario? Should we not move over to Promises until we allow better adoption (Titanium APIs) of them?

jquick-axway:
Updated PR. Renamed "_ti" directory to "ti.internal".

@infosia, this will work-around the build warnings caused by the leading underscore for Windows app builds.

jquick-axway:
Updated PR:
- Added "bootstrap.json" support.
  * Optional feature where JSON file is expected to be created by app build system.
  * Avoids recursive directory traversal for bootstraps if JSON found. Huge performance boost.
- Removed "extension.loader.js" file.
  * Now requires-in extensions per file in "ti.main".
  * Changed because iOS `getFile()` and `getDirectoryListing()` does not support "encrypted" asset directories.

jquick-axway:
Updated PR:
- Modified Android build to generate a "boostrap.json" file.
- On iOS, added missing `Ti.Filesystem.File` methods `isFile()` and `isDirectory()` if file references an encrypted asset.
- On Android, fixed `File.read()` to not return an empty string when referencing an encrypted asset.

jquick-axway:
Updated PR:
- Modified iOS build to generate a "boostrap.json" file.
- Modified iOS `getFile()` to support acquiring encrypted JSON files for test/production builds.

garymathews:
@jquick-axway @cb1kenobi @sgtcoolguy I want to use this functionality to load ACA on startup ([TIMOB-25955](https://jira.appcelerator.org/browse/TIMOB-25955))

What's the status of this? I'd rather we not make use of Promises just yet until we adopt Promise support with our Titanium APIs which we can do at a later date. There are too many complications to address by supporting Promises that are beyond the scope of this PR.

What's left to get this merged?

hansemannn (approved):
The iOS changes look good and I love the idea of this! If our benchmarking-tests regarding performance hits will allow this to be integrated, it will be a great addition to the platform.

sgtcoolguy:
@jquick-axway @garymathews I'm good with the current state of the PR. The only change I could see would be to remove the unnecessary String extensions for ES6. (We'll be transpiling by default, polyfilling, and in 8.0.0 we're dropping iOS 8 support)

Josh, is it Ok to merge as-is?

jquick-axway:
@sgtcoolguy, I'll get rid of the "String.js" extension. Especially since (after all of my changes) I'm only using its `endsWith()` function in one place now. Thanks.

lokeshchdhry:
Testing in progress ......

Test comments in the appropriate tickets. 
1. https://jira.appcelerator.org/browse/TIMOB-26144 - FR Passed.
2. https://jira.appcelerator.org/browse/TIMOB-26128 - FR Passed.

jquick-axway:
PR updated:
- Removed ES6 String extension. Setting "tiapp.xml" `<transpile>true</transpile>` now supports polyfill too. This is the preferred solution.
- Internal use of `String.endsWith()` changed to `String.search()`. Only used by when a "bootstrap.json" not provided (slowest method) and is only used when debugging "titanium_mobile" via IDE.

lokeshchdhry:
Jenkins build fails. Can someone please take a look at it. Thanks !!

hansemannn:
@jquick-axway 
```
[ERROR] :  TiExceptionHandler: (main) [54,54] ti:/module.js:303
[ERROR] :  TiExceptionHandler: 	throw new Error('Requested module not found: ' + request); // TODO Set 'code' property to 'MODULE_NOT_FOUND' to match Node?
[ERROR] :  TiExceptionHandler:  ^
[ERROR] :  TiExceptionHandler: Error: Requested module not found: ./ti.internal/extensions/String
```

PR #10123, opened 2018-06-20
[TIMOB-26109] [TIMOB-26110] Use legacy build system for now

hansemannn:
**JIRA's:** 
- https://jira.appcelerator.org/browse/TIMOB-26109
- https://jira.appcelerator.org/browse/TIMOB-26110

Until [an issue](https://openradar.appspot.com/40906897) in the new build system is fixed, we can use the legacy build system. There is nothing wrong with the old one and the improvements using the new build system are mainly around Swift compile time, which was deadly slow in some circumstances before. But since our core is Obj-C based, it does not effect us at all. To unblock us from supporting Xcode 10, this PR will solve both module- and app-builds.


cb1kenobi (approved):

craigrouse:
@hansemannn moving my comment here: I'm experiencing this issue now (and it's my first time attempting to build a Titanium app). I'm using Appcelerator Studio. Is there a workaround for this problem that I can use right now?

hansemannn:
Hey again! So you are using Xcode 10 Beta? In this case, you should use the latest SDK 7.4.0 build (`appc ti sdk install -b master`) or apply this change manually.

craigrouse:
Thanks @hansemannn. I applied the change manually and got it working. Studio is really slow, though, so I've just switched to Atom with the plugin. Thanks for your support.

hansemannn:
Yey, have fun!

PR #10125, opened 2018-06-20
[TIMOB-24589] TiAPI: Remove Ti.Yahoo

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24589

> `no tests` since there were no tests and we are..well...removing it 😙


apidoc/Titanium/Yahoo/Yahoo.yml
```diff
@@ -14,6 +14,10 @@ methods:
       - name: callback
         summary: the function to execute when the query completes.
         type: Callback<YQLResponse>
+deprecated:
+    since: "8.0.0"
```
sgtcoolguy: didn't we deprecate this before?

sgtcoolguy (approved):
Should have been killed off a very long time ago...

apidoc/Titanium/Yahoo/Yahoo.yml
hansemannn: I think we planned to in 7 but never came to it

android/modules/contacts/src/java/ti/modules/titanium/contacts/CommonContactsApi.java
```diff
@@ -143,8 +143,6 @@ protected static String getImTextType(int type)
 			key = "custom";
 		} else if (type == ContactsContract.CommonDataKinds.Im.PROTOCOL_MSN) {
 			key = "msn";
-		} else if (type == ContactsContract.CommonDataKinds.Im.PROTOCOL_YAHOO) {
```
sgtcoolguy: Should this be removed? Does Yahoo still have IM? I don't think it's related to our Ti.Yahoo API.

android/modules/contacts/src/java/ti/modules/titanium/contacts/ContactsApiLevel5.java
```diff
@@ -286,8 +286,6 @@ protected int processIMProtocol(String serviceName)
 			return Im.PROTOCOL_QQ;
 		} else if (serviceName.equals("Skype")) {
 			return Im.PROTOCOL_SKYPE;
-		} else if (serviceName.equals("Yahoo")) {
```
sgtcoolguy: Should this be removed? Does Yahoo still have IM? I don't think it's related to our Ti.Yahoo API.

sgtcoolguy (changes requested):
I think you removed a couple yahoo related things from the contacts API that wren't tied to our Ti.Yahoo API in any way and should have been retained. 

Also, This should probably be targeting next branch now.

sgtcoolguy (approved):

PR #10126, opened 2018-06-20
[TIMOB-26142] TiAPI: Deprecate Ti.XML

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26142


android/modules/xml/src/java/ti/modules/titanium/xml/XMLModule.java
```diff
@@ -25,6 +25,7 @@
 import org.appcelerator.kroll.common.Log;
 import org.xml.sax.SAXException;
 
+// DEPRECATED IN SDK 8.0.0
```
sgtcoolguy: Might as well make use of the actual @deprecated annotation available in Java: https://stackoverflow.com/questions/9031832/how-to-declare-or-mark-a-java-method-as-deprecated

```java
/**
 * @deprecated Will be removed from core SDK in 9.0.0 and moved out to a native module
 **/
@Deprecated
@Kroll.module
public class XMLModule extends KrollModule
```

iphone/Classes/XMLModule.m
```diff
@@ -12,6 +12,15 @@
 
 @implementation XMLModule
 
+- (id)_initWithPageContext:(id<TiEvaluator>)context
+{
+  if (self = [super _initWithPageContext:context]) {
+    DebugLog(@"[WARN] The Ti.XML namespace has been deprecated in favor of a standalone module in SDK 8.0.0.");
```
sgtcoolguy: We should try and use a more formal way of marking the code as deprecated too: https://stackoverflow.com/questions/1443599/how-do-i-flag-a-function-as-being-deprecated-in-an-iphone-objective-c-header-fil

So, I guess in this case that'd mean marking all of the XML module/types methods/properties in the header(s) as deprecated?

sgtcoolguy (changes requested):
Please use the proper compiler annotations/attributes to mark the code as deprecated.

Additionally, I think this removal/deprecation raises the issue of any other APIs which may use the xml code under the hood. Specifically, Ti.Network.HttpClient.responseXML comes to mind.

I assume that would *also* need to be deprecated/removed? Maybe it's an API addition that the native module would make if loaded up?

I don't know if we have any real usage numbers for devs using xml via that route, as I don't think our API analytics would capture it properly.

hansemannn:
Not going to deprecate.

PR #10128, opened 2018-06-22
[TIMOB-26149] Android: Ti.Filesystem.File#getDirectoryListing() does not return null for file/non-existent directory

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26149

**Description:**
Adds unit tests around Ti.Filesystem.File#getDirectoryListing() to confirm behaviors.
- returns `null` if it's a file or doe not exist
- returns an empty `Array` for an empty directory
- returns an `Array` of `String`s for a typical directory containing files.

Modifies Android's implementation to match iOS/docs.


cb1kenobi:
I would argue that `getDirectoryListing()` should throw an `Error` if the file does not exist. This is what Node.js does for `fs.readdirSync()`.

jquick-axway (approved):
CR: Pass

PR #10129, opened 2018-06-22
[TIMOB-26141] Android: Prevent lookup of invalid reference key

garymathews:
- Prevent lookup of invalid reference key `0`
- Remove annoying error log `No valid root or current activity found for application instance`

###### TEST CASE
```JS
console.log(Ti.App.Android.R);
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26141)

### DIFF
--- a/android/runtime/v8/src/native/JavaObject.cpp
+++ b/android/runtime/v8/src/native/JavaObject.cpp
@@ -76,8 +76,7 @@ jobject JavaObject::getJavaObject()
 	}
 	if (useGlobalRefs) {
 		return javaObject_;
-	} else {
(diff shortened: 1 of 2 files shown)


ypbnv (approved):
CR: Pass

ypbnv:
@garymathews I tried to track why does it end with a refTableKey_ equal to 0, but I did not got it. Any idea about that?

garymathews:
@ypbnv `refTableKey_` is initialised as `0`. This is considered an invalid key, only keys greater than zero should be looked up. `getJavaObject()` is called before a Java object is attached, which causes this invalid lookup.

PR #10130, opened 2018-06-24
feat(android): add stopAnimation() to View

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-5931

**Optional Description:**

Android has no possibility to cancel/stop animations like moving an object from left to right before the duration is over. This PR adds a `stopAnimation()` method to `Ti.UI.View`

```

	text: "-",
	width: Ti.UI.SIZE,
	height: Ti.UI.SIZE
});

	backgroundColor: 'red',
	height: 100,
	width: 100,
	top: 0,
	left: 0
});

const btn1 = Ti.UI.createButton({
	title: 'Animate',
	width: 100,
	height: 40,
	bottom: 20,
	left: 10
});
btn1.addEventListener('click', function() {
	view.left = 0;
	view.animate(ani);
});

const btn2 = Ti.UI.createButton({
	title: 'Cancel',
	width: 100,
	height: 40,
	bottom: 20,
	right: 10
});
btn2.addEventListener('click', function() {
	view.stopAnimation();
	view.left = 0;
});

const ani = Ti.UI.createAnimation({
	left: 100,
	duration: 3000
})

ani.addEventListener("start", function() {
	lbl.text = "start";
});

ani.addEventListener("complete", function() {
	lbl.text = "complete";
})

ani.addEventListener("cancel", function() {
	lbl.text = "cancel";
})

win.add([view, btn1, btn2, lbl]);
win.open();
```


sgtcoolguy (commented):
Hey @m1ga Thanks for yet another PR!

Some general thoughts:
- It feels odd to be calling this on the view itself, but I suppose we don't really return an object to call this on otherwise (we don't return something like a `Promise` to represent the operation). We could potentially call it on a `Ti.UI.Animation` object that you passed into `Ti.UI.View#animate`. Makes me curious how this would play with an animation that transitions to a second view...
- It also doesn't really give you control over exactly *which* animation it'd try and stop. I suppose only one animation per view at a time is executing given our API.
- Nor is there any feedback as to whether/when it gets canceled. Seems like for parity there should be a `'canceled'` event fired when an animation is canceled. Or maybe at least a boolean return value for this method to indicate that it will attempt to cancel (i.e. if we know there's no animation or it finished already we'd return false)? I don't see a nice backwards-compatible way to modify the `complete` event to denote an animation was canceled/stopped.

m1ga:
@sgtcoolguy thanks for your feedback! As far as I remember you can do `animate(null)` on iOS to stop an animation, so that would be another way to stop the animation on Android. But I have to check that.
Yes, since there is only one animation playing I thought it would be good to stop it on the view. But I'm open for the `Ti.UI.Animation` way! Just means that you have to keep track of the animation object. In my case I couldn't access it in the 2nd button.
A callback sounds like a good idea. I'll add a `cancelled` event, so we have `start`, `complete` and `cancelled` to listen for.

m1ga:
I think I can't make something like `animation.stop()` since it looks like that there is no connection between `TiAnimation` (just storing the parameters) and the actual TiAnimationBuilder that is created from the view where the animation is created

sgtcoolguy:
@m1ga Sorry about the delay in responding!

Ok, yeah it makes sense that We call stopAnimation() on the view then. I'd still like to see some sort of feedback/mechanism like a callback to handle when an animation is canceled.

Also, my OCD hates that both `canceled` and `cancelled` are [technically correct](http://grammarist.com/spelling/cancel/), and looking at our API docs I do not see that we stuck with one over the other...

m1ga:
Added the `cancel` event, updated the example, added the JIRA ticket

jquick-axway:
@m1ga, calling `View.animate()` should stop an existing animation too. On iOS, it does. On Android, the previous animation will continue to animate until its duration has ended, which causes the 2nd given animation to be ignored . This is a pre-existing bug in Titanium and not an issue on your end, but this is good opportunity to fix it. Thanks.

m1ga:
@jquick-axway I've added `animate()` to call stop, too. Should I remove `stopAnimation()`?

jquick-axway:
> Should I remove `stopAnimation()`?

Good question... because I'm wondering if setting `View.animate()` to `null` would be a good alternative to stopping an animation. Currently, setting it to `null` will throw an exception on Android, but we can change that behavior. On iOS, a `null` animation argument is simply ignored, but doesn't stop the current animation (we can change that too).

**Side Note:**
I've been experimenting with animations this week. Your fix to the `View.animate()` method solves an edge case where a translation/scale animation needs to change when the app orientation changes. So, thanks for this. :)

m1ga:
great...my master branch is broken after testing the github issue PR :disappointed: need to fix that

PR #10131, opened 2018-06-25
[TIMOB-24776] iOS: ListView item template support CommonJS modules

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24776

**Description:**
iOS Parity for #9955 
Also very close to #9767

This is an amateurish hack for trying to fall back to loading an alloy widget by name in templates. First we try to original code path of loading a native proxy. If that throws an exception because no proxy exists by that name, I use some questionable JS to try and load a widget by name and construct one.

Please feel free to hack this to pieces, but it's a good starting point and is roughly equivalent to what Android is doing.

Also, I *hate* that this codes alloy specific logic directly into the SDK core. Ideally alloy itself would have some hook to process the template names/provide widgets as templates. Or maybe that we had a more generic mechanism to pass in the path of the module to load as the template and we require that? I don't know...


iphone/Classes/TiViewProxy.m
```diff
@@ -2935,13 +2935,44 @@ + (TiViewProxy *)unarchiveFromTemplate:(id)viewTemplate_ inContext:(id<TiEvaluat
...
-    TiViewProxy *proxy = [[self class] createProxy:viewTemplate.type withProperties:nil inContext:context];
-    [context.krollContext invokeBlockOnThread:^{
-      [context registerProxy:proxy];
-      [proxy rememberSelf];
-    }];
-    [proxy unarchiveFromTemplate:viewTemplate];
-    return proxy;
+    TiViewProxy *proxy = nil;
+    // First try a native proxy class...
+    @try {
```
hansemannn: The approach I took in https://github.com/appcelerator/titanium_mobile/pull/9767 was to use a "module" tag that does class swizzling internally to construct classes from native modules. Obviously, this is not possible for common-js modules, but we could try to require the .js file manually and receive the proxy from there, similar to [here](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiHost.m#L151). We could therefore add the Alloy-lookup into our require-scope temporary and keep it's logic hidden from the actual proxy. 

I agree that our core should **not** have any Alloy references as it becomes a huge mess across versions of both products over time, but I understand the state as a workaround. @vijaysingh-axway also spent some time on this already and also had issues with Alloy specifically and may be able to add something here as well.

sgtcoolguy: @hansemannn @vijaysingh-axway Well, this PR and the Android one are high priority due to customer request for this. It originated as an Alloy feature request and support in Alloy was added at the end of April...

iphone/Resources/app.js
```diff
@@ -5,18 +5,103 @@
  * Change the below code to fit your use-case. By default, it included a button
  * to trigger a log that is displayed in the Xcode console.
  */
+'use strict';
```
hansemannn: We should not change our internal app.js

iphone/Resources/alloy/widgets/custom/controllers/widget.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
```
hansemannn: This seems to be part of your local testing environment and should not be committed.

hansemannn (changes requested):

iphone/Classes/TiViewProxy.m
hansemannn: Can't a developer just use the widget view inside the list-view as a workaround? And are we planning to ship this in a 7.2.1, 7.3.0 or related? The Alloy PR should never been merged at this point, the SDK implementation was not in place.

hansemannn:
Fixed the review comments and crash in https://github.com/sgtcoolguy/titanium_mobile/pull/3

iphone/Classes/TiViewProxy.m
vijaysingh-axway: @sgtcoolguy @hansemannn  Isn't it good to use if-else instead of try-catch. e.g                                                                
if ([viewTemplate.type hasPrefix:@"Ti."] || [viewTemplate.type hasPrefix:@"Titanium.”]) {
//Native proxy
} else {
// Alloy widget
}

hansemannn: @vijaysingh-axway Good idea! That's what also done inside the `createProxy:` factory method. By checking it before calling it, we also prevent a debug log that warns the user that the proxy cannot be constructed. See the change in my PR to this PR.

hansemannn:
@sgtcoolguy Ping!

iphone/Resources/alloy/widgets/custom/controllers/widget.js
sgtcoolguy: This was intentionally committed so you guys could work from the same example quickly. It is intended to be reverted.

hansemannn (approved):

ludolphus:
This fails when using modules such as https://github.com/AndreaVitale/av.imageview in combination with ListView templates. Type used for this module is 'AvImageview.ImageView'. As this is a native module but not prefixed with Ti or Titanium it fails that test and only commonJS modules inside /alloy/widgets are tested after that. Anything else will fail.

joshualambert:
^ Can confirm what @ludolphus is reporting here. I'm seeing this same behavior on 7.5.0 builds. Breaks the AVImageView module for me also.

ludolphus:
@joshualambert what I did is take the 'unarchiveFromTemplate' method from 7.4.1.GA (TiViewProxy.m) and replaced it in this 7.5.0 test version. Everything works great now.

janvennemann:
cc @sgtcoolguy I've created https://jira.appcelerator.org/browse/TIMOB-26499 for this. We may want to revert this to not break apps in 7.5.0

PR #10134, opened 2018-06-27
[TIMOB-26157] Android: Avoid Android 8.0 crash when applying fixed orientation to semi-transparent or modal window

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26157

**Summary:**
- As of Android 8.0, Google no longer supports assigning a fixed orientation (ie: portrait-only or landscape-only) to a semi-transparent/translucent window. Will now catch Google's exception when attempting to do so and log it as an error.
- Removed API Level 9 checks when assigning orientation. No longer relevant.
- The variable `activityOrientationMode` used to be initialized to `-1` which happened to exactly match the value assigned to Android constant [SCREEN_ORIENTATION_UNSPECIFIED](https://developer.android.com/reference/android/content/pm/ActivityInfo.html#SCREEN_ORIENTATION_UNSPECIFIED). Simplified code to use the constant.

**Test:**
1. Build and run the below code on an Android 8.x device.
2. Verify that a semi-transparent orange window is displayed over the splash screen. (It used to crash here.)
3. Verify that you see the following message in the Android log, `[ERROR]: TiWindowProxy: (main) [57,57] Only fullscreen activities can request orientation`.
4. Note that the window orientation is not portrait-only and allows you to rotate the app to landscape. This is because fixed-orientation translucent windows are no longer supported.

```javascript
	opacity: 0.5,
//	modal: true,
	orientationModes: [Ti.UI.PORTRAIT],
	backgroundColor: "orange",
});
window.add(Ti.UI.createLabel({ text: "Translucent Window Test" }));
window.open();
```


garymathews (approved):
CR: **PASS**

PR #10140, opened 2018-06-28
[TIMOB-26164] : iOS TableView + run-on-main-thread - insertRowAfter directly after updateRow causes error / race crash

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26164


iphone/Classes/TiUITableViewProxy.m
```diff
@@ -421,7 +421,7 @@ - (void)updateRow:(id)args
     TiUITableViewAction *action = [[[TiUITableViewAction alloc] initWithObject:newrow animation:anim type:TiUITableViewActionUpdateRow] autorelease];
     [table dispatchAction:action];
   },
-      NO);
+      YES);
```
hansemannn: We have some more cases where `dispatchAction` is called directly. Should these be put on a sync-queue as well or are we fine?

vijaysingh-axway: Ideally it should be. Added. As we discussed we can create a ticket to validate TiThreadPerformOnMainThread in whole sdk.

PR #10143, opened 2018-06-29
Fix warnings in doc generation

ewanharris:
Fix the warnings output when generating the jsduck docs

- Cannot find API: String
	- Replace Array<String> with `Array<String>`, I think this should look ok on the generated docs site?
- Cannot find API: Titanium.Media.MusicPlayer.volume
	- Remove the volumechange event as the volume property was removed,
- Cannot find API: Titanium.Media.VideoPlayer.naturalSizeAvailable
	- Event should be all lowercase
- Cannot find API: Titanium.Media.VideoPlayer.sourceChange
	- Event should be all lowercase
- Cannot find class: Ti.UI.2DMatrix.MatrixCreationDict
	- Correct reference is MatrixCreationDict
- Cannot find API: Titanium.UI.iPhone.showStatusBar
- Cannot find API: Titanium.UI.iPhone.hideStatusBar
- Cannot find API: Titanium.UI.iPhone.statusBarStyle
	- Variety of fixes


apidoc/Titanium/Media/MusicPlayer.yml
```diff
@@ -92,13 +92,6 @@ events:
         playback state.
     since: '3.0.0'
 
-  - name: volumechange
-    summary: Fired when the volume changes.
-    description: |
-        Use the [volume](Titanium.Media.MusicPlayer.volume) property to set or determine
-        the current volume level for this player.
-    since: '3.0.0'
-
```
hansemannn: Sorry, I was initially wrong here: The `volumechange` event is still fired once the volume changes, different to the `volume` property that is the read-only version to receive the volume at any time.

hansemannn (changes requested):
One small change required.

apidoc/Titanium/Media/MusicPlayer.yml
ewanharris: Ah cool, looking at the code this is specific to the MusicPlayer right? I'll just remove the description for this entry, or should we note that it's not possible to retrieve the volume value?

hansemannn: But it is possible to receive the `volume` using the getter on a music-player instance.

hansemannn (approved):

PR #10150, opened 2018-07-03
[TIMOB-26179] Check for property existence using hasProperty

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26179

**Optional Description:**
Implements the `hasProperty` callback to properly check for property existence in our proxy objects. Without this callback `getProperty` is which results in property checks always returning `true`.

@hansemannn Not sure if this implies a breaking change or can just be considered as a re-definition of previously undefined behavior.


hansemannn:
LGTM. But please add a unit-test for it since it should be somehow testable from our proxy implementation. Also, do you have a practical use case for this that is worth mentioning?

sgtcoolguy:
I think it'd be pretty easy to add a quick unit test that generated a proxy and then just tested for a non-existent and existing own property. I believe iOS always returned true before?
```js
should(label.hasOwnProperty('madeup')).eql(false);
should(label.hasOwnProperty('text')).eql(true);
label.madeup = 123;
should(label.hasOwnProperty('madeup')).eql(true);
```

PR #10151, opened 2018-07-03
Updated Window.yml

jawa9000:
Replaced "Windows Phone 8.1" with "Windows Phone" per https://jira.appcelerator.org/browse/TIDOC-3206


infosia (approved):
👍

PR #10152, opened 2018-07-03
[TIMOB-26181] Android: Do not init kroll object if krollruntime is disposed

garymathews:
- Do not initialise `KrollObject` if `KrollRuntime` is disposed

##### TEST CASE
```JS
win.open();
```
- Press Back, should not see crash log

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26181)


jquick-axway (approved):
CR: Pass

PR #10153, opened 2018-07-03
[TIMOB-26182] Android: Validate external storage temp directory

garymathews:
- Validate we have access to external storage before using it for a temp directory

##### TEST CASE
- On a device with external storage, without external storage permissions
```JS
const win = Ti.UI.createWindow({ backgroundColor: 'gray' }),
      img = Ti.UI.createImageView({ image: 'https://apod.nasa.gov/apod/image/1807/PlaneMoon_Kim_960.jpg' });

win.add(img);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26182)


android/runtime/common/src/java/org/appcelerator/kroll/util/TiTempFileHelper.java
```diff
@@ -167,7 +170,13 @@ private void updateTempDir()
 	{
 		String extState = Environment.getExternalStorageState();
 		if (!extState.equals(previousExternalStorageState)) {
-			if (Environment.MEDIA_MOUNTED.equals(extState)) {
+			if (Environment.MEDIA_MOUNTED.equals(extState)
+				&& (Build.VERSION.SDK_INT < 23
+					|| (Build.VERSION.SDK_INT >= 23
```
ypbnv: Maybe we can get away with removing `Build.VERSION.SDK_INT >= 23` check ? It should be evaluated only if the first check `Build.VERSION.SDK_INT < 23` is `false`, which will leave us with it always being `true`.

garymathews: What if `Environment.MEDIA_MOUNTED.equals(extState)` is `false`?

sgtcoolguy: If `Environment.MEDIA_MOUNTED.equals(extState) === false`, then yes the rest of the `if` won't get evaluated. I believe Yordan is correct and that you could remove `Build.VERSION.SDK_INT >= 23` since it is implied (being the exact opposite of `Build.VERSION.SDK_INT < 23`)

lokeshchdhry:
@garymathews , the build is failing linting. Can you please take a look.

lokeshchdhry:
FR Passed.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2

PR #10155, opened 2018-07-05
[TIMOB-26184] iOS: Allow modification of global timer functions

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26184

**Optional Description:**
Allow modifying some of our global function like `setTimeout` or `setInterval`. This is possible in other JS environments too and required for using Jasmin for example.

Note that this is already possible on Android.

### DIFF
--- a/iphone/Classes/KrollContext.m
+++ b/iphone/Classes/KrollContext.m
@@ -1098,7 +1098,7 @@ static TiValueRef StringFormatDecimalCallback(TiContextRef jsContext, TiObjectRe
     TiObjectRef global = TiContextGetGlobalObject(context);
     TiObjectSetProperty(context, global,
         invokerFnName, invoker,
-        kTiPropertyAttributeReadOnly | kTiPropertyAttributeDontDelete,
+        kTiPropertyAttributeNone,
         NULL);
   }
   TiStringRelease(invokerFnName);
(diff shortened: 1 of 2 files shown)


hansemannn:
Code change LGTM. Please add a unit test for modifying the methods as part of our mocha-test-suite. Thanks!

sgtcoolguy:
Just confirmed that on Node the setTimeout function is writable/enumerable/configurable, which is what this change should do... 👍 

You *could* write a test that overrode the impl and then tried to set it back, but I worry that might interfere with mocha actually properly running the tests. It's probably sufficient to call Object.getOwnPropertyDescriptor(global, 'setTimeout') and verify that it's enuemrable/writable/configurable.

ewanharris:
Just to double check my reading, this will make the 4 timeout/interval based functions, alert, L and require overwrite-able. Will that the same across all platforms? (not saying we shouldn't, just making sure it's the expected)

janvennemann:
@ewanharris Yup, that is exactly the intention of this PR. On Android this is already the case, although i'm not familiar enough with the V8 sources to say where it is configured there.

I'll write the appropriate unit tests after doc day to see if it is working as intended on both Android and iOS.

PR #10156, opened 2018-07-06
[TIMOB-26115] TiAPI: Add support for console.timeLog

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26115

Implement console.timeLog for Android and iOS

Test:

```js
console.time('mytimer'); // Start timer
console.timeLog('mytimer'); // Log time taken so far
console.timeLog('mytimer', 'with', 'some', 'extra', 'info'); // Log time taken with extra logging
console.timeLog('mytimer', [ 'a', 'b', 'c' ], { objects: true }); // Should handle Arrays and Objects
console.timeEnd('mytimer');
```
Should log something like the below, there's some slight discrepancies between how Arrays/Objects are logged on platforms.
```
[INFO]  mytimer: 1ms
[INFO]  mytimer: 2ms with some extra info
[INFO]  mytimer: 3ms ["a","b","c"] {"objects":true}
[INFO]  mytimer: 3ms
```

### DIFF
--- a/android/runtime/common/src/js/console.js
+++ b/android/runtime/common/src/js/console.js
@@ -26,6 +26,22 @@ function join(args) {
 	}).join(' ');
 }
 
+function logTime(label, ...logData) {
+	label = `${label}`;
+	const startTime = times.get(label);
+	if (!startTime) {
+		exports.warn(`Label "${label}" does not exist`);
+		return true;
+	}
+	const duration = Date.now() - startTime;
+	if (logData) {
+		exports.log(`${label}: ${duration}ms`, logData);
+	} else {
+		exports.log(`${label}: ${duration}ms`);
+	}
+	return false;
+}
+
 exports.log = function () {
 	Titanium.API.info(join(arguments));
 };
@@ -56,13 +72,12 @@ exports.time = function (label = 'default') {
 };
 
 exports.timeEnd = function (label = 'default') {
-	label = `${label}`;
-	const startTime = times.get(label);
-	if (!startTime) {
-		exports.warn(`Label "${label}" does not exist`);
-		return;
+	const warned = logTime(label);
+	if (!warned) {
+		times.delete(label);
 	}
-	const duration = Date.now() - startTime;
-	exports.log(`${label}: ${duration}ms`);
-	times.delete(label);
+};
+
+exports.timeLog = function (label = 'default', ...logData) {
+	logTime(label, logData);
 };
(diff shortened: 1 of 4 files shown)


iphone/Classes/TiConsole.m
```diff
@@ -50,6 +50,27 @@ - (void)timeEnd:(id)label
   [_times removeObjectForKey:label];
 }
 
+- (void)timeLog:(id)args
+{
+  NSString *label = [args objectAtIndex:0];
```
hansemannn:
```objc
NSString *label = [args objectAtIndex:0] ?: @"default";
```

iphone/Classes/TiConsole.m
```diff
@@ -50,6 +50,27 @@ - (void)timeEnd:(id)label
...
+  if (label == nil) {
+    label = @"default";
+  }
+  NSDate *startTime = _times[label];
+  if (startTime == nil) {
+    NSString *logMessage = [NSString stringWithFormat:@"Label \"%@\" does not exist", label];
+    [self logMessage:[logMessage componentsSeparatedByString:@" "] severity:@"warn"];
+    return;
+  }
+  double duration = [startTime timeIntervalSinceNow] * -1000;
```
hansemannn: Use `NSTimeInterval` instead of `double`. They are the same, but `NSTimeInterval` is more known for storing time values.

iphone/Classes/TiConsole.m
```diff
@@ -50,6 +50,27 @@ - (void)timeEnd:(id)label
...
+  NSString *logMessage = [NSString stringWithFormat:@"%@: %0.fms ", label, duration];
+  if ([logData count]) {
```
hansemannn: Use `[logData count] > 0` for clearance.

hansemannn (approved):
iOS **approved**

android/runtime/common/src/js/console.js
```diff
@@ -56,13 +72,12 @@ exports.time = function (label = 'default') {
 };
 
 exports.timeEnd = function (label = 'default') {
-	label = `${label}`;
-	const startTime = times.get(label);
-	if (!startTime) {
-		exports.warn(`Label "${label}" does not exist`);
-		return;
+	const warned = logTime(label);
```
sgtcoolguy: the use of `warned` here is a little confusing. What you're really determining is if the label existed.

sgtcoolguy (approved):
Looks good. The ti.shell/common JS bootstrap stuff that Josh Quick is doing could lead us to define this just once in JS code across all platforms and thereby allow the fancier unit tests I added to work on iOS too. But for now, this works.

PR #10159, opened 2018-07-10
Update SearchBar and SearchView apidoc

infosia:
Update apidoc for `SearchBar` and `SearchView` for the Day of Docs July 2018.

- Changed some defaults, it turns out `<>` and `"` don't work at `default` section.
- Added more descriptions for Windows
- Added more examples

@jawa9000 Could you upload screenshot image of `SearchBar` [searchbar_windows.png](https://user-images.githubusercontent.com/1661068/42489732-b0d96276-8447-11e8-9525-e6043955cece.PNG) onto `images/searchbar/searchbar_windows.png`? It is used in `SearchBar.yml`.

```html
<td><img src="images/searchbar/searchbar_windows.png" height="25" /></td> 
```

searchbarwindows

Additional note: I'm wondering if we could get rid of `table` HTML element from `SearchBar.yml`. I thought the table HTML tag such as `<table>`, `<td>` and `<th>` are obsolete 😮

### DIFF
--- a/apidoc/Titanium/UI/ScrollView.yml
+++ b/apidoc/Titanium/UI/ScrollView.yml
@@ -27,7 +27,7 @@ description: |
 
     Pinching a `ScrollableView` to zoom in and out of content is not supported on Android. On iOS, this action
     is natively supported by the UIScrollView class, but on Android, the native ScrollView class does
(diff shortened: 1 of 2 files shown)


jawa9000 (approved):
Looks good.

PR #10161, opened 2018-07-10
[TIMOB-17546] ScrollableView paging controls same as in iOS

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17546

https://github.com/appcelerator/titanium_mobile/pull/10161/commits/f5387cd8a2a05acb298fc6e4ddab1d57c5cd7e72
- move all properties to `ScrollableViewProxy`
- instantiate `TiUIScrollableView` only when it needed
- update `ViewPagerAdapter` (getItemPosition returns correct value, previously always `POSITION_NONE`)

https://github.com/appcelerator/titanium_mobile/pull/10161/commits/547d36b4b8d4edc2ef402d79d5dbcc39dcf1a603:
- create paging control only when needed

https://github.com/appcelerator/titanium_mobile/pull/10161/commits/95c18fc2202bdabdc0190983ae0eefc4fd477e4c:
- For backward compatibility TiDotPagingControl used only if `useLegacyControl` property of ScrollableView is set to `false` (default is `true`).
- TiDotPagingControl acts as default iOS control with `overlayEnabled: true`
- add properties: `currentPageIndicatorColor`, `pageIndicatorColor`, `pagingControlHeight`, `pagingControlOnTop`

UPDATE: also https://github.com/appcelerator/titanium_mobile/pull/10161/commits/f5387cd8a2a05acb298fc6e4ddab1d57c5cd7e72 fixes existing tests which were marked as broken for android


drauggres:
Here is my thoughts about this PR:
- changing paging control for ScrollableView would be breaking change, so `TiDotPagingControl` is not enabled by default
- `useLegacyControl` will be set to `true` on 7.4.0, changed to `false` on 8.0.0 and some time later deprecated and removed along with legacy paging control

If we are agreed on that, I'll update docs and tests appropriately. Or we could postpone this PR to 8.0.0.

sgtcoolguy:
ping @garymathews @jquick-axway Please take a moment to review this community PR

lokeshchdhry:
I am seeing few differences on Android & IOS:
1. On IOS, the `paging controls` are always visible where as on Android they disappear after few seconds.
2. On IOS, Clicking on the `paging control dot` scrolls the `scrollableView` to either right or left depending on which dot is clicked, whereas on Android clicking the dot does nothing.
3. Also, on Android if the paging dots are not visible they will only become visible after you scroll a full view left or right. I think best thing to do here if possible will be to bring up or make visible the `paging control dots` as soon as a user touches the screen. 

Not sure if these differences are due to platform restrictions though.

drauggres:
1 and 3 are happening because [`pagingControlTimeout`](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.ScrollableView-property-pagingControlTimeout) has default value `3000`. I can reset it to zero when `useLegacyControl : false` and `pagingControlTimeout` is not set explicitly.

About 2: I wasn't aware of this. But if click only performs left/right scrolling (and not reset current page corresponding to dot) I think this feature is useless.

drauggres:
Closing in favor of #10316

PR #10162, opened 2018-07-10
[DOC] Minor doc update on Ti.UI.AlertDialog

ssjsamir:
https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.AlertDialog

+ Minor text update 
+ Updated Alloy test case so the application window opens


jawa9000 (approved):
Looks good.

PR #10165, opened 2018-07-10
Remove invalid character in media docs

ewanharris:
fyi, @jawa9000 this seems to be the cause of all the build failures, not sure why. Had to PR as I can't push to master


ewanharris:
For those as boring as me, that's charcode 924 a.j.a Greek Capital Letter MU

PR #10166, opened 2018-07-10
[TIMOB-26196] iOS: Remove SCListener, use AVAudioRecorder instead

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26196

> `no tests` since Jenkins can probably not speak to the Simulator. We could add type-tests for the properties, but that didn't really change.


iphone/Classes/MediaModule.m
```diff
@@ -670,28 +669,71 @@ - (void)requestAudioRecorderPermissions:(id)args
 
 - (void)startMicrophoneMonitor:(id)args
 {
-  [[SCListener sharedListener] listen];
+  if (_microphoneRecorder != nil) {
+    DebugLog(@"[ERROR] Trying to start new microphone monitoring while the old one is still active. Call \"stopMicrophoneMonitor()\" before and try again.");
+    return;
+  }
+
+  NSURL *url = [NSURL fileURLWithPath:@"/dev/null"];
```
vijaysingh-axway: Is value @"/dev/null" is correct for path or we should get path using [NSFileMange defaultManage] api to get path?

iphone/Classes/MediaModule.m
```diff
@@ -670,28 +669,71 @@ - (void)requestAudioRecorderPermissions:(id)args
...
+  NSDictionary *settings = [NSDictionary dictionaryWithObjectsAndKeys:
+                                             [NSNumber numberWithFloat:44100.0], AVSampleRateKey,
+                                         [NSNumber numberWithInt:kAudioFormatAppleLossless], AVFormatIDKey,
+                                         [NSNumber numberWithInt:1], AVNumberOfChannelsKey,
+                                         [NSNumber numberWithInt:AVAudioQualityMax], AVEncoderAudioQualityKey,
+                                         nil];
+
```
vijaysingh-axway: Are these default value of setting ? Is there any doc for recommending these settings?

iphone/Classes/MediaModule.m
```diff
@@ -670,28 +669,71 @@ - (void)requestAudioRecorderPermissions:(id)args
...
+  NSError *error;
+
+  _microphoneRecorder = [[AVAudioRecorder alloc] initWithURL:url settings:settings error:&error];
+
```
vijaysingh-axway: Should check for permission 'NSMicrophoneUsageDescription'

vijaysingh-axway (changes requested):
Can you please address the comments. Thanks

iphone/Classes/MediaModule.m
hansemannn: They are recommended for backwards compatibility with the old API. Especially the sample rate (44100.0) is for that. The other ones ensure a certain quality of the recording.

iphone/Classes/MediaModule.m
hansemannn: We can do that, yes. But that was not part of the old API as well, so it would only be a log-warning.

vijaysingh-axway (approved):
CR passed.

PR #10169, opened 2018-07-11
[TIMOB-26095] iOS 12: Replace UIWebView API in favor of WKWebView

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26095


apidoc/Titanium/UI/WebViewConfiguration.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    E.g.
+
+    var config = Ti.UI.createWebViewConfiguration({
+        allowsPictureInPictureMediaPlaback: true
+    });
+
+    var webView = WK.createWebView({
+        configuration: config
+    });
```
hansemannn: The `url` property is always required. Did you test this example code? We should always test code before putting it into the docs, since some docs expect copy and paste examples to just work.

apidoc/Titanium/UI/WebViewConfiguration.yml
```diff
@@ -0,0 +1,107 @@
...
+
+extends: Titanium.Proxy
+since: "7.4.0"
+platforms: [iphone, ipad]
+
+properties:
+
+  - name: preferences
+    summary: The preference settings to be used by the web view.
+    type: PreferencesObject
```
hansemannn: We always need to prefix these kind of API's, e.g. `WebViewPreferencesObject`.

apidoc/Titanium/UI/WebViewConfiguration.yml
```diff
@@ -0,0 +1,107 @@
...
+    platforms: [iphone, ipad]
+    default: true
+
+  - name: processPool
+    summary: The process pool from which to obtain the  Web Content process of view.
+    description: |
+        When a web view is initialized, either a new web content process is created for it from the 
+        specified pool or an existing process in that pool is used.
+    type: Titanium.UI.WebViewProcessPool
+    platforms: [iphone, ipad]
```
hansemannn: All of these are missing the minimum iOS version.

apidoc/Titanium/UI/WebViewProcessPool.yml
```diff
@@ -0,0 +1,33 @@
+
+---
+name: Titanium.UI.WebViewProcessPool
```
hansemannn: Remove empty line in L1.

apidoc/Titanium/UI/WebViewProcessPool.yml
```diff
@@ -0,0 +1,33 @@
...
+        processPool: pool
+    });
+
+    var firstWebView = Ti.UI.createWebView({
+        configuration: config1
+    });
+
+    var secondWebView = Ti.UI.createWebView({
+        configuration: config2
+    });
```
hansemannn: Please provide full example, with URL's and opening windows.

iphone/Classes/TiUIWebView.h
```diff
@@ -6,51 +6,34 @@
  */
 #ifdef USE_TI_UIWEBVIEW
 
+#import "TiDimension.h"
```
hansemannn: General question here: Did you copy the changes into this file (replaced) or did you copy the Ti.WKWebView  classes over and renamed them? I'd highly prefer the latter option to not mix up API's.

iphone/Classes/TiUIWebView.m
```diff
@@ -7,66 +7,19 @@
 #ifdef USE_TI_UIWEBVIEW
```
hansemannn: Not explicitly reviewing this file, since it should be a 1:1 copy of the existing file from Ti.WKWebView + the macros. If not, please change.

iphone/Classes/TiUIWebViewConfigurationProxy.h
```diff
@@ -0,0 +1,34 @@
...
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2009-2017 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+#ifdef USE_TI_UIWEBVIEW
+
+#import "TiProxy.h"
+
+@interface TiUIWebViewConfigurationProxy : TiProxy {
```
hansemannn: Noting one more time: Move to iOS-only `TiUIiOSWebViewConfigurationProxy` namespace.

tests/Resources/ti.ui.webview.addontest.js
```diff
@@ -0,0 +1,48 @@
...
+			url: 'https://google.com'
+		});
+
+		webView.startListeningToProperties([ 'title' ]);
+		webView.addEventListener('title', function (e) {
+			finish();
+		});
+		win.add(webView);
+		win.open();
+	});
```
hansemannn: We need unit tests for every new property and method. It will for sure be some work but worth it for a stable API and future testing.

hansemannn (changes requested):
Left my first couple of review comments after round 1 of CR. Let me know once it's ready for the next round, but **great work** so far! This API is so huge.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -368,6 +463,69 @@ events:
...
+        summary: |
+            A Boolean value indicating whether the frame is the web site's main frame or a subframe.
+        type: Boolean
+
+  - name: progress
+    summary: Fired when webpage download progresses.
+    description: |
+        This value ranges from 0.0 to 1.0 based on the total number of bytes expected to be received, 
+        including the main document and all of its potential subresources.
+    platforms: [ipad, iphone]
```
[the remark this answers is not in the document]
jquick-axway: Yes. On Android, we can detect page download progress via [WebChromeClient.onProgressChanged()](https://developer.android.com/reference/android/webkit/WebChromeClient.html#onProgressChanged(android.webkit.WebView,%20int)).

Would you also mind rephrasing the description please? How about to something like this?
> Provides a normalized value between 0.0 and 1.0 where 0.0 indicates nothing was downloaded and 1.0 means the page and its resources has finished downloading.

**Question:**
Does this value include iframes? I suspect it does not since an iframe can timeout and fail to download, but the main page can finish fine. I tried looking for info on the Android side about iframes, but it doesn't say anything either.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -437,6 +595,13 @@ properties:
     availability: creation
     default: true
 
+  - name: disableZoom
```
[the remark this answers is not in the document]
jquick-axway: Yes, Android's zoom controls (such as pinch-zoom and its hover +/- buttons) can be disabled.

I prefer the `enableZoomControls` property name as well. Most frameworks typically use the term "enable" instead of "disable" when naming properties/methods.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -244,20 +244,114 @@ methods:
...
+
+  - name: takeSnapshot
+    summary: Takes a snapshot of the view's visible viewport.      
+    platforms: [iphone, ipad]
+    since: "7.4.0"
+    osver: {ios: {min: "11.0"}}
+    parameters:
+      - name: callback
+        summary: Function to call upon snapshot captured.
+        type: Callback<SnapshotResult>
```
jquick-axway: Does `takeSnapshot()` take a screenshot of the WebView's contents?

We already have a `toImage()` method for taking a screenshot, but it would include the WebView's border too which might not be desired. But we may want to double check if `toImage()` is still able to capture the WebView's contents or not. Would you mind double checking please?

And perhaps we should consider renaming `takeSnapshot()` to something else since `toImage()` already exists to avoid confusion between the two. Like `contentToImage()`? Or `pageToImage()`? Any other suggestions?

apidoc/Titanium/UI/WebView.yml
```diff
@@ -244,20 +244,114 @@ methods:
...
+      - name: mainFrameOnly
+        summary: |
+            A Boolean value indicating whether the script should be injected only into the 
+            main frame (true) or into all frames (false).
+        type: Boolean
+
+  - name: removeAllUserScripts
+    summary: Removes all associated user scripts.      
+    platforms: [iphone, ipad]
+    since: "7.4.0"
```
jquick-axway: How will `addUserScript()` and `removeAllUserScripts()` impact our existing `evalJS()` function and our support for Titanium JS APIs within a locally loaded HTML file? Both of those things involve script injection as well. Also, will the `addUserScript()` and `removeAllUserScripts()` method only work **after** the "load" event has been received? If so, then we should document this... or have the given scripts queued to be added once loaded.

One more thing. Please run the test in PR #9931 . This loads local HTML in string form and verifies that it has access to our Titanium APIs... and it supports 2 way interop between the WebView and the Titanium "app.js".

apidoc/Titanium/UI/WebView.yml
```diff
@@ -368,6 +463,69 @@ events:
...
+    platforms: [ipad, iphone]
+    since: "7.4.0"
+    properties:
+      - name: handler
+        summary: Page title of webpage.
+        type: String
+      - name: url
+        summary: URL of the web document being loaded.
+        type: String
+
```
jquick-axway: I think the `handleurl` summary is wrong. This even is fired for every URL and before the WebView does the HTTP request, right?

Also, doesn't `handleurl` require immediate feedback? So, if "tiapp.xml" property `run-on-main-thread` is set to `false`, then there is no way to respond to this in-time. We'll run into the same issue as our `onLink` feature in PR #9459.

Feedback?

apidoc/Titanium/UI/WebView.yml
```diff
@@ -368,6 +463,69 @@ events:
...
+    summary: Fired when a web view receives a server redirect.
+    platforms: [ipad, iphone]
+    since: "7.4.0"
+    properties:
+      - name: title
+        summary: Page title of webpage.
+        type: String
+      - name: url
+        summary: URL of the web document being loaded.
+        type: String
```
jquick-axway: Does "redirect" need to be its own event? Can we consolidate it with the "handleurl" event instead and provide a property indicating what triggered it?

apidoc/Titanium/UI/WebView.yml
hansemannn: It captures the actual visible area of the web-view (see [the Apple docs here](https://developer.apple.com/documentation/webkit/wkwebview/2873260-takesnapshotwithconfiguration)). @vijaysingh-axway Please check if we can have a < iOS 11 version as well and do the snapshot internally by overriding "toImage".

apidoc/Titanium/UI/WebView.yml
hansemannn: "handleurl" is only for handling url-schemes, e.g. `fb://`. How does Android signalize redirects? Via beforeload?

vijaysingh-axway:
I have tried to fix all issues. Thanks @hansemannn  and @jquick-axway for your detailed feedback. @jquick-axway  is right with 'handleurl' event, if it is not on main-thread. Currently it is crashing. @hansemannn Lets discuss on this. I think we can implement similar to 'onlink' event . Rather invoking decision handler, we should return  Ti.UI.iOS.ACTION_POLICY_* .

apidoc/Titanium/UI/WebViewConfiguration.yml
vijaysingh-axway: @hansemannn Should we really mention iOS version, which is iOS 8.0? As we are going to support these with SDK 8.0.0+ and min iOS version for that will be 9.0.

apidoc/Titanium/UI/WebView.yml
vijaysingh-axway: takeSnapshot supported for iOS < 11 using 'toImage'.

hansemannn:
@vijaysingh-axway Can you fix the thread issue or should I go for it via ti.wkwebview first? For the "onlink" property, we can unwrap the boolean from our kroll-core like I did for the current implementation and return the WKWebView policy based on true or false :-).

vijaysingh-axway:
Which thread issue? Do you mean  'handleurl',  which is crashing on main thread?

hansemannn:
I thought it's only crashing when not on main thread?

apidoc/Titanium/UI/WebView.yml
jquick-axway: On Android 6.x and older OS versions doesn't distinguish. All URL requests are passed to the Java [WebViewClient.shouldOverrideUrlLoading()](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)) method. This includes redirects and custom URL schemes. The only way to know if it's a redirect is for us track the URL usage ourselves, which we can do. Unfortunately, this method never gets called when loading URLs via the web page's JavaScript code, which is a bummer. We can only detect JavaScript loaded URLs after the HTTP request has been sent via the [WebViewClient.onPageStarted()](https://developer.android.com/reference/android/webkit/WebViewClient.html#onPageStarted(android.webkit.WebView,%20java.lang.String,%20android.graphics.Bitmap)) method. At least that's how I remember it back in the Android 2.x/3.x days.

Android 7.0 has a new [WebResourceRequest](https://developer.android.com/reference/android/webkit/WebResourceRequest) to determine if it's a redirect, but if we track it ourselves for Android 6.x and below, then there is no point in using it.

jquick-axway:
@vijaysingh-axway,
The `onlink` PR #10180 that @hansemannn wrote up is a `handleurl` equivalent and is thread safe, right? We're doing `onlink` on Android as well. Perhaps you should do the same thing Hans is doing with your `WKWebView`?

jquick-axway:
@vijaysingh-axway, can you check if it's possible to do mixed https/http content in a `WKWebView` please? For example, an "https" web page with an http iframe or http image. I'm asking because this was a recent feature request here: [TIMOB-19618](https://jira.appcelerator.org/browse/TIMOB-19618)

Based on what I've read on stackoverflow, `WKWebView` does not support this, but it doesn't hurt to double check. Such as if there are any plist settings to enable it. Thanks.
(And if it's not possible, then so be it.)

apidoc/Titanium/UI/WebView.yml
hansemannn: the handler is a completion handler that can be called async, passing a constant to allow or disallow the request. there is an example in the ti.wlwebview docs

hansemannn:
@vijaysingh-axway I changed the base branch to `next` as we are coming closer to approving this. Can you please resolve the merge conflicts? Especially be very careful with the .pbxproj changes and let me know if I can help.

hansemannn:
Thanks! Is the failing `ios.Titanium.UI.WebView` test related to this?

vijaysingh-axway:
It was failing previously also with this PR. I have to find out reason.

android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
```diff
@@ -612,12 +604,7 @@ public void startService(IntentProxy intentProxy)
 	@Kroll.method
 	public void stopService(IntentProxy intentProxy)
 	{
-		TiApplication app = TiApplication.getInstance();
-		if (app != null) {
-			app.stopService(intentProxy.getIntent());
-		} else {
-			Log.w(TAG, "Application instance no longer available. Unable to stopService.");
-		}
+		TiApplication.getInstance().stopService(intentProxy.getIntent());
```
hansemannn: This seem to be incorrectly solved merged conflicts. Please verify this for each file.

vijaysingh-axway: This file was not in list of conflict files which I resolved. There were 3 iOS files only. Thanks!

jquick-axway: This part was merged correctly. I changed the code to this last week.
But thanks for double checking.

_**Side Note:** The `TiApplication` instance is guaranteed to exist for the lifetime of the app, which means that there's no need to do a `null` check. So, I simplified the code a bit._

vijaysingh-axway:
@janvennemann, Ping for review.

iphone/Classes/TiUIiOSProxy.m
```diff
@@ -795,5 +802,42 @@ - (id)createFeedbackGenerator:(id)args
...
+{
+  return [[[TiUIiOSWebViewProcessPoolProxy alloc] _initWithPageContext:[self executionContext] args:args] autorelease];
+}
+
+MAKE_SYSTEM_PROP(CREDENTIAL_PERSISTENCE_NONE, NSURLCredentialPersistenceNone);
+MAKE_SYSTEM_PROP(CREDENTIAL_PERSISTENCE_FOR_SESSION, NSURLCredentialPersistenceForSession);
+MAKE_SYSTEM_PROP(CREDENTIAL_PERSISTENCE_PERMANENT, NSURLCredentialPersistencePermanent);
+MAKE_SYSTEM_PROP(CREDENTIAL_PERSISTENCE_SYNCHRONIZABLE, NSURLCredentialPersistenceSynchronizable);
+
+MAKE_SYSTEM_UINTEGER(AUDIOVISUAL_MEDIA_TYPE_NONE, WKAudiovisualMediaTypeNone);
```
janvennemann: Is this new macro necessary? For example, the above constants for credential persistence also are `NSUInteger` enum values and they use `MAKE_SYSTEM_PROP `

janvennemann (changes requested):
Added a short note. Can you also unit tests for the new WKWebView specific stuff like web view configuration, process pools and decision handlers?

iphone/Classes/TiUIiOSProxy.m
vijaysingh-axway: This is of type NSUInteger.
MAKE_SYSTEM_PROP(AUDIOVISUAL_MEDIA_TYPE_ALL, WKAudiovisualMediaTypeAll); will give following error message-
"Implicit conversion from 'NSUInteger' (aka 'unsigned long') to 'int' changes value from 18446744073709551615 to -1"
because WKAudiovisualMediaTypeAll = NSUIntegerMax

janvennemann: Ah ok, can you rename it to `MAKE_SYSTEM_PROP_UINTEGER` to align with `MAKE_SYSTEM_PROP `?

vijaysingh-axway: Agreed. Updated.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiBase.h
```diff
@@ -330,7 +330,7 @@ void TiExceptionThrowWithNameAndReason(NSString *exceptionName, NSString *reason
     return [NSNumber numberWithInt:map]; \
   }
 
-#define MAKE_SYSTEM_UINTEGER(name, map)              \
+#define MAKE_SYSTEM_PROP_UINTEGER(name, map)         \
```
hansemannn: Please do not rename but add a new one and deprecate the old one. Otherwise, a good hand full of native modules will be broken in SDK 8+.

janvennemann (approved):

lokeshchdhry:
FR Passed.

OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.9-3
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
Simulator: Iphone XS Max IOS 12.1

PR #10113, opened 2018-06-18
Update docs

hansemannn:
Happy pre-docs-day! This PR:
- updates the Ti.UI.TextField and Ti.Geolocation docs
- removes docs that have been deprecated/removed in 3.x
- removes pre iOS 8 (and a few pre Android 4.1) references
- adds parity for the `heading` event to include the `success` flag on Android as well

PR #10170, opened 2018-07-11
[DOC] Review Android.yml

ypbnv:
Rewording in a few places ( it sounded a bit better in my head, not sure if it is the proper way ).
Indentation here and there.
Filling in some dots and missing fields.
Move a detailed description to the proper constant.

I noticed a few properties were removed due to being deprecated/removed a long time ago. Are we safe to remove them in the code as well ?
Here they are:
https://github.com/appcelerator/titanium_mobile/pull/10113/files#diff-de28ec13bdb41ca9ccff28d09119a758L787


sgtcoolguy:
@jawa9000 ping

sgtcoolguy:
cc @garymathews RE: properties removed in docs long ago, but not code(?)

hansemannn:
@jawa9000 @garymathews Can you take a look please? This is a left-over from docs-day.

jawa9000 (changes requested):
For https://github.com/appcelerator/titanium_mobile/pull/10113/files#diff-ef87022d8eb2735f703832e1077f7f58, I would change "Before, a contact can be selected without requiring User permission." to "Previously, a contact can be selected without requiring User permission." Otherwise, it looks good.

ypbnv:
@jawa9000 Updated.

PR #10171, opened 2018-07-11
[DOC] Review of Event.yml

ypbnv:
We now do support recurring events. I am not sure if a more detailed explanation is needed here or simply a reference to the RecurringRule API doc.

### DIFF
--- a/apidoc/Titanium/Android/Calendar/Event.yml
+++ b/apidoc/Titanium/Android/Calendar/Event.yml
@@ -3,8 +3,7 @@ name: Titanium.Android.Calendar.Event
 summary: An object that represents a single event in an Android calendar.
 description: |
     The API supports retrieving information about existing events and creating new events.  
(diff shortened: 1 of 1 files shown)


hansemannn:
@ypbnv Something is odd here: Do we still maintain an "Event" API on "Ti.Android.Calendar"? Why not use https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Calendar.Event which seems to be cross-platform. Also, there it misses the recurrence rule change.

ypbnv:
@hansemannn My bad. Went through the wrong document. I will close this and open a PR with the proper YML file.

hansemannn:
No worries! Can you please also create a ticket to remove the (since 3.2.0 deprecated) API in 8.0.0? And we should do a recursive search for other API's like this (3.x, 4.x, 5.x deprecated) and remove those as well.

ypbnv:
Yes, I created a ticket. If there are other APIs that need to be removed maybe we can put them together in a story?

PR #10172, opened 2018-07-11
[TIMOB-25635] Android: Multi-View Viewpager (ScrollableView)

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25635

Android: add `clipViews` and `padding` properties to ScrollableView.

```js
  title: 'ScrollableView clipViews',
  backgroundColor:'#eee'
});

// Common params
var params = {
  clipViews: false,
  showPagingControl: false,
  top: 0
};

if (Ti.Platform.name === 'android') {
  // Android specific params
  params.padding = {
    left: 40,
    right: 40
  };
} else {
  // iOS specific params
  params.width = '90%';
}
var scrollableView = Titanium.UI.createScrollableView(params);

var view1 = Ti.UI.createView({ id: 'view1', backgroundColor: '#836' });
var view2 = Ti.UI.createView({ id: 'view2', backgroundColor: '#246' });
var view3 = Ti.UI.createView({ id: 'view3', backgroundColor: '#48b' });
scrollableView.setViews([view1, view2, view3]);

win.add(scrollableView);
win.open();
```


hansemannn:
Hey @drauggres, I just added the docs for your PR, so we're moving forward! One parity issue I noticed is that iOS supports the "width" of the scrollable-view, but Android the padding. Couldn't we "translate" the width to the padding internally? I know there are some good dimension utils that could be used to get the actual width (even when passed via percentage, dp or other). Let me know your thoughts on this one!

drauggres:
- using "width" property same way as iOS could be complicated, for sure it will require more than 50-lines patch,
- "padding" looks like good compromise between "full parity" and no feature at all,
- "padding" is more flexible than just "width", maybe iOS someday should have it.

hansemannn:
I am fine with that, thanks for the thoughts @drauggres !

garymathews (approved):
CR: **PASS**

PR #10081, opened 2018-05-30
[TIMOB-25910] Android: Fix Ti.UI.TableView memory leak

garymathews:
- Remove references to TableView rows

###### TEST CASE
```JS
const root = Ti.UI.createWindow({
        backgroundColor: 'gray'
    }),
    btn = Ti.UI.createButton({
        title: 'OPEN TABLE WINDOW'
    });

btn.addEventListener('click', () => {
    const win = Ti.UI.createWindow({
            backgroundColor: 'red'
        }),
        tableView = Ti.UI.createTableView();

    let rows = [];

    for (let i = 0; i < 64; i++) {
        const row = Ti.UI.createTableViewRow();
        const textField = Ti.UI.createTextField();

        row.add(textField);
        rows.push(row);
    }

    tableView.data = rows;

    win.add(tableView);
    win.open();
});

root.add(btn);
root.open();
```
- Open and close `TABLE WINDOW` multiple times
- Observe memory usage to not increase dramatically after GC
- Application should not crash

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25910)

PR #10174, opened 2018-07-11
[TIMOB-26202] Android: Fix references when releasing views

garymathews:
- Amend references when releasing views

###### TEST CASE #1
```JS
let win = null;

new function create() {

    // create new window
    const newWin = Ti.UI.createWindow({ backgroundColor: 'gray', layout: 'vertical' }),
          btn = Ti.UI.createButton({ title: 'RELOAD' });

    btn.addEventListener('click', create);
    newWin.add(btn);

    // create multiple labels
    for (let i = 0; i < 16; i++) {
        const lbl = Ti.UI.createLabel({ text: '#' + i });
        newWin.add(lbl);
    }

    // display new window
    newWin.open();

    // close existing window
    if (win != null) {
        win.close();
    }
    win = newWin;
}
```
- Should not observe memory leak when reloading window multiple times
  - May need to force a GC to see results
###### TEST CASE #2
```JS
let win = Ti.UI.createWindow({ backgroundColor: 'gray', layout: 'vertical' }),
    lbl = Ti.UI.createLabel({ text: 'LABEL' }),
    btn = Ti.UI.createButton({ title: 'RE-USE' });

    // create a new window
    const newWin = Ti.UI.createWindow({ backgroundColor: 'gray', layout: 'vertical' });
    newWin.open();

    // remove components so we can re-use them
    win.remove(btn);
    win.remove(lbl);

    // close existing window
    win.close();
    win = newWin;

    // re-use previous components
    win.add([btn, lbl]);
});

win.add([btn, lbl]);
win.open();
```
- Both the button and label should be able to be re-used by a new window

##### NOTE: QE should also run the tests from https://github.com/appcelerator/titanium_mobile/pull/10081 (including @ypbnv test case)

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26202)

### DIFF
--- a/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java
+++ b/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java
@@ -134,4 +134,16 @@ public final class ReferenceTable
 		}
 		return ref;
 	}
+
+	/**
+	 * Determines if the reference is strong
+	 *
+	 * @param key the key for the reference.
+	 * @return returns true if the reference is strong
+	 */
+	public static boolean isStrongReference(long key)
+	{
+		Object ref = getReference(key);
+		return !(ref instanceof WeakReference || ref instanceof SoftReference);
+	}
 }
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
@@ -134,10 +134,24 @@ public class TableViewProxy extends TiViewProxy
 	public void releaseViews()
 	{
 		super.releaseViews();
+
 		if (localSections != null) {
 			for (TableViewSectionProxy section : localSections) {
 				section.releaseViews();
 			}
+		}
+	}
+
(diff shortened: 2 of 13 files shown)


android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java
```diff
@@ -134,4 +134,16 @@ public static Object getReference(long key)
...
+	/**
+	 * Determines if the reference is strong
+	 *
+	 * @param key the key for the reference.
+	 * @return returns true if the reference is strong
+	 */
+	public static boolean isStrongReference(long key)
+	{
+		Object ref = getReference(key);
+		return !(ref instanceof WeakReference || ref instanceof SoftReference);
```
sgtcoolguy: what if getReference() returns null? This will return `true`, that it is a strong reference. Is that what we expect/want?

garymathews: I thought about that too 👍 

Yes, `null` returning `true` is okay here since we default to strong references and have validation for `null` references.

PR #10175, opened 2018-07-12
[TIMOB-26203] Android: Prevent exception in Log for HTTP response

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26203

**Description:**
Guard for a `null` response from the server.

Note: Can we have a unit test for that?

**Test case:**
```JS
var url = "https://www.appcelerator.org/exit.xml";

Ti.API.info("checkURL - Checking URL: " + url);
// createHTTPClient       
  onload : function(e) {
    // if success do           
    Ti.API.info("checkURL Success - URL Found: " + url);
  },
  onerror : function(e) {
    // on error do            
    Ti.API.info("checkURL Error - File URL Not Found");
  },
});

// open the connection with client.open with method and url
client.open('HEAD', url);
// Send the request.
client.send(); 
```


garymathews (approved):
CR: **PASS**

Yes, we should be able to write a test case for this

sgtcoolguy:
@ypbnv I mean does that manual test case you wrote above work to expose the issue and passes after the fix? If so, that's pretty easy to write up as a mocha test:

```javascript
describe('Ti.Network.HTTPClient', function () {
  it('handles null input stream in response', function (finish) {
    var url = "https://www.appcelerator.org/exit.xml";
      onload : function(e) {   
        finish(new Error('Expected network call to fail'));
      },
      onerror : function(e) {
      // TODO: Test some of the event callback properties?
       finish();
      },
    });

    client.open('HEAD', url);
    client.send();
  });
});
```

ypbnv:
@sgtcoolguy If I understand the test case you proposed right it will pass if the `onerror` callback is called, right? Which is the case before this fix, but it will log an Android exception too.

lokeshchdhry:
FR Passed.

If `url` is not present & server returns `null` then `checkURL Error - File URL Not Found` is logged.

Studio Ver:     5.1.0.201807181611
SDK Ver:        7.4.0 local build
OS Ver:         4.2.13
Xcode Ver:      7.0.4
Appc NPM:       1.1.3
Appc CLI:       10.13.5
Daemon Ver:     Xcode 9.4.1
Ti CLI Ver:     5.1.1
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.1

PR #10177, opened 2018-07-12
[TIMOB-26197] iOS: Deprecate Ti.Media.SystemAlert

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26197


sgtcoolguy:
@jquick-axway @infosia Before I merge this PR, is there the idea of an equivalent API for Ti.Media.SystemAlert for Android or Windows? The API simply plays a single sound (I am not sure what formats it accepts, or max length though there's mention of < 30 second) using the ringtone volume (not music volume).

We only have this on iOS, so we should support another platform or move it out (as Hans does here). Even if we do support multiple platforms, it may still make sense to be moved to it's own module as it seems to be a pretty specific and small use case.

hansemannn:
@sgtcoolguy See the discussion around it in the linked JIRA ticket. It seems like Android and Windows only have the AudioPlayer API for this kind, but no system-internal API to handle special sounds.

jquick-axway:
We can replicate something similar on Android. We would just use AudioPlayer's implementation hard-coded to use the ringer volume. And I would ignore the 30 second limitation too. I am okay with doing this in the core if you are if you think many devs use this API.

sgtcoolguy:
@Topener Can you ask around the community about this API? (AKA, is it used much at all? So we can determine if it'd make sense to keep core or move out and what the priority would be on trying to achieve parity)

My personal feeling is that it's unlikely it's widely used, and that we likely should move it out to a separate native module - but that even then it may make sense to have an equivalent Android impl that is backed by AudioPlayer (though it'd be unlikely to be high priority as a non-core module API).

hansemannn:
@Topener ping

Topener:
I've been asking around and have yet to find a person who has ever used it. So makes sense to strip out imho.

jquick-axway:
Okay. Sounds good to me.

PR #10178, opened 2018-07-12
Use Karma for unit testing

janvennemann:
This is a first iteration to replace our mocha unit testing suite with Karma using the new Titanium launcher. This will remove a lot of manual steps and streamline our unit testing to use established tools.

- [x] Use Karma to utilize the unit testing framework and reporters we need
- [x] Allows us to write unit tests directly in this repo making the mocha unit testing suite obsolete (No more copying of tests and manual report parsing, yay!).
- [ ]  Use mocha as the testing framework. This will allow us to just copy the tests from the mocha suite and reuse them here.
- [ ] Changes to the Jenkins build need some love. I just did the minimal changes to get it working but the mocha suite did some additional stuff behind the scenes to install and clean up the SDK built by Jenkins.
- [ ] Test behavior when running tests in parallel


package.json
```diff
@@ -18,7 +18,9 @@
   ],
   "author": "Appcelerator, Inc. <npmjs@appcelerator.com>",
   "scripts": {
-    "test": "grunt"
+    "test": "grunt",
+    "test:unit:android": "grunt karma:android",
+    "test:unit:ios": "grunt karma:ios"
```
hansemannn: 🙏

Gruntfile.js
```diff
@@ -35,6 +35,21 @@ module.exports = function (grunt) {
...
+				configFile: 'test/unit/karma.unit.config.js'
+			},
+			android: {
+				junitReporter: {
+					outputFile: 'junit.android.xml'
+				}
+			},
+			ios: {
+				junitReporter: {
+					outputFile: 'junit.android.xml'
```
sgtcoolguy: junit.ios.xml

Jenkinsfile
```diff
@@ -268,11 +244,8 @@ timestamps {
 
 		// Run unit tests in parallel for android/iOS
 		stage('Test') {
-			parallel(
-				'android unit tests': unitTests('android', nodeVersion, npmVersion, targetBranch),
-				'iOS unit tests': unitTests('ios', nodeVersion, npmVersion, targetBranch),
-				failFast: true
-			)
+			unitTests('android', nodeVersion, npmVersion, targetBranch),
```
sgtcoolguy: why not in parallel? you can set failFast: false if you want both to run to completion even if one of them fails... (so you can try and fix issues in either/both each round of commits 😄 )

sgtcoolguy:
So the reason we have the horribly convoluted separate repo/override stuff is because the suite is intended to be common cross *all* platforms, including Windows which is not in this repo.

Moving the suite here certainly makes life easier to Android/iOS devs, but presents an issue for Kota and the Windows SDK.

sgtcoolguy:
Hey @janvennemann I pushed a couple commits to resolve the package-lock.json merge conflict.

janvennemann:
Closing this for now. I'll open another PR inside our unit testing suite to use the karma-launcher there.

PR #10184, opened 2018-07-17
Update to node-titanium-sdk 0.5.0

sgtcoolguy:
* [TIMOB-26173] Android: Add the Android Studio default SDK/NDK install paths to the Android lookup
* [TIMOB-26126] Update emulator path

**Optional Description:**
Incorporate recent bug fixes from node-titanium-sdk

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1247,9 +1247,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30000856",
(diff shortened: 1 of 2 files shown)


hansemannn:
Should https://github.com/appcelerator/node-titanium-sdk/pull/40 bump it to 0.6.0?

sgtcoolguy:
@hansemannn Yes, that looks like a new "feature" to me, so you should bump node-titanium-sdk master to 0.6.0, then open a PR to update the SDK to use that.

PR #10186, opened 2018-07-18
[TIMOB-19618] Android: Implement Ti.UI.WebView.mixedContentMode

garymathews:
- Allows the loading of insecure resources from a secure origin

###### TEST CASE
```JS
const win = Ti.UI.createWindow({ backgroundColor: 'gray' }),
          url: 'https://www.bennish.net/mixed-content.html',
          mixedContentMode: true
      });

win.add(webView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19618)


apidoc/Titanium/UI/WebView.yml
```diff
@@ -608,6 +608,13 @@ properties:
     since: "1.8.0"
     default: true
 
+  - name: mixedContentMode
+    summary: If `true`, allows the loading of insecure resources from a secure origin.
+    type: Boolean
+    platforms: [android]
+    since: "7.4.0"
+    default: false
+
```
jquick-axway: Since we only set this upon creation, we need...
`availability: creation`

The rest looks great. Thanks!

jquick-axway (changes requested):

apidoc/Titanium/UI/WebView.yml
jquick-axway: And perhaps in the description we should state that iOS users should use the `NSAllowsArbitraryLoads` plist setting.

jquick-axway (approved):
CR: Pass

PR #10190, opened 2018-07-19
[TIMOB-25955] Support crash analytics refactor

garymathews:
- Support recent crash analytics refactor

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25955)


iphone/Classes/KrollContext.m
```diff
@@ -659,7 +659,6 @@ - (void)invoke:(KrollContext *)context
     id excm = [KrollObject toID:context value:exception];
     [[TiExceptionHandler defaultExceptionHandler] reportScriptError:[TiUtils scriptErrorValue:excm]];
     pthread_mutex_unlock(&KrollEntryLock);
-    @throw excm;
```
sgtcoolguy: why not throw the exception?

garymathews: We already handle the exception on the lines above. Throwing an exception causes the app to crash with an exception that cannot be caught since it's throwing a `KrollObject` and not a valid exception.

sgtcoolguy:
@garymathews With the merge from `next` branch, the APSAnalytics and Kroll classes have changed locations in the repo and are part of the pre-built TitaniumKit framework. I assume that's the cause of the conflicts.

PR #10193, opened 2018-07-20
[TIMOB-26119] TiAPI: Deprecate getter/setter accessor methods for properties

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26119

**Description:**
I wrote a manual test for this as there's no way in our current mocha suite to assert on the output that a test produces in its logs.

Basically, we want to spit out a deprecation for getters/setter methods which just wrap a property. In JS style, a property should be access using typical obj.property, obj['property] fashion.

These generated getter/setters are more of a Java-style language thing. Now under the hood we may hook a JS property to a method named like this internally, but we shouldn't expose that.

If a method *doesn't* wrap a property but happens to start with get or set, we shouldn't log a notice.

There are possible pairs of methods that should be moved to properties (i.e. Ti.Filesystem.File#getProtectionKey()/setProtectionKey()), but I didn't try to "fix" these.

I *did* fix the exposure of Ti.Filesystem.File.directoryListing property on Android. There is no such property and we should be exposing #getDirectoryListing() only.

This PR is in anticipation of cleaning up the API to remove these. The first platform that'd benefit would be Windows to avoid the boilerplate code to re-expose getter/setters for properties. Additionally, it'd avoid the need to do the same for any proxies we eventually moved to using the Obj-C JSC API (#10069). On Android, there would be minimal cleanup since under the hood we use this naming style anyhow, but we could remove tagging with both the @Kroll.method and @Kroll.getProperty on many proxy methods.

```javascript
/*
 * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
 */
/* eslint-env mocha */
/* global Ti */
'use strict';
var should = require('./utilities/assertions'),
	utilities = require('./utilities/utilities');

describe.only('Example', function () {
	it('Should pass', function () {
		var file = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, 'app.js'),
			dir = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory),
			label = Ti.UI.createLabel({ text: 'original text' });

		console.log('Ti.Filesystem.File#getDirectoryListing() - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		dir.getDirectoryListing();

		console.log('Ti.Filesystem.File#getHidden() - should log deprecation notice'); // Android: ✅, iOS: ✅
		file.getHidden();

		console.log('Ti.Filesystem.File.hidden - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		file.hidden;

		console.log('Ti.UI.Label#getText() - should log deprecation notice'); // Android: ✅, iOS: ✅
		label.getText();
		console.log('Ti.UI.Label.text - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		label.text;
		console.log('Ti.UI.Label#setText() - should log deprecation notice'); // Android: ✅, iOS: ✅
		label.setText('new text');
		console.log('Ti.UI.Label.text= - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		label.text = 'new text';

		// Test with getter/setter methods that aren't for a property!
		console.log('Ti.App.Properties.getString() - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		Ti.App.Properties.getString('ti.defaultunit');
		console.log('Ti.App.Properties.setString() - should NOT log deprecation notice'); // Android: ✅, iOS: ✅
		Ti.App.Properties.getString('whatever', 'yeah');

		if (utilities.isIOS()) {
			console.log('Ti.Filesystem.File#setHidden() - should log deprecation notice'); // iOS: did log, but did so immediately after setText deprecation notice!
			file.setHidden(true); // FIXME: AAAaaahhh Android doesn't have this accessor!

			console.log('Ti.Filesystem.File.hidden= - should NOT log deprecation notice'); // iOS: ✅
			file.hidden = true; // FIXME: AAAaaahhh Android doesn't have this accessor!

			// Test get/set methods taking typical 0 and 1 arg that aren't "property" accessors
			// Note that these are a good candidate for moving to typical properties though!
			console.log('Ti.Filesystem.File#getProtectionKey() - should NOT log deprecation notice'); // iOS: ✅
			file.getProtectionKey();
			console.log('Ti.Filesystem.File#setProtectionKey() - should NOT log deprecation notice'); // iOS: ✅
			file.setProtectionKey(Ti.Filesystem.IOS_FILE_PROTECTION_NONE);
		}
	});
});

```


hansemannn (approved):
Looks great, works great! Thanks @sgtcoolguy!

hansemannn (commented):
@sgtcoolguy We should also deprecate the accessors via the docs, e.g. [here](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/docgen.js#L330) (getter) and [here](https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/docgen.js#L346) (setter).

apidoc/docgen.js
```diff
@@ -327,7 +327,7 @@ function generateAccessors(apis, className) {
 			rv.push({
 				name: 'get' + api.name.charAt(0).toUpperCase() + api.name.slice(1),
 				summary: 'Gets the value of the <' + className + '.' + api.name + '> property.',
-				deprecated: api.deprecated || '8.0.0',
+				deprecated: api.deprecated || { since: '8.0.0' },
```
hansemannn: Can we also add `note: 'Access the property directly` or related phrasing?

hansemannn:
@sgtcoolguy I changed the base to `next` and it now includes the current master changes as well, which should be fine to keep it in sync. Can we merge this?

sgtcoolguy:
@hansemannn I want to keep the history easier to work with on "next", so I want to squash all my commits down to one. I'll manually merge, I suppose.

sgtcoolguy:
Manually merged to 'next' branch

PR #10194, opened 2018-07-20
[TIMOB-26211] iOS: Round layout percentage down instead of up (parity)

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26211

> `no_tests` since UI-related change.


jquick-axway (approved):
CR: Pass

Also, Android's layout code will floor the percentage value too...
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java#L384

PR #10199, opened 2018-07-24
[TIMOB-25391] iOS: Fix zombie crash in Ti.UI.TableView

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25391

> `no tests` as no new API's. All existing tests should still pass.


vijaysingh-axway:
@hansemannn  I guess problem is that we are not retaining headerViewProxy/footerViewProxy while these are getting set in - (void)setHeaderView_:(id)args / - (void)setFooterView_:(id)args. Due to this it may be that  headerViewProxy/footerViewProxy has already released . And inside dealloc we are calling a method on these.
I think we should retain it in above methods and should be released inside dealloc. Can you please check once? Thanks!

hansemannn:
@vijaysingh-axway The crash did not occur prior to the last change that was now partly reverted and replaced by a more solid solution that is also used in the TiUIListView implementation. We have been unable to receive an actual test-case, but with this fix, it did not crash at all. If you can give a concrete example of your solution, feel free to submit it!

vijaysingh-axway (approved):
CR passed.

jatinj:
What fix do I need to make in Appcelerator or Titanium studio in order to get free of this error. I am getting this error in SDK 7.4.1, Appclerator Studio 5.0.0, testing on iPhone 6 with v ios 12.1 

Thread 1: EXC_BAD_ACCESS (code=1, address=0x10)

 if (footerViewProxy != nil) {
    [footerViewProxy setProxyObserver:nil];
    [[self proxy] forgetProxy:footerViewProxy];
    footerViewProxy = nil;
  }

PR #10204, opened 2018-07-24
[TIMOB-26209] Only error out if selected xcode EULA has not been accepted.

feons:

If any Xcode installs don't have EULA accepted (even one not being used), build won't work.


cb1kenobi (approved):
CR'd. I haven't tested ioslib@1.7.x, but it seems the only changes are a few minor npm dep updates, xcode 10 support, and the eula bug fix.

PR #10205, opened 2018-07-24
[TIMOB-26209] Only error out if selected xcode EULA has not been acce…

feons:
…pted

If any Xcode installs don't have EULA accepted (even one not being used), build won't work.


cb1kenobi:
@feons This PR also needs to update the `package-lock.json`. Apparently running `npm install` will fix it, but that will probably also update a bunch of other things too.

PR #10065, opened 2018-05-18
[TIMOB-26055] Android: Handle injected build variables

garymathews:
- Handle injected build variable `applicationId` in `AndroidManifest.xml`
- Related to https://github.com/appcelerator/titanium_mobile/pull/10002

###### tiapp.xml
```XML
<module platform="android">facebook</module>
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26055)

PR #10206, opened 2018-07-24
[TIMOB-26055] Android: Fixed build failure if AAR AndroidManifest.xml contains multiple ${applicationId}

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26055

**Summary:**
- Continuation of PR #10065 
- Previous fix was only replacing 1st instance of `${applicationId}` in an AAR library's "AndroidManifest.xml" instead of all instances.
- Note that JS `String.replace(string, string)` only replaces one instance, but `String.replace(regex, string)` replaces all instances.

**Test:**
1. Create a project targeting Titanium 7.3.0 or higher.
2. Add module: [titanium-firebase-core](https://github.com/hansemannn/titanium-firebase-core) v2.2.0
3. Build for Android.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -3888,9 +3888,9 @@ AndroidBuilder.prototype.generateAndroidManifest = function generateAndroidManif
 		if (fs.existsSync(libraryManifestPath)) {
 			let libraryManifestContent = fs.readFileSync(libraryManifestPath).toString();
 
-			// handle injected build variables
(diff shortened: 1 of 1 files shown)


hansemannn (approved):
CR + FT passed, PR **approved**. Also tested with the related modules and they work fine now.

jquick-axway:
@hansemannn, thanks for testing it out. 👍

PR #10210, opened 2018-07-25
[TIMOB-26201] Remove all deprecated API's

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26201

> **Note 1**: I left out the Android Ti.Geolocation "legacy mode" deprecation removal. It's deprecated since 2.0.0, but the affected components are too complex to squeeze it into this PR. It should be properly removed by the Android team. I created [TIMOB-26132](https://jira.appcelerator.org/browse/TIMOB-26132) to track that.

> **Note 2**: As of for now, I also removed the docs for deprecated and previously removed API's. Maybe we should leave the `removed:` tag to let our devs know that is was removed?


hansemannn:
@sgtcoolguy This is the last 8.0.0 PR that could mess up my Swift-setup environment at this point. Once merged, I can rebase the other one again and we have a solid iOS core.

iphone/Classes/TiCalendarCalendar.m
```diff
@@ -174,99 +174,6 @@ - (NSArray *)getEventsBetweenDates:(id)args
   return [TiCalendarEvent convertEvents:events withContext:[self executionContext] calendar:[self calendar] module:module];
 }
 
-- (NSArray *)getEventsInDate:(id)arg
```
sgtcoolguy: I don't see this as deprecated in our apidocs, and I don't think this was removed on Android...

iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -883,10 +801,6 @@ - (void)handleLoadStateChangeNotification:(NSNotification *)note
         [self fireEvent:@"load" withObject:nil];
       }
 
-      if ([self _hasListeners:@"preload"]) {
-        DEPRECATED_REPLACED(@"Media.VideoPlayer.preload", @"7.0.0", @"Media.VideoPlayer.load");
```
sgtcoolguy: our apidocs do not note this as deprecated, and it seems to be supported on Android? (though looking at the impl, they just fire at the same time_.

sgtcoolguy (changes requested):
There's a handful of changes still to be made, and I think we need to be more consistent in how we mark apis as removed or delete them.

On a related note, reviewing this makes it abundantly clear that we suck at marking methods/events/types/properties deprecated in a nice way that would make removing them easier on us in the future. We're not making use of the nice `@deprecated` annotation in Java at all here, and we're giving no tips about which version it was deprecated in. We should try and come up with some nice standard for each platform so it's as simple as doing some search and replace to find them.

hansemannn:
I feel like we should rework this PR and introduce a different concept:
- Leave all `removed` docs to let our developers know that something was removed
- Introduce the following "API's":
  - iOS: We already make heavy use of `DEPRECATED_REPLACED` and `DEPRECATED_REMOVED` across the SDK
  - Android: Use `@deprecated("7.0.0")` and `@removed("8.0.0")` annotations?
  - Windows: Something related to Android? (cc @infosia)

Working on this PR was not critical, only time-consuming. Let's start this over and do it right this time.

infosia:
Added [TIMOB-26236](https://jira.appcelerator.org/browse/TIMOB-26236) to track this for Windows.

PR #10211, opened 2018-07-25
[TIMOB-26104] Android: Picker inside Tableview resets value

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26104

**Description:**
Persist the selection in a Picker in case it has to be refreshed\added again after being removed.

**Note:** This may not be the best experience if during the absence of the Picker from the screen it's adapter has been changed. Do you think it will be fine to reset the persisted index in that case?

**Test case:** 
Removing and adding the picker again
_app.js_
```js
var picker = Ti.UI.createPicker({ width: '50%' });
var rows = [];

for (var index = 0; index < 20; index++) {
	rows.push(Ti.UI.createPickerRow({ title: 'Item ' + (index + 1).toString() }));
}

picker.add(rows);

window.add(picker);

function createButtonText() {
	return window.children.includes(picker) ? 'Remove Picker' : 'Add Picker';
}

	title: createButtonText(),
	bottom: '20dp',
});

	if (window.children.includes(picker)) {
		window.remove(picker);
	} else {
		window.add(picker);
	}

	button.title = createButtonText();
});

window.add(button);
window.open();
```

Using a Picker in a TableView
_app.js_
```
var currentwindow = Ti.UI.createWindow();

var tabledata = [];

var picker = Ti.UI.createPicker({ top:20 });

var data = [];
data[0]=Ti.UI.createPickerRow({title:'Bananas'});
data[1]=Ti.UI.createPickerRow({title:'Strawberries'});
data[2]=Ti.UI.createPickerRow({title:'Mangoes'});

picker.add(data);

var myrow=Ti.UI.createTableViewRow({ 'height' : '50sp' });

myrow.add(picker);
tabledata.push(myrow);

var infotableview = Titanium.UI.createTableView({ data:tabledata });

currentwindow.add(infotableview);
currentwindow.open();
```

Unit test added.


jquick-axway:
**Question:**
Do we need to store the last selection for each column too for pickers that have columns? We probably do, right? I'm okay with us writing that up as a separate ticket in order to resolve the community's issues today. Single selection pickers are the most common case.

> This may not be the best experience if during the absence of the Picker from the screen it's adapter has been changed. Do you think it will be fine to reset the persisted index in that case?

To support this, our proxy's getter would have to return the proxy's `lastSelectedIndex` member variable instead of returning the index from native picker.

garymathews:
@hansemannn @ypbnv iOS failed `ios.Ti.UI.Picker | Selected index persistance`

Is there an issue here?

hansemannn:
@garymathews `postlayout` on iOS is only triggered if the layout actually changed. If the frame / origin of the view remains (like in this case), it is not fired. Can we change the test to not use the picker's `postlayout` event but the window's `open` / `focus` event? I rather would not change the iOS side to trigger a postevent on row-changes, since it seems Android-specific.

ypbnv:
@hansemannn I will see if the test can use something different. I tried at first with events from the window, but got some mismatched timings. 
And BTW is the disparity of when the postlayout is fired on Android and iOS something we should address? I have used it in a number of unit tests and it sounds like I may have to revisit that.

ypbnv:
@hansemannn I am not particularly proud with the workaround for the unit test, but if I got it right now it should be fine on iOS too. 

@jquick-axway Let's separate the multiple column persistence in another ticket yes. I think it can be put together with the persistence across different data sets for both single and multiple column pickers.

jquick-axway (approved):
CR: Pass

jquick-axway:
@hansemannn, @ypbnv,

I just tested add/removing the following picker types:
- For multicolumn spinner, both Android and iOS forget last selection.
- For date picker, iOS resets to current time. (Android remembers last selection.)
- For time picker, iOS resets to current time. (Android remembers last selection.)

PR #10215, opened 2018-07-26
[TIMOB-26237] (7_3_X) iOS: Fix local notifications triggered when the app is closed

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26237

> `no_tests`, since it's a UI-related change that requires the app to be killed and relaunched.

Also adding missing deprecation notes that have been left over from 7.3.0.


vijaysingh-axway (approved):
CR passed

PR #10219, opened 2018-07-30
[TIMOB-26240] Android: Removed harmless activity errors/warnings logged on startup

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26240

**Summary:**
- Removed `null` activity error/warning log messages.
  * It's up to the caller to decide if they should be logged, not the method.
  * It's okay for these methods to return `null` if there are no activities, such as when the app was launched in the background via a `BroadcastReceiver`.
- Changed TiApplication.getInstance() handling:
  * Changed "tiApp" from weak reference to strong. Guaranteed to exist for lifetime of app, so, it was unnecessary.
  * Removed TiApplication.getInstance() null checks. Not necessary.

**Test:**
1. Create a new Titanium app project.
2. Build and run on an Android device.
3. Wait for the app to startup.
4. View the Android log and verify the following have **not** been logged.

```
[ERROR] :  TiApplication: (main) [3,913] No valid root or current activity found for application instance
[WARN] :   TiAndroid: (main) [2,915] Application instance no longer available. Unable to get current activity.
```

### DIFF
--- a/android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
+++ b/android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
@@ -580,26 +580,18 @@ public class AndroidModule extends KrollModule
 	public ActivityProxy getCurrentActivity()
 	// clang-format on
 	{
-		TiBaseActivity resultBaseActivity = (TiBaseActivity) TiApplication.getAppCurrentActivity();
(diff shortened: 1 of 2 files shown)


garymathews (approved):
LGTM!

PR #10221, opened 2018-07-31
[TIMOB-26242] Android: Attempt to fix memory behaviour

garymathews:
- Attempt to fix regressions from recent memory management refactor
- The changes for [TIMOB-25910](https://jira.appcelerator.org/browse/TIMOB-25910) in `releaseKroll()` are not necessary since the recent memory management refactor
- V8 does not account for the size of objects in JVM, so it does not GC as often as it should. This PR aims to increase the rate of V8 GC by coinciding with JVMs GC, allowing JVM to free objects.

#### OBJECT LIFECYCLE

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26242)

### DIFF
--- a/android/runtime/v8/src/native/V8Runtime.cpp
+++ b/android/runtime/v8/src/native/V8Runtime.cpp
@@ -348,9 +348,13 @@ JNIEXPORT jboolean JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nati
 	// while (v8::platform::PumpMessageLoop(V8Runtime::platform, V8Runtime:v8_isolate)) continue;
 	// v8::platform::RunIdleTasks(g_platform, isolate, 50.0 / base::Time::kMillisecondsPerSecond);
 
-	// FIXME What is a good value to use here? We're basically giving it 100 ms to run right now
-	double deadline_in_s = V8Runtime::platform->MonotonicallyIncreasingTime() + 0.1;
-	return V8Runtime::v8_isolate->IdleNotificationDeadline(deadline_in_s);
+	// notify V8 of low memory to suggest a full GC
+	V8Runtime::v8_isolate->LowMemoryNotification();
+	return true;
+
+	// give GC time to perform cleanup (1 second)
+	// double deadline_in_s = V8Runtime::platform->MonotonicallyIncreasingTime() + 1;
+	// return V8Runtime::v8_isolate->IdleNotificationDeadline(deadline_in_s);
 }
 
 /*
--- a/android/runtime/v8/src/native/V8Runtime.cpp
+++ b/android/runtime/v8/src/native/V8Runtime.cpp
@@ -199,7 +199,7 @@ using namespace titanium;
  * Method:    nativeInit
  * Signature: (Lorg/appcelerator/kroll/runtime/v8/V8Runtime;)J
  */
-JNIEXPORT void JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nativeInit(JNIEnv *env, jobject self, jboolean useGlobalRefs, jobject debugger, jboolean DBG, jboolean profilerEnabled)
+JNIEXPORT void JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nativeInit(JNIEnv *env, jobject self, jobject debugger, jboolean DBG, jboolean profilerEnabled)
 {
 	if (!V8Runtime::initialized) {
 		// Initialize V8.
@@ -215,7 +215,6 @@ JNIEXPORT void JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nativeIn
 
 	titanium::JNIScope jniScope(env);
 
-	JavaObject::useGlobalRefs = useGlobalRefs;
 	V8Runtime::DBG = DBG;
 
 	V8Runtime::javaInstance = env->NewGlobalRef(self);
--- a/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java
+++ b/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java
@@ -1414,32 +1414,8 @@ public class KrollProxy implements Handler.Callback, KrollProxySupport, OnLifecy
 	 */
 	public void releaseKroll()
 	{
-		// TIMOB-25910: attempt to make any strong references into weak references
-		// NOTE: there is an issue where proxies created in another context (such as an event listener callback)
-		// may not be destroyed when the context has ended. This workaround allows the GC to free these objects
-		// if memory is constraint.
-		if (referenceKey != -1) {
-			try {
-				final Class referenceTableClass = Class.forName("org.appcelerator.kroll.runtime.v8.ReferenceTable");
-				final Method getReferenceMethod = referenceTableClass.getDeclaredMethod("getReference", long.class);
-
-				final Object reference = getReferenceMethod.invoke(null, this.referenceKey);
-				if (!(reference instanceof WeakReference || reference instanceof SoftReference)) {
-					final Method weakReferenceMethod =
-						referenceTableClass.getDeclaredMethod("makeWeakReference", long.class);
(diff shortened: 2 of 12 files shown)


lokeshchdhry:
FR Passed.

Ran previous test cases & test case in the ticket & GC seems to work as expected.

Studio Ver:     5.1.0.201807191252
SDK Ver:        7.4.0.v20180801124615
OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.4
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.12.0
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.1

android/runtime/v8/src/native/V8Runtime.cpp
```diff
@@ -348,9 +348,13 @@ JNIEXPORT jboolean JNICALL Java_org_appcelerator_kroll_runtime_v8_V8Runtime_nati
 	// while (v8::platform::PumpMessageLoop(V8Runtime::platform, V8Runtime:v8_isolate)) continue;
 	// v8::platform::RunIdleTasks(g_platform, isolate, 50.0 / base::Time::kMillisecondsPerSecond);
 
-	// FIXME What is a good value to use here? We're basically giving it 100 ms to run right now
-	double deadline_in_s = V8Runtime::platform->MonotonicallyIncreasingTime() + 0.1;
-	return V8Runtime::v8_isolate->IdleNotificationDeadline(deadline_in_s);
+	// notify V8 of low memory to suggest a full GC
+	V8Runtime::v8_isolate->LowMemoryNotification();
```
sgtcoolguy: I think this is a pretty fundamentally different method to be calling here.

This basically informs that memory is low and V8 should try to GC pretty aggressively, which isn't necessarily the case here.

I think really this method should take the deadline increment as an arg and the Java side should probably use some sort of backoff function to increasingly allocate more time to do GC so long as we're still idle and the idle notification deadline call returns false (meaning it wants to do more GC).

android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java
```diff
@@ -1414,32 +1414,8 @@ public void release()
...
-				}
-			} catch (Exception e) {
-				// do nothing...
-			}
-		} else {
-			if (krollObject != null) {
-				krollObject.release();
-			}
+		if (krollObject != null) {
+			krollObject.release();
```
sgtcoolguy: I certainly think that this revert cleans up a not so elegant change that was made to try and fix memory... so +1 to making it cleaner.

sgtcoolguy (commented):
So I see what you're aiming for here, but I still think it's not necessarily the right way to do things. 

It looks like you're basically offering up a sacrificial lamb to the JVM to GC and when it does you then tell V8 that we're low on memory and it tries to aggressively GC itself (at least on the next idle callback in the JVM). So in a way, you're sort of syncing up the GC on each side, but this seems pretty brute force.

I do think that the memory model/management is difficult to deal with now and I was just trying to get something working when I updated V8 long ago.

I think the fundamental issue here is that we have two GC'ing VMs that each have references to one half of the full object which lives across environments.

To my mind, that means we should be keeping objects alive on both halves until each side has agreed that they don't need it anymore. Or else we get JS objects with no backing Java proxy, or Java proxies whose JS object/pointer has been cleared.

We've been trying to get this right with reference tables using weak/soft/strong references, etc. I'm still not sure we're doing it right.

From the looks of the recents string of PRs, it appears we're not.

Your flowchart here does represent the ideal scenario but only from the V8 side. We create a JS object and a Java proxy, stick a (strong) reference to the Java proxy in a Map, mark the JS object as weak/independent and wait for a V8 GC callback.
When that comes, we revive the JS object and move the Java reference to become weak and then wait for the JVM to GC it. We then wait for finalize() to get called on the object, call V8Object.nativeRelease() and that will delete the native proxy if the JS side is "detached".

So I think there's a handful of things we haven't really looked into here:
- What if we get to the end and the native proxy/JS object is not "detached"? Won't it stay alive in V8 forever? (and be dead on the JVM side?)
- What if we want to initiate the cleanup from the JVM side? It looks like the reflection stuff in KrollObject was intended to try and fix that?

I think we basically need to have a two-pass GC system on both native and JVM sides; that we need to simplify the code on the native side; and that neither side gets to get fully GC'd unless both sides agree for the same object.

We have no real mechanism for "reviving" the Java half if the JVM wants to GC it. It (may) call finalize() and try to release the JS half, but if the JS half hasn't been asked to be GC'd yet, it won't be.

garymathews:
@sgtcoolguy I'll try to explain what's happening here and the reasons behind the changes in this PR which should answer some of your points.

The idea here is to have JVM as the dominant GC because it takes into account the actual memory consumption of the allocated objects.

However, due to the ReferenceTable JVM cannot collect objects currently referenced by V8. This is why we need to suggest V8 to do a _full_ GC with `LowMemoryNotification()`. Otherwise, it would do a partial GC and a lot of objects in our ReferenceTable will not be updated to allow JVM to collect them.

This is why it's hard to make each runtime _agree_ that an object can be collected. V8 may never agree some objects should be collected as V8 can't see the actual resources they are consuming. It only sees small proxy object that may not be worth its time to GC. When in reality, that object could be consuming lots of memory in JVM. That's why I'm forcing V8 to do a full collection.

Forcing V8 to do a full GC isn't too taxing either as the objects its collecting are small in size. And it will only do a GC at the intervals of JVMs GC (which would be at idle times).

- If a V8 proxy is destroyed it will remove its Java counterpart from the ReferenceTable, allowing it to be collected by JVM.
- A Java object should never be collected prematurely, it can only be collected if V8 has allowed its reference to be downgraded or removed from the ReferenceTable. So the scenario of a V8 object being alive without its Java counterpart should not be possible.

sgtcoolguy:
So, certainly the JVM is aware of how big all the Java objects are - but it's unaware of how much memory V8 is taking, and vice versa.

There is an API to try and inform V8 about external memory being held by objects: 
http://peerigon.github.io/v8-docs/classv8_1_1Isolate.html#aaeda5fa60961a3d9d476c46200e30711

We likely should be trying to make use of that. I would have assumed perhaps there'd be a way to associate specific objects with external memory, but I couldn't find that. But also, it may be rather difficult to get the size of java objects quickly/easily. I do see a jvmti method for it, but that is meant for tooling (i.e. debuggers/profilers).

I suppose given our current state, this PR seems Ok, but again I am very hesitant to be moving from our idle notifications to a full out low memory notification. If called enough (or with enough time for it's deadline) the idle notification can also trigger full GCs.

I tried to clean this code up locally, but it's a jumble. One thing I think we can do is remove the useGlobalRefs code branches since it seems we've already modified the code so it's always false now... Also The KrollProxy/ReferenceTable passing of reference keys can now be removed as it's unused (since you removed the reflection hack here).

garymathews:
I did look at [AdjustAmountOfExternalAllocatedMemory()](http://peerigon.github.io/v8-docs/classv8_1_1Isolate.html#aaeda5fa60961a3d9d476c46200e30711), but it would pretty much achieve the same result as what this PR does.

> Registering externally allocated memory will trigger global garbage collections more often than it would otherwise in an attempt to garbage collect the JavaScript objects that keep the externally allocated memory alive.

But we would have the added complexity of sampling our memory usage to update it. And a full GC may not happen soon enough, which could prevent functionality that requires large memory resources (image loading etc..) if memory is constraint. That's why I opted not to use `IdleNotificationDeadline()` with a longer time, as I could not guarantee a full GC would occur when it needs to.

sgtcoolguy:
@garymathews OK, I'm convinced 😄 

Please remove the useGlobalRefs related code in JavaObject/V8Runtime since it's no longer used - should make the code a little easier to follow. Also, please remove the ReferenceTable/KrollProxy setReferenceKey method/calls since those are also now unused. Then I think we're good to merge.

PR #10222, opened 2018-08-01
[TIMOB-24605] Android: Fixed crashes caused by loading image URLs when JS is running on main thread

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-24605
- https://jira.appcelerator.org/browse/TIMOB-26249

**Summary:**
- Fixed image URL crash if JS running on main thread for:
  * [TIMOB-24605](https://jira.appcelerator.org/browse/TIMOB-24605) `View` object's "backgroundImage" property.
  * [TIMOB-26249](https://jira.appcelerator.org/browse/TIMOB-26249) `ImageView` object's `image` property while "autorotate" is `true`.
  * `ImageView` object's "defaultImage" property.
- Modified Java "TiImageHelper" class to support reading JPEG EXIF orientation from a stream.
  * Allows `ImageView` to auto-rotate JPEG downloaded from the Internet if needed.
- Added new Java class "TiInputStreamWrapper".
  * Used to close the `HttpUrlConnection` when the stream has been closed.
- Added redirect endpoint fetching methods to Java "TiResponseCache".
  * Needed to fetch cached image when using a URL that redirects. Major optimization.

---
**Test 1:**
1. Uninstall the app from the device if already installed. (Will delete cached web responses.)
2. Make sure the Android device has Internet access.
3. Set "tiapp.xml" file property "run-on-main-thread" to `true`.
4. Build and run the below code.
5. Verify that the image was successfully displayed without crashing.

```javascript
window.add(Ti.UI.createView({
	// This URL redirects from HTTP to HTTPS.
	backgroundImage: "http://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_1.jpg",
}));
window.open();
```

---
**Test 2:**
1. Uninstall the app from the device if already installed. (Will delete cached web responses.)
2. Make sure the Android device has Internet access.
3. Set "tiapp.xml" file property "run-on-main-thread" to `true`.
4. Build and run the below code.
5. Verify that the image was displayed "upright" successfully, without crashing.

```javascript
var imageView = Ti.UI.createImageView({
	// This URL redirects from HTTP to HTTPS.
	image: "http://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_3.jpg",
	autorotate: true,
});
window.add(imageView);
window.open();
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
@@ -88,7 +88,7 @@ public class TiDownloadManager implements Handler.Callback
 	 * <p>
 	 * Returns null if failed to download content or if given an invalid argument.
 	 */
-	public InputStream blockingDownload(URI uri)
+	public InputStream blockingDownload(final URI uri)
 	{
 		// Validate.
 		if (uri == null) {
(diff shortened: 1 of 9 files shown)


android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
```diff
@@ -50,19 +60,200 @@ public static TiDownloadManager getInstance()
...
+			}
+		} catch (Exception ex) {
+		}
+
+		// If running on main UI thread, then block this thread while doing download on another thread.
+		// Note: Using "HttpUrlConnection" on UI thread will cause a "NetworkOnMainThreadException" to be thrown.
+		if (TiApplication.isUIThread()) {
+			// Anonymous class used to call this blockingDownload() method on another thread.
+			// Downloaded content will be made available via Titanium's "TiResponseCache".
+			class DownloadOperation implements Runnable
```
garymathews: Instead of creating an anonymous class, you should be able to do this:
```Java
public InputStream blockingDownload(final URI uri)
{
...
	// Perform the blocking download.
	try {
		Thread downloadThread = new Thread(new Runnable() {
			@Override
			public void run() {
				try (InputStream stream = blockingDownload(uri)) {
					if (stream != null) {
						KrollStreamHelper.pump(stream, null);
					}
				} catch (Exception ex) {
				}
			}
		});
		downloadThread.start();

		// wait for thread to finish
		downloadThread.join();
	} catch (Exception ex) {
	}
	...
...
```

android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
```diff
@@ -50,19 +60,200 @@ public static TiDownloadManager getInstance()
...
+						String redirectUrl = httpConnection.getHeaderField("Location");
+						if ((redirectUrl == null) || (redirectUrl.length() <= 0)) {
+							break;
+						}
+						url = new URL(redirectUrl);
+
+						// Close the connection and attempt to connect to the redirect URL.
+						httpConnection.disconnect();
+						continue;
+					}
```
garymathews: Maybe log a warning if there's a bad response (404?), or throw an exception?

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
jquick-axway: Good point. The `image` property used to log a 404 warning and now it doesn't. I'll add some log messages.

Also, `backgroundImage` used to crash the app for a 404 response when "run-on-main-thread" was `false`. I've inadvertently fixed that. :)

garymathews (approved):
CR: **PASS**

PR #10233, opened 2018-08-05
[TIMOB-26268] iOS: Support UIRefreshControl for UIScrollView on iOS < 10

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26268

> `no tests`, since we already have tests for this.


vijaysingh-axway (approved):
CR passed.

PR #10238, opened 2018-08-07
[TIMOB-26165] iOS: Fix requiring a JSON file with single quotes

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26165

> A new file `json_files/amber.json` was [added](https://github.com/appcelerator/titanium-mobile-mocha-suite/commit/eb95f50a679167de95a632445b9f3cd10db7f202) to the `titanium-mocha-test-suite` repo.

### DIFF
--- a/tests/Resources/ti.require.addontest.js
+++ b/tests/Resources/ti.require.addontest.js
@@ -12,7 +12,7 @@ var should = require('./utilities/assertions');
 
 describe('require()', function () {
 	it('JSON-based require() with single-quotes', function () {
-		var amber = require('json_files/amber');
+		var amber = require('./json_files/amber');
 		should(amber).have.property('sdk');
 		should(amber.sdk).be.eql('7.4.0.v20180627024922');
 	});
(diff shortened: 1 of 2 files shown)


tests/Resources/ti.require.addontest.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe('require()', function () {
+	it('JSON-based require() with single-quotes', function () {
+		var amber = require('json_files/amber');
```
sgtcoolguy: Don't use a bad require style 😏 
```javascript
var amber = require('./json_files/amber');
```

sgtcoolguy (approved):
LGTM, but please avoid the ambiguous require path used in your unit test. It works, but we will eventually end support for this bad Titanium style require.

Also, you could have added the test input file into tests/Resources/jsonfiles/amber.json here in this PR and not pushed it to the suite. (Would make it easier to see together and would make it easier to cherry-pick/port to other branches).

tests/Resources/ti.require.addontest.js
hansemannn: Updating now. I was not sure if changes to the the `tests` repository will be picked up during CI build, so I placed it in the main repo instead. Good to know!

PR #10239, opened 2018-08-08
[TIMOB-25032] iOS: Resolve asset-catalog hashes back to real names, fail build when detecting duplicates

Sajoha:
JIRA: https://jira.appcelerator.org/browse/TIMOB-25032

Removed hashing from file names, first in the build step, so assets are stored under their real names, with checks in place to make sure duplicates are not used. And also in the module code to ensure de-hashing isn't attempted


iphone/Classes/FilesystemModule.m
```diff
@@ -196,20 +196,10 @@ - (id)getAsset:(id)args
...
-        NSMutableString *sha = [[NSMutableString alloc] init];
-        for (int i = 0; i < CC_SHA1_DIGEST_LENGTH; i++) {
-          [sha appendFormat:@"%02x", digest[i]];
-        }
-        [sha appendString:[newpath substringFromIndex:[newpath length] - 4]];
-        image = [UIImage imageNamed:sha];
-        RELEASE_TO_NIL(sha)
-      }
-    }
+    NSString *fileName = [imageArg lastPathComponent];
```
hansemannn: I think you can reuse the existing `imageArg` property:
```objc
imageArg = [imageArg lastPathComponent];
```
and align it to line 198 to be right after the other string manipulations. Same for other occurences of this code in the core.

iphone/cli/commands/_build.js
```diff
@@ -5706,15 +5706,34 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 				imageSets = {},
 				imageNameRegExp = /^(.*?)(@[23]x)?(~iphone|~ipad)?\.(png|jpg)$/;
 
+			const checked = new Set();
+			const duplicates = new Set();
+
+			Object.keys(imageAssets).forEach(function (assetName) {
```
hansemannn: Maybe some ES6 here?
```js
Object.keys(imageAssets).forEach(assetName => {
```

iphone/cli/commands/_build.js
```diff
@@ -5706,15 +5706,34 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
...
+
+				if (!checked.has(asset)) {
+					checked.add(asset);
+				} else if (!duplicates.has(asset)) {
+					duplicates.add(asset);
+				}
+			});
+
+			if (duplicates.size > 0) {
+				duplicates.forEach(function (asset) {
```
hansemannn: =>

iphone/cli/commands/_build.js
```diff
@@ -5706,15 +5706,34 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
...
+					this.logger.error(__('Image asset name "%s" used multiple time, duplicate names are not permitted', asset));
```
hansemannn:
```js
this.logger.error(__('Image asset name "%s" used multiple time, asset names should be unique in Titanium SDK 8.0.0+', asset));
```

hansemannn (changes requested):
Left a few changes. Also requesting Chris' review for the CLI part of this.

hansemannn (approved):

cb1kenobi (commented):
I think this ticket is a bad idea and that we should **NOT** merge it. Please see my comment on the ticket: https://jira.appcelerator.org/browse/TIMOB-25032.

vijaysingh-axway:
Closing in favor of PR #11437

PR #10110, opened 2018-06-14
[TIMOB-26130] Android: Remove flag that breaks NDK compatibility

garymathews:
- Remove `-flto` link-time optimization to preserve NDK 12+ compatibility

###### TEST CASE
- Set `ANDROID_NDK` environment variable to NDK 17
- Compile Titanium SDK successfully

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26130)

PR #10240, opened 2018-08-08
[7_3_X][TIMOB-26130] Android: Remove flag that breaks NDK compatibility

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26130

**Cherry-pick of:** #10110


hansemannn:
Is there a reason for not taking this into GA? It's pretty developer specific and the

garymathews (approved):
CR: **PASS**

PR #10245, opened 2018-08-10
[TIMOB-26284] Android: Cannot add child-views to views extending TiUIFragment

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26284

**Optional Description:**
Children added before the fragment transaction of the component are left under the fragment. These additions handle them and draws them once the fragment transaction has been committed. 

**Note:** Currently this takes care of only `add()` method. Dealing with positions for `insertAt()` will be added once we clear if it needs reworking.

Do we want any special treatment for TiUIFragment implementations?

**Test case:**
This requires a Ti.Map module built against a SDK with these changes merged.
_app.js_
```js
var Map = require('ti.map');
var mapview = Map.createView();
mapview.add(Ti.UI.createView({ width: 50, height: 50, backgroundColor: 'red'}));
win.add(mapview);
win.open();
```
I have to think a bit for a possible unit test.


android/titanium/src/java/org/appcelerator/titanium/view/TiUIFragment.java
```diff
@@ -44,12 +49,49 @@ public boolean dispatchTouchEvent(MotionEvent ev)
...
+		{
+			// Draw all the children that have been added prior the fragment transaction commit.
+			realizeFragmentViews();
+			fragmentCommitted = true;
+		}
+	};
+
+	public void realizeFragmentViews()
+	{
+		for (TiUIView tiUIView : childrenToRealize) {
```
garymathews: could we rename `tiUIView` into `child` ?

android/titanium/src/java/org/appcelerator/titanium/view/TiUIFragment.java
```diff
@@ -44,12 +49,49 @@ public boolean dispatchTouchEvent(MotionEvent ev)
...
+		// If the fragment transaction has been committed add the children the usual way
+		if (fragmentCommitted) {
+			super.add(child);
+		} else {
+			// If the fragment has not been added to the native view add the children in
+			// the array to be realized
+			childrenToRealize.add(child);
+		}
+	}
+
```
jquick-axway: @ypbnv, we also need to override the `insertAt(TiUIView, int)` and `remove(TiUIView)` methods. It's an unlikely edge-case, but it should be easy to add support, no?

For example, see our `TiUIScrollView` code here...
https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java#L1065

jquick-axway (changes requested):

android/titanium/src/java/org/appcelerator/titanium/view/TiUIFragment.java
ypbnv: Yes, I thought about `insertAt()`, but what I found out took the focus at that moment. I have described it in here: 
https://jira.appcelerator.org/browse/TIMOB-26283
Let me know what Gary and you think about that. If `insertAt()` needs a fix it will be better to do it for the source before overriding it here. 

I will override `remove()` , too.

ypbnv:
@jquick-axway I tried a few approaches about `insertAt()` and this update seems to not break anything else and consist of at least overriding as I was able to do. For `remove()` I was not able to time a test properly to break an application without the override, but the way it is now it should prevent drawing views that are not supposed to be there.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

`add` & `insertAt` methods works as expected & the child views are properly displayed.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2

ypbnv:
@jquick-axway , @garymathews I couldn't come up with a proper unit test for that. On the JS everything looks fine so I wasn't able to write a test that is failing before these changes. Do you have any ideas?

jquick-axway:
@ypbnv, perhaps we should keep it simple and do a unit test like the `ScrollView` "add-insert-remove" one shown below. It doesn't wait for the window to open. It just verifies that you can add/remove child views.
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.ui.scrollview.test.js#L165

PR #10249, opened 2018-08-13
[TIMOB-26209] Moved the Xcode EULA assertion from pre-validate to val…

cb1kenobi:
[TIMOB-26209] Moved the Xcode EULA assertion from pre-validate to validate after an Xcode version has been chosen for the build.


sgtcoolguy:
@cb1kenobi I assume this PR now needs to be updated to include ioslib 1.7.5...

sgtcoolguy:
Also that we'll need a back port for `7_3_X`

PR #10250, opened 2018-08-13
[TIMOB-26209] (7_3_X) Moved the Xcode EULA assertion from pre-validate to val…

cb1kenobi:
[TIMOB-26209] Moved the Xcode EULA assertion from pre-validate to validate after an Xcode version has been chosen for the build.


sgtcoolguy:
@cb1kenobi Something's clearly not right here:
```
/Users/build/Library/Application Support/Titanium/mobilesdk/osx/7.3.0.v20180814113856/iphone/cli/commands/_build.js:1979
				if (!this.xcodeEnv.eulaAccepted) {
                       ^
TypeError: Cannot read property 'eulaAccepted' of null
    at iOSBuilder.checkEULA (/Users/build/Library/Application Support/Titanium/mobilesdk/osx/7.3.0.v20180814113856/iphone/cli/commands/_build.js:1979:24)

```

cb1kenobi:
The only way that appears to be possible is when the build lets "ioslib suggest an iOS version".

Perhaps that block should move down the chain further? I'm investigating.

cb1kenobi:
OK, fixed. Turns out the scenario that causes the issue was removed in master (7.4.0).

PR #10251, opened 2018-08-13
[TIMOB-26295] Android: Listview with template causes a nullpointer on itemclick

javamanrj:
Adding a missing null check in itemProperties

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26295

Test case is on the issue.


garymathews (approved):
CR: **PASS**

PR #10254, opened 2018-08-14
[TIMOB-26300] iOS: Fix crash when using (deprecated) kroll-thread

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26300

> Removed some dirty old hacks that caused the internal iOS API to go mad when not using main-thread. I confirmed that the ticket linked to the initial code (from SDK 3.0.0) still works.


hansemannn:
@sgtcoolguy Something is odd here. For example, the `UserNotificationCenter` test fails, but it really cannot fail, since it simply checks for being an object, which it is always.

PR #10256, opened 2018-08-14
Updated Titanium.UI.TableViewRow

jawa9000:
Set selectedBackgroundColor to deprecated (as of 8.0.0) according to https://github.com/appcelerator/titanium_mobile/pull/10210. This was request in https://jira.appcelerator.org/browse/AC-5803.


hansemannn:
As per [TIMOB-26304](https://jira.appcelerator.org/browse/TIMOB-26304), we will aim for a code-based solution to un-deprecate the Android API. Sorry for the false warning here!

PR #10257, opened 2018-08-15
[TIMOB-26306] Android: Improved Titanium SDK (titanium_mobile) ant/ndk builds

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-26306
- https://jira.appcelerator.org/browse/TIMOB-25932

**Summary:**
- [TIMOB-26306](https://jira.appcelerator.org/browse/TIMOB-26306) Can now do fast parallel execution `-j8` NDK builds on Windows.
- [TIMOB-25932](https://jira.appcelerator.org/browse/TIMOB-25932) Can now build with JDK 10.
  * Removed "javah" tool usage. Tool was removed from JDK 10, which caused build failures.
  * Removed usage of "javah" generated header files. Not needed. Functions on C++ side are already exported via `JNIEXPORT`.
- Refactored Android make files.
  * Moved code generation parts to "ant" as a pre-build step. _(Fixes dependency order and file locking issues.)_
  * Removed files "common.mk" and "genSources.mk".
  * Added missing `APP_PLATFORM` setting. _(We were wrongly using `TARGET_PLATFORM` instead.)_
- Modified gperf and python code generators to create `*.h` files instead of `*.cpp` files.
  * Greatly simplifies make file configuration.
  * Dependency order between `*.cpp` no longer needs to be defined. _("KrollBindings.cpp" was `#include`ing these `*.cpp` files.)_
  * Simplifies `*.cpp` file listing in make file. _(No longer have to omit generated `*.cpp` files.)_
- Fixed majority of Android Java/C/C++ compiler warnings.
- Modified "javac" to ignore Java 1.7 bootstrap warning via "-Xlint:-options".
  * Only way to not ignore it is to reference Java 7 "rt.jar" which is not bundled with JDK 8 or higher.
  * Resolves `warning: [options] bootstrap class path not set in conjunction with -source 1.7`
- Modified NDK "clean" to delete the x86 intermediates/binaries.
  * Was missing `BUILD_X86` environment variable setting.
- Updated Titanium project files to use "min" API Level 16.
  * Note that not all files were updated when we last raised it. Particularly the modules and ndk side.
  * Fixes warnings with NDK build files.
- I've tested this PR with both Android NDK `r12b` and `r17b`.

**Test:**
Do the below on both Mac and Windows.
1. Install JDK 10 if not done already.
2. Open the Terminal or Command Prompt.
3. `cd` to the `titanium_mobile/build` directory.
4. Enter: `node scons.js cleanbuild`
5. Verify that the build finishes successfully.
6. Search the Terminal for "warning". Verify that no warnings appear.
7. Build and run a Titanium app for Android using the "7.4.0" version you just built.
8. Verify that the app launches successfully on Android.


android/build/common.xml
```diff
@@ -378,22 +382,68 @@ Common ant tasks and macros for building Android-based Titanium modules and proj
...
 	<target name="build.kroll.v8.so" depends="generate.kroll.v8.bindings,copy.kroll.v8.jars">
-		<condition property="ndk.build.args" value="" else="-j8">
-			<os family="windows"/>
-		</condition>
 		<property name="kroll.v8.build.x86" value=""/>
+
+		<!-- Generate a "KrollNativeBindings.h" file with perfect hashes via gperf tool. -->
+		<!-- Note: "in.file" string is inserted into file. Use relative path since absolute may contain user name. -->
+		<gperf
+			base.directory="${kroll.v8.project.dir}"
```
garymathews: Is this meant to be on the same line as
```XML
<gperf
```
?

garymathews (approved):
CR: **PASS**

android/build/common.xml
jquick-axway: I purposely put the attribute on the next line, lining it up with the rest of the wrapped attributes. We do this in other places in the xml file as well. Admittedly, it looks better with 4 spaced tabs.

jquick-axway:
@ypbnv, would you mind doing a `node scons.js build android` on your Windows machine with this PR please? It should make your builds much much faster.

I've already tested it on a Windows PC in the office here, but it's a super slow machine. It would probably be better to benchmark it off of your machine. Thanks.

m1ga:
@jquick-axway not had a time to test that but how much faster is it?

jquick-axway:
> not had time to test that but how much faster is it?

This mostly depends on the number of CPU cores your machine has. The more the better.

My old Windows 10 test machine (4 core CPU; solid state drive; active anti-virus scanner) is unusually slow, but here were my benchmarks when doing an Android build...
- Before: 477 seconds
- After: 218 seconds

So, it's about `2.2x` faster.
(I would hope a "better" Windows PC would be faster.)

If you're building from a Mac, then you can use the "ccache" feature (see link below) which was something that was already supported. It provides a huge performance gain.
https://wiki.appcelerator.org/display/eng/Speeding+up+Titanium+Mobile+Builds

m1ga:
:thumbsup: nice! I only can give you the before numbers for Linux:

cleanbuild:
1st (with all downloads): 2m49,858s
2nd (no downloads): 1m21,402s
(Dell XPS 15, Linux, i7-8750H, 12 core)

will test it soon and give you some feedback

ypbnv (approved):
**LGTM!**

I ran two builds - one with our current master and one from this PR. Both from scratch, nothing downloaded or prebuilt. The build times were as follow:
master:  10 minutes 47 seconds (I believe I had an overdue scheduled scan running at this point, but the lowest I got with recent builds was about 8 minutes.) 
this PR:  4 minutes 20 seconds
So the optimization for my machine will be around 2-2.5 times faster build times.

jquick-axway:
Thanks for testing it out @ypbnv. :)

PR #10260, opened 2018-08-16
[TIMOB-26299] : iOS Ti.Media.AudioPlayer test-suite crashes in some situations

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26299


hansemannn:
@vijaysingh-axway It does not seem to fix it so far, see the latest crash attached.

hansemannn (approved):
Looks good now! The approach is similar to what we do for iOS as well.

PR #10265, opened 2018-08-16
Update Android.yml

jawa9000:
Updated code sample in the Permissions chapter.

This is related to https://jira.appcelerator.org/browse/TIDOC-3222


jawa9000:
Unnecessary change.

PR #10266, opened 2018-08-17
[TIMOB-26313]iOS: Fix accessibility on Ti.UI.TableView

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26313

> `no tests`, since it's a UI-related change that is verified by the Appium suite


sgtcoolguy (approved):
FR passed with Appium desktop. Tested with a Ti.Ui.TableView having an accessibilityLabel value with a master SDK build and the name/accessibility id wasn't shown in the Appium inspector. Re-tried with the SDK built by this PR and it was. Simple as that.

PR #10267, opened 2018-08-20
[TIMOB-26319] iOS 10: Fix parity in "notificationsettings" event

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26319


vijaysingh-axway (approved):
CR passed.

PR #10269, opened 2018-08-20
[FRAME-11] Remove uses-sdk setting

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/FRAME-11

**Optional Description:**
Removes the now unnecessary `uses-sdk` setting. This was a leftover from when the template used the Material theme as a base instead of the AppCompat one.


jquick-axway (approved):
CR: Pass

Thanks @janvennemann :)

PR #10273, opened 2018-08-21
(7_3_X) [TIMOB-26325] Android: Ti.UI.WebView not firing events in 7.x

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26325

**Description:**
This reverts the boolean flag hack to prevent multiple `load`/`beforeLoad` events getting fired when changing the `userAgent` after construction.

When running through `processProperties` we set the user agent string before loading the html/url/data content to avoid immediate reload.

I think so long as you provide the user agent in the construction properties this should only ever fire the `load`/`beforeLoad` events once. If you construct a WebView with an Url and then change the user agent after (mid-load), it's entirely possible you'll get multiple `load`/`beforeLoad` events fired based on the timing of the code.

I included a quick blurb in the Ti.UI.WebView.userAgent apidocs about this behavior.

### DIFF
--- a/apidoc/Titanium/UI/WebView.yml
+++ b/apidoc/Titanium/UI/WebView.yml
@@ -635,6 +635,10 @@ properties:
     description: |
         On the iOS platform, this is not per-webview. Once you have set this property for a webview
         it will not change for same. But while creating new webview it can be changed to new user agent.
+
+        On Android, changing the user agent after the webview has begun loading content may cause
+        the webview to reload and fire multiple `load` or `beforeLoad` events. Developers should provide the
+        user agent value in the creation properties to avoid the reload and multiple events firing.
     type: String
     default: System default user-agent value.
     platforms: [android, iphone, ipad]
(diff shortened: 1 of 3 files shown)


garymathews (approved):
CR: **PASS**
FT: **PASS**

apidoc/Titanium/UI/WebView.yml
```diff
@@ -635,6 +635,10 @@ properties:
     description: |
         On the iOS platform, this is not per-webview. Once you have set this property for a webview
         it will not change for same. But while creating new webview it can be changed to new user agent.
+
+        On Android, changing the user agent after the webview has begun loading content may cause
+        the webview to reload and fire multiple `load` or `beforeLoad` events. Developers should provide the
```
hansemannn: It is [beforeload](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.WebView-event-beforeload).

sgtcoolguy: Thanks for the catch Hans!

sgtcoolguy:
Merged manually.

PR #10274, opened 2018-08-21
[TIMOB-25941] Android: Optimized HTTPClient download performance

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-25941

**Summary:**
- Download performance improved by about 25%. (Depends on network speed/quality.)
- Increased response buffer size from 4kb to 8kb.
- Removed undocumented event property `blob` from callback `ondatastream()`.
  * Improves performance by avoiding extra buffer copy.
  * This undocumented property does not exist on iOS.
- Modified `responseText` property handling only create its string when accessed.
  * Was creating and transcoding the string for every response segment received. Was a huge performance hit.
  * Fixed `responseText` to provide total text downloaded like iOS instead of the buffered response segment's text.
- Synchronized file write to storage upon download completion.
  * Avoids race condition where reading `responseText` or `responseData` in `onload()` callback may provide an incomplete response.

**Test:**
@lokeshchdhry, I recommend that you test this at home. The office network is too congested and download speed varies wildly.
1. Acquire an Android device with Internet access.
2. Build and run the below code with Titanium 7.3.0.
3. Tap the "Download" button. 
4. Wait for the download to complete.
5. Note down the download duration time. (You may want to do this a few times and record the average.)
6. Build and run with this PR and repeat steps 3-5.
7. The download time should have improved.

```javascript
var url = "https://github.com/appcelerator-modules/ti.admob/releases/download/ios-2.2.0/ti.admob-iphone-2.2.0.zip";
var downloadFilePath = Ti.Filesystem.applicationDataDirectory + "downloaded.zip";

var button = Ti.UI.createButton({ title: "Download" });
var progressBar = Ti.UI.createProgressBar({
	min: 0,
	max: 100,
	bottom: "40dp",
	left: "5dp",
	right: "5dp",
});
button.addEventListener("click", function(e) {
	var startTime = new Date();
	var httpClient = Ti.Network.createHTTPClient({
		ondatastream: function(e) {
			progressBar.value = e.progress * 100;
		},
		onload: function(e) {
			var duration =  (new Date()) - startTime;
			Ti.API.info("@@@ onload() duration: " + duration + "ms");
			alert("Download Time: " + duration + "ms");
			progressBar.value = 100;
			button.enabled = true;
		},
		onerror: function(e) {
			Ti.API.info("@@@ onerror()");
			alert("Download Failed");
			button.enabled = true;
		},
	});
	progressBar.value = 0;
	button.enabled = false;
	httpClient.open("GET", url);
	httpClient.setRequestHeader("Accept-Encoding", "identity");
	httpClient.file = downloadFilePath;
	httpClient.send();
});
window.add(button);
window.add(progressBar);
window.open();
```


android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -350,15 +350,10 @@ private void handleEntityData(byte[] data, int size, long totalSize, long conten
...
 		callbackData.put("totalCount", contentLength);
 		callbackData.put("totalSize", totalSize);
-		callbackData.put("size", size);
+		callbackData.put(TiC.PROPERTY_SIZE, size);
 
-		byte[] blobData = new byte[size];
-		System.arraycopy(data, 0, blobData, 0, size);
-
-		TiBlob blob = TiBlob.blobFromData(blobData, contentType);
-		callbackData.put("blob", blob);
```
sgtcoolguy: Technically this is a breaking Api change (and should go into 8.0.0) since this is removed. Although, it's hard to see how many people would be using this undocumented event callback property...

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
```diff
@@ -245,7 +246,7 @@ private void handleResponse(HttpURLConnection connection) throws IOException
 				Log.d(TAG, "Content length: " + contentLength, Log.DEBUG_MODE);
 				int count = 0;
 				long totalSize = 0;
-				byte[] buf = new byte[4096];
+				byte[] buf = new byte[8192];
```
sgtcoolguy: Extract the magic constant as a final static at top so it's easier to tweak the buffer size?

(in fact looking at this file it looks like we have a number of magic constants around buffers/sizes: 512 bytes for progress notifications/entity data buffer; 8MB for blob/file I/O buffer; 512Kb for dumping response to file first rather than keep in-memory)

jquick-axway: It's not worth changing to a constant if it's only used in 1 place in the code. It's better to keep it simple.

Yes, there are other place in this class where there is duplication, but that's a refactoring for another day.

android/modules/network/src/java/ti/modules/titanium/network/TiHTTPClient.java
jquick-axway: It is. But from searching our repo's and JIRA's history, it doesn't appear to be used. We've discussed it internally in the office and decided to remove it.

Also note that we have a customer that wants a solution to this performance issue before Titanium 8.0.0. You may want to talk to Eric about this.

PR #10290, opened 2018-08-28
[TIMOB-26339] (7_3_X) iOS: Fix "scheduleLocalNotification" API from not repeating triggers properly

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26339

> `no tests` because only testable via UI-tests.


vijaysingh-axway (approved):
CR passed.

PR #10296, opened 2018-08-30
[TIMOB-26344] iOS: Pass blocks with copy to retain them properly

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26344

> `no tests` because of the low-level change inside the application-delegate that requires user-interaction (in this case by triggering shortcut-items from the home screen).


vijaysingh-axway (approved):
CR passed

PR #10303, opened 2018-08-31
Fix for missing parameters of setContentOffset

jawa9000:
This PR fixes the missing parameters of setContentOffset. The parameter contentOffset shared a property with the same name. This was causing the rendered HTML to fail to display all the setContentOffset parameters.

https://jira.appcelerator.org/browse/TIDOC-3232


hansemannn:
@jawa9000 The [scroll-view docs](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.ScrollView) seems to work fine again now, even without this change. Or did you publish the changes from your local branch?

PR #10304, opened 2018-08-31
[TIMOB-13642] Android: Fixed bug where "relative" image paths won't load

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-13642

**Summary:**
- Relative image path was broken due to extra slash added to path, which Android asset manager could not handle.
  * Solved by trimming extra slashes.
  * Not an issue with external files since filesystem can handle extra slashes.
- Fixed resolution of paths containing `"../"` when loading an image from a JS file in a subdirectory.
- Added support for "content://" image URLs. (Used to only work via blobs.)
- Added support for "android.resource://" image URLs. (Used to only work via blobs.)
  * Allows `Ti.Filesystem.getResRawDirectory() + "MyImage"` to work.
  * Allows `"android.resource://" + Ti.App.id + "/drawable/MyImage"` to work.

**Note:**
Relative paths are currently handled differently between platforms.
- On Android, paths are relative to the JavaScript file.
- On iOS and Windows, paths are relative to the app's "Resources" directory.

I have not changed this behavior since that would be a breaking-change.

**Test:**
1. Create a "Classic App" project.
2. Replace the project's "app.js" with the code below. _(This code assumes the project has a "tab1.png".)_
3. Build and run the app on Android.
4. Scroll through the app, verifying that every black box has loaded an image.

```javascript
function createImageGroupFor(path) {
	var container = Ti.UI.createView({
		layout: "vertical",
		top: "5dp",
		left: "5dp",
		right: "5dp",
		backgroundColor: "#222222",
		width: Ti.UI.FILL,
		height: Ti.UI.SIZE,
	});
	container.add(Ti.UI.createLabel({
		text: path,
		color: "#CCCCCC",
		top: "5dp",
		left: "5dp",
		right: "5dp",
	}));
	container.add(Ti.UI.createImageView({
		image: path,
		preventDefaultImage: true,
		bottom: "5dp",
		width: "80dp",
		height: "80dp",
	}));
	return container;
}

	layout: "vertical",
	scrollType: "vertical",
	backgroundColor: "gray",
});
scrollView.add(createImageGroupFor("/assets/images/tab1.png"));
scrollView.add(createImageGroupFor("assets/images/tab1.png"));
scrollView.add(createImageGroupFor("./assets/images/tab1.png"));
scrollView.add(createImageGroupFor("././assets/images/.././images/tab1.png"));
scrollView.add(createImageGroupFor(Ti.Filesystem.getFile("assets/images/tab1.png").nativePath));
if (Ti.Platform.name === "android") {
	var externalFilePath = Ti.Filesystem.applicationDataDirectory + "copy.png";
	Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "assets/images/tab1.png").copy(externalFilePath);
	scrollView.add(createImageGroupFor(externalFilePath));
	externalFilePath = Ti.Filesystem.getFile(externalFilePath).nativePath;
	scrollView.add(createImageGroupFor(externalFilePath));
	externalFilePath = externalFilePath.substring("file:///".length);
	scrollView.add(createImageGroupFor("content://" + Ti.App.id + ".tifileprovider/filesystem/" + externalFilePath));
	scrollView.add(createImageGroupFor(Ti.Filesystem.resourcesDirectory + "assets/images/tab1.png"));
	scrollView.add(createImageGroupFor("file:///android_asset/Resources/assets/images/tab1.png"));
	scrollView.add(createImageGroupFor("android.resource://" + Ti.App.id + "/drawable/appicon"));
}
scrollView.add(Ti.UI.createView({ width: Ti.UI.FILL, height: "5dp" }));
window.add(scrollView);
window.open();
```


android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
```diff
@@ -122,6 +154,34 @@ private static String formPath(String path, String parts[])
...
+	 * Removes all characters matching the given "trimCharacter" from the front of given "sourceString"
+	 * and returns the result.
+	 * @param sourceString The string to have characters trimmed from. Can be empty or null.
+	 * @param trimCharacter The character to be trimmed from given "sourceString".
+	 * @return
+	 * Returns a new string if characters were trimmed from the front.
+	 * Returns the same string reference if no characters were trimmed.
+	 * Returns null if given a null string.
+	 */
+	private static String trimFront(String sourceString, char trimCharacter)
```
garymathews: I was able to simplify this a little:
```Java
private static String trimFront(String sourceString, char trimCharacter)
{
	if (sourceString != null) {
		for (int i = 0; i < sourceString.length(); i++) {
			if (sourceString.charAt(i) != trimCharacter) {
				sourceString = sourceString.substring(i);
				break;
			}
		}
	}
	return sourceString;
}
```

android/titanium/src/java/org/appcelerator/titanium/util/TiUrl.java
```diff
@@ -273,21 +273,21 @@ public static String absoluteUrl(String defaultScheme, String url, String baseUr
...
-					} else if (baseUrl.length() == 1) {
-						combined = url;
-					} else if (url.length() == 1) {
-						combined = baseUrl;
+				} else if (url.startsWith(PATH_SEPARATOR)) {
+					// The URL is an absolute path. Not relative to base URL.
+					// Note: A "file:///" URL needs 3 slashes to reference localhost.
+					if (defaultScheme != null) {
+						combined = defaultScheme + PATH_SEPARATOR;
+						if (defaultScheme.equals("file:")) {
```
garymathews: Could we use `TiFileFactory.FILE_URL_SCHEME` here? Or even `FILE_URL_PREFIX`

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
jquick-axway: This doesn't handle the case where all characters in the string are trimmable. That's why I do the substring handling outside of the loop.

android/titanium/src/java/org/appcelerator/titanium/util/TiUrl.java
jquick-axway: Yeah, we have several "file:" string literals in the "TiURL.java" code.
I also don't like how "TiURL.java" expects a scheme string to end with a trailing colon ':' either. That's not part of the scheme. That's the character which separates it from the scheme from the path.

I didn't feel like mucking with this part and left it alone.

android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
garymathews: I think that a `for` is a little easier to read than a `while` in this case:
```Java
private static String trimFront(String sourceString, char trimCharacter)
{
  if (sourceString != null) {
    int i = 0;
    for (; i < sourceString.length(); i++) {
      if (sourceString.charAt(i) != trimCharacter) {
        sourceString = sourceString.substring(i);
        break;
      }
    }
    if (i == sourceString.length()) {
      sourceString = "";
    }
  }
  return sourceString;
}
```

But they both work, so I'll let you decide.

jquick-axway: I originally wrote it with a `for` like this...
```java
for (; (index < sourceString.length()) && (sourceString.charAt(index) == trimCharacter); index++);
```

But our lint step didn't like it. _(shrugs)_

lokeshchdhry:
FR Passed.

`Relative path` images load successfully.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ samsung SM-G955U1 (Android 8.0.0)

lokeshchdhry:
Waiting for FR to pass.

garymathews (approved):
CR: **PASS**

PR #10305, opened 2018-09-01
[TIMOB-26349] iOS: Fix a couple of long-running issues with push notifications

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26349

> `no tests`, because a quite complex setup of certificates, provisioning profiles, plist-entries, entitlements and an actual hardware device are required to test this.


vijaysingh-axway (approved):
CR passed.

PR #10307, opened 2018-09-04
[TIMOB-26261] Android: Compile with API Level 28 (Android P)

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26261

**Summary:**
- Modified Titanium to compile against API Level 28 (Android P).
- Does not target API Level 28 by default yet.
  * To be done once all Android P breaking changes have been resolved.
  * App dev can opt-in to targeting API Level 28 via "tiapp.xml" `<uses-sdk/>` setting.
- Updated Android build-tools max version from `27.x` to `28.x`.
- Updated Android NDK max version from `r16c` to `r17b`.

**Test:**
Build and run KitchenSinkV2 on Android 9.0 and Android 4.1. Traverse app's UI features and verify that they work.

### DIFF
--- a/android/.idea/misc.xml
+++ b/android/.idea/misc.xml
@@ -24,7 +24,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="false" project-jdk-name="Android API 27 Platform" project-jdk-type="Android SDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="false" project-jdk-name="Android API 28 Platform" project-jdk-type="Android SDK">
     <output url="file://$PROJECT_DIR$/out" />
   </component>
 </project>
--- a/android/.idea/workspace.xml
+++ b/android/.idea/workspace.xml
@@ -675,7 +675,7 @@
         </state>
       </provider>
     </entry>
-    <entry file="file://$USER_HOME$/Library/Android/sdk/sources/android-27/android/app/Activity.java">
+    <entry file="file://$USER_HOME$/Library/Android/sdk/sources/android-28/android/app/Activity.java">
       <provider selected="true" editor-type-id="text-editor">
         <state relative-caret-position="207">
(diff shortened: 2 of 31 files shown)


sgtcoolguy:
So, technically speaking bumping the min node version is a "breaking change" and would need to go in a major release (like 8.0.0).

However, I think we kind of already have unofficially bumped the node minimum required in our toolset anyhow? I know Chris Barber wants it to be like Node 8.11 which is quite new, for `appcd`.

But I'm not sure who's keeping track of our broad-based Node min supported versions - it'd be good to ensure they're aligned between the SDK and all our 1st-party npm packages that we use in the CLI/SDK.

@mukherjee2 @eric34 Do we have an official Node minimum we're shooting for across the board? If we look at https://github.com/nodejs/Release I think we should be trying to support 6.x - 10.x right now.

garymathews (approved):
CR: **PASS**

jquick-axway:
@sgtcoolguy, right, I was told verbally that our min was 8.0. And QE has only been testing with 8.x as well. So, I got the impression that someone forgot to bump up the node versions in our "package.json".

We document the min as 8.0 here...
https://docs.appcelerator.com/platform/latest/#!/guide/Titanium_Compatibility_Matrix-section-src-29004837_TitaniumCompatibilityMatrix-Node.js

But we also document it as 6.0 in the same doc below. Although from a customer perspective, I think this would be confusing and I think we should only show one min version (keep the 8.0 one).
https://docs.appcelerator.com/platform/latest/#!/guide/Titanium_Compatibility_Matrix-section-src-29004837_TitaniumCompatibilityMatrix-Node.js.1

How about this. I'll change the min version back to what it was until we have consensus from the rest of the team.

lokeshchdhry:
FR Passed.

Ran KS2 on API 28 & API 16 & it seems to work fine.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
Emulator:       Android API28, Android API16

PR #10312, opened 2018-09-07
TIMOB-26369 Restored notification of unhandled exceptions in iOS

rlustemberg:
Restored notification of unhandled exceptions in iOS

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26369


hansemannn:
The change looks good! We should add a unit-test for it, but our team can help getting that in shape. Thanks @rlustemberg !

PR #10313, opened 2018-09-07
[TIMOB-26371] Ensure buildAssetsDir exists before writing app props file

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26371

Due to the parallelisation of this processes the assets dir doesn't already exist. Just make sure it exists before we write it.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -5816,6 +5816,8 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			const contents = JSON.stringify(props);
 			if (!fs.existsSync(appPropsFile) || contents !== fs.readFileSync(appPropsFile).toString()) {
 				this.logger.debug(__('Writing %s', appPropsFile.cyan));
+
+				fs.existsSync(this.buildAssetsDir) || wrench.mkdirSyncRecursive(this.buildAssetsDir);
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -5819,6 +5819,8 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			const contents = JSON.stringify(props);
 			if (!fs.existsSync(appPropsFile) || contents !== fs.readFileSync(appPropsFile).toString()) {
 				this.logger.debug(__('Writing %s', appPropsFile.cyan));
+
+				fs.existsSync(this.buildAssetsDir) || wrench.mkdirSyncRecursive(this.buildAssetsDir);
```
cb1kenobi: Shouldn't you only do this if encryption is enabled?

```js
if (this.encryptJS && fs.existsSync(this.buildAssetsDir)) {
    wrench.mkdirSyncRecursive(this.buildAssetsDir);
}
```

ewanharris: Good point, I had missed that we're not always working with buildAssetsDir, and looking at the flow `xcodeAppDir` will exist before this so I'll edit to only do this when we're encrypting

cb1kenobi (approved):

PR #10315, opened 2018-09-08
[TIMOB-26372] Fire change event only after actual text change

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26372

**Optional Description:**
The text of the internal UITextView is not updated yet in shouldChangeTextInRange so calling noteValueChange here can cause unwanted side effects. See the ticket for an example of such side effects.


hansemannn:
@janvennemann Please add a unit test for this change.

janvennemann:
@hansemannn, this requires user input to reproduce the test case from the ticket. Programmatically firing the change event won't trigger the wrong behavior so unit tests are not suited to verify this change. Do we have an UI testing suite for such cases?

hansemannn (approved):

PR #10317, opened 2018-09-11
[TIMOB-26378] Android: backgroundDisabledColor is ignored

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26378

**Description:**
Fix the check for having a color state background for a view.

Note: There is a unit test for that, but needs some attention apparently.

**Test case:**
_app.js_
```js
    layout: 'vertical'}),
view = Ti.UI.createView({
    top: 50,
    width: 100,
    height: 100,
    backgroundColor: 'red',
    backgroundDisabledColor: 'gray'}),
button = Ti.UI.createButton({
    title: 'toggle'
});
view.addEventListener('click', function () {
    Ti.API.info('Click!');
});
    view.enabled = !view.enabled;
});
win.add([view, button])
win.open();
```


garymathews (approved):
CR: **PASS**

NOTE: Is this still `work in progress` ?

ypbnv:
@garymathews Yes, it was. I was trying to get the unit test working. I did it, but with a bit more code that anticipated. Let me know what you think about it. The thing that bothers me the most is the trend to limit usage of non SDK APIs in Android P ( supposing that's the way Android is taking ) which may cause the reflection to fail at some point. I have ran the tests on a Pixel with Android P and it is working fine, but that may not be the case in the future.
I have not put the code for the disabled state in TiUiHelper for now. If we decide to keep it I will move it there and add getters for DEFAULT, FOCUSED and SELECTED states as well. 
That opens the question if we can do something similar for images for different states? And if we can have the same on iOS?

CC: @jquick-axway, @hansemannn

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -1166,6 +1172,46 @@ public TiViewProxy getParent()
...
+		if (backgroundDrawable == null) {
+			return null;
+		} else {
+			try {
+				// Get the backgroundDrawable background as a StateListDrawable.
+				StateListDrawable stateListDrawable = ((StateListDrawable) backgroundDrawable.getBackground());
+				// Get the reflection methods.
+				Method getStateDrawableIndexMethod =
+					StateListDrawable.class.getMethod("getStateDrawableIndex", int[].class);
+				Method getStateDrawableMethod = StateListDrawable.class.getMethod("getStateDrawable", int.class);
```
garymathews: I feel we should be a little more graceful if the methods do not exist instead of printing a stack trace

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -1166,6 +1172,46 @@ public TiViewProxy getParent()
...
+				// Get the disabled state's (as defined in TiUIHelper) index.
+				int index =
+					(int) getStateDrawableIndexMethod.invoke(stateListDrawable, TiUIHelper.BACKGROUND_DISABLED_STATE);
+				// Get the drawable at the index.
+				Drawable drawable = (Drawable) getStateDrawableMethod.invoke(stateListDrawable, index);
+				// Parse it as a ColorDrawable.
+				ColorDrawable colorDrawable = ((ColorDrawable) ((LayerDrawable) drawable).getDrawable(0));
```
garymathews: Validate we can do this cast?

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -1166,6 +1172,46 @@ public TiViewProxy getParent()
...
+				// Transcript the color int to HexString.
+				String strColor = String.format("#%06X", 0xFFFFFF & colorDrawable.getColor());
```
garymathews: Could this return a color with an alpha channel?

garymathews (changes requested):

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
ypbnv: I have missed to check that. It can, so I changed the resulting String format.

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
ypbnv: Maybe print Warnings instead of the stack trace?

garymathews (approved):
CR: **PASS**

ypbnv:
If we are OK with this, I think doing the same for the rest of the states of the color background is a good next step. What do you think ?

longton95 (approved):
**FR Passed**

`backgroundDisabledColor` is no longer ignored, The view now turns grey when disabled.

**ENV**

```
Pixel 2 XL (9.0)
Emulator (8.1, 7.1)
Appc NPM: 4.2.13
Appc CLI: 7.0.6
Ti CLI: 5.1.1
Node: 8.9.4
NPM: 6.3.0
```

PR #10320, opened 2018-09-12
[TIMOB-26374] Android: Ti.UI.Button can't be disabled until displayed as of 6.1.0

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26374

**Description:**
Link `clickable` native property instead of `enabled` with the Titanium `touchEnabled`.
The result from the test case makes sense to me - if a button has `touchEnabled` set to true and is disabled it consumes the click event and prevents it from bubbling to the parent. But if we set the `touchEnabled` to false it always propagates to the parent no matter button's state.

**WARNING:** This change exposes a problem - if you have a button which is enabled, but has `touchEnabled` set to false it triggers the ripple effect if you click on its parent. I am still investigating why is that happening. I will open a separate ticket for it if we decide to go for this fix in its current form, otherwise the mentioned bug may not be relevant.

Note: I don't think we can have unit tests for that since it requires user interaction.

**Test case:**
_app.js_
```js
  layout: 'vertical'
});
win.addEventListener('click', function () {
    alert('Touch from Window!');
});
var btnEnabledHasTouch = Ti.UI.createButton({
    title: 'Enabled Touch',
    enabled: true,
    touchEnabled: true,
});
var btnDisabledHasTouch = Ti.UI.createButton({
    title: 'Disabled Touch',
    enabled: false,
    touchEnabled: true
});
var btnEnabledNoTouch = Ti.UI.createButton({
    title: 'Enabled No Touch',
    enabled: true,
    touchEnabled: false
});
var btnDisabledNoTouch = Ti.UI.createButton({
    title: 'Disabled No Touch',
    enabled: false,
    touchEnabled: false
});
btnEnabledHasTouch.addEventListener('click', function () {
    alert('Enabled Has Touch!');
});
btnDisabledHasTouch.addEventListener('click', function () {
    alert('Disabled Has Touch!');
});
btnEnabledNoTouch.addEventListener('click', function () {
    alert('Enabled No Touch!');
});
btnDisabledHasTouch.addEventListener('click', function () {
    alert('Disabled No Touch!');
});
win.add([btnEnabledHasTouch, btnDisabledHasTouch, btnEnabledNoTouch, btnDisabledNoTouch]);
win.add(btnEnabledHasTouch);
win.open();
```


jquick-axway:
**Question:**
I know that setting the "enabled" property dynamically after the window has been open works. I've enabled/disabled the button in my download test code in PR #10274 . I disable the button while it's in the middle of the button. So, why is this only an issue upon creation?
(Admittedly, the code is difficult to follow here.)

ypbnv:
Here is what I found out:
On handling the `getOrCreateView()` after the views are realized we call `registerForTouch()` on the native view:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java#L587
It always calls `registerForTouch` with the view as a parameter here: 
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java#L1580
And that gets us to here:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java#L1773
Since the properrty `PROPERTY_TOUCH_ENABLED` is `true` by default it overrides the the value of the `PROPERTY_ENABLED` set during `processProperties()` here:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java#L1054

jquick-axway:
Thanks for stepping me through it.
So, it's going to call `View.setEnable()` in the `registerTouchEvents()`. Sounds good.

jquick-axway (approved):
CR: Pass

PR #10321, opened 2018-09-13
[TIMOB-26388] iOS 12: support iPhone XS Max and iPhone XR launch images

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26388


Topener:
I tested "Default-Landscape-2688h@3x.png" and I can compress it with 60%. I guess that needs to be done before this should be merged

hansemannn:
@Topener Already done, for Alloy as well. Pushed now.

Topener:
Awesome! Love the quick work 👍

PR #10322, opened 2018-09-13
[TIMOB-26388] (7_4_X) iOS 12: support iPhone XS Max and iPhone XR launch images

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26388


iphone/Classes/TiUtils.h
```diff
@@ -662,11 +662,23 @@ typedef enum {
 + (BOOL)isRetinaiPhone6Plus;
 
 /**
- Whether or not the current device has a 5.8 inch retina display (iPhone X).
- @return _YES_ if the current device has a 5.8 inch retina display, _NO_ otherwise.
+ Whether or not the current device has a 5.8 inch super retina display (iPhone X).
+ @return _YES_ if the current device has a 5.8 inch super retina display, _NO_ otherwise.
  */
 + (BOOL)isRetinaiPhoneX;
```
sgtcoolguy: Should this be renamed to "isRetinaiPhoneXorXS"? something along those lines? Seems the Xs and X share the same resolution/density: https://www.cnet.com/news/iphone-xs-specs-vs-iphone-xs-max-xr-and-x-whats-new-and-better/

iphone/Classes/TiUtils.h
```diff
@@ -662,11 +662,23 @@ typedef enum {
...
 
+/**
+ Whether or not the current device has a 6.5 inch super retina display (iPhone X Max).
```
sgtcoolguy: `iPhone X Max` -> `iPhone XS Max`

iphone/Classes/TiUtils.h
```diff
@@ -662,11 +662,23 @@ typedef enum {
...
+ @return _YES_ if the current device has a 6.5 inch super retina display, _NO_ otherwise.
+ */
++ (BOOL)isRetinaiPhoneXSMax;
+
+/**
+ Whether or not the current device has a 6.1 inch retina display (iPhone X Max).
```
sgtcoolguy: `iPhone X Max` -> `iPhone XR`
May want to consider adding the actual resolution/density values here in the comments as well?

iphone/Classes/TiUtils.m
```diff
@@ -112,6 +112,18 @@ + (BOOL)isRetinaiPhoneX
   return (mainScreenBoundsSize.height == 812 || mainScreenBoundsSize.width == 812);
 }
 
++ (BOOL)isRetinaiPhoneXSMax
+{
+  CGSize mainScreenBoundsSize = [[UIScreen mainScreen] bounds].size;
+  return (mainScreenBoundsSize.height == 896 || mainScreenBoundsSize.width == 896) && [TiUtils isRetinaHDDisplay];
```
sgtcoolguy: The naming of isRetinaHDDisplay is confusing here for me (and the next method). Presumably `[TiUtils isRetinaHDDisplay]` is equivalent to asking if it's a 3x density (458pp)?

Maybe rename TiUtils.isRetinaHDDisplay to something like `TiUtils.is3xDensity` or `TiUtils.is458ppDensity`? And then below (iPhone XR) checking that we're at 2x density rather than checking we're *not* at 3x?

sgtcoolguy (commented):
I think it'd be helpful to clarify the comments better here where you seem to have copy-pasted the phone name to multiple methods.

Even more important/useful I think would to be consider renaming some of the utility methods to better reflect what they're checking. It's not super obvious to me that `isRetinaHDDisplay` means asking if we're at 3x density/scale.

(And on an even wonkier note: I think our dpi calculations in TiUtils are technically off and should reflect the values from here: https://en.wikipedia.org/wiki/Retina_display

Instead we have some relatively simple logic to use one of: 130, 160, 260, 320, 480
But I assume this may have implications on our layout engine for conversions to inches.)

iphone/Classes/TiUtils.m
hansemannn: +1 for the 3xDensity

iphone/Classes/TiUtils.h
hansemannn: We used to use the same for the `*S` version (like for the iPhone 4 and 6), but we can change this, sure. The only concern I'm having is that some people may use this public method in their native modules, which would break now.

iphone/Classes/TiUtils.m
sgtcoolguy: Can we add the new name and have the old redirect to the new name and deprecate (the old)?

hansemannn: Yep!

hansemannn:
@sgtcoolguy I updated the PR to use an inch-based approach. 

#### New API's
- [x] `is2xRetina`
- [x] `is3xRetina`
- [x] `isRetina4_7Inch`
- [x] `isRetina5_5Inch`
- [x] `isRetina5_8Inch`
- [x] `isRetina6_1Inch`

#### Deprecated API's
- [x] `isRetinaDisplay` -> ``
- [x] `isRetinaHDDisplay`
- [x] `isRetinaiPhone6`
- [x] `isRetinaiPhone6Plus`
- [x] `isRetinaiPhoneX`

PR #10323, opened 2018-09-14
[TIMOB-26389] (7_4_X): iOS Application throws an error when building with a blank app

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26389


hansemannn:
Closing since this will only be in 7.5.0.

PR #10324, opened 2018-09-14
[TIMOB-26389] : [iOS] Application throws an error when building with a blank app

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26389


hansemannn (approved):
CR looks good, PR **approved**!

PR #10331, opened 2018-09-20
[TIMOB-26402] Android: Maintain backwards compatibility with Titanium 7.X.X native modules

garymathews:
#### REQUIRES https://github.com/appcelerator/v8_titanium/pull/38

- Update V8 to `7.0.276.13`
- Maintain backwards compatibility with existing Titanium 7.X.X native modules

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26402)


sgtcoolguy:
@garymathews This needs to be done on the `next` branch, not `master`. You should be able to just rebase your single commit on top of `next`. I tried changing the base here, but it resulted in a ton of extra commits being added to the PR.

PR #10333, opened 2018-09-21
[7_4_X][TIMOB-26404] Update to hyperloop 3.1.3

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26404


ewanharris:
Looks like the tests are failing as they're testing iOS 12 only APIs and they ran on iOS 11.4. It looks like we tie to xcode-10 though? Maybe we need to allow specifying an iOS version in the mocha suite?

PR #10338, opened 2018-09-24
[TIMOB-26030] Android: Fixed TCP read()/write() to not crash when JS running on main thread

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26030

**Note:**
Unit test already exists for this in mocha suite. It's been failing on Android the whole time.

**Test:**
1. Open the project's "tiapp.xml" file.
2. Make sure property "run-on-main-thread" is set to `true`.
3. Set up an Android device with Internet access.
4. Build and run `TcpClientBlockingTest.js` attached to [TIMOB-26030](https://jira.appcelerator.org/browse/TIMOB-26030).
5. On app startup, verify that an HTTP response is shown onscreen.
6. Back out of the app.
7. Disconnect the device from the Internet.
8. Restart the app.
9. Verify that you see an "Unable to connect" message in the JSON output onscreen.
10. Open "tiapp.xml" file and change property "run-on-main-thread" to `false`.
11. Repeat steps 3-9.

### DIFF
--- a/android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
+++ b/android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
@@ -7,7 +7,6 @@
 package ti.modules.titanium.network.socket;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
@@ -19,6 +18,7 @@ import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.kroll.common.Log;
 import org.appcelerator.titanium.io.TiStream;
+import org.appcelerator.titanium.TiApplication;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.util.TiStreamHelper;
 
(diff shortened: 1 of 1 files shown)


android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
```diff
@@ -422,22 +421,51 @@ public int read(Object args[]) throws IOException
 			throw new IllegalArgumentException("Invalid number of arguments");
 		}
 
-		if (inputStream == null) {
```
sgtcoolguy: I assume we had this null guard for a good reason? We do retain the input stream as a field here, so I think we should still be re-using/assigning to it

sgtcoolguy (approved):
I think we need to retain the assignment/re-use of the inputStream field, but otherwise LGTM.

This is approximately 1000x easier/more elegant than what I did. I do think however, that we should consider deprecating the sync APIs (maybe?) or at the very least extending them to accept a callback function to operate async (it is very, very odd to me that we have an async variant on some separate utility proxy in a different namespace). But, that should be done in another PR. If this fixes the crash/blocking then it is a good fix that should be merged ASAP.

lokeshchdhry:
FR Passed. 

`read()` and `write()` methods do not cause a crash & run as expected when `run-on-main-thread = true`.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       1.8.0_161

PR #10340, opened 2018-09-25
Remove notice of deprecation

Topener:
I removed notice of deprecation as the mentioned property is removed, not deprecated. Since it is too long ago already there should not have to be a mention anymore imho.

The other way around there still is a mention.


jawa9000 (commented):
Looks good!

PR #10344, opened 2018-09-25
added style property & code sample

Topener:
Added the missing style property and added a small code sample to demonstrate how it should work.


jawa9000 (commented):
Looks good!

apidoc/Titanium/UI/iOS/FeedbackGenerator.yml
```diff
@@ -44,6 +54,13 @@ properties:
     constants: Titanium.UI.iOS.FEEDBACK_GENERATOR_TYPE_*
     availability: creation
     optional: false
+  - name: style
+    type: Number
+    summary: The style of the feedback generator you want to create. This property is only required when using <Titanium.UI.iOS.FEEDBACK_GENERATOR_TYPE_IMPACT>.
+    constants: Titanium.UI.iOS.FEEDBACK_GENERATOR_STYLE_*
```
sgtcoolguy: I think this should be `FEEDBACK_GENERATOR_IMPACT_STYLE_*`, which is why the apidoc validator is complaining.

Topener: Yep indeed! Let me fix that

PR #10346, opened 2018-09-25
Update View.yml

lokeshchdhry:
Removed old IOS 2.0 default unit behaviour description.


jawa9000 (approved):
Simple text change. Looks good to me.

PR #10348, opened 2018-09-25
Update View.yml

lokeshchdhry:
1. Removed 2.0 reference in `auto` property in `Size & Position` section.
2. Removed `\n` from `backgroundDisabledImage` property.
3. Removed `\n` from `backgroundFocusedImage` property.
4. Removed `\n` from `backgroundImage` property.
5. Removed `\n` from `backgroundSelectedImage` property.


jawa9000 (changes requested):
While the changes look good, I discovered several issues with the use of the `default` parameter. I created https://github.com/appcelerator/titanium_mobile/pull/10348/files#diff-0cc7dd4ed3fbabcc1ea4d811f65da9dc in response.

PR #10351, opened 2018-09-26
[TIMOB-26413] (7_4_X) iOS: Use correct dpi-scaling for launch-screens until Utils are refactored

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26413


vijaysingh-axway (approved):
CR passed.

PR #10352, opened 2018-09-26
[TIMOB-26391] Protect js objects for proxies

janvennemann:

**Optional Description:**
This will protect JS objects temporarily until they are passed back into the js context so they don't get CG'd during the small timeframe of their creation and until they are actually referenced in the JS object graph.

### DIFF
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -668,6 +668,8 @@ CFMutableSetRef krollBridgeRegistry = nil;
   ourKrollObject = [[KrollCoverageObject alloc] initWithTarget:proxy context:context];
 #else
   ourKrollObject = [[KrollObject alloc] initWithTarget:proxy context:context];
+#endif
+#ifdef USE_JSCORE_FRAMEWORK
   [ourKrollObject protectJsobject];
 #endif
 
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -668,6 +668,7 @@ CFMutableSetRef krollBridgeRegistry = nil;
   ourKrollObject = [[KrollCoverageObject alloc] initWithTarget:proxy context:context];
 #else
   ourKrollObject = [[KrollObject alloc] initWithTarget:proxy context:context];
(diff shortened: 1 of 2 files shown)


hansemannn (approved):
Working good on my site! Approved if @sgtcoolguy agrees!

sgtcoolguy (approved):
Generally LGTM. 

I'm not 100% sure these are the exact right positions to be protecting/unprotecting the values. I was looking at HAL to see where it was done on Windows, and it's basically done right in the constructor for objects created on the native side, and unprotected in the destructor (Of the JSObject c++ type which wraps the JSObjectRef c type).

I think if we were to translate that sort of policy over, then it'd probably be done in the KrollObject initWithTarget:, and unprotected in dealloc? (Though there appear to be other places we call TiObjectMake to generate js objects, but I'm not sure if any of those are kept around long-term as fields - they look to be returned as values from a method or set as properties on some other object)

sgtcoolguy:
Just for reference, since we're dealing with the exact same JS engine in both cases, but in pretty different ways - but the concepts remain applicable to both.

Generally speaking, the flow for Windows/HAL is:

- We decide to new up a proxy, which calls...
- [JSObjectCallAsConstructorCallback](https://github.com/appcelerator/HAL/blob/master/include/HAL/detail/JSExportClass.hpp#L747) - This news up a JSObject that conforms to the proxy's JS type we've defined, which we do in...
- [JSContext::CreateObject](https://github.com/appcelerator/HAL/blob/master/src/JSContext.cpp#L103) - Here's where we new up the object with the class for a given context, which calls the constructor...
- [JSObject::JSObject](https://github.com/appcelerator/HAL/blob/master/src/JSObject.cpp#L261) - the constructor of creating a new JS object, at the end we "register" it with the context it was created in for doing lookups...
- [JSObject::RegisterJSContext](https://github.com/appcelerator/HAL/blob/master/src/JSObject.cpp#L326) - here's where we keep a map of contexts to objects. If this is a new object, we then protect the js object until the destructor is called on the C++ side.

alitele:
Please give permission to me to download SDK .

PR #10358, opened 2018-09-28
[TIMOB-26354] Android: Refactor TabGroup and introduce new style

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26354

**Description:**
Refactor the implementation of TabGroup - remove deprecated ActioBar tabs UI and replace it with TabLayout. Introduce a new style for using BottomNavigationView. A more detailed description will be available once I am done with everything. 

Currently still trying to fix the removal of tabs before opening the TabGroup by simplifying the relation between the tab proxies added and their TiUiAbstractTab counterparts.

Update: 
How things work after the refactoring:
 - all the code related to the deprecated ActionBar Tabs UI has been removed - that way all the problems with themes originating from ActionBar are no longer present. For instance now TabGroup can be used with a Theme.AppCompat.NoTitleBar and children of it without issues;
 - handling of the ViewPager that is used to show the Tabs is moved to `TiUIAbstractTabGroup`;
 - `TiUIAbstractTabGroup` provides an interface for communication with different "controllers";
 - this allows us to have the separate styles as different implementations making it easier to support and introduce new styles;
 - backwards compatibility with the JS code for the TabGroup implementation before this refactoring. Not defining a style at creation falls back to using the DEFAULT style which is as close as possible the ActionBar UI, but following the Material Design Guidelines ( one difference is the position of the icon relative to the title );
 - fixes the behavior of some properties ( tabsBackgroundColor, tabsFocusedBackgroundColor ), adds some new properties ( style, shiftMode ) and brings parity with iOS for other ( titleColor, activeTitleColor );

Note: The constants for style will be used from the same namespace for the `TabbedBar` component. I will change them in the PR for it once this gets the green light for merging.

Tickets that should be resolved by this refactoring:
 - https://jira.appcelerator.org/browse/TIMOB-24961
 - https://jira.appcelerator.org/browse/TIMOB-19695
 - https://jira.appcelerator.org/browse/TIMOB-15496
 - https://jira.appcelerator.org/browse/TIMOB-23271
 - https://jira.appcelerator.org/browse/TIMOB-25200
 - https://jira.appcelerator.org/browse/TIMOB-17931
 - https://jira.appcelerator.org/browse/TIMOB-16900 ( this one was already fixed, listing it here as the same behavior is kept )

**Test case:**
_app.js_
```js
var mainWindow = Ti.UI.createWindow({
		layout: 'vertical'
	}),
	buttonDefault = Ti.UI.createButton({
		title: 'Default'
	}),
	buttonBottom = Ti.UI.createButton({
		title: 'Bottom'
	}),
	winDefaultA = Ti.UI.createWindow({
		backgroundColor: 'red'
	}),
	winDefaultB = Ti.UI.createWindow({
		backgroundColor: 'green'
	}),
	winDefaultC = Ti.UI.createWindow({
		backgroundColor: 'blue'
	}),
	tabDefaultA = Ti.UI.createTab({
		title: 'Tab A',
		window: winDefaultA
	}),
	tabDefaultB = Ti.UI.createTab({
		title: 'Tab B',
		window: winDefaultB
	}),
	tabDefaultC = Ti.UI.createTab({
		title: 'Tab C',
		window: winDefaultC,
		backgroundColor: 'cyan',
		backgroundFocusedColor: 'magenta',
		titleColor: 'red',
		activeTitleColor: 'blue'
	}),
	tabGroupDefault = Ti.UI.createTabGroup({
		// for backwards compatibility if no style is defined, the component falls back to the default one
		style: Ti.UI.Android.TABS_STYLE_DEFAULT,
		tabs: [tabDefaultA, tabDefaultB, tabDefaultC],
		tabsBackgroundColor: '#88AAAAAA',
		tabsBackgroundSelectedColor: '#AAAAAA'
	}),
	winBottomA = Ti.UI.createWindow({
		backgroundColor: 'red'
	}),
	winBottomB = Ti.UI.createWindow({
		backgroundColor: 'green'
	}),
	winBottomC = Ti.UI.createWindow({
		backgroundColor: 'blue'
	}),
	tabBottomA = Ti.UI.createTab({
		title: 'Tab 1',
		window: winBottomA
	}),
	tabBottomB = Ti.UI.createTab({
		title: 'Tab 2',
		window: winBottomB,
	}),
	tabBottomC = Ti.UI.createTab({
		title: 'Tab 3',
		window: winBottomC,
		backgroundColor: 'cyan',
		backgroundFocusedColor: 'magenta',
		titleColor: 'red',
		activeTitleColor: 'blue'
	})
	tabGroupBottom = Ti.UI.createTabGroup({
		tabs: [tabBottomA, tabBottomB, tabBottomC],
		// The following two properties override the colors from TabGroup
		tabsBackgroundColor: '#88994444',
		tabsBackgroundSelectedColor: '#883333',
		// Comment the row below or set it to true to see the difference
		shiftMode: false
	}),
	buttonDisableTabNavigation = Ti.UI.createButton({
		title: 'Disable Tab Navigation'
	}),
	flagTabNavigationDisabled = false,
	buttonSetActiveTab = Ti.UI.createButton({
		title: 'Go to Tab A'
	});

buttonDisableTabNavigation.addEventListener('click', function () {
	flagTabNavigationDisabled = !flagTabNavigationDisabled;
	tabGroupDefault.disableTabNavigation(flagTabNavigationDisabled);
});
winDefaultB.add(buttonDisableTabNavigation);

buttonSetActiveTab.addEventListener('click', function () {
	tabGroupDefault.setActiveTab(tabDefaultA);
});
winDefaultC.add(buttonSetActiveTab);

buttonDefault.addEventListener('click', function () {
	tabGroupDefault.open();
});

buttonBottom.addEventListener('click', function () {
	tabGroupBottom.open();
});

mainWindow.add(buttonDefault);
mainWindow.add(buttonBottom);
mainWindow.open();

tabGroupDefault.addEventListener('open', function () {
	Ti.API.info('Default TabGroup OPEN event fired');
});
tabGroupDefault.addEventListener('close', function () {
	Ti.API.info('Default TabGroup CLOSE event fired');
});
tabGroupDefault.addEventListener('focus', function () {
	Ti.API.info('Default TabGroup FOCUS event fired');
});
tabGroupDefault.addEventListener('blur', function () {
	Ti.API.info('Default TabGroup blur event fired');
});
tabDefaultA.addEventListener('focus', function () {
	Ti.API.info('Default Tab A FOCUS event fired');
});
tabDefaultA.addEventListener('blur', function () {
	Ti.API.info('Default Tab A BLUR event fired');
});
tabDefaultA.addEventListener('selected', function () {
	Ti.API.info('Default Tab A SELECTED event fired');
});
tabDefaultA.addEventListener('UNSELECTED', function () {
	Ti.API.info('Default Tab A UNSELECTED event fired');
});

```

Note: I haven't added a new example in the docs. I was thinking the test case from here should do, but maybe it is not in the best format to go there (too long?).


hansemannn:
Yey! Will try it out asap.

jquick-axway:
Cool. I like where this is heading. :smile:

I'm curious about the ripple effect you are adding. Does it work like the Google Play tabs? Where tapping a tab changes the background color of the whole tab bar with that circle scaling effect?

Oh and the `TiUIAbstractTab` class should probably stay `abstract` eh?

ypbnv:
Yes, ActionBar tabs had the ripple effect by default and since I am messing with the background drawables to support different colors for selected/unselected states I needed to add it manually.

`TiUIAbstractTab` is not abstract anymore so I renamed it in the most recent commit.

jawa9000 (approved):
This yml update looks good to me.

lokeshchdhry:
@ypbnv , On Android 4.1:
1. Clicking on `DEFAULT`  I get:
```
[ERROR] :  [Android SDK built for x86] TiExceptionHandler: (main) [26149,2715682]
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.widget.tabgroup.TiUIAbstractTabGroup.addTab(TiUIAbstractTabGroup.java:190)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.TabGroupProxy.handlePostOpen(TabGroupProxy.java:442)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.TabGroupProxy.windowCreated(TabGroupProxy.java:416)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiActivityWindows.windowCreated(TiActivityWindows.java:33)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiBaseActivity.windowCreated(TiBaseActivity.java:558)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiBaseActivity.onCreate(TiBaseActivity.java:681)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiActivity.onCreate(TiActivity.java:23)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.Activity.performCreate(Activity.java:5008)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1079)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2023)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2084)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.access$600(ActivityThread.java:130)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread$H.handleMessage(ActivityThread.java:1195)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.os.Handler.dispatchMessage(Handler.java:99)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.os.Looper.loop(Looper.java:137)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.main(ActivityThread.java:4745)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     java.lang.reflect.Method.invokeNative(Native Method)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     java.lang.reflect.Method.invoke(Method.java:511)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:786)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     com.android.internal.os.ZygoteInit.main(ZygoteInit.java:553)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     dalvik.system.NativeStart.main(Native Method)
```
2. Clicking on `BOTTOM` I get :
```
[ERROR] :  [Android SDK built for x86] TiExceptionHandler: (main) [97539,2813221]
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.widget.tabgroup.TiUIAbstractTabGroup.addTab(TiUIAbstractTabGroup.java:190)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.TabGroupProxy.handlePostOpen(TabGroupProxy.java:442)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     ti.modules.titanium.ui.TabGroupProxy.windowCreated(TabGroupProxy.java:416)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiActivityWindows.windowCreated(TiActivityWindows.java:33)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiBaseActivity.windowCreated(TiBaseActivity.java:558)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiBaseActivity.onCreate(TiBaseActivity.java:681)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     org.appcelerator.titanium.TiActivity.onCreate(TiActivity.java:23)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.Activity.performCreate(Activity.java:5008)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1079)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2023)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2084)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.access$600(ActivityThread.java:130)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread$H.handleMessage(ActivityThread.java:1195)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.os.Handler.dispatchMessage(Handler.java:99)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.os.Looper.loop(Looper.java:137)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     android.app.ActivityThread.main(ActivityThread.java:4745)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     java.lang.reflect.Method.invokeNative(Native Method)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     java.lang.reflect.Method.invoke(Method.java:511)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:786)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     com.android.internal.os.ZygoteInit.main(ZygoteInit.java:553)
[ERROR] :  [Android SDK built for x86] TiExceptionHandler:     dalvik.system.NativeStart.main(Native Method)
```

ypbnv:
@lokeshchdhry I made a change that is exactly on the line reported in the console log. Would you be able to give it a shot now?
TBH I don't see a reason why it should crash there, but the field turned out to be unnecessary as it is provides the same value as the default one.

lokeshchdhry:
@ypbnv , Thanks. I don't the exception anymore.

lokeshchdhry:
@yordan, I am seeing one more issue:
When using the bottom style tabgroup. When I scroll to go to other tabs using swipe the tab title does not show the tab title zoom effect which happens for the active tab. But, when I click the tab & then start scrolling using the swipe I do see the tab title zoom effect.

I have attached a short clip of the issue below.
tabgroup_issue

Also, I am not sure what to see when `shiftMode` is `true`.

ypbnv:
@lokeshchdhry Nice catch! TabLayout automatically selects the first Tab that is added in it, but that is not the case for BottomNavigatioView. I updated the PR with a fix.

For the shift mode - it kicks in when you have 4 or 5 tabs, so to see it in action it would need one more tab to be added in the sample code. What it does is that it hides the title of the unselected tabs and shrinks their icon by half.

lokeshchdhry:
FR Passed.

Refactor of `tabgroup` looks good. All properties seems to work as expected.

Studio Ver:     5.1.2.201810080801
SDK Ver:        7.5.0. local build
OS Ver:         10.14
Xcode Ver:      Xcode 10.0
Appc NPM:       4.2.14-2
Appc CLI:       7.0.7-master.9
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.3
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

jquick-axway:
If you run the "kitchesink-v2" project with this PR, notice that the text color for each tab is now "black" instead of "white" like it was before. This is because the "custom_theme.xml" it is using set the tab text color via an "ActionBar.TabText" style which the new code no longer uses.
https://github.com/appcelerator/kitchensink-v2/blob/master/app/platform/android/res/values/custom_theme.xml

It's not a bug with this PR, but it does mean we need to update the kitchensink-v2 custom theme file to change the `TabLayout` text color style as well.

Also, when you show the `TabGroup` in landscape form, the tabs are no longer embedded within the top ActionBar. The tabs are below the ActionBar. This is a major change, but it now matches Google's material design, which is what we want. We may get app devs complaining about landscape eating up too much vertical space. Yordan, perhaps we should add "theme" property support to `TabGroup` in this PR so that app devs can set it to "Theme.AppCompat.NoTitleBar". I know many app devs want this ability anyways. And it saves vertical space. (I've tested this theme via the manifest and it works. Great job!)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIAbstractTabGroup.java
```diff
@@ -60,12 +326,140 @@ public void processProperties(KrollDict d)
...
+		// Android docs say we must override the default implementation if item position can change
+		// The FragmentPagerAdapter uses the ID to construct the fragment tag (android:switcher:containerID:ID)
+		// and then checks the FragmentManager for presence of said tag. The default Android implementation
+		// returns the fragment position in the ViewPager, which is of course wrong when a tab is removed
+		// and fragments in higher positions move. Thus we maintain the position and IDs in an ArrayList ourselves.
+		@Override
+		public long getItemId(int position)
+		{
+			long id = tabs.get(position).hashCode();
+			return id;
```
jquick-axway: The returned hash code is not guaranteed to be unique.
How did this work before?

apidoc/Titanium/UI/Android/Android.yml
```diff
@@ -797,6 +797,20 @@ properties:
...
 
+  - name: TABS_STYLE_DEFAULT
+    summary: The default TabGroup style that is an implementation similar to the (deprecated in Android) ActionBar.Tab implementation.
+    type: Number
+    permission: read-only
+    platforms: [android]
+    since: "7.5.0"
+
+  - name: TABS_STYLE_BOTTOM_NAVIGATION
+    summary: TabGroup style taking advantage of a bottom navigation controller for swithing between tabs.
```
jquick-axway: Typo: "swithing"

apidoc/Titanium/UI/Android/Android.yml
```diff
@@ -797,6 +797,20 @@ properties:
     permission: read-only
     since: "5.2.0"
 
+  - name: TABS_STYLE_DEFAULT
+    summary: The default TabGroup style that is an implementation similar to the (deprecated in Android) ActionBar.Tab implementation.
```
jquick-axway: I don't think we should mention the word "deprecated" here.
How about we just say this implements tabs above the content, but below the title bar?

hansemannn:
Hey there!

We heavily rely on a stable BottomNavigationView. Can someone confirm that the following use-cases work:
- Using bottom-tab-group in a drawer
- Using bottom-tab-group with only icons (which are vertically centered in the tabs)
- Swipe between tabs
- Proper focus/blur events when entering/leaving tabs.

Thank you guys!

android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
```diff
@@ -456,7 +485,8 @@ public void closeFromActivity(boolean activityIsFinishing)
 		for (TabProxy tab : tabs) {
 			tab.close(activityIsFinishing);
 		}
-
+		tabs.clear();
+		tabs = null;
```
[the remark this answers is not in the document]
ypbnv: That makes sense. I will remove the setting to null.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIAbstractTabGroup.java
ypbnv: It had an AtomicLong that generated IDs for each added tab. I will push an update with a simplified flow like it worked before.

ypbnv:
@jquick-axway Thanks for the feedback!
I tried to come up with a good way to deal with having space around tabs when they can not fill the screen width (being too few or being in Landscape orientation), but I was not able to do so. What I have done is I put back `backgroundColor` as a supported property for TabGroup. It will change the color of the space behind the tabs and in cases they do not fill the width this color will be seen. I think this is a good compromise for not complicating the layout too much, stick to the native behavior and provide an option for users to make it fit with their content. The only thing that bothers me is that way we will not have parity with iOS ( IIRC it uses `tabsBackground` for coloring the space around the tabs ). 

And I have totally overlooked the `theme` property. I extracted the `fillIntent()` method in TiWindowProxy and put there all the common properties between Window and TabGroup.

ypbnv:
> 
> 
> Hey there!
> 
> We heavily rely on a stable BottomNavigationView. Can someone confirm that the following use-cases work:
> 
>     * Using bottom-tab-group in a drawer
> 
>     * Using bottom-tab-group with only icons (which are vertically centered in the tabs)
> 
>     * Swipe between tabs
> 
>     * Proper focus/blur events when entering/leaving tabs.
> 
> 
> Thank you guys!

@hansemannn 

TabGroup will not work together with DrawerLayout. The latter expects Ti.UI.View instances as content for left/center/right view and TabGroup is a child of Ti.UI.Window. 

Bottom navigation style sticks to the native behavior where `title` is a requirement for creating the items in the navigation bar. The TextView is there even if we use an empty String for a title and the icon is placed above it, thus not centered vertically. One way would to achieve that would be with custom views, but we do not support that at this moment.

If you mean using swipe gesture to change the currently selected tab instead of clicking on the controller - yes this works.

As far as I have tested the code the order and the timing of the select/unselect and focus/blur events is the same as before the refactoring.

hansemannn:
Thanks @ypbnv! Not being able to use a drawer is kind of a blocker for us :(. In native Android (proof-of-concept), we are easily able to attach the activity that holds the `BottomNavigationView` to a `DrawerLayout`, since the bottom-nav is a simple view on Android. We would expect that this is the same in Titanium, where the Window instances are being unwrapped and the underlaying view is used for it's tabs.

For the centered icon, we'd like to prevent custom views. Again, on native Android, we would just use `app:labelVisibilityMode="unlabeled"` (see ref [here](https://stackoverflow.com/a/51574662/5537752), which may be possible with custom theming in Titanium as well.

jquick-axway:
@hansemannn, what you want is `TabbedBar` which we don't support yet. A `TabbedBar` can be placed in whatever view you want... including a `DrawerLayout` side panel.

hansemannn:
I agree! Maybe we need to make an own module for now. The parity would be a bit add in that case, since the iOS UITabBar remains visible even in child-windows, which wouldn't work in Android anyway.

jquick-axway:
@ypbnv, one of the biggest features this PR offers is bottom tabs support. Since the bottom tab bar doesn't fill the bottom of the window, I'm thinking we should delay this PR to 8.0.0. I definitely like the rest of the changes such as the ability to theme it without an ActionBar/Title and the removal of the deprecated ActionBar tabs. But I think the bottom tabs is what most of app devs will be after. This is the only thing that's blocking it.

Also, my "safeAreaPadding" feature will miscalculate the dimensions with this PR, but I was expecting this. Currently, my code assumes the activity's "layout" view is the `ViewPager` which hosts the tab window, but that's no longer a valid assumption because it'll include the top `TabLayout` view or `BottomNavigationView`. I want the tabs to excluded from the padding. Hmm... maybe my "safeAreaPadding" PR should simply search for a `ViewPager` under layout when using a TabGroup activity. Kind of a cheat, but would work with or without your PR changes.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java
```diff
@@ -0,0 +1,290 @@
...
+		this.mBottomNavigationHeightValue = activity.getResources().getDimensionPixelSize(resourceID);
+
+		this.mBottomNavigationView = new BottomNavigationView(activity);
+		setTabMaxWidth();
+
+		// Set the colorPrimary as backgroundColor by default if do not have the backgroundColor set.
+		if (proxy.hasPropertyAndNotNull(TiC.PROPERTY_BACKGROUND_COLOR)) {
+			this.mBottomNavigationView.setBackgroundColor(
+				TiColorHelper.parseColor(proxy.getProperty(TiC.PROPERTY_BACKGROUND_COLOR).toString()));
+		}
```
jquick-axway: I've figured out how to make the tab bar fill the window. You need to replace the above if-condition with this...
```java
if (proxy.hasPropertyAndNotNull(TiC.PROPERTY_TABS_BACKGROUND_COLOR)) {
	this.mBottomNavigationView.setBackgroundColor(
		TiColorHelper.parseColor(proxy.getProperty(TiC.PROPERTY_TABS_BACKGROUND_COLOR).toString()));
} else {
	try {
		int resourceId = org.appcelerator.titanium.util.TiRHelper.getResource("attr.colorPrimary");
		android.util.TypedValue typedValue = new android.util.TypedValue();
		activity.getTheme().resolveAttribute(resourceId, typedValue, true);
		this.mBottomNavigationView.setBackgroundColor(typedValue.data);
	} catch (Exception ex) {
		Log.e(TAG, "Failed to set BottomNavigationView background color.", ex);
	}
}
```

Note that you need to change `PROPERTY_BACKGROUND_COLOR` to `PROPERTY_TABS_BACKGROUND_COLOR`. The "backgroundProperty" is not supposed to be applied to the tab bar. You need to use "tabsBackgroundColor" property instead. I've verified that this is how it works on iOS too. Our "backgroundProperty" is already correctly applying that color to the `ViewPager`, which is the correct thing to do and matches iOS.

The above works because `BottomNavigationView` is correctly filling the width of the window. It just didn't have a background color applied to it (was transparent by default). So, the trick was to apply the same background color to it as the tabs. I've verified that this works with 1 bottom tab in landscape. It works great. Also, setting the background color this way matches what Google recommends below...
https://material.io/develop/android/components/bottom-navigation-view

The only thing I don't like in my code is how I'm assuming that the default tab color is "colorPrimary". That said, Google does the same in the above link's example, but I wish there was a better way. I tried to do the below in my `else` condition, but it didn't work. Not sure why.
```java
int resourceId = this.mBottomNavigationView.getItemBackgroundResource();
this.mBottomNavigationView.setBackgroundResource(resourceId);
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUITabLayoutTabGroup.java
```diff
@@ -0,0 +1,197 @@
...
+	public void addViews(TiBaseActivity activity)
+	{
+		this.mTabLayout = new TabLayout(activity);
+		// Set the colorPrimary as backgroundColor by default if do not have the backgroundColor set.
+		if (proxy.hasPropertyAndNotNull(TiC.PROPERTY_BACKGROUND_COLOR)) {
+			this.mTabLayout.setBackgroundColor(
+				TiColorHelper.parseColor(proxy.getProperty(TiC.PROPERTY_BACKGROUND_COLOR).toString()));
+		} else {
+			this.mTabLayout.setBackgroundColor(this.colorPrimaryInt);
+		}
```
jquick-axway: We need to use `PROPERTY_TABS_BACKGROUND_COLOR` here, not `PROPERTY_BACKGROUND_COLOR`.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java
jquick-axway: I just noticed you have a `this.colorPrimaryInt`. Sounds like we should use that in my `else` condition instead since you've already fetched it.

jquick-axway (changes requested):

jquick-axway:
@ypbnv, I've found a solution for the bottom tab bar. Woo-hoo! :)
Please see my latest code review comments with attached code sample. I've already ran it on my end and it works on Android 4.4 and Android 9.0.

Also, I don't think we need the "maxTabWidth" solution anymore. Since it's using reflection to modify this setting, I'm thinking we should get rid of it. It's no longer needed.

jquick-axway (approved):
CR: Pass

jquick-axway:
There is 1 new problem with this PR. It ignores the "extendSafeArea" property for `TabGroup` which I added in my PR. I spent a few hours trying to add "extendSafeArea" support to this PR, but I've concluded that it would take me a couple of days to do it. My current `TiActivitySafeAreaMonitor.java` code currently obtains the tab height via the `ActionBar`, but since tabs are now now implemented via separate views (which is a good thing), I need another means of obtaining those additional insets. I'm thinking I'll need to create a new Java `InsetProvider` class to do this.

I think this PR is fine. I'm okay with us dropping "extendSafeArea" support in tab groups for now. This is something we can add later.

ypbnv:
@hansemannn
 - I am OK with such a change in the name. ( CC @jquick-axway )
 - In Design support libraries 27.1 ( what we use in Titanium ) shifting mode is determined enabled\disabled based on the number of items in the BottomNavigationView. If you have 3 tabs or less it does not kick in. Can you share any details about how it is disabled by default in native Android with you? I may have overlooked something.
 - I suppose `visibility` refers to `labelVisibilityMode`. If that is the case the proper way to handle this would be to update the Design support libraries to a newer version ( since this property is non-existent in 27.1 ) and after that expose all the new additions for the component. Otherwise forcing it with current version will need a "hackish" workaround which may result in faulty behavior.
 - No, support for `activeTabIconTint` and `unselectedItemTintColor` has not been added on Android side with this refactoring.
 - The ripple effect uses the color defined in the theme. Overriding `colorControlHighlight` in your theme ( See: https://developer.android.com/reference/android/R.attr#colorControlHighlight ) should allow you to change the color of the ripple effect.  
 - `swipeable` is the property you are looking for.
 - Using a `TabGroup` you can't have a special treatment for a button as of now. All the items in the navigation controller are bound to a `Tab` added in the `TabGroup`. Similar UI would be possible with the upcoming `TabbedBar` on Android where the items in the controller are not bound with anything, so it will simply be a matter of defining a listener for the `click` event.

hansemannn:
Thanks Yordan! Then all we need is the `labelVisibilityMode` for now I think.

jquick-axway:
Changing the property from `shiftMode` to `shiftModeEnabled` sounds fine to me.

Note that we pushed this PR out to Titanium 8.0.0. We really wanted to get this into 7.5.0, but the team here has made a commitment to get an RC out today. So, this didn't make the cut (even though I really wanted it too). Well, it gives us time to perfect the remaining aspects of it and for me to resolve the safe-area handling in the near future.

hansemannn:
Very sad to hear. Any chances to get it into GA? We need to release with a 7.x SDK soon. Let me try to cherry pick it for now then. Can the design lib update be performed then? Greets to Cali!

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java
```diff
@@ -0,0 +1,260 @@
...
+		// Set the click listener.
+		menuItem.setOnMenuItemClickListener(this);
+		// Set the title.
+		if (tabProxy.hasPropertyAndNotNull(TiC.PROPERTY_TITLE)) {
+			menuItem.setTitle(tabProxy.getProperty(TiC.PROPERTY_TITLE).toString());
+		}
+		// Set the icon.
+		if (tabProxy.hasPropertyAndNotNull(TiC.PROPERTY_ICON)) {
+			Drawable drawable = TiUIHelper.getResourceDrawable(tabProxy.getProperty(TiC.PROPERTY_ICON));
+			menuItem.setIcon(drawable);
```
jquick-axway: There is an issue here with loading image icons for the top tabs from APK "assets". The old version of `TabBar` used to scale the images loaded from "assets". If I remember right, the old code assumed asset images were "mdpi" and scaled them to the device's current density. This new code loads the images as-is, which means they end up being extra tiny on a xxxhdpi device, such as my Pixel 2 phone.

Our `BottomNavigationView` does not have this problem and scales "asset" images. Perhaps you should use the same code?

jquick-axway: You can test this with our "tab1.png" and "tab2.png" files that our Titanium Classic app template comes with.

ypbnv: Updated.

hansemannn:
Quick note: You could add even more parity by exposing the `badge` property via:
```java
bottomNavigation.setNotification(notification, bottomNavigation.getItemsCount() - 1);
```

ypbnv:
@jquick-axway I did some testing and it seems that prior to the refactoring the image was still passed as a drawable with its raw dimensions. Here a two screenshots for the same project with the same resources:
refactored
current

I also thought that they were scaled according to the device, but I suppose the difference was more noticeable due to the new layout. That opens the question if we still want to scale them in the refactored code?
Technically it will be a change to the behavior before that, but I am OK if we decide to add it. I tried a couple of things to deal with it, but so far the best I got is related to a change in one of the layout files (https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/design/res/layout/design_layout_tab_icon.xml) in the Design library we package with the app. I think somewhere in the process of updating the ImageView in this file with a new source is the problem, because this layout does not have the same behavior in a new clean native Android project, provided with the same source.

jquick-axway:
@ypbnv, I'm sorry man. You are exactly right. The old tab group handling did not scale the tab icons. My bad.
`Titanium 7.5.0`
tabgroupicononly_7 5 0
`Titanium 8.0.0`
tabgroupicononly_8 0 0

jquick-axway:
@lokeshchdhry, this PR is ready for you.

lokeshchdhry:
Passing final FR (first was done earlier before CR) as all CR comments seems to be looked at & appropriate changes made.

OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.8
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

android/modules/ui/src/js/tabgroup.js
```diff
@@ -59,6 +59,18 @@ exports.bootstrap = function(Titanium) {
...
+	TabGroup.prototype.removeTab = function(options) {
+		if (this.currentState != this.state.opened) {
+			var index = this._tabs.indexOf(options);
+				if (index > -1) {
+				  this._tabs.splice(index, 1);
+				}
+		} else {
+			_removeTab.call(this, options);
+		}
+	}
```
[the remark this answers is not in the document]
hansemannn: @sgtcoolguy Great work on PR validation!

jquick-axway:
@juliengatt , Titanium does not support tinting the tab icons on Android. So, the icon PNG you assign to the tab is displayed as-is.

hansemannn:
We could use `itemIconTint` together with the `ColorStateList` for the checked and unchecked state:
```xml
<selector xmlns:android="http://schemas.android.com/apk/res/android">
  <item
      android:state_checked="true"
      android:color="@color/colorPrimary" />
  <item
      android:state_checked="false"
      android:color="@color/grey" />
 </selector>
```
Should be possible via XML.

jquick-axway:
Yes, it is possible to do it natively on Android, but Titanium does not currently support it.

hansemannn:
But shouldn't it be possible with global styling. In any case, it's one of the last tabgroup parity issues effecting devs in real life use cases

jquick-axway:
> But shouldn't it be possible with global styling?

On Android, you would normally do this via a drawable in XML. In that drawable's XML, you can define the different states. I think our internal code will load a drawable as-is in this case (not assume it's a BitmapDrawable) when loading from a `res/drawable` folder. So, perhaps it's theoretically possible to do it now.

juliengatt:
If already possible, an example would be greatly appreciated!

jquick-axway:
You would need to create a "selector" in XML kind of like what is shown in the link below. This XML file would go under your project's `./platform/android/res/drawable` folder. Let's say you call it "my_tab_icon1.xml".
https://stackoverflow.com/a/49854402

And then you would assign the above to the tab's icon as follows...
```javascript
Ti.UI.createTab({
	icon: "android.resource://" + Ti.App.id + "/drawable/my_tab_icon1",
});
```

Note that I haven't tested the above. But "something" like that should theoretically work. You'll have to figure it out from there.

I've also written up a feature request ticket below.
https://jira.appcelerator.org/browse/TIMOB-26734

juliengatt:
@jquick-axway I haven't been able to make it work. I tried the following:

Ti.UI.createTab({
	icon: "android.resource://" + Ti.App.id + "/drawable/my_tab_icon1"
        //Tried this too: Ti.App.Android.R.drawable.my_tab_icon1
});

//my_tab_icon1.xml and selected.png & unselected.png in drawable folder
<selector xmlns:android="http://schemas.android.com/apk/res/android">
  <item android:drawable="@drawable/selected" android:state_selected="true"/>
   <item android:drawable="@drawable/unselected" android:state_selected="false"/>
</selector>

I have found out that whatever icon color I add to the folder, the selected icon color will always be the opposite. I don't think this should be the expected behavior.

Please advise.
Thanks

juliengatt:
> We could use `itemIconTint` together with the `ColorStateList` for the checked and unchecked state:
> 
> ```
> <selector xmlns:android="http://schemas.android.com/apk/res/android">
>   <item
>       android:state_checked="true"
>       android:color="@color/colorPrimary" />
>   <item
>       android:state_checked="false"
>       android:color="@color/grey" />
>  </selector>
> ```
> 
> Should be possible via XML.

@hansemannn did you make it work?
Thanks

juliengatt:
I have seen the following ticket but not able to reproduce:
https://jira.appcelerator.org/browse/TIMOB-3179

Brianggalvez:
@juliengatt try with a custom theme

customTheme.xml

```
<resources>

  <!-- Base application theme. -->
  <style name="rehaviour" parent="Theme.AppCompat.Light">
    <item name="colorPrimary">@color/colorPrimary</item>
    <item name="colorAccent">@color/colorAccent</item>
    <item name="android:navigationBarColor">@color/turquoise</item>
    <item name="android:statusBarColor">@color/colorPrimaryDark</item>
    <item name="android:textColorPrimary">@color/turquoise</item>
    <item name="android:textColorPrimaryInverse">@color/colorPrimaryDark</item>
    <item name="itemIconTint">@drawable/nav_item_color_state</item>
  </style>
</resources>
```
drawable/nav_item_color_state.xml
```
<selector xmlns:android="http://schemas.android.com/apk/res/android">
    <item android:state_checked="true" android:color="@color/colorAccent"/>
    <item android:state_checked="false" android:color="@color/colorPrimaryDark"/>
</selector>
```

PR #10359, opened 2018-10-01
[TIMOB-26424] Android: Implement missing Ti.App events

garymathews:
- Implement missing `Titanium.App` events

###### TEST CASE
```JS

Ti.App.addEventListener('resume', () => {
    Ti.API.info('RESUME');
});
 
Ti.App.addEventListener('pause', () => {
    Ti.API.info('PAUSE');
});


Ti.App.addEventListener('resumed', () => {
    Ti.API.info('RESUMED');
});
 
Ti.App.addEventListener('paused', () => {
    Ti.API.info('PAUSED');
});
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26424)


android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -831,6 +835,35 @@ private void stopExternalStorageMonitor()
...
+			public void onReceive(Context context, Intent intent)
+			{
+				unlockedScreen = true;
+			}
+		};
+
+		IntentFilter filter = new IntentFilter();
+		filter.addAction(Intent.ACTION_SCREEN_ON);
+		registerReceiver(lockScreenReceiver, filter);
+	}
```
jquick-axway: One thing that bothers me is if `hasUnlockedScreen()` was called twice, it'll get reset to `false` on the second call. So, I'm thinking we should listen to both `ACTION_SCREEN_ON` and `ACTION_SCREEN_OFF`. Then in the `onReceive()`, you can set the `unlockedScreen` member variable based on which intent action was received like this...
```java
public void onReceive(Context context, Intent intent)
{
	switch (intent.getAction()) {
		case Intent.ACTION_SCREEN_ON:
			unlockedScreen = true;
			break;
		case Intent.ACTION_SCREEN_OFF:
			unlockedScreen = false;
			break;
	}
}
```

jquick-axway: And maybe we should add the following the if-check to the top of this method in case`startLockScreenMonitor()` was called twice. This will make it bullet proof and prevent a hanging listener that'll continue to receive intents.
```java
if (lockScreenReceiver != null) {
	return;
}
```

The rest looks great!

jquick-axway (changes requested):

lokeshchdhry:
FR Passed.

The `PAUSE`, `PAUSED`, `RESUME` & `RESUMED` events are fired as expected.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

PR #10360, opened 2018-10-01
[TIMOB-26425] Always use thread-safe access to timers array

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26425

**Optional Description:**
Timers are always started in their own thread. Locking should also be done when running on main thread to make the access to the timers array thread-safe.

### DIFF
--- a/iphone/Classes/KrollContext.h
+++ b/iphone/Classes/KrollContext.h
@@ -40,10 +40,10 @@
 #endif
   TiGlobalContextRef context;
   NSMutableDictionary *timers;
+  NSRecursiveLock *timerLock;
   void *debugger;
 
 #ifdef TI_USE_KROLL_THREAD
-  NSRecursiveLock *timerLock;
   NSString *krollContextId;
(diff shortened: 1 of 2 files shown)


vijaysingh-axway (changes requested):
@janvennemann  Your logic looks good to me. You need to adjust this PR a bit e.g declaration and definition of variable timerLock should be out of TI_USE_KROLL_THREAD macro.
Can you create PR for 7_4_X so we can ask reporter to test? Thanks!

janvennemann:
Addressed the missing declarations and moved them out of the kroll macro.

sgtcoolguy:
[mocha_2018-09-27-000653_morimoto-macos.crash.txt](https://github.com/appcelerator/titanium_mobile/files/2438902/mocha_2018-09-27-000653_morimoto-macos.crash.txt)

janvennemann:
@sgtcoolguy Updated the PR and the Jenkins build is looking good now.

PR #10327, opened 2018-09-19
[TIMOB-26391](7_4_X) : iOS App randomly freezes on iOS 12 beta when using main-thread

vijaysingh-axway:
**JIRA:** https://jira.appcelerator.org/browse/TiMOB-26391

PR #10361, opened 2018-10-02
[TIMOB-26391](7_4_X) : iOS App randomly freezes on iOS 12 beta when using built-in JSCore

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26391

### DIFF
--- a/iphone/Classes/KrollBridge.m
+++ b/iphone/Classes/KrollBridge.m
@@ -680,6 +680,14 @@ CFMutableSetRef krollBridgeRegistry = nil;
 
 - (void)unregisterProxy:(id)proxy
 {
+#ifdef USE_JSCORE_FRAMEWORK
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
I believe that this should likely be rejected? Sounds like Jan is trying to look into a better location for the unprotect call?

Also, I'm not 100% certain that the issue for "freezes" is necessarily an GC issue. If we mess up GC (protect/unprotect) that should cause a crash. If an app is "freezing" it's likely due to a deadlock or some threading issue. Perhaps we should roll back the change we made to timers to flip the "wait" boolean flag when it invoked the block on the main thread (our first speculative fix for the freeze/crash ticket)?

vijaysingh-axway:
Agreed. It is holding the proxy for lifetime. So it can cause some different issue. I will revert the changes of [PR-10327](https://github.com/appcelerator/titanium_mobile/pull/10327). Thanks!

PR #10364, opened 2018-10-03
[TIMOB-26403] Update node-titanium-sdk, pass logger to analyzeJs

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26403

- Update node-titanium-sdk@0.6.4 to include fix
- Pass logger into to analyzeJs so we can log deprecation warning

To test:

1. Build kitchensink-v2 to device or simulator
- You should see a warning log about "implicit global scope being deprecated
- It should launch successfully


sgtcoolguy:
CR and FR pass!

PR #10369, opened 2018-10-04
[TIMOB-26435] Update gradle for JDK 11 support

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26435

- Update to gradle 4.10.2
- Use https to download gradle
- Change "Failed to run dexer" warning to "Failed to run gradle"
	- We have (had) 3 potential error cases where we log "Failed to run dexer" so tracking down _what_ went wrong is kind of a PITA
	- I'd like to reduce this to 1 "Failed to run dexer" error case, so if anyone has suggestions for [this](https://github.com/ewanharris/titanium_mobile/blob/77fd3f2acd5976dd63eb489a4f42d60d9426c2f8/android/cli/commands/_build.js#L4398) error log, send them in on a postcard 🙂 I think it's still the dexer step but can we make it a little more clear what the task is?

**JDK 11 Test:**
1. Install JDK 11. _(Note that this will currently break Appc Studio.)_
2. Build and run a Titanium app for Android.
3. Verify that it builds and runs successfully.

**Multidexing Test:**
Run the test shown in [TIMOB-25597](https://jira.appcelerator.org/browse/TIMOB-25597), but use this PR's Titanium version. Verify that the app can launch successfully on an Android 4.x without crashing.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -4376,7 +4376,7 @@ AndroidBuilder.prototype.runDexer = function runDexer(next) {
 				'-Pconfiguration=' + quotePath(baserules) + pathArraySeparator + quotePath(mainDexProGuardFilePath)
 			], { shell: true, windowsHide: true }, function (code, out, err) {
 				if (code) {
-					this.logger.error(__('Failed to run dexer:'));
+					this.logger.error(__('Failed to run gradle:'));
 					this.logger.error();
 					err.trim().split('\n').forEach(this.logger.error);
 					this.logger.log();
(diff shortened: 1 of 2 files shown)


jquick-axway (approved):
CR: Pass

Thanks @ewanharris! :)

jquick-axway:
> Change "Failed to run dexer" warning to "Failed to run gradle"

I'm okay with this... even though we're not technically building with "gradle" yet. We only use "gradle" to download the newest ProGuard, which is used to scan which Java classes are needed to be in the main dex file when multidexed. But it's a critical step that all app builds need to go through.

janvennemann:
@ewanharris for the error message, how about: "Generating main class list for the dexer failed" At least that's what the command is supposed to do, right?

lokeshchdhry:
FR Passed.

Build to device, emulator works fine with JDK 11 from CLI. Also, packaging & module creation works fine as well.

OS Ver:         10.13.5
Xcode Ver:      Xcode 9.4.1
Appc NPM:       4.2.13
Appc CLI:       7.0.6
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.2
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       11

lokeshchdhry:
Multidex tests are pending ....

PR #10370, opened 2018-10-04
[TIMOB-26391](7_4_X) Improve proxy protection logic

janvennemann:

**Optional Description:**
Using the existing protect/unprotect methods caused unwanted side effects. This introduces two similar methods for the sole purpose of safeguarding the JS object during our proxy creation flow.


vijaysingh-axway:
Both apps are working fine at my end.

vijaysingh-axway:
Two methods which you have introduced-
-(void)applyGarbageCollectionSafeguard 
-(void)removeGarbageCollectionSafeguard
Is similar to
- (void)protectJsobject
- (void)unprotectJsobject

Only difference I can see is updating location of bool ‘gcSafeguarded’ and ‘protecting’ in theses functions. Is bool value updating location is cause of some issue?

vijaysingh-axway (approved):
CR passed.

PR #10376, opened 2018-10-10
Fix for CalendarProxy.yml

jawa9000:
The name property was changed from Titanium.Android.Calendar.CalendarProxy to Titanium.Android.Calendar.Calendar. This fix corrects that error so the document can be published again.


sgtcoolguy:
@jawa9000 The doc validation script doesn't like this change, because other references to the type name weren't also updated.

sgtcoolguy:
@jawa9000 This is still failing the doc validation.

jawa9000:
After chatting with Chris W, we sorted out the issue here and decided no action is needed here.

PR #10379, opened 2018-10-16
[TIMOB-26391] Improve proxy protection logic

janvennemann:
Since we didn't do any further changes to the proxy protection logic after testing with 7.4.1 we should take this into 7.5.0 and continue from here if we get more input from https://jira.appcelerator.org/browse/TIMOB-26391


vijaysingh-axway:
@janvennemann  Can you please fix lint  issue.

janvennemann:
@vijaysingh-axway linting issues fixed.

vijaysingh-axway (approved):
CR passed.

PR #10380, opened 2018-10-16
fix(android): optimize imageAsResized

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26456


garymathews (approved):
CR: **PASS**

m1ga:
The test is failing because the image will be 50x50 (since the source image is squared) and it is not stretching the result. So this method can only be used if the ratio is correct or it has to be stretched in the last step e.g. use `imageResized = Bitmap.createBitmap(img, 0, 0, imgWidth, imgHeight, matrix, true);` for non-squared images at the end. But I need to verify that

lokeshchdhry:
FR Passed.

The PR does save some memory.

Studio Ver:     5.1.2.201903111843
SDK Ver:        8.1.0 local build
OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.10
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.9
Node Ver:       8.15.1
NPM Ver:        6.4.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

PR #10384, opened 2018-10-18
[TIMOB-26462] Improve error handling and logging when failing to connect to log server

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26462

Totally unsure of the verbiage, any comments on it are more than welcome.

#### Test steps

1. Edit your `/etc/hosts` file to remote the localhost mapping
2. Build for iOS
    - Build should error informing you that we could not resolve localhost, **not** that another process is using the port

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -2558,12 +2558,11 @@ iOSBuilder.prototype.determineLogServerPort = function determineLogServerPort(ne
 				if (error.code === 'ENOTFOUND') {
 					_t.logger.error(__('Unable to connect to log server on localhost'));
 					_t.logger.error(__('Please ensure your /etc/hosts file contains a valid entry for `localhost`'));
-					process.exit(1);
 				} else {
 					_t.logger.error(__('Another process is currently bound to port %d', _t.tiLogServerPort));
 					_t.logger.error(__('Set a unique <log-server-port> between 1024 and 65535 in the <ios> section of the tiapp.xml') + '\n');
-					process.exit(1);
 				}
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -2553,11 +2553,17 @@ iOSBuilder.prototype.determineLogServerPort = function determineLogServerPort(ne
...
-				_t.logger.error(__('Set a unique <log-server-port> between 1024 and 65535 in the <ios> section of the tiapp.xml') + '\n');
-				process.exit(1);
+				if (error.code === 'ENOTFOUND') {
+					_t.logger.error(__('Unable to connect to log server on localhost'));
+					_t.logger.error(__('Please ensure your /etc/hosts file contains a valid entry for `localhost`'));
+					process.exit(1);
+				} else {
+					_t.logger.error(__('Another process is currently bound to port %d', _t.tiLogServerPort));
+					_t.logger.error(__('Set a unique <log-server-port> between 1024 and 65535 in the <ios> section of the tiapp.xml') + '\n');
+					process.exit(1);
```
cb1kenobi: No need to duplicate `process.exit(1)`. A single call outside the `if` block would suffice.

cb1kenobi (changes requested):

iphone/cli/commands/_build.js
ewanharris: Updated

cb1kenobi (approved):

cb1kenobi:
**APPROVED**

PR #10388, opened 2018-10-19
[TIMOB-26476] Android: Validate proxy property definitions to identify possible clashes and unexposed accessor methods

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-26476

**Description:**
This PR is a result of an issue Yordan raised, which I was aware of in the past (I believe V8 actually would error when a property was defined multiple times).

When we generate the native bindings for a proxy, I do a last pass at validating usage of the `@Kroll` annotations to attempt to find clashes/duplicates or other possible issues.

In some cases it will error out the build until you reconcile the issue. In others it will suggest potential fixes.

I also modified our binding template to avoid the worst clashes:
- When a property is listed in `@Kroll.proxy's` `propertyAccessor` listing **and** has `@Kroll.getProperty` or `@Kroll.setProperty` methods attached. The property would get defined twice and it was undefined whether it'd create it like a typical JS property or use the native method binding.
- When a property is listed in `@Kroll.proxy's` `propertyAccessor` listing **and** a `@Kroll.method` had the same name as the get/setProperty() accessor. We do not try to define the method twice anymore, and the `@Kroll.method` one wins.

The validation attempts to find cases like these and suggest changes. It also tries to find when a `@Kroll.method` matches a property accessor's default set/getProperty() name but has no `@Kroll.getProperty/setProperty` annotation (breaking `obj.property` and `obj.property = value;` usage being the exact same as calling the accessor methods). It also find the exact opposite, when we have a property accessor method with no `@Kroll.method` - In 99% of those cases we do want that method exposed (at least until we remove accessor methods in SDK 9.0.0).


android/kroll-apt/src/java/org/appcelerator/kroll/annotations/generator/KrollBindingGenerator.java
```diff
@@ -282,10 +284,170 @@ protected void generateBindings() throws ParseException, IOException
...
+						+ proxyClassName + "." + propertyName + "'.");
+					Map<String, Object> dynamicProperty = (Map<String, Object>) dynamicProperties.get(propertyName);
+					if ((Boolean) dynamicProperty.get("set")) { // there's a setter
+						// is getter defined?
+						if ((Boolean) dynamicProperty.get("get")) {
+							System.err.println(
+								"Likely fix is to remove from proxy.propertyAccessors listing, as both getter and setter methods are defined.");
+						} else {
+							System.err.println(
+								"Likely fix is to remove from proxy.propertyAccessors listing, as a setter method is already defined. A getter IS NOT defined, so you may want to add a @kroll.getProperty implementation as well (or rely n the default getter generated, which uses #onPropertyChanged() to react).");
```
ypbnv: "@kroll.getProperty" - should be with an uppercase 'K'. A typo in "rely n the".

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationChannelProxy.java
```diff
@@ -116,14 +119,31 @@ public void setEnableVibration(boolean vibration)
 		setProperty(TiC.PROPERTY_ENABLE_VIBRATION, vibration);
```
ypbnv: This manual setting of the JS property should be gone, too?

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationChannelProxy.java
```diff
@@ -156,14 +192,31 @@ public void setImportance(int importance)
 		setProperty(TiC.PROPERTY_IMPORTANCE, importance);
```
ypbnv: Same as above - should the manual setting of the property be gone?

android/modules/android/src/java/ti/modules/titanium/android/notificationmanager/NotificationChannelProxy.java
```diff
@@ -186,14 +247,36 @@ public void setName(String name)
 		setProperty(TiC.PROPERTY_NAME, name);
```
ypbnv: Manual property setting like the previous cases - should be removed?

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/AndroidModule.java
```diff
@@ -25,12 +25,12 @@
 import android.os.Message;
 
 /**
- * AndroidModule exposes all Android specific methods and properties relating to geolocation behavior 
- * associated with Ti.Geolocation.Android to the Titanium developer.  Cross platform API points should 
+ * AndroidModule exposes all Android specific methods and properties relating to geolocation behavior
+ * associated with Ti.Geolocation.Android to the Titanium developer.  Cross platform API points should
```
ypbnv: Double whitespace before "Cross plaform API..."

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/AndroidModule.java
```diff
@@ -134,10 +140,10 @@ public void setManualMode(boolean manualMode)
 	}
 
 	/**
-	 * Creates a new instance of a location provider.  Mimics that normal proxy mechanism 
-	 * provided via annotations.  This is needed due to a flaw in how the annotation driven 
+	 * Creates a new instance of a location provider.  Mimics that normal proxy mechanism
+	 * provided via annotations.  This is needed due to a flaw in how the annotation driven
```
ypbnv: Double whitespace before "This is needed..."

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/AndroidModule.java
```diff
@@ -164,10 +170,10 @@ public LocationProviderProxy createLocationProvider(Object creationArgs[])
 	}
 
 	/**
-	 * Creates a new instance of a location rule.  Mimics that normal proxy mechanism 
-	 * provided via annotations.  This is needed due to a flaw in how the annotation driven 
+	 * Creates a new instance of a location rule.  Mimics that normal proxy mechanism
+	 * provided via annotations.  This is needed due to a flaw in how the annotation driven
```
ypbnv: Double whitespace before "This is needed..."

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/AndroidModule.java
```diff
@@ -195,11 +201,11 @@ public void addLocationProvider(LocationProviderProxy locationProvider)
 	}
 
 	/**
-	 * Adds the specified location provider to the list of manual location providers.  If a location 
-	 * provider with the same "name" property already exists in the list of manual location 
+	 * Adds the specified location provider to the list of manual location providers.  If a location
```
ypbnv: Double whitespace before "If a location..."

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/LocationProviderProxy.java
```diff
@@ -59,9 +59,9 @@
 	}
 
 	/**
-	 * Constructor.  Used primarily when creating a location provider via 
+	 * Constructor.  Used primarily when creating a location provider via
```
ypbnv: Double whitespace before "Used primarily when...".

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/android/LocationProviderProxy.java
```diff
@@ -78,13 +78,13 @@ public LocationProviderProxy(Object[] creationArgs, LocationProviderListener pro
...
```
ypbnv: Double whitespace.

ypbnv (changes requested):
Besides the comments about NotificationChannelProxy there are few places with minor things - double whitespace and typos.

PR #10392, opened 2018-10-23
[TIMOB-26479] : TiUtils.m missing check for iPhone XR and iPhone XS Max images

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26479


garymathews (approved):
CR: **PASS**

PR #10393, opened 2018-10-23
[TIMOB-26390] Android: Connect getters with background drawables for Unit test purposes.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26390

**Description:**
Connect the getters for different color states for Ti.UI.View with the Android object representing it.
Extract the method for getting colors for state in Ti.UIHelper.
Expose the states we assign for custom background drawables.
Guard for backwards compatibility.
Replace the String properties in Ti.UI.View with the proper TiC constants.

**NOTE:** This introduces behavior that may be classified as "breaking". Prior to this PR color state properties that have not been defined would return `undefined` and now that is replaced with `null`. Also colors set with predefined words ( like "green", "cyan", "yellow" ) will now return the hexadecimal value with alpha channel. For instance setting a `backgroundColor` to `blue` will afterwards return the value `#FF0000FF`. Please, let me know what do you think  about these changes and if they need reworking, mention in the docs or anything that this PR is missing.
The failed unit tests show pretty well what I had in mind with "breaking" change. Having a list with predefined color words for specific color values does solve the problem ( I personally don't like this approach because there may be a difference in naming specific color values across platforms** ). I see we have a defined color map, but it may need a bit of a rework to help in this case.
** A good example is Android - they are supporting both words "green" and "lime" according to their docs:
https://developer.android.com/reference/android/graphics/Color.html#parseColor(java.lang.String)
Which is great, but the result of the call of `Color.parseColor()` with both parameters is the same. Judging by the code it is more likely a bug, because there is no point in having two keys with the same value in their color map. 

**Test case:**
The added unit test should suffice, but here is a code sample 
_app.js_
```js
	view = Ti.UI.createView({
		width: 100,
		height: 100,
		borderColor: 'green',
		borderWidth: 5,
		backgroundColor: 'red',
		backgroundSelectedColor: 'green',
		backgroundFocusedColor: 'blue',
		backgroundDisabledColor: 'gray'
	}),
	button = Ti.UI.createButton();

win.add(view);
win.add(button);

	Ti.API.info('backgroundColor is ' + view.getBackgroundColor());
	Ti.API.info('backgroundSelectedColor is ' + view.getBackgroundSelectedColor());
	Ti.API.info('backgroundFocusedColor is ' + view.backgroundFocusedColor);
	Ti.API.info('backgroundDisabledColor is ' + view.backgroundDisabledColor);
});
win.open();

```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -8,8 +8,6 @@ package org.appcelerator.titanium.proxy;
 
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Array;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
 import java.util.HashMap;
@@ -42,9 +40,6 @@ import android.app.Activity;
 import android.graphics.Bitmap;
 import android.graphics.Bitmap.Config;
 import android.graphics.drawable.ColorDrawable;
-import android.graphics.drawable.Drawable;
-import android.graphics.drawable.LayerDrawable;
-import android.graphics.drawable.StateListDrawable;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Message;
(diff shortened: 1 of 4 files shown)


keerthi1032:
FR Passed. Waiting for CR and Merge. Hexadecimal values displayed for colors.
Test Environment:
Operating System
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = Local 7.5.0.v20181023053605 build
Device                            =Samsung s5 Android 6 
Emulator				= pixel android 7, Sony Xperia S  android 4.1.1

lokeshchdhry:
@ypbnv , Can you please fix the unit test failures. Thanks.

ypbnv:
@lokeshchdhry Removing this out of 7.5.0. Will schedule it for 8.0.0, since we need to clarify what approach will be the most suitable for the PRs purpose.

lokeshchdhry:
@ypbnv , Did we finalize any approach for this PR ? Is it good to merge ?

ypbnv:
@lokeshchdhry Not really.
@garymathews, @jquick-axway When you get the chance can you take a look at the note in the description?

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -57,44 +52,50 @@
...
 	// others
-	"focusable",
-	"touchEnabled",
-	"visible",
-	"enabled",
-	"opacity",
-	"softKeyboardOnFocus",
-	"transform",
-	"elevation",
-	"touchTestId",
```
jquick-axway: This change is removing "touchTestId", but that's probably fine since it's not referenced anywhere else in the code. Our iOS code doesn't have any references to it either.

android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
```diff
@@ -1221,4 +1222,64 @@ public static Uri getRedirectUri(Uri mUri) throws MalformedURLException, IOExcep
...
+			Drawable simpleDrawable = backgroundDrawable.getBackground();
+			if (simpleDrawable instanceof PaintDrawable) {
+				// For backwards compatibility return null if the required state is not the default one.
+				if (state != TiUIHelper.BACKGROUND_DEFAULT_STATE_1) {
+					return null;
+				}
+				return transcriptColorIntToString(((PaintDrawable) simpleDrawable).getPaint().getColor());
+			}
+			// Get the backgroundDrawable background as a StateListDrawable.
+			StateListDrawable stateListDrawable = ((StateListDrawable) simpleDrawable);
```
jquick-axway: We need to do an `instanceof` check on the `StateListDrawable` to avoid the issue mentioned here...
https://github.com/appcelerator/titanium_mobile/pull/10744

android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
```diff
@@ -1221,4 +1222,64 @@ public static Uri getRedirectUri(Uri mUri) throws MalformedURLException, IOExcep
...
+		try {
+			// TiBackgroundDrawable's background can be either PaintDrawable or StateListDrawable.
+			// Handle the cases separately.
+			Drawable simpleDrawable = backgroundDrawable.getBackground();
+			if (simpleDrawable instanceof PaintDrawable) {
+				// For backwards compatibility return null if the required state is not the default one.
+				if (state != TiUIHelper.BACKGROUND_DEFAULT_STATE_1) {
+					return null;
+				}
+				return transcriptColorIntToString(((PaintDrawable) simpleDrawable).getPaint().getColor());
```
jquick-axway: We need to do an `instanceof` check on `PaintDrawable`.

android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
```diff
@@ -1221,4 +1222,64 @@ public static Uri getRedirectUri(Uri mUri) throws MalformedURLException, IOExcep
...
+							   + drawable.getClass().toString());
+				return null;
+			}
+		} catch (NoSuchMethodException e) {
+			Log.w(TAG, "Unable to get a method for reflection.");
+		} catch (IllegalAccessException e) {
+			Log.w(TAG, "Unable to access a method for reflection.");
+		} catch (InvocationTargetException e) {
+			Log.w(TAG, "Unable to invoke a method for reflection.");
+		}
```
jquick-axway: Let's catch all exceptions here as shown in my other PR.
I'll close my PR once this PR resolves the same issue.

android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
```diff
@@ -1221,4 +1222,64 @@ public static Uri getRedirectUri(Uri mUri) throws MalformedURLException, IOExcep
...
+		return null;
+	}
+
+	public static String transcriptColorIntToString(int colorInt)
+	{
+		return String.format("#%08X", 0xFFFFFFFF & colorInt);
+	}
```
jquick-axway: I think this function should be moved to our `TiColorHelper` class.

And would you mind renaming the function from `transcriptColorIntToString(int)` to `hexStringFrom(int)` please?
I say this because Titanium also supports an "rgba(0,0,0,0)" color string as well, which if we were to support it in the future, I would name it `rgbaStringFrom(int)`.

jquick-axway (dismissed):

garymathews:
CR: **PASS**

##### TEST CASE from https://github.com/appcelerator/titanium_mobile/pull/10744
2. Verify that the app does not crash on startup.
3. Uncomment the 2 lines of code below.
4. Rebuild and rerun the app on Android.
5. Verify that the displayed `TextField` now has a gray background.
6. In the log, verify that you see the following message. _("#FF888888" is the color "gray".)_
`[INFO] :   ### textField.backgroundDisabledColor: #FF888888`

```javascript
	value: "Hello World",
	width: "80%",
	color: "white",
	backgroundColor: "black",
	borderColor: "white",
//	backgroundDisabledColor: "gray",
//	enabled: false,
});
window.add(textField);
window.open();
Ti.API.info("### textField.backgroundDisabledColor: " + textField.backgroundDisabledColor);
Ti.API.info("### JSON.stringify(textField): " + JSON.stringify(textField));
```

garymathews (approved):

jquick-axway (approved):

garymathews:
Reverted to returning the property string by default for parity and to satisfy our test cases.

keerthi1032:
FR Passed.Works as expected. 
Test Environment:
Operating System
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.1.0.v20190501153539
CLI =7.0.11-1
Studio =5.1.2.201903111843
Device =Samsung s5 android 6,pixel android 9
Emulator =pixel 3xl android 9

PR #10401, opened 2018-10-26
fix(ios): use native NSTimer to avoid thread deadlocks

janvennemann:

**Optional Description:**
Our KrollTimer implementation is using his own thread to wait until the timer is fired. This causes thread deadlocks under iOS 12. Using iOS native timer to operate directly on the run loop to try mitigate this problem.

The implementation is somewhat naive and may require some adjustments for proper memory management of the timer callbacks.


iphone/Classes/KrollContext.m
```diff
@@ -1529,3 +1536,59 @@ - (void)dealloc
...
+  return timers;
+}
+
++ (int)setIntervalInContext:(JSContext *)context withArguments:(NSArray<JSValue *> *)arguments shouldRepeat:(BOOL)repeats {
+  JSValue *callbackFunction = [[arguments objectAtIndex:0] retain];
+  double interval = [[arguments objectAtIndex:1] toDouble] / 1000;
+  NSNumber *timerIdentifier = [NSNumber numberWithUnsignedInteger:[[[[NSUUID alloc] init] autorelease] hash]];
+  NSTimer *timer = [NSTimer scheduledTimerWithTimeInterval:interval repeats:repeats block:^(NSTimer * _Nonnull timer) {
+    [callbackFunction callWithArguments:nil];
+    [callbackFunction release];
```
janvennemann: @vijaysingh-axway not sure if it's a good idea to directly release the callback again here. Any thoughts?

joshualambert:
I've been testing with this code change on my app. Super stable and fast now on iOS12. Fantastic work @janvennemann!

iphone/Classes/KrollContext.m
vijaysingh-axway: @janvennemann 
1. In case of setTimeout we should remove timer from self.timers array.
2. Is there any reason you have retained in callbackFunction. There is neither need of retain in callbackFunction nor release. Above code is crashing for me.
Test case -
`setTimeout(function(e){
      Ti.API.info('timeout called');
}, 2000);

setTimeout(function(e){
      Ti.API.info('timeout called second time');
}, 4000);

setInterval(function(e){
      Ti.API.info('time interval 1 called');
}, 3000);

setInterval(function(e){
      Ti.API.info('time interval 2 called');
}, 4000);`

iphone/Classes/KrollContext.m
```diff
@@ -1529,3 +1536,59 @@ - (void)dealloc
 }
 
 @end
+
```
vijaysingh-axway: Please wrap whole class implementation  inside macro
#if defined(USE_JSCORE_FRAMEWORK) && !defined(TI_USE_KROLL_THREAD)

iphone/Classes/KrollContext.m
```diff
@@ -1177,10 +1179,15 @@ - (void)main
   [kroll release];
   TiStringRelease(prop);
 
+#if defined(USE_JSCORE_FRAMEWORK) && !defined(TI_USE_KROLL_THREAD)
```
vijaysingh-axway: I think the problem is with JSCore only irrespective of main thread or kroll thread. If so we should update the # if defined statement to - 
#if defined(USE_JSCORE_FRAMEWORK)

iphone/Classes/KrollContext.m
```diff
@@ -1177,11 +1179,18 @@ - (void)main
   [kroll release];
   TiStringRelease(prop);
 
-  [self bindCallback:@"setTimeout" callback:&SetTimeoutCallback];
+#ifdef USE_JSCORE_FRAMEWORK
+  JSContext *jsContext = [JSContext contextWithJSGlobalContextRef:context];
+  [JSTimerManager initializeInContext:jsContext];
+#else
+  [self bindCallback:@"setTimeout"
+            callback:&SetTimeoutCallback];
```
janvennemann: No idea why it suddenly wants to format it like this. Total non-sense if you ask me. If you remove the `#ifdef` the break is not required anymore.

janvennemann:
@vijaysingh-axway i addressed your suggested changes and also reworked a few bits of this. Every `KrollContext` now has it's own timer manager instance instead of working with static class methods. I also added some clean up code and excluded some of the legacy timer code with macros if using jscore.

iphone/Classes/KrollContext.m
```diff
@@ -1407,6 +1422,9 @@ - (void)main
     [delegate performSelector:@selector(willStopNewContext:) withObject:self];
   }
 
+#ifdef USE_JSCORE_FRAMEWORK
+  [JSTimerManager invalidateAllTimers];
```
vijaysingh-axway: It should be -
[timerManager invalidateAllTimers];

iphone/Classes/KrollContext.m
```diff
@@ -1529,3 +1548,80 @@ - (void)dealloc
...
+  NSMutableArray *callbackArgs = [arguments.mutableCopy autorelease];
+  JSValue *callbackFunction = [callbackArgs objectAtIndex:0];
+  [callbackArgs removeObjectAtIndex:0];
+  double interval = [[callbackArgs objectAtIndex:0] toDouble] / 1000;
+  [callbackArgs removeObjectAtIndex:0];
+  NSNumber *timerIdentifier = [NSNumber numberWithUnsignedInteger:[[[[NSUUID alloc] init] autorelease] hash]];
+  NSTimer *timer = [NSTimer scheduledTimerWithTimeInterval:interval
+                                                   repeats:shouldRepeat
+                                                     block:^(NSTimer *_Nonnull timer) {
+                                                       [callbackFunction callWithArguments:callbackArgs];
```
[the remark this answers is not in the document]
janvennemann: I tested with Kroll thread and timers are not properly firing here. I think it's because we are permanently blocking the run loop with our infinite while loop. For now i only made the new timer management available when using jscore *and* running on main thread.

package-lock.json
```diff
@@ -477,7 +477,7 @@
     },
     "babel-helper-is-nodes-equiv": {
       "version": "0.0.1",
-      "resolved": "http://registry.npmjs.org/babel-helper-is-nodes-equiv/-/babel-helper-is-nodes-equiv-0.0.1.tgz",
+      "resolved": "https://registry.npmjs.org/babel-helper-is-nodes-equiv/-/babel-helper-is-nodes-equiv-0.0.1.tgz",
```
vijaysingh-axway: I think this file should not be commit.

iphone/Classes/KrollTimer.m
```diff
@@ -108,7 +108,7 @@ - (void)main
       TiThreadPerformOnMainThread(^{
         [kroll invokeOnThread:self method:@selector(invokeWithCondition:) withObject:invokeCond condition:nil];
       },
-          NO);
```
vijaysingh-axway: @janvennemann Please add KrollTimer.m and KrollTimer.h file's under macro.

iphone/Classes/KrollContext.h
```diff
@@ -175,3 +183,29 @@ KrollContext *GetKrollContext(TiContextRef context);
 @property (nonatomic, readwrite, retain) id invocationArg4;
 
 @end
+
+#if defined(USE_JSCORE_FRAMEWORK) && !defined(TI_USE_KROLL_THREAD)
+
+/**
+ * Handles creating and clearing timers when running with JavaScriptCore
+ */
+@interface JSTimerManager : NSObject
```
hansemannn: Small peak: This may clash with internal API's some day, since it uses the JavaScriptCore framework prefix. To be save, naming it like `TiTimerManager` or `TiJSTimerManager` may help.

iphone/Classes/KrollContext.h
```diff
@@ -175,3 +183,29 @@ KrollContext *GetKrollContext(TiContextRef context);
...
+
+/**
+ * Map of timer identifiers and the underlying native NSTimer.
+ */
+@property (nonatomic, strong) NSMutableDictionary<NSNumber *, NSTimer *> *timers;
```
hansemannn: If timers are supposed to be unique identified, an `NSMutableSet` may be more appropriate.

iphone/Classes/KrollContext.m
```diff
@@ -1529,3 +1544,84 @@ - (void)dealloc
...
+- (instancetype)initInContext:(JSContext *)context
+{
+  self = [super init];
+  if (!self) {
+    return nil;
+  }
+
+  self.timers = [NSMutableDictionary new];
+
+  int (^setInterval)(void) = ^() {
```
hansemannn: Any reason to prefer the C `int` type vs. ObjC `NSInteger`?

iphone/Classes/KrollContext.m
```diff
@@ -1529,3 +1544,84 @@ - (void)dealloc
...
+                                                         [self.timers removeObjectForKey:timerIdentifier];
+                                                       }
+                                                     }];
+  self.timers[timerIdentifier] = timer;
+  return [timerIdentifier intValue];
+}
+
+- (void)clearIntervalWithIdentifier:(int)identifier
+{
+  NSNumber *timerIdentifier = [NSNumber numberWithInteger:identifier];
```
hansemannn: In general: All `[NSNumber numberWith****]` convenience constructors should use the boxed version of it (`@(identifier)`) since Objective C 2.0.

iphone/Classes/KrollContext.h
janvennemann: Since i need to access timers by their id in various places a set is somewhat cumbersome to work with in that use case.

PR #10406, opened 2018-10-29
[TIMOB-26496] Android: Include missing exception message

garymathews:
- Java exceptions do not print the exception message

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26496)


jquick-axway (approved):
CR: Pass

PR #10410, opened 2018-10-30
[7_5_X][TIMOB-26499]: Remove namespace restriction in item templates

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26499

Removes the restriction for only `Ti` or `Titanium` as a namespace to allow third-party modules again in list view item templates.

Tries to construct the class first, catches the exception that is thrown if the class does not exists and only then falls back to the commonjs module loading.

fixes regression introduced in #10131


iphone/Classes/TiViewProxy.m
```diff
@@ -2946,57 +2946,57 @@ - (void)unarchiveFromTemplate:(id)viewTemplate_
...
+      [proxy rememberSelf];
+    }];
+  } @catch (NSException *exception) {
+    // No such class exists, so fall back to trying to load an Alloy widget or CommonJS module
+    // TODO Cache by name? Otherwise it seems to keep trying to load native class repeatedly (and failing) and has to re-eval this below!
+    // TODO eval once and pass in the controller name and props as args?
+    DebugLog(@"[DEBUG] Failed to load native class %@, trying Alloy widget / CommonJS module", viewTemplate.type);
+    NSString *code = [NSString stringWithFormat:@"var result;"
+                                                 "try {"
+                                                 "  var jsModule = require('/alloy/widgets/%@/controllers/widget');"
```
garymathews: Unrelated, but should this be either:
```
require('./alloy/widgets/%@/controllers/widget');
require('alloy/widgets/%@/controllers/widget');
```

garymathews (approved):
CR: **PASS**
FT: **PASS**

##### NOTE: You will need to update the `av.imageview` config `module.xcconfig` so its `CURRENT_MODULE_VERSION` is `1.5.0`. This is an issue with the module.

sgtcoolguy (approved):
LGTM

PR #10412, opened 2018-10-31
Improve resilience of git checkout

whitfin:
This PR improves the reliability of the Jenkins checkout. The timeout has been extended to 5 minutes (due to repository size), and there's a `gc` run before each update. 

@garymathews verified that the `gc` calls appear to help (if nothing else, it's not any worse).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -127,6 +127,7 @@ timestamps {
 				dir('..') {
 					if (fileExists('titanium_mobile.git')) {
 						dir('titanium_mobile.git') {
+							sh 'git gc'
 							sh 'git remote update -p' // update the clone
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
@whitfin The timeout here is in minutes, not seconds. So the timeout was already 30 minutes(!), and this PR would make it 5 hours.

The gc may help, but the git operations intermittently run gc's anyways.

whitfin:
It did help; @garymathews observed builds that wouldn't clone and after the gc, it worked fine. Either way, no harm. 

As for the timeout, most StackOverflow answers appear to be using 100+, which would suggest seconds (surely nobody needs two hours for checkout?). Can't find actual docs though, so no idea!

sgtcoolguy:
Manually merged, bumped timeout back down.

PR #10417, opened 2018-11-01
fix(ios): use native NSTimer to avoid thread deadlocks

janvennemann:
JIRA: https://jira.appcelerator.org/browse/TIMOB-26391

Since we only have JSCore and main thread execution i completely removed the legacy timer implementation and replaced it with native `NSTimer` using the new `KrollTimerManager`


sgtcoolguy (approved):
Given that this passes the unit tests, and that the changes are verified to be equivalent to those found in #10401, I'm merging this.

sgtcoolguy:
@janvennemann So I've merged this to master, and we should have the same iOS timer fixes on `7_4_X`, `7_5_X` and `master` now. However, looking at the comments here: https://jira.appcelerator.org/browse/TIMOB-26391?focusedCommentId=443268&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-443268

I'm guessing there's still a bug in the implementation, and I strongly suspect it's from not using JSManagedValues - so that the callback functions can get GC'd while the timer itself is still valid. I think we need to register the callback JSValues as JSManagedValues until he timer is invalidated/cleared.

iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollTimerManager.m
```diff
@@ -0,0 +1,92 @@
...
+{
+  for (NSNumber *timerIdentifier in self.timers.allKeys) {
+    [self clearIntervalWithIdentifier:timerIdentifier.unsignedIntegerValue];
+  }
+}
+
+- (NSUInteger)setIntervalFromArguments:(NSArray<JSValue *> *)arguments shouldRepeat:(BOOL)shouldRepeat
+{
+  NSMutableArray *callbackArgs = [arguments.mutableCopy autorelease];
+  JSValue *callbackFunction = [callbackArgs objectAtIndex:0];
```
sgtcoolguy: Here I think we need to be registering a dictionary/map of JSManagedValues to retain the callbacks and prevent them from GC until the timer is invalidated/cleared.

PR #10418, opened 2018-11-02
[TIMOB-26513]: fix(ios) call remote notification delegate only if available

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26513

the application:didReceiveRemoteNotification: selector is only available when our built in remote push notification implementation was used. when using third-party services like OneSignal this method is not available and also not required.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
@@ -537,7 +537,9 @@ TI_INLINE void waitForMemoryPanicCleared(); //WARNING: This must never be run on
   if (showInForeground) {
     completionHandler(UNNotificationPresentationOptionBadge | UNNotificationPresentationOptionAlert | UNNotificationPresentationOptionSound);
   } else {
-    [self application:[UIApplication sharedApplication] didReceiveRemoteNotification:notification.request.content.userInfo];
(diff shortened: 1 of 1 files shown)


iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
```diff
@@ -546,14 +548,16 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNot
 - (void)userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNotificationResponse:(UNNotificationResponse *)response
              withCompletionHandler:(void (^)(void))completionHandler
 {
-  if ([[[[[response notification] request] content] userInfo] valueForKey:@"aps"] != nil) {
```
hansemannn: sorry :D

vijaysingh-axway (approved):
CR passed.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
janvennemann: Haha, i just couldn't stand looking at all those brackets :D

hansemannn: Let's dream about a "TiApp.swift"!

PR #10425, opened 2018-11-06
[TIMOB-26523] iOS 11+: Expose UIFontTextStyleLargeTitle for dynamic fonts

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26523


hansemannn:
It seems like one of them returns undefined, although the test case displays all of them properly. Unfortunately I don't have Jenkins permissions anymore, so someone else needs to take a look.

sgtcoolguy:
@hansemannn The test if failing on Android, because these constants are not defined on Android. The test needs to be guarded like so:
{code:javascript}
it.androidMissing('#constants', function () {
{code}

May also want to rename it to `'TEXT_STYLE_* constants'`

iphone/TitaniumKit/TitaniumKit/Sources/API/WebFont.m
```diff
@@ -34,25 +34,10 @@ - (BOOL)isSizeNotSet
...
-      [theStyle isEqualToString:UIFontTextStyleTitle3] ||
-      [theStyle isEqualToString:UIFontTextStyleTitle2] ||
-      [theStyle isEqualToString:UIFontTextStyleTitle1]);
-}
-
 - (UIFont *)font
 {
   if (font == nil) {
-    if (textStyle != nil && [self isValidTextStyle:textStyle]) {
+    if (textStyle != nil && [textStyle isKindOfClass:[NSString class]]) {
```
janvennemann: Any particular reason you removed this check? What will happen if the text style is not one of the valid constants?

janvennemann (commented):
Just a minor note regarding the removed validation check. The rest looks good!

iphone/TitaniumKit/TitaniumKit/Sources/API/WebFont.m
```diff
@@ -34,25 +34,10 @@ - (BOOL)isSizeNotSet
...
-      [theStyle isEqualToString:UIFontTextStyleCallout] ||
-      [theStyle isEqualToString:UIFontTextStyleTitle3] ||
-      [theStyle isEqualToString:UIFontTextStyleTitle2] ||
-      [theStyle isEqualToString:UIFontTextStyleTitle1]);
-}
-
 - (UIFont *)font
 {
   if (font == nil) {
-    if (textStyle != nil && [self isValidTextStyle:textStyle]) {
```
janvennemann: Any particular reason you removed this check? What will happen if the text style is not one of the valid constants?

hansemannn: I wanted to remove the restriction to allow future font-styles to be applied even if the constant is not exposed so far, e.g. a raw-value of `large-title` which can map to `UIFontTextStyleLargeTitle`. It does allow invalid values as well, but that's caught by the `UIFont` convenience constructor anyway, so double-checking this every time does not make much sense. In the end, it's the developers choice what do pass and if it's invalid, he/she/it will be informed.

vijaysingh-axway (approved):
CR passed.

keerthi1032:
FR passed. tests returned large title  style correctly.Works fine 

Test Environment:
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local 8.1 SDK
Cli =7.0.10
Studio =5.1.2.201903111843
Device = iPhone 6s iOS 12
Simulator = iPhone 8 iOS 12,iphone 6 iOS 11

PR #10426, opened 2018-11-06
fix(ios): Fix iOS 8/9 compatibility of timers on main thread, register exception handler

sgtcoolguy:

**Description:**
Extension of #10401 

This combines the best of both Jan and my solutions for fixing up usage of NSTimer for JS timers. Specifically, we use target/select based timers, not block calls. We use an `NSMapTable` to hold timer id -> timer mappings, so that the NSTimer/RunLoop code in iOS can deal with cleaning up timers (specifically single-shot timers). When a timer is invalidated it'll invalidate the new TimerTarget object which holds the JS callback and optional an array of additional arguments from JS - so those references will get GC'd too as a result of normal operation.

Additionally, this PR fixes handling setTimeout/setInterval called with no specified interval time (which should default to 0 by spec).


janvennemann:
I had to refactor the timer implementation anyway for compatibility with iOS 9 and 8 in #10438 and i think that also solved any potential GC issues since i maintain strong references to the callback and arguments now. I only opened it up for the 7_4_X branch for now and i'm waiting for a complete build to pass. The last one failed in the Android tests but iOS seemed ok.

PR #10428, opened 2018-11-06
[7_4_X] fix(ios): Fix iOS 8/9 compatibility of timers on main thread, register exception handler

sgtcoolguy:

**Optional Description:**
Backport of #10426

### DIFF
--- a/iphone/Classes/KrollContext.m
+++ b/iphone/Classes/KrollContext.m
@@ -1622,7 +1622,8 @@ static TiValueRef StringFormatDecimalCallback(TiContextRef jsContext, TiObjectRe
 
 - (void)invalidateAllTimers
 {
-  for (NSNumber *timerIdentifier in self.timers) {
+  NSArray<NSNumber *> *keys = [[self.timers keyEnumerator] allObjects];
+  for (NSNumber *timerIdentifier in keys) {
     [self clearIntervalWithIdentifier:timerIdentifier.unsignedIntegerValue];
   }
 }
--- a/iphone/Classes/KrollContext.m
+++ b/iphone/Classes/KrollContext.m
@@ -1566,6 +1566,40 @@ static TiValueRef StringFormatDecimalCallback(TiContextRef jsContext, TiObjectRe
 
 @end
 
+@implementation KrollTimerTarget
+
+- (instancetype)initWithCallback:(JSValue *)callback arguments:(NSArray<NSValue *> *)arguments
+{
+  self = [super init];
+  if (!self) {
+    return nil;
+  }
+
+  self.callback = callback;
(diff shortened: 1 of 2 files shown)


iphone/Classes/KrollContext.m
```diff
@@ -1617,29 +1628,67 @@ - (void)invalidateAllTimers
...
 - (NSUInteger)setIntervalFromArguments:(NSArray<JSValue *> *)arguments shouldRepeat:(BOOL)shouldRepeat
 {
-  NSMutableArray *callbackArgs = [arguments.mutableCopy autorelease];
+  NSMutableArray<JSValue *> *callbackArgs = [arguments.mutableCopy autorelease];
+  // Retain the callback
   JSValue *callbackFunction = [callbackArgs objectAtIndex:0];
   [callbackArgs removeObjectAtIndex:0];
-  double interval = [[callbackArgs objectAtIndex:0] toDouble] / 1000;
-  [callbackArgs removeObjectAtIndex:0];
+  JSVirtualMachine *vm = callbackFunction.context.virtualMachine;
```
hansemannn: For SDK < 8, this should likely use the Ti prefixed JSCore calls, otherwise this could break apps not using JSCore so far.

iphone/Classes/KrollContext.m
```diff
@@ -1610,36 +1622,54 @@ - (void)dealloc
...
+  NSTimer *timer = [NSTimer timerWithTimeInterval:interval target:self selector:@selector(callJsCallback:) userInfo:callbackArgs repeats:shouldRepeat];
+  [[NSRunLoop mainRunLoop] addTimer:timer forMode:NSRunLoopCommonModes];
+
+  [self.timers setObject:timer forKey:timerIdentifier];
   return [timerIdentifier unsignedIntegerValue];
 }
 
+- (void)callJsCallback:(NSTimer *_Nonnull)timer
+{
+  NSArray<JSValue *> *args = [timer userInfo];
```
sgtcoolguy: I'm not sure if trying to optimize for the most common case of no additional arguments and setting just the callback as the user info and then checking type here would improve performance or not...

iphone/Classes/KrollContext.m
```diff
@@ -1610,36 +1622,54 @@ - (void)dealloc
 
 - (void)invalidateAllTimers
 {
-  for (NSNumber *timerIdentifier in self.timers.allKeys) {
+  for (NSNumber *timerIdentifier in self.timers) {
```
janvennemann: Is it save to iterate directly over the NSMapTable and remove entries?

sgtcoolguy: ah good catch.

PR #10440, opened 2018-11-08
[TIMOB-26461]: iOS Add "safeAreaPadding" property to Ti.UI.Window

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26461


iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
```diff
@@ -166,8 +166,6 @@ - (BOOL)belongsToContext:(id<TiEvaluator>)context
 
 - (void)add:(id)arg
 {
-#if IS_XCODE_9
-#ifdef USE_TI_UIWINDOW
```
janvennemann: Can you explain why these could all be removed?

jquick-axway:
@vijaysingh-axway, I have 4 test projects attached to [TIMOB-26427](https://jira.appcelerator.org/browse/TIMOB-26427) (including screenshots) that can be used to test this. They were partly intended to demonstrate how to use this new feature to Titanium app developers.

jquick-axway:
@vijaysingh-axway, does the returned `safeAreaPadding` compensate for the top status bar? That is, the top padding should be the height of the status bar so that content won't overlap it.

Android's `safeAreaPadding` does this in case the top status bar is set up to be translucent. Since Android is the wild west of operating systems, we can't assume the height of the top status bar like you can on iOS. I remember this being problematic with Amazon's old Kindle Fire tablets which didn't follow Google's rules.

hansemannn:
one more question: does this ensure to return 0 for top/bottom when using native container views (concrete: tab-group and navigation-window)? Those are already layed out to expand their size for the safe-area and should not receive additional padding for their sub-views.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
vijaysingh-axway: IS_XCODE_9 and  USE_TI_UIWINDOW macro is not available at Titanium framework. It was only ensuring to unnecessarily code compilation

vijaysingh-axway:
> @vijaysingh-axway, does the returned `safeAreaPadding` compensate for the top status bar? That is, the top padding should be the height of the status bar so that content won't overlap it.
> 
> Android's `safeAreaPadding` does this in case the top status bar is set up to be translucent. Since Android is the wild west of operating systems, we can't assume the height of the top status bar like you can on iOS. I remember this being problematic with Amazon's old Kindle Fire tablets which didn't follow Google's rules.

Yes 'safeAreaPadding'  compensate for top status bar.

vijaysingh-axway:
> One more question: Does this ensure to return 0 for top/bottom when using native container views (concrete: tab-group and navigation-window)? Those are already layed out to expand their size for the safe-area and should not receive additional padding for their sub-views.

Yes.

janvennemann (approved):
LGTM!

jquick-axway:
@vijaysingh-axway, I've found one other issue.

The top safe-area padding is always zero on iOS 9 and 10 when the status bar is shown.

Is it because iOS' inset API is only available on iOS 11 and higher? If so, then perhaps we need to hardcode the status bar height on older iOS versions. I believe it's always `20dp` tall.

vijaysingh-axway:
> @vijaysingh-axway, I've found one other issue.
> 
> The top safe-area padding is always zero on iOS 9 and 10 when the status bar is shown.
> 
> Is it because iOS' inset API is only available on iOS 11 and higher? If so, then perhaps we need to hardcode the status bar height on older iOS versions. I believe it's always `20dp` tall.

Safe are insets are only available in iOS 11 and above.

PR #10450, opened 2018-11-09
[TIMOB-26539] Android: Fixed Ti.UI.Label default font size to be "dp" unless "font" property is set

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26539

**Summary:**
- This is a breaking-change.
- If `Ti.UI.Label` property "font" is not set, then label should default to `15dp`.
  * Was defaulting to `15` and whatever the "ti.ui.defaultunit" was set to. So, if unit was set to "px", then the label would have been `15px` which would be too hard to read.
  * Now has parity with iOS which defaults to "dp" as well.
- Note that no other Android UI (ie: `Button`, `TextField`, `Picker`, etc.) has this problem. This was only an issue with `Label`.

**Test:**
1. Follow the steps-to-reproduce attached to [TIMOB-26539](https://jira.appcelerator.org/browse/TIMOB-26539).
2. On app launch, verify that the "Hello World" label is no longer tiny. (See ticket's screenshots.)


lokeshchdhry:
FR Passed.

Text are now `15px` when default unit used is `px` & the text appears to display the right size.

SDK Ver:        8.0.0 local build.
OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.8-5
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2

lokeshchdhry:
Waiting for CR to merge.

garymathews (approved):
CR: **PASS**

PR #10451, opened 2018-11-09
fix(android): TIMOB-26541 setTimeout and setInterval do not support calls without interval specified

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26541

**Description:**
* Update timers test suite
* fix Android implementation to support unspecified interval for timer

The update unit test suite really covers the use cases much better now, and specifically tests for this bug. Also note that the iOS implementation was fixed in #10426 (https://github.com/appcelerator/titanium_mobile/pull/10426/files#diff-4c4d01060a96380d965a014b8b0c3186R116)

### DIFF
--- a/android/titanium/src/java/ti/modules/titanium/TitaniumModule.java
+++ b/android/titanium/src/java/ti/modules/titanium/TitaniumModule.java
@@ -1,6 +1,6 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2009-2017 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-Present by Appcelerator, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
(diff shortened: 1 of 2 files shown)


jquick-axway:
Regarding `setInterval()`, according to Mozilla and Node.js, the `delay` parameter is required.
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval
https://nodejs.org/api/timers.html#timers_setinterval_callback_delay_args

However, I tested the below in Node.js and it works. If the `delay` parameter is not provided, then it defaults to 1ms. At least on MacOS (I would expect the min to be 10-16ms on Windows). This appears to be an undocumented feature in Node.js.
```javascript
setInterval(function() {
	console.log("@@@ UnixTime: " + (new Date()).getTime());
});
```

jquick-axway (approved):
CR: Pass

PR #10454, opened 2018-11-10
[TIMOB-26545] fix(android): use correct default value in L()

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26545

### DIFF
--- a/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
+++ b/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
@@ -128,6 +128,10 @@ public class LocaleModule extends KrollModule
 	public String getString(String key, @Kroll.argument(optional = true) String defaultValue)
 	// clang-format on
 	{
+		if (defaultValue == null) {
+			defaultValue = key;
+		}
+
 		try {
 			int resid = TiRHelper.getResource("string." + key.replace(".", "_"));
 			if (resid != 0) {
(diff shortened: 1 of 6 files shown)


android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -128,6 +128,10 @@ public void setLanguage(String language)
 	public String getString(String key, @Kroll.argument(optional = true) String defaultValue)
 	// clang-format on
 	{
+		if (defaultValue == null) {
```
sgtcoolguy: So the only problem here is that I don't believe Android splits the notion of `null` from `undefined` in the type translations to native. So, the test where we explicitly send in null as the second argument we have no way of differentiating that against just not specifying an argument.

I'd argue given the method name and usage, we should probably treat any non-staring second arguments as invalid and just return the key as the default value on all platforms (though technically, say Numbers could be coerced to string, null could be coerced to `'null'`, etc. - but we'd run into the same issue again of Android's no-differentiation of `undefined`/`null`)

sgtcoolguy: I opened a ticket for Android to distinguish between the two here: https://jira.appcelerator.org/browse/TIMOB-26553

hansemannn: I agree. The effort could be used to get all parity issues resolved (including default values). I might have the time to dig more into this, but having the default value for now resolves a release blocker issue for us, so 8.0.0 would be great for this.

PR #10458, opened 2018-11-12
fix(android): Only remove x86 from valid abis if user does not specify abi tag

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26548

We should let a user target x86 if they want to. Just want to call out that the change that caused this bug should have been done in a semver major release.

Test steps: See jira

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1270,13 +1270,15 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 
 	// determine the abis to support
 	this.abis = this.validABIs;
-	if (this.deployType === 'production') {
-		// by default, remove 'x86' from production builds
-		// 'x86' devices are scarce; this is predominantly used for emulators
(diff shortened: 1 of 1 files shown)


garymathews (approved):
CR: **PASS**

sgtcoolguy:
If we want this in 7.5.0 GA, we'll need to cherry-pick back to 7_5_X

lokeshchdhry:
@ewanharris , The build fails. Can you please take a look. Thanks.

lokeshchdhry:
FR Passed.

x86 can be specified as an abi for production builds.

Studio Ver: 5.1.2.201810301430
SDK Ver: 8.0.0 local build
OS Ver: 10.14
Xcode Ver: Xcode 10.1
Appc NPM: 4.2.13
Appc CLI: 7.0.7
Daemon Ver: 1.1.3
Ti CLI Ver: 5.1.1
Alloy Ver: 1.13.2
Node Ver: 8.9.1
NPM Ver: 5.5.1
Java Ver: 10.0.2

PR #10459, opened 2018-11-13
fix(ios): Check for error existence before accessing property

ewanharris:
Fixes TIMOB-26551

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26551

Test steps: see jira


sgtcoolguy (approved):
CR looks fine to me.

PR #10467, opened 2018-11-14
[TIMOB-26537] Android: Module "AndroidManifest.xml" <uses> elements overwrites "tiapp.xml" elements

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26537

**Description:**
Fix overriding tags in manifest when merging modules descriptors.
Since the build script first reads the tags from tiapp.xml I have made it to skip any duplicating tags from modules or tags with the same name. This is not an ideal merge, but should be enough to avoid overriding tags of the application. I don't think there is an universal way to determine which values to be used for tags with clashing names.

**Note:** In this current implementation there is something that concerns me. If we have a node declared in `tiapp.xml` that is also declared in any of the modules the application uses, but with a different value what would be the better way to handle it? If we keep the node from `tiapp.xml` it can break the module that also uses it, because modules would have in their `timodule.xml` files only required nodes. On the other hand if we overwrite the declared in `tiapp.xml` property with the one from the module we go back to the initial problem of this PR. That also applies for different modules depending on some node with different values. I am not sure if we have such a case somewhere, but it may happen I guess.

**Test case:**
Add these lines in tiapp.xml of any project.
```xml
    <manifest>
      <uses-feature android:name="android.hardware.camera.autofocus" android:required="false"/>
      <uses-feature android:name="android.hardware.camera" android:required="false"/>
    </manifest>
  </android>
```
Add the `Ti.Map` module in the project and build it. No need to make use of anything in the module. The result `AndroidManifest.xml` under `build/android/` should contain the defined in tiapp.xml `<uses-feature>` tags together with the `<uses-feature android:glEsVersion="0x00020000" android:required="true"/>` (and without duplicating the latter, since it is defined in ti.playservices)


android/cli/lib/AndroidManifest.js
```diff
@@ -488,9 +488,24 @@ function AndroidManifest(filename) {
...
+									// for identical names to preserve the value from tiapp.xml.
+									if (alreadyAddedItem === tagItem
+										|| alreadyAddedItem.name === tagItem.name) {
+										skip = true;
+									}
+								});
+							}
+							if (!skip) {
+								this[tag].push(tagItem);
+							}
```
jquick-axway: Slight nit-pick. It would be better to use the array's `find()` function instead because you can return out once a duplicate is found and stop iteration.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Something like the below. _(Note that I haven't tested this.)_
```javascript
src[tag].forEach(function (tagItem) {
	let duplicateItem = this[tag].find(function (nextItem) {
		return (nextItem === tagItem) || (nextItem.name === tagItem.name);
	});
	if (!duplicateItem) {
		this[tag].push(tagItem);
	}
});
```

jquick-axway (changes requested):

jquick-axway:
@ypbnv, regarding duplicate "AndroidManifest.xml" elements, let's not worry about that for now... but I'm glad you brought it up.

We plan on updating Titanium to use gradle in the near future. Gradle will merge manifest settings for us for free. We can look into this then. Thanks.
https://developer.android.com/studio/build/manifest-merge

Your changes will favor the "tiapp.xml" manifest settings, which allows the app developer to resolve the conflict on their end. Such as with a conflicting `<uses-feature>` setting where they can determine if its "required" attribute should be set true or not. This will do for now.

ypbnv:
@jquick-axway Updated with the improved duplication detection. Leaving the logic as is. It will be in the developers hand to deal with conflicts until we move to gradle.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

The 2 defined `<uses-feature>` are present along with `<uses-feature android:glEsVersion="0x00020000" android:required="true"/>` & is not duplicated.

OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.8-5
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2

lokeshchdhry:
FR passed with the new changes.

OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.8
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

PR #10478, opened 2018-11-16
[TIMOB-26480] Android: Titanium.UI.ProgressBar 'color' setting portion of progressBar - should be text only

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26480

**Description:**
Fix `color` and `tintColor` properties of ProgressBar for Android. Add `trackTintColor` for parity with iOS. Note that `tintColor` and `trackTintColor` are working on Android API level 21 and above. For API levels 16 to 19 we would need to apply color filters on embedded drawables which IMO is not a good solution because it can result in broken behavior among different vendors and devices that have changed the default behavior.

**Note:** For now I won't be adding unit tests for checking the colors, because again we will need to address the problem with readable words for colors we have in:
https://github.com/appcelerator/titanium_mobile/pull/10393
Once we are clear on what's the best way to approach this I will add them for the properties.

**Test case:**
_app.js_
```js
let win = Ti.UI.createWindow(),
	progressBar = Titanium.UI.createProgressBar({
	top: 0,
	width: Ti.UI.FILL,
	height: 'auto',
	left: 5,
	right: 5,
	min:0,
	max:10,
	value:5,
	tintColor: 'blue',
	color: 'red',
	message: 'Messaage',
	trackTintColor: 'green'
});

win.add(progressBar);
win.open();
```


keerthi1032:
FR Passed. Color setting for progress bar works fine.waiting for CR to merge
Test Environment:
  # CPUs                      = 8
  Memory                      = 17179869184
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.0.0.v20181119102636
Device                            =PIXEL 2 Android 9,samsung s5 android 6
Emulator                        = Nexus 6p android 6

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIProgressBar.java
```diff
@@ -141,4 +151,20 @@ protected void handleSetMessageColor(int color)
 	{
 		label.setTextColor(color);
 	}
+
+	protected void handleSetTintColor(int color)
+	{
+		ColorStateList singleColorStateList = ColorStateList.valueOf(color);
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
garymathews: Minor, it may as well go inside the `if`
```Java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
	final ColorStateList singleColorStateList = ColorStateList.valueOf(color);
	progress.setProgressTintList(singleColorStateList);
}
```

ypbnv: Nice catch!

garymathews (approved):
CR: **PASS**

PR #10481, opened 2018-11-20
fix(ios): avoid multiple dismiss calls of UISearchController from SearchBar delegate

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26521


iphone/Classes/TiUIListView.m
```diff
@@ -945,6 +945,10 @@ - (void)cleanup:(id)unused
   if (_footerViewProxy != nil) {
     [_footerViewProxy windowWillClose];
   }
+
+  if ([searchController isActive]) {
+    [searchController setActive:NO];
```
janvennemann:
```suggestion
    searchController.active = NO;
```
Please use property style accessors whenever possible here and in the other places where you check if the search bar controller is active.

iphone/Classes/TiUIListView.m
```diff
@@ -2023,9 +2027,7 @@ - (void)searchBarTextDidEndEditing:(UISearchBar *)searchBar
   if ([searchBar.text length] == 0) {
     self.searchString = @"";
     [self buildResultsForSearchText];
-    if ([searchController isActive]) {
-      [searchController setActive:NO];
-    }
+    [self performSelector:@selector(dismissSearchController) withObject:nil afterDelay:.2];
```
janvennemann: Why do we need a delay here?

janvennemann (changes requested):

iphone/Classes/TiUIListView.m
janvennemann: Would it make sense to introduce a flag to check for this case? Using a delay may be error prone when the transition takes longer than expected for whatever reason or Apple decides to change the transition duration.

As far as i understand this would also introduce an unnecessary delay if the user just empties the search string and ends editing, wouldn't it?

vijaysingh-axway: I don't think we should introduce a new flag. We are checking [searchController isActive] for active state. This is an edge case in which this issue is raised. Actually there needs only time difference to get value updated. When we end editing, Search controller delegate - (void)willDismissSearchController:(UISearchController *)searchController  first called before
- (void)searchBarTextDidEndEditing:(UISearchBar *)searchBar. 
So it may be possible that there is no need to call
searchController.active = NO; 

ListView's Search functionality needs refactoring, I'll plan in future.

janvennemann: Alright, that sounds reasonable. Let's take the fix as is and you'll schedule a refactor.

janvennemann (approved):

ssjsamir (approved):
**FR Passed:** On searching an item in a search bar and selecting a row, the popover no longer stays with blank white background and instead of disappears. 

Tested using the following test case:

```
var win = Ti.UI.createWindow({backgroundColor: 'white'});

var button = Ti.UI.createButton({title: 'Open Popover!', top : 100});
button.addEventListener('click', function(e){
popover.show({ view: button });
})
win.add(button);
 
var contentWindow = Ti.UI.createWindow({
width : Ti.UI.SIZE,
height : Ti.UI.SIZE,
borderColor : '#CCC'
});
 
var rows = [];
for (var i = 0; i < 4; i++) {
var title = i ? 'Row '+ i : 'Close Window'
rows.push({ properties: { title:title  , backgroundColor: 'red', searchableText:title}});
}
 
var sb = Ti.UI.createSearchBar();
 
var ls = Ti.UI.createListSection({
items: rows
});
var lv = Ti.UI.createListView({
sections: [ls],
searchView: sb,
});
 
lv.addEventListener('itemclick', function(e) {
Ti.API.info('click at index: ' + e.itemIndex);
popover.hide();
});
 
contentWindow.add(lv);
 
var popover = Ti.UI.iPad.createPopover({
width:  400,
height: 220,
backgroundColor: 'green',
contentView: contentWindow
});
win.open();

```
**Test Environment**

```
iPhone 7 Sim (iOS 12.1)
APPC CLI: 7.0.8
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10484, opened 2018-11-20
feat(api): Add cross-platform node-compatible 'path' module

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18582

**Description:**
This is intended to be a JS implementation of a Node-compatible `path` core module.

Included are a suite of tests to verify compatibility.

### DIFF
--- a/android/runtime/common/src/js/kroll.js
+++ b/android/runtime/common/src/js/kroll.js
@@ -6,6 +6,11 @@
  */
 'use strict';
 
+/**
+ * main bootstrapping function
+ * @param  {object} kroll [description]
+ * @return {void}       [description]
+ */
 (function (kroll) { // eslint-disable-line no-unused-expressions
 	var global = this;
 
--- a/common/Resources/ti.internal/extensions/binding.js
+++ b/common/Resources/ti.internal/extensions/binding.js
@@ -1,0 +1,48 @@
+/**
+ * This file is used to hijack the standard require to allow for JS
+ * implementations of "core" modules.
+ *
+ * You add a binding from the "core" module id to the under the hood JS
+ * implementation. We then intercept require calls to handle requests for these modules
+ * and lazily load the file.
+ */
+'use strict';
+
+const bindings = new Map();
+
+// FIXME: Use a cache to avoid the redirection? Maybe just make the map hold either the filepath *or* the cached require result?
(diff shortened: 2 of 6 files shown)


sgtcoolguy:
So, some notes here: I am hacking require to "bind" the implementation in our "bootstrap" files to the module id ""path". it's possible this hack might screw with live view, and it may not work with our require implementation on Windows (I haven't tested that).

I think the pure JS implementation is definitely the right way to go here, just curious whether longer-term we may need a better mechanism for hooking the JS impls in as "core" modules.

jquick-axway:
> just curious whether longer-term we may need a better mechanism for hooking the JS impls in as "core" modules

You don't have to put the "path.js" under the "ti.internal/extensions" folder. You can put it under the root directory next to "ti.main.js" instead. That way no hacks are required. A `require('path')` will just-work. I think this is the simplest solution.

I've also noticed that node.js supports `module.paths`. Are these lookup paths that `require()` method can use kind of like how the `PATH` environment variable works? Perhaps we can implement something like this as well and append our own internal folder path to it. I would prefer the simpler solution above, but this would be an alternative. (When I used to work with Lua scripts, there was a similar feature called [package.path(http://pgl.yoyo.org/luai/i/package.path) and [package.loaders](http://pgl.yoyo.org/luai/i/package.loaders) which I could modify at runtime and add paths that the `require()` function would use.)

**Side Note:**
If you don't know already, it's easier to test your core JS code changes under an appc project instead of under the SDK. I've set up the build on all platforms to copy the SDK's common JS files 1st and then have the appc project's JS files copied 2nd, overwriting any files of the same name (ex: "ti.main.js"). I've made it this way on purpose for easy testing and JS debugging (such as via the Chrome browser). And I get the same build/runtime experience a real app goes through. Once you're done, simply copy the updated JS files to the SDK's common directory. That's how I do it. :)

sgtcoolguy:
@jquick-axway The use of requires like `require('module')` are not intended to be resolved relatively as your solution suggests. They were treated by Titanium in that way in the past (inconsistently), but that leads to name clashes and ambiguity. A require call with a module id and no path separators or '.' is considered a "core" module and should be loaded differently than a require call with a relative or absolute path.

PR #10487, opened 2018-11-22
fix(ios): Default background  color of TiRootViewController in framework using shared configuration

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26557


janvennemann (changes requested):
Please use property accessors whenever possible, e.g. `UIColor.blackColor` or `TiSharedConfig.defaultConfig.defaultBackgroundColor`

vijaysingh-axway:
For colors, I used property accessor. 'defaultConfig' is a static method.  Setter method is written for defaultBackgroundColor because previous properties  set by using setters, so this will be more aligned with other setters.

janvennemann (approved):
Thanks, that's a good trade-off for now. Approved!

ssjsamir (approved):
**FR Passed** Following the instructions mentioned in https://jira.appcelerator.org/browse/TIMOB-26557

**Test Environment** 
```
iPhone 7 Sim (iOS 12.1)
APPC CLI: 7.0.8
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10490, opened 2018-11-26
[TIMOB-26585] Android: Catch all TCP socket read and write exceptions

garymathews:
- Catch all `readySync` and `writeSync` exceptions
- Close socket upon release

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26585)


jquick-axway (approved):
CR: Pass

jquick-axway:
I don't think this is a regression in our `HTTPClient` object. This error will happen when you attempt to send an HTTP request via a **closed** `HTTPClient`, which can happen in an older Titanium version.

Perhaps this is an issue in LiveView or our app restart code?

PR #10375, opened 2018-10-09
[TIMOB-24582] Android: Implement Ti.UI.NavigationWindow

garymathews:
- Implement `Titanium.UI.NavigationWindow`

###### TEST CASE
```JS
var win2 = Titanium.UI.createWindow({
    backgroundColor: 'red',
    title: 'Red Window'
});

var win1 = Titanium.UI.createNavigationWindow({
    window: win2
});

var win3 = Titanium.UI.createWindow({
    backgroundColor: 'yellow',
    title: 'Yellow Window'
});

var win4 = Titanium.UI.createWindow({
    backgroundColor: 'green',
    title: 'Green Window'
});

var button = Titanium.UI.createButton({
    title: 'Open Yellow Window'
});
button.addEventListener('click', function () {
    win1.openWindow(win3);
});

win2.add(button);
var button2 = Titanium.UI.createButton({
    title: 'Open Green Window'
});
button2.addEventListener('click', function () {
    win1.openWindow(win4);
});

win3.add(button2);
win1.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24582)

PR #10503, opened 2018-11-30
chore(lint): Add Android runtime JS to files listed, fix ESLint errors/warnings

sgtcoolguy:
Intended to clean up the internal JS files in Android's runtime and modules, as #10375 is running into listing issues.

I had forgotten to include these files in our listing globs, so they had pre-existing lint errors/warnings.

The review here is not really intended to be about the lint fixes, but about the changes it introduced:
- Removal of the url.js file which appeared to be unused.
- No longer attempting to assign updated values to the Ti.Android.currentActivity property in our JS wrapper code here - as the property is read-only and will throw errors under strict mode (so it should have been silently doing nothing anyways).


sgtcoolguy:
Merged manually

PR #10510, opened 2018-12-04
fix(ios): validate setNeedsUpdateOfHomeIndicatorAutoHidden selector

garymathews:
- iOS 11.0.0 issue where a crash can occur if the selector does not exist
  - Validate the selector exists

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26628)


vijaysingh-axway (approved):
CR passed.

PR #10512, opened 2018-12-05
fix(android): validate KrollObject in KrollProxy

garymathews:
- Validate `getKrollObject` result before use

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26632)


jquick-axway (approved):
CR: Pass

PR #10414, opened 2018-11-01
[TIMOB-25839] Android/iOS: Modified touch events to use default units defined in "tiapp.xml"

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-25839

**Summary:**
- This is a breaking-change.
- Android touch coordinates used to be hard-coded to pixels.
- iOS touch coordinates used to be hard-coded to dips.

**Test:**
_Run the below test on both Android and iOS._
1. Create a classic app using the test code attached to [TIMOB-25839](https://jira.appcelerator.org/browse/TIMOB-25839).
2. Set "tiapp.xml" property "ti.ui.defaultunit" to `dp`.
3. Run the project.
4. Drag your finger within the light-gray rectangle.
5. Verify that a green square is centered under your finger while you drag.
6. Set "tiapp.xml" property "ti.ui.defaultunit" to `px`.
7. Repeat steps 3-5.

PR #10516, opened 2018-12-07
fix(android): Fix touch events coordinates for TableView.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26635

**Description:**
After switching to using the units defined in `tiapp.xml` (https://github.com/appcelerator/titanium_mobile/pull/10414), TableView needed to be brought up to speed with the change when determining which item was clicked from the Touch event coordinates.

Note: I did not find other places in the code where the coordinates are used before firing the event.

**Test case:**
_app.js_
```js
	tableData = [
		{title: 'Apples'},
		{title: 'Bananas'},
		{title: 'Carrots'},
		{title: 'Potatoes'}
	],
	table = Ti.UI.createTableView({
		data: tableData
	});

table.addEventListener('singletap', function (e) {
	Ti.API.info('-----SINGLE TAP TRIGGERED-----');
	Ti.API.info('Index is ' + e.index);
	if (e.row) {
		Ti.API.info('Row title is ' + e.row.title);
	}
});

win.add(table);
win.open();

```
Before this change touch events will be registered with wrong indexes and rows and even below the TableView's items.


ssjsamir (approved):
**FR Passed** singleTap event now returns the correct index. Tested using the test case in the description above. 

**Test Environment**
```
Pixel XL Device (9.0)
APPC CLI: 7.0.8
Node.js Version: 8.9.1
```

jquick-axway (approved):
CR: Pass

PR #10521, opened 2018-12-10
fix(ios) : Convert property ‘value’ to string inside hasText function

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26639


tests/Resources/ti.ui.textfield.addontest.js
```diff
@@ -0,0 +1,48 @@
...
+		win.add(textFieldA);
+
+		var textFieldB = Ti.UI.createTextField({
+			top: '120dip',
+			value: 0
+		});
+
+		win.add(textFieldB);
+
+		textFieldA.addEventListener('change', function () {
```
janvennemann: What's the change event listener for? How will it be triggered in a unit test?

janvennemann (changes requested):
Please also verify the string value in the unit test, e.g. make sure `0` is being converted to `"0"`.

tests/Resources/ti.ui.textfield.addontest.js
vijaysingh-axway: When value get set , at that time also 'change' event listener get called.

janvennemann: This event will only be triggered by user input, not when changing the text programmatically.

vijaysingh-axway: In this particular case when we set number, 'change' event listener is getting called. I guess you are correct, ideally it should not call if it set programatically.
Though I have updated unit test, I'll create a ticket to check the behaviour of 'change' event listener.

vijaysingh-axway: For verifying the conversion of 0 to "0" -
There is no getter method implemented in this to get value from textfield. Whenever we call getter it will simply fetch value from proxy (default getter). So with getter it will always give the same value, which is number. Other places also I can see default getter (value from proxy) returned.

janvennemann (approved):

PR #10524, opened 2018-12-10
fix(ios): avoid leaking objects in hasProperty callback

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26452

**Optional Description:**
This fixes two issues at once:

1. [TIMOB-26452](https://jira.appcelerator.org/browse/TIMOB-26452): Incorrect reporting of false when property's value is null 
2. [TIMOB-26391](https://jira.appcelerator.org/browse/TIMOB-26391): Continued reports about app crashes in 7.5.0 were caused by the introduction of the `hasProperty` callback in #10150 and hanging `JSValueRef`s that were not added to the JS object graph after retrieval. This reworks the whole implementation by checking for all the magic that is happening in `- (id)valueForKey:(NSString *)key` and just returning true or false without accessing the actual value of the property.

@sgtcoolguy This introduces some duplicate code since the checks are now done in `hasProperty` and in `valueForKey`. I have some ideas how to resolve this by isolating the checks and match them with an enum. We would then just check for the enum's value and perform the appropriate actions in both methods. However, i didn't want to refactor too much of the existing code in such a critical point of our core. We might want to pair this with setter and getter removal in 9.0.0 when we need to touch `valueForKey` anyway.


sgtcoolguy (approved):
changes LGTM. I'd really like to see some changes to the test suite to sanity check this.

Ideally just a handful of spot-checks on our proxies that they respond properly to each of these cases. You do a nice job enumerating them in the doc string there.

janvennemann:
@sgtcoolguy added tests to verify the property checks. This revealed some differences between Android and iOS we might want to fix for parity.

tests/Resources/core.runtime.addontest.js
```diff
@@ -0,0 +1,114 @@
...
+var should = require('./utilities/assertions');
+
+describe('Core', () => {
+	describe('Runtime', () => {
+		describe('hasProperty', () => {
+			describe('Top-Module', () => {
+				describe('Submodules', () => {
+					it('should check for sub-module', () => {
+						Ti.should.have.property('UI');
+						// @fixme this check throws an error on android
```
sgtcoolguy: The way to mark this in the test is to do `it.androidBroken('should check for sub-module', () => {`

tests/Resources/core.runtime.addontest.js
```diff
@@ -0,0 +1,114 @@
...
+						if (Ti.Platform.osname !== 'android') {
+							Ti.should.not.have.property('Foo');
+						}
+					});
+				});
+
+				describe('Custom properties', () => {
+					it.ios('should check for custom properties', () => {
```
sgtcoolguy: why limit this to iOS only?

tests/Resources/core.runtime.addontest.js
```diff
@@ -0,0 +1,114 @@
...
+			describe('Proxy', () => {
+				describe('Getter/Setter', () => {
+					it('should check for existing getter method', () => {
+						const tabGroup = Ti.UI.createTabGroup();
+						tabGroup.should.have.property('getTabs');
+					});
+
+					it.ios('should check for dynamic getter method', () => {
+						const view = Ti.UI.createView();
+						view.should.have.property('getFoo');
```
sgtcoolguy: why should this pass? It shouldn't respond true here as there is no "foo" property.

tests/Resources/core.runtime.addontest.js
```diff
@@ -0,0 +1,114 @@
...
+					});
+
+					it('should check for existing setter method', () => {
+						const webView = Ti.UI.createWebView();
+						webView.should.have.property('setHtml');
+					});
+
+					it.ios('should check for dynamic setter method', () => {
+						const view = Ti.UI.createView();
+						view.should.have.property('setFoo');
```
sgtcoolguy: again, why does this pass? This looks like a bug to me, there's no foo property.

tests/Resources/core.runtime.addontest.js
janvennemann: That's iOS getter/setter magic. It first tries to find an actually existing property on the proxy and tries to set that. If not found, it will fallback to set it as a "dynamic property" on the proxy. So everything starting with `set` or `get` is automatically available on iOS.

https://github.com/appcelerator/titanium_mobile/blob/abc04066bd0ec9ac29066ba9a3b46fb489d51044/iphone/TitaniumKit/TitaniumKit/Sources/API/TiProxy.m#L1100

This results in these statements being equivalent on iOS:
```js
view.foo = 'bar;
console.log(view.foo);
```

```js
view.setFoo('bar);
console.log(view.getFoo());
```

tests/Resources/core.runtime.addontest.js
janvennemann: Android throws an error when doing `.not.have.property` checks directly on modules. I'll mark it as `it.androidBroken` together with the other one above.

janvennemann:
@sgtcoolguy marked the two tests as broken on android. For the weird/magical iOS behavior see my other comment.

PR #10525, opened 2018-12-10
fix(ios): avoid leaking objects in hasProperty callback (7_5_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26452

7_5_X backport of #10524


sgtcoolguy (approved):
Same feedback as original change, code looks OK, I'd just prefer we add tests to our suite to verify it.

PR #10527, opened 2018-12-10
feat(api): create node-compatible os core module

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18584

**Description:**
- Adds new API to Ti.Platform to support this:
  - `totalMemory()` (analog to `availableMemory()`)
  - `uptime` property
  - `cpus()` (only on Android)

BREAKING CHANGES: This "fixes" iOS to report `Ti.Platform.availableMemory` in bytes, not megabytes - to achieve cross-platform parity.

Additionally, I modified some of the hacks built into `Ti.Platform.model`. On iOS, it would rename the original `"iPhone 2,1"` style strings to become `"iPhone 3GS"` - but only for a handful of very old models. Additionally, It would rename any simulator as simply `"Simulator"`. I now just pass through he original `"iPhone 2,1"` style strings, but if we know we're on a Simulator I append:
`" (Simulator)"`

This way we can still sniff for Simulator and the iPhone/iPad model it's emulating.

This is used to report iOS cpu information by using a lookup table from model to CPU.

### DIFF
--- a/android/modules/platform/src/java/ti/modules/titanium/platform/PlatformModule.java
+++ b/android/modules/platform/src/java/ti/modules/titanium/platform/PlatformModule.java
@@ -29,10 +29,19 @@ import android.net.wifi.WifiInfo;
 import android.net.wifi.WifiManager;
 import android.os.BatteryManager;
 import android.os.Build;
+import android.os.SystemClock;
 
 import com.appcelerator.aps.APSAnalytics;
 import com.appcelerator.aps.APSAnalyticsMeta;
 
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
 @Kroll.module
@@ -50,6 +59,8 @@ public class PlatformModule extends KrollModule
 	public static final int BATTERY_STATE_FULL = 3;
 
 	protected DisplayCapsProxy displayCaps;
+
+	private List<Processor> processors;
 
 	protected int batteryState;
 	protected double batteryLevel;
@@ -139,6 +150,15 @@ public class PlatformModule extends KrollModule
 	// clang-format on
 	{
 		return Runtime.getRuntime().freeMemory();
+	}
+
+	// clang-format off
+	@Kroll.method
+	@Kroll.getProperty
+	public double getTotalMemory()
+	// clang-format on
+	{
+		return Runtime.getRuntime().totalMemory();
 	}
 
 	// clang-format off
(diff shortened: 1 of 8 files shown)


janvennemann:
@sgtcoolguy

tests/Resources/os.addontest.js
```diff
@@ -0,0 +1,236 @@
...
+		});
+
+		it('returns a String', function () {
+			os.arch().should.be.a.String;
+		});
+
+		// node values: 'arm', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 's390', 's390x', 'x32', and 'x64'.
+		// iOS values: "arm64", "armv7", "x86_64", "i386", "Unknown"
+		// Android values: "armeabi", "armeabi-v7a", "arm64-v8a", "x86", "x86_64", "mips", "mips64", "unknown"
+		// Windows values: "x64", "ia64", "ARM", "x86", "unknown"
```
janvennemann: Why do they all have `unknown` included?

janvennemann (approved):
CR passed! Just some general questions and notes but the code looks good to me.

common/Resources/ti.internal/extensions/os.js
```diff
@@ -0,0 +1,529 @@
...
+			PRIORITY_LOW: 19,
+			PRIORITY_BELOW_NORMAL: 10,
+			PRIORITY_NORMAL: 0,
+			PRIORITY_ABOVE_NORMAL: -7,
+			PRIORITY_HIGH: -14,
+			PRIORITY_HIGHEST: -20
+		}
+	};
+} else if (isAndroid) {
+	OS.cpus = () => Ti.Platform.cpus();
```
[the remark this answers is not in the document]
sgtcoolguy: It's technically supposed to change over time and actually report timings for cpu in various modes - values we could conceivably get on Android (but for now I am stubbing to 0 values)

common/Resources/ti.internal/extensions/process.js
```diff
@@ -5,7 +5,9 @@ const process = {
 	cwd: function () {
 		return __dirname;
 	},
-	platform: Ti.Platform.name
+	// FIXME: Should we try and adopt 'windowsphone'/'windowsstore' to 'win32'?
+	// FIXME: Should we try and adopt 'ipad'/'iphone' to 'darwin'? or 'ios'?
```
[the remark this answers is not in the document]
sgtcoolguy: I think this is such a ridiculously simple API we don't have that we should somewhere. Alloy even has pretend macro/constants for IS_IOS/ISO_ANDROID to solve this.

I'm just not sure *where* it should go. methods on Ti.Platform? just hijack `process.platform`?

tests/Resources/os.addontest.js
sgtcoolguy: I think it's just a fallback in case some never expected exception/usage happens.

This does remind me, that I enumerated the values because I wanted to see if I could/should converge the equivalent arches together to match what node reports.

janvennemann: That might be a good idea. At least for stuff that's the same, e.g. the sim archs:
- `x86`, `i386` => `x32`
- `x86_64` => `x64`
Not sure what's the difference between the individual `arm` archs, though.

janvennemann: Ah, i see you already pushed a commit. If `i386`, `x86` -> `ia32`, what's node's `x32` then?

sgtcoolguy: I modified our return values from Ti.Platform.architecture to match the equivalents from Node (which are all lowercase). And yes, os.arch just returns process.arch. As for the difference between it returning the compiled target versus actual, I don't think that can ever differ in our case. I suppose it could if they're running a 32-bit build of an app on a 64-bit Android device? 

my best read of "x32" is an attempt to make x86/ia32 more sane when dealing with code trying to access 64-bit stuff: https://lwn.net/Articles/456731/

android/modules/platform/src/java/ti/modules/titanium/platform/PlatformModule.java
```diff
@@ -337,6 +357,83 @@ public String getRuntime()
...
+		}
+		final int processorCount = getProcessorCount();
+		processors = new ArrayList<Processor>(processorCount);
+
+		// Now read in their details
+		BufferedReader in = null;
+		List<Map> groups = new ArrayList<Map>();
+		Map<String, String> current = new HashMap<String, String>();
+		try {
+			String[] args = { "/system/bin/cat", "/proc/cpuinfo" };
```
jquick-axway: @sgtcoolguy, are you sure reading Android `/proc/cpuinfo` is reliable? Our APS library used to do this, and because of this, the **majority** of Android devices have been reported to our analytics server with an "unknown" architecture. Please see the following ticket. You can see our test results with different devices too.
https://jira.appcelerator.org/browse/TIMOB-25878

For APS, it could be because it made a bad assumption about a "Processor" field always existing. But regardless, we switched over to using a public Android API for obtaining CPU architecture which is definitely reliable. I see your code is scanning for "processor" (lowercase), so, you might get bit by this too.

PR #10529, opened 2018-12-11
refactor(android): ScrollView now overrides toImage method

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26630

**Description:**
ScrollView now passes its inner layout to the UIHelper in order to allow conversion to image of its
whole content and not only the visible part.

**Test case:**
There is a good code snippet in the JIRA ticket.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
@@ -17,6 +17,7 @@ import org.appcelerator.titanium.proxy.TiViewProxy;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.util.TiRHelper;
 import org.appcelerator.titanium.TiBaseActivity;
+import org.appcelerator.titanium.util.TiUIHelper;
 import org.appcelerator.titanium.view.TiCompositeLayout;
 import org.appcelerator.titanium.view.TiCompositeLayout.LayoutArrangement;
 import org.appcelerator.titanium.view.TiUIView;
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
```diff
@@ -1121,4 +1122,13 @@ public void registerForKeyPress()
 			registerForKeyPress(this.scrollView);
 		}
 	}
+
+	@Override
+	public KrollDict toImage() {
+		if (scrollView instanceof TiVerticalScrollView) {
+			return TiUIHelper.viewToImage(proxy.getProperties(), ((TiVerticalScrollView) scrollView).getLayout());
```
sgtcoolguy: Why not introduce a very simple interface internally here that lists the methods common between the two implementations:
```java
public interface TiScrollView {
  public TiScrollViewLayout getLayout();
}
```

It looks like this code isn't the only place where we're doing `instanceof` checks just to cast and ask for the exact same method signature. I imagine you could clean up the code nicely by introducing it and having the `scrollView` field be declared as a `TiScrollView`.

sgtcoolguy (changes requested):

jquick-axway:
@ypbnv, I did a quick test on iOS and that platform captures the `ScrollView` container (including the border)... and not the whole inner offscreen contents. So, the captured image will be exactly what you see onscreen.

Android and iOS have parity here. Only discrepancy I saw was that Android does not capture the border. (Minor issue, but the customer isn't applying a border so we're fine for now.)

The ticket isn't a bug. I think the person is just asking our advise on how to capture the inner contents.

ypbnv:
@sgtcoolguy Nice proposal. I will see to it.
@jquick-axway In case we have parity with iOS I am closing the PR. Thanks for checking it out!

PR #10539, opened 2018-12-14
[TIMOB-26655] fix(ios): fix application shortcuts parity

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26655

> Please add `no tests`, since this is a UI related change that requires explicit user interaction.


iphone/Classes/TiUIApplicationShortcutsProxy.m
```diff
@@ -141,13 +141,13 @@ - (void)removeDynamicShortcut:(id)identifier
...
 
-  NSMutableArray *shortcuts = (NSMutableArray *)[UIApplication sharedApplication].shortcutItems;
+  NSMutableArray<UIApplicationShortcutItem *> *shortcuts = (NSMutableArray *)[UIApplication sharedApplication].shortcutItems.mutableCopy;
 
-  for (UIApplicationShortcutItem *item in shortcuts) {
-    if ([item.type isEqualToString:key]) {
-      [shortcuts removeObject:item];
+  [shortcuts enumerateObjectsUsingBlock:^(UIApplicationShortcutItem * _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+    if ([obj.type isEqualToString:key]) {
+      [shortcuts removeObject:obj];
```
janvennemann: It's not a good idea to mutate an array while enumerating it. Consider storing the indices that need to be removed in a `NSMutableIndexSet` and then remove them using `removeObjectsAtIndexes`. I know it was already wrong in the first place but since you are touching it here anyway let's make it right ;)

Also any particular reason you switched to enumerating using block instead of `for (... in ...)`. I don't see any benefit but instead it makes it more verbose.

iphone/Classes/TiUIiOSApplicationShortcutsProxy.m
```diff
@@ -141,13 +141,13 @@ - (void)removeDynamicShortcut:(id)itemtype
...
 
-  NSMutableArray *shortcuts = (NSMutableArray *)[UIApplication sharedApplication].shortcutItems;
+  NSMutableArray<UIApplicationShortcutItem *> *shortcuts = (NSMutableArray *)[UIApplication sharedApplication].shortcutItems.mutableCopy;
 
-  for (UIApplicationShortcutItem *item in shortcuts) {
-    if ([item.type isEqualToString:key]) {
-      [shortcuts removeObject:item];
+  [shortcuts enumerateObjectsUsingBlock:^(UIApplicationShortcutItem * _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+    if ([obj.type isEqualToString:key]) {
+      [shortcuts removeObject:obj];
```
janvennemann: Same as above, do not mutate the array while enumerating it.

iphone/Classes/TiUIApplicationShortcutsProxy.m
```diff
@@ -166,12 +166,6 @@ - (void)addDynamicShortcut:(id)args
     return;
   }
 
-  if ([self typeExists:[args objectForKey:@"identifier"]]) {
-    NSLog(@"[ERROR] Ti.UI.ApplicationShortcuts: The identifier for the shortcut %@ already exists. This field must be unique.",
-        [args objectForKey:@"identifier"]);
-    return;
-  }
-
   UIApplicationShortcutItem *shortcut = [[[UIApplicationShortcutItem alloc] initWithType:[args objectForKey:@"identifier"]
```
janvennemann: Does it make sense to rename this to `itemtype` too? Not sure why it's named `identifier` here and `itemtype` in the other place.

janvennemann (changes requested):
Just some notes about the mutating enumerations and param naming. cc @vijaysingh-axway

janvennemann:
@hansemannn please fix the linting errors too, thanks!

iphone/Classes/TiUIApplicationShortcutsProxy.m
hansemannn: SDK 7.5.0 renamed it to `identifier` for the new API on the Ti.UI namespace which uses `identifier` for better parity with Android. I don't exactly remember why the old one wasn't just used here as well.

iphone/Classes/TiUIApplicationShortcutsProxy.m
hansemannn: Damn, you're right with the mutation, I oversaw that! Regarding the `enumerateObjectsUsingBlock `: It's the faster and more ObjC'ish way of iterating. Some inspiring words can be found [here](https://stackoverflow.com/questions/4486622/when-to-use-enumerateobjectsusingblock-vs-for).

hansemannn:
@janvennemann PR updated

vijaysingh-axway:
@hansemannn 
If we are not making 'identifier' unique, there are following issues can arise -
1. In [-(void)see getDynamicShortcut:(id)identifier](https://github.com/appcelerator/titanium_mobile/blob/16b7631a9a01798adf3386efd5fde81706fabd7f/iphone/Classes/TiUIApplicationShortcutsProxy.m#L115), it always compare identifier and return the shortcut. If identifier is not unique we will always get first item.
2. In [-(void)removeDynamicShortcut:(id)identifier](https://github.com/appcelerator/titanium_mobile/blob/16b7631a9a01798adf3386efd5fde81706fabd7f/iphone/Classes/TiUIApplicationShortcutsProxy.m#L129), it will remove all shortcuts with this identifier. There will not be a way to remove a particular shortcut.

I think we should use ‘identifier’ (unique string to identify the shortcut) and ‘itemtype’ (app-defined type). Backward compatibility is also we have to take care .
I am not sure how android use ‘identifier’. Is it simple string to identify the shortcut uniquely or app-defined type as iOS?

hansemannn:
@vijaysingh-axway Going back to this one, can you provide an example on how that API (`identifier` and `itemtype`) would look like? This is the only thing left blocking us from 8.0.0, so I'm optimistic!

hansemannn:
I found a better way around it by actually making the identifiers unique. By using a prefix like `OpenDetails_`, I can use the ID of the model to make them unique + by iterating over that prefix, I can find the shortcuts to remove. 

Still, I would keep the doc-changes as those have been a regression in 7.5+.

hansemannn:
@mukherjee2 @vijaysingh-axway Can this **please** be integrated into 8.0.0? We're patching the SDK with this change since 7.4.1 and it's getting messy to update to newer SDK's without this being included.

Regarding the sensibility of this change: The fix is isolated to one SDK component and does not influence others. Also, it's more of a objc language-specific change than a change to the native layer or proxy. It is always reproducible and can easily be tested. If you need more test plans, we can provide those asap. Thank you.

vijaysingh-axway (approved):
CR passed.

lokeshchdhry:
FR Passed.

Studio Ver:     5.1.2.201812191831
OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.10-master.5
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.7
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2

PR #10540, opened 2018-12-14
fix(android): Fix toImage method for views with border

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26650

**Description:**
`toImage` method now uses `outerView` instead of `nativeView` in order to provide a proper support for
views with border. Previously there were two problems with converting views with border to an image:
 - the image result will be without the border ( I believe with this change we achieve parity for iOS ).
 - you will get a transparent image if you don't attach the view to a parent.

Note: The change is visual, so I don't think we can have a unit test for it.

**Test case:**
_app.js_
```js
var win = Ti.UI.createWindow({ layout: 'vertical'}),
	button = Ti.UI.createButton(),
	view = Ti.UI.createView({
		width: 200,
		height: 200,
		borderRadius: 15,
		backgroundColor: 'blue'
	});

	var newImageView = Ti.UI.createImageView({
		image: view.toImage()
	})
	win.add(newImageView);
});

win.add(button);
win.open();

```


garymathews (approved):
CR: **PASS**

ssjsamir (approved):
**FR Passed:** Able to get views and their borders using the test case provided above. 

**Test Environment**
```
Google pixel xl 9.0 sim
APPC CLI: 7.10.0-master.5
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10542, opened 2018-12-17
[TIMOB-26662] Android: TableViewRow bugs

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26662

This PR fixes following bugs in TableViewRow:

1. `TableViewRow.getRect()` always returns zeroes
2. `TableViewRow` height does not grow beyond original size (but can shrink)
3. NPE in `TableViewRowProxy`

```
java.lang.NullPointerException: Attempt to invoke virtual method 'ti.modules.titanium.ui.widget.TiUITableView ti.modules.titanium.ui.TableViewProxy.getTableView()' on a null object reference
    at ti.modules.titanium.ui.TableViewRowProxy.handleMessage(TableViewRowProxy.java:209)
    at android.os.Handler.dispatchMessage(Handler.java:98)
    at android.os.Looper.loop(Looper.java:154)
    at android.app.ActivityThread.main(ActivityThread.java:6144)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)```

### CODE REVIEW

**Comment by drauggres** [comment]
Related to #4137 ([TIMOB-12658](https://jira.appcelerator.org/browse/TIMOB-12658))

File: android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableViewRowProxyItem.java [hunk 1]
```diff
@@ -649,6 +645,32 @@ public Drawable getSelectorDrawable()
 		return selectorDrawable;
 	}
 
+	public void populateRectDict(KrollDict d)
+	{
+		if (content != null) {
+			int position[] = new int[2];
+			content.getLocationInWindow(position);
+
+			TiDimension nativeWidth = new TiDimension(content.getWidth(), TiDimension.TYPE_WIDTH);
```
sgtcoolguy: probably want to explicitly state that these values are in pixels by calling `setUnits(TypedValue.COMPLEX_UNIT_PX)`. I think this happens to end up being treated as pixels solely by some luck internally.

On that note, it is absurd we do not have a constructor that takes in the unit types on `TiDimension`. We should add one.

sgtcoolguy: @drauggres I've added the TiDimension constructor that accepts the units as the third argument here (on master branch): 
https://github.com/appcelerator/titanium_mobile/commit/84fc4f2b3db74531d40cf959864a9a4850cef943

jquick-axway: > I think this happens to end up being treated as pixels solely by some luck internally.

On Android, the default native coordinates are pixels. This is the assumed units when accessing Google's Java APIs. Our `TiDimension` class was designed to default to pixels as well. So, it's not luck. Although it does this in a rather obtuse manner in this case, that I agree with. I would prefer the units to be explicit set as well.

Honestly, I don't like the design of our Java `TiDimension` class anyways. It's fine for storing coordinates received from JavaScript. But for unit coordinate conversion like this, it's cumbersome to use since it wants to be fed a left/top/right/bottom type which is only relevant to percentage based units. In the future, we should really extract the conversion code to a separate class for convenience.

drauggres: This method is "95% copypaste" of [TiViewProxy.getRect()](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java#L271)

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewRowProxy.java
```diff
@@ -201,9 +217,14 @@ public boolean handleMessage(Message msg)
 			if (tableViewItem != null) {
 				tableViewItem.setRowData(this);
 				// update/refresh table view when a row's data changed.
-				TiUITableView table = getTable().getTableView();
-				table.setModelDirty();
-				table.updateView();
+				TableViewProxy proxy = getTable();
+				if (proxy != null) {
+					TiUITableView table = (TiUITableView) proxy.peekView();
```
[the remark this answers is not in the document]
drauggres: Reason is the crash (see stacktrace in the PR description)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableViewRowProxyItem.java
drauggres: For some reason (I don't know it) TableViewRowProxy does not save its "native view" to `view` property and uses additional property `content`, so `super` method can't calculate "rect" because `view` is `null`.

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewRowProxy.java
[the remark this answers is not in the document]
jquick-axway: Right. The issue is that `getTable()` can return `null`.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableViewRowProxyItem.java
jquick-axway: The Android team already knows that the default units are pixels. I can go either or on this.

Passing the decor view to `TiDimension` is the right move here. It only needs it to fetch the screen DPI in this case. The proxy's actual view is only needed for percentage units, which is never needed here (that was my gripe with the `TiDimension` class up above). The decor view can be `null` if the activity was destroyed. So, the `null` case can theoretically happen when backing out of the app and you fetch the row's rect via the activity's `onDestroy` listener. Although with the changes we've made in Titanium 8.0.0 already, this may no longer be an issue anymore. But it's better to be safe than sorry. I'm fine with the `null` check.

drauggres: I moved part of  `TiViewProxy.getRect()` to `TiViewProxy.getViewRect(View v)` and changed `TableViewRowProxy.getRect()` to use it.

I didn't understand everything above about pixels, dpi, etc, but now all suspicious code again in [one place](https://github.com/appcelerator/titanium_mobile/pull/10542/files#diff-bec22e2efdd8b352a7754ea42f435dc7R281), and you can deal with it regardless of this PR.

drauggres: This discussion is not related anymore for this PR, so I'm closing it as resolved.

drauggres:
Commits are rebased on `master` and reasonably squashed.
FYI: [this](https://github.com/appcelerator/titanium_mobile/pull/10542#issuecomment-449015824) was caused by `/* global Ti */` in addontest-file (that wasn't clear at that time):
> /Users/build/jenkins/workspace/ium-sdk_titanium_mobile_PR-10542/tests/Resources/ti.ui.tableview.addontest.js
>8:11  error  'Ti' is already defined as a built-in global variable  no-redeclare

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -273,30 +273,36 @@ public boolean handleMessage(Message msg)
...
+
+			// TiDimension needs a view to grab the window manager, so we'll just use the decorview of the current window
+			View decorView = TiApplication.getAppRootOrCurrentActivity().getWindow().getDecorView();
+			if (decorView != null) {
+				d.put(TiC.PROPERTY_WIDTH, nativeWidth.getAsDefault(decorView));
+				d.put(TiC.PROPERTY_HEIGHT, nativeHeight.getAsDefault(decorView));
+				d.put(TiC.PROPERTY_X, localLeft.getAsDefault(decorView));
+				d.put(TiC.PROPERTY_Y, localTop.getAsDefault(decorView));
+				d.put(TiC.PROPERTY_X_ABSOLUTE, nativeLeft.getAsDefault(decorView));
+				d.put(TiC.PROPERTY_Y_ABSOLUTE, nativeTop.getAsDefault(decorView));
```
jquick-axway: We actually don't need the DecorView here. We can pass in the `v` variable as an argument to `getAsDefault()` and will get the same results.

I know you didn't make it this way. That's how the old code worked. Now that I look at it, the `TiApplication.getAppRootOrCurrentActivity().getWindow()` part will crash if Developer Option "Don't keep activities" is enabled and the app is backgrounded. The `getAsDefault()` method simply wants a `Context` to fetch `DisplayMetrics` from and we can get that from any `View`.

garymathews (approved):
CR: **PASS**

jquick-axway (approved):
CR: Pass

ssjsamir (commented):
@drauggres When testing this PR against the following test case `TableViewRow` height does not grow beyond the original size but it seems like `TableViewRow.getRect()` return the correct value.
```
win = Ti.UI.createWindow({
backgroundColor: '#AAAAFF',
layout: 'vertical'
});
 
var top = Ti.UI.createView({
backgroundColor: '#FFAAAA',
layout: 'horizontal',
height: Ti.UI.SIZE,
width: Ti.UI.FILL
});
 
var btnPlus = Ti.UI.createButton({title: '+'});
btnPlus.addEventListener('click', function() {
console.log('plus click');
view.height += 50;
});
 
var btnMinus = Ti.UI.createButton({title: '-'});
btnMinus.addEventListener('click', function() {
console.log('minus click');
view.height -= 50;
});
 
var tableView = Ti.UI.createTableView();
 
height: Ti.UI.SIZE,
width: Ti.UI.FILL
});
 
var view = Ti.UI.createView({
height: 150,
backgroundColor: 'blue'
});
 
row.add(view);
tableView.setData([ row ]);
tableView.addEventListener('postlayout', function onPostLayout() {
console.log('postlayout', row.rect.height, view.rect.height);
label.text = [row.rect.height, view.rect.height].join(' \| ');
});
 
top.add(btnPlus);
top.add(btnMinus);
top.add(label);
 
win.add(top);
win.add(tableView);
win.open();
```

**Test Environment**
```
MacOS Catalina 10.15.1 beta
Node.js 10.16.3
"NPM":"4.2.15-1","CLI":"7.1.2"
Google Pixel 7.1.1 Emulator
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableViewRowProxyItem.java
```diff
@@ -552,6 +552,10 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec)
 				} else {
 					h = Math.max(minRowHeight, height.getAsPixels(this));
 				}
+				// Make sure the height is greater than 1 (not 0 since image views default to 1)
```
drauggres: Removing these lines was the fix for "not growing beyond original height" error.
Now they are back and the error too.

Once again, this is the only place in the entire SDK where used assignment to LayoutParams.

garymathews:
@ssjsamir Updated PR

ssjsamir (approved):
FR Passed: Using the test case mentioned above. 

**Test environment*Test Environment
```
MacOS Catalina: 10.15.1 beta
Xcode: 11.3
Java Version: 1.8.0_131
Android NDK: 18.1.5063045
Node.js: 10.16.3
Pixel xl (7.1.1) Emulator
```

PR #10549, opened 2018-12-19
fix(android): Fix localization support for newer API levels.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-13966

**Description:**
Rework the localization system in order to take advantage of the new way of configuring contexts.
Also adds a helper method for TiActivity to allow users to apply changes immediately.

Since from 8.0.0 we are supporting minSDK 19 I have removed the direct assign of `locale` property for configuration, because it was deprecated and is supposed to be replaced by `setLocale()` method.
This may not be the case if we go for a backport of this PR.

Note: I don't think we can unit test this change.

**Test case:**
You should provide translated strings for `hello_world` key in `i18n\en` and `i18n\ar` folders.

Also add `android:supportsRtl="true"` in tiapp.xml for `application` node of Android.

You can see the list's sections change they layout direction as well as the title of the window with the list.

_app.js_
```js
    layout: 'vertical',
    backgroundColor: '#333'
});

    width: 300,
    backgroundColor: 'red',
    text: 'Current Locale: ' + Ti.Locale.currentLanguage + ' (' + L('hello_world') + ')',
    top: 100,
    color: 'black'
});

    title: 'Change'
});

    var newLang = Ti.Locale.currentLanguage  == 'ar' ? 'en' : 'ar';
    Ti.Locale.setLanguage(newLang);
    win.activity.recreate();
    label.text = 'Current Locale: ' + Ti.Locale.currentLanguage + ' (' + L('hello_world') + ')';
});

var listView = Ti.UI.createListView();
var sections = [];

var fruitDataSet = [
];
sections.push(fruitSection);

var vegSection = Ti.UI.createListSection({ headerTitle: 'Vegetables'});
var vegDataSet = [
];
vegSection.setItems(vegDataSet);
sections.push(vegSection);

listView.sections = sections;

var fishSection = Ti.UI.createListSection({ headerTitle: 'Fish'});
var fishDataSet = [
    {properties: { title: 'Cod'}},
    {properties: { title: 'Haddock'}},
];
fishSection.setItems(fishDataSet);
listView.appendSection(fishSection);
win2.add(listView);

var button = Ti.UI.createButton({title: 'Open list'});
    win2.open();
});

win.add(label);
win.add(btn);
win.add(button);
win.open();

```

Sorry for the ugly code snippet!


android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -106,17 +104,8 @@ public void setLanguage(String language)
 				locale = new Locale(parts[0]);
 			}
 
-			Locale.setDefault(locale);
+			TiLocaleManager.setLocale(locale);
 
-			Configuration config = new Configuration();
-			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
sgtcoolguy: Why was the version guarded as Android N here? setLocale was added in api level 17. Just a mistake, or maybe we had a good reason? @garymathews appears to have been the one to add it, so maybe he knows?

android/titanium/src/java/org/appcelerator/titanium/proxy/ActivityProxy.java
```diff
@@ -82,6 +82,14 @@ protected Activity getWrappedActivity()
 		return TiApplication.getInstance().getRootActivity();
 	}
 
+	@Kroll.method
+	public void recreate()
+	{
+		if (activity.get() != null) {
```
sgtcoolguy: there *could* be a timing issue here since you're not assigning the result to a temp variable. It's extremely unlikely, but the reference could become null after this check but before you call it again to call `recreate()` on it.

android/titanium/src/java/org/appcelerator/titanium/util/TiLocaleManager.java
```diff
@@ -0,0 +1,39 @@
...
+		config.setLocale(currentLocale);
+		return ctx.createConfigurationContext(config);
+	}
+
+	public static void setLocale(Locale locale)
+	{
+		Locale.setDefault(locale);
+		// Resorting to the deprecated method updateConfiguration in order to immediately change
+		// the localized strings in the Activity that has triggered the locale change.
+		Context context = TiApplication.getInstance().getBaseContext();
```
sgtcoolguy: Why not retain the old code that did this: `TiApplication.getAppRootOrCurrentActivity().getBaseContext()`?

android/titanium/src/java/org/appcelerator/titanium/util/TiLocaleManager.java
```diff
@@ -0,0 +1,39 @@
...
+		Configuration configuration = new Configuration();
+		configuration.setLocale(locale);
+		context.getResources().updateConfiguration(configuration, context.getResources().getDisplayMetrics());
```
sgtcoolguy: This is deprecated as you state. Do we need to look towards a more comprehensive solution like the ContextWrappers in some of the answers here: https://stackoverflow.com/questions/40221711/android-context-getresources-updateconfiguration-deprecated ?

android/titanium/src/java/org/appcelerator/titanium/proxy/ActivityProxy.java
jquick-axway: Chris is right. There is a race condition here. You should do it like this...
```java
Activity activity = this.activity.get();
if (activity != null) {
	activity.recreate();
}
```

Assigning it to a local hard reference like the above keeps the object alive and makes it safe to use after the null check.

jquick-axway: Also, are you sure calling the `recreate()` method is a good idea? My understanding is that it will destroy the activity and then create a new instance. You would typically only do this if the activity loads its UI from XML. But since we create the UI dynamically via code, all of the displayed views will be lost and the new activity will be blank. Or is our "AndroidManifest.xml" `configChanges` attribute overriding this behavior and it's calling the `Activity.onConigurationChanged()` method instead?

And requiring the app developer to call `recreate()` for every existing activity window would be rather cumbersome too. I'd rather not expose this method in JavaScript and we invoke this method for them automatically for every activity (assuming we need this method at all). This means we create our own event listener handling for this or see if we can leverage the activity `onConfigurationChange()` method somehow.

What about the `ActionBar` title text? By default, the text comes from "strings.xml" and uses the same title resource used to display the app name under the app icon. Things like this tend to only be re-fetched when inflated from XML on Google's end, and that may not happen here. That's partially why I had to create that Java `TiActionBarHandler` class and do it myself.

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
jquick-axway: It's because the `Configuration` class' public `locale` member variable was deprecated in Android N.
https://developer.android.com/reference/android/content/res/Configuration.html#locale

jquick-axway:
I talked to @ypbnv about this.

To make things clear, there is an undocumented breaking-change that happens when you target Android P (aka: API Level 26) where `Context` objects no longer share the same `Configuration` object. This is why language selection broke in Titanium 7.5.0 since it targets Android P by default. The below article provides technical details about this.
https://proandroiddev.com/change-language-programmatically-at-runtime-on-android-5e6bc15c758

jquick-axway (approved):
CR: Pass

Changing language dynamically is broken if you target Android P. This PR solves it and we need a solution as soon as possible. We can worry about using the other deprecated Java API later.

I've verified that this change is "mostly" on par with iOS. The only difference is that iOS now saves the language setting via a plist preference to be restored the next time the app starts up. Android does not currently support this, but the app developer can do this manually.

longton95 (approved):
**FR Passed**

Verified using the test cases above,  localization works on newer API levels, the text switches sides when selecting AR for right-left reading.

**ENV**

```
One Plus 6T (9.0)
Emulator (9.0, 6.0, 4.4.2)
Appc NPM: 4.2.13
Appc CLI: 7.0.9
Ti CLI: 5.1.1
Node: 8.12.0
NPM: 6.4.1
```

jquick-axway:
We have a problem.

This PR solves the bug with `L()`, but now `titleid` and `textid` don't use the selected language/locale anymore.

jquick-axway:
The below is my follow-up PR which resolves the above issue. This changes things so that all strings are loaded from the same context and fixes the inconsistency.
https://github.com/appcelerator/titanium_mobile/pull/10591

PR #10554, opened 2018-12-20
feat: native Promise APIs

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24549

I suggest to reopen [TIMOB-24549](https://jira.appcelerator.org/browse/TIMOB-24549).
This PR introduces an interface for creating asynchronous methods (for native Android modules) that return ES6 Promise.

Sample application:
```javascript
let win = Ti.UI.createWindow({
  title: 'Promise API POC',
  layout: 'vertical',
  backgroundColor: '#fff'
});

let btn1 = Ti.UI.createButton({title: 'Promise'});
btn1.addEventListener('click', (event) => {
  console.log(event.type, event.source.title);
  Ti.Filesystem.requestStoragePermissions()
    .then(function(result) {
        console.log(event.source.title, JSON.stringify(result));
    })
    .catch(function(error) {
        console.error(event.source.title, JSON.stringify(error));
    });
});

let btn2 = Ti.UI.createButton({title: 'async/await'});
btn2.addEventListener('click', onClick);

async function onClick(event) {
  console.log(event.type, event.source.title);
  try {
    let result = await Ti.Filesystem.requestStoragePermissions();
    console.log(event.source.title, JSON.stringify(result));
  } catch (error) {
    console.error(event.source.title, JSON.stringify(error));
  }
}

let btn3 = Ti.UI.createButton({title: 'callback'});
btn3.addEventListener('click', function(e) {
  console.log(e.type, e.source.title);
  Ti.Filesystem.requestStoragePermissions(function(result) {
      if (result && result.success) {
        console.log(e.source.title, JSON.stringify(result));
      } else {
        console.error(e.source.title, JSON.stringify(result));
      }
  })
});

win.add(btn1);
win.add(btn2);
win.add(btn3);

win.open();
```


brentonhouse:
Thanks for the contribution @drauggres !

drauggres:
#### How to use it
Java:
1. Create `void` method on `KrollProxy` and annotate it with `@Kroll.method`
2. Add `KrollPromise` as the **last argument** to this method
3. Call `resolve` or `reject` on `KrollPromise` instance when async result is ready

Generated native method will not be `void`, but will return ES6 Promise and also will pass to java method newly created `KrollPromise` object.

JS:
1. Call the method, receive `Promise` as the result
2. Await for `Promise` fulfillment (call `then`/`catch` or use `async/await`)

sgtcoolguy:
@drauggres I would absolutely love to get this in so we can offer Promise based APIs for any async methods. However, I'm wary of doing so without some tests in our suite for it. We can't test permission requests in our unit test suite, so the changes here aren't something we can test. But maybe if you can modify our WebViewProxy.java to support Promises in place of a callback function?

Or does that not work because it's return type is `Object`? Maybe `GeolocationModule.forwardGeocoder()` and `reverseGeocoder()`?

sgtcoolguy:
Another general idea: This clearly coincides with usage of KrollFunction in the APIs. Maybe there's some way to unify the idea of an optional callback function which if not specified becomes a Promise?

I dunno, just throwing ideas out there. We clearly have some funky legacy APIs that can be run sync or async and may return values in sync mode so the idea of enforcing a void return type for this to happen might not cover all cases.

garymathews:
@sgtcoolguy I agree, similar to the behaviour I have in `ti.es6` https://github.com/appcelerator/ti.es6/blob/master/plugins/ti.es6/1.3.0/Titanium/Geolocation.js#L352

I would prefer we have this backwards compatibility.

jquick-axway:
Instead of implementing promises natively per platform, we can keep the native APIs as-is (still using `KrollFunction`), and instead wrap/override these APIs using our core JavaScript code which is shared on all platforms. That way we can implement promise support on all platforms (Android, iOS, Windows) in 1 shot.
https://github.com/appcelerator/titanium_mobile/tree/master/common/Resources

Something like the below...
```javascript
var oldRequestMethod = Ti.Filesystem.requestStoragePermissions;
Ti.Filesystem.requestStoragePermissions = function(callback) {
	// Don't use a promise if a callback was given. (Legacy behavior.)
	if (callback) {
		oldRequestMethod(callback);
		return;
	}

	// Callback not provided. Use promises solution instead.
	return new Promise(function(resolve, reject) {
		oldRequestMethod(function(result) {
			if (result.success) {
				resolve(result);
			} else {
				reject(result);
			}
		});
	});
};
```

The only issues that are blocking us from doing the above are:
- iOS 9 doesn't support promises if transpile/polyfill is disabled.
- iOS doesn't allow us to overwrite core module JS methods.
- On Android, every required-in JS file gets their own copy of Titanium's core modules. So while we can override JS core module methods on Android, it's not a global change.

drauggres:
@jquick-axway This is not about migrating permissions-requests-API to promises. What I want is to have ability to create native modules with modern API. Of course we can wrap any callbacks into promises, there are already [projects](/appcelerator/titanized) doing this.

@sgtcoolguy 
> Another general idea: This clearly coincides with usage of KrollFunction in the APIs. Maybe there's some way to unify the idea of an optional callback function which if not specified becomes a Promise?

I would prefer to not closely integrate KrollFunction and KrollPromise. It works fine like this. Changes in PermReqAPI allow to use it both ways without breaking anything (sample is in the PR description). Somedays some new APIs will be created and there are no reason for them to be callback-based.

> I dunno, just throwing ideas out there. We clearly have some funky legacy APIs that can be run sync or async and may return values in sync mode so the idea of enforcing a void return type for this to happen might not cover all cases.

Yeah, there are always some legacy, but, once again, this PR is not about changing any current APIs. The second commit here is mostly for example.

> Maybe `GeolocationModule.forwardGeocoder()` and `reverseGeocoder()`?
OK, I'll modify this methods and add tests soon.

I was thinking about creating a module for testing interface generation itself with method/tests like:
- resolve/reject promise with primitive values
- resolve/reject promise with `Object`
- resolve/reject promise with `Array`
- etc.
I don't think it would be right to create it in core, but also it looks like there is no easy way to put an additional module to test suite. Any ideas on this?

sgtcoolguy:
We're planning to do 8.0.0 RC this week, so just looking at our timeline this will definitely need to get bumped to 8.1.0 - updated the milestone for now.

tests/Resources/ti.geolocation.addontest.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe.android('Titanium.Geolocation', function () {
+	// Methods
+
+	it('async #forwardGeocoder()', function (finish) {
+		this.timeout(6e4); // 60 sec
+
+		should(Ti.Geolocation.forwardGeocoder).be.a.Function;
+		Ti.Geolocation.forwardGeocoder('440 N Bernardo Ave, Mountain View')
+			.then(function (data) {
```
sgtcoolguy: 👌🏻
Thanks for the tests!

tests/Resources/ti.geolocation.addontest.js
```diff
@@ -0,0 +1,62 @@
...
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
```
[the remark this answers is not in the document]
drauggres: I had to change this test to fix linter warnings, so I also rewrote it in es6 syntax.

BTW linter does not allow **ES6+** syntax in tests (only es6), so no `async`/`awiat` in tests (but test app works fine). And current `es6.*.test.js`-files from `mocha-suite` do not pass linter check if you'll put them inside `test/Resources`

>you can do arrow functions to shorten the syntax in these `it`/`describe` methods

Actually in this test you can't use arrow for `describe`, because it will break
```javascript
this.timeout(6e4); // 60 sec
```
which after transpilation will become
```javascript
global.timeout(6e4);
```

jquick-axway:
@sgtcoolguy , my only issue with this is that we're passing the `KrollPromise` as an argument when it should be the method's return type. Ideally, the JS method and Java method signatures should match like how it currently works on the iOS side.

sgtcoolguy:
@jquick-axway I'm going to rework this one. The iOS side more or less function as intended. As you noted, the Android side has. mismatch in that we're passing in a promise as like an in/out parameter rather than declaring it as a return type, which feels odd. It "works" in the sense that it pre-creates the native/JS promise before calling in to Java to avoid another bridge traversal. But ultimately here, the code isn't really running fully async. Or at least not explicitly so, and hiding the need to create and schedule the promise from the Java side sort of obscures that you most likely will end up running sync and returning an already finished Promise.

So I'll see if I can't come up with a representation that makes explicit that you need to generate/schedule a promise whose implementation will run "later" and immediately return the Promise. Something akin to Future/Callable/Task/Runnable.

sgtcoolguy:
@jquick-axway OK, I reworked this so that the KrollPromise is the return type - and it's basically a `Runnable` with `reject`/`resolve` methods that will go through JNI to fire it on the JS Promise. When you create an instance of a `V8Promise`, you must override the `public void call()` method and on instantiation we generate a `v8::Promise::Resolver` under the hood and tie the instance to it. We also then post the promise on the runtime looper/handler to be run.

What I'm unsure of is if this is causing a memory/object leak. I'm not sure if/when we may need to call release on the V8Promise to clean it up. Maybe we make `KrollPromise` abstract and have the `reject`/`resolve` methods set some `done` flag and call release under the covers? Will it get collected/cleaned up automatically if the JS engine GC's the Promise? I suspect not. cc @garymathews

hansemannn:
Please please please make this into 9.3.0. Such an important future-topic that could be crossed from our list! :-)

iphone/Classes/GeolocationModule.m
```diff
@@ -494,19 +501,23 @@ - (void)performGeo:(NSString *)direction address:(NSString *)address callback:(G
   [callback start:params];
 }
 
-- (void)reverseGeocoder:(double)latitude longitude:(double)longitude withCallback:(JSValue *)callback
+- (JSValue *)reverseGeocoder:(double)latitude longitude:(double)longitude withCallback:(JSValue *)callback
 {
 #ifndef __clang_analyzer__ // Ignore static analyzer error here, memory will be released. See TIMOB-19444
-  ReverseGeoCallback *rcb = [[ReverseGeoCallback alloc] initWithCallback:callback];
+  KrollPromise *promise = [[KrollPromise alloc] initInContext:[self currentContext]];
```
vijaysingh-axway, line 507: It will leak memory here. I think it should be -
```suggestion
  KrollPromise *promise = [[[KrollPromise alloc] initInContext:[self currentContext]] autorelease];
```

iphone/Classes/GeolocationModule.m
```diff
@@ -494,19 +501,23 @@ - (void)performGeo:(NSString *)direction address:(NSString *)address callback:(G
...
+  return promise.JSValue;
 #endif
 }
 
-- (void)forwardGeocoder:(NSString *)address withCallback:(JSValue *)callback
+- (JSValue *)forwardGeocoder:(NSString *)address withCallback:(JSValue *)callback
 {
 #ifndef __clang_analyzer__ // Ignore static analyzer error here, memory will be released. See TIMOB-19444
-  ForwardGeoCallback *fcb = [[ForwardGeoCallback alloc] initWithCallback:callback];
+  KrollPromise *promise = [[KrollPromise alloc] initInContext:[self currentContext]];
```
vijaysingh-axway, line 517: Memory leak. Same as above.

iphone/Classes/GeolocationModule.m
```diff
@@ -794,18 +805,26 @@ - (void)requestLocationPermissions:(CLAuthorizationStatus)authorizationType with
     [callback.context.virtualMachine addManagedReference:authorizationCallback withOwner:self];
   }
 
+  // Promise to return
+  KrollPromise *promise = [[KrollPromise alloc] initInContext:[self currentContext]];
```
vijaysingh-axway, line 809: Memory leak. Same as above.

sgtcoolguy: It's explicitly retained and assigned to `authorizationPromise` a few lines below this.

vijaysingh-axway: Yes, it should be autoreleased.

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/V8Promise.java
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+
+		return super.handleMessage(message);
+	}
+
+	@Override
+	public void doRelease()
+	{
+		long resolverPointer = getPointer();
```
garymathews, line 74: I don't think this is necessary as [super.doRelease()](https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/V8Object.java#L79) will release the pointer.

###### NOTE: Can be removed from `V8Function` too.

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -271,6 +274,50 @@ Local<Function> TypeConverter::javaObjectToJsFunction(Isolate* isolate, JNIEnv *
...
+		return NULL;
+	}
+	return TypeConverter::jsObjectToJavaPromise(isolate, env, jsObject);
+}
+
+jobject TypeConverter::jsObjectToJavaPromise(Isolate* isolate, JNIEnv *env, Local<Object> jsObject)
+{
+    Local<Promise::Resolver> resolver = jsObject.As<Promise::Resolver>();
+    Persistent<Promise::Resolver, CopyablePersistentTraits<Promise::Resolver>> persistent(isolate, resolver);
+    persistent.MarkIndependent();
```
garymathews, line 290: `MarkIndependent()` has been removed from V8, we only added it back for compatibility with older SDKs. This line can safely be removed, the persistent object won't be GC'ed anyway (unless `SetWeak()` is called)

android/runtime/v8/src/native/V8Promise.cpp
```diff
@@ -0,0 +1,110 @@
...
+	Local<Context> context = V8Runtime::v8_isolate->GetCurrentContext();
+	TryCatch tryCatch(V8Runtime::v8_isolate);
+	MaybeLocal<Promise::Resolver> maybeResolver = v8::Promise::Resolver::New(context);
+	Local<Promise::Resolver> resolver;
+	if (!maybeResolver.ToLocal(&resolver)) {
+ 		titanium::V8Util::fatalException(V8Runtime::v8_isolate, tryCatch);
+ 		return 0;
+ 	}
+    Persistent<Promise::Resolver, CopyablePersistentTraits<Promise::Resolver>> persistent(V8Runtime::v8_isolate, resolver);
+    persistent.MarkIndependent();
```
garymathews, line 40: Remove this line as noted above

garymathews (changes requested):

android/kroll-apt/build.gradle
```diff
@@ -72,6 +72,6 @@ jar {
 
 dependencies {
 	implementation 'com.googlecode.json-simple:json-simple:1.1'
-	implementation 'org.freemarker:freemarker:2.3.14'
+	implementation 'org.freemarker:freemarker:2.3.28'
```
garymathews, line 75: The latest version of `org.freemarker:freemarker` is now `2.3.30`

android/kroll-apt/src/main/resources/org/appcelerator/kroll/annotations/generator/ProxyBinding.fm
```diff
@@ -689,19 +708,18 @@ Outputs the complete statement that performs the JNI call and
 stores the return value in "result". A nested section can
 beprovided to handle the result.
 ---------------------------------------------------------------->
-<#macro callJNIMethod methodArgs hasInvocation returnType methodID jobjectVar argsVar hasResult>
+<#macro callJNIMethod methodArgs hasInvocation isAsync returnType methodID jobjectVar argsVar hasResult>
```
garymathews, line 711: `isAsync` isn't used and can be removed

garymathews:
Not all APIs containing callbacks have been converted to promises. Here's a list of all current APIs that contain callbacks:

[Titanium.Proxy.addEventListener](https://appcelerator.github.io/titanium-docs/api/titanium/proxy.html#addeventlistener)
[Titanium.Proxy.removeEventListener](https://appcelerator.github.io/titanium-docs/api/titanium/proxy.html#removeeventlistener)
[Titanium.UI.View.animate](https://appcelerator.github.io/titanium-docs/api/titanium/ui/view.html#animate)
[Titanium.UI.View.toImage](https://appcelerator.github.io/titanium-docs/api/titanium/ui/view.html#toimage)
[Titanium.Android.Activity.startActivityForResult](https://appcelerator.github.io/titanium-docs/api/titanium/android/activity.html#startactivityforresult)
[Titanium.Android.requestPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/android.html#requestpermissions)
[Titanium.App.iOS.SearchableIndex.addToDefaultSearchableIndex](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/searchableindex.html#addtodefaultsearchableindex)
[Titanium.App.iOS.SearchableIndex.deleteAllSearchableItems](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/searchableindex.html#deleteallsearchableitems)
[Titanium.App.iOS.SearchableIndex.deleteAllSearchableItemByDomainIdenifiers](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/searchableindex.html#deleteallsearchableitembydomainidenifiers)
[Titanium.App.iOS.SearchableIndex.deleteSearchableItemsByIdentifiers](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/searchableindex.html#deletesearchableitemsbyidentifiers)
[Titanium.App.iOS.UserNotificationCenter.getPendingNotifications](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/usernotificationcenter.html#getpendingnotifications)
[Titanium.App.iOS.UserNotificationCenter.getDeliveredNotifications](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/usernotificationcenter.html#getdeliverednotifications)
[Titanium.App.iOS.UserNotificationCenter.requestUserNotificationSettings](https://appcelerator.github.io/titanium-docs/api/titanium/app/ios/usernotificationcenter.html#requestusernotificationsettings)
[Titanium.Calendar.requestCalendarPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/calendar.html#requestcalendarpermissions)
[Titanium.Calendar.requestEventsAuthorization](https://appcelerator.github.io/titanium-docs/api/titanium/calendar.html#requesteventsauthorization)
[Titanium.Contacts.requestContactsPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/contacts.html#requestcontactspermissions)
[Titanium.Contacts.requestAuthorization](https://appcelerator.github.io/titanium-docs/api/titanium/contacts.html#requestauthorization)
[Titanium.Database.DB.executeAsync](https://appcelerator.github.io/titanium-docs/api/titanium/database/db.html#executeasync)
[Titanium.Database.DB.executeAllAsync](https://appcelerator.github.io/titanium-docs/api/titanium/database/db.html#executeallasync)
[Titanium.Filesystem.requestStoragePermissions](https://appcelerator.github.io/titanium-docs/api/titanium/filesystem.html#requeststoragepermissions)
[Titanium.Geolocation.forwardGeocoder](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#forwardgeocoder)
[Titanium.Geolocation.getCurrentHeading](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#getcurrentheading)
[Titanium.Geolocation.getCurrentPosition](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#getcurrentposition)
[Titanium.Geolocation.requestLocationPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#requestlocationpermissions)
[Titanium.Geolocation.requestTemporaryFullAccuracyAuthorization](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#requesttemporaryfullaccuracyauthorization)
[Titanium.Geolocation.reverseGeocoder](https://appcelerator.github.io/titanium-docs/api/titanium/geolocation.html#reversegeocoder)
[Titanium.Media.Android.scanMediaFiles](https://appcelerator.github.io/titanium-docs/api/titanium/media/android.html#scanmediafiles)
[Titanium.Media.requestMusicLibraryPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestmusiclibrarypermissions)
[Titanium.Media.requestCameraPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestcamerapermissions)
[Titanium.Media.requestPhotoGalleryPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestphotogallerypermissions)
[Titanium.Media.requestCameraAccess](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestcameraaccess)
[Titanium.Media.takeScreenshot](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#takescreenshot)
[Titanium.Media.requestAuthorization](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestauthorization)
[Titanium.Media.requestAudioPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestaudiopermissions)
[Titanium.Media.requestAudioRecorderPermissions](https://appcelerator.github.io/titanium-docs/api/titanium/media.html#requestaudiorecorderpermissions)
[Titanium.Media.VideoPlayer.requestThumbnailImagesAtTimes](https://appcelerator.github.io/titanium-docs/api/titanium/media/videoplayer.html#requestthumbnailimagesattimes)
[Titanium.Network.BonjourService.publish](https://appcelerator.github.io/titanium-docs/api/titanium/network/bonjourservice.html#publish)
[Titanium.Network.BonjourService.resolve](https://appcelerator.github.io/titanium-docs/api/titanium/network/bonjourservice.html#resolve)
[Titanium.Network.BonjourService.stop](https://appcelerator.github.io/titanium-docs/api/titanium/network/bonjourservice.html#stop)
[Titanium.Platform.openURL](https://appcelerator.github.io/titanium-docs/api/titanium/platform.html#openurl)
[Titanium.UI.WebView.evalJS](https://appcelerator.github.io/titanium-docs/api/titanium/ui/webview.html#evaljs)
[Titanium.UI.WebView.takeSnapshot](https://appcelerator.github.io/titanium-docs/api/titanium/ui/webview.html#takesnapshot)
[Titanium.UI.WebView.createPDF](https://appcelerator.github.io/titanium-docs/api/titanium/ui/webview.html#createpdf)
[Titanium.UI.WebView.createWebArchive](https://appcelerator.github.io/titanium-docs/api/titanium/ui/webview.html#createwebarchive)
[Titanium.UI.WebView.findString](https://appcelerator.github.io/titanium-docs/api/titanium/ui/webview.html#findstring)

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/V8Promise.java
sgtcoolguy: I think the issue here was that the `nativeRelease` I needed to call was the custom one for V8Promise because we keep the underlying Resolvers in a separate map. `V8Object`'s nativeRelease is defined as static and private, so I couldn't override it in the subclass.

So, I suppose I could eliminate this and just override the `nativeRelease` if we make it non-static and protected in `V8Object`. Same goes for `V8Function`.

sgtcoolguy:
@garymathews This PR was really intended to enable native promises and modified some APIs as a proof of concept. But yeah we should look to enable returning promises for all async APIs we can once this is in.

garymathews (approved):
CR: **PASS**

vijaysingh-axway (approved):

jquick-axway:
@sgtcoolguy, regarding our direct `V8Promise` usage, normally we would use the `Kroll*` prefixed classes instead. I doubt we would ever switch JS engines to take advantage of this design, but we may want to change things over to use `KrollPromise` instead for consistency.

Java 8 supports static methods within an interface. So, it could be set up to be used like this...
```java
KrollPromise promise = KrollPromise.create((promise) -> {
	if (successful) {
		promise.resolve("I did it.");
	} else {
		promise.reject("I failed.");
	}
});
return promise;
```

And ideally you should be able to use it without a callback. This may make it easier to integrate into other APIs.
_(Do we even need the callback? It would be easier to integrate into our existing code without it.)_
```java
KrollPromise promise = KrollPromise.create();
if (successful) {
	promise.resolve("I did it.");
} else {
	promise.reject("I failed.");
}
return promise;
```

We could implement `KrollPromise` as shown below. The key thing here is to create the `V8Promise` via an abstract `KrollRuntime.createPromise()` method that the `V8Runtime` derived class can implement.
```java
public interface KrollPromise {
	interface OnExecuteCallback {
		void onExecute(KrollPromise promise);
	}

	void resolve(Object value);
	void reject(Object value);

	static KrollPromise create() {
		return create(null);
	}

	static KrollPromise create(OnExecuteCallback callback) {
		KrollPromise promise = KrollRuntime.getInstance().createPromise();
		if (promise == null) {
			promise = new MockPromise();
		}
		if (callback != null) {
			TiMessenger.postOnRuntime(() -> {
				callback.onExecute(promise);
			});
		}
		return promise;
	}

	private static class MockPromise implements KrollPromise {
		@Override
		public void resolve(Object value) {}
		@Override
		public void reject(Object value) {}
	}
}

public abstract class KrollRuntime {
	public abstract KrollPromise createPromise();
}

public class V8Runtime extends KrollRuntime {
	@Override
	public KrollPromise createPromise() {
		if (KrollRuntime.isDisposed()) {
			return null;
		}
		return new V8Promsie();
	}
}
```

I'm not trying to cause trouble. Just tossing in some ideas here.

sgtcoolguy:
@jquick-axway re: the callback - yes, it's definitely needed, otherwise there's no way to ensure the promise "body" is going to be async. If we simply create a promise and keep on executing, we're running the code sync but pretending it's async - the promise will already have rejected/resolved by the time the method returns it.

I do see your point that we have typically hid the JS-engine specific subclass vs the `Kroll` abstract type so we could swap engines, though as you mentioned it's highly unlikely we're swapping engines again at this point (this is a relic from then they swapped from rhino to v8). I'll see if I can hide the implementation class as you suggest.

sgtcoolguy:
For reference, I've opened https://jira.appcelerator.org/browse/TIMOB-28279 to track enabling more of our APIs to return Promises.

PR #10555, opened 2018-12-20
iOS: Application crashes on launch on iOS9

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26672

**Description:**
Here I introduce usage of gulp, gulp-changed, and gulp-babel to transpile the `common` JS code to the iOS and Android targets using babel-preset-env.

This way the code is already transpiled in the distributed SDK so even if the user opts out of transpiration it *should* work. (note that I'm not explicitly polyfilling, so maybe it'll fall over not he Map usage there?)


build/packager.js
```diff
@@ -257,12 +263,46 @@ Packager.prototype.package = function (next) {
...
+			const minSupportedIosSdk = version.parseMin(require('../iphone/package.json').vendorDependencies['ios sdk']);
+			gulp.src(this.srcDir + '/common/**/*.js')
+				.pipe(changed(destDir))
+				.pipe(gulpBabel({
+					presets: [
+						[ '@babel/env', {
+							targets: {
+								// TODO: filter to only targets relevant for platforms we're building?
+								ios: minSupportedIosSdk,
+								chrome: chromeVersion
```
ewanharris: Do we maybe want to add safariVersion for Windows too? Just incase iOS moves past that? It should be `safariVersion: '10'` based off the [current code](https://github.com/appcelerator/titanium_mobile_windows/blob/2ae2ef9c6e601181356a0ffc9e43ab1b6b851a71/cli/commands/_build/copy.js#L629)

ewanharris (approved):
CR/FR and the changes look good to me. Just one comment which really isn't an issue right now

I built locally and verified that I can now launch apps on iOS 9, Map appears to not be an issue on iOS 9, unfortunately kanagax compat-table appears to have dropped this version from the table so it wont be easy for us to know what we can and cannot use in this code.

Unsure on the failure in CI, it's failed twice now but builds fine locally for me 🤷‍♂️

sgtcoolguy:
@ewanharris I modified the PR to not use gulp, which seemed to have a timing issue on clean builds (to just using babel-core directly).
Map *should* work on iOS 9: https://github.com/kangax/compat-table/blob/gh-pages/data-es6.js#L6801 (I think I'm reading it right that basic functionality was in Safari 7.1 and most fo the rest in safari 9)

build/packager.js
sgtcoolguy: my concern here is that e don't extract that version out to the windows package.json right now, it appears to be hard-coded. We should extract it out and then we can reference it from the same place.

hansemannn:
quick question - sorry to interrupt: Will transpiling be enabled by default in SDK 8?

sgtcoolguy:
@hansemannn Yes, transpiling is on by default in SDK 8 - though this PR is about transpiling the JS code we ship in the SDK to extend our own APIs. (basically we transpile down to the minimum JS engines of all the platforms so that even if the user turns off transpilation, the common JS SDK code we ship should still be pre-transpiled and work).

ssjsamir (approved):
FR Pass application no longer crashes on launch when using iOS9.
**Test Environment**
```
iPhone 6s plus (9.3) Sim
APPC CLI: 7.0.0
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10556, opened 2018-12-20
fix(ios): schedule KrollCallback on main thread

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26673

According to the docs, accessing the default `UNNotificationCenter` is thread safe.

> You may use the shared user notification center object simultaneously from any of your app's threads. The object processes requests serially in the order they were initiated.

The docs also point out that the completion handler blocks for methods can be executed on a background thread.

Wrapping the whole `UNNotificationCenter` is therefore unnecessary. Calling the `KrollCallback` in the completion block without scheduling it on the main thread however can cause unwanted side-effects since a lot of JS operations will be rescheduled to the main-thread one by one. This scatters instructions on the run loop leaving them vulnerable to internal JS deadlocks.


vijaysingh-axway:
Changes looks good.
@janvennemann  There are many places we are using TiThreadPerformOnMainThread. I think we should scan those places as well.

janvennemann:
@vijaysingh-axway I was thinking if we could force main thread execution in 
https://github.com/appcelerator/titanium_mobile/blob/33498aeaddac7409a3f411a60455bfc5ddaeabf6/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollCallback.m#L108

We do that already in the async call
https://github.com/appcelerator/titanium_mobile/blob/33498aeaddac7409a3f411a60455bfc5ddaeabf6/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollCallback.m#L103

I'll give it a try and test this out.

janvennemann:
@vijaysingh-axway reworked the PR so every `KrollCallback` automatically re-schedules if not on main-thread. Also removed the unnecessary `TiThreadPerformOnMainThread` around `UNNotificationCenter`.

iphone/Classes/TiUIScrollView.m
```diff
@@ -362,7 +362,6 @@ - (void)setContentHeight_:(id)value
 
 - (void)setRefreshControl_:(id)args
 {
-#ifdef USE_TI_UIREFRESHCONTROL
```
vijaysingh-axway: Why this macro removed?

keerthi1032:
FR Passed. No freezes, counter increases fine and works as expected. Waiting for resolving  merge conflicts to merge.

Operating System
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 
Device            = iPhone 6s iOS 12 
Simulator       =iPhone 6 iOS 11

iphone/Classes/TiUIScrollView.m
janvennemann: Oh, this was removed by accident because changes from master where pushed in and then i reverted back to a previous state. Fixed!

janvennemann:
@keerthi1032 merge conflicts resolved!

vijaysingh-axway (approved):
CR passed.

iphone/Classes/TiUIScrollView.m
vijaysingh-axway: Thanks!

PR #10519, opened 2018-12-08
refactor(android): Refactored app resume, intent-filter, and "newintent" handling

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-26075
- https://jira.appcelerator.org/browse/TIMOB-26636
- https://jira.appcelerator.org/browse/TIMOB-26637
- https://jira.appcelerator.org/browse/TIMOB-26638
- https://jira.appcelerator.org/browse/TIMOB-26645

**Summary:**
- Fixed bug where an intent can create a 2nd app instance stuck at the splash screen if an app instance already exists.
  * Typically happens when app is launch via URLs, tapping notifications, etc.
  * Fixed by immediately destroying 2nd instance, resuming existing activity, and firing a "newintent" event which simulates Android "singleTask" handling.
- Fixed bug where `startActivityForResult()` can create a 2nd app instance stuck at the splash screen if an app instance already exists.
  * Fixed by destroying 1st instance (all windows are closed) and then starting up a new Titanium activity instance.
- [TIMOB-26636](https://jira.appcelerator.org/browse/TIMOB-26636) Removed deprecated "newIntent" event in favor of "newintent".
- [TIMOB-26637](https://jira.appcelerator.org/browse/TIMOB-26637) Added new "rootActivity" property to `Ti.Android`:
  * Improves intent-filter usability since only the root activity will receive the new intent.
  * Note that existing "currentActivity" property always references the top-most activity opened by `Ti.UI.Window`.
- [TIMOB-26638](https://jira.appcelerator.org/browse/TIMOB-26638) Root `Ti.Android.Activity` "intent" property will now be updated when "newintent" has been fired.
  * This "intent" property used to be assigned the intent that originally launched/created the activity and would never change.
  * This is needed in case "newintent" event gets fired before the "app.js" has a chance to set up a listener for it. Can happen when receiving a URL intent during a cold start of the app.
  * Note that `Ti.App.Android` already provides a "launchIntent". So, we're not losing anything.
- [TIMOB-26645](https://jira.appcelerator.org/browse/TIMOB-26645) Calling `finish()` method on root activity should exit app.
  * This makes it easier for Titanium apps launched via `startActivityForResult()` since the app needs to call `setResult()` and `finish()` methods on root activity to return a result.
[...]

PR #10565, opened 2019-01-02
fix(ios): check if app is booted before sending notifications

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26687


jquick-axway:
@hansemannn, in another PR, I'm looking into adding a new `Ti.App` event "started" which gets fired **after** the "app.js" gets executed. This intends to solve the same problem. Please see...
https://github.com/appcelerator/titanium_mobile/pull/10519/files#diff-722bd9f159d5a9feed5efb028c22ab04

I think your code change is fine. I just wanted to let you know that I plan on transitioning the code to use the above solution in the future. It's especially needed when we start adding bootstrap scripts which will display UI before the "app.js" gets executed.

What do you think?

jquick-axway:
@hansemannn, I think we might be talking about 2 different things. Sorry about the confusion.

The `Ti.App` event "started" is a new feature that I'm adding. I only added it because the system needs some kind of clue that the "app.js" has been executed and has had a chance to add its event listeners... because we don't want to fire events until the "app.js" has had a chance to add listeners for them. I added the "started" event to my intent PR because my code change also needed to defer showing a "JSActivity" until after the root activity has been displayed. This is an Android only issue I'm solving and the new "started" event provides a guaranteed means for doing this.

Now here's a potential issue that can happen on iOS in the future. If we did set up a bootstrap script to show UI on startup (let's say to show a dialog), the "app.js" won't be loaded until after that bootstrap UI has been cleared. This means that a "handleurl" event fired while a bootstrap UI is displayed won't be received by the "app.js" listeners since the "app.js" hasn't been loaded yet. That's the potential future issue. For "handleurl", perhaps this behavior is okay, but this behavior might not be okay for notification handling. That's the potential issue that the "started" event solves.

hansemannn:
You mean linting? Can you post the command to lint again? `clang-format -style=file -i <file>` seems to not pickup the .clang-format we've put into `/iphone/` back then.

vijaysingh-axway:
@hansemannn  Yes. Linting :). 
I believe you are running command -
`clang-format -i -style=file TitaniumKit/TitaniumKit/Sources/API/TiApp.m` at location '/iphone' in this case.
I have run the same command and its working. I have fixed listing issue. Thanks!

vijaysingh-axway (approved):
CR passed.

ssjsamir (approved):
**FR Passed** Using a variation of test case provided above. 
```
  color: '#900',
  font: { fontSize:48 },
  shadowColor: '#aaa',
  shadowOffset: {x:5, y:5},
  shadowRadius: 3,
  text: 'A simple label',
  top: 30,
  width: Ti.UI.SIZE, height: Ti.UI.SIZE
});

Ti.App.iOS.addEventListener('handleurl', function (event) {
    Ti.API.warn(event);
    win.add(label1);
});

var win = Ti.UI.createWindow({ backgroundColor: '#fff' });
win.open();

```
**Test Environment**
```
iPhone 6 Sim (12.1)
APPC CLI: 7.0.9
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10566, opened 2019-01-03
feat: enable transpile by default, opt-out option now

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26146

**Description:**
Make transpiling an opt-out by default now.

Relates to #10555


hansemannn:
Another quick question here: Was there a recent change already that finally allows the async / await usage? We use it heavily on the server-side + web-client and would love to see it in Titanium as well.

hansemannn:
@ewanharris Pretty impressive, nice! The next step would be to return a promise for the callback based API's :-).

PR #10567, opened 2019-01-03
feat: inject polyfills to common JS code bundled in SDK

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26729

**Optional Description:**
This injects the core-js/regenerator polyfills into the bundle `common` JS SDK code we ship when we build.
This way we can inject the polyfills once at startup and have the polyfills packaged as part of the SDK.

Pros:
- We can remove polyfilling from node-titanium-sdk based on usage. The full set of possible polyfills are loaded once at the beginning of startup.
- We can use pretty much every JS feature we want in our common/bootstrap code in the SDK
- Should be faster overall to just require/import the polyfills once at the beginning rather than sporadically throughout the code (given enough usage of them. If the user doesn't use them, then it's overkill)

Cons:
- We're technically shipping `@babel/polyfill` (and `core-js`/`regenerator-runtime`) twice in the SDK (once under `node_modules`, once under `common/Resources/node_modules`)
- Requires change to node-titanium-sdk to avoid doing the polyfilling (which would also remove the duplicated packages above)
- With little or no usage of these polyfills in our code or user code, there's no real gain. (and right now we're not using them, so it's based on the user)

Relates to #10555

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -5815,6 +5815,7 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 			series(this, [
 				function processJSFiles(next) {
 					this.logger.info(__('Processing JavaScript files'));
+					const sdkCommonFolder = path.join(this.titaniumSdkPath, 'common', 'Resources');
 
 					async.eachSeries(Object.keys(jsFiles), function (file, next) {
 						setImmediate(function () {
@@ -5853,10 +5854,12 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 										// Read the possibly modified file contents
 										const source = r.contents;
 										// Analyze Ti API usage, possibly also minify/transpile
+										// DO NOT TRANSPILE CODE inside SDK's common folder. It's already transpiled!
+										const transpile = from.startsWith(sdkCommonFolder) ? false : this.transpile;
 										const analyzeOptions = {
 											filename: from,
 											minify: this.minifyJS,
-											transpile: this.transpile,
+											transpile: transpile,
 											sourceMap: this.sourceMaps || this.deployType === 'development',
 											resourcesDir: this.xcodeAppDir,
 											logger: this.logger,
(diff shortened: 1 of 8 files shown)


iphone/cli/commands/_build.js
```diff
@@ -5853,10 +5854,12 @@ iOSBuilder.prototype.copyResources = function copyResources(next) {
 										// Read the possibly modified file contents
 										const source = r.contents;
 										// Analyze Ti API usage, possibly also minify/transpile
+										// DO NOT TRANSPILE CODE inside SDK's common folder. It's already transpiled!
+										const transpile = from.startsWith(sdkCommonFolder) ? false : this.transpile;
 										const analyzeOptions = {
 											filename: from,
 											minify: this.minifyJS,
```
ewanharris: I think we should also toggle off minification for the sdk common dir, while testing this PR it kept choking on the core-js libs

ewanharris (commented):
I took a not really scientific look at the startup difference here for a project that does not use polyfilling currently (movies).

Using an iPhone 6 on iOS 12.1.2, comparing this PR (with the node-titanium-sdk changes in appcelerator/node-titanium-sdk/pull/72) vs SDK 8.1.0.v20190212084015, I launched the app on device 5 times and took an average. I got the following,

This PR: 858ms average 
8.1.0.v20190212084015: 741ms average

I think this PR is the right way to go, but I think _before_ we merge it'd be good to spend some more time seeing what this impact is, and if we can potentially reduce that (bundling the core-js libs for example)

iphone/cli/commands/_build.js
sgtcoolguy: ok, turned off minifying those files as well.

sgtcoolguy:
Some follow-up from Ewan after the rollup/bundling change:

> So here's my latest findings on your polyfill change, master build is 8.1.0.v20190219213249, ran each case 5 times
> Cold start (turn device off and on, wait a couple minutes, launch app)
> Current master average: 1086.8
> With your changes average: 1084.8
> Warm start (device always running, open app >2 minutes after closing previous time)
> Current master average: 805.6
> With your changes average: 836

I think we're close enough performance-wise that it's worth it to merge this in.

PR #10568, opened 2019-01-04
fix(android): Fix proxy release for SearchView attached to a TableView.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26665

**Description:**
SearchView proxy is now released when the TableView proxy is released instead of when the views are
removed from the screen.

**Test case:**
_app.js_
```js
var win1 = Ti.UI.createWindow({ layout: 'vertical'}),
	tableData = [ {title: 'Apples'}, {title: 'Bananas'}, {title: 'Carrots'}, {title: 'Potatoes'} ],
	button1 = Ti.UI.createButton({ title: 'Remove'}),
	button2 = Ti.UI.createButton({ title: 'Add'}),
	searchView = Ti.UI.Android.createSearchView(),
	table = Ti.UI.createTableView({
		height: '80%',
		search: searchView,
		data: tableData
	});

button1.addEventListener('click', function () {
	win1.remove(table);
});

button2.addEventListener('click', function () {
	win1.add(table);
});

win1.addEventListener('open', function () {
	win1.add(button1);
	win1.add(button2);
	win1.add(table);
});

win1.open();

```
Before this PR removing and adding the table view again will result in a crash.

Note: Unit test is on the way.


tests/Resources/ti.ui.tableview.addontest.js
```diff
@@ -0,0 +1,35 @@
...
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2015-Present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
```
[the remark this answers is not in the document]
sgtcoolguy:
```suggestion
var should = require('./utilities/assertions'); // eslint-disable-line no-unused-vars
```

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
```diff
@@ -137,6 +139,13 @@ public void releaseViews()
 	@Override
 	public void release()
 	{
+		// Release search bar proxy if there is one
+		if (getTableView().getNativeView() instanceof RelativeLayout) {
```
sgtcoolguy: Can getTableView() return null?
Also, assign a temp var to hold result of `getTableView().getNativeView()` since you're calling it twice consecutively.

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
```diff
@@ -137,6 +139,13 @@ public void releaseViews()
...
+			((RelativeLayout) getTableView().getNativeView()).removeAllViews();
+			TiViewProxy searchView = (TiViewProxy) (getProperty(TiC.PROPERTY_SEARCH));
```
sgtcoolguy: no need to wrap getProperty in parens.
```suggestion
			TiViewProxy searchView = (TiViewProxy) getProperty(TiC.PROPERTY_SEARCH);
```

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
sgtcoolguy: Isn't all of this redundant with TiUITableView.java.release()? Why do we need to release the search view in both? Doesn't one "control" the flow and call release on the other one?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUITableView.java
```diff
@@ -275,11 +275,11 @@ public void onDestroy(Activity activity)
 	@Override
 	public void release()
 	{
-		// Release search bar if there is one
+		// Release search bar views if there is one
 		if (nativeView instanceof RelativeLayout) {
 			((RelativeLayout) nativeView).removeAllViews();
 			TiViewProxy searchView = (TiViewProxy) (proxy.getProperty(TiC.PROPERTY_SEARCH));
-			searchView.release();
+			searchView.releaseViews();
```
sgtcoolguy: can searchView be null? Or is the layout instanced check an implicit way of "knowing" there's a search view?

tests/Resources/ti.ui.tableview.addontest.js
```diff
@@ -0,0 +1,35 @@
...
+				height: '80%',
+				search: searchView,
+				data: tableData
+			}),
+			win = Ti.UI.createWindow();
+		function removeAndAddTable() {
+			table.removeEventListener('postlayout', removeAndAddTable);
+			win.remove(table);
+			win.add(table);
+			finish();
```
sgtcoolguy: When a function is used as a callback/event listener it will execute "async" and can behave a little strange under unit tests if there's an Error thrown. Typically you'd want to wrap the body of the function where you're doing the work in a `try` block and place the `finish()` call last inside the block - with a `catch(err)` block that calls `finish(err);` to handle error cases.

i.e.
```javascript
function removeAndAddTable() {
  try {
 	table.removeEventListener('postlayout', removeAndAddTable);
 	win.remove(table);
 	win.add(table);
 	finish();
  } catch (err) {
    finish(err);
  }
}
```

Without that I've seen different platforms behave differently - Windows sometimes crashes, the other sometimes just time out the test, leading you to think that there was a timeout/hang (when there wasn't).

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
ypbnv: I think not. It either returns the already available table view or creates one. But maybe checking for having PROPERTY_SEARCH assigned a value would be a better solution. We don't really care if the search view was drawn on the screen or not before releasing the proxy it holds. I am on it.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUITableView.java
ypbnv: Yes, the layout check should only pass if we have a searchView and it is not part of the TableView (Its `searchAsChild` property is set to `false`).

ypbnv:
@sgtcoolguy Updated.

sgtcoolguy (approved):

PR #10572, opened 2019-01-07
chore(TIMOB-26695): only use debug logs in core-sdk development

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26695

Note: This does not require unit- or ui-tests, since it is an internal change only and is only applied to the `Debug` version of the target.


janvennemann:
+1 for the general idea of this. All the module loading debug is sometimes really cluttering the logs and hides the output that i'm actually interested in.

However, this would remove this debugging info completely for end users, right? It might be nice to still display this info just in the case that a file wasn't found. So a user knows where we looked for his required file and can check what's wrong with his required statement (typos or whatever).

hansemannn:
I agree, but even for that case, a reference to the Node.js algorithm could be posted. Since the lookup is incremental, we would need to store the lookup until a successful or failed lookup and discard it if not necessary. I'd really align to Node.js here and as it seems, they only show a trace of the require statement itself.

sgtcoolguy:
We do report when a file is not found, but yeah we don't report the full listing of files we try. The output here is incidental - it's not really intended to be exposed and with my new fixes for require the listing of files/directories we try to load won't be displayed anyways. This only spits out when we try to load an encrypted asset. So, long story short: I think this should get merged in for 8.0.1.

sgtcoolguy (approved):
This change is fine to me. Most of the original problem is actually solved by PR #10640 but this is debug level output that we shouldn't really be spitting out typically.

sgtcoolguy:
Mental note to self: needs to be cherry-picked to `8_0_X`

hansemannn:
Can this be merged in master? We'd love to use this in our app.

ssekhri:
FR Passed. The debug logs related to Node.js paths are not shown in the console.
Environment:
Studio: 5.1.2.201812191831
SDK: 8.1.0 local build
Mac OS: 10.14.3
Appc NPM: 4.2.13
Appc CLI: 7.0.10-master.17
Node: 8.12.0
NPM Ver: 6.4.1
Java Ver: 9.0.4
XCode: 10.1

PR #10577, opened 2019-01-08
fix(android): amend Titanium ListView widget lookup

garymathews:
- Fix ListView widget lookup for Titanium widgets

##### TEST CASE
- See JIRA

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26580)


android/modules/ui/src/js/listview.js
```diff
@@ -83,26 +83,30 @@ exports.bootstrap = function (Titanium) {
...
+		const namespaceIndex = namespace.lastIndexOf('.');
+		const proxyName = namespace.slice(namespaceIndex + 1);
+		const parentNamespace = namespace.substring(0, namespaceIndex);
+		const segments = parentNamespace.split('.');
+		let parentProxy = global;
+		for (let i = 0; i < segments.length; i++) {
+			parentProxy = parentProxy[segments[i]];
+		}
+		const method = parentProxy['create' + proxyName];
+		if (parentProxy && method) {
```
jquick-axway: You're doing the `parentProxy` null check after using it.

Perhaps it should be coded like the below?
```javascript
let method = null;
if (parentProxy) {
	method = parentProxy['create' + proxyName];
}
if (!method) {
	throw new Error('Could not lookup constructor for namespace: "' + namespace + '"');
}
return method;
```

lokeshchdhry:
FR Passed.

No `TypeError: creationFunction is not a function` is seen.

Studio Ver:     5.1.2.201812191831
OS Ver:         10.14
Xcode Ver:      Xcode 10.1
Appc NPM:       4.2.13
Appc CLI:       7.0.9
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.4
Node Ver:       8.9.1
NPM Ver:        5.5.1
Java Ver:       10.0.2
                ⇨ google Pixel (Android 9)

android/modules/ui/src/js/listview.js
garymathews: Good catch!

PR #10603, opened 2019-01-10
fix(ios): allow TiExceptionHandler to obtain source line from encrypted asset

garymathews:
- Use `TiUtils loadAppResource` to load encrypted assets
- Allow source line to be present in exceptions for production builds

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26712)


garymathews:
Verified with a production build app in TestFlight, monitored by Chimera:
```
/app.js:15\nconsole.log(object.test.crash);
                        ^
undefined is not an object (evaluating 'object.test.crash')
    at file:///var/containers/Bundle/Application/F08D0770-9FB7-435E-B827-C10F3548C967/test.app/app.js:15:24)
```
Source line of crash is visible.

PR #10620, opened 2019-01-16
fix(android): Fixed bug where closing root window won't exit app as of 8.0.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26735

**Summary:**
- Regression introduced by PR #10519.
  * Caused by child activity's intent being changed when root activity's intent has received a new intent. Original intent extras were lost, such as the one providing the `exitOnClose` flag.
- Also removed redundant firing of shortcut event. Is already being fired by `TiRootActivity.onNewIntent()` method.

**Test:**
2. Wait for the app to start up.
3. Press the "back" button.
4. Verify that you are exited out of the app. (Should **not** back out to splash screen.)
5. Un-comment the below code's `exitOnClose: false`.
6. Rebuild the app and run on Android.
7. Press "back" button after app launches.
8. Verify that app backs out to the splash screen.

```javascript
//	exitOnClose: false,
});
window.add(Ti.UI.createLabel({ text: "Press back to exit." }));
window.addEventListener("androidback", function() {
	window.close();
});
window.open();
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -1424,19 +1424,6 @@ public abstract class TiBaseActivity extends AppCompatActivity implements TiActi
 		tiApp.setCurrentActivity(this, this);
 		TiApplication.updateActivityTransitionState(false);
 
-		// handle shortcut intents
-		Intent intent = getIntent();
-		String shortcutId =
-			intent.hasExtra(TiC.EVENT_PROPERTY_SHORTCUT) ? intent.getStringExtra(TiC.EVENT_PROPERTY_SHORTCUT) : null;
-		if (shortcutId != null) {
-			KrollModule appModule = TiApplication.getInstance().getModuleByName("App");
-			if (appModule != null) {
-				KrollDict data = new KrollDict();
-				data.put(TiC.PROPERTY_ID, shortcutId);
-				appModule.fireEvent(TiC.EVENT_SHORTCUT_ITEM_CLICK, data);
-			}
-		}
-
 		if (activityProxy != null) {
 			activityProxy.fireEvent(TiC.EVENT_RESUME, null);
 		}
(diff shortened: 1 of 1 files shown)


ssjsamir (approved):
**FR Passed** Root window now closes when pressing the back button, tested using the test case mentioned above. 

**Test Environment**
```
Google pixel xl 7.1.1 sim
APPC CLI: 7.0.9
Node.js Version: 8.9.1
Xcode 10.1
```

**Awaiting CR**

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -1724,7 +1713,11 @@ protected void fireOnDestroy()
 
 	protected boolean shouldFinishRootActivity()
 	{
-		return canFinishRoot && getIntentBoolean(TiC.INTENT_PROPERTY_FINISH_ROOT, false);
+		boolean isIntentRequestingFinishRoot = false;
```
sgtcoolguy: you could check and exit early `if (!TiBaseActivity.canFinishRoot)` to avoid the "work" of checking null/grabbing the boolean.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -1615,8 +1602,10 @@ protected void onDestroy()
 		// If the activity is finishing, remove the windowId and supportHelperId so the window and supportHelper can be released.
 		// If the activity is forced to destroy by Android OS, keep the windowId and supportHelperId so the activity can be recovered.
 		if (isFinishing) {
-			int windowId = getIntentInt(TiC.INTENT_PROPERTY_WINDOW_ID, -1);
-			TiActivityWindows.removeWindow(windowId);
+			if (this.launchIntent != null) {
+				int windowId = this.launchIntent.getIntExtra(TiC.INTENT_PROPERTY_WINDOW_ID, -1);
+				TiActivityWindows.removeWindow(windowId);
```
sgtcoolguy: I was going to tag this that we should guard for invalid indices, but I see that it uses a SparseArray under the hood - so presumably it "handles" invalid indices by doing nothing?

sgtcoolguy (approved):
LGTM

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
jquick-axway: I thought about doing that too, but decided to maintain the old behavior since I've never mucked with our `TiActivityWindows` handling behavior. I can re-evaluate that side of it later.

PR #10634, opened 2019-01-18
fix(android): Ti.App wrongly fires pause/resume events when opening/closing child windows

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26746

**Summary:**
- Now only fires "pause" and "paused" event when app is backgrounded, like iOS.
- Now only fires "resume" and "resumed event when app is put into the foreground (including app startup), like iOS.
- Updated `Ti.App` docs to indicate Android supported these events since 7.5.0.
- We can't use Google's `ProcessLifecycleOwner` Java classes until the Titanium SDK can be built with gradle. This is because we need to use Google's lifecycle compiler JAR to handle its annotations. (Alternatively, we compile with Java 8, but then we could have a potential min Android OS version issue.)

**Test:**
2. Tap the "Open Child" button.
3. Observe log and verify that you do **NOT** see any pause/resume events.
4. Press the "Home" button to put app in the background.
5. Observe log and verify that you **do** see "pause" and "paused" events.
6. Resume the app.
7. Observe log and verify that you **do** see "resume" and "resumed" events.
8. Tap on the "Close Child" button.
9. Observe log and verify that you do **NOT** see any pause/resume events.

```javascript
function appEventHandler(e) {
	Ti.API.info("@@@ '" + e.type + "' event received.");
	if (Ti.App.Android || Ti.App.Windows) {
		Ti.UI.createNotification({
			message: e.type,
		}).show();
	}
}

Ti.App.addEventListener("resume", appEventHandler);
Ti.App.addEventListener("resumed", appEventHandler);
Ti.App.addEventListener("pause", appEventHandler);
Ti.App.addEventListener("paused", appEventHandler);
Ti.App.addEventListener("close", appEventHandler);

var rootWindow = Ti.UI.createWindow({ title: "Parent Window" });
var openButton = Ti.UI.createButton({ title: "Open Child" });
openButton.addEventListener("click", function() {
	var childWindow = Ti.UI.createWindow({ title: "Child Window" });
	var closeButton = Ti.UI.createButton({ title: "Close Child" });
	closeButton.addEventListener("click", function() {
		childWindow.close();
	});
	childWindow.add(closeButton);
	childWindow.open();
});
rootWindow.add(openButton);
rootWindow.open();
```


android/titanium/src/java/org/appcelerator/titanium/TiApplicationLifecycle.java
```diff
@@ -29,19 +30,49 @@ public void onActivityCreated(Activity activity, Bundle savedInstanceState)
 	@Override
 	public void onActivityStarted(Activity activity)
 	{
-		if (activityCount == 0 && tiApp != null && tiApp.isAnalyticsEnabled()) {
-			APSAnalytics.getInstance().sendAppForegroundEvent();
+		// If no activities have been started, then app is going to be put into the foreground.
+		if (this.activityCount == 0) {
+			// Fire Ti.App resume events.
+			KrollModule appModule = this.tiApp.getModuleByName("App");
```
garymathews: Should we also put `appModule` in `TiApplicationLifecycle`, so we don't have to grab it each time?

garymathews (approved):

android/titanium/src/java/org/appcelerator/titanium/TiApplicationLifecycle.java
jquick-axway: I noticed that we add a new "App" module instance every time a new JavaScript runtime gets created (ie: back-out of the app and relaunch).

garymathews (approved):
CR: **PASS**

jquick-axway:
I was verbally told that iOS fires a "pause" event when an alert gets displayed. I'm going to double check. (Not sure if I agree with this behavior, but app devs might be designing around this.)

ssekhri:
The 'resume' and 'resumed' events are fired on app launch, and app foreground. They do not get fired when a window is opened.
The 'pause' and 'paused' events are fired on app background only. They do not get fired on window close.

However there is a slight difference between android and iOS related to events fired during app launch. On iOS only 'resumed' event is fired whereas for android both 'resume' and 'resumed' events are fired.

jquick-axway:
Update PR so that a "resume" event only gets fired after a "pause" event (like iOS).

You will no longer see a "resume" event on app startup... but you will still see a "resumed" event (with a 'd') on app startup.

hansemannn:
Can this be merged soon? I see the PR is approved since a while. Thx!

ssekhri:
FR Passed.
The 'resumed' event is fired on app launch, and 'resume', 'resumed' events are fired on app foreground. They do not get fired when a window is opened.
The 'pause' and 'paused' events are fired on app background only. They do not get fired on window close.
Test Env:
Mac OS 10.14.3
Ti SDK: 8.1.0.v20190320134035
Appc CLI: 7.0.10
Node: 8.12.0
JDK: 9.0.4
Studio: 5.1.2.201903111843
Android Device: Pixel(v9.0), Nexus6 Emulator(v6.0)

PR #10640, opened 2019-01-22
TiAPI: App launch slowed down by 2-5 seconds in SDK 8

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26742

**Description:**

This is a performance fix to improve iOS require behavior to be more in line with Android's.

On Hans' app, this took startup **from 5.5 seconds to load app.js down to ~400-500ms** for me. For reference, Android takes roughly ~217ms.

The general theme here is to:
- Make use of an index.json file to "cheat" so we know which JS/JSON files exist in the user's app. (which is what we do on Android)
- Additionally mark the files as encrypted or not by assigning different values to them in the index.json (thereby eliminating the dumb behavior on iOS of trying encrypted first for everything and falling back to disk, which alone resulted in a halving of the times in my last commit introducing it)
- Rewriting the require process to separate resolution of the path to a file from actually loading the module. This way we can resolve everything and keep a much better cache around with the resolved values.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
@@ -31,6 +31,45 @@ NSString *TitaniumModuleRequireFormat = @"(function(exports){"
 //Defined private method inside TiBindingRunLoop.m (Perhaps to move to .c?)
 void TiBindingRunLoopAnnounceStart(TiBindingRunLoop runLoop);
 
+typedef NS_ENUM(NSInteger, FileStatus) {
+  DoesntExist,
+  ExistsOnDisk,
+  ExistsEncrypted
+};
+
+typedef NS_ENUM(NSInteger, ModuleType) {
+  Native,
+  JS,
+  JSON,
+  NativeJS
+};
+@interface ResolvedModule : NSObject {
+  @public
+  ModuleType type;
+  NSString *path;
+}
+- (id)initWithType:(ModuleType)modType andPath:(NSString *)modPath;
+@end
+
+@implementation ResolvedModule
+- (id)initWithType:(ModuleType)modType andPath:(NSString *)modPath
+{
+  if ((self = [super init])) {
+    type = modType;
+    path = [modPath retain];
+  }
+
+  return self;
+}
+
+- (void)dealloc
+{
+  [path release];
+
+  [super dealloc];
+}
+@end
+
 @implementation TitaniumObject
 
 - (NSDictionary *)modules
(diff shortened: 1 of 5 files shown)


sgtcoolguy:
cc @hansemannn

sgtcoolguy:
no tests as we already have a good suite of tests around require (which caught a bug during the process!)

Note to reviewer(s)/QE: I'd prefer to merge this myself so I can squash it and get the message/body correct.

keerthi1032:
@sgtcoolguy im getting the following error with this PR build and app stuck at the splash screen 
steps:
1.created default classic app 
2.run against iOS device with local SDK 8.0.0.v20190123072053

 DEBUG] :  Loading: /var/containers/Bundle/Application/665A2CF9-ED9D-472F-86B4-80476F7B74E3/K2test.app/ti.main.js, Resource: ti_main_js
[DEBUG] :  Loading: /var/containers/Bundle/Application/665A2CF9-ED9D-472F-86B4-80476F7B74E3/K2test.app/index.json, Resource: index_json
**[ERROR] :  Script Error Couldn't find module: ./ti.internal/extensions/Error for architecture: arm64**
[DEBUG] :  Application booted in 54.376960 ms

sgtcoolguy:
Looks like this is due to encrypted assets being renamed during the build. So the file `ti.internal/extensions/Error.js` is renamed to `ti_internal/extensions/Error_js`, and that's the entry we're stuffing in our require index.json file.
I'll have to modify the encryption stuff to retain the original name as well, since I can't trust doing a reverse replacement will work (`_` to `.`, since file may actually have had underscores in its name originally).

sgtcoolguy:
ok, I rebase all the commits down to one with a proper commit message.

I included the fix for encrypted JS/JSON assets being required (tested on my local device). I also renamed the file we write to be `_index_.json` rather than `index.json` to help avoid clashes with possible file existing in the user app. Lastly, this file is also encrypted when we're encrypting JS/JSON assets.

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -31,6 +31,45 @@
 //Defined private method inside TiBindingRunLoop.m (Perhaps to move to .c?)
 void TiBindingRunLoopAnnounceStart(TiBindingRunLoop runLoop);
 
+typedef NS_ENUM(NSInteger, FileStatus) {
+  DoesntExist,
```
hansemannn: Only code style, but usually in ObjC, the enumeration cases are prefixed with the enumeration itself, e.g. FileStatusDoesntExist.

keerthi1032:
FR Passed. 
Test Environment:
Operating System
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.0.0.v20190124091001
  Studio Version                =5.1.2.201812191857
Device         = iPhone xs max iOS 12
Simulator   =iPhone 6s iOS 11

PR #10645, opened 2019-01-23
feat(android): hide radio buttons OptionDialog

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26793

Create an OptionDialog without the radio buttons:

```JS
	title: 'Click window to test OptionDialog',
	backgroundColor: 'white'
});

var opts = {
	title: 'Select an option',
	options: ['Option 1', 'Option 2', 'Option 3', 'Option 4'],
	buttonNames: ['Cancel']
}

var dialog;
win.addEventListener('click', function() {
	dialog = Ti.UI.createOptionDialog(opts);

	dialog.addEventListener('click', onSelectDialog);
	dialog.addEventListener('cancel', function(e) {
		alert('Dialog canceled! e.cancel = ' + e.cancel + ', e.index = ' + e.index);
	})

	dialog.show();
});

function onSelectDialog(e) {
	console.log(e);
}

	title: "toggle selected index",
	bottom: 5
});
	if (opts.selectedIndex == 0) {
		opts.selectedIndex = -1;
	} else {
		opts.selectedIndex = 0;
	}
});
win.add(btn);
win.open();
```

ti_option


jquick-axway:
I have different thoughts on how this should be handled. _(I've put some thought about this in the past, but haven't done anything about it yet.)_

First, I agree that we should not show a list of radio buttons where none of them are checked. It looks ridiculous. By default, it should be a vertical list of buttons. I'm pretty sure that's how most app developers want this to be handled.

However, the OptionDialog does have a "selectedIndex" property which when set will check one of the radio buttons. In this case, using radio buttons are fine. This is also an Android and Windows only property.

So, how about this? Instead of introducing a new "textOnly" property, we could simply check if the "selectedIndex" property is set. If it is, then use radio buttons. If not, then use normal buttons. What do you think?

CC: @hansemannn

m1ga:
Updated the example and using `selectedIndex`. If it is -1 (unset) it will display normal text options. Makes sense since a radio button should have at least one selection. 
A proper template selection would be nice to add in the future (radio, checkbox, text, images,...) but for now this gives a better parity with iOS in the default case.

hansemannn:
Using this in production now, works really great! Who ever reads this, please do not bump it to post 8.1.0 :).

garymathews (approved):
CR: **PASS**

hansemannn:
P.S.: Actually just moved to the [BottomSheet API](https://github.com/hansemannn/titanium-android-bottom-sheet) since it's even more used in best-practice apps these days.

m1ga:
yes, looks more like the iOS part. But still it is great to remove the circles when they are not needed for most users :smile:

ssjsamir (approved):
**FR Passed** Able to see dialog options with and without radio buttons depending on the toggle of the selected index. Tested with the test case mentioned above. 

**Test Environment**
```
Google pixel xl 7.1.1 sim
APPC CLI: 7.0.10-17
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10649, opened 2019-01-25
fix(ios): auto-select cert and/or profile for ios if only one valid entry

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26767

**Description:**
This is a simple usability enhancement that irked me when testing on my own personal device. I have only one valid non-expired cert and one wildcard profile. Yet I got prompted to select them every time I built to device for iOS. If there's only one possible value for those config options, have it get auto-selected rather than stop and prompt the user to select the only option.


sgtcoolguy:
This should be cherry-picked to 8_0_X for an eventual 8.0.1 release (once 8.0.0 GA is out).

iphone/cli/commands/_build.js
```diff
@@ -1136,6 +1137,7 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
 				relistOnError: true,
 				complete: true,
 				suggest: false,
+				autoSelectOne: true,
```
janvennemann: Is this flag already implemented or where will this be handled?

cb1kenobi: Yeah, `autoSelectOne` has been a property for "select" fields for years.

janvennemann: Ah sweet, thanks for the clarification!

janvennemann (approved):

ssekhri:
FR Passed.
When building a project via cli and there is only a single appropriate certificate it gets auto-selected and not prompted for. Similarly the profile gets auto-selected if there is only a single profile for the selected certificate and app-id.
If there is more than one certificate or more than one profile for the selected certificate and app-id, the user is prompted to select the cert and/or profile.
Env:
Mac OS 10.14.3
Ti SDK: 8.1.0.v20190125091143
Appc CLI: 7.0.10
Node: 8.12.0
JDK: 9.0.4
XCode: 10.1

ssekhri:
@sgtcoolguy Need to cherry-pick to 8_0_X? The ticket is marked with fix version 8.0.1

PR #10650, opened 2019-01-25
fix(android): Fixed bug where app launches without animation when launched with data intent

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26715

**Summary:**
- This issue was introduced as of Titanium 8.0.0 due to the intent refactoring.
- The issue is that the app suddenly appears when resumed. It used to do a transition animation such as a fade-in, slide-in, etc.
- This fix will now show a fade-in effect when receiving a data intent.

**Note:**
This animation bug only happens for data intents. It does not happen when launching the app normally via the app list screen (ie: main launcher intent) or when shown via `startActivityForResult()`.

**Test:**
1. Create a classic Titanium app.
2. Set project name to "IntentTest". (This is `<name/>` in "tiapp.xml".)
3. Set project's "Application Id" to "com.appc.intent.test". (This is `<id>` in "tiapp.xml".)
4. Set up the "app.js" with the below code.
5. Build and run on Android.
6. Tap the device's "Home" button to put the app in the background.
7. Open the Mac "Terminal".
8. CD (change directory) to: `~/Library/Android/sdk/platform-tools`
9. In the terminal, enter the following...
`./adb shell am start -n com.appc.intent.test/.IntenttestActivity -a android.intent.action.VIEW -d https://www.google.com`
10. Verify that the app "fades-in" when resumed. _(This is the fix.)_

```javascript
Ti.Android.rootActivity.addEventListener("newintent", function(e) {
	label.text = "New Intent:\n" + JSON.stringify(e.intent, null, 4);
});

	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
	text: "Launch Intent:\n" + JSON.stringify(Ti.Android.rootActivity.intent, null, 4),
	width: Ti.UI.FILL,
	height: Ti.UI.SIZE,
});
scrollView.add(label);
window.add(scrollView);
window.open();
```


garymathews (approved):
CR: **PASS**

ssekhri:
@jquick-axway, with the fix the fade-in effect does work when app is launched upon data intent. However when i tried it more than once then usually on the third launch or so the app launches with a flicker. Check the attachment. It consists of a screen record of the emulator. The first time it does launch with a fade in effect however the second time you would notice a flicker. 
[ScreenRecord.zip](https://github.com/appcelerator/titanium_mobile/files/2942896/ScreenRecord.zip)

jquick-axway:
@ssekhri, my Pixel XL sometimes does this too. I don't think there is anything we can do about it. Some devices are showing a screenshot of the app when transitioning into the app which I can't prevent. Most devices don't seem to do this.

I think this is as good as it gets.

ssekhri:
FR Passed. The app launches with a fade-in animation when launched based on a data intent.
There was a flicker issue observed which is not easily noticeable and is intermittent and only on few devices. 
Test Env:
Mac OS 10.14.3
Ti SDK: 8.1.0.v20190125154911
Appc CLI: 7.0.10
Node: 8.12.0
JDK: 9.0.4
Studio: 5.1.2.201903111843
Android Device: Pixel(v9.0), Pixel XL(v9.0), Galaxy S5(6.0), Nexus6 Emulator(v6.0)

PR #10652, opened 2019-01-26
[TIMOB-26769] feat(android): add parity for Ti.UI.MenuPopup API

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26769

| iOS | Android | 
|-----|---------|
|iOS|Android|


android/modules/ui/src/java/ti/modules/titanium/ui/MenuPopupProxy.java
```diff
@@ -0,0 +1,123 @@
...
+
+		if (tiView == null) {
+			Log.e(TAG, "Missing required property \"view\"");
+			return;
+		}
+
+		if (!(tiView instanceof TiViewProxy)) {
+			Log.e(TAG, "Property \"view\" must be a Titanium View, currently is " + tiView.getClass().getName());
+			return;
+		}
```
hansemannn: This is currently work in progress, because the tiView always fails through here. If someone know how to extract a view from a `KrollDict`, let me know! I couldn't find a related API that does this so far.

jquick-axway:
I love the idea of having parity for this.

Note that Android already has a [Ti.Android.Menu](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Android.Menu) and [Ti.Android.MenuItem](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Android.MenuItem) types defined already. As well as Alloy `<Menu/>` and `<MenuItem/>` for them. Currently, these are limited to extended the built-in menu in the Android `ActionBar` and `Toolbar`. You can see code examples of this here...
https://github.com/appcelerator/titanium_mobile/pull/9147

Since the existing Android `Menu` and iOS `PopupMenu` APIs don't line-up, I suppose we should keep the existing `Ti.Android.Menu` and `Ti.Android.MenuItem` types as-is. That's fine for `ActionBar`. Although the parity difference with the `Toolbar` kind of bugs me. Displaying a popup menu from an iPad toolbar is part of Apple's UI guidelines, right? Popup menus displayed by an Android toolbar is common place as well.

@ypbnv, do you have any feedback? You've messed around with toolbar menus before.

hansemannn:
It seems like the `Ti.UI.Picker` already offers this UI options. Closing.

PR #10653, opened 2019-01-26
refactor(docs): use markdown-it to parse markdown

janvennemann:
JIRA: https://jira.appcelerator.org/browse/TIDOC-3279

@jawa9000 this is a first try at moving to a new markdown parser to support fenced code blocks. I compared the output of the HTML generator and apart from a few newlines and some characters that are now escaped (like quotes and angle brackets) it's identical.

I'm not familiar with the complete process for generating the docs so it would be awesome if you can take another look and see if there are any issues i might have missed. For example, i'm not sure how the code highlighting was done in the old one. markdown-it has the ability to use a custom code highlighter via the `highlight ` option.


jawa9000 (approved):
I think these changes are good and stable (so far). I completed a local build twice now, reviewed about two dozen documents (both guide and API documents) and I didn't see any issues.

PR #10658, opened 2019-01-30
fix(android)(8_0_X): Ti.App wrongly fires pause/resume events when opening/closing child windows

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26746

**Cherry-pick of:** #10634


ssekhri:
FR Passed.
The 'resumed' event is fired on app launch, and 'resume', 'resumed' events are fired on app foreground. They do not get fired when a window is opened.
The 'pause' and 'paused' events are fired on app background only. They do not get fired on window close.
Test Env:
Mac OS 10.14.3
Ti SDK: 8.0.0.v20190315171321
Appc CLI: 7.0.10
Node: 8.12.0
JDK: 9.0.4
Studio: 5.1.2.201903111843
Android Device: Pixel(v9.0), Nexus6 Emulator(v6.0)

PR #10659, opened 2019-01-30
fix(ios): memory leak in require

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26783

Adds a missing `autorelease` for the returned `ResolvedModule *` instance that is returned with a +1 reference count.


vijaysingh-axway (approved):
CR passed.

PR #10660, opened 2019-01-30
fix(ios): memory leak in require (8_0_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26783

8_0_X backport of #10659

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
@@ -1202,19 +1202,19 @@ CFMutableSetRef krollBridgeRegistry = nil;
     if (json) {
       type = JSON;
     }
-    return [[ResolvedModule alloc] initWithType:type andPath:path];
+    return [[[ResolvedModule alloc] initWithType:type andPath:path] autorelease];
   }
 
   // 2. If X.js is a file, load X.js as JavaScript text.  STOP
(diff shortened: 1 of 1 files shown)


iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -1366,7 +1366,7 @@ - (id)require:(KrollContext *)kroll path:(NSString *)path
 {
   NSURL *oldURL = [self currentURL];
   NSString *workingPath = [oldURL relativePath];
-  ResolvedModule *resolved = [self resolveRequire:path withWorkingPath:workingPath];
+  ResolvedModule *resolved = [[self resolveRequire:path withWorkingPath:workingPath] autorelease];
```
vijaysingh-axway: Isn't it good if we use autorelease, where we are allocating the object. So better we should go inside function`-(ResolvedModule *)resolveRequire:(NSString *)path withWorkingPath:(NSString *)workingPath ` and use autorelease while allocating.

vijaysingh-axway (changes requested):

vijaysingh-axway (approved):
CR passed.

sgtcoolguy:
This PR should be merged post 8.0.0.GA since the fix on master has already been merged.

sgtcoolguy (approved):

PR #10661, opened 2019-01-30
TiAPI: Create Node-compatible assert module API

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26669

**Description:**
This is an attempt to create a Node-compatible 'assert' module. This module *was* the underpinning of many of the NPM/JS test frameworks.

https://nodejs.org/api/assert.html

There's the Browserify shim, which is beginning to show it's age as I don't believe it supports Maps/Sets/Symbols. Our own unit testing framework for the SDK uses a very old version of should.js with that shim embedded in it.

Required merge of #10567 first, which is now done.

Additionally, there is overlap with #10718 (which should likely be merged first)


sgtcoolguy:
Fun stuff - using the latest official `should` npm package works, but exposes some bugs in our tests themselves that relied on "loose" equality checks, so this should fail 3 unit tests right now. Also of note, `should` no longer uses `assert` under the hood itself AFAICT and re-implements the methods itself pretty much (Presumably to work outside of node?).

So, unfortunately this PR is not actually exercising my custom `assert` core module for every test as I had hoped (a fun sort of bootstrap/eat-your-own-dogfood exercise I was aiming for). But it does show that:
- we can get rid of the old hacked version of `should`
- that we can use the off-the-shelf version of `should` from npm
- I have ~115 unit tests explicitly testing my `assert` module, so it is being tested some and appears to be pretty darn close to node's module (I don't yet have `#rejects()`, `#doesNotReject()`, nor do I yet support testing Error's against Object that have Regexp property values).

sgtcoolguy:
To be able to provide the `rejects`/`doesNotReject` methods requires polyfilling (#10567). So this PR is done for now, or would require merging that PR first to add those two.

sgtcoolguy:
Now that the polypill PR has been merged, this should be good to go, and now includes `#rejects()` and `#doesNiotReject()` implementations.

PR #10666, opened 2019-02-01
docs(ListView)Minor updates

ssjsamir:
+ Minor text change 
+ Update example to see Listview items on both iOS and Android


jawa9000 (approved):
Looks good

PR #10667, opened 2019-02-01
docs(api): remove notes about heavyweight windows

janvennemann:
Removes all the confusing instructions about heavyweigth and leightweight windows which are from pre SDK 3.2.0 era.

I have found not mention about a way to create leightweight windows these days, but just to be sure cc @garymathews @jquick-axway


jawa9000 (approved):
Looks good.

janvennemann:
@jquick-axway just to double check, there is no way to create a lightweight window these days, correct?

jquick-axway:
Our current Android code creates either an `Activity` or a tab `Fragment` when opening a `Ti.UI.Window`. Neither of those are lightweight.

Our old `Ti.UI.Window` used to support creating a view within an existing activity. This was considered a lightweight window back then. I can see it in our old 3.x.x code below. We definitely do not support this anymore
https://github.com/appcelerator/titanium_mobile/blob/3_1_X/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java

janvennemann:
@jquick-axway, perfect, thanks for the clarification. That's exactly what i thought.

@jawa9000 this is good to go!

PR #10669, opened 2019-02-01
fix(ios,android): require of dir with package.json main pointing at dir

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26788

**Description:**
Found this one when trying to move to `mocha` npm package in our test suite (rather than `ti-mocha` using lots of shims).

Requiring a node_module or directory that had a `package.json` in it, it failed to properly handle when that file's `main` property pointed at another directory.

⚠️ 8_0_X/8.0.1 back port is #10672


janvennemann (approved):
LGTM

PR #10675, opened 2019-02-05
docs: Update Calendar docs

ypbnv:
Documentation updates - some fixes, removed unnecessary whitespaces and slightly changed samples.
There is a problem on Android side with getting recurrence rules from events in a specific case. I will fix it in a separate ticket and adjust the code sample (and any APIs description if necessary) there.


jawa9000 (changes requested):
Yordan, can you also address https://jira.appcelerator.org/browse/TIDOC-3283 while this PR is open?

jawa9000 (approved):
Looks good!

PR #10676, opened 2019-02-05
fix(android): skip inclusion of soasta module

garymathews:
- Skip inclusion of `com.soasta.touchtest` module for Android builds

##### TEST CASE
- Build an existing project that has SOASTA included
- SOASTA module should be skipped during build and not present in built application

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26785)

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1518,9 +1518,18 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 			const manifestHashes = [],
 				nativeHashes = [],
 				bindingsHashes = [],
-				jarHashes = {};
+				jarHashes = {},
+				blacklist = [ 'com.soasta.touchtest' ];
 
 			modules.found.forEach(function (module) {
+
+				// skip modules from blacklist
+				// TODO: remove SOASTA files from project in 8.1.0
+				if (blacklist.includes(module.id)) {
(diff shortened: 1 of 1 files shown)


keerthi1032:
FR Passed. Waiting for the CR to merge. Test services are not enabled when app created with this PR and runs successfully.

**Test Environment:**
Name = Mac OS X
Version = 10.13.6
Architecture = 64bit
Memory = 17179869184
Node.js
Node.js Version = 8.12.0
npm Version = 6.4.1
Titanium CLI
CLI Version = 5.1.1
Titanium SDK
SDK Version = local master 8.1.0.v20190205135845
Studio version =5.1.2.201812191831
CLI =7.0.10 Master 7
Device = Samsung s5 android6 ,iPhone 6s iOS 12
Simulator =Ios 11 iPhone 6
emulator =Pixel 3xl android 8

jquick-axway (approved):
CR: Pass

And I've tested an app with the soasta `<service/>` injected in the "AndroidManifest.xml" (because it's defined in the "tiapp.xml"), but it doesn't cause any issues. The app launches fine. I don't think Google validates the service java class references in the "AndroidManifest.xml". (Just don't try to start it. That for sure would throw an exception, but the same would happen if the service name was defined in the XML as well.)

PR #10681, opened 2019-02-06
fix(ios): fix leak in TiBlob imageAsCompressed method (7_4_X)

rlustemberg:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26800

The TiBlob.imageAsCompressed method leaks memory as it's not initialising the blob with the page context. For that reason, is not deallocated once it's JS proxy is garbage collected.


keerthi1032:
@janvennemann  seeing issue with lint in jenkin. not able to test this PR. can you please look at it.

PR #10687, opened 2019-02-08
fix(android): Fixes TabLayout gravity when changing orientation

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26698

**Description:**
Takes care of switching a TabLayout's TabGravity depending on current orientation. Due to Titanium not recreating activities on orientation changes the style defined for the TabLayout in Design support library will not be applied, because the TabLayout won't be redrawn. This PR makes the change on configurationChanged event and effectively changes the component according to the new style. The change from the default style is actually only the tabGravity property.

Edit: No unit test for this one - it is a visual change.

**Test case:**
There is a good code sample in the JIRA ticket.


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUITabLayoutTabGroup.java
```diff
@@ -90,6 +90,17 @@ protected void onLayout(boolean hasChanged, int left, int top, int right, int bo
...
+			@Override
+			protected void onConfigurationChanged(Configuration newConfig)
+			{
+				super.onConfigurationChanged(newConfig);
+				if (newConfig.orientation == Configuration.ORIENTATION_PORTRAIT) {
+					mTabLayout.setTabGravity(TabLayout.GRAVITY_FILL);
+				} else if (newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE) {
+					mTabLayout.setTabGravity(TabLayout.GRAVITY_CENTER);
+				}
+			}
```
jquick-axway: Minor nit-pick. You can get rid of the `mTabLayout.` part since you're overriding the `TabLayout` class' `onConfigurationChanged()` method. The rest is fine. Nice and simple solution.

jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
The layout of the tabs is as expected in both portrait and landscape orientation. In portrait the tabs, one or more, in the tab group would fill the entire width and in landscape mode the tabs would be centred and each tab taking the width of its title text.
Environment:
Studio: 5.1.2.201812191831
SDK: 8.1.0 local build
Mac OS: 10.14.3
Xcode: Xcode 10.1
Appc NPM: 4.2.13
Appc CLI: 7.0.10-master.17
Node: 8.12.0
NPM Ver: 6.4.1
Java Ver: 9.0.4

PR #10693, opened 2019-02-11
feat(android): Add "barColor" property of Ti.UI.Window to Android.

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26817

**Description:**
Adds parity for `barColor` property of Ti.UI.Window. Also removed an unused import.
Note: As for a unit test - the scenario is the same as in https://github.com/appcelerator/titanium_mobile/pull/10393, so for now I am skipping it.

**Test case:**

With a default ActionBar:
_app.js_
```js
	button = Ti.UI.createButton();
win.add(button);
win.open();
	win.barColor = 'red';
});
```

With a Toolbar instance used as an ActionBar. You need to use a theme without an ActionBar for this sample to work.
_app.js_
```js
	button = Ti.UI.createButton(),
	toolbar = Ti.UI.createToolbar({ top: 0, barColor: 'blue'});
win.activity.supportToolbar = toolbar;
win.add([toolbar, button]);
win.open();
	win.barColor = 'red';
});
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
@@ -23,6 +23,7 @@ import org.appcelerator.titanium.TiDimension;
 import org.appcelerator.titanium.TiTranslucentActivity;
 import org.appcelerator.titanium.proxy.ActivityProxy;
 import org.appcelerator.titanium.proxy.TiWindowProxy;
+import org.appcelerator.titanium.util.TiColorHelper;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.util.TiRHelper;
 import org.appcelerator.titanium.view.TiUIView;
@@ -39,6 +40,8 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Message;
 import android.support.annotation.Nullable;
+import android.support.v7.app.ActionBar;
+import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
 import android.transition.ChangeBounds;
 import android.transition.ChangeClipBounds;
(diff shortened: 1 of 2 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -374,7 +377,18 @@ public void onPropertyChanged(String name, Object value)
 				}
 			}
 		}
-
+		if (name.equals(TiC.PROPERTY_BAR_COLOR)) {
+			// Get a reference to the ActionBar.
+			ActionBar actionBar = ((AppCompatActivity) activity.get()).getSupportActionBar();
```
jquick-axway: We need to do a `null` check on the activity reference in case the property was changed after the activity window was closed/destroyed.

Also, the above code uses the `windowActivity` member variable instead. Would you mind switching to that for consistency please? I can see that `windowActivity` is assigned via `windowCreated()`. So, we know it's never assigned to the previous activity.

jquick-axway (changes requested):

jquick-axway:
@ypbnv, for your info, on iOS the "barColor" property only applies when using the `NavigationWindow` feature. This is because an iOS window normally does not display a title bar. The below will show a blue title bar.
```javascript
var parentWindow = Ti.UI.createWindow({
	title: "My Title",
	barColor: "blue",
});
parentWindow.add(Ti.UI.createLabel({ text: "Hello World!" }));
var navigationWindow = Ti.UI.createNavigationWindow({
	window: parentWindow,
});
navigationWindow.open();
```

Android's current `NavigationWindow` implementation does not support the above either. Although our Android implementation is really controller design patter instead of a real window container. This means that setting the bar color like you're doing will work for our Android `NavigationWindow` as well.

Can you look into the following please?
- Make sure "barColor" can be set via a **creation** property.
- Test your changes with the above code.

Thanks! :)

ypbnv:
@jquick-axway Updated. I totally missed the availability in creation dictionary. Thanks! I ran the changed PR with the code sample you provided and it works fine.

jquick-axway:
Changed PR's milestone to 8.1.0 since this is for the "master" branch.

jquick-axway (approved):
CR: Pass

longton95 (approved):
**FR Passed**

Verified using the test cases above, `barColor` works as expected.

**ENV**

```
One Plus 6T (9.0)
Emulator (9.0, 8.0)
Appc NPM: 4.2.13
Appc CLI: 7.0.10-17
Ti CLI: 5.1.1
Node: 8.15.0
NPM: 6.4.1
```

PR #10699, opened 2019-02-13
fix(iOS): swift module’s app.js template made correct (8_0_X)

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26710

### DIFF
--- a/iphone/templates/module/swift/template/ios/Classes/{{ModuleIdAsIdentifier}}Module.swift.ejs
+++ b/iphone/templates/module/swift/template/ios/Classes/{{ModuleIdAsIdentifier}}Module.swift.ejs
@@ -49,17 +49,15 @@ class <%- moduleIdAsIdentifier %>Module: TiModule {
     return "hello world!"
   }
   
-  @objc(exampleProp)
-  func exampleProp() -> String {
-    // Example property getter. 
-    // Call with "MyModule.exampleProp"
-    
-    return "Titanium rocks!"
-  }
-
-  @objc(setExampleProp:)
-  func setExampleProp(value: Any?) {
-    // Example property setter. 
-    // Call with "MyModule.exampleProp = 'newValue'"
-  }
+  @objc public var exampleProp: String {
+     get { 
+        // Example property getter
+        return "Titanium rocks!"
+     }
+     set {
+        // Example property setter
+        // Call with "MyModule.exampleProp = 'newValue'"
+        self.replaceValue(newValue, forKey: "exampleProp", notification: false)
+     }
+   }
 }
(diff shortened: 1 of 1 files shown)


iphone/templates/module/swift/template/ios/Classes/{{ModuleIdAsIdentifier}}Module.swift.ejs
```diff
@@ -40,11 +40,26 @@ class <%- moduleIdAsIdentifier %>Module: TiModule {
...
+    // Example method. 
+    // Call with "MyModule.example(args)"
+
+    return "hello world!"
+  }
   
-  @objc(tryThis:)
-  func tryThis(arguments: Array<Any>?) -> String {
-    guard let arguments = arguments, let message = arguments.first else { return "No arguments" }
+  @objc(exampleProp)
```
janvennemann: This would be a great place to demonstrate Swift property getter/setter instead of the ObjC style methods:

```swift
@objc public var exampleProp: String {
    get { 
        // Example property getter
        return "Titanium rocks!"
    }
    set {
        // Example property setter
        // Call with "MyModule.exampleProp = 'newValue'"
        self.replaceValue(newValue, forKey: "exampleProp", notification: false)
    }
}
```

vijaysingh-axway: Good idea. Updated. Thanks!

ssjsamir (approved):
**FR Passed** The example in the swift module now works by default.

example.js
```
// This is a test harness for your module
// You should do something interesting in this harness
// to test out the module and to provide instructions
// to users on how to use it by example.


// open a single window
	backgroundColor:'white'
});
win.add(label);
win.open();

// TODO: write your module tests here
var test = require('com.example.tese');
Ti.API.info("module is => " + test);

label.text = test.example();

Ti.API.info("module exampleProp is => " + test.exampleProp);
test.exampleProp = "This is a test value";

if (Ti.Platform.name == "android") {
	var proxy = test.createExample({
		message: "Creating an example Proxy",
		backgroundColor: "red",
		width: 100,
		height: 100,
		top: 100,
		left: 150
	});

	proxy.printMessage("Hello world!");
	proxy.message = "Hi world!.  It's me again.";
	proxy.printMessage("Hello world!");
	win.add(proxy);
}

```
**Test Environment**
```
iPhone 6 Sim (12.1)
APPC CLI: 7.0.10-14
Node.js Version: 8.9.1
Xcode 10.1
```
**Awaiting CR** Has this been approved to go in to 8_0_X?

PR #10702, opened 2019-02-14
fix(android): fix copying of resources

garymathews:
- Relocate `removeOldFiles` to before new resources are copied over
- Remove unnecessary `setImmediate` call
- Do not symlink a symlink when linking the `appicon.png`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26824)

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1758,11 +1758,11 @@ AndroidBuilder.prototype.run = function run(logger, config, cli, finished) {
 		},
 
 		'createBuildDirs',
+		'removeOldFiles',
 		'copyResources',
 		'generateRequireIndex',
 		'processTiSymbols',
 		'copyModuleResources',
-		'removeOldFiles',
 		'copyGradleTemplate',
 		'generateJavaFiles',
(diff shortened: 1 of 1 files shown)


sgtcoolguy (approved):
LGTM, but may want @cb1kenobi to take a peek too?

Sounds like you have confirmed it fixes the issue locally, though.

garymathews:
At the moment, I seem to be the only one that has ran into this issue. I'm not sure if it's exposed due to having a fast SSD. If I delete my `build` folder and build I can consistently reproduce this issue.

janvennemann:
@garymathews These changes are making it hard to implement #10805 for Android. 

The `removeOldFiles` step removes all files in the `lastBuildFiles` map. This map is generated at the beginning by the builder and then the individual steps modify it to control which files to remove. Moving `removeOldFiles` to the top has rendered this logic pretty much useless and almost all files will be deleted now.

m1ga:
I think this PR might be the source for https://jira.appcelerator.org/browse/AC-6360 I see random files not appearing in my resources folder/APK.

PR #10706, opened 2019-02-15
fix(android): resolve template compatibility with Java 9+

garymathews:
- Amed template for compatibility with Java 9+
- Fix building of hybrid native modules

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26826)


sgtcoolguy (changes requested):
Have you tested with Java <= 6 and Java 7+? We had to split the titanium_prep binaries based on JDK version due to changes in he includes or something long ago. Maybe we don't support older JDKs now? Looks to me like the app build still uses the different binaries, but the module build doesn't. So there's an issue here:
- either we can use one binary and it works (or we no longer support JDK 6-)
- or the module build needs to swap macOS binaries based own version as well

Also, why the removal of the titanium_prep exes for windows? AFAIK we still need them.

garymathews:
@sgtcoolguy We don't support Java 1.6 anymore since 1.7+ is required to build for Android (Studio also installs Java 7 by default). I have updated the `build.js` to remove this check. Also, the Windows binaries are included, they are just ~60% smaller than the previous binaries which makes them red on Github 'Files changed'.

sgtcoolguy (approved):

ssjsamir (approved):
**FR Passed** Able to build Ti.playservices and then include it in titanium application

**Test Steps**

1. Downloaded ti.playservices 
2. cd in to ti.playservices 
3. cd into Android folder 
4. Created lib folder
5. cd in to ti.playservices root
6. cd into the updater folder
7. did npm install node .
8. did npm install && node .
9. cd in to ti.playservices root
10. cd into Android folder 
11. run appc run -p android --build-only
12. added the module to a titanium application
13. added the code in to the {{app.js}}
```
const PlayServices = require('ti.playservices');

const btn = Ti.UI.createButton({ title: 'CHECK PLAY SERVICES' });

    PlayServices.makeGooglePlayServicesAvailable((e) => {
        if (e.success) {
            alert(`Play Services: ${PlayServices.GOOGLE_PLAY_SERVICES_VERSION_CODE}`);
            // Use Play Services
        } else {
            alert(`Play Services is not available.`);
        }
    });
});

win.add(btn);
win.open();
```
14. ran the program 
15. pressed check play services 
16. version number was returned 


**Test Environment** 
```
Google pixel xl 7.1.1 sim
APPC CLI: 7.0.10
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10708, opened 2019-02-18
fix(ios): remove startup log from titaniumkit

ewanharris:
This is now done via the ti.main script. Closes TIMOB-26833

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26833

⚠️ Note: the log that was being output had the SDK down as (0.0.0.(null)), and `TITANIUM_VERSION` is being set to `0.0.0` I can't find anywhere where we're referencing that? But if we are, whatever uses it is using invalid information?

### Test steps 

* Build an app to iOS device/sim, on startup only one log like `plainalloy 1.0 (Powered by Titanium 8.1.0.849c15e098)` should be output.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -64,9 +64,13 @@ async function checkJIRA() {
 // Check that if we modify the Android or iOS SDK, we also update the tests
 // Also, assign labels based on changes to different dir paths
 async function checkChangedFileLocations() {
+	let modifiedTopTiModule = false;
 	const modified = danger.git.modified_files.concat(danger.git.created_files);
 	const modifiedAndroidFiles = modified.filter(p => p.startsWith('android/') && p.endsWith('.java'));
 	const modifiedIOSFiles = modified.filter(p => {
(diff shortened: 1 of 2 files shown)


ewanharris:
Added a (naive) check to add a warning if the TopTiModule.m changes during scons build are committed.

jquick-axway:
Nice find. I did remove that log message in my PR below, but it looks like it was copied into TitaniumKit later.
https://github.com/appcelerator/titanium_mobile/pull/10112/files#diff-4c87204e7b1bd7f7c84399ab50226410

jquick-axway (approved):
CR: Pass

sgtcoolguy:
@ewanharris Note that I also (secretly) added a git hook commit check that enforces no one checked in static values in the TopTiModule.m file. Obviously the hook can be bypassed, but I wanted to try and avoid this situation too.

sgtcoolguy (approved):

PR #10718, opened 2019-02-21
TiAPI: Create Node-compatible util module API

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26670

**Description:**
- This does not includes every method off of util.types (which is new as of Node 10+).
- This does not include `util._extends()`
- This does not yet include `util.isDeepStrictEqual()`, which would be done after #10661 was also merged.
- I did not attempt to handle `util.inspect` options too far. We support `depth` and `showHidden`, and to some degree `breakLength`; I didn't handle `showProxy`, `compact`, `maxArrayLength`, `colors`, `customInspect`, `sorted` custom functions (just Boolean). I also assume there's lots of edge cases like WeakMaps, WeakSets, Map/Set iterators, etc that we don't properly support yet.

Note that due to the way iOS proxies are set up, I can't verify the format of util.log/error/debug/puts/print messages because we cannot override `console.log` or `console.error`.

#10069 would fix that issue.

### DIFF
--- a/common/Resources/ti.internal/extensions/process.js
+++ b/common/Resources/ti.internal/extensions/process.js
@@ -46,14 +46,87 @@ function standardizeArch(original) {
 	}
 }
 
+const listeners = {};
 const process = {
 	arch: standardizeArch(Ti.Platform.architecture),
+	noDeprecation: false,
+	throwDeprecation: false,
+	traceDeprecation: false,
+	pid: 0,
 	cwd: function () {
 		return __dirname;
 	},
 	// FIXME: Should we try and adopt 'windowsphone'/'windowsstore' to 'win32'?
 	// FIXME: Should we try and adopt 'ipad'/'iphone' to 'darwin'? or 'ios'?
-	platform: Ti.Platform.osname
+	platform: Ti.Platform.osname,
+	on: function (eventName, callback) {
+		const eventListeners = listeners[eventName] || [];
+		eventListeners.push(callback);
+		listeners[eventName] = eventListeners;
+		return this;
+	},
+	// TODO: Add #once which is like #on, but should get wrapped to remove itself before getting fired
+	emit: function (eventName, ...args) {
+		const eventListeners = listeners[eventName] || [];
+		for (const listener of eventListeners) {
+			listener.call(this, ...args);
+		}
+		return eventListeners.length !== 0;
+	},
+	eventNames: () => Object.getOwnPropertyNames(listeners),
+	emitWarning: function (warning, options, code, ctor) {
+		let type;
+		let detail;
+		if (typeof options === 'string') {
+			type = options;
+		} else if (typeof options === 'object') {
+			type = options.type;
+			code = options.code;
+			detail = options.detail;
+		}
+		if (typeof warning === 'string') {
+			// TODO: make use of `ctor` arg for limiting stack traces? Can only really be used on V8
+			// set stack trace limit to 0, then call Error.captureStackTrace(warning, ctor);
+			warning = new Error(warning);
+			warning.name = type || 'Warning';
+			if (code !== undefined) {
+				warning.code = code;
+			}
+			if (detail !== undefined) {
+				warning.detail = detail;
+			}
+		}
+		// TODO: Throw TypeError if not an instanceof Error at this point!
+		const isDeprecation = (warning.name === 'DeprecationWarning');
+		if (isDeprecation && process.noDeprecation) {
+			return; // ignore
+		}
+		if (isDeprecation && process.throwDeprecation) {
+			throw warning;
+		}
+		this.emit('warning', warning);
+	}
 };
 global.process = process;
+// handle spitting out warnings
+const WARNING_PREFIX = `(titanium:${process.pid}) `;
+process.on('warning', warning => {
+	const isDeprecation = (warning.name === 'DeprecationWarning');
+	// if we're not doing deprecations, ignore!
+	if (isDeprecation && process.noDeprecation) {
+		return;
+	}
+	// TODO: Check process.traceDeprecation and if set, include stack trace in message!
+	let msg = WARNING_PREFIX;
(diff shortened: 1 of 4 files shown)


common/Resources/ti.internal/extensions/util.js
```diff
@@ -0,0 +1,497 @@
...
+		// Produces output like: "21 Feb 10:04:23 - message"
+		console.log(`${date.getDate()} ${MONTHS[date.getMonth()]} ${time} - ${string}`);
+	},
+	print: (...args) => console.log(args.join('')), // FIXME: Shouldn't add trailing newline like console.log does!
+	puts: (...args) => console.log(args.join('\n')),
+	error: (...args) => console.error(args.join('\n')),
+	debug: string => console.error(`DEBUG: ${string}`)
+};
+
+util.isBuffer = () => false; // FIXME: Check for Ti.Buffer? for node/browserify buffer?
```
janvennemann: Maybe a `Buffer` interface around our `Ti.Buffer`?

common/Resources/ti.internal/extensions/process.js
```diff
@@ -46,14 +46,87 @@ function standardizeArch(original) {
...
 	// FIXME: Should we try and adopt 'ipad'/'iphone' to 'darwin'? or 'ios'?
-	platform: Ti.Platform.osname
+	platform: Ti.Platform.osname,
+	on: function (eventName, callback) {
+		const eventListeners = listeners[eventName] || [];
+		eventListeners.push(callback);
+		listeners[eventName] = eventListeners;
+		return this;
+	},
+	// TODO: Add #once which is like #on, but should get wrapped to remove itself before getting fired
```
janvennemann: We also might want to move the event stuff into a separate module so it can be reused.

common/Resources/ti.internal/extensions/util.js
```diff
@@ -0,0 +1,497 @@
...
+util.isDate = value => util.types.isDate(value);
+util.isError = value => util.types.isNativeError(value);
+util.isRegexp = value => util.types.isRegexp(value);
+
+// FIXME: Our String.format is not very forgiving. It sort-of is supposed to do the same thing, but blows up easily
+// util.format = String.format;
```
janvennemann: Can this be removed? since you wrote your own format function further down?

common/Resources/ti.internal/extensions/util.js
```diff
@@ -0,0 +1,497 @@
...
+		const objType = typeof obj;
+		if (objType === 'object' || objType === 'function') {
+			if (obj === null) {
+				return 'null';
+			}
+
+			// Guard against circular references
+			// FIXME: Need to push/pop the references, not store forever!
+			mergedOptions.memo = mergedOptions.memo || [];
+			if (mergedOptions.memo.includes(obj)) {
```
janvennemann: What about repeated occurrences of the same object, for example: `const a={id:1}; util.inspect([a,a]);`

common/Resources/ti.internal/extensions/process.js
sgtcoolguy: agreed. As I go I am eventually extracting the code out to it's proper place, but didn't want to go nuts writing every other module as these PRs are large enough. But, yeah we need to define EventEmitter eventually.

common/Resources/ti.internal/extensions/util.js
sgtcoolguy: yeah, it was a reminder as I tried to use our String.format as a placeholder and it broke right away.

common/Resources/ti.internal/extensions/util.js
sgtcoolguy: good catch, and yes that test is a nice way to check. I've since added this as a unit test and fixed it.

common/Resources/ti.internal/extensions/process.js
janvennemann: Alright, sounds good!

common/Resources/ti.internal/extensions/util.js
```diff
@@ -0,0 +1,498 @@
...
+		throw new TypeError(`The "constructor" argument must be of type Function. Received type ${typeof constructor}`);
+	}
+	if (superConstructor == null) { // eslint-disable-line eqeqeq, no-eq-null
+		throw new TypeError(`The "superConstructor" argument must be of type Function. Received type ${typeof superConstructor}`);
+	}
+	if (superConstructor.prototype === undefined) {
+		throw new TypeError('The "superConstructor.prototype" argument must be of type Object. Received type undefined');
+	}
+
+	constructor.super_ = superConstructor; // FIXME: Don't make enumerable!
```
janvennemann: Can't this be fixed with `Object.defineProperty`?
```suggestion
	Object.defineProperty(constructor, 'super_', superConstructor);
```

janvennemann (commented):
Added two notes about todo/fixes which sound rather trivial and could be done right now.

Apart from that this already looks pretty damn solid for simple shimming purposes. Nice one!

common/Resources/ti.internal/extensions/util.js
sgtcoolguy: yup, I'll push a fix.

common/Resources/ti.internal/extensions/util.js
```diff
@@ -0,0 +1,498 @@
...
+			}
+			lastIndex = foundIndex + 2;
+		}
+
+		// If we haven't reached end of string, append rest of it with no replacements!
+		str += firstArg.slice(lastIndex, firstArg.length);
+
+		// If we have args remaining, need to...
+		// loop over rest of args and coerce to Strings and concat joined by spaces.
+		// FIXME: Unless typeof === 'object' or 'symbol', then do util.inspect() on them
```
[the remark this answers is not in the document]
sgtcoolguy: yeah, I'll do it now.

PR #10719, opened 2019-02-21
feat(android): Adds extended support for color customization of Ti.UI.Android.SearchView

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26822

**Description:**
Adds a few new properties for color customization of Ti.UI.Android.SearchView.

**Test case:**
app.js
```js
		title: 'Search Test',
		layout: 'vertical'
	}),
	searchViewActionBar = Ti.UI.Android.createSearchView({
		hintText: 'Action Bar Search',
		hintTextColor: 'cyan',
		iconSearchColor: 'magenta',
		iconSearchHintColor: 'yellow'
	}),
	searchViewIconified = Ti.UI.Android.createSearchView({
		hintText: 'Action Bar Default',
		hintTextColor: 'blue',
		iconSearchColor: 'green',
		iconSearchHintColor: 'red',
		iconSearchCloseColor: 'orange'
	}),
	searchViewNonIconified = Ti.UI.Android.createSearchView({
		hintText: 'Action Bar Non Iconified',
		iconifiedByDefault: false,
		hintTextColor: 'red',
		iconSearchColor: 'green',
		iconSearchHintColor: 'blue',
		iconSearchCloseColor: 'purple'
	});

	e.menu.add({
		title: 'Search',
		actionView: searchViewActionBar,
		icon: Ti.Android.R.drawable.ic_menu_search,
		showAsAction: Ti.Android.SHOW_AS_ACTION_IF_ROOM | Ti.Android.SHOW_AS_ACTION_COLLAPSE_ACTION_VIEW,
	});
};
window.add([searchViewIconified, searchViewNonIconified]);
window.open();

```


android/titanium/src/java/org/appcelerator/titanium/util/TiMenuSupport.java
```diff
@@ -47,6 +51,16 @@ public boolean onCreateOptionsMenu(boolean created, Menu menu)
 		if (onCreate != null || onPrepare != null) {
 			created = true;
 		}
+
+		Drawable searchIconDrawable = null;
+		try {
+			searchIconDrawable = TiApplication.getInstance().getResources().getDrawable(
+				TiRHelper.getResource("drawable.abc_ic_search_api_material"), null);
+			searchIconDrawable.setColorFilter(Color.RED, PorterDuff.Mode.SRC_IN);
```
jquick-axway: I think you need to call `searchIconDrawable.mutate()` since you're receiving a global reference to the drawable here.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
```diff
@@ -177,4 +163,98 @@ public void release()
...
+			ssb.append("");
+			// In order to stick to the default behavior we need to manually add the hint text,
+			// if there is one, to the newly created SpannableString. This is necessary because
+			// calling setQuery method of the SearchView triggers the default getDecoratedHint
+			// method that will create a new ImageSpan without the custom color filter.
+			if (getProxy().hasPropertyAndNotNull(TiC.PROPERTY_HINT_TEXT)) {
+				ssb.append(getProxy().getProperty(TiC.PROPERTY_HINT_TEXT).toString());
+			}
+			// Add it as a hint.
+			editText.setHint(ssb);
```
jquick-axway: ^ The above makes me nervous. It makes assumptions about the internal implementation details on Google's end, which could change in the future. Is there really no other nice way to do this? I was able to replace this icon via a theme/style, except when the `SearchView` was embedded within the `ActionBar`.

jquick-axway: I mean, you might be right. There might be no other way to do this without hacks... but we then have to be concerned about potential breakage/issues in the future.

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -1923,6 +1923,21 @@
...
+	/**
+	 * @module.api
+	 */
+	public static final String PROPERTY_ICON_SEARCH_CLOSE_COLOR = "iconSearchCloseColor";
+
+	/**
+	 * @module.api
+	 */
+	public static final String PROPERTY_ICON_SEARCH_HINT_COLOR = "iconSearchHintColor";
+
```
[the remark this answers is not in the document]
ypbnv: I agree with the proposed name changes and I will update them.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/searchview/TiUISearchView.java
ypbnv: From what I tested providing a custom theme with overridden properties for close, hintIcon and searchIcon worked just fine. The case where I embedded the SearchView as a Menu item in the ActionBar required to separately pass the drawable I used for `<item name="searchIcon">@drawable/pictureduo</item>` because it is different from the already embedded resources which are referenced by ID (like the sample `icon: Ti.Android.R.drawable.ic_menu_search,`)
Possibly the biggest benefit would be that we can have partial parity with the properties of Ti.UI.SearchBar, but that leaves us at "Is it worth it to have that at the price of blindly assuming what is happening under the AppCompat library?" Maybe we can discuss this in details offline ?

PR #10725, opened 2019-02-23
fix(ios): properly copy completion handlers, remove duplicate UIApplicationDelegate selectors

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26854

This pull fixes 2 major issues:
1. Properly copy completion handlers to prevent a hard crash in those delegate methods
2. Remove the old methods that have been moved from TiApp.m (core project) to `TiApp-Addons.m` (TitaniumKit). 

Note: I created that file back then to be able to handle preprocessor-based methods (`#ifdef TI_APPIOS*`), since TitaniumKit is precompiled and could not handle this. There should be some more validation if there are not more methods like the already listed ones.


vijaysingh-axway:
@hansemannn Changes looks good. Can you please give test case for crash to be verified by QE? Thanks!

hansemannn:
@vijaysingh-axway The steps are now linked in the JIRA ticket.

vijaysingh-axway (approved):
CR passed.

ssekhri:
FR passed. No crash observed during handoff. Tested between three devices one iOS 9 and two iOS 12 device. Also tested the handoff immediately after launch on first device and after a while. No crash observed.
Verified on:
Mac OS: 10.14.3
Ti SDK: 8.1.0.v20190515121020
Appc CLI: 7.0.11
XCode: 10.2.1

PR #10729, opened 2019-02-27
fix(ios): add missing runtime checks

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26829

The `__IPHONE_XX_X` macros are evaluated at compile time which could lead to crashes on devices running a lower version than what the macros checked against. To prevent that i added runtime checks for the required version in the media module which was affected by the crash.


iphone/Classes/MediaModule.m
```diff
@@ -148,30 +148,32 @@ + (NSDictionary *)filterableItemProperties
...
-                                              MPMediaItemPropertyRating, @"rating",
-                                              MPMediaItemPropertyAssetURL, @"assetURL",
-                                              MPMediaItemPropertyIsExplicit, @"isExplicit",
-                                              MPMediaItemPropertyReleaseDate, @"releaseDate",
-                                              MPMediaItemPropertyBeatsPerMinute, @"beatsPerMinute",
-                                              MPMediaItemPropertyComments, @"comments",
-                                              MPMediaItemPropertyLastPlayedDate, @"lastPlayedDate",
-                                              MPMediaItemPropertyUserGrouping, @"userGrouping",
-                                              MPMediaItemPropertyBookmarkTime, @"bookmarkTime",
-#ifdef __IPHONE_10_0
```
hansemannn: These have been used to ensure Xcode compatibility (in the above case to support Xcode 7 although the API would require Xcode 8). Is there a concept to support this in the future? I could think of class extensions that are injected outside TitaniumKit.

janvennemann: Ahhh thanks for the explanation, that makes sense. So i think we still need these but with an additional runtime check like in the WebView below to prevent the runtime crash on devices that are actually below the checked iOS version.

janvennemann:
@vijaysingh-axway can you give this another look?

vijaysingh-axway:
@janvennemann @hansemannn  As per previous history in titanium sdk, I can say that we guard current Xcode sdk and current  Xcode sdk - 1. e.g in titanium SDK 8.0.0 we guard against Xcode 10 and Xcode 9.
iOS 10 comes with Xcode 8. Here we are guarding for Xcode 8 using #ifdef __IPHONE_10_0. Should we guard for Xcode 8 as well? Though this will not create any problem. Am I missing anything?

janvennemann:
Well, technically it is possible to use old iOS SDKs with newer Xcode versions. So i think the check against the iOS SDK is more sophisticated and generally more expressive. It makes clear what the check is meant to do: check the iOS SDK version.

Although Xcode versions are (usually) tied to a specific iOS SDK checking for the Xcode version seems confusing to me. If you don't know the SDK version shipped with Xcode you need to look that up.

vijaysingh-axway (approved):
CR passed.

ssjsamir (approved):
**FR Passed** Titanium.Media.openMusicLibrary no longer returns empty items on iOS 9.3.5 and no longer crashes the application. Tested using the test case from: https://jira.appcelerator.org/browse/TIMOB-26829

**Test Environment**
```
iphone 5s (9.3.5)
APPC CLI: 7.0.10
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10735, opened 2019-02-28
fix: remove unmarking of @babel/polyfill during android build

ewanharris:
@babel/polyfill is now bundled and distributed via the core ti.internal scripts rather than being placed into the app at build time

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26869


sgtcoolguy (approved):

sgtcoolguy:
cc @garymathews

sgtcoolguy:
Merged since I do see from he build that this fixed the issue we were running into (even though Android is still experiencing issues, due to the new emulators)

PR #10741, opened 2019-03-01
build: check destination exists before trying to get stats info

ewanharris:
If the destination dir (`/Users/eharris/Library/Application\Support/Titanium/mobilesdk/osx/8.1.0` or similar), doesn't exist then this will currently throw checking for the stats. Just check if it exists before that

#### Verification steps

1. Nuke your dir that matches `/Users/eharris/Library/Application\Support/Titanium/mobilesdk/osx/8.1.0`
2. Run `node build/scons cleanbuild`

No errors should occur and you should have an SDK copied over

### DIFF
--- a/build/utils.js
+++ b/build/utils.js
@@ -260,7 +260,7 @@ Utils.installSDK = async function (versionTag, symlinkIfPossible = false) {
 	}
 
 	const destDir = path.join(dest, 'mobilesdk', osName, versionTag);
-	if (await fs.pathExists(destDir)) {
+	try {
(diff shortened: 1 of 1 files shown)


ypbnv (approved):
Looks good to me!

sgtcoolguy:
According to the node fs docs, they prefer you just try and do whatever operation and catch errors rather than check if something exists first (since it actually has to do equivalent of stat under the hood first).

ewanharris:
@sgtcoolguy updated to try/catch and ignore the error if thrown

PR #10744, opened 2019-03-02
fix(android): Fixed crash caused by reading View's "backgroundDisabledColor" if background/border properties are set

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26841

**Summary:**
Fixed bug where if both a background color property and a border property was set, then reading the "backgroundDisabledColor" property would cause a crash.

**Note:**
@garymathews , @ypbnv , I added a new `TiCast.java` class used to provide a C#/Swift `as` keyword like functionality where if it's unable to cast to the given type, then it'll return `null`. It allows us to do the following. What do you think?
```java
BitmapDrawable result = TiCast.as(getDrawable(), BitmapDrawable.class);
if (result != null) {
	// Got it!
}
```

Note that I'd prefer to use generics more heavily, but Java generics cannot provide type information at runtime, making the **below** impossible. The only viable solution that I know of is to pass in a `Class<T>` type like I'm doing in this PR's code.
```java
public static <T> T as(Object value)
{
	if (value instanceof T) { // <- This won't work.
		return (T) value;
	}
	return null;
}
```

**Test:**
2. Verify that the app does not crash on startup.
3. Uncomment the 2 lines of code below.
4. Rebuild and rerun the app on Android.
5. Verify that the displayed `TextField` now has a gray background.
6. In the log, verify that you see the following message. _("#FF888888" is the color "gray".)_
`[INFO] :   ### textField.backgroundDisabledColor: #FF888888`

```javascript
	value: "Hello World",
	width: "80%",
	color: "white",
	backgroundColor: "black",
	borderColor: "white",
//	backgroundDisabledColor: "gray",
//	enabled: false,
});
window.add(textField);
window.open();
Ti.API.info("### textField.backgroundDisabledColor: " + textField.backgroundDisabledColor);
Ti.API.info("### JSON.stringify(textField): " + JSON.stringify(textField));
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
@@ -29,6 +29,7 @@ import org.appcelerator.titanium.TiBlob;
 import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.TiDimension;
 import org.appcelerator.titanium.util.TiAnimationBuilder;
+import org.appcelerator.titanium.util.TiCast;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.util.TiUrl;
 import org.appcelerator.titanium.util.TiUIHelper;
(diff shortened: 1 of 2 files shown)


ypbnv (approved):
LGTM!

ypbnv:
For the bug - besides the complaint for formatting from Danger it looks good to me.
And seeing that I introduced it I would like to know what do you think about getting properties this way. On one hand I am not sure if it is worth it to make reflections like this for every place we have drawables, but on the other this could affect our unit tests in a good way - making sure we do not regress straightforward properties. Recently I have stumbled upon similar situations on a few occasions and I am starting to think if it will be good to try to apply it wherever possible. Of course other concern I have about such a move is the one mentioned in here: https://github.com/appcelerator/titanium_mobile/pull/10393 ( Long story short - backwards compatibility with human readable color names ).

I am OK with the introduction of the `as` method. I see myself taking advantage of it.

jquick-axway:
@ypbnv, I think I should close this PR in favor of your PR #10393. Thanks for bringing that up.

I'll see about introducing `TiCast` later. I'm not sure if I'm 100% settled on it yet. I may just get rid of its instance methods and only keep the static `as()` method instead.

garymathews:
Closing in favor of https://github.com/appcelerator/titanium_mobile/pull/10393

###### NOTE: You should make another PR containing TiCast

PR #10764, opened 2019-03-08
refactor(common): move to es6 import/export syntax for common sdk js when possible

sgtcoolguy:

**Description:**
Since we are already both polyfilling and transpiling the common SDK JS code, we should make use of more modern syntax. The gain here is that using import/export means more of the extensions are bundled up using rollup into the ti.main.js file to help reduce startup performance impact. I had to change my "binding" hack to support two variants: redirecting requires from a core id to the underlying file path to actually load (no longer used right now); and to add a pre-loaded binding, attaching an object to return when requiring the module id (now used to bind util/assert/tty/path/os impls to the module ids to "expose" them as via require).

Additionally, I'd say that moving to es6 syntax helps clean up some of the code (for me, the move from `var` to `const`/`let` make for much more natural code around variables).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -44,7 +44,15 @@
       }
     },
     {
-      "files": [ "common/Resources/ti.main.js" ],
+      "files": [
+        "common/Resources/ti.main.js",
+        "common/Resources/ti.internal/extensions/binding.js",
+        "common/Resources/ti.internal/extensions/os.js",
+        "common/Resources/ti.internal/extensions/path.js",
+        "common/Resources/ti.internal/extensions/tty.js",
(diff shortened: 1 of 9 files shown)


.eslintrc
```diff
@@ -44,18 +44,20 @@
       }
     },
     {
-      "files": [ "common/Resources/ti.main.js" ],
+      "files": [
+        "common/Resources/ti.main.js",
```
ewanharris: Can we glob this so that we don't have to keep updating in future?

ewanharris (approved):
Everything looks good to me.

Nit on the eslint conf but not going to block on that.

sgtcoolguy:
Merged manually.

PR #10770, opened 2019-03-12
fix(android): Modified "Ti.Database" install() and open() methods

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-26878
- https://jira.appcelerator.org/browse/TIMOB-26877

**Summary:**
- [TIMOB-26878](https://jira.appcelerator.org/browse/TIMOB-26878) Fixed bug where `Ti.Database.install()` would fail if destination directory did not exist.
  * Fixed by creating directory tree via `mkdirs()` before copying source file.
  * Note that this issue commonly happens on Android 6.0.
- [TIMOB-26877](https://jira.appcelerator.org/browse/TIMOB-26877) Modified `Ti.Database.install()` to throw an exception if given source database file was not found.
  * Old behavior would create a new database file instead, which was causing confusion. If you want to create a new/clean database, then use the `open()` method instead.
  * Now matches iOS' behavior.
- Modified `Ti.Database.open()` to check if SQLite returned `null` or not.
  * Old code was assuming it was not `null` and attempting to log the database's name, causing an exception.
- Removed logging of exception before re-throwing it in `install()` and `open()` methods.
  * Not needed since Titanium's unhandled exception handler will log it anyways. Prevents the exception from being logged twice.

**Test:**
1. Create a Classic Titanium app.
2. Set the "tiapp.xml" property "analytics" to `false` as shown below.
3. Set up the "app.js" with the below code.
4. Copy the `Database1.sqlite` file from [TIMOB-26878](https://jira.appcelerator.org/browse/TIMOB-26878) to your project's "Resources" directory.
5. Build and run on Android 6.0.
6. Verify that the following was logged on startup.
```
[INFO] :   - Key: 'CompanyName', Value: 'Axway'
[INFO] :   - Key: 'TestMessage', Value: 'Hello World'
```
7. Back out of the app.
8. Restart the app.
9. Verify that `- Key: 'LastLaunchTime'...` was logged. 

---
**tiapp.xml**
```xml
	<analytics>false</analytics>
</ti:app>
```

---
**app.js**
```javascript
var dbConnection = Ti.Database.install("Database1.sqlite", "Database1");
var resultSet = dbConnection.execute("SELECT key, value FROM properties");
Ti.API.info("@@@ Database Table 'properties' row count: " + (resultSet ? resultSet.rowCount : "<null>"));
if (resultSet) {
	for (; resultSet.isValidRow(); resultSet.next()) {
		Ti.API.info("- Key: '" + resultSet.field(0) + "', Value: '" + resultSet.field(1) + "'");
	}
}
dbConnection.execute("INSERT OR REPLACE INTO properties(key, value) VALUES (?, ?)", "LastLaunchTime", new Date());
dbConnection.close();
```


android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
+			if (db != null) {
 				dbp = new TiDatabaseProxy(db);
 			} else {
-				String name = TiConvert.toString(file);
-				SQLiteDatabase db = TiApplication.getInstance().openOrCreateDatabase(name, Context.MODE_PRIVATE, null);
+				String badPath = (absolutePath != null) ? absolutePath : "(null)";
+				badPath = "'" + badPath + "'";
+				throw new RuntimeException("SQLiteDatabase.openDatabase() returned null for path: " + badPath);
+			}
+		} else {
```
garymathews: Should we check if `file` is a `String` and return an invalid argument error if it is not?

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
+			Log.d(TAG, "Opening database from filesystem: " + absolutePath);
+
+			SQLiteDatabase db = SQLiteDatabase.openDatabase(
+				absolutePath, null, SQLiteDatabase.CREATE_IF_NECESSARY | SQLiteDatabase.NO_LOCALIZED_COLLATORS);
+			if (db != null) {
 				dbp = new TiDatabaseProxy(db);
 			} else {
-				String name = TiConvert.toString(file);
-				SQLiteDatabase db = TiApplication.getInstance().openOrCreateDatabase(name, Context.MODE_PRIVATE, null);
+				String badPath = (absolutePath != null) ? absolutePath : "(null)";
```
garymathews: I'm not sure this check is necessary? Appending `null` to a string will output `null` into the string:
```Java
throw new RuntimeException("SQLiteDatabase.openDatabase() returned null for path: '" + absolutePath + "'");
```

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
+				badPath = "'" + badPath + "'";
+				throw new RuntimeException("SQLiteDatabase.openDatabase() returned null for path: " + badPath);
+			}
+		} else {
+			String name = TiConvert.toString(file);
+			SQLiteDatabase db = TiApplication.getInstance().openOrCreateDatabase(name, Context.MODE_PRIVATE, null);
+			if (db != null) {
 				dbp = new TiDatabaseProxy(name, db);
+			} else {
+				String badName = (name != null) ? name : "(null)";
```
garymathews: Same as above

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
+		}
+
+		// Fetch a path to the source database file. This is the file to be copied/installed.
+		// Throw an exception if the source database was not found.
+		Log.d(TAG, "db url is = " + url, Log.DEBUG_MODE);
+		TiUrl tiUrl = TiUrl.createProxyUrl(invocation.getSourceUrl());
+		String sourcePath = TiUrl.resolve(tiUrl.baseUrl, url, null);
+		TiBaseFile srcDb = TiFileFactory.createTitaniumFile(sourcePath, false);
+		if (srcDb.isFile() == false) {
+			if (url == null) {
```
garymathews: Again, I don't think this is necessary?

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
-					}
-					try {
-						os.close();
-					} catch (Exception ig) {
-					}
-				}
+		// open an empty one to get the full path and then close and delete it
+		if (name.startsWith("appdata://")) {
+			String path = name.substring(10);
+			if (path != null && path.length() > 0 && path.charAt(0) == '/') {
```
garymathews: Should we improve validation here?
```Java
if (path == null || path.isEmpty()) {
  // throw error?
} else if (path.charAt(0) == '/') {
  path = path.substring(1);
}
```

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,98 @@ public DatabaseModule()
...
+		// Copy the source database file to the destination directory. (ie: Do the install.)
+		try (InputStream inputStream = new BufferedInputStream(srcDb.getInputStream());
+			 OutputStream outputStream = new BufferedOutputStream(new FileOutputStream(dbPath))) {
+
+			byte[] byteBuffer = new byte[8096];
+			int byteCount = 0;
+			while ((byteCount = inputStream.read(byteBuffer)) > 0) {
+				outputStream.write(byteBuffer, 0, byteCount);
+			}
+			outputStream.flush();
```
garymathews: Close both `inputStream` and `outputStream `? `flush` wouldn't be necessary with `close` either.

garymathews (changes requested):

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
jquick-axway: Oh wow, you're right. Concatenating a `null` reference to a string does work.

I remember Java throwing an exception when doing this in the past, but it looks like it's safe now.

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
jquick-axway: From my past experience, not all stream classes flush when closed. So, I think it's better not to assume and do it explicitly (better safe than sorry).

jquick-axway:
Updated PR

jquick-axway:
Updated PR:
Added integer boundary DB unit test.
Not related to fixes made via PR, but we should be doing this kind of test anyways.

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,108 @@ public DatabaseModule()
...
+			if (dbname.equals(name)) {
+				return open(name);
+			}
+		}
+
+		// Fetch a path to the source database file. This is the file to be copied/installed.
+		// Throw an exception if the source database was not found.
+		Log.d(TAG, "db url is = " + url, Log.DEBUG_MODE);
+		if (invocation == null) {
+			throw new RuntimeException("Ti.Database.install() was given a null 'KrollInvocation' object.");
```
sgtcoolguy: this seems like "inside baseball" leaking out here. This argument is a special one that the user has no control over and does not deal with and this exception leaking to them may be rather confusing. If this argument is null, then the SDK did something wrong internally.

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
```diff
@@ -53,101 +53,108 @@ public DatabaseModule()
...
+		}
+		TiUrl tiUrl = TiUrl.createProxyUrl(invocation.getSourceUrl());
+		String sourcePath = TiUrl.resolve(tiUrl.baseUrl, url, null);
+		TiBaseFile srcDb = TiFileFactory.createTitaniumFile(sourcePath, false);
+		if (srcDb.isFile() == false) {
+			String message = "Ti.Database.install() failed to find 1st argument's source database file: " + url;
+			throw new java.io.FileNotFoundException(message);
+		}
+
+		// open an empty one to get the full path and then close and delete it
```
sgtcoolguy: this comment seems to no longer describe reality here

sgtcoolguy (approved):

keerthi1032:
FR Passed. Ti.Database.install() not throwing error or exception. Works as expected. 
Test Environnent:
Operating System
Node.js
Titanium CLI
Titanium SDK =local sdk 8.1.0.v20190501151420
Cli = 7.0.11-70X.2
Device =NEXUS 5 android 5,Samsung s5 android 6
Emulator=pixel 3XL android 9

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
jquick-axway: Admittedly, I haven't really changed this code block other than its indentation and added extra null checks.

Having a harder look at it now, this can be handled better. It should go through our `TiFileFactory` class which is where we handle all of our internal customer URL schemes if given. From this we can fetch the filesystem path if applicable.

jquick-axway: I've confirmed that `install()` method was missing support for URLs referencing files on the file system such as `file://` and  `appdata-private://`. The `open()` method didn't support any URLs file referencing schemes at all. I've updated both to support this and have already tested it on my end.

android/modules/database/src/java/ti/modules/titanium/database/DatabaseModule.java
jquick-axway: Fair enough. I've modified the code to only resolve/normalize the URL if the `KrollInvocation` object is not `null`.

keerthi1032:
Tested updated PR. FR passed. No error or exception on Ti.Database.install() 
Operating System
Name = Mac OS X
Version = 10.13.6
Architecture = 64bit
Node.js
Node.js Version = 8.9.1
npm Version = 5.5.1
Titanium CLI
CLI Version = 5.1.1
Titanium SDK =local sdk 8.1.0.v20190515211536
Cli = 7.0.11
Device =NEXUS 5 android 5,Samsung s5 android 6
Emulator=pixel 3XL android 9,pixel xl android 8

PR #10771, opened 2019-03-12
fix(ios): delay js object creation to prevent accidental GC

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26852

We used to re-submit the block that calls an event callback asynchronously to the main queue. This would push it to the back of the run loop. In very rare cases this can cause issues when JSCore runs a GC in the meantime because when the block finally executes the JS object could have been already finalized.


janvennemann:
After further investigation it turns out that this issue is more likely to occur during events or firing callbacks, but those are not the real reason.

Very similar to our previous JSCore crash this is related to JS object refs being created but not inserted in the JS object graph in time. As long as they are not protected during this time they are subject to GC collection. In our previous fix we addressed this for proxies but not for method calls.

Since there is no viable place tor protect/unprotect them this will now postpone the call to `JSObjectMake` until the JS object is really needed by delegating access through lazy getters.

iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollObject.m
```diff
@@ -148,9 +148,6 @@ void KrollInitializer(JSContextRef ctx, JSObjectRef object)
 
   if ([o isKindOfClass:[KrollObject class]]) {
     [o retain];
-    JSObjectRef propsObject = JSObjectMake(ctx, NULL, NULL);
-    JSObjectSetProperty(ctx, object, kTiStringTiPropertyKey, propsObject, kJSPropertyAttributeDontEnum, NULL);
-    [o setPropsObject:propsObject];
```
vijaysingh-axway: @janvennemann Why removed?

vijaysingh-axway (commented):
@janvennemann  Everything looks good. One minor note. Can you please check? Thanks!

iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollObject.m
janvennemann: This will now be created lazily in the getter for `propsObject`. The js object only needs to be created once `jsObject` or `propsObject` are used so i moved everything related to creating those objects into the corresponding getter.

vijaysingh-axway (approved):
CR passed.

ssjsamir (approved):
FR Passed, a reduction in crashes can be seen (e.g. against customer applications).

**Test environment** 
```
iPhone 6S plus (12.2 Sim)
APPC CLI: 7.0.11-1
Node.js Version: 8.9.1
Xcode 10.2.1
```

PR #10775, opened 2019-03-13
fix(android)(8_0_X): increase java max memory

garymathews:
- Cherry-pick of #10774 

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26898)


lokeshchdhry:
FR Passed.

Could not repro. But support was able to & was confirmed as fixed by support.

PR #10783, opened 2019-03-19
fix: update template for latest webpack and angular versions

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26798

Webpack dependency was already updated via Greenkeeper but the Webpack config template was never adjusted. This bumps all dependencies to the latest version and updates the Webpack template to be compatible with Webpack 4.x


janvennemann:
Closing in favor of our new Webpack build pipline coming with #11346. A follow up PR will update the template to work with the appcd plugin.

PR #10794, opened 2019-03-21
fix(ios)(8_0_X): Supported tel protocol similar to previous implementation

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26835


sgtcoolguy:
This should ideally have at least one unit test defined to confirm the fix/feature.

Is the use case trying to "open" a tel: style url to dial a phone number? Is the issue that it actually would exit out app with no way to return if you invoked it?

janvennemann (dismissed):
Please update the log message to explain that this a iOS only recommendation.

janvennemann:
@sgtcoolguy, to goal is that `tel:` style links will be properly handled and a user is able to dial the number. Currently is doing nothing and just ignoring the link click.

I 'm not sure this can be tested with a unit test since it requires user interaction. Or can we emulate user clicks in a unit test in our suite? cc @vijaysingh-axway to confirm.

ssjsamir (approved):
FR Passed: Able to get call pop menu from a webView

**Test Environment**
```
iPhone 6 (12.1)
APPC CLI: 7.0.10
Node.js Version: 8.9.1
Xcode 10.1
```

PR #10798, opened 2019-03-22
fix(ios)(8_0_X): Unnecessary import removed and code wrapped inside proper macro

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-26681


janvennemann (approved):
LGTM!

PR #10802, opened 2019-03-25
refactor(build): clean up sdk build scripts

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/

**Description:**
This is an internal dev refactoring of our build scripts. The goal was to move to using async/await rather than the `async` library, and to clean the code up substantially so that we can reduce duplication.


build/lib/android/ant.js
```diff
@@ -0,0 +1,87 @@
...
+async function getJavaHome() {
+	// TODO: Use same code as node-appc's jdk detection!
+	if (process.env.JAVA_HOME) {
+		return process.env.JAVA_HOME;
+	}
+
+	// TODO: Sniff possible locations on Windows?
+	if (os.platform() !== 'darwin') {
+		throw new Error('No JAVA_HOME environmnet variable set. Please set JAVA_HOME before building for Android.');
+	}
```
jquick-axway: The **best** way to do this on Windows is via a registry lookup. The JDK installer always writes the install path to the registry and you shouldn't assume it's installed to the default location under `C:\`.
https://docs.oracle.com/javase/9/install/installation-jdk-and-jre-microsoft-windows-platforms.htm

You'll want to look under registry key/folder:
`HKLM\Software\JavaSoft\Java Development Kit` _(JDK 8 and older)_
...or...
`HKLM\Software\JavaSoft\JDK` _(JDK 9 and newer)_

Under the above registry key, it'll have registry value `CurrentVersion` under it which will have the newest JDK version stored to it. Oracle/Sun officially supports multiple JDK installations on Windows side-by-side and `CurrentVersion` will indicate the newest version installed.

After reading the above value, you then open registry key/folder:
`HKLM\Software\JavaSoft\Java Development Kit\<CurrentVersion>`

Under the above registry key, it'll have registry value `JavaHome` which provides the installation path of that version of the JDK.

^ The above is how I used to do it on Windows natively in C/C++. It's been a while, but it looks like it hasn't changed much (thankfully) other than JDK 9 writing under "JDK" instead.

sgtcoolguy: Thanks for the info, yeah I assumed there was a way to do this, but seeing as how most of us build the SDK on Mac, I really worried more about a host macOS system when I wrote this.

We already have jdk/android sdk/ndk detection code baked into our cli across `node-titanium-sdk` and `node-appc`. I think longer term it'd be best to re-use that logic and enhance it as we can so that the fixes apply to our own SDK builds as well as actual app builds. I haven't gotten to that stage yet, just added TODOs.

jquick-axway: Yeah, probably not worth doing for an SDK build (it's a lot of work and testing is a pain). Yordan is the only dev on the team who builds with a Windows laptop, but he sets the `JAVA_HOME` environment variable.

I would hope we do something like the above for the CLI though.
(I wonder if there is an existing node module that does all this too.)

ewanharris: Just incase you wanna keep it in style with the promise based code you're going for, you may prefer [androidlib](http://github.com/appcelerator/androidlib), and [jdklib](https://github.com/appcelerator/jdklib) over node-appc/node-titanium-sdk. They're what the daemon plugins are using today

sgtcoolguy: I would love to use whatever it is our CLI/tooling will be using. I think there's a broader issue here that we're circling around which is that we are not using the same code constantly for these things:
- I wrote up some quick and dirty stuff for our SDK builds
- the titanium CLI uses `node-appc`/`node-titanium-sdk` to detect android sdks, java, ndk
- the daemon uses `androidlib`/`jdklib`

That's not good. If androidlib and jdklib are the new blessed way to detect this stuff, I'm ok with that but then we should be moving the SDK/CLI to use them as well.

But all of that is beyond the scope of this PR for now, I think.

sgtcoolguy:
Merged manually. I'll be opening up an architecture-related JIRA ticket to try and tackle the inconsistent usage of code for android sdk/ndk/jdk detection.

sgtcoolguy:
Here's the ticket to track for us to stop re-inventing the wheel: https://jira.appcelerator.org/browse/TIMOB-26944

PR #10805, opened 2019-03-25
perf: improve incremental app build times

janvennemann:
JIRA: https://jira.appcelerator.org/browse/TIMOB-27043

This is a follow up for #10789 which further improves the build time on incremental builds.

Some timings i made with our mocha test suite (average of three runs each):

| Scenario | Build time (master) | Build time (PR) |
| --- | --- | --- |
| Clean build | 24417 ms | 22856 ms |
| Incremental, no changes | 6779 ms | 645 ms |
| Incremental, one file changed | 7126 ms | 955 ms |

The clean build difference can be neglected i guess, that's probably just random due to other system usage. Incremental build time improvements however are huge!

I tried to change as little as possible and just moved the related code to a task that automatically does file change detection. The async stuff is now handled in promises and i refactored the existing code into the two methods `processJsFile` and `transformAndCopy`. The rest is just utility stuff for the task to load and save state and merging data from the previous run with incremental changes back into our builder.

This is currently iOS only but once we decide to go for this i'll update the Android build too.


iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,330 @@
...
+					};
+
+					this.builder.cli.createHook('build.ios.compileJsFile', this.builder, (r, from, to, cb2) => {
+						// Read the possibly modified file contents
+						const source = r.contents;
+
+						// If the file didn't change from previous run, return early
+						const currentHash = this.generateHash(source);
+						const previousHash = this.data.contentHashes[from];
+						if (previousHash && previousHash === currentHash) {
```
janvennemann: This is basically where all the magic happens. If the source file content did not change BEFORE we run `jsanalyze` just skip the whole thing, re-use the existing file and load used Ti symbols from previous run.

cb1kenobi: We cannot skip `jsanalyze`. We need to AST parse everything so we can detect all `Ti.*` and `Titanium.*` symbols and enable the various APIs such as Ti.Media. Same goes for Android (and maybe Windows?).

I highly recommend testing using iOS device builds of KitchenSink. There's a lot of sensitive code here.

janvennemann: I store the symbols at the end so they can be re-used on the next run. If nothing in the source file changed i don't see why we would need to run the `jsanalyze` again. We can simply [load the symbols](https://github.com/appcelerator/titanium_mobile/pull/10805/files#diff-597ce7d1e6d7ef73f52de3af52512094R309) from the last run and inject them back [into the builder](https://github.com/appcelerator/titanium_mobile/pull/10805/files#diff-597ce7d1e6d7ef73f52de3af52512094R98).

I tested this with our KitchenSink app using device builds and this works just fine. The symbols for unchanged files will be re-used from the last build and changed files will run `jsanalyze` again and have their Ti symbols updated. If a file gets removed then it's stored Ti symbols also get cleared.

sgtcoolguy: So I think the concern here is that we need to take into account more than just the original source file changing. Did any options baked into jsanalyze change? minify, transpile, etc. Basically anything that goes into `analyzeOptions`. Additionally, and this is implicit and not properly handled as-is, but the user could have a .babelrc/config that will get loaded any modify babel usage. I think @brentonhouse is doing this. We haven't officially supported/fixed handling that yet, but it's an expected use case. (right now I think for example we bomb out if a babel config ignores a file because transform returns null and we expect an object with a `code` property).

Second, is obviously we are generating both a transformed file and gathering Ti API usage. You need to handle "storing" both of those and handling merging the Ti API usage symbols stuff back into the build properly. I don't think that's happening correctly here.

janvennemann: The `jsanalyze` options are a good point. I can check those before the task starts and fallback to a full run if anything changed. I can also feed in possible .babelrc/config paths and force a full rebuild if any of those changed. So that's also not a problem once that is officially supported.

Regarding your second point, I'm doing exactly that. The TI API usage symbols are stored per file in the task data that get's written to disk and will be loaded and merged back into the builder on subsequent runs. The transformed file is preserved between builds anyway so it's also kind of "stored".

janvennemann:
Changes to `jsanalyze` options will now also be detected and a fallback to processing all JS files is done.

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+
+/**
+ * Task that processes JS files by applying several transforms and copying them to their
+ * destination inside the build directory.
+ */
+class ProcessJsTask extends IncrementalFileTask {
+	/**
+	 * Constructs a new processing task.
+	 *
+	 * @param {Object} taskInfo Configuration object for this task
```
sgtcoolguy: please document the optional/required properties on the taskInfo object

```suggestion
	 * @param {object} taskInfo Configuration object for this task
	 * @param {string} [taskInfo.name='process-js'] name for the task
	 * @param {string[]} taskInfo.jsFiles array of file paths to consider
	 * @param {string[]} taskInfo.jsBootstrapFiles array of bootstrap scripts to consider
	 * @param {object} taskInfo.builder the iOS Builder object driving the CLI build
	 * @param {boolean} taskInfo.builder.minifyJS whether to minify the JS files
	 * @param {boolean} taskInfo.builder.transpile whether to transpile the JS files
	 * @param {boolean} taskInfo.builder.sourceMaps whether to generate source maps
	 * @param {string} taskInfo.builder.deployType 'production' or 'development'
	 * @param {string} taskInfo.builder.xcodeAppDir path to app directory
	 * @param {string} taskInfo.builder.minSupportedIosSdk minimum iOS version to support in transpiling
	 * @param {string} taskInfo.sdkCommonFolder path to the SDK's common js API folder
```

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+		super(taskInfo);
+
+		this.jsFiles = taskInfo.jsFiles;
+		this.jsBootstrapFiles = taskInfo.jsBootstrapFiles;
+		this.builder = taskInfo.builder;
+		this.sdkCOmmonFolder = taskInfo.sdkCOmmonFolder;
+
+		const minify = this.builder.minifyJS;
+		const transpile = this.builder.transpile;
+		this.defaultAnalyzeOptions = {
```
sgtcoolguy: feels like this object should just get passed in whole, rather than re-generated from various values pulled from the builder

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+		this.resetTaskData();
+		return Promise.all(Array.from(this.inputFiles).map(filePath => this.processJsFile(filePath)));
+	}
+
+	/**
+	 * Does an incremental task run, processing only changed files.
+	 *
+	 * For backwards compatibiliy this will currently also call processJsFile for unchanged files.
+	 *
+	 * @param {Map} changedFiles Map of file paths and their current file state (created, changed, deleted)
```
sgtcoolguy: Map<string, string>

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+			}
+		}
+
+		const info = this.data.jsFiles[file];
+		if (info) {
+			delete this.data.contentHashes[info.src];
+			delete this.data.tiSymbols[info.dest];
+		}
+
+		return Promise.resolve();
```
sgtcoolguy: This method and return value seem self defeating. Doesn't this run sync and then return an empty Promise? Why not just make the method async so it basically wraps the body in a Promise? Or explicitly wrap the full body in a Promise?

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+	generateHash(value) {
+		return crypto.createHash('sha1').update(value).digest('hex');
+	}
+
+	/**
+	 * Loads data from the previous task run.
+	 *
+	 * @return {Boolean} True if the data was sucessfully loaded, false if not.
+	 */
+	loadTaskData() {
```
sgtcoolguy: make async?
```suggestion
	async loadTaskData() {
	  if (!await fs.exists(this.dataFilePath)) {
	    return false;
	  }
	  
	  try {
	    this.data = await fs.readJson(this.dataFilePath);
	    return true;
	  } catch (err) {
	    return false;
	  }
	}
```

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+	 * @return {Promise}
+	 */
+	loadResultAndSkip() {
+		const loaded = this.loadTaskData();
+		const fullBuild = !loaded || this.requiresFullBuild();
+		if (fullBuild) {
+			return this.doFullTaskRun();
+		}
+
+		return Promise.resolve();
```
sgtcoolguy: This seems self-defeating. Why not make the method async so it just does the right thing? Seems to me like in the non full build scenario it will run the first two lines sync and then return an empty Promise.

```suggestion
	async loadResultAndSkip() {
		const loaded = await this.loadTaskData();
		const fullBuild = !loaded || this.requiresFullBuild();
		if (fullBuild) {
			return this.doFullTaskRun();
		}
	}
```

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+						}
+
+						this.transformAndCopy(source, from, to, () => {
+							this.data.contentHashes[from] = currentHash;
+							cb2();
+						});
+					})(r, from, to, cb);
+				})(info.src, info.dest, resolve);
+			} catch (ex) {
+				ex.message.split('\n').forEach(m => this.logger.error(m));
```
sgtcoolguy: So I think this sort of code should only be in the builder itself. This should just throw an Error out and let the builder decide whether it wants to just log it, or log and exit the process forcibly.

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+			this.builder.jsFilesChanged = true;
+
+			done();
+		} catch (err) {
+			err.message.split('\n').forEach(m => this.logger.error(m));
+			this.logger.error(err.stack);
+			if (err.codeFrame) { // if we have a nicely formatted pointer to syntax error from babel, use it!
+				this.logger.error(err.codeFrame);
+			}
+			process.exit(1);
```
sgtcoolguy: again, I think the builder should be the one making the decision how to handle the Errors

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+	 * it to the destination path.
+	 *
+	 * @param {String} source JavaScript source
+	 * @param {String} from Path to the file that contains the JavaScript source
+	 * @param {String} to Path where the transformed source should be saved to
+	 * @param {Function} done Callback function
+	 */
+	transformAndCopy(source, from, to, done) {
+		// DO NOT TRANSPILE CODE inside SDK's common folder. It's already transpiled!
+		const transpile = from.startsWith(this.sdkCommonFolder) ? false : this.defaultAnalyzeOptions.transpile;
```
sgtcoolguy: extract out the result of `from.startsWith(this.sdkCommonFolder)` to a const to avoid calculating it twice (my bad!)

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+	/**
+	 * Transforms the given JavaScript source by transpiling and minifying it and then copying
+	 * it to the destination path.
+	 *
+	 * @param {String} source JavaScript source
+	 * @param {String} from Path to the file that contains the JavaScript source
+	 * @param {String} to Path where the transformed source should be saved to
+	 * @param {Function} done Callback function
+	 */
+	transformAndCopy(source, from, to, done) {
```
sgtcoolguy: make async, use the async fs methods in here, and avoid passing a callback

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+	 * Processes a single JavaScript file by applying several transforms and then copying it to
+	 * its destination path.
+	 *
+	 * If the content of the unprocessed JavaScript file did not change compared to the last run,
+	 * any further processing will be skipped.
+	 *
+	 * @param {String} filePathAndName Full path to the JavaScript file
+	 * @return {Promise}
+	 */
+	processJsFile(filePathAndName) {
```
sgtcoolguy: make the whole method async....

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+			}
+			this.builder.jsFilesEncrypted.push(file); // original name
+			file = file.replace(/\./g, '_');
+			info.dest = path.join(this.builder.buildAssetsDir, file);
+			this.builder.jsFilesToEncrypt.push(file); // encrypted name
+		}
+
+		return new Promise((resolve) => {
+			try {
+				this.builder.cli.createHook('build.ios.copyResource', this.builder, (from, to, cb) => {
```
sgtcoolguy: I'd like to break these nesting levels up and try to minimize the mismatches here between Promises and callback functions if we can. Basically extract these callback functions out to their own async methods (using async fs APIs), then have just one method where you interface between the Promises and callbacks (maybe just using `util.callbackify`/`util.promisify`?)

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+					const originalContents = fs.readFileSync(from).toString();
+
+					const r = {
+						original: originalContents,
+						contents: originalContents,
+						symbols: []
+					};
+
+					this.builder.cli.createHook('build.ios.compileJsFile', this.builder, (r, from, to, cb2) => {
```
sgtcoolguy: Also, dumb to never look into this myself, but do we need to create a hook *for every file*? Can't we generate these two hooks once, and just call/pass in different values when we invoke them?

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+		return crypto.createHash('sha1').update(value).digest('hex');
+	}
+
+	/**
+	 * Loads data from the previous task run.
+	 *
+	 * @return {Boolean} True if the data was sucessfully loaded, false if not.
+	 */
+	loadTaskData() {
+		if (!fs.exists(this.dataFilePath)) {
```
sgtcoolguy: does this even work properly? You're calling the async API but checking the result which is presumably a Promise (and therefore will always evaluate to "true")

iphone/cli/lib/process-js-task.js
```diff
@@ -0,0 +1,374 @@
...
+			return true;
+		} catch (e) {
+			return false;
+		}
+	}
+
+	/**
+	 * Saves current task data for reuse on next run.
+	 */
+	saveTaskData() {
```
sgtcoolguy: make async?
```suggestion
	async saveTaskData() {
		return fs.writeJson(this.dataFilePath, this.data);
	}
```

sgtcoolguy (changes requested):

sgtcoolguy:
Forgot to mention: I don't know how to avoid it, but I'm not a huge fan of having to pass in the builder object here and rely on all sorts of properties/values internal to it. In the past I ran into this as well where we pass this massive object holding all the state and it makes it near impossible to test if you want to just hand craft a simple mock object yourself.

janvennemann:
@sgtcoolguy this is ready for another review. I updated the Android build too and made it a universal task that works for both Android and iOS.

I wasn't able to decouple the builder from the task without creating a huge options object which basically was a "builder light" so i settled with just passing the complete build in anyway. The task requires so many options and even some functions from the builder that IMHO it doesn't made much sense to compose a new options object from it.

android/cli/commands/_build.js
```diff
@@ -2675,6 +2676,67 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+				};
+				return jsFilesInfo;
+			}, {}),
+			jsBootstrapFiles,
+			sdkCommonFolder: path.join(this.titaniumSdkPath, 'common', 'Resources'),
+			defaultAnalyzeOptions: {
+				minify: this.minifyJS,
+				transpile: this.transpile,
+				sourceMap: this.sourceMaps || this.deployType === 'development',
+				resourcesDir: this.xcodeAppDir,
```
sgtcoolguy: I don't think this field will exist on Android...?

Probably should be:
```suggestion
				resourcesDir: this.buildBinAssetsResourcesDir,
```

android/cli/commands/_build.js
```diff
@@ -2719,209 +2781,116 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
-					titaniumPrepHook(
-						path.join(this.platformPath, titaniumPrep),
-						args,
-						opts,
-						function (err) {
-							if (err) {
-								fatal(err);
-							}
-							next();
+				// windows 64-bit failed, try again using 32-bit
```
sgtcoolguy: this comment no longer matches reality. I believe the exact opposite is happening: we try 32-bit and then try 64-bit.

cli/lib/tasks/process-js-task.js
```diff
@@ -0,0 +1,381 @@
...
+	 * @param {String} [options.name='process-js'] Name for the task.
+	 * @param {String[]} options.inputFiles Array of input files this task will use.
+	 * @param {String[]} options.incrementalDirectory Path to a folder where incremental task data will be stored.
+	 * @param {Object} options.logger The logger instance used by this task.
+	 * @param {Object} options.builder iOS builder instance.
+	 * @param {String[]} options.jsFiles Array with info about JS files to resolve paths.
+	 * @param {String[]} options.jsBootstrapFiles Array of bootstrap scripts to consider. The task will directly modify this array.
+	 * @param {String} options.sdkCommonFolder Path to common JS files from the SDK.
+	 * @param {Object} options.defaultAnalyzeOptions Default configuration options for jsanalyze.
+	 * @param {Boolean} options.defaultAnalyzeOptions.minify Wether to minify the JS files or not.
```
sgtcoolguy: Whether (same in the next few instances as well)

cli/lib/tasks/process-js-task.js
```diff
@@ -0,0 +1,381 @@
...
+	 * @param {Boolean} options.defaultAnalyzeOptions.transpile Wether to transpile the JS files or not.
+	 * @param {Boolean} options.defaultAnalyzeOptions.sourceMaps Wether to generate source maps or not.
+	 * @param {String} options.defaultAnalyzeOptions.resourcesDir .
```
sgtcoolguy: This is used in node-titanium-sdk to know where to insert the polyfills, but is no longer used in master/8.1.0

cli/lib/tasks/process-js-task.js
```diff
@@ -0,0 +1,381 @@
...
+		}
+
+		const bootstrapPath = file.substr(0, file.length - 3);  // Remove the ".js" extension.
+		if (bootstrapPath.endsWith('.bootstrap') && !this.jsBootstrapFiles.includes(bootstrapPath)) {
+			this.jsBootstrapFiles.push(bootstrapPath);
+		}
+
+		const info = this.jsFiles[file];
+		if (this.builder.encryptJS) {
+			if (this.platform === 'ios') {
```
sgtcoolguy: May need to check/normalize from 'iphone' as well. I think the CLI still accepts that (and not sure it normalizes it to 'ios' itself).

sgtcoolguy (changes requested):

cli/lib/tasks/process-js-task.js
janvennemann: The CLI will normalize this to `ios`. Test with `ti build -p ios|iphone|ipad` and all resolved to `ios`.

sgtcoolguy (approved):

lokeshchdhry:
I am seeing the following timings building for KS V2:

**8.1.0 (no fix):**
Studio:
1. Clean  build - 27 sec.
2. Rebuild no change - 7s 180ms.
3. Change in 1 file - 6s 188ms.

CLI:
1. Clean  build - 23s 321ms.
2. Rebuild no change - 7s 136ms.
3. Change in 1 file - 6s 157ms.

**This PR:**
Studio:
1. Clean  build - 24s 328ms.
2. Rebuild no change - 3s 661ms.
3. Change in 1 file - 4s 511ms.

CLI:
1. Clean  build - 21s 773ms.
2. Rebuild no change - 2s 736ms.
3. Change in 1 file - 5s 556ms.

@janvennemann , I definitely see an improvement in the timings but do not see a vast improvement as you see in incremental builds. 

Studio Ver:     5.1.2.201903111843
OS Ver:         10.14.5
Xcode Ver:      Xcode 10.2.1
Appc NPM:       4.2.13
Appc CLI:       7.0.12-70X.4
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.10
Node Ver:       8.15.1
NPM Ver:        6.4.1
Java Ver:       10.0.2
IOS Simulator:  ⇨ iPhone 7 Plus (iOS 12.2)

lokeshchdhry:
FR Passed.

hansemannn:
@lokeshchdhry Most of the benefits of this PR can be seen on larger projects. We've cut our build times to at least 1/3 of the original build time.

whitfin:
I was actually working on some of these changes locally in the last few days because it's definitely causing slowness.

On my local project (which is small, but has a fairly large amount of node modules) it reduces my build time from ~200s to ~12s. Definitely a good improvement!

lokeshchdhry:
@janvennemann , Can you please resolve the conflicts. Thanks.

cli/lib/tasks/process-js-task.js
```diff
@@ -0,0 +1,381 @@
...
+const crypto = require('crypto');
+const fs = require('fs-extra');
+const jsanalyze = require('node-titanium-sdk/lib/jsanalyze');
+const path = require('path');
+const pLimit = require('p-limit');
+const { promisify } = require('util');
+
+const i18n = appc.i18n(__dirname);
+const __ = i18n.__;
+const limit = pLimit(8);
```
whitfin: Why we need to limit the promises at all? Are we not slowing things down artificially? Or is there some reason we need to do this? If so, why 8 as a limit?

hansemannn:
Can someone resolve the merge conflicts and merge this? We need to cherry pick this since months now and it's a pain.

hansemannn:
@sgtcoolguy The merge conflicts come from your env-var changes. I think @janvennemann is on holiday, so can you resolve those? I currently needed to fix the merge problems locally but reverted the env-changes there.

whitfin:
I resolved the merge conflicts and this should be good to merge. I'm still curious about why we need `p-limit` and why we need a limit on the Promise throughput at all, but I can save that question for another day.

cb1kenobi:
Packages like `p-limit` and `promise-limit` are essentially the same thing as `async.eachLimit()` where you want to limit the number of concurrent tasks. If you do a `Promise.all()` on many, many things, you could exhaust the memory and crash.

I've always put a limit when synchronizing an open-ended number of tasks.

I've never done any real world benchmarks, so I can't say what the upper limit is, but it probably depends on the task workload. If you're copying files, 20 may seem low, but 20k, sure seems high. I'd be interested to see some data on it.

whitfin:
@cb1kenobi sure, but 8 seems pretty low - I agree that tuning it would be nicer (in future). Given that most of the operations are just waiting on IO, there shouldn't be much impact of waiting on a fair bit more than that.

PR #10813, opened 2019-03-28
fix(ios): fix extension app group and entitlements support 

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26948

Fixes a typo where it was enabling the app group by setting `true` instead of `1`.

Also correctly reference the entitlements file, specifically when the entitlements is referenced by the `CODE_SIGN_ENTITLEMENTS` build setting that is a relative path from the extension project's root.


iphone/cli/commands/_build.js
```diff
@@ -3562,32 +3562,55 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
...
+						if (fs.existsSync(src)) {
+							extBuildSettings.CODE_SIGN_ENTITLEMENTS = '"' + path.join(ext.relPath, entFile) + '"';
+							targetInfo.entitlementsFile = path.join(this.buildDir, ext.relPath, entFile);
+						} else {
+							src = path.join(ext.basePath, targetName, entFile);
+							if (fs.existsSync(src)) {
+								extBuildSettings.CODE_SIGN_ENTITLEMENTS = '"' + path.join(ext.relPath, targetName, entFile) + '"';
+								targetInfo.entitlementsFile = path.join(this.buildDir, ext.relPath, targetName, entFile);
+							} else {
+								delete extBuildSettings.CODE_SIGN_ENTITLEMENTS;
```
janvennemann: I think we should spit out a warning here that we couldn't resolve the configured entitlements file instead of just silently deleting the `CODE_SIGN_ENTITLEMENTS` setting.

janvennemann (changes requested):

iphone/cli/commands/_build.js
cb1kenobi: That's what I originally thought too, but I realized that not every extension target has entitlements. For example, File Provider UI targets do not have entitlements. So, having a warning leads to false positives and that would be confusing.

janvennemann: True, but wouldn't `CODE_SIGN_ENTITLEMENTS` then be empty anyway and we could check for that?

cb1kenobi: We do! https://github.com/appcelerator/titanium_mobile/pull/10813/files#diff-f430483233aa01af5b10df8390f9635dR3582

janvennemann: Yes, so in that specific case it should be fine to print a warning, shouldn't it? If {{CODE_SIGN_ENTITLEMENTS}} is explicitly set, warn if we can't properly resolve it. If it's not set, e.g. for the File Provider UI extension, that code path is ignored anyway.

cb1kenobi: I added a warning. Should be good to go now.

janvennemann (approved):

keerthi1032:
FR Passed. App group showed and enabled correctly and entitlements files listed correctly. 
  Node.js
Titanium CLI
Studio =5.1.2.201903111843 
Titanium SDK
  SDK Version                 = local 8.0.1 sdk and 8.1 master sdk 
Cli =7.0.10

keerthi1032:
Danger is failing and not able to merge. can someone please resolve it

PR #10817, opened 2019-04-01
chore(android): update V8 to 7.3.492.26

garymathews:
- Update V8 runtime to `7.3.492.26`
  - Use snapshot blobs to improve startup time by 50%

[TIMOB-26957](https://jira.appcelerator.org/browse/TIMOB-26957)
[TIMOB-26909](https://jira.appcelerator.org/browse/TIMOB-26909)


sgtcoolguy:
I assume this breaks native modules (i.e. we need to hold for 9.0 and bump the module api version)?

Or did you manage to retain our patches to v8 that retained the removed APIs?

garymathews:
@sgtcoolguy Currently working on that, should be able to get backwards compatibility.

garymathews:
@sgtcoolguy Fixed native module compatibility, not sure why the tests are failing though.

sgtcoolguy:
@garymathews this still crashes the android emulator - presumably backwards native module compat is still broken?

sgtcoolguy:
Gary and I spoke about this on a back channel chat. Before merging I'd like to see the  generation of the snapshot header file automated.

Long-term I think that means mksnapshot binaries become packaged with the V8 libraries in the zip file on the v8_titanium side; and on this side, our build scripts to use the mksnapshot binaries to generate the header when the v8 library is updated and then hooked into our builds.

That does present an issue if we're not generating a windows compatible mksnapshot binary and devs try to build the SDK on Windows (like Yordan does).

garymathews:
Made the necessary changes to V8 so that:
- Modules compiled with 7.X.X - 8.0.X are compatible with 8.1.0
- Modules compiled with 8.1.0 are compatible with 7.X.X+

Once [7.3-lkgr](https://jenkins.appcelerator.org/job/titanium-sdk/job/v8_titanium/job/7.3-lkgr/8/) has built, we can re-build this PR.

build/lib/android/index.js
```diff
@@ -27,25 +28,32 @@ class Android {
...
-			'build.githash': this.gitHash,
-			'android.sdk': sdk.getAndroidSDK(),
-			'android.platform': sdk.getPlatformDir(),
-			'google.apis': sdk.getGoogleApisDir(),
-			'kroll.v8.build.x86': 1,
-			'android.ndk': this.androidNDK
-		};
-		return ant.build(ANDROID_BUILD_XML, [ 'full.build' ], properties);
+		// Clean build and download V8
+		await this.clean().catch(error => console.error(error));
```
sgtcoolguy: really don't need the catch chained here, it'll bubble up the Error if there is one. adding a catch will actually just log it and move on.

build/lib/android/index.js
```diff
@@ -27,25 +28,32 @@ class Android {
...
+
+		// Generate snapshots
+		const snapshot = require('./snapshot');
+		await snapshot.build().catch(error => console.error(error));
```
sgtcoolguy: same here, no need for a catch really.

build/lib/android/snapshot.js
```diff
@@ -0,0 +1,110 @@
...
+async function generateBlob(target) {
+	const V8_LIB_TARGET_DIR = path.resolve(V8_LIB_DIR, target);
+	const MKSNAPSHOT_PATH = path.join(V8_LIB_TARGET_DIR, 'mksnapshot');
+	const BLOB_PATH = path.join(V8_LIB_TARGET_DIR, 'blob.bin');
+	const args = [
+		'--startup_blob=' + BLOB_PATH
+	];
+
+	// Set correct permissions for 'mksnapshot'
+	fs.chmodSync(MKSNAPSHOT_PATH, 0o755);
```
sgtcoolguy: the method is async, so you could do `await fs.chmod(MKSNAPSHOT_PATH, 0o755);`

build/lib/android/snapshot.js
```diff
@@ -0,0 +1,110 @@
...
+
+	return new Promise((resolve, reject) => {
+
+		// Snapshot already exists, skip...
+		if (fs.existsSync(BLOB_PATH)) {
```
sgtcoolguy: this could be moved out of the Promise and use `await fs.exists(BLOB_PATH)`

build/lib/android/snapshot.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Generates V8Snapshots.h header from architecture blobs
+ * @returns {Promise<void>}
+ */
+async function generateHeader() {
+	const blobs = {};
+
+	// Load snapshots for each architecture
+	for (const target of TARGETS) {
```
sgtcoolguy: again this method is async, so you can use the async fs methods. 

Additionally, you could use probably refactor this very slightly to read in all the blobs in parallel by using `Promise.all(TARGETS.map());`

build/lib/android/snapshot.js
```diff
@@ -0,0 +1,110 @@
...
+		const V8_LIB_TARGET_DIR = path.resolve(V8_LIB_DIR, target);
+		const BLOB_PATH = path.join(V8_LIB_TARGET_DIR, 'blob.bin');
+
+		if (fs.existsSync(BLOB_PATH)) {
+			blobs[target] = Buffer.from(fs.readFileSync(BLOB_PATH, 'binary'), 'binary');
+		}
+	}
+
+	return new Promise(async (resolve, reject) => {
```
sgtcoolguy: I'm confused by the async here...

Generally I try to make use of util.promisify to handle older callback style APIs so I can avoid the hand-written Promise instances. so in this case you could do:
```javascript
const promisify = require('util').promisify; // at the top of the file
const renderFile = promisify(require('ejs').renderFile);
// ...
const output = await renderFile(path.join(__dirname, 'V8Snapshots.h.ejs'), blobs, {});
return fs.writeFile(path.join(ANDROID_DIR, 'runtime', 'v8', 'src', 'native', 'V8Snapshots.h'), output);
```

build/lib/android/snapshot.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Generates empty snapshot blobs for each supported architecture
+ * and creates a header to include the snapshots at build time.
+ * NOTE: SNAPSHOT GENERATION IS ONLY SUPPORTED ON macOS
+ * @returns {Promise<void>}
+ */
+async function build() {
+	return new Promise(async (resolve, reject) => {
```
sgtcoolguy: again, if the method is async, you shouldn't need to hand write a Promise instance:
```javascript
// Only macOS is supports creating snapshots
if (os.platform() !== 'darwin') {
    return;
}

// generate snapshots in parallel
await Promise.all(TARGETS.map(t => generateBlob(t)));
return generateHeader();
```

sgtcoolguy (approved):
The PR looks fine to me, passes all the tests and looks to be a nice performance boost - awesome work Gary!

The comments from me are just stylistic JS nitpicks

PR #10822, opened 2019-04-03
docs(typescript): update sdk version in typescript definition header

ewanharris:
Not sure whether this is what you had in mind but it looks like the docgen script only reads in the version for generating changes, figured it made sense to just require in the package.json in the write header func.

### DIFF
--- a/apidoc/Titanium/UI/ShortcutItem.yml
+++ b/apidoc/Titanium/UI/ShortcutItem.yml
@@ -21,7 +21,7 @@ methods:
     platforms: [android]
 
   - name: pin
-    description: Pin shortcut to launcher.
+    summary: Pin shortcut to launcher.
     platforms: [android]
     osver: {android: {min: "8.0"}}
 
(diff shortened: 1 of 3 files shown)


ewanharris:
Validating locally using the linting in DefinitelyTyped throws some errors, will fix those in this PR

janvennemann:
@ewanharris Are the changes to the .yml files requires to pass linting or why are they in this PR?

ewanharris:
@janvennemann, yes they are required to pass the linting step in DefinitelyTyped.

It was throwing the following errors
* `propertyList` didn't have typed array
* `description/pin/title` didn't have a jsdoc description when generated, the docgen typescript template pulls from summary which [per the tdoc spec is required](https://wiki.appcelerator.org/display/guides2/TDoc+Specification#TDocSpecification-PropertySpecification) and description is optional so I switched description to summary for those
* Double new line at the end of the file which is the change on line 384

janvennemann (approved):

PR #10840, opened 2019-04-10
fix(ios): module builds broken

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26981

**Optional Description:**
- Add redirecting headers to `iphone/include` that point at the real copies in `iphone/Classes`
- Add redirecting headers to `iphone/include` that point to real headers in TitaniumKit framework
- fixed module templates to point `FRAMEWORK_SEARCH_PATHS` at `iphone/Frameworks`
- added warning log to module build if `titanium.xcconfig` doesn't have value for `FRAMEWORK_SEARCH_PATHS` so dev is at least warned they made need to append the value if their build is failing

### DIFF
--- a/build/lib/ios.js
+++ b/build/lib/ios.js
@@ -54,17 +54,23 @@ class IOS {
 	}
 
 	async package(packager) {
-		// FIXME This is a hot mess. Why can't we place artifacts in their proper location already like Windows?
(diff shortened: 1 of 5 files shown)


janvennemann:
Should we copy headers from `Classes` too? So we can have one unified include path?

janvennemann (approved):
LGTM!

PR #10842, opened 2019-04-11
fix(android): module clean should delete only generated libraries

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26980

**Optional Description:**

Instead of deleting libs folder, delete lib<moduleid>.so files


sgtcoolguy:
Easiest way to test is to check out appcelerator.encryptedatabase, cd `android` `ti clean -p android`, it shouldn't wipe the `libsqlcipher.so` files in the lib subdirs.

PR #10845, opened 2019-04-12
chore(android): cleanup snapshot generation

garymathews:
- Address review comment from https://github.com/appcelerator/titanium_mobile/pull/10817

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26909)

### DIFF
--- a/build/lib/android/snapshot.js
+++ b/build/lib/android/snapshot.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const spawn = require('child_process').spawn; // eslint-disable-line security/detect-child-process
+const promisify = require('util').promisify;
+const exec = require('child_process').execFile; // eslint-disable-line security/detect-child-process
 const os = require('os');
 const fs = require('fs-extra');
 const path = require('path');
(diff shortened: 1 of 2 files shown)


build/lib/android/snapshot.js
```diff
@@ -65,26 +60,20 @@ async function generateHeader() {
...
-			}
-			fs.writeFileSync(path.join(ANDROID_DIR, 'runtime', 'v8', 'src', 'native', 'V8Snapshots.h'), output);
-			resolve();
-		});
-	});
+	console.log(`Generating V8Snapshots.h for ${Object.keys(blobs).join(', ')}...`);
+
+	// Generate 'V8Snapshots.h' from template
+	const output = await promisify(ejs.renderFile)(path.join(__dirname, 'V8Snapshots.h.ejs'), blobs, {});
+	return fs.writeFile(path.join(ANDROID_DIR, 'runtime', 'v8', 'src', 'native', 'V8Snapshots.h'), output);
```
sgtcoolguy: No need to address this, but just an FYI since I used to do the same thing. You don't need to worry about using '/' as the file separators in paths passed in to `path.join()` it does the right thing on Windows under the hood. I used to break each segment up live this thinking Node would be too stupid to apply the right separator otherwise, but it's not - it silently converts. And then you can avoid all the extra quoting/commas.

sgtcoolguy (approved):
👍

PR #10848, opened 2019-04-13
fix(android): Fixed bug where setting SearchBar "hintText" after window open crashes as of 7.0.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-25832

**Test:**
2. Verify that the app starts up without crashing.
3. Verify that the top `SearchBar` reads "Hint Text".

```javascript
var searchBar = Ti.UI.createSearchBar({
	top: 0,
	width: Ti.UI.FILL,
	height: "50dp",
});
window.add(searchBar);
window.add(Ti.UI.createLabel({ text: "SearchBar Test" }));
window.addEventListener("open", function(e) {
	searchBar.hintText = "Hint Text";
});
window.open();
```


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -347,7 +347,10 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
 		} else if (key.equals(TiC.PROPERTY_COLOR)) {
 			tv.setTextColor(TiConvert.toColor((String) newValue));
 		} else if (key.equals(TiC.PROPERTY_HINT_TEXT)) {
-			int type = proxy.getProperties().getInt(TiC.PROPERTY_HINT_TYPE);
+			int type = UIModule.HINT_TYPE_STATIC;
+			if (proxy.getProperties() != null) {
```
sgtcoolguy: assign to a temp var to avoid grabbing it via getter twice?

Also, we should consider using a null object pattern here to avoid the null check altogether. If a KrollProxy has a null `properties` field, can't we return an empty KrollDict and get the same behavior typically? (i.e. returns null for everything, returns default values if passed in)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -921,7 +924,11 @@ public void setAttributedStringHint(AttributedStringProxy attrString)
 	{
 		Spannable spannableText = AttributedStringProxy.toSpannable(attrString, TiApplication.getAppCurrentActivity());
 		if (spannableText != null) {
-			int type = getProxy().getProperties().getInt(TiC.PROPERTY_HINT_TYPE);
+			int type = UIModule.HINT_TYPE_STATIC;
+			KrollProxy proxy = getProxy();
```
sgtcoolguy: another potential place we could return a null object variant? i.e. if a TiUIView's proxy is null, return a special NullProxy instance that has empty properties?

tests/Resources/ti.ui.searchbar.addontest.js
```diff
@@ -0,0 +1,54 @@
...
+				return;
+			}
+			setTimeout(function () {
+				try {
+					should(searchBar.hintText).eql(NEW_HINT_TEXT);
+					finish();
+				} catch (err) {
+					finish(err);
+				}
+			}, 100);
```
sgtcoolguy: Shouldn't need to add any specific delay in milliseconds here, I wouldn't think. You could put 1 and it should basically allow the UI thread to dispatch before we fire. (also note that technically the delay arg is optional, and under the hood it enforces a minimum delay anyways - so maybe just omit the delay)

sgtcoolguy (approved):
LGTM

keerthi1032:
FR passed. Searchbar did not crash and works as expected.
Test Enviornment:
Operating System
Node.js
Titanium CLI
APPC CLI  =7.0.11-70X.1
Titanium SDK
  SDK Version                 = local 8.0.1.v20190423101725 and local master  8.1.0.v20190412182932 
Device=Oneplus 5t android 9,samsung s5 android 6
Emulator =pixel 2xl android 9,nexus android 6

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
jquick-axway: I was going to change this to use `proxy.getProperty(TiC.PROPERTY_HINT_TYPE)` instead which will return `null` if the properties dictionary is `null`, but I can see this PR is already merged. That said, I think the current code is fine anyways.

PR #10850, opened 2019-04-15
chore(deps): update dev dependencies

ewanharris:
Includes update to danger that should fix the status checks issue we're seeing.

Specifically see https://github.com/danger/danger-js/blob/master/CHANGELOG.md#712 
 https://github.com/danger/danger-js/pull/854


ewanharris:
clang-format seems to be enforcing the following (curly on same line) across _all_ iOS files pretty much. Backing that update out
```diff
- (void)dealloc
- {
+ (void)dealloc {
  [self stopGeneratingEvents];
  [super dealloc];
}
```

PR #10856, opened 2019-04-19
feat(android): add tabMode to TabGroup

m1ga:
https://jira-archive.titaniumsdk.com/TIMOB/TIMOB-26990

Switch between fixed and srollable tabMode:

```javascript

});
var btn1 = Ti.UI.createButton({
	title: "fixed",
	top: 5
});
var btn2 = Ti.UI.createButton({
	title: "scrollable",
	top: 60
});
win.add([btn1, btn2]);
	openTabGroup(Ti.UI.Android.TAB_MODE_FIXED);
})
	openTabGroup(Ti.UI.Android.TAB_MODE_SCROLLABLE);
})
win.open();

function openTabGroup(tabModeValue) {

	var tabs = [];

	for (var i = 0; i < 10; ++i) {
		var w = Ti.UI.createWindow({});
		var t = Ti.UI.createTab({
			window: w,
			title: 'win'
		});
		tabs.push(t);
	}

		tabs: tabs,
		tabMode: tabModeValue
	});
	tabGroup.open();
}
```

https://user-images.githubusercontent.com/4334997/160287343-79483db5-d6ea-42de-929f-c8a2b30282c1.mp4


sgtcoolguy:
cc @ypbnv

apidoc/Titanium/UI/TabGroup.yml
```diff
@@ -443,12 +443,27 @@ properties:
...
     type: String
     since: {android: "3.0.0", iphone: "3.0.0", ipad: "3.0.0"}
     platforms: [android, iphone, ipad]
 
+  - name: tabMode
+    summary: Property defining which tabMode is used for the TabGroup.
+    description: |
+        Sets the style of the TabGroup tabs.
+
+        [TAB_MODE_FIXED](Titanium.UI.Android.TAB_MODE_FIXED) is the style default style and will show all tabs at once.
```
ypbnv: One of the "style" words is redundant.

ypbnv (changes requested):
Thank you for the contribution! Just a single comment in the documentation and it will need a bump of the version number in there - 8.1.0 should be changed to 8.3.0.

m1ga:
changed the text and the version number

apidoc/Titanium/UI/Android/Android.yml
```diff
@@ -837,13 +837,27 @@ properties:
     platforms: [android]
     since: "8.0.0"
 
+  - name: TAB_MODE_FIXED
+    summary: Set the TabGroup tab mode to fixed (default).
+    type: Number
+    availability: creation
+    platforms: [android]
+    since: "8.1.0"
```
ypbnv: This version number should also be 8.3.0.

apidoc/Titanium/UI/Android/Android.yml
```diff
@@ -837,13 +837,27 @@ properties:
...
+
+  - name: TAB_MODE_SCROLLABLE
+    summary: Set the TabGroup tab mode to scrollable.
+    type: Number
+    availability: creation
+    platforms: [android]
+    since: "8.1.0"
```
ypbnv: This version number should also be 8.3.0.

ypbnv:
@m1ga Thanks! Can you update the version number in the other two places I have commented on?

m1ga:
missed them, sorry! Updated too

jquick-axway:
@m1ga, @ypbnv,

Our `TabGroup` in Titanium 7.x.x scrolled horizontally by **default**, but our new top tab style in 8.0.0 no longer scrolls without this tab mode. Which means we did not maintain the old behavior. (I didn't notice this until now, so, it got past me too.)

So, do you think we need a new "tabMode" property at all? I'm thinking we should make the top tabs scrollable by default and drop this new property if we can.

Perhaps a better question is would anyone want to use the "static" tab mode setting where it shows ellipsis if there isn't enough room for all tabs (this kind of looks bad).

m1ga:
@jquick-axway I can change the default value but it would be nice to have this option e.g. if you tabs are only some pixel too big it will be scrollable and with fixed you set it to be static.

m1ga:
I've changed the default value to scrollable again (inside the XML that caused it to be fixed: https://github.com/appcelerator/titanium_mobile/blame/1a2343d1500a47ea5adabe8a52fcc077dd4b5169/android/modules/design/res/values-land/values-land.xml)

ypbnv:
@jquick-axway , @m1ga To me it feels OK to have the default value match the old behavior ( I have no clue, why I did not pay attention to that doing the refactoring ) and have a new property exposed. I understand it does not make much sense parity-wise, but for edge cases like the one pointed out by Michael it would be useful.

And that rises a question - having a different default tabMode from the refactoring until now - which functionality would it make more sense to keep backward compatible?

jquick-axway:
The only thing that bugs me is that this new "tabMode" property only applies to Android's top tab style. It seems like a waste to me, which is why I was thinking about simplifying it. Hmm...

jquick-axway:
@m1ga, @ypbnv, does setting tab mode to "fixed" make the tabs fully span the width of the screen when in landscape mode? Or in portrait mode on a tablet? (I haven't tested this for myself yet.)

I ask because one thing that has changed as of Titanium 8.0.0 when we switched to the Java `TabLayout` class is that the top tabs do not fully fill the width of the screen in portrait mode on a tablet... versus the old tab group behavior did. I don't think it's a big deal, but there have been a couple app devs who have complained about it.

What I've seen on stackoverflow below suggests that this will do the trick...
https://stackoverflow.com/questions/30843775/tab-not-taking-full-width-on-tablet-device-using-android-support-design-widget

m1ga:
There is a difference between opening the Window in landscape mode vs turning it while open:
http://migaweb.de/20190822_192849.mp4

If you open it in portrait fixed mode and turn it it will take the whole space. If you open it in landscape fixed mode it will be centered.
Scrolling mode is always aligned to the left side

m1ga:
closing in favor of https://github.com/tidev/titanium-sdk/pull/13877

PR #10862, opened 2019-04-22
fix(ios): don't add exception handler to js context, breaks usage when timers throw exceptions

sgtcoolguy:

**Description:**

From testing with the `mocha` package off npm and other node compatibility improvements locally, I found that the
exception handler hung here on timers results in the test suite behaviing oddly. It looks like it catches the exception,
spits out details on the log but doesn't propagate it and then the mocha test suite gets confused about it's lifecycle.

I think we may still want an exception handler generally that fills out the nativeStack details and then propagates the modified error?


janvennemann:
@sgtcoolguy yeah i think we should add a general exception handler to our context. Have you tried moving it to the `KrollBrdige` or `KrollContext`? Do you wan't to handle that with this PR too? Or should we do this in a separate ticket?

PR #10863, opened 2019-04-22
feat(process): add initial implementations of setImmediate and process.nextTick

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26574

**Description:**
This is an implementation of `process.nextTick()` and global `setImmediate`/`clearImmediate`

I tested this usage in our mocha test suite with other node compatibility changes (some of the `fs` library, modified `require()`)

The two implementations now *do* work in concert. ticks are drained fully in a while loop. `Immediates` are done afterwards with some caveats:
- we give `Immediates` a 100ms window to process the queue. After that we basically exit the loop and use setTimeout to re-schedule processing the ticks/`Immediates` remaining.
- we check if any ticks have been added after an `Immediate` is run. If so, we immediately drain the tick queue fully again before proceeding in our while loop to process `Immediates` (and again, the 100ms window limitation may cause us to not process any more immediate right now but do a setTimeout to re-schedule draining the queues)


sgtcoolguy:
Follow-on to  #10782

common/Resources/ti.internal/extensions/node/process.js
```diff
@@ -147,7 +147,9 @@ process.emitWarning = function (warning, options, code, ctor) { // eslint-disabl
 	}
 	this.emit('warning', warning);
 };
-process.env = {};
+process.env = {
+	// DEBUG: '*' // uncomment to make use of npm 'debug' module. TODO: have CLI pass along env vars in some file/property for dev builds?
```
hansemannn: Would be pretty cool to pass environment settings! In our backend, we use https://github.com/motdotla/dotenv for handling environments. And for production builds, things like API keys could be managed by this.

sgtcoolguy: Ah, yeah that's a nice solution for production builds. Yeah I was trying to debug some low-level node-api compatibility issues in using mocha on our test suite and hacked in the `process.env.DEBUG` value here for it and then realized it'd be super-useful to hook our CLI up to generate some files that passed along the ENV for development builds so you could do `DEBUG=* ti build -p ios` and it just magically worked. I'll probably open a ticket and add that.

common/Resources/ti.internal/extensions/node/process.js
```diff
@@ -253,3 +255,103 @@ Ti.App.addEventListener('uncaughtException', function (event) {
 });
 
 export default process;
+
+// Use a nice predictable class/structure for our timers
+// JS engine should be able to optimize easier
+class Timer {
```
janvennemann: This name is a little misleading since it's not really a timer. It's a structure to hold a callback function and its arguments so something like `BoundCallback` would fit better IMHO.

In Node the object is called `Immediate` but since we use it for both `setImmediate` and `nextTick` that might cause confusion (although it's only used internally for `nextTick`).

common/Resources/ti.internal/extensions/node/process.js
```diff
@@ -253,3 +255,103 @@ Ti.App.addEventListener('uncaughtException', function (event) {
...
+	constructor(fun, args) {
```
janvennemann: `func` is a more clear abbreviation here

```suggestion
	constructor(func, args) {
```

janvennemann (dismissed):
Just two minor notes on some variable/class naming. The rest looks good!

sgtcoolguy:
manually merged after rebasing to single commit (and addressing Jan's feedback)

PR #10866, opened 2019-04-25
Update RecurrenceRule.yml

jawa9000:
Fix for missing Android platform for `daysOfTheWeekDictionary` as requested in TIDOC-3283.


apidoc/Titanium/Calendar/RecurrenceRule.yml
```diff
@@ -102,7 +102,7 @@ properties:
 name: daysOfTheWeekDictionary
 summary: Dictionary containing `daysOfWeek` and `week`.
 since: "3.1.0"
```
hansemannn: This should also be changed, since Android only supports it since 7.3.0, see https://github.com/appcelerator/titanium_mobile/pull/9412 for details.

PR #10872, opened 2019-05-01
[TIMOB-27049] fix(ios): call super after implementation in "sendEvent:" delegate selector

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27049


janvennemann (approved):
CR passed!

hansemannn:
Can someone please merge this? It's a language-specific fix that should be pretty clear. Rather than that, there won't be a test case for this since it very rarely happens.

PR #10893, opened 2019-05-15
fix(ios): remove busy loop and change waitForMemoryPanicCleared to noop since it's always on main thread

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27080

**Description:**
Was previously used in kroll thread entry points to wait for main thread to handle memory warnings.
Now it will just result in an infinite sleep loop because JS runs on main thread.

All credit goes to @janvennemann for finding this, once he pointed it out to me I was able to do a local SDK build with it as a no-op and confirm the fix on @hansemannn 's Lambus app.

PR #10894, opened 2019-05-15
fix(ios): remove busy loop and change waitForMemoryPanicCleared to noop since it's always on main thread (8_0_X)

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27080

**Description:**
Was previously used in kroll thread entry points to wait for main thread to
handle memory warnings. Now it will just result in an infinite sleep loop
because JS runs on main thread.

This is an `8_0_X` backport of #10893

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -798,8 +798,11 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 											_t.keystoreAliases.pop();
 										}
 
+										// Parse the keystore's alias name and signature algorithm.
+										// Note: Algorithm can return "MD5withRSA (weak)" on JDK 8 and higher.
+										//       Only extract 1st token since we need a valid algorithm name.
 										const aliasRegExp = /Alias name: (.+)/,
-											sigalgRegExp = /Signature algorithm name: (.+)/;
+											sigalgRegExp = /Signature algorithm name: (.[^\s]+)/;
 										out.split('\n\n').forEach(function (chunk) {
 											chunk = chunk.trim();
 											const m = chunk.match(aliasRegExp);
@@ -807,7 +810,7 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 												const sigalg = chunk.match(sigalgRegExp);
 												_t.keystoreAliases.push({
 													name: m[1],
-													sigalg: sigalg && sigalg[1]
+													sigalg: sigalg && sigalg[1] && sigalg[1].trim()
 												});
 											}
 										});
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1857,7 +1857,6 @@ AndroidBuilder.prototype.initialize = function initialize(next) {
 		return appc.string.capitalize(word.toLowerCase());
 	}).join('');
 	/^[0-9]/.test(this.classname) && (this.classname = '_' + this.classname);
-	this.mainActivity = '.' + this.classname + 'Activity';
 
 	this.buildOnly = argv['build-only'];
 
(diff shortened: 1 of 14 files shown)


janvennemann (approved):

ssekhri:
Still see the issue with the SDK build fetched from Jenkins for this PR. It consistently crashes when scrolling through the trips (had around 15 trips created and it crashes after scrolling through 8-10 trips). Following logs were shown in console during one of the crash
`
[DEBUG] :  Firing app event: memorywarning
[DEBUG] :  Firing app event: memorywarning
[DEBUG] :  Firing app event: memorywarning
[ERROR] :  The application has crashed with an uncaught exception 'NSGenericException'.
[ERROR] :  Reason:
[ERROR] :  *** Collection <__NSCFDictionary: 0x1754ea40> was mutated while being enumerated.
[ERROR] :  Stack trace:
[ERROR] :  0   CoreFoundation                      0x212ff933 <redacted> + 150
[ERROR] :  1   libobjc.A.dylib                     0x20a9ae17 objc_exception_throw + 38
[ERROR] :  2   CoreFoundation                      0x212ff3a1 <redacted> + 0
[ERROR] :  3   TitaniumKit                         0x01962cc3 -[KrollBridge didReceiveMemoryWarning:] + 196
[ERROR] :  4   CoreFoundation                      0x212b1735 <redacted> + 12
[ERROR] :  5   CoreFoundation                      0x212b113f <redacted> + 390
[ERROR] :  6   CoreFoundation                      0x212b0f1d <redacted> + 40
[ERROR] :  7   CoreFoundation                      0x21307c6b <redacted> + 1334
[ERROR] :  8   CoreFoundation                      0x21211083 _CFXNotificationPost + 486
[ERROR] :  9   Foundation                          0x21a525df <redacted> + 74
[ERROR] :  10  Foundation                          0x21a57113 <redacted> + 30
[ERROR] :  11  UIKit                               0x25b26813 <redacted> + 166
[ERROR] :  12  libdispatch.dylib                   0x20e6d80f <redacted> + 22
[ERROR] :  13  libdispatch.dylib                   0x20e85db3 <redacted> + 2030
[ERROR] :  14  libdispatch.dylib                   0x20e80d19 <redacted> + 712
[ERROR] :  15  libdispatch.dylib                   0x20e7b73f <redacted> + 394
[ERROR] :  16  CoreFoundation                      0x212c1b6d <redacted> + 8
[ERROR] :  17  CoreFoundation                      0x212c0067 <redacted> + 1574
[ERROR] :  18  CoreFoundation                      0x2120f229 CFRunLoopRunSpecific + 520
[ERROR] :  19  CoreFoundation                      0x2120f015 CFRunLoopRunInMode + 108
[ERROR] :  20  GraphicsServices                    0x227ffac9 GSEventRunModal + 160
[ERROR] :  21  UIKit                               0x258e3189 UIApplicationMain + 144
[ERROR] :  22  Lambus                              0x000d0871 Lambus + 51313
[ERROR] :  23  libdyld.dylib                       0x20eb7873 <redacted> + 2
`

Env:
SDK: 8.0.2.v20190522114611, 8.1.0.v20190515121638
Device: iPad mini (v9.3.5)
XCode: 10.2.1
tested using Lambus app versioned 1.2.0.9

sgtcoolguy:
I pushed another commit to try and address the issue @ssekhri saw - looks like when we get a memory warning, we notify the proxies. If they happen to unregister as a result, I suspect the error he saw occurs as we're iterating over the dictionary containing the proxies.

The "fix" is to iterate over a copy of the keys instead.

As an aside, @janvennemann and @vijaysingh-axway  I am a little confused about ow this method works. Looks like we get a memory warning and start notifying proxies, but we stop if the proxy count doesn't change - so if a proxy doesn't do anything about it, why wouldn't we let the other proxies handle it too?

vijaysingh-axway:
> I pushed another commit to try and address the issue @ssekhri saw - looks like when we get a memory warning, we notify the proxies. If they happen to unregister as a result, I suspect the error he saw occurs as we're iterating over the dictionary containing the proxies.
> 
> The "fix" is to iterate over a copy of the keys instead.
> 
> As an aside, @janvennemann and @vijaysingh-axway I am a little confused about ow this method works. Looks like we get a memory warning and start notifying proxies, but we stop if the proxy count doesn't change - so if a proxy doesn't do anything about it, why wouldn't we let the other proxies handle it too?

@sgtcoolguy 
Inside 'for loop' we are sending  memory warning to all registeredProxies. (for loop inside while loop)
If any proxy from registeredProxies get unregistered, while we are sending memory warning,  we restart sending memory warning to all proxies inside registeredProxies. 
In this way we are sending memory warning to other proxies. 
One issue I can see is, if any proxy get unregistered while we are sending  memory warning, this will again send memory warning to all registeredProxies though we have already sent warning to some proxies.
Am I missing anything?

sgtcoolguy:
@ssekhri Did the latest commit fix the issue?

ssekhri:
@sgtcoolguy I will check the latest commit. Unfortunately the ticket is still with In Review status hence didnt get the focus

ssekhri:
@sgtcoolguy I tried the commit on the PR (8.0.2.v20190530051406) and tried on three different iOS devices. 
One device i.e. iPad mini (v9.3.5) on which I reported the issue earlier still shows the crash however did not see any issues when working with iPhone5(v10.3.3) and iPhone 6S(v12.1.4)

sgtcoolguy:
merged manually.

PR #10895, opened 2019-05-15
feat(process): load ENV variables into process.env

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27045

**Description:**
- write _env_.json file with ENV variables
- lazily load those values into process.env at runtime


android/cli/commands/_build.js
```diff
@@ -2825,6 +2825,19 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
 			delete this.lastBuildFiles[appPropsFile];
 
+			// write the env variables file
+			const envVarsFile = path.join(buildAssetsPath, '_env_.json');
+			Object.keys(this.tiapp.properties).forEach(function (prop) {
+				props[prop] = this.tiapp.properties[prop].value;
+			}, this);
+			fs.writeFileSync(
+				envVarsFile,
+				// for non-development builds, DO NOT WRITE OUT ENV VARIABLES TO APP
```
janvennemann: Maybe i'm missing something here, but why is this disabled for production builds? What if i want to use environment variables to control certain things in my app?

Also, the `deployType` for device builds is `test` which would make this unreliable when switching between sim and device builds during development.

janvennemann (dismissed):

android/cli/commands/_build.js
sgtcoolguy: So my basic concern here was that people tend to forget and have some identifying information, api keys, passwords, whatever in their environment variables and sneaking those by default into the production/released build of their app seems like a security accident waiting to happen. I think for production builds developers will likely want to be much more particular about what actually goes in. For those cases, I suspect they'd want an approach more like https://github.com/motdotla/dotenv

But, you're saying when we target a physical device from `ti build` the `deployType === 'test'`? If os, then I guess I can just change the guard here to be `deployType !== 'production'`

sgtcoolguy:
For QE testing:

run a build with something like:
```
MY_ENV_VAR=testing ti build -p ios
```
And an app.js with:
```js
Ti.API.info(process.env. MY_ENV_VAR);
```
It should print out "testing" in the log.

ssjsamir (approved):
**FR Passed** Able to load ENV variables into process.env.

**Test Steps:**

1. Created a new Titanium application with the build from this PR
2. Added the following in to the app.js:
`Ti.API.info(process.env.MY_ENV_VAR);`
3. Ran the following command:
`MY_ENV_VAR=testing ti build -p ios`
4. Able to see the following in the console:
`[INFO]  testing`

**Test Environment** 
```
iPhone 6S plus (12.2 Sim)
APPC CLI: 7.0.12-70X.4
Node.js Version: 8.11.3
Xcode 10.2.1
```

hansemannn:
I think this pull request broke incremental builds. Can someone confirm? I noticed full rebuilds on every incremental build since last week. It can also be #10805 but I think it worked fine with that one.

PR #10897, opened 2019-05-15
[TIMOB-27082] fix(ios): restore SDK < 8.1.0 compatibility

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27082

Note: There may be more classes that are affected by moving to the JSCore ObjC API, so this may need to be extended further. We're currently testing all cases.

cc @sgtcoolguy 

**EDIT**: No more affected cases for us.


janvennemann:
@sgtcoolguy i did a quick check of #10381 and it looks like a few more proxies are affected:

- TiCalendarAlert
- TiCalendarAttendee
- TiCalendarCalendar
- TiCalendarEvent
- TiCalendarRecurrenceRule

sgtcoolguy:
@hansemannn Technically, anything that went from TiProxy -> ObjcProxy hierarchy would be affected in this way, but obviously we don't expect people to instantiate any modules as they're intended to be singletons.

So I guess that'd mean these particular types are affected:
- TiBlob
- TiDatabaseProxy
- TiDatabaseResultSetProxy
- TiCalendarAlert
- TiCalendarAttendee
- TiCalendarCalendar
- TiCalendarEvent
- TiCalendarRecurrenceRule

And just adding back the old init methods seems to fix the issue for you?

hansemannn:
Yep, that works, because the linked (compiled) modules only search for the selectors and performs them if found. For classic proxies, having the default `_initWithPageContext:` constructor in `TiProxy` will fix most use-cases. In `TiBlob`, having custom constructors that accepted more than the context caused the issues  and this PR resolves those.

sgtcoolguy:
@hansemannn I've added back the old init page context methods for the other proxies I listed above. Can you confirm this works fine for you, and then I'll merge it?

iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
```diff
@@ -111,6 +111,26 @@ - (NSUInteger)size
 }
 GETTER_IMPL(NSUInteger, size, Size);
 
+- (id)_initWithPageContext:(id<TiEvaluator>)pageContext
+{
+  return [[TiBlob alloc] init];
```
hansemannn: The default `_initWithPageContext: `constructor should be re-added to the parent class (`TiViewProxy`) again, since other proxies (especially in modules) may be constructed that way as well.

sgtcoolguy: ok, I can do that too, I suppose.

Related: why do we need to alloc calls here? Typical usage before was:
```obj-c
[[TiBlob alloc] _initWithPageContext:context];
```

so should the redirecting version just be:
```obj-c
- (id)_initWithPageContext:(id<TiEvaluator>)pageContext
{
  return [self init];
}
```

hansemannn: You're right, no alloc required!

iphone/Classes/TiCalendarRecurrenceRule.m
```diff
@@ -15,6 +15,11 @@ @implementation TiCalendarRecurrenceRule
 
 #pragma mark - Internals
 
+- (id)_initWithPageContext:(id<TiEvaluator>)context rule:(EKRecurrenceRule *)rule_
+{
+  return [self initWithRule:rule];
```
hansemannn: @sgtcoolguy `rule_`

hansemannn: You can run the project locally in iphone/iphone/Titanium.xcodeproj.

PR #10898, opened 2019-05-17
build(jenkins): add timeout for adb commands

janvennemann:
The adb commands will wait for a device when none was found which blocks the build forever.

See [this](https://jenkins.appcelerator.org/blue/organizations/jenkins/titanium-sdk%2Ftitanium_mobile/detail/PR-10668/7/pipeline) build for an example.

@sgtcoolguy i guess this shouldn't happen in the first place. Are there some nodes that don't have an emulator installed?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -108,9 +108,11 @@ def unitTests(os, nodeVersion, npmVersion, testSuiteBranch) {
 							} finally {
 								// Kill the emulators!
 								if ('android'.equals(os)) {
-									sh returnStatus: true, script: 'adb -e shell am force-stop com.appcelerator.testApp.testing'
-									sh returnStatus: true, script: 'adb -e uninstall com.appcelerator.testApp.testing'
-									killAndroidEmulators()
+									timeout(5) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -108,9 +108,11 @@ def unitTests(os, nodeVersion, npmVersion, testSuiteBranch) {
 							} finally {
 								// Kill the emulators!
 								if ('android'.equals(os)) {
-									sh returnStatus: true, script: 'adb -e shell am force-stop com.appcelerator.testApp.testing'
-									sh returnStatus: true, script: 'adb -e uninstall com.appcelerator.testApp.testing'
-									killAndroidEmulators()
+									timeout(5) {
```
sgtcoolguy: may only want to wrap these two add commands and not the killAndroidEmulators() call, which I think just does killall and shouldn't block. (Also looks like only the uninstall command actually blocks)

sgtcoolguy:
All the nodes should have emulators, so something is up with ginsu-macos02 reporting no emulators found... cc @ahutton1217

PR #10901, opened 2019-05-19
fix(ios): force alert() messages to be a string

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27090

No unit tests since it's a UI related change.


sgtcoolguy:
sucks that we can't basically programmatically close the alert so that we could test this...

sgtcoolguy (approved):

hansemannn:
Is there a timeline when the Appium suite will be deployed? I'd have a good bunch of tests (especially for sensors / geolocation) that could benefit from it directly

ssekhri:
FR passed.
The computed alert message is converted to string and then displayed in alert. No app crash or issues with message shown in alert noticed.
Verified on:
Mac OS:10.14.3
Ti SDK: 8.1.0.v20190523110359
Appc CLI: 7.0.11
XCode: 10.2.1

PR #10902, opened 2019-05-19
fix(ios): properly retain geolocation callback

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27089

Fixes a regression from #10381.

Currently not unit-testable, since the geolocation permissions dialog required user interaction. I used the static code analyzer to make sure there is no retain leak.


sgtcoolguy (approved):
LGTM

PR #10905, opened 2019-05-22
fix(ios): exclude all frameworks from optimizing files step

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27097

**Optional Description:**
If a framework contains .png or .plist files our build will try to optimize those. Xcode will update code signing on the first build but fails to do so on subsequent builds. Frameworks should be left untouched to avoid this issue.


iphone/cli/commands/_build.js
```diff
@@ -6495,7 +6495,7 @@ iOSBuilder.prototype.optimizeFiles = function optimizeFiles(next) {
 				}
 			}
 		});
-	}(this.xcodeAppDir, /^(PlugIns|Watch|TitaniumKit\.framework)$/i));
+	}(this.xcodeAppDir, /^(PlugIns|Watch|\.framework)$/i));
```
cb1kenobi: Do you mean `/^(PlugIns|Watch|.+\.framework)$/i`?

iphone/cli/hooks/frameworks.js
```diff
@@ -579,7 +579,7 @@ class FrameworkIntegrator {
 			isa: 'PBXBuildFile',
 			fileRef: fileRefUuid,
 			fileRef_comment: frameworkPackageName,
-			settings: { ATTRIBUTES: [ 'CodeSignOnCopy' ]  }
+			settings: { ATTRIBUTES: [ 'CodeSignOnCopy', 'RemoveHeadersOnCopy' ] }
```
cb1kenobi: I believe Xcode would automatically add `RemoveHeadersOnCopy` when you opened the generated Xcode project. I tested explicitly adding it as you've done, but as I recall it wasn't necessary, so I ripped it out. I'm fine with you adding it.

cb1kenobi (changes requested):

iphone/cli/hooks/frameworks.js
hansemannn: Just tested this pull - it works fine without this setting.

janvennemann: I just added it for consistency because that's the attributes we use when adding TitaniumKit as well.

iphone/cli/commands/_build.js
janvennemann: Oh yeah sure, didn't notice the `^` at the beginning of the regex.

janvennemann:
@cb1kenobi ping! Can you give this another look? Thanks!

cb1kenobi (approved):

ssekhri:
FR Passed. The project with a framework having png/plist files builds successfully each time.
Mac OS 10.14.3
Ti SDK: 8.2.0.v20190709154013
Appc CLI: 7.0.12
Node: 10.13.0
Xcode: 10.2.1

PR #10907, opened 2019-05-23
fix(ios): don't log error about missing use of filesystem api

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27099

**Optional Description:**
Backport/cherry-pick of #10906 

when trying to access Ti.Blob.file without a backing file/path we'd erroneously spit an error about
not using filesystem API


vijaysingh-axway (approved):
CR passed.

ssekhri:
FR Passed. No errors logged during debug.

PR #10920, opened 2019-05-29
feat(android): implement async Ti.Database.DB methods

garymathews:
- Implement asynchronous `Ti.Database.DB` methods for executing SQL queries.
  - `Ti.Database.DB.executeAsync(query, parameters[], callback)`
  - `Ti.Database.DB.executeAllAsync(queries[], callback)`
- Implement method for executing multiple queries in one shot. (Useful for transactions)

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27111)


android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+	public void executeAllAsync(final KrollFunction callback, final String[] queries)
+	{
+		executingQueue.set(true);
+		try {
+			queue.put(new Runnable() {
+				@Override
+				public void run()
+				{
+					final Object[] results = executeAll(queries);
+					callback.callAsync(getKrollObject(), new Object[] { results });
```
sgtcoolguy: does results need to be "double-wrapped" here? I assume otherwise it'd expand out and the callback would get separate arguments for each results rather than a single array?

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+					} catch (InterruptedException e) {
+					}
+				}
+			});
+			this.thread.start();
+		}
+
 		this.name = name;
 		this.db = db;
-		statementLogging = false;
```
sgtcoolguy: why did we have these fields/flags? Did we have a documented way to open a Database to be read-only? It looks like it only protected against removal, but not actual data manipulation....

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+	 * @param db SQLiteDatabase instance.
+	 */
 	public TiDatabaseProxy(String name, SQLiteDatabase db)
 	{
 		super();
+
+		if (this.thread == null) {
+
+			// Query execution thread.
+			this.thread = new Thread(new Runnable() {
```
sgtcoolguy: when do you clean up this thread? I assume once a DB is closed, this should be interrupted and stop running.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 @Kroll.proxy(parentModule = DatabaseModule.class)
 public class TiDatabaseProxy extends KrollProxy
 {
 	private static final String TAG = "TiDB";
 
+	private static Thread thread;
```
sgtcoolguy: why are the thread/queue static? Shouldn't they be per-database?

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
-			sqlArgs = (Object[]) args[0];
-		} else {
-			sqlArgs = args;
+		// Validate and parse parameter objects.
+		if (parameterObjects != null && parameterObjects.length == 1 && parameterObjects[0] instanceof Object[]) {
+			parameterObjects = (Object[]) parameterObjects[0];
 		}
+		for (int i = 0; i < parameterObjects.length; i++) {
+			if (parameterObjects[i] instanceof TiBlob) {
+				parameterObjects[i] = ((TiBlob) parameterObjects[i]).getBytes();
```
sgtcoolguy: This is a behavioral change. We should add some unit tests around this. This is probably the right thing to do for "binary" blobs, but may be wrong and breaking behavior for blobs holding string contents.

Compounding this is the ambiguous case of "file" blobs whose contents may be binary or not.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+	}
+
+	/**
+	 * Asynchronously execute a single SQL query.
+	 * @param callback Result callback for query execution.
+	 * @param query SQL query to execute on database.
+	 * @param parameters Parameters for `query`
+	 */
+	@Kroll.method
+	public void executeAsync(final KrollFunction callback, final String query, final Object... parameters)
```
sgtcoolguy: So in JS do we actually have to pass in the callback as the first argument here, like so?
```javascript
const db = Ti.Database.open('whatever');
db.executeAsync(result => { Ti.API.info(result); }, 'SELECT * FROM TABLE WHERE name = ?', myName);
```

Because if so, that is very odd and we should avoid it if possible. I'm assuming the use of the varargs for parameters here is why the callback needed to come first? Can we re-arrange it so the callback is the last JS arg? If we can't "natively" with our binding layer, we could add JS code to basically do the swap under the hood in the same way we do some hacking of say Ti.App.Properties here: https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/app/src/js/properties.js

sgtcoolguy: Second, it's also not idiomatic to have the async variant have the extra name suffix.

Assuming we can hack around the arguments in JS or whatever, why not just use the same method name and if the last argument is a callback function then treat it as async, otherwise sync?

I suppose that boxes us in if we want to return a Promise...

Maybe introduce a db.executeSync, db.executeAsync where they are explicitly sync/async, have the existing method do either sync or async based on if callback is passed in, and spit out a deprecation warning if no callback is passed to it (basically saying, hey in the future this will return a Promise - so if you want sync behavior use db.executeSync)? I think at some point we want to remove the default sync nature of the method and push people towards async naturally - and eventually support Promises.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
+		} catch (InterruptedException e) {
+		}
+	}
+
+	/**
+	 * Synchronously execute a multiple SQL queries.
+	 * @param queries SQL queries to execute on database.
+	 */
+	@Kroll.method
+	public Object[] executeAll(final String[] queries)
```
sgtcoolguy: We should look to something like this node sqlite3 package to consider how they've approached their API: https://github.com/mapbox/node-sqlite3/wiki/API#database

Maybe not introduce prepare/Statements, but having explicit methods to say whether you want to run something without caring about the result, or wanting the first result, or calling the callback with the batch results once vs calling for each row seems useful to me.

Note that there is also this package which does explicitly run sync intentionally for performance: https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/api.md

I recall @jquick-axway mentioning usage of transactions helped performance significantly.  Would we wrap this batch of queries in a transaction for the user? Would we expect them to include their own transaction management statements in this batch?

sgtcoolguy (changes requested):

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
```diff
@@ -18,145 +19,244 @@
...
-			Log.d(TAG, "Database is not open, ignoring close for " + name, Log.DEBUG_MODE);
 		}
 	}
 
+	/**
+	 * Synchronously execute a single SQL query.
+	 * @param query SQL query to execute on database.
+	 * @param parameterObjects Parameters for `query`
+	 */
 	@Kroll.method
```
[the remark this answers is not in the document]
garymathews: Good catch, didn't intent to make this `private`

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
garymathews: Yes it does, each element of the `new Object[] { }` is a parameter in the callback. This will return a callback with a parameter containing an array of results.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
garymathews: No idea, it was intended for internal use. But it's unnecessary.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
garymathews: The idea was to have a single background thread for all TiDatabase instances, to conserve order (Multiple TiDatabase proxies can refer to the same database). But on second thought, it's not necessary anymore. So I've reverted to a thread for each TiDatabase instance.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
garymathews: This is actually the same code as what's [here](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java#L136). I've just moved it.

android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
garymathews: I've modified the parameters so the callback is the last parameter, we need to keep `Ti.Database.execute()` as is to not cause breaking changes.

```
Ti.Database.DB.executeAsync(query, parameters[], result => { ... });

results[] = Ti.Database.DB.executeAll(queries[]);
Ti.Database.DB.executeAllAsync(queries[], results[] => { ... });
```

I'd prefer to introduce as few API changes possible until we make the jump to Promise based APIs in the near future. I didn't intend to refactor all of our Ti.Database.DB APIs with this PR.

I think these APIs allow the flexibility to do the primary things, including transactions if the user chooses to do so. We should create a ticket to refactor the methods in `9.0.0`?

sgtcoolguy (approved):
LGTM.

My only remaining feedback here is that the batch execute methods only take strings and there's no way to do the string + params style you'd do in a single query. I think that'd be a fairly common annoyance here, and you may want to consider being more expansive on how to handle elements in the `queries` array so it's a `Object[]` and could hold either a single `String` in each entry, or another `Object[]` with first element the sql `String`, the remaining the varargs params.

Something like:
```javascript
const queries = [
  'DELETE FROM testTable',
  ['INSERT INTO testTable (text, number) VALUES (?, ?)', testName, testNumber]
]
```

lokeshchdhry:
FR Passed.

`executeAsync`, `executeAll` & `executeAllAsync` work as expected.

Studio Ver:     5.1.3.201906102126
SDK Ver:        8.2.0 local build
OS Ver:         10.14.5
Xcode Ver:      Xcode 10.2.1
Appc NPM:       4.2.13
Appc CLI:       7.0.12
Daemon Ver:     1.1.3
Ti CLI Ver:     5.1.1
Alloy Ver:      1.13.10
Node Ver:       8.15.1
NPM Ver:        6.4.1
Java Ver:       10.0.2
Devices:        ⇨ google Pixel (Android 9)

lokeshchdhry:
@garymathews , Its complaining about unit tests. Does it need any ?

PR #10922, opened 2019-05-29
fix(ios)(8_0_X) : Added log functionality

vijaysingh-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27054

### DIFF
--- a/iphone/Classes/TiUIWebView.m
+++ b/iphone/Classes/TiUIWebView.m
@@ -444,7 +444,7 @@ static NSString *const baseInjectScript = @"Ti._hexish=function(a){var r='';var 
     window.webkit.messageHandlers._Ti_.postMessage({name:'warn', method: 'log', callback: Ti._JSON({level:'warn', message:message},1)},'*'); \
     }, \
     log: function(level, message){ \
-    window.webkit.messageHandlers._Ti_.postMessage({name:'level', method: 'log', callback: Ti._JSON({level:'level', message:message},1)},'*'); \
+    window.webkit.messageHandlers._Ti_.postMessage({name: level, method: 'log', callback: Ti._JSON({level: level, message:message},1)},'*'); \
     }, \
     }; \
     ";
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIWebView.m
```diff
@@ -443,6 +443,9 @@ - (WKUserScript *)userScriptTitaniumInjectionForAppEvent
     warn: function(message){ \
     window.webkit.messageHandlers._Ti_.postMessage({name:'warn', method: 'log', callback: Ti._JSON({level:'warn', message:message},1)},'*'); \
     }, \
+    log: function(level, message){ \
+    window.webkit.messageHandlers._Ti_.postMessage({name:'level', method: 'log', callback: Ti._JSON({level:'level', message:message},1)},'*'); \
```
sgtcoolguy: You're passing in the string literal `'level'` instead of the actual value of the `level` argument.

Also is there any way to add a test for this? Open a Ti.UI.WebView, have it call Ti.API.log('warn', 'message'); and verify it actually occurred somehow? (I'm guessing not, since it goes directly to our APIModule and we can't intercept it)

sgtcoolguy (changes requested):

iphone/Classes/TiUIWebView.m
vijaysingh-axway: Good catch. Thanks!
I don't think we can write test case for same reason as you have mentioned. Probably we can add event listener in APIModule and fire some event when log api get called. For this we have  to make some modification inside APIModule. I am not sure if it is right way.

keerthi1032:
FR Passed. logs shown as expected in console .
Test Environment:
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.0.2.v20190530104542
CLi =7.0.11
Device =iPhone 6s ios12 
Simulator =iphone xr ios12

keerthi1032:
Jenkin failed. Not able to merge PR. waiting for resolving jenkin's failure

PR #10924, opened 2019-05-30
build(android): verify the libv8 tar.bz2 file against a checksum

sgtcoolguy:
**Description:**
To avoid 'bad' builds of libv8 being used, verify not only the version/mode
from package.json, but also a SHA512 checksum hash of the tar.bz2 file.

This is an attempt to help avoid the bad builds/crashes we saw yesterday on Android. Now if the libv8 library content doesn't match expectations we'll bomb out immediately during build time and let the developer know the libv8 looks wrong and show the expected/actual hashes.

This does mean that when we update the v8 build we target/use we'll need to not just bump the version but also ensure the checksum is updated.

### DIFF
--- a/android/build/common.xml
+++ b/android/build/common.xml
@@ -773,6 +773,10 @@ Common ant tasks and macros for building Android-based Titanium modules and proj
 			<arg value="-p"/>
 			<arg value="require('../package.json').v8.mode"/>
 		</exec>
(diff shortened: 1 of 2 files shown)


garymathews (approved):
CR: **PASS**

But one of our build nodes is having trouble launching the emulator

PR #10932, opened 2019-06-04
feat(ios) : add support for Xcode 11 / iOS 13 dev environment

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27125


janvennemann:
@vijaysingh-axway are there any changes suggested by Xcode to our project templates, like adjusting build settings? If so we should also apply those.

vijaysingh-axway:
> @vijaysingh-axway are there any changes suggested by Xcode to our project templates, like adjusting build settings? If so we should also apply those.

@janvennemann  I don't see anything like that.

vijaysingh-axway:
@sgtcoolguy/@janvennemann  Can you please review this? If everything looks fine can it be merged please. Thanks!

hansemannn:
This one is useful for our upcoming pull requests as well, especially the new precompile flag.

iphone/iphone/Titanium_Prefix.pch
```diff
@@ -11,10 +11,10 @@
 #include "ThirdpartyNS.h"
 #import <AvailabilityInternal.h>
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
-#define IS_XCODE_9 true
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+#define IS_XCODE_11 true
```
janvennemann: Since this is a new constant i would propose to change the naming. This is not really a check for the Xcode version but rather for the version of the iOS SDK that is used to compile the code.

Although a specific version fo the SDK  usually is tied to a specific Xcode version, i find it more expressive to check for the SDK version. That way one doesn't have to remember the Xcode and iOS SDK version number coupling.

```suggestion
#define IS_IOS_SDK_13 true
```

iphone/iphone/Titanium_Prefix.pch
```diff
@@ -11,10 +11,10 @@
 #include "ThirdpartyNS.h"
 #import <AvailabilityInternal.h>
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000
-#define IS_XCODE_9 true
```
janvennemann: Since you removed the Xcode 9 check this means we will bump the minimum supported Xcode version to 9, right?

janvennemann (changes requested):

iphone/iphone/Titanium_Prefix.pch
vijaysingh-axway: Yes.

iphone/iphone/Titanium_Prefix.pch
vijaysingh-axway: I agree with your naming convention. We have other flag IS_XCODE_10 as well. I think we should change naming for both at same time to avoid any confusion. I think creating an internal ticket to change constant name, will be good way to deal with this.

janvennemann: Sounds good. Please change the name in this PR for the new constant already.

janvennemann (approved):

iphone/package.json
```diff
@@ -19,8 +19,8 @@
 	"minIosVersion": "9.0",
 	"minWatchosVersion": "2.0",
 	"vendorDependencies": {
-		"xcode": ">=8.0 <=10.x",
-		"ios sdk": ">=9.0 <=12.x"
+		"xcode": ">=9.0 <=11.x",
```
ewanharris: Does raising this and the below `ios sdk` values cause breakage? As I understand it this is used to handle our minimum versions right?

PR #10945, opened 2019-06-06
fix(android): Fixed incremental build issues introduced by [TIMOB-27043] in 8.1.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27135

**Summary:**
- This bug was caught before release.
- Fixed build failure when a local HTML file references a local JS file.
  * Added unit test to exercise this feature on Android and iOS. (Not supported on Windows.)
- Fixed incremental build failure when building for "device" or "production".
  * This is really an encrypted assets issue.

**Test:**
1. Create a Classic app project from default template.
2. Build and run to an Android device. (Not the emulator. Must be a device.)
3. Build and run to the same Android device again.
4. Verify that the build succeeds and runs successfully on device. (This used to fail.)

PR #10951, opened 2019-06-07
fix(ios): expose --source-maps CLI flag, default to source maps for non-production builds

sgtcoolguy:

**Description:**
- Create a `--source-maps` CLI flag to force generation of source maps
- fix `--hide-error-controller` to be a flag, not an option (internal refactoring, it's a flag used by test suite so we can continue despite errors)
- Change source mapping behavior to behave like so:
  - if `--source-maps` flag is passed to CLI, turn them on
  - if tiapp.xml has a `source-maps` tag respect it's value
  - if neither, turn on source maps by default on non-`'production'` (deploy type) builds

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -923,7 +923,16 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 
 	// Transpilation details
 	this.transpile = cli.tiapp['transpile'] !== false; // Transpiling is an opt-out process now
-	this.sourceMaps = cli.tiapp['source-maps'] === true || cli.argv['source-maps'] || this.deployType !== 'production'; // opt-in to generate inline source maps
+	// If they're passing flag to do source-mapping, that overrides everything, so turn it on
+	if (cli.argv['source-maps']) {
+		this.sourceMaps = true;
+		// if they haven't, respect the tiapp.xml value if set one way or the other
+	} else if (cli.tiapp.hasOwnProperty['source-maps']) { // they've explicitly set a value in tiapp.xml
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
cc @brentonhouse Can you try this build? Does it help/fix the problems?

sgtcoolguy:
relates to appcelerator/node-titanium-sdk#106 and appcelerator/titanium_studio#1134

jquick-axway:
@sgtcoolguy, are you seeing any incremental build issues on iOS when enabling/disabling `source-maps`?

I was running into this issue on Android. The build directory's JS files were not being re-processed. This is because our `ProcessJsTask`'s base class currently only process the JS files if 1 of the source JS files have changed. It doesn't check if the `encryptJS` or `sourceMaps` settings have changed since the last build. As a quick solution on Android, I've modified "_build.js" to do a full rebuild if these 2 settings have changed. It's a bit heavy handed, but it'll do for the moment... and I would expect these settings to rarely change too.
https://github.com/appcelerator/titanium_mobile/pull/10945/commits/062194a1eb22db44a7a2030e1199c811f77e6ea1

sgtcoolguy:
@jquick-axway I haven't looked into it deeper. I was doing `ti clean` between builds whenever I tweaked the logic in our source mapping (from appcelerator/node-titanium-sdk#106) because otherwise no JS files changed, and no setting changed so it wouldn't use my new logic for generating source maps. 

It should be doing a full build if the jsanalyze options change: https://github.com/appcelerator/titanium_mobile/blob/master/cli/lib/tasks/process-js-task.js#L323-L327 which includes the sourceMaps options, but I don't see it checking about the `this.encryptJS` flag.

PR #10954, opened 2019-06-10
fix: pass along eventual destination for js file in jsanalyze

sgtcoolguy:

Relates to appcelerator/node-titanium-sdk#106

**Optional Description:**
Passes in to the jsanalyze options a `dest` property pointing at the eventual file path of the generated JS file we write after transpilation.


sgtcoolguy:
no longer needed

PR #10961, opened 2019-06-14
[TIMOB-27150] Add ability to alias required modules in Titanium

brentonhouse:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27150

This PR adds the bindings/redirects to global so that they are accessible externally for aliasing modules.


common/Resources/ti.internal/extensions/binding.js
```diff
@@ -83,3 +83,8 @@ export function redirectCoreModuleIdToPath(coreModuleId, internalPath) {
 export function bindObjectToCoreModuleId(coreModuleId, object) {
 	bindings.set(coreModuleId, object);
 }
+
+global.redirectCoreModuleIdToPath = redirectCoreModuleIdToPath;
```
sgtcoolguy: So this is a useful method internally for me to hijack require to expose our shims via the "core" names used by Nodes.

I worry, because I never actually wrote this or intended it as publicly available API. This would re-expose it, but I never put any thought into making a special new API to do this - or whether we'd want to.

I would think we'd want to at least refactor it to some better API naming/usage if we do want to expose it.

sgtcoolguy (changes requested):

PR #10962, opened 2019-06-14
fix(android): prevent multiple exception logs

garymathews:
- Prevent multiple unnecessary `TiExceptionHandler` logs when handlers are present

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27153)


jquick-axway (approved):
CR: Pass

ssjsamir (approved):
**FR Passed** Their is no longer multiple unnecessary TiExceptionHandler logs when handlers are present. (Only 1 set now). Tested with the following test case:

- Include {{com.appcelerator.aca}} module
- Throw an exception
`throw new Error('exception');`
- Should NOT see multiple {{TiExceptionHandler}} logs




**Test Environment**
```
Pixel XL (7.1)
7.0.12-70X.4
Node.js Version: 8.11.3
```

PR #10972, opened 2019-06-18
refactor(android): consolidate incremental build fixes back into task

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27160

**Optional Description:**
Refactors the workarounds from #10945 and moves the related logic back into the task.

- Removes manual force rebuild checks, changes for the `sourceMaps` and `transpile` flags will be checked here: https://github.com/appcelerator/titanium_mobile/blob/673a3b9117e3d185a09882b5f0409ead47454179/cli/lib/tasks/process-js-task.js#L323-L327
- Removes manual deletion of the incremental data dir and now correctly checks the `forceRebuild` flag on Android


jquick-axway (approved):
CR: Pass

jquick-axway:
I've tested this PR on Android as follows. Everything is working:
- Incremental builds with deployment type "test". (This encrypts JS files.)
- Did a rebuild twice with deployment type "test".
- Incremental builds with deployment type "development". (Plain JS files.)
- Did a rebuild twice with deployment type "development".
- Incremental build after changing "transpile" setting with deployment type "development".
- Incremental build after changing "source-maps" setting with deployment type "development".
- Incremental build after changing Titanium SDK version. (Will do rebuild.)

The only interesting thing I noted was that source maps were always included at the bottom of every JS files when I set "transpile" to `false` and "source-maps" to `false` in my "tiapp.xml" file. Hoewever, I've noticed this happens in Titanium 8.0.0 as well. So, this doesn't appear to be a new issue. @janvennemann , have you noticed the same? (Not related to your incremental build changes of course.)

sgtcoolguy:
@jquick-axway I suspect that the issue you mention about ignoring the tiapp.xml value for source maps is fixed by #10951

Before we turned on source maps if the deploy type was `'development'` or the tiapp.xml value was true. Now it should behave like so:
- if you pass along a `--source-maps` flag to the CLI, it'll turn them on (and ignore tiapp.xml's value)
- if not, and there is a value in tiapp.xml use that value
- if neither, turn on if deploy type is not `'production'`

jquick-axway:
Thanks @sgtcoolguy . Right, I see that you've changed the following line in that PR. That's definitely it.
```javascript
sourceMap: this.sourceMaps || this.deployType === 'development'
```

lokeshchdhry:
@sgtcoolguy , when I set "source-maps" to "false" & build it to simulator(IOS) which is "development" I still do see the source map getting generated.

janvennemann:
@lokeshchdhry should be fixed with #11076

lokeshchdhry:
FR Passed.

Studio Ver:     5.1.3.201907112159
SDK Ver:        8.2.0 local build
OS Ver:         10.14.5
Xcode Ver:      Xcode 11.0
Appc NPM:       4.2.15-1
Appc CLI:       7.1.1-master.5
Daemon Ver:     1.1.3
Ti CLI Ver:     5.2.1
Alloy Ver:      1.14.1
Node Ver:       10.16.1
NPM Ver:        6.9.0
Java Ver:       10.0.2
Devices:        ⇨ google Pixel (Android 9)
Emulator: Android 7.1.1

PR #10975, opened 2019-06-18
fix(ios): fixed debug log server issues (TIMOB-27074)

cb1kenobi:
fix(ios): Fixed bug where the log server was being started when the iOS app was quitting instead of stopping it.
fix(ios): Changed the log server port test to use 127.0.0.1 instead of localhost to avoid bad dns lookup.
fix(ios): Fixed log server port test timeout to properly handle the timeout. (TIMOB-27074)

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27074


janvennemann:
@cb1kenobi , i can still reproduce the issue with the following steps:

1. Build and run any app using {{ti build -p ios}}
1. Put the app into background and quit the ti build command
1. Build and run the app again using {{ti build -p ios}}
1. The build fails with the following error:
```
[DEBUG] Checking if log server port 59584 is available
[DEBUG] Log server port 59584 is in use, testing if it's the app we're building
[ERROR] Another process is currently bound to port 59584
[ERROR] Set a unique <log-server-port> between 1024 and 65535 in the <ios> section of the tiapp.xml
```
1. Put the app back into foreground, build and run again with {{ti build -p ios}}. This will now work again since the app is not in the background anymore.

cb1kenobi:
@janvennemann OK, I walked through things with @ewanharris and @vijaysingh-axway and came up with a totally rad solution. Give it a spin.

janvennemann:
Hmm, so this will still fail the build, just because the app is backgrounded? This seems like an unnecessary inconvenience to me. As a user i don't want to manually bring back the app to foreground or kill it just so i can start a build.

Is there any chance we could improve this process on the simulator? Like alternating the ports between builds slightly so the running app in the background does not block the entire build? Or sending a proper quit message before exiting from our CLI that will make the app stop listening if no other connections are open?

cb1kenobi:
@janvennemann Just chatted with @ewanharris and talked about the tradeoffs that were made with the way things are today. I proposed that we revisit these tradeoffs in the next architecture meeting and come to a consensus how best to handle this going forward.

cb1kenobi:
@janvennemann OK, I've updated the code to pick a random port if the deterministic port is unavailable. Please give it a spin!

hansemannn:
This works fine, we're using it in our fork already. Hope this can be merged on master soon!

sgtcoolguy (approved):
LGTM

ssjsamir (approved):
**FR Passed** Rebuilding the app with simulator no longer fails due to log port being occupied. 

**Tested with the following steps:** 
1. Build and launch app on Simulator
2. Put the app in the background
3. Rebuild and see the error happening
4. Bring the app to foreground, do another rebuild and everything works again

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 10.2.1
Node.js ^8.12.0
iPhone 6s Sim(12.1).
"NPM":"4.2.14-3","CLI":"7.0.10"
```

sgtcoolguy:
I cloned this branch, fixed the ESLint issues, rebase down to single commit that matched our conventions and merged it to master.

PR #10976, opened 2019-06-18
fix(ios): allow changing WebView read access when loading local file

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27159


apidoc/Titanium/UI/WebView.yml
```diff
@@ -170,6 +170,22 @@ methods:
     since: {android: "8.1.0", iphone: "2.0.0", ipad: "2.0.0"}
     platforms: [android, iphone, ipad]
 
+  - name: loadFile
```
hansemannn: Is this for loading html files? Or for other resources?

vijaysingh-axway: Both.

janvennemann:
@vijaysingh-axway does this need parity with Android?

vijaysingh-axway:
> @vijaysingh-axway does this need parity with Android?

Android do not have  anything similar. @jquick-axway has mentioned in comment  of this ticket.

apidoc/Titanium/UI/WebView.yml
hansemannn: All html files should be handled via "url" which is already pretty smart handling different types of urls. In additions, assets should be loaded through the bridge between the WKWebView and the web-view handler.

jquick-axway:
When setting this to a local HTML file, wouldn't everyone want to set `allowingReadAccessToURL` to the HTML file's directory?

I don't see anyone ever setting it to a different directory. That doesn't seem practical.

Also, I can seeing this being as a non-stop tech-support issue for us. We would end up telling everyone to call the `loadFile()` method and pass it the HTML file's directory. Also, only being able to do this via a `loadFile()` method isn't Alloy friendly where we want to use properties instead.

I'm leaning toward us automatically setting `allowingReadAccessToURL` to the local HTML's file directory. This makes it backward compatible, avoids tech-support issues, and has feature parity between platforms.

Is there any reason not to set this automatically?

janvennemann:
@jquick-axway The ticket author mentioned it would be nice to change `allowingReadAccessToURL` so that different HTML files in the cache dir could reuse shared resources like css that are already in the app's resources folder.

However, i think it would be best to copy the local CSS files to the cache dir and then set `allowingReadAccessToURL` automatically. This is basically what you are doing in your example posted in the ticket, correct @vijaysingh-axway? I was also wondering if this could be integrated in `url` property setter so we don't need a new method?

sgtcoolguy:
So my 2 cents here: I really don't like the idea of adding some epochal extra method to web view that is both iOS-specific **and* specific to a particular use case of files in the app cache. 

As @jquick-axway @hansemannn and @janvennemann are alluding to, we should try and "do the right thing" under the hood in our existing APIs (like the url property).

vijaysingh-axway:
Agreed. What if I modify the [setUrl](https://github.com/appcelerator/titanium_mobile/blob/b735f547a94b35a1d8ac899580879df4f751c364/iphone/Classes/TiUIWebView.m#L176) to take extra optional parameter  'readAccessUrl' similar to [setHtml](https://github.com/appcelerator/titanium_mobile/blob/b735f547a94b35a1d8ac899580879df4f751c364/iphone/Classes/TiUIWebView.m#L240)
and modify implementation of  internal function [loadLocalURL](https://github.com/appcelerator/titanium_mobile/blob/b735f547a94b35a1d8ac899580879df4f751c364/iphone/Classes/TiUIWebView.m#L668) . If 'readAccessUrl' is available we can use the value, otherwise default value which will be directory of passed 'url'. So developer need not to modify their implementation until they really need e.g. test case given in ticket. In android, we can modify implementation of 'setUrl' to ignore extra parameter, if it required.

jquick-axway:
The default handling of all other `WebViews` is to use the local HTML file's directory as the base directory. There is no equivalent to `allowingReadAccessToURL` on other platforms.

If a developer wants the base directory to be different, then we already have an existing solution. They can call the `setHtml(htmlString, { baseUrl: path })` method, where they can read in the local HTML file as a string and pass it to that method. For that API, setting a base directory makes sense since the `WebView` needs to be told what paths should relative to since it did not load HTML from file itself.

hansemannn:
I'd extend the `url` setter and add an iOS-only property `assetsDirectory` with defaults to the current setting. Internally, you can just set a helper property and re-use it in the `url` setter. You just need to ensure that the `keySequence` lists the `assetsDirectory` before `url`, so the setters are always called after each other. Pretty light change actually

jquick-axway:
Guys. We're waaayyy over thinking this.
Setting the base directory to the local HTML file is enough.

Also, if you look at the ticket's attached sample code (Resources.zip), you'll see that the HTML file is loading the CSS file via a relative path like this "../css/test.css". Meaning that the HTML is assuming that the base directory is the same location as the HTML file. We just need to restore the old behavior. We've never offered the ability to set the base directory to something different than the HTML file's directory before. So, we're losing nothing.

Let's not waste time implementing a new feature that no one is going to use.

vijaysingh-axway:
@jquick-axway As we discussed, when there is any directory structure like ../Cache/html/example.html ,  ../Cache/css/test.css if we use default access directory ../Cache/html/, it will not whitelist ../Cache/css/test.css for webkit. For whitelisting ../Cache/css/test.css we have to whitelist ../Cache/ directory. I agree with Hans's proposal to make iOS specific property.

janvennemann:
think this is a good middle ground between getting parity of Android and iOS and having too much platform specific APIs. With this change loading local resources that are in the same directory as the html works out of the box on both platforms. Referring resources from parent directories like the `../css/test.css` for example will need setting the property on iOS.

@vijaysingh-axway I think we should add a few notes to the docs explaining when the `assetDirectory` is required and what its default value is.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -657,6 +657,15 @@ properties:
     since: "3.0.0"
     default: false
 
+  - name: assetsDirectory
+    summary: Path of file or directory to allow read access.
```
janvennemann:
```suggestion
    summary: Path of file or directory to allow read access by the WebView.
```

apidoc/Titanium/UI/WebView.yml
```diff
@@ -657,6 +657,15 @@ properties:
...
+    description: |
+        If assetsDirectory references a single file, only that file may be loaded. If assetsDirectory references a directory,
```
janvennemann:
```suggestion
        Use this to change the resources the web view has access to when loading the content of a local file. By default
        the web view only has access to files inside the same directory as the loaded file. To reference resources from
        other directories (e.g. a parent directory) change this property accordingly.
        
        If assetsDirectory references a single file, only that file may be loaded. If assetsDirectory references a
        directory, files inside that directory may be loaded.
```

apidoc/Titanium/UI/WebView.yml
```diff
@@ -657,6 +657,15 @@ properties:
     since: "3.0.0"
     default: false
 
+  - name: assetsDirectory
+    summary: Path of file or directory to allow read access.
+    description: |
+        If assetsDirectory references a single file, only that file may be loaded. If assetsDirectory references a directory, 
+        files inside that directory may be loaded. This property should be set with [url](Titanium.UI.WebView.url) property or before it.
```
janvennemann:
```suggestion
        
        This property needs to be set before [url](Titanium.UI.WebView.url) is assigned to a local file.
```

janvennemann (changes requested):

vijaysingh-axway:
> I think this is a good middle ground solution between getting parity of Android and iOS and having too much platform specific APIs. With this change loading local resources that are in the same directory or in a sub directory as the html works out of the box on both platforms. Referring resources from parent directories like the `../css/test.css` for example will need setting the property on iOS.
> 
> @vijaysingh-axway I think we should add a few notes to the docs explaining when the `assetDirectory` is required and what its default value is.

@janvennemann Updated as per your suggestion.

janvennemann (approved):
LGTM!

janvennemann:
@vijaysingh-axway Please add a unit test for the new property, thanks!

jigneshkasundra:
hello,

can you please review and confirm is it related to same ? file stored under applicationDataDirectory not loading into webview

https://jira.appcelerator.org/browse/AC-6319

keerthi1032:
FR passed. WebView works as expected .loading HTML files . Waiting for 8.1.0 GA to merge.

**_Test Environment:_**
Operating System
**Node.js**
  Node.js Version             = 10.16.0
**Titanium CLI**
**Titanium SDK**
  SDK Version                 = local sdk 8.2.0.v20190724052958
Device -iphone x iOS 11
Simulator -iPhone 6s iOS 12.4
Studio =5.1.3.201907112159
Cli -7.1.0.-24

janvennemann:
@vijaysingh-axway can you please add the missing unit tests before this gets merged? Also there seems to be a related issue with `setHtml`, see [AC-6319](https://jira.appcelerator.org/browse/AC-6319).

sgtcoolguy:
@vijaysingh-axway This is repeatedly failing some of the WebView unit tests due to a timeout. I don't think we can merge this until that is resolved.

sgtcoolguy (changes requested):
4 WebView tests are timing out with this PR. They need to be working for this to be ready to merge.

vijaysingh-axway:
Looks like iOS 13 broken WKWebView's API - (nullable WKNavigation *)loadFileURL:(NSURL *)URL allowingReadAccessToURL:(NSURL *)readAccessURL. https://forums.developer.apple.com/thread/120566
In iOS 12 it is working fine. Looking in this.

vijaysingh-axway:
Closing in favor of PR #11431  .

PR #10981, opened 2019-06-19
chore(package): update node-titanium-sdk to version 3.2.0

sgtcoolguy:

**Description:**
Updates node-titanium-sdk to 3.2.0 which includes the fix for TIMOB-27098

This is targeting SDK release 8.1.0 (greenkeeper presumably will generate a PR for master)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,16 +13,16 @@
       }
     },
     "@babel/core": {
-      "version": "7.4.4",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.4.4.tgz",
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
QE - To test:
- create a classic app
- run `ti build -p iOS --source-maps`
- Debug under Safari
- On the left hand side, you should see an entry for `app.js` (with a child `app.js` under it)
- one is the original source file's contents, the other is the "generated" contents in-app loaded by the JS engine).

Before the fix, using source maps resulted in app.js and other app JS file not to get displayed.

ssekhri:
FR Passed. The app.js (main and child) are both shown during debug using safari web inspector and chrome dev-tools. Also the breakpoints are hit as expected.

hansemannn:
Has this landed on master already? Keen to test it out!

sgtcoolguy:
@hansemannn It was merged to 8_1_X so that's likely already available via ti sdk. I just pushed the same fix to master now, so hopefully it'll be available within half an hour (unless the build fails for some reason, the office network has been flaky today)

PR #10984, opened 2019-06-19
Update Global.yml

jawa9000:
Fix for https://jira.appcelerator.org/browse/TIDOC-3203


jawa9000 (commented):
Tested this changed locally and all was good.

PR #10985, opened 2019-06-19
docs: updates per TIDOC-3203

jawa9000:
https://jira.appcelerator.org/browse/TIDOC-3203


ewanharris:
Looking at the [TDoc spec](https://wiki.appcelerator.org/display/guides2/TDoc+Specification#TDocSpecification-MarkdownLinksinFree-FormTextValues) an angle bracket is correct to use when we don't want any custom text for the link. Seeing as that document refers to it as a "Markdown standard" maybe this is something that's changed when we switched the markdown parser?

jawa9000:
@ewanharris, I agree that that is the standard but when the angle brackets appear inside a deprecation property, the angle brackets stay (rendered on the page). Hans complained about that visual artifact. I only changed the links that appears within deprecation notes.

PR #10986, opened 2019-06-20
feat(android): Allow touch feedback for image and gradient backgrounds

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26315

**Description:**
Touch feedback property is now available for views with an image or a gradient backgrounds.

No unit tests since this is a visual improvement.

Test case:
```js
var win = Ti.UI.createWindow({ layout: 'vertical'}),
	button = Ti.UI.createButton({ top: 10,
				backgroundImage: '<your-image-here>',
				touchFeedback: true,
				touchFeedbackColor: 'blue' }),
	button2 = Ti.UI.createButton({ top: 10,
				backgroundColor: 'yellow',
				touchFeedback: true,
				touchFeedbackColor: 'green'}),
	button3 = Ti.UI.createButton({ top: 10,
				backgroundGradient: { type: 'linear',
				startPoint: { x: '0%', y: '50%' },
				endPoint: { x: '100%', y: '50%' },
				colors: [ { color: 'red', offset: 0.0},
					{ color: 'blue', offset: 0.25 },
					{ color: 'red', offset: 1.0 } ],
				},
				touchFeedback: true,
				touchFeedbackColor: 'blue'}),
	button4 = Ti.UI.createButton({ top: 10,
				title: 'Change colors'});

win.add([button, button2, button3, button4]);

button4.addEventListener('click', function () {
	button.touchFeedbackColor = 'yellow';
	button2.touchFeedbackColor = 'red';
	button3.touchFeedbackColor = 'green';
});
win.open();
```


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1018,25 +1015,26 @@ public void processProperties(KrollDict d)
...
+				background.setBackgroundColor(bgColor);
+			} else {
+				nativeView.setBackgroundColor(bgColor);
 			}
 		}
 
+		if (canApplyTouchFeedback(d)) {
+			applyTouchFeedback(bgColor, background,
+							   d.containsKey(TiC.PROPERTY_TOUCH_FEEDBACK_COLOR)
+								   ? TiConvert.toColor(d, TiC.PROPERTY_TOUCH_FEEDBACK_COLOR)
```
sgtcoolguy: I understand that this is way outside the scope of this PR, but this screams to me that we should have a better API to handling cases like this. Specially why not just have `TiConvert.toColor()` handle the case where ether key doesn't exist and have it return a default value (in this case `null`)?

Presumably it's because it returns an `int`? I think it may useful to have a `TiColor` class that can encapsulate the concept of colors and incorporate the `TiColorHelper` logic and support outputting any representations (like an int) that we'd need. In fact, Android itself appears to already have the same concept in `android.graphics.Color`. Bonus is that you can then return `null` in a case like this.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1195,7 +1193,7 @@ protected boolean canApplyTouchFeedback(@NonNull KrollDict props)
 	 * @param backgroundColor The background color of the view.
 	 * @param rippleColor The ripple color.
 	 */
-	private void applyTouchFeedback(@NonNull Integer backgroundColor, @Nullable Integer rippleColor)
+	private void applyTouchFeedback(Integer backgroundColor, Drawable backgroundDrawable, @Nullable Integer rippleColor)
```
sgtcoolguy: If we're using `@Nullable` to mark rippleColor, we should use the annotations consistently as we go through our code to help us out. So, presumably all three arguments here are `@Nullable` now (since I can see null cheeks below)?

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -910,6 +896,17 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
 			if (!nativeViewNull) {
 				nativeView.postInvalidate();
 			}
+		} else if (key.equals(TiC.PROPERTY_TOUCH_FEEDBACK_COLOR) || key.equals(TiC.PROPERTY_TOUCH_FEEDBACK)) {
+			KrollDict d = proxy.getProperties();
```
sgtcoolguy: before the logic called:
```java
 // Update first before querying.
proxy.setProperty(key, newValue);
```
before calling `proxy.getProperties();` Is it still necessary?

ypbnv: From what I see the properties dictionary of the proxy is updated here:
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java#L1125
This happens before calling the `propertyChanged`, so I think this line would not be necessary in any scenario. I see the line is from quite some time, so it may be a leftover from something not in use anymore?

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
ypbnv: Yes, all three parameters can be null. I will add the annotations.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
ypbnv: We have a few more places where we fallback to a default color, so it would definitely be an improvement to deal with that in one single method. I can try to do it in this PR, but I think a new one dedicated to the refactoring would be better.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -831,19 +831,6 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
...
-					if (!nativeViewNull) {
-						if (canApplyTouchFeedback(d)) {
-							applyTouchFeedback(bgColor, d.containsKey(TiC.PROPERTY_TOUCH_FEEDBACK_COLOR)
-															? TiConvert.toColor(d, TiC.PROPERTY_TOUCH_FEEDBACK_COLOR)
-															: null);
-						} else {
-							nativeView.setBackgroundColor(bgColor);
-						}
-					}
-				}
```
jquick-axway, line 846: I think removing this block of code will cause a regression where you can no longer change a view's `backgroundColor` dynamically anymore because `setBackgroundColor()` won't get called.

jquick-axway (changes requested):

jquick-axway:
Closing in favor of PR #11795 .

PR #10992, opened 2019-06-24
fix(ios): restore correct Ti.Platform.id value

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27175

**Optional Description:**
The UUID returned by `Ti.Platform.id` changes when migrating to SDK 8.0.0. This fixes the issues and will return the expected identifier again.

### DIFF
--- a/iphone/Classes/TiUtils+Addons.h
+++ b/iphone/Classes/TiUtils+Addons.h
@@ -1,0 +1,25 @@
+/**
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2019-present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#import <TitaniumKit/TiUtils.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface TiUtils (Addons)
+
+/**
+ Returns a unique identifier for this app.
+ 
+ This will change upon a fresh install.
+ 
+ @return UUID for this app.
+ */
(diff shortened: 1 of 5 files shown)


vijaysingh-axway:
@janvennemann  Can you explain reason of the issue? Is string 'appcelerator' is replaced with app name in SDK < 8.0.0?

janvennemann:
@vijaysingh-axway We will scrub our iOS source files to replace strings like `Titanium` or `appcelerator` in our build [here](https://github.com/appcelerator/titanium_mobile/blob/e677dbe127c86248f5d37a028d6a85a3c80a549d/iphone/cli/commands/_build.js#L4448). To prevent possible app store rejections i guess. In this case the key `com.appcelerator.uuid` was replaced to be `com.<app-name>.uuid`.

Now, since 8.0.0 the `TiUtils.m` is part of TitaniumKit, so it gets pre compiled and the build won't replace the key that is used to store the unique app id. And since the key changes when migrating from 7.x to 8.x, the id changes. Moving the relevant code back to a category outside of TitaniumKit solves this issue as the build can now properly replace the key again.

vijaysingh-axway (approved):
CR passed.

ssjsamir (approved):
FR Passed: Correct Ti.Platform.id value is now restored to meet 7.5.2.GA, tested with the steps mentioned in: https://jira.appcelerator.org/browse/TIMOB-27175. 

**Test Environment**
```
iPhone 7 (12.2 Sim)
APPC CLI: 7.0.12-70X.4
Node.js Version: 8.11.3
Xcode 8.9.1

PR #10998, opened 2019-06-26
feat(api): add Node-compatible fs module

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-18583

**Description:**
This adds a node-compatible `fs` module in our common JS code shipped with the SDK.

The goal is to have a compatible shim that allows use of more npm packages out there since this is a pretty heavily used core module of Node. When possible I attempted to make the async variants more than simple wrappers that scheduled the sync version in a `setTimeout()` call (though I used that pattern for a number of the simpler ones).

For example the basic read/write/copy operations I attempt to make use of our async stream APIs.

Note that the async variants are overall **slower** than the sync variants, but give the main thread/app time to "breathe" - which should make the app more usable from an end user perspective. I used `setTimeout` with a minimum wait threshold, so async variants will have an initial delay from the scheduling and enforcement of a minimum wait time - and that may be compounded if the implementation uses other async methods that use setTimeout to schedule.

CAN'T/WON'T DO:
- chown/chmod
- watch/unwatch

FOR LATER:
- New promises API/FileHandle

REQUIRES SDK CHANGES TO DO:
- `utimes`
- Any link related methods (link, lstat, symlink, readLink) - though, how relevant are these in actual apps?

REQUIRES OTHER SHIMS TO DO:
- `createReadStream`
- `createWriteStream`

Both of these require creation of a `stream` shim first.

### DIFF
--- a/common/Resources/ti.internal/extensions/node/fs.js
+++ b/common/Resources/ti.internal/extensions/node/fs.js
@@ -1,0 +1,1704 @@
+import assertArgumentType from './_errors';
+import path from './path';
+
+const isWindows = Ti.Platform.name === 'windows';
+const isAndroid = Ti.Platform.name === 'android';
+
+// Keep track of printing out one-time warning messages for unsupported operations/options/arguments
+const printedWarnings = {};
+function oneTimeWarning(key, msg) {
+	if (!printedWarnings[key]) {
+		console.warn(msg);
+		printedWarnings[key] = true;
+	}
+}
+/**
+ * Prints a one-time warning message that we do not support the given API and performs an effective no-op
+ * @param {string} moduleName name of the module/object
+ * @param {string} name name of the function.property we don't support
+ * @returns {Function} no-op function
+ */
+function unsupportedNoop(moduleName, name) {
+	return () => {
+		const fqn = `${moduleName}.${name}`;
+		oneTimeWarning(fqn, `"${fqn}" is not supported yet on Titanium and uses a no-op fallback.`);
+		return undefined;
+	};
+}
+
+/**
+ * @param {string} moduleName name of the module/object
+ * @param {string} name name of the function.property we don't support
+ * @param {Function} callback async callback we call in a quick setTimeout
+ */
+function asyncUnsupportedNoop(moduleName, name, callback) {
+	callback = maybeCallback(callback); // enforce we have a valid callback
+	unsupportedNoop(moduleName, name)();
+	setTimeout(callback, 1);
+}
+
+// Used to choose the buffer/chunk size when pumping bytes during copies
+const COPY_FILE_CHUNK_SIZE = 8092; // what should we use here?
+
+// Keep track of integer -> FileStream mappings
+const fileDescriptors = new Map();
+let fileDescriptorCount = 4; // global counter used to report file descriptor integers
+
+// Map file system access flags to Ti.Filesystem.MODE_* constants
+const FLAGS_TO_TI_MODE = new Map();
+FLAGS_TO_TI_MODE.set('a', Ti.Filesystem.MODE_APPEND);
+FLAGS_TO_TI_MODE.set('ax', Ti.Filesystem.MODE_APPEND);
+FLAGS_TO_TI_MODE.set('a+', Ti.Filesystem.MODE_APPEND);
+FLAGS_TO_TI_MODE.set('ax+', Ti.Filesystem.MODE_APPEND);
+FLAGS_TO_TI_MODE.set('as+', Ti.Filesystem.MODE_APPEND);
+FLAGS_TO_TI_MODE.set('r', Ti.Filesystem.MODE_READ);
+FLAGS_TO_TI_MODE.set('r+', Ti.Filesystem.MODE_READ);
+FLAGS_TO_TI_MODE.set('rs+', Ti.Filesystem.MODE_READ);
+FLAGS_TO_TI_MODE.set('w', Ti.Filesystem.MODE_WRITE);
+FLAGS_TO_TI_MODE.set('wx', Ti.Filesystem.MODE_WRITE);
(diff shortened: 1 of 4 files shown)


common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+/**
+ * @param {string|Buffer|URL} path path to file
+ * @param {string} [flags='r'] file system access flags
+ * @param {integer} [_mode=0o666] file mode to use when creating file
+ * @returns {integer}
+ */
+fs.openSync = (path, flags = 'r', _mode = 0o666) => {
+	const tiFile = getTiFileFromPathLikeValue(path);
+	if (!tiFile.exists()) {
+		// TODO: Support creating file with specific mode
```
janvennemann: Print warning that mode is not supported yet.

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+	const file = getTiFileFromPathLikeValue(path);
+	if (!file.exists()) {
+		throw noSuchFile('scandir', path);
+	}
+	if (!file.isDirectory()) {
+		throw notADirectory('scandir', path);
+	}
+	// options = mergeDefaultOptions(options, { encoding: 'utf-8', withFileTypes: false });
+	// TODO: if options.withFileTypes === true, return fs.Dirent objects
+	// TODO: if options.encoding === 'buffer', loop through listing and return Ti.Buffer objects for each
```
janvennemann: Print warning that `options` are not supported yet.

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+ */
+fs.statSync = (path, options) => new fs.Stats(path);
+
+fs.symlink = (target, path, type, callback) => {
+	if (typeof type === 'function') {
+		callback = type;
+	}
+	setTimeout(() => callback(new Error('fs.symlink is not supported')), 1);
+};
+fs.symlinkSync = NO_OP; // TODO: Throw error about not being supported?
```
janvennemann: We should have a unified behavior for not implemented methods. For example, `fs.symlink` throws an error, the sync version of it is a noop and other methods are simply not implemented at all.

Not quite sure what would be the best behavior in terms of compatibility with third-party modules, though. Just print a one-time warning and run a no-op (see my comment at the top of the file)? Or completely error out by throwing?

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+			return buffer.toString(encoding);
+	}
+}
+
+/**
+ * @param {string|Buffer|URL} path file path
+ * @return {Ti.Filesystem.File}
+ */
+function getTiFileFromPathLikeValue(path) {
+	// Android can't properly handle file: URIs if they don't have trailing '//'
```
janvennemann: It looks like you have actually fixed that in the Android implementation, right?

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+	// if (path.startsWith('file:') && !path.startsWith('file://')) {
+	// 	console.log(`Fixing bad file: URI from ${path} -> file://${path.slice(5)}`);
+	// 	path = `file://${path.slice(5)}`;
+	// }
+	// FIXME: Handle Buffers or URLs!
```
janvennemann: Error out here for unsupported types.

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1622 @@
...
+ * @param {string|Buffer} data data to append to file
+ * @param {object|string} [options] options
+ * @param {string} [options.encoding='utf8'] encoding to use
+ * @param {integer} [options.mode=0o666] mode to create file, if not created
+ * @param {string} [options.flag='a'] file system flag
+ * @param {Function} callback function to call back with error if failed
+ */
+fs.appendFile = (file, data, options, callback) => {
+	callback = callback || options;
+	// TODO: Verify the callback is now a function!
```
janvennemann: This needs to be done in most async function throughout the file. See node's [maybeCallback](https://github.com/nodejs/node/blob/53816cce699d02fb28a49b258e1fbc474568bbf8/lib/fs.js#L142) which does exactly this.

janvennemann (changes requested):

sgtcoolguy:
I pushed up come commits to address the first round of feedback...

common/Resources/ti.internal/extensions/node/fs.js
```diff
@@ -0,0 +1,1700 @@
...
+		throw noSuchFile('scandir', filepath);
+	}
+	if (!file.isDirectory()) {
+		throw notADirectory('scandir', filepath);
+	}
+	options = mergeDefaultOptions(options, { encoding: 'utf-8', withFileTypes: false });
+	const listing = file.getDirectoryListing();
+	if (options.withFileTypes === true) {
+		// TODO: if options.withFileTypes === true, return fs.Dirent objects
+		oneTimeWarning('fs.readdir\'s optins.withFileTypes is unsupported by Titanium and strings will be returned');
```
janvennemann:
```suggestion
		oneTimeWarning('fs.readdir\'s options.withFileTypes is unsupported by Titanium and strings will be returned');
```

janvennemann (changes requested):

janvennemann (approved):
LGTM!

sgtcoolguy:
All right, need to land one of the bugfixes in this PR (as #11098) and rebase all these commits down, then I'll merge it in...

PR #11002, opened 2019-06-27
test: remove .only from database tests

ewanharris:


jquick-axway:
I don't know what `describe.only()` does. @sgtcoolguy wrote it. I merely back-ported it.
Was it causing problems?

ewanharris:
@jquick-axway, it's a [feature of mocha](https://mochajs.org/#exclusive-tests) to run an exclusive set of tests and/or suites. So without this change we're only running those 10 or so tests.

PR #11003, opened 2019-06-27
fix(android): Fix Up button flickering when clicked in NavigationWindow

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27190

**Description:**
The Up button disappears before the Window is actually closed when this is done in a
NavigationWindow's Windows stack.
This was happening due the displayOptions flags being set after the ActionBar is drawn and immediately before closing the Window. In this PR setting the child Window's `showHomeAsUp` is done through the `ActionBarProxy` class instead of directly the `SupportActionBar` instance.

Note: This is a visual fix so there is no unit test for it.

**Test case:**
```js
const win1 = Ti.UI.createWindow(),
	win2 = Ti.UI.createWindow(),
	navWin = Ti.UI.createNavigationWindow({ window: win1}),
	but = Ti.UI.createButton({ title: 'Nav'});
navWin.open();
win1.add(but);
butt.addEventListener('click', () => {
	navWin.openWindow(win2);
});
```
There is a short clip in the JIRA ticket that demonstrates what is the mentioned flicker.


jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
No flicker when closing the Nav window using the Nav window back button or android's back button.
Verified using:
Mac OS 10.14.3
Ti SDK: 8.2.0.v20190627060333
Appc CLI: 7.1.0-24
Node: 8.16.0
JDK: 10.0.2

PR #11005, opened 2019-06-28
Updated Menu.yml per TIDOC-2376

jawa9000:
https://jira.appcelerator.org/browse/TIDOC-2376


apidoc/Titanium/Android/Menu.yml
```diff
@@ -304,5 +304,5 @@ examples:
 
             // Ensure menu is displayed
             function doOpen(e) {
-                $.win.invalidateOptionsMenu();
+                $.win.invalidateOptionsMenu(); // should be --> $.win.activity.invalidateOptionsMenu();
```
sgtcoolguy: This doesn't look right

sgtcoolguy (changes requested):
Looks like a bad insert here

sgtcoolguy:
manually fixed and merged

apidoc/Titanium/Android/Menu.yml
jawa9000: Thanks for the assist, Chris.

PR #11008, opened 2019-06-28
fix(ios): fallback to non-widget require in catch

ewanharris:
This ensures that when the alloy require fails we try the normal require

Fixes TIMOB-27195

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27195


vijaysingh-axway (approved):
CR passed.

ssjsamir (approved):
FR Passed: Able to now use a commonjs module in an itemtemplate without the build throwing an error. Tested with the following application: https://github.com/ewanharris/ti.imageview-listview-example

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 11 beta 5
Node.js ^8.11.1
iPhone 6s (12.2)
```

PR #11010, opened 2019-07-01
build: use faster method for copying production npm packages into SDK

sgtcoolguy:
**Description:**
When we package up the SDK in our build process I used to handle the `node_modules` by doing:
- straight copy of the folder to the built SDK
- running `npm prune --production` to let npm remove all dev depndencies
- doing extra hacks we need around node-ios-device prebuilt binaries and node-titanium-sdk

This PR changes the first two steps to instead traverse the packages and copy over only the production dependencies. This came out of a PR I submitted to Brenton's package: brentonhouse/titanium-module-copier#2

He's using that package to run in some custom alloy build hook or something os needs it to run sync and do some other things that I don't think align with what I wanted to do here.

The goal here is to eventually also re-use this code in our CLI build process to allow copying production `node_modules` into the user's app so that they can have a top-level `package.json` and not have to stick it in `Resources/` or `app/lib` or something. I assume at that point we'll have to add extra knowledge about our re-packaged native modules and run things through the jsanalyze code.

### DIFF
--- a/build/lib/module-copier.js
+++ b/build/lib/module-copier.js
@@ -1,0 +1,114 @@
+'use strict';
+
+const copier = {};
+module.exports = copier;
+
+const fs = require('fs-extra');
+const path = require('path');
+
+const NODE_MODULES = 'node_modules';
+
+/**
+ * @param {string} projectPath absolute filepath for project root directory
+ * @param {string} targetPath absolute filepath for target directory to copy node_modules into
+ * @param {object} [options] options object
+ * @param {boolean} [options.includeOptional=true] whether to include optional dependencies when gathering
+ * @returns {Promise<void>} A Promise that resolves on completion
+ */
+copier.execute = async (projectPath, targetPath, options = { includeOptional: true }) => {
+	if (projectPath === null || projectPath === undefined) {
+		throw new Error('projectPath must be defined.');
+	}
+	if (targetPath === null || targetPath === undefined) {
+		throw new Error('targetPath must be defined.');
+	}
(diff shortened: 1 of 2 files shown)


cb1kenobi:
FWIW, `node-ios-device@2.0.0` will no longer require downloading from S3 or compiling thanks to bundled prebuilds.

build/lib/module-copier.js
```diff
@@ -0,0 +1,114 @@
...
+
+	const dirSet = new Set(directoriesToBeCopied); // de-duplicate
+	// back to Array so we can #map()
+	const deDuplicated = Array.from(dirSet);
+
+	// Then copy them over
+	return Promise.all(deDuplicated.map(async directory => {
+		const relativePath = directory.substring(projectPath.length);
+		const destPath = path.join(targetPath, relativePath);
+		return fs.copy(directory, destPath, { overwrite: true }); // TODO: Allow incremental copying! Maybe use gulp/vinyl?
```
ewanharris: It seems setting `dereference: true` on the opts here allows us to copy over a module that has been linked via `npm link`. That's the only difference I can see between the old method and this method (the old method would just completely ignore the link and reinstall everything).

sgtcoolguy: Before I think usage of `npm link` with our build process actually broke the build.

With this, the logic *looks* like it should copy the symlink. Setting to `dereference: true` would copy the dereferenced symlink. So I think we'd prefer to retain the symlink if we can for local development.

PR #11019, opened 2019-07-02
 chore(package): update ioslib to version 1.7.9 (8_0_X)

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27124

**Optional Description:**
This backports ioslib 1.7.9 to SDK 8.0.3

While 8.2.0 is our Xcode 11/ iOS 13 target release, installing Xcode 11 betas can break SDK usage even if the user retains Xcode 10 and keeps it as the selected one. Porting this back helps fix that issue for users.

Note that I also snuck in a run of `npm audit fix` as it was reporting a fairly large number of vulnerabilities before that. Most were fixed by updating `devDependencies`, but this also updates `liveview` to 1.5.0


package.json
```diff
@@ -72,8 +72,8 @@
     "fs-extra": "^7.0.1",
     "glob": "^7.1.2",
     "humanize": "0.0.9",
-    "ioslib": "^1.7.7",
-    "liveview": "^1.4.5",
+    "ioslib": "^1.7.9",
+    "liveview": "^1.5.0",
```
ewanharris: Could you back the liveview change out (and maybe change to `~1.4.5`), it intended to only land on the 8_1_X branch to align with the node-titanium-sdk changes that landed there

ewanharris (changes requested):
The liveview bump should be backed out, the PR to release it (appcelerator/liveview/pull/128) mentioned to not land on 8_0_X, but I've no idea why now..

PR #11022, opened 2019-07-02
TIMOB-27164 iOS: Implement async Ti.Database.DB methods

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27164

**Description:**
First, this modifies the test suite to run on iOS and Android. It also fixes expectations so that callbacks expect to receive a first argument of a possible Error object, second argument of results.

Second, this implements the new APIs on iOS. I'm not real sure of the best mechanism for lock/queueing the operations on iOS so that's probably the most important thing for @janvennemann or @vijaysingh-axway to look at - basically now that we have async APIs we need to synchronize access to the database so that only one base DB operation can run simultaneously under the hood. For that I used `@synchronize(self) {}` blocks to lock the `database.prepareStatement`, `statement.executeQuery` and the internal lazy add/removal of statements to an array used to retain/clean them up.

### DIFF
--- a/android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
+++ b/android/modules/database/src/java/ti/modules/titanium/database/TiDatabaseProxy.java
@@ -6,6 +6,7 @@
  */
 package ti.modules.titanium.database;
 
+import org.appcelerator.kroll.JSError;
 import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollFunction;
 import org.appcelerator.kroll.KrollProxy;
@@ -23,8 +24,11 @@ import android.database.DatabaseUtils;
 import android.database.sqlite.SQLiteDatabase;
 import android.os.Looper;
 
+import java.lang.Exception;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -37,6 +41,7 @@ public class TiDatabaseProxy extends KrollProxy
 	private Thread thread;
 	private BlockingQueue<Runnable> queue = new LinkedBlockingQueue<>();
 	private AtomicBoolean executingQueue = new AtomicBoolean(false);
+	private boolean isClosed = false;
 
 	protected SQLiteDatabase db;
 	protected String name;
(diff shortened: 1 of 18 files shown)


garymathews (approved):
Android changes LGTM

vijaysingh-axway (commented):
@sgtcoolguy  Changes looks good. 
New APIs are not documented.  Please add doc for same and backport it. Thanks!

jquick-axway:
Sorry for my late feedback.
There are 2 edge cases we need to consider.

1st, what should happen if the database's `close()` method is called while it's in the middle of executing async SQL queries? On Android, we currently abort the async SQL queries. (We may need to give up the file lock in case the developer wants to copy the database file.)

2nd, what should happen if the database object is garbage collected while it's in the middle of executing async SQL queries? Currently, both Android and iOS internally close the database. For Android, this means it'll abort the async SQL queries. But for the garbage collected case, ideally we shouldn't close the database until the async SQL queries are done... because I can see a developer potentially doing something like the below without ever closing the database explicitly. I've seen developer code doing the same with our currently blocking execute methods as well.
```javascript
function updateDatabase(sqlStatements) {
	var dbConnection = Ti.Database.open("myDatabase.sqlite");
	dbConnection.executeAllAsync(sqlStatements, (err, results) => {
		if (!err) {
			updateUI();
		}
	});
}
```

sgtcoolguy:
So right now I don't believe there's any lock around `#close()` for iOS, so my guess is that we'll crash if they close and then there's still queries around that will get run (because the database field gets set to nil).

I could use the same lock on the proxy instance around close so that it'll block until any query running finishes, then cleans up, and add code that checks the database field isn't nil in our internal query method that will throw an Error (or fire an error to the callback). That's probably the best scenario - close when we can, and if there's any in-progress queries not done throw an Error to the app to handle.

For the second issue, the DB object should *never* get GC'd until any async queries are done. We have some complicated C++/Java interior code to try and track references so a JS object won't get GC'd while it's Java object lives on, and vice versa. Same for iOS, which I believe JavaScriptCore framework takes care of. But it'd be good to try and write some test that verifies this.

sgtcoolguy:
@jquick-axway ok, I pushed another commit with a couple additional tests and fixed them for iOS. Basically the first test just queues up thousands of inserts and then sets a timer to close the database (which should happen in the middle of all the queries). I had to tweak iOS to lock internally on close, and to check when executing a query if the database pointer is nil, if so bubble up an error.

The second test tries to recreate the scenario you mention, but I'm not sure how to better test avoiding a GC.

Both tests passed locally on iOS Sim and Android emulator. But I think Android is actually just blocking on close until all the queries finish, which differs from iOS - there we'll close before starting the next query in an array, then try the next element, see the pointer is nil and throw an Error.

jquick-axway:
Thanks. Happy to help.

@garymathews, it sounds like we need parity between Android/iOS' database `close()` behavior. Chris says the `close()` method blocks on Android until async execution is done. It should probably abort. I say this because back navigating out of the app will terminate the JS runtime, which in turn will automatically close the database connection when garbage collected. We don't want that to block the main UI thread in case there is an insane amount if async SQL queries going on, because it would block relaunch of the app.

@sgtcoolguy, iOS has an undocumented `Ti.App.garbageCollect()` method. Perhaps we could add the same on Android and Windows for testing purposes?

garymathews:
@sgtcoolguy @jquick-axway Made changes to `close()` for parity with iOS; tests are passing now.

vijaysingh-axway (approved):

sgtcoolguy:
@vijaysingh-axway @garymathews @jquick-axway ok, so I force-pushed some changes, and I had to overwrite Gary's changes so that we could more easily provide the details on the exact index that failed and provide partial results. I still need to update the docs here.

But basically if `#executeAll()` fails in the middle for any reason, we throw a JS Error with the properties `index` and `results`, which hold the index we failed on; and the partial results we had collected. This way a dev can catch the error and have more detail on exactly what failed and still operate on partial results.

If `#executeAllAsync()` fails we call the callback with an `Error` argument having an `index` property pointing at the bad query index, and the`results` callback argument holds the partial results.

The interesting part here is that I had to make a more generic mechanism for passing along Java `Throwable`s from Java to JS and converting them to JS `Error`s, with a new interface to provide additional key/value pairs that get hung as properties on the JS `Error`. We already handled if Java threw an Exception, converting that, but I had to decouple the handling/re-throwing from the conversion.

On iOS, the code to do that was quite a bit easier, so it was really about extracting the code that created an Error from the code that "threw" the error so we could create an error to pass to the callback (again unbundle creation from throwing).

ssjsamir (approved):
**FR Passed**, able to use the new DB implemented methods without any issues on iOS:
```
Ti.Database.DB.executeAsync(callback, query, parameters...);
Ti.Database.DB.executeAll(queries);
Ti.Database.DB.executeAllAsync(callback, queries);
```

**Test Environment**
MacOS Mojave version 10.14.4
Xcode 10.2.1
Node.js ^8.12.0
iPhone 6s Sim (12.1).
"NPM":"4.2.13","CLI":"7.1.0-master.24"

PR #11023, opened 2019-07-03
fix(ios): regression in incremental device builds

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27202

Fixes a regression in incremental device builds that caused missing entries of encrypted JS files in the `index.json` used in our require.


sgtcoolguy (approved):
LGTM, did not do FR

ssekhri:
FR Passed. Incremental builds working fine for iOS.

PR #11024, opened 2019-07-03
fix(ios): regression in incremental device builds (8_1_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27202

8_1_X backport of #11023

### DIFF
--- a/cli/lib/tasks/process-js-task.js
+++ b/cli/lib/tasks/process-js-task.js
@@ -116,7 +116,7 @@ class ProcessJsTask extends IncrementalFileTask {
 		this.jsFiles = this.data.jsFiles;
 		this.jsBootstrapFiles.splice(0, 0, ...this.data.jsBootstrapFiles);
 		return Promise.all(Object.keys(this.jsFiles).map(relPath => {
-			return limit(() => this.processJsFile(this.jsFiles[relPath].src))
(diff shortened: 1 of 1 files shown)


cli/lib/tasks/process-js-task.js
```diff
@@ -110,7 +113,11 @@ class ProcessJsTask extends IncrementalFileTask {
 			return this.doFullTaskRun();
 		}
 
-		Object.keys(this.data.jsFiles).forEach(relPath => this.builder.unmarkBuildDirFile(this.data.jsFiles[relPath].dest));
+		this.jsFiles = this.data.jsFiles;
```
sgtcoolguy: OK, so looking at the code, it appears we already compare hashes inside the hooks that processJsFile does, so we effectively skip the work on them anyways?

sgtcoolguy (approved):
LGTM, though did not do FR.

cli/lib/tasks/process-js-task.js
janvennemann: Yes, it will still skip unchanged files. It just goes through the processing flow to make sure the hooks are called and that the builder gets populated with required info about encrypted files.

ssekhri:
FR Passed. Incremental builds working fine for iOS.

PR #11027, opened 2019-07-03
fix(ios): verify module class type

garymathews:
- Prevent interpreting CommonJS module as native module

##### TEST CASE
- Create basic `utils/test.js` module
```JS
const test = require(‘utils/test.js’);
```
- Should not see crash

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27174)


janvennemann (approved):

sgtcoolguy:
Why no test here? This seems like something we should be able to test...

PR #11036, opened 2019-07-09
fix(ios): video player natural size property and event

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27234

Since 7.0.0 VideoPlayers property `naturalSize` and event `naturalSizeAvailable` are broken on iOS. They return video layer size instead of "movie" size.

P.S. I'm not obj-c developer, so please check carefully (it's working for my case).


iphone/Classes/TiMediaVideoPlayerProxy.m
```diff
@@ -702,9 +702,13 @@ - (NSNumber *)repeatMode
 
 - (id)naturalSize
 {
+  CGSize size = CGSizeZero;
+  if (movie && [movie player] && [[movie player] currentItem]) {
```
hansemannn: ObjC supports optional chaining, so `if (movie.player.currentItem)` should work fine, even if `movie.player` is `nil.`

drauggres: I was looking for Axway coding style conventions about "dots vs brackets" (didn't find any), but instead I found [this](https://docs.axway.com/bundle/Titanium_SDK_allOS_en/page/objective-c_and_objective-c___coding_standards.html):
>
> It is illegal to use a statement such as {{ if (objcObject) { ... } }}. Instead directly compare to nil, only where required.

So, is `if (movie.player.currentItem)` OK?

hansemannn: Well, it's really up to the developer. Nowadays, the dot-notation is preferred (introduced in ObjC 2.0+), but it's not part of the linting / clang-format. In regards to the nil check, the `movie.player.currentItem` is more strict, like `===` on JS.

drauggres: OK. Thank you, `movie.player.currentItem` it is.

vijaysingh-axway (approved):
CR passed.

vijaysingh-axway:
Thanks @drauggres for PR and @hansemannn  for review.

keerthi1032:
@vijaysingh-axway with the fix  naturalSize & naturalSizeAvailable  does not return values .shows blank .added screenshot for reference. can you please look at it 
Screen Shot 2019-08-05 at 2 40 42 PM

drauggres:
@keerthi1032 Actually it is a problem with `console.log`.
Please try with
```javascript
    console.log(e.type + ' e.naturalSize ' + JSON.stringify(e.naturalSize));
    console.log('player.naturalSize ' + JSON.stringify(e.source.naturalSize));
```

drauggres:
Also please consider opening a new ticket for `console.log` bug.
>`console.log( message )`
>Log a message at the `info` level.
>The message to log can either be a single argument, or **any number of arguments**, which will be converted to strings and then concatenated together with a space character.

keerthi1032:
Thanks .it is working now.

PR #11044, opened 2019-07-12
fix(build): remove unresolved dependencies

garymathews:
- Do not attempt to resolve and copy an unresolved dependency

### DIFF
--- a/build/lib/module-copier.js
+++ b/build/lib/module-copier.js
@@ -60,7 +60,7 @@ class Dependency {
 			return [ this.directory ]; // just need our own directory!
 		}
 
-		const children = await Promise.all(childrenNames.map(name => this.resolve(name)));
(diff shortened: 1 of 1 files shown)


sgtcoolguy (approved):
LGTM

build/lib/module-copier.js
```diff
@@ -60,7 +60,7 @@ class Dependency {
 			return [ this.directory ]; // just need our own directory!
 		}
 
-		const children = await Promise.all(childrenNames.map(name => this.resolve(name)));
+		const children = (await Promise.all(childrenNames.map(name => this.resolve(name)))).filter(child => child !== null);
```
cb1kenobi: On a side note, you can just do this in the future:

```js
const children = (await Promise.all(childrenNames.map(name => this.resolve(name)))).filter(Boolean);
```

PR #11045, opened 2019-07-12
refactor(android): asset handling

garymathews:
- Refactor Android asset encryption to utilize new [ti.cloak](https://github.com/appcelerator/ti.cloak) module

##### TEST CASE
- Build application with encrypted assets
  - `appc run -p android --deploy-type test`
- Application should run without any issues

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26043)


hansemannn:
Can we test this already? Which characteristics of our app should we check with this included?

jquick-axway:
@hansemannn, I'm pretty sure this PR won't work with hybrid modules such as "ti.playservices". Not yet anyways. But we'll get there.

hansemannn:
You mean it cannot be part of a project with other modules so far? Do they need to be recompiled? We're happy to extensively test this

garymathews:
@hansemannn It should work fine, just might have issues with `ti.playservices` as I may need to refactor how hybrid-modules load resources.

android/cli/commands/_build.js
```diff
@@ -2817,91 +2819,51 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
-		// figure out which titanium prep to run
-		let titaniumPrep = 'titanium_prep';
-		if (process.platform === 'darwin') {
-			titaniumPrep += '.macos';
-		} else if (process.platform === 'win32') {
-			titaniumPrep += '.win32.exe';
-		} else if (process.platform === 'linux') {
-			titaniumPrep += '.linux' + (process.arch === 'x64' ? '64' : '32');
-		}
+		} else if (cloak) {
```
sgtcoolguy: Minor style nitpick I prefer:
The else is not necessary because the block above returns. Then you could invert the if check to handle the error case and return/exit early, i.e.
```javascript
if (!cloak) {
  this.logger.error('Could not encrypt assets!');
  process.exit(1);
}

// the main logic for encrypting goes here now...
```

Personally I prefer to handle early exists/pre-condition checks at top and avoid nesting the main codepath/logic as a result.

android/cli/commands/_build.js
```diff
@@ -2817,91 +2819,51 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+			this.logger.info('Encrypting javascript assets...');
+			appc.async.parallel(this,
```
sgtcoolguy: I know we haven't started to use async/await yet in our CLI code but there's no reason *not* to at this point. We're Node 8 minimum and about to move to 10 as min in SDK 9.

I think this code would be much more readable if we moved to calling an async function rather than nesting these callbacks.

I know our ESLint rules don't like mixing callbacks and promises, but we kind of need to to start transitioning over.

android/cli/commands/_build.js
```diff
@@ -2817,91 +2819,51 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+				jsFilesToEncrypt.map((file) => {
```
sgtcoolguy: drop the surrounding parens for single arg arrow functions?

android/cli/commands/_build.js
```diff
@@ -2817,91 +2819,51 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
+					return (done) => {
+						const from = path.join(this.buildAssetsDir, file);
+						const to = path.join(this.buildBinAssetsResourcesDir, file);
+						fs.readFile(from, (error, content) => {
+							if (error) {
+								throw error;
```
sgtcoolguy: will throwing the errors inside nested callbacks do what we want? Don't you need to call `done(err)`?

android/cli/commands/_build.js
```diff
@@ -4104,6 +4066,10 @@ AndroidBuilder.prototype.compileJavaClasses = function compileJavaClasses(next)
 
 	classpath[path.join(this.platformPath, 'lib', 'titanium-verify.jar')] = 1;
 
+	if (this.encryptJS) {
+		classpath[path.join(this.platformPath, 'ti.cloak.jar')] = 1;
```
sgtcoolguy: Just curious why this JAR isn't under the `lib` folder like titanium-verify is...

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -147,10 +151,25 @@ public static InputStream openAsset(String path)
...
 	}
 
+	public static ByteArrayOutputStream readInputStream(InputStream in) throws IOException
+	{
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		byte buffer[] = new byte[1024];
+		int count = 0;
+
+		while ((count = in.read(buffer)) != -1) {
+			if (out != null) {
```
sgtcoolguy: why would `out` ever be null here?

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -147,10 +151,25 @@ public static InputStream openAsset(String path)
 		return null;
 	}
 
+	public static ByteArrayOutputStream readInputStream(InputStream in) throws IOException
```
sgtcoolguy: Is there no common util code for piping streams?

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -163,15 +182,7 @@ public static String readAsset(String path)
...
-			ByteArrayOutputStream out = new ByteArrayOutputStream();
-			byte buffer[] = new byte[1024];
-			int count = 0;
-
-			while ((count = in.read(buffer)) != -1) {
-				if (out != null) {
-					out.write(buffer, 0, count);
-				}
-			}
+			ByteArrayOutputStream out = readInputStream(in);
```
sgtcoolguy: I assume we don't need to worry about charsets here? The default is OK?

android/cli/commands/_build.js
```diff
@@ -4104,6 +4066,10 @@ AndroidBuilder.prototype.compileJavaClasses = function compileJavaClasses(next)
 
 	classpath[path.join(this.platformPath, 'lib', 'titanium-verify.jar')] = 1;
```
sgtcoolguy: do we need titanium-verify anymore?

android/templates/build/AssetCryptImpl.java
```diff
@@ -1,93 +1,94 @@
...
 	}
 
-	private static byte[] filterDataInRange(byte[] data, int offset, int length)
+	private static byte[] getAssetBytes(String path)
 	{
 		try {
-			Class clazz = Class.forName("org.appcelerator.titanium.TiVerify");
-			Method method = clazz.getMethod("filterDataInRange", new Class[] { data.getClass(), int.class, int.class });
-			return (byte[]) method.invoke(clazz, new Object[] { data, offset, length });
+			InputStream in = KrollAssetHelper.getAssetManager().open(path);
```
sgtcoolguy: Rather than read the full input stream into a byte array output stream, then doing a cipher on the byte array; why not use a CipherInputStream so it can stream through?

You're basically reading the full file into memory, decrypting, and returning those bytes - but for some calls you could return the InputStream unread but set up to read as needed (when called by `openAsset`) versus the fully read in version (when called by `readAsset`).

android/cli/commands/_build.js
```diff
@@ -4208,7 +4174,8 @@ AndroidBuilder.prototype.runDexer = function runDexer(next) {
 		}),
 		injarsCore = [
 			this.buildBinClassesDir,
-			path.join(this.platformPath, 'lib', 'titanium-verify.jar')
+			path.join(this.platformPath, 'lib', 'titanium-verify.jar'),
```
sgtcoolguy: extract constants for these files paths since you're re-building them in multiple places.

sgtcoolguy (changes requested):

android/cli/commands/_build.js
cb1kenobi: titanium-verify is responsible for validation Titanium modules, so yeah, I think we still need it.

jquick-axway: We definitely need it. When you include at least 1 module, we make a method call to this library on app startup to verify module licenses. Our mocha test suite is already indirectly testing this since it includes the "ti.map" module.

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -300,18 +300,6 @@ private static String normalizeDirectoryPath(String path)
...
-		final String ROOT_PATH_NAME = "Resources/";
-
-		if (path != null) {
-			if (path.startsWith(ROOT_PATH_NAME)) {
-				path = path.substring(ROOT_PATH_NAME.length());
-			}
-		}
-		return path;
-	}
-
```
jquick-axway: Theoretically, removing the `normalizeAssetCryptPath()` method to strip off the leading "Resources/" from the path would break backward compatibility with a hybrid module's `AssetCryptImpl` generated code.

This means our "ti.playservices" module's `makeGooglePlayServicesAvailable()` method might not be accessible anymore

To be fair, I don't like that our old code was stripping off the "Resources/" directory from the path to begin with. It's a bad assumption on our part. But we kept this part of the code alive for backward compatibility reasons. If we can prove that this is no longer needed anymore, then yes, let's remove this method.

garymathews: Existing hybrid modules work fine, I tested `ti.playservices`. I also removed the dependency of `titanium_prep` when compiling hybrid modules https://github.com/appcelerator/titanium_mobile/pull/11045/commits/1d47632d2ef06057e58bd6a1a27a2e5a1385b09a

garymathews:
@sgtcoolguy @jquick-axway @ypbnv Updated PR, ready for review.

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
jquick-axway: Okay cool. Thanks for testing it out. Then yes, we should get rid of this method.

android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
```diff
@@ -38,6 +39,7 @@
 
 	public interface AssetCrypt {
 		InputStream openAsset(String path);
+		byte[] readAssetBytes(String path);
```
jquick-axway: Adding a method to the `AssetCrypt` interface will cause our "appc-verify" module to crash since it implements it. And we distribute "appc-verify" via the CLI, so, I'd prefer to avoid this. (I also remember some other library implementing this interface too, but I don't remember which.)

Because of this, let's not add a method to this interface. However, we should definitely keep the `KrollAssetHelper.readAssetBytes()` method, which can create the `byte[]` via the interface's `openAsset()` returned `InputStream`.

What if we changed `AssetCryptImpl.getAssetBytes()` to return an `InputStream` type (really the `CypherInputStream`) instead of `byte[]`? Then its `openAsset()` method implementation would like this...
```javascript
@Override
public InputStream openAsset(String path)
{
	return AssetCryptImple.getAssetBytes(path);
}
```

garymathews: Resolved with https://github.com/appcelerator/titanium_mobile/pull/11045/commits/d38282b8401ede36b1f2efca33de624c82ad3ec7

garymathews:
@sgtcoolguy @jquick-axway Updated PR

garymathews:
I ran some benchmarks across four devices to compare the performance increases of `8_3_X` including the following PRs:
- [refactor(android): asset handling](https://github.com/appcelerator/titanium_mobile/pull/11045)
- [build(android): include common bundle in snapshot](https://github.com/appcelerator/titanium_mobile/pull/11054)
- [fix(android): remove usage of getResourceAsStream()](https://github.com/appcelerator/titanium_mobile/pull/11058)

The results are the time taken to launch [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) and execute `app.js`, averaged over four consecutive runs.

|Device|8.0.2|8.1.0|8.3.0|
|:-|:-| :-|:-|
|OnePlus 6T|1175ms|586ms|431ms|
|Pixel 3a|1605ms|805ms|577ms|
|Samsung A50|4225ms|3020ms|774ms|
|Nexus 4|6077ms|3543ms|2079ms|

hansemannn:
Holy sh*t … Gary thats amazing!! Can't wait to test this once merged.

m1ga:
Impressive! Even that older phones (<Android 5) start a lot faster and are still supported! 
Now do the same with the [apk filesize (apk golf)](https://fractalwrench.co.uk/posts/playing-apk-golf-how-low-can-an-android-app-go/) :smile:

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -163,6 +163,30 @@ Local<Value> TypeConverter::javaStringToJsString(Isolate* isolate, JNIEnv *env,
...
+Local<Value> TypeConverter::javaBytesToJsString(Isolate* isolate, JNIEnv *env, jbyteArray javaBytes)
+{
+	if (!javaBytes) {
+		return Null(isolate);
+	}
+
+	jboolean isCopy;
+	jsize nativeBytesLength = env->GetArrayLength(javaBytes);
+	uint8_t* nativeBytes = reinterpret_cast<uint8_t*>(env->GetByteArrayElements(javaBytes, &isCopy));
+	Local<String> jsString = v8::String::NewFromOneByte(V8Runtime::v8_isolate, nativeBytes, v8::NewStringType::kNormal, nativeBytesLength).ToLocalChecked();
```
sgtcoolguy: Ok, so having just done the node string_encoder/Buffer shims this raises a red flag for me. Basically you're making an implicit assumption that the byte array contains a string with a one-byte charset (i.e. ascii, latin-1, binary). That seems like a big no-no here. I think any UTF-8 content outside the typical ASCII range would break this, no? I just think you probably have been lucky and only tried the overlapping character range that exists in both ASCII and UTF-8 and requires only 1 byte to "encode" in UTF-8 - like here: https://ascii.cl
Write some JS code with some umlauts/euros/stuff in strings or identifiers and I'll bet you end up with some garbled source out at the end here.

I would think you'd need to pass along the original encoding of the file plus the bytes that make it up. In reality, I would think UTF-8 would be a good assumption for a charset, but if you had any input files that were say utf-16 you'd have to transcode before encrypting, etc. (And you're already making some assumptions that the bytes are actually utf-8 in some Java code I see in this PR). Note that UTF-8 is the most widely used and for good reason as it packs the most common characters into 1 byte and can grow up to 4 bytes for the esoteric characters (whereas UTF-16 uses 2 or 4 bytes, so is less efficient for common ASCII chars in source code)
(Also, for reference Java uses UTF-16 for strings normally, and so does Javascript - so passing an actual byte array through here as you do should be more efficient memory-wise)

So I guess my suggestion here would be:
- have your encryption generate a byte array, sure, but make sure it holds the utf-8 encoded character data
- change this to ` v8::String::NewFromUtf8()`

sgtcoolguy (changes requested):
I think you're due for a nasty surprise if any source doe contains utf-8 data outside the normal ASCII charset range. Try some files with characters like emoji, or umlauts or euros in strings/identifiers.

android/runtime/v8/src/native/TypeConverter.cpp
garymathews: So I done a quick test using this test case:
```JS
    color: 'black',
    font: { fontSize: '18pt' },
    text: 'Let\'s test UTF-8 characters: \nCopyright: ©\nBritish Pount: £\nEuro: €\nGreek: αβγπ\nTamil: ௵௶௷\nEmoji: ✌😁😺🍔📦'
});

// Let's also test UTF-8 characters in a comment: అೠቜ

win.add(lbl);
win.open();
```

And it works correctly, this is because the byte array is already guaranteed to be UTF-8 encoded before being passed in. But ideally I should be using `v8::String::NewFormUtf8` as `v8::String::NewFromOneByte` assumes Latin-1 data.

.gitignore
```diff
@@ -43,3 +43,4 @@ junit.*.xml
 
 .vscode/
 /android/runtime/v8/src/native/V8Snapshots.h
+/android/templates/build/AssetCryptImpl.java
```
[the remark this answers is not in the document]
garymathews: To prevent the linter from messing up the template

android/cli/commands/_build.js
```diff
@@ -2801,110 +2809,68 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
...
-				}
+		this.logger.info('Encrypting javascript assets...');
+		try {
+			await Promise.all(
+				jsFilesToEncrypt.map(async file => {
+					const from = path.join(this.buildAssetsDir, file);
+					const to = path.join(this.buildBinAssetsResourcesDir, file);
+
+					this.logger.debug(__('Encrypting: %s', from.cyan));
+					const content = await fs.readFile(from);
```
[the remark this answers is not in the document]
garymathews: Ah, I added that after I wrote the CLI changes. Thanks for catching that :)

sgtcoolguy (approved):

sgtcoolguy:
For QE: Please verify that this doesn't break apps with encrypted assets, and in particular try out some JS files that contain UTF-8 multi-byte characters inside of strings (like say emoji and characters like €):
``` javascript
const stringWithSpecialCharacters = '€100.00 to buy ⚽️🏀🏈🐃 ';
```

drauggres:
[ti.cloak](https://github.com/appcelerator/ti.cloak) link from the PR description returns 404. Is `ti.cloak` going to be open sourced?

jquick-axway:
@drauggres, the "ti.cloak" repo is a private repository.

jquick-axway:
@garymathews, this PR "might" break our remote encryption policy feature. Try testing this PR with the following setting in your "tiapp.xml" file.
```xml
	<property name="appc-sourcecode-encryption-policy">remote</property>
</ti:app>
```

When the above is set, our "appc-cli-titanium" node module will fetch all of the JS files from Android build directory "bin/assets" (which is APK assets), encrypt them, generate Java code with the encrypted JS inside of it, and then delete the JS files from assets. I'm worried that these JS files will end up being encrypted twice and will fail to load.
https://github.com/appcelerator/appc-cli-titanium/blob/e7c65ebfb4a62c1eea8f223f2afb21d6be70dd84/hook/titanium.js#L826

drauggres:
> @drauggres, the "ti.cloak" repo is a private repository.

@jquick-axway So, it's going to be proprietary?

jquick-axway:
> So, it's going to be proprietary?

Yes. And we were already using a private repo to handle our current asset/JS encryption. We're just swapping out the older library with this one.

jquick-axway:
FYI: Gary's CLI code changes for the below (including yesterday's commit to this PR) resolves the "remote" policy issue I mentioned above.
https://jira.appcelerator.org/browse/CLI-1370

lokeshchdhry:
@garymathews , can you please resolve the conflicts.

lokeshchdhry:
Building default app/KSv2 to device (deploy type: test) throws:
```
[INFO] :   Encrypting javascript assets...
(node:86607) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of undefined
    at AndroidBuilder.data.fn (/Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/appc-cli-titanium/hook/titanium.js:626:25)
    at /Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/lib/hook.js:260:13
    at /Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:3888:9
    at /Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:473:16
    at replenish (/Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:1006:25)
    at /Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:1016:9
    at eachOfLimit (/Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:1041:24)
    at /Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:1046:16
    at _parallel (/Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:3879:5)
    at Object.series (/Users/lchoudhary/.appcelerator/install/7.1.1/package/node_modules/titanium/node_modules/async/dist/async.js:4735:5)
(node:86607) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:86607) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

expergefactor:
Can confirm that building .APK using the test case results in the same error as shown above.

ssaddique:
Update: 

I have gotten the build to encrypt the files.

**Steps taken:**
1. Used the following line in tiapp.xml
`<property name="appc-sourcecode-encryption-policy">remote</property>`

2. Built app using the terminal command as shown in OP
`appc run -p android --deploy-type test`

**Output**
```
[ERROR] TiAssetHelper: Error while reading asset "Resources/_app_props_.json":
[ERROR] TiAssetHelper: java.io.FileNotFoundException: Resources/_app_props_.json
[ERROR] TiAssetHelper:  at android.content.res.AssetManager.nativeOpenAsset(Native Method)
[ERROR] TiAssetHelper:  at android.content.res.AssetManager.open(AssetManager.java:744)
[ERROR] TiAssetHelper:  at android.content.res.AssetManager.open(AssetManager.java:721)
[ERROR] TiAssetHelper:  at org.appcelerator.kroll.util.KrollAssetHelper.readAssetBytes(KrollAssetHelper.java:179)
[ERROR] TiAssetHelper:  at org.appcelerator.kroll.util.KrollAssetHelper.readAsset(KrollAssetHelper.java:189)
[ERROR] TiAssetHelper:  at org.appcelerator.titanium.TiApplication.loadAppProperties(TiApplication.java:324)
[ERROR] TiAssetHelper:  at org.appcelerator.kroll.KrollRuntime.init(KrollRuntime.java:95)
[ERROR] TiAssetHelper:  at com.axway.timob.Timob26043Application.onCreate(Timob26043Application.java:47)
[ERROR] TiAssetHelper:  at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1154)
[ERROR] TiAssetHelper:  at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5871)
[ERROR] TiAssetHelper:  at android.app.ActivityThread.access$1100(ActivityThread.java:199)
[ERROR] TiAssetHelper:  at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1650)
[ERROR] TiAssetHelper:  at android.os.Handler.dispatchMessage(Handler.java:106)
[ERROR] TiAssetHelper:  at android.os.Looper.loop(Looper.java:193)
[ERROR] TiAssetHelper:  at android.app.ActivityThread.main(ActivityThread.java:6669)
[ERROR] TiAssetHelper:  at java.lang.reflect.Method.invoke(Native Method)
[ERROR] TiAssetHelper:  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
[ERROR] TiAssetHelper:  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

hansemannn:
Same here, it doesn't work so far. @garymathews Maybe some rebasing necessary? Would love to test this on our larger scale app!

garymathews:
For remote policy encryption to work you will need a CLI with the following fix [CLI-1370](https://jira.appcelerator.org/browse/CLI-1370)

garymathews:
Pushed a fix to prevent
```
TypeError: Cannot read property 'length' of undefined
```
And maintain backwards compatibility with older CLIs.

lokeshchdhry:
@garymathews , I am getting this when I build with the remote encryption policy in the tiapp.xml:
I am using the CLI having fix for https://jira.appcelerator.org/browse/CLI-1370.
Seems like a CLI issue ?
```
[TRACE] :  sending build-update {"json":{"session":"ec2a5a57e145e3118b6404a50ee237da9cf032fc JP65hDQ+bP0stNGNJC3W5g4bOL6M9MSJh/YmVgwEQTieVEfTVIFvQdCDSYNjHI0JjgV8/032eHyOf+ByspHThFq5DvtuJidfmHNHIGY41RuoioLm6dxqbSB1g6LQk8tUPd9lLCvRDLHtlgZdEiSXMzLScre65u12eC/h9G8K6GxaA+/yvKo+MGPEGB1dWxWKZX6xmETBdiVesnsmAsuBNbZ/kCdGj/q739VZRQa8UTAlQXohZq3yLVgowe0vvRoC+MC8J/Ob2qiYMZDFSkxwACH+oH0u56cZXeTu/a8qI0Tx9yNfN/nCt+o25n2LC5HeEuB3ap3szcNhhvoX/BgGnerdzUb857Ziv9lRUTURF0WP7sM6MRUYw2XlMud/sz7UdSYPtIudbXzHFwuVWZRrWD7oMNYnnP2smLXPADwJiwBhYHmt0XTPwg8rKmQScvc4asDMZofQkFAJ2n43nKouQR962HA/G0/0IbgEEtztDqTn1y4Qt1lt7A6PIbpkL9NtsbW5sFORV9BVsKxOIufGG5xwijIlIaaRE7hA6humb/wScixdUHnns0lpZf5UrdQ1vgf3560fOc+yNzgpOWgO02OjFdsFmiKXTL1db4LoYZabIETma7TGDNusrql6HpioepKAhx0unjS1P1ngtYDzU7cncas1eFNV3FBwOlTEVgMBGxhrFKvgJMtwcRg+llLky5Ycgbgtq2RrQgk7bmnPKVodtSqrWJFrmSYeiPbCS7o=","buildid":"5db0d823c1059411bdce16e2","buildsha":"6d3ee159c0f682c025143e0621f241474a7dd63e","keys":{"app.js":"b840e55eed2fda05502e8ca50bd4c0c5","ti.cloud/ti.cloud.js":"c693d1c579e2c01aaa4eab490923eb8b","_app_props_.json":"1fef4514df164e3cf95d2738cd22add1","ti.internal/bootstrap.json":"ff6a9170f5f99e64d62e1441c00c749a"}},"followAllRedirects":true,"gzip":true,"url":"https://platform.axway.com/api/v1/auth/build-update"}
[TRACE] :  result from /build-update=> {"success":true}, err=null
[DEBUG] :  TypeError: callback.apply is not a function
    at CLI._fireHookCallback (/Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/lib/hook.js:273:12)
    at /Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/lib/hook.js:252:10
    at /Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:3888:9
    at /Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:473:16
    at replenish (/Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:1006:25)
    at /Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:1016:9
    at eachOfLimit (/Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:1041:24)
    at /Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:1046:16
    at _parallel (/Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:3879:5)
    at Object.series (/Users/lchoudhary/.appcelerator/install/7.1.2-master.3/package/node_modules/titanium/node_modules/async/dist/async.js:4735:5)
[ERROR] :  error copying files. callback.apply is not a function

```

lokeshchdhry:
FR Passed.

Building `test` deploytype apps with encrypted assets run fine.
Special characters in encrypted assets does not break the app build.

OS Ver:           10.14.5
Xcode Ver:        Xcode 11.1
Appc NPM:         4.2.15
Appc CLI:         7.1.2-master.3
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.1
Node Ver:         10.16.1
NPM Ver:          6.9.0
Java Ver:         10.0.2
Android Devices:  ⇨ google Pixel (Android 9)
emulator: Android 4.4

drauggres:
@garymathews 
@jquick-axway 

`ti.cloak` does not provide binaries for linux. Now it's impossible to build the SDK on it.

m1ga:
I've created a ticket: https://jira.appcelerator.org/browse/AC-6420

PR #11049, opened 2019-07-12
chore(android): detailed getter and setter deprecation warning

garymathews:
- Display a verbose deprecation message detailing the occurrence.
- Prevent logs displaying in production builds.

##### TEST CASE
```JS

console.log(win.getBackgroundColor());
win.setBackgroundColor('white');

win.open();
```
```
Getter method deprecated, please use `obj.backgroundColor;` or `obj['backgroundColor'];` instead.
console.log(win.getBackgroundColor());
                ^
    at (/app.js:3:17)
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27242)


sgtcoolguy (approved):
I guess I'm ok, since this wouldn't "spam" the logs so much - but it does vary in an important way in that:
- iOS would still complain at runtime for usage
- This just blanket mentions the deprecation, but doesn't basically "flag" usage

If we want to do it at build time, we could have a babel plugin that flagged the exact usages - and possibly also easily convert it to "fix" usages.

ssekhri:
FR Passed.
The detailed warnings for deprecation of getters and setters is shown.
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200715172045
Appc CLI: 8.0.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Device: Nexus4(v5.1.1) device, Pixel3(v10.0) emulator

PR #11052, opened 2019-07-16
Variety of doc fixes

ewanharris:
This fixes all the warnings generated by the docgen script. Some by fixing typos, and one by introducing support for `Array<type>` syntax in the html generator.

It also proposes a change to the `create2DMatrix` deprecation notice, by proposing `createMatrix2D` instead (which is a more correct replacement). I'd like to do this for `create3DMatrix` too, but I would need to create an entry in `UI.yml` for it with a correct parameter type, and I can't fully figure out what the actual structure for a `Matrix3DCreationDict` would be, if anyone knows, please ping me!

I'd like to propose we backport this into 8.1.0 otherwise we'll be shipping some invalid docs in places

### DIFF
--- a/apidoc/Titanium/Android/Notification.yml
+++ b/apidoc/Titanium/Android/Notification.yml
@@ -447,13 +447,13 @@ properties:
         The application needs to also set the `android.permission.WAKE_LOCK` permission
         in the Android manifest section of the `tiapp.xml` file.
         
-        <ti:app>
-            <android>
-                <manifest>
-                    <uses-permission android:name="android.permission.WAKE_LOCK" />
-                </manifest>
-            </android>
-        </ti:app>
+            <ti:app>
+                <android>
+                    <manifest>
+                        <uses-permission android:name="android.permission.WAKE_LOCK" />
+                    </manifest>
+                </android>
+            </ti:app>
     type: wakeLockOptions
     default: 0
     since: "6.2.0"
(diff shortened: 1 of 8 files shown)


apidoc/Titanium/Android/Notification.yml
```diff
@@ -447,13 +447,13 @@ properties:
...
         in the Android manifest section of the `tiapp.xml` file.
         
-        <ti:app>
-            <android>
-                <manifest>
-                    <uses-permission android:name="android.permission.WAKE_LOCK" />
-                </manifest>
-            </android>
-        </ti:app>
+            <ti:app>
```
ewanharris: The lack of indentation here was causing us to parse this as an API

sgtcoolguy:
@ewanharris So a 3D Matrix on iOS appears to support taking in a `scale` float value: https://github.com/appcelerator/titanium_mobile/blob/master/iphone/TitaniumKit/TitaniumKit/Sources/API/Ti3DMatrix.m#L32

The type also has the following float properties (I assume you can pass in values for them to the `createMatrix3D()` method?):
- m11
- m12
- m13
- m14
- m21
- m22
- m23
- m24
- m31
- m32
- m33
- m34
- m41
- m42
- m43
- m44

In effect we're mirroring the underlying type here: https://developer.apple.com/documentation/quartzcore/catransform3d?language=objc

ewanharris:
@sgtcoolguy, Thanks for the pointers! I played around with `createMatrix3D` to try see what it takes. From what I can tell it only accepts an object with a `scale` property. I was unable to set any of the mXY properties in the creator (tried setting via properties on an object and just passing values into the function).

I've updated the PR to reflect that

sgtcoolguy:
@ewanharris That seems like it should be a bug? Not sure how high priority it is, but I don't see why you shouldn't be able to initialize a 2D or 3D matrix with seed values (the default is to use an identity matrix, I believe, which is presumably what you'd want to start off with in 99.9% of the cases)

PR #11054, opened 2019-07-16
build(android): include common bundle in snapshot

garymathews:
- Include `common` bundle in V8 snapshot for improved launch times

|Device|8.0.2.GA|8.3.0|8.3.0 + PR|
|:-|:-| :-|:-|
|Pixel 3a|1133ms|387ms (reduced by 65%)|303ms (reduced by 73%)|
|Nexus 4|4723ms|3378ms (reduced by 28%)|1666ms (reduced by 65%)|

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27248)


build/lib/packager.js
```diff
@@ -152,8 +113,9 @@ class Packager {
 			this.generateManifestJSON(),
 			// copy misc dirs/files over
 			this.copy([ 'CREDITS', 'README.md', 'package.json', 'cli', 'templates' ]),
-			// transpile/bundle and copy common/ JS files
-			this.transpile(),
+			// copy over 'common' bundle
+			fs.ensureDir(path.join(this.zipSDKDir, 'common', 'Resources')),
```
sgtcoolguy: These two won't run in series, these are all things run in parallel! while this is similar to the `async` library usage we do in the CLI, I'm gathering the all up as Promises to run in parallel via `Promise.all`. If you have specific code that needs to run in series, extract it out to an async function you can call here instead.

android/titanium/src/java/org/appcelerator/titanium/TiLaunchActivity.java
```diff
@@ -96,7 +97,11 @@ protected void loadScript()
 	{
 		try {
 			String fullUrl = resolveUrl(this.url);
-			KrollRuntime.getInstance().runModule(KrollAssetHelper.readAsset(fullUrl), fullUrl, activityProxy);
+			if (KrollAssetHelper.assetExists(fullUrl)) {
+				KrollRuntime.getInstance().runModule(KrollAssetHelper.readAsset(fullUrl), fullUrl, activityProxy);
+			} else {
+				KrollRuntime.getInstance().runModule("global._startup(global)", fullUrl, activityProxy);
```
sgtcoolguy: Don't you need to check against the specific url you want to hijack with a snapshot? Or are we just assuming that ti.main.js won't actually exist because we're not copying it during app build?

sgtcoolguy (changes requested):

hansemannn:
How does this affect build times? Tweaking performance is probably only relevant for production builds, but still interesting to test.

android/titanium/src/java/org/appcelerator/titanium/TiLaunchActivity.java
```diff
@@ -181,6 +186,7 @@ protected void onResume()
 		if (!hasLoadedScript) {
 			hasLoadedScript = true;
 			loadScript();
+			Log.i(TAG, "Launched in " + (System.currentTimeMillis() - TiApplication.START_TIME_MS) + " ms");
```
hansemannn: This is probably a debug left over? Or maybe keep here but on debug level?

jquick-axway: Right. This should be removed since it's for debugging.

Especially since launching the app, backing-out, then re-launching will cause this to log again but with the old start time set upon cold start.

android/titanium/src/java/org/appcelerator/titanium/TiLaunchActivity.java
jquick-axway: I think it's okay for the `else` block to assume the snapshot exists. The old code was already assuming "ti.main.js" was being copied-in by our CLI anyways. And this is the route our mocha test suite will take, so, we'll know when something goes wrong.

Also, "fullUrl" will exist if we copy the "titanium_mobile/common/Resources" folder ("ti.main.js" and it's "ti.internal" directory) to the app project (to be ran via `appc run`) for easy testing/debugging of our core JS changes. This is something I want to make sure we continue to support for "productivity" reasons. In this case the "ti.main.js" we drop in will be favored and the snapshot will be ignored.

jquick-axway:
> How does this affect build times?

We don't currently create snapshots for `appc run` built apps. We create the snapshots when doing a "titanium_mobile" SDK build. Partly because we don't have Windows versions of the V8 command lines tools needed to create the snapshots.

What we're currently doing for 8.1.0 is creating an empty snapshot. This has made V8 "cold startup" time about 2-3x faster. It's pretty significant. You've likely noticed it.

With Gary's change, we're taking a snapshot of our rolled-up common "ti.main.js" code, which has gotten heftier since we're babel transpile/polyfilling it now. This improves the cold startup time and relaunch time. Gary's other PR #11058 further improves the startup time as well.

Note that the newest Android OS versions have a much more aggressive battery optimizers, causing apps to be force quit more often while backgrounded. This makes improving our app cold startup time a high priority.

build/lib/android/startup.ejs
```diff
@@ -0,0 +1,3 @@
+this._startup = function (global) {
+<%- script %>
+}
```
jquick-axway: First, I think what you're doing via the "startup.ejs" is a nice elegant solution for "capturing" a snapshot without running it. I like it.

I do think we should delete/undefine the startup function when invoked to remove it from the global namespace. This will work fine for LiveView since it terminates/restarts the JS runtime when a file changes. And perhaps we should give the startup function a more meaningful name to make it more obvious what we're doing in the `TiLaunchActivity.java` side. What do you think?

```javascript
global.startSnapshot = function (global) {
	delete global.startSnapshot;
<%- script %>
};
```

Oh and is passing the "global" object as an argument necessary?

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -133,6 +134,8 @@ public static void updateActivityTransitionState(boolean state)
 		}
 	}
 
+	public static long START_TIME_MS = 0;
```
garymathews: ##### NOTE: I'm putting this here so we can easily access the start time elsewhere in our SDK.

build/lib/builder.js
```diff
@@ -82,9 +124,51 @@ class Builder {
 		this.program.gitHash = hash || 'n/a';
 	}
 
+	async transpile() {
```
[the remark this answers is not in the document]
garymathews: @sgtcoolguy See https://github.com/appcelerator/titanium_mobile/pull/11054/commits/dc6659ebbf8057cab2738c3f5f9e09018c9f4579 if you're happy with this approach?

garymathews:
@sgtcoolguy @jquick-axway Updated PR

build/lib/android/snapshot.js
```diff
@@ -34,16 +36,32 @@ async function generateBlob(target) {
...
+		'--startup_blob=' + BLOB_PATH,
+		STARTUP_PATH,
+		'--print-all-exceptions'
 	];
 
 	// Snapshot already exists, skip...
 	if (await fs.exists(BLOB_PATH)) {
-		console.warn(`Snapshot blob for ${target} already exists, skipping...`);
-		return;
+		const blobTime = fs.statSync(BLOB_PATH).mtime;
```
sgtcoolguy: already in an async function, why not `await fs.stat(BLOB_PATH)`? Or better: stat both files in parallel and grab the times:
```javascript
const { blobStat, commonStat} = await Promise.all(fs.stat(BLOB_PATH), fs.stat(TI_MAIN_PATH));
if (commonStat.mtime < blobStat.mtime) {
  return;
}
```

Admittedly, this is small potatoes, but we should try and do things in parallel/async when we can. It's made a huge difference in SDK and app build times.

build/lib/builder.js
```diff
@@ -28,6 +38,47 @@ function thisOS() {
 	return osName;
 }
 
+function determineBabelOptions(platform) {
```
sgtcoolguy: Ok, so you're now generating platform-specific versions of the ti.main.js. I definitely is tricky in the sense that technically act platform will get a different copy of the shared code, but obviously means it only gets what it needs (and we trust babel to do the right thing). I think it's a pretty sound approach so long as it doesn't bloat the SDK size (God, I hope the shared code isn't that big!)

But, I would definitely rewrite this SDK build method here still - you're grabbing the android and iOS data you need and then basically deleting the other platform after the fact. The easiest "fix" would be to just construct all of the object we need and then have a switch or if/else blocks to add in the platform-specific `targets` key/value pair(s).

In fact, given that we'd have differing values between platforms, it may make sense to have the platforms code grab the data it needs and return it (i.e. have the IOS, Android and Windows classes have a shared method called babelTarget() or something that returns an Object we can inject into the common babel options object we're creating, like: `{ iOS: minSupportedIosSdk }`.

I get that I'm being super-picky here...

 If you go up the call stack here you're calling #transpile() right before creating an instance of the Platform class and calling build() on it anyways. maybe just construct the object and pass it in to the transpile method here, or come up with a shared base class for the Platform builder classes that holds the common transpile logic and have the subclasses override some method to return the relevant `targets` object.

build/lib/builder.js
```diff
@@ -82,11 +133,53 @@ class Builder {
...
+		await fs.remove(TMP_COMMON_DIR);
+	}
+
 	async build() {
 		await this.ensureGitHash();
 		console.log('Building MobileSDK version %s, githash %s', this.program.sdkVersion, this.program.gitHash);
+
 		// TODO: build platforms in parallel
 		for (const item of this.platforms) {
+			await this.transpile(item);
```
sgtcoolguy: Here's the part I was referring to in the above comment - right below this you're constructing the Platform-specific "builder" where the logic to supply the babel `targets` value would make sense to live.

sgtcoolguy (approved):
I think this general approach can be useful in that each platform would get only the set of polyfills/shims it needed. I'd tweak the SDK build logic to not have to produce the ios/android data and then just wipe it out of the object if it's a different platform.

Also, this presumably means we'd need to tweak the windows CLI code to point at common/Resources/windows like you do for Android/IOS here (Unless it's just all smart enough to pick up the platform-specific subfolder magically)

sgtcoolguy:
@garymathews Also, just a note that the work you've done on Android startup is *awesome*. Those timing numbers above are insane.

I think the community is in for a real surprise between our app build time improvements and then app startup improvements on Android in 8.1 - 8.3 timeframe!

garymathews:
@sgtcoolguy Thanks for the feedback, updated PR

android/titanium/src/java/org/appcelerator/titanium/TiLaunchActivity.java
```diff
@@ -105,6 +113,9 @@ protected void loadScript()
 	@Override
 	protected void onCreate(Bundle savedInstanceState)
 	{
+		// set start time
+		TiApplication.START_TIME_MS = System.currentTimeMillis();
+
```
jquick-axway: We should **not** use the `System.currentTimeMillis()` method since it is affected by system clock changes. We should use the `SystemClock.uptimeMillis()` method instead, which is not affected by clock changes and only counts up. It's perfect for timing measurements and for implementing timeouts.
https://developer.android.com/reference/android/os/SystemClock
https://developer.android.com/reference/android/os/SystemClock#uptimeMillis()

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -48,6 +48,7 @@
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
+import java.lang.System;
```
sgtcoolguy: not used?

sgtcoolguy (approved):
LGTM

build/lib/ios.js
```diff
@@ -27,6 +29,17 @@ class IOS {
 		this.timestamp = options.timestamp;
 	}
 
+	babelOptions() {
+		// eslint-disable-next-line security/detect-non-literal-require
+		const minSupportedIosSdk = version.parseMin(require(path.join(ROOT_DIR, 'iphone', 'package.json')).vendorDependencies['ios sdk']);
```
ewanharris: This'll need updating to use `minIosVersion` per #11091/[TIMOB-27292](https://jira.appcelerator.org/browse/TIMOB-27292)

garymathews:
Updated PR, squashed commits; ready for FR.

jquick-axway:
@garymathews, this PR "might" break our remote encryption policy feature. Try testing this PR with the following setting in your "tiapp.xml" file.
```xml
	<property name="appc-sourcecode-encryption-policy">remote</property>
</ti:app>
```

When the above is set, our "appc-cli-titanium" node module will fetch all of the JS files from Android build directory "bin/assets" (which is APK assets), encrypt them, generate Java code with the encrypted JS inside of it, and then delete the JS files from assets. I'm worried that these JS files will end up being encrypted twice and will fail to load.
https://github.com/appcelerator/appc-cli-titanium/blob/e7c65ebfb4a62c1eea8f223f2afb21d6be70dd84/hook/titanium.js#L826

lokeshchdhry:
Building an app with artifact from this PR errors with:
```
[ERROR] :  TiExceptionHandler: (main) [21,21] ti:/module.js:302
[ERROR] :  TiExceptionHandler: 	throw new Error('Requested module not found: ' + request); // TODO Set 'code' property to 'MODULE_NOT_FOUND' to match Node?
[ERROR] :  TiExceptionHandler:  ^
[ERROR] :  TiExceptionHandler: Error: Requested module not found: ../../../../../../../../../../semantic.colors.json
[ERROR] :  TiExceptionHandler:     at Module.require (ti:/module.js:302:8)
[ERROR] :  TiExceptionHandler:     at require (ti:/module.js:566:15)
[ERROR] :  TiExceptionHandler:     at startSnapshot (<embedded>:7:39)
[ERROR] :  TiExceptionHandler:     at /ti.main.js:1:8
[ERROR] :  TiExceptionHandler:     at Module._runScript (ti:/module.js:591:17)
[ERROR] :  TiExceptionHandler:     at Module.load (ti:/module.js:107:7)
[ERROR] :  TiExceptionHandler:     at Function.Module.runModule (ti:/module.js:75:9)
[ERROR] :  TiExceptionHandler:
[ERROR] :  TiExceptionHandler:     org.appcelerator.kroll.runtime.v8.V8Runtime.nativeRunModule(Native Method)
[ERROR] :  TiExceptionHandler:     org.appcelerator.kroll.runtime.v8.V8Runtime.doRunModule(V8Runtime.java:162)
[ERROR] :  TiExceptionHandler:     org.appcelerator.kroll.KrollRuntime.runModule(KrollRuntime.java:211)
[ERROR] :  TiExceptionHandler:     org.appcelerator.titanium.TiLaunchActivity.loadScript(TiLaunchActivity.java:106)
[ERROR] :  TiExceptionHandler:     org.appcelerator.titanium.TiRootActivity.loadScript(TiRootActivity.java:480)
[ERROR] :  TiExceptionHandler:     org.appcelerator.titanium.TiLaunchActivity.onResume(TiLaunchActivity.java:194)
[ERROR] :  TiExceptionHandler:     org.appcelerator.titanium.TiRootActivity.onResume(TiRootActivity.java:499)
[ERROR] :  TiExceptionHandler:     android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1412)
[ERROR] :  TiExceptionHandler:     android.app.Activity.performResume(Activity.java:7292)
[ERROR] :  TiExceptionHandler:     android.app.ActivityThread.performResumeActivity(ActivityThread.java:3776)
```
This does not happen with master SDK from jenkins.

jquick-axway:
@garymathews,

The exception @lokeshchdhry ran into might be because we are doing a `require()` on the new semantic color JSON file instead of loading it via our file APIs (which is how we load our "bootstrap.json" file).
https://github.com/appcelerator/titanium_mobile/pull/11097/files#diff-e66c10e6fedb17f9d452e0cfd2d1ff83R45

garymathews:
@lokeshchdhry I've fixed the `require` in `ti.ui.js`. It was an unrelated issue to this PR, but a simple fix that I've included here.

ewanharris:
@garymathews, that shouldn't have been needed, this works fine on master, it looks like the rollup configuration to ignore the semantic colors require when rollup does it stuff was just removed in this PR [master](https://github.com/appcelerator/titanium_mobile/blob/28907e0c3808ca019ee02cdc167d6ef4314f0b0b/build/lib/packager.js#L274-L281), [this PR](https://github.com/garymathews/titanium_mobile/blob/9bf98405645a435b768a4d20cb78da175097259f/build/lib/builder.js#L127-L132)

jquick-axway:
@lokeshchdhry, regarding that exception, I assume the app didn't crash. The exception was simply logged right?

I'm asking because I see a try/catch block around the `require()` JS function that was causing the exception. I want to make sure that safety mechanism still works.

lokeshchdhry:
@jquick-axway , yes the app does not crash but a runtime error is thrown. It says at the splash screen though.

jquick-axway:
That would suggest our try/catch handling isn't working here.

And yet the try/catch block we have in "ti.main.js" does work when attempting to require-in the ACA module that isn't there.

Hmm...

garymathews:
@jquick-axway Rollup pulled the `require` out of the try catch. But it's fine now, I've addressed the issue.

lokeshchdhry:
FR passed.

Checked it with default app & KS. Improvement in launch time is seen.

OS Ver:           10.14.5
Xcode Ver:        Xcode 11.0
Appc NPM:         4.2.15
Appc CLI:         7.1.1
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.1
Alloy Ver:        1.14.1
Node Ver:         10.16.1
NPM Ver:          6.9.0
Java Ver:         10.0.2
Android Devices:  ⇨ google Pixel (Android 9), Android 4.4, android 6.0

hansemannn:
We just tried this change and it's causing even longer launch times (~3s on a Mi A2) compare to before. You can test our app internally to reproduce the issue. Possible regression reasons:
- We have an own `our_app.bootstrap.js` file that loads things like Firebase, which might cause issues
- We have some polyfills in that file that might cause issues

In any case, it should't get worse with this change, but it seems like it does.

jquick-axway:
Well if it reduced your launch time by 0.5 seconds, then I would say that is still an improvement. Every bit counts. And there are a couple other pending PRs that will improve launch times as well. One removes our usage of the Java `getResourcesAsStream()` and the other cuts down on the needless UTF-8/UTF-16 transcoding we were doing when requiring-in scripts.

PR #11057, opened 2019-07-17
 chore(ios): prevent modal windows from being swiped down

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27169


janvennemann:
@vijaysingh-axway the commits don't appear to be related to the ticket 😄

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -558,6 +558,13 @@ - (void)openOnUIThread:(NSArray *)args
           [theController setModalPresentationStyle:style];
         }
       }
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+      if ([TiUtils isIOSVersionOrGreater:@"13.0"]) {
+        isForceModalDisable = ![TiUtils boolValue:@"forceModal" properties:dict def:NO];
```
hansemannn: So it's creation only?

vijaysingh-axway: This should be pass as parameter while opening window.

janvennemann: The double negation here and in the following line is confusing. Consider changing the variable name so you don't have to negate the result twice, for example:

```suggestion
        forceModalBehavior = [TiUtils boolValue:@"forceModal" properties:dict def:NO];
```

janvennemann (changes requested):

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -558,6 +558,13 @@ - (void)openOnUIThread:(NSArray *)args
           [theController setModalPresentationStyle:style];
         }
       }
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
```
janvennemann: Use the new `IS_IOS_SDK_13` constant

vijaysingh-axway: This macro will not work inside framework.

janvennemann: Duh, of course.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
vijaysingh-axway: Updated.

hansemannn:
@vijaysingh-axway Can you squash the commits please? Otherwise it's hard to cherry pick

hansemannn:
We tested this in two places of our app and it does not work. Did you try this in an Alloy based app (inside the tss)?

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -46,6 +46,7 @@ - (void)_destroy
 
 - (void)_configure
 {
+  forceModal = YES;
```
janvennemann: Why is this set to `YES` here when it's default value is defined as `NO`?

janvennemann: Ah, I was just wondering what's the reason behind that. Thanks for the explanation.

janvennemann (approved):

ssjsamir (approved):
**FR Passed** Modal windows no longer being swiped down. Tested with the two test case from https://jira.appcelerator.org/browse/TIMOB-27169

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 11 beta 5
Node.js ^8.11.1
iPhone 8 sim (13.0)
```

PR #11076, opened 2019-07-24
fix(ios): correctly read source-maps setting from tiapp.xml

janvennemann:
@sgtcoolguy this one slipped through in #10951


sgtcoolguy (approved):
@janvennemann LGTM, but ESLint wants you to change it to:
```javascript
else if (Object.prototype.hasOwnProperty.call(cli.tiapp, 'source-maps')) {
```

sgtcoolguy:
Manually merged after fixing ESLint issue

PR #11082, opened 2019-07-26
feat(android): Add contentSize to scroll event

m1ga:
- Parity for `contentSize` inside the `scroll` event of a `ScrollView`

##### TEST CASE
```js
    backgroundColor: 'red',
    top: 0,
    width: 20,
    height: 1020
});
const scrollView = Ti.UI.createScrollView({
    height: Ti.UI.FILL,
    width: Ti.UI.FILL
});

scrollView.addEventListener('scroll', e => {
	console.log('contentSize:', e.contentSize);
});

win.add(scrollView);
scrollView.add(view);
win.open();
```
```
contentSize: {"width":411,"height":1020}
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27283)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
```diff
@@ -448,9 +445,26 @@ protected void onScrollChanged(int l, int t, int oldl, int oldt)
...
+			int contentWidth = 0;
+			int contentHeight = 0;
+
+			if (getChildCount() > 0) {
+				final View child = getChildAt(0);
+				TiDimension dimW = new TiDimension(child.getWidth(), TiDimension.TYPE_WIDTH);
+				TiDimension dimH = new TiDimension(child.getHeight(), TiDimension.TYPE_HEIGHT);
+				contentWidth = TiConvert.toInt(dimW.getAsDefault(child), 0);
+				contentHeight = TiConvert.toInt(dimH.getAsDefault(child), 0);
+			}
```
jquick-axway: Variables `contentWidth` and `contentHeight` should be of type `double` and not `int`. And we should not convert the `double` returned by `getAsDefault()` to an integer.

The reason is because the `contentSize` you are providing needs to use the units defined in the "tiapp.xml", which will typically be "dp" by default (not pixels). "dp" based values can be **fractional**. Especially on an "hdpi" Android device which has a `1.5` scaling factor. Test it out in an "hdpi" emulator and you'll see what I mean.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
```diff
@@ -575,13 +589,28 @@ protected void onScrollChanged(int l, int t, int oldl, int oldt)
...
 				getProxy().fireEvent(TiC.EVENT_DRAGSTART, data);
 			}
+
+			if (getChildCount() > 0) {
+				final View child = getChildAt(0);
+				TiDimension dimW = new TiDimension(child.getWidth(), TiDimension.TYPE_WIDTH);
+				TiDimension dimH = new TiDimension(child.getHeight(), TiDimension.TYPE_HEIGHT);
+				contentWidth = TiConvert.toInt(dimW.getAsDefault(child));
+				contentHeight = TiConvert.toInt(dimH.getAsDefault(child));
+			}
```
jquick-axway: Same as above. Content width/height should be doubles, not ints.

apidoc/Titanium/UI/ScrollView.yml
```diff
@@ -114,9 +114,9 @@ events:
       - name: contentSize
         summary: |
             The current content size of the scroll view defined by its `width` and `height` properties.
-            Available in Titanium SDK 5.2.0 and later.
         type: Dictionary
-        platforms: [iphone, ipad]
+        platforms: [android, iphone, ipad]
+        since: { iphone: "5.2.0", ipad: "5.2.0", android: "8.2.0" }
```
jquick-axway: Titanium 8.2.0 is planned to be an iOS 13 only release.
No Android or Windows changes are allowed.

Would you mind changing the version from 8.2.0 to 8.3.0 please?

jquick-axway (changes requested):

jquick-axway:
Thanks for the contribution @m1ga .
I wasn't aware of the event's "contentSize" property on iOS. I'm sure this will help.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
m1ga: done

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
m1ga: done

m1ga:
Thanks, all issues fixed

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
```diff
@@ -448,9 +445,26 @@ protected void onScrollChanged(int l, int t, int oldl, int oldt)
 				KrollDict data = new KrollDict();
 				getProxy().fireEvent(TiC.EVENT_DRAGSTART, data);
 			}
+
+			double contentWidth = 0;
```
garymathews: Maybe make a method that returns a `contentData` `KrollDict` so we don't have duplicate code?

m1ga: good idea! Will change that

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollView.java
```diff
@@ -1061,6 +1063,27 @@ public void scrollToTop()
...
+	private KrollDict contentSize()
+	{
+		double contentWidth = 0;
+		double contentHeight = 0;
+		KrollDict contentData = new KrollDict();
+		View nativeView = this.nativeView;
+		TiScrollViewLayout lay = getLayout();
+
+		if (lay.getChildCount() > 0) {
+			final View child = lay.getChildAt(0);
```
m1ga: I need to check that again. I think it might return the wrong child

garymathews (approved):
Updated PR, LGTM!

No tests as this requires user input.

keerthi1032:
FR passed.shows content size inside scroll  view
Test Environment:
 Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.3.0.v20190903144732
Device -samsung s5 android 6, oneplus 5t android 9
Simulator-pixel android 9,nexus 6p android 7

sgtcoolguy:
manually rebased on top of master and merged.

PR #11088, opened 2019-07-27
docs: replace docgen.js with titanium-docgen

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIDOC-3337

**Optional Description:**
This moves all JS files in our `apidocs` folder into a new module so the code can be re-used. The module can be found at https://github.com/appcelerator/docs-devkit/tree/develop/packages/titanium-docgen

There is one notable change: The docgen script now requires at least one apidocs path as an argument that it can work on. Previously this would default to the `apidocs` directory in this repo, which is obviously no longer possible.


sgtcoolguy:
@janvennemann So one major thing that Danger is flagging here is that this removes our apidoc validation. Basically I had written up a JS script to validate the format of our API Docs on `npm test` so that it could flag if we used incorrect types, didn't follow spec, etc. I think that's a pretty useful thing to try and retain and would need to be kept in `titanium-docgen` and exposed in some way. (Now, I didn't hook it up to lint-staged because it does doc validation as a batch - I think it needs all doc files to look up types)

janvennemann:
@sgtcoolguy Yup, i noticed that too. I already have the script ported to the new module and setup an additional binary to run it. Just waiting for Jenkins to pass so i can update the PR here.

sgtcoolguy:
@janvennemann I updated this PR to not remove the images folder from apidocs, and opened a PR on titanium-docgen to support copying images folder from input paths.

PR #11092, opened 2019-07-29
Update WebView.yml

jawa9000:
Added WKWebView section per https://jira.appcelerator.org/browse/TIDOC-3342


jquick-axway:
We may want to document a scaling difference between `WKWebView` and `UIWebView` differences.

To have a `WKWebView` scale the page the same way as `UIWebView`, add the following meta tag to the HTML header...
```html
<html>
	<head>
		<meta name="viewport" content="width=device-width, initial-scale=1.0">
	</head>
</html>
```

PR #11094, opened 2019-07-30
build(android): rework merging of external library's resources

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27252

**Description:**
Rework the merging of third party native libraries resources. In order to try to keep backwards compatibility the merging now:
- appends children nodes with different names for parents with the same name;
- overwrites only children with the same names for parents with the same name (previously it overwrote the parent). This happens in order of expanding the libraries.

**Test case:**
There is an archive in the JIRA ticket with a version of a third party native library. Put the AAR files under `app/platform/android` for an Alloy project and build the project.


sgtcoolguy:
@ypbnv Given that this is an Android fix/improvement, shouldn't this be scheduled for 8.3.0 milestone/release?

ypbnv:
@sgtcoolguy Ah, my bad. I will change it. BTW this change may be irrelevant once we move to using `gradle`, because it is dealing with merging resource files. Maybe we can consider backporting it for a release before it?

janvennemann (changes requested):

ypbnv:
@janvennemann Updated.

janvennemann (approved):

jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.14.5
SDK: 8.3.0.v20191008074512
Appc CLI: 7.1.1
JDK: 1.8.0_162
Node: 10.5.0
Studio: 5.1.4.201909061933

PR #11095, opened 2019-07-31
feat(node): add string_decoder module

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27286

**Description:**
Follow on to #11084 which adds Buffer

This adds the `string_decoder` module, which is sort of a safer way to interact with Buffers when dealing with character data.

Basically a string decoder will temporarily cache/buffer up incomplete multi-byte characters until they're completed before outputting them. This is handy if you expect multi-byte utf8/utf16/base64 data and want to spit it out as you go.

**Caveats**
V8 and Node handle invalid UTF-8 multi-byte characters by replacing invalid/incomplete multi-byte sequences with '\ufffd'. This implementation does not - and on a platform like iOS can result in `undefined` being returned. I've left comments about this in here for now - but the "fix" here is 2-fold:
- Have iOS/Windows explicitly handle UTF-8 by injecting the replacement characters (there's a link to a gist that shows this for iOS, but it's a fair amount of code) - Note that this would fix our Ti.Buffer handling of invalid data as well.
- Have this module do the checking/replacement. Given that the data lives across the JS/native boundary this could be performance-heavy. If our Buffer module got changed to use Uint8Array in the majority of cases, this may not be such an issue - but being backed by a Ti.Buffer and traversing the bridge for every byte means it would be pretty expensive to iterate through to find and validate every multi-byte char sequence.

The only "invalid" case we do handle is when you call `decoder.end()` and it has a buffered incomplete multi-byte character.


sgtcoolguy:
⚠️ **Note that this should be held until #11084 is merged (and then this should be rebased)**

janvennemann (approved):
Looks good! Just added two minor notes in the test suite but nothing important.

I guess we have to see if and how the caveats have any impact on third-party modules that use `string_decoder`. Regarding a fix for this i'd vote for the actual fix in `Ti.Buffer` on iOS to get the correct behavior natively, assuming that we don't make the switch to UIInt8Array for our buffers anytime soon.

sgtcoolguy:
Merged manually to master.

PR #10881, opened 2019-05-10
feat(android): Added Ti.Platform.canOpenURL() support

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-14460

**Summary:**
- [TIMOB-14460](https://jira.appcelerator.org/browse/TIMOB-14460) Added `Ti.Platform.canOpenURL()` support. _(Parity with iOS and Windows.)_
- Added ability to open app's sandboxed files via `openURL()` and `canOpenURL()`.
  * iOS supports something like this via [Ti.UI.iOS.DocumentViewer](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.iOS.DocumentViewer) which displays file in-app.
  * This provides a similar equivalent on Android. (We should do the same for `startActivity()` method in the future.)
- Updated `TiFileProvider.java` to provide external access to embedded files within the APK's "assets" and "res" folders.
- Updated `TiMimeTypeHelper.java` to fetch mime type from image, audio, and video files if they don't have a file extension.
  * Requires scanning the binary. Needed for APK "res" files not under "res/raw".
- Unit test already written in mocha suite. Need to remove `it.androidMissing` when merged.

**Test:**
1. Create a new Titanium Classic "Default Project".
2. Copy the below code to the proejct's "app.js" file.
3. Build and run on Android.
4. Notice that a table view of tests are displayed. Tap on the "canOpenURL" and "openURL" buttons for each row and verify it matches the results below.

| URL | canOpenURL() | openURL() |
| --- | ------------ | --------- |
| `"http://www.appcelerator.com` | `true` | Opens Appcelerator website in web browser app. |
| `https://www.appcelerator.com` | `true` | Opens Appcelerator website in web browser app. |
| `mailto:user@domain.com` | `true` | Opens new draft in e-mail app with "user@domain.com" assigned to "To" field. |
| `mailto:user@domain.com?cc=johndoe@domain.com ...` | `true` | Opens new draft in e-mail app. "TO" field set to "user@domain.com". "CC" field set to "johndoe@domain.com" and "janedoe@domain.com". "BCC" field set to "bsmith@domain.com". "Subject" field set to "The Subject". "Body" field set to "This is a test. Hello World!". |
| `tel:123-456-7890` | `true` on phone. `false` on tablet. | Opens phone app with phone number set to "(123) 456-7890". |
| `geo:37.774,-122.431?z=12` | `true` | Displays San Francisco in maps app. |
[...]

PR #11098, opened 2019-07-31
fix(android): handle file: URIs without // after scheme

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27191

**Optional Description:**
Fix extracted from work on #10998

This specifically fixes support for `file:` URIs that are of the form:
- `file:/path/to/file`
- `file:relative/path`

Before it only supported file URIs that were of the form:
- `file:///path/to/file`

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
@@ -1,12 +1,14 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2009-2010 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-2019 by Appcelerator, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 package org.appcelerator.titanium.io;
 
 import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Date;
 import java.util.HashSet;
 
@@ -34,6 +36,10 @@ public class TiFileFactory
 	private static final String CONTENT_URL_PREFIX = CONTENT_URL_SCHEME + "://";
 	private static final String FILE_URL_SCHEME = ContentResolver.SCHEME_FILE;
 	private static final String FILE_URL_PREFIX = FILE_URL_SCHEME + "://";
+	// strip file:// prefix off the special URL we need to handle like app:
+	private static final String ANDROID_ASSET_RESOURCES =
+		TiC.URL_ANDROID_ASSET_RESOURCES.substring(FILE_URL_PREFIX.length());
+	private static final String TI_URL_SCHEME = "ti";
 	private static HashSet<String> localSchemeSet;
 
 	static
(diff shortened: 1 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
```diff
@@ -71,87 +77,94 @@ public static TiBaseFile createTitaniumFile(String path, boolean stream)
...
+			return new TitaniumBlob(possibleURI); // TODO: Forward along the actual URI instance?
+		}
+
+		if (FILE_URL_SCHEME.equals(scheme)) {
+			// check for fake "file:///android_asset/Resources/" URL, treat like app:
+			if (path.startsWith(ANDROID_ASSET_RESOURCES)) {
+				// Strip this fake base path
+				path = path.substring(ANDROID_ASSET_RESOURCES.length());
+				return new TiResourceFile(trimFront(path, '/')); // remove leading '/' characters
+			}
```
jquick-axway: The "file:///android_asset/" URL is not a fake path. It's officially supported by Google and is used to access files from the APK's assets folder. But I'm pretty sure it cannot be loaded by a Java `File` class. This seems to be an issue with the old code. We need to do the below else-if block to turn the given URL into a "content://" so that it can be loaded by our own content provider. Note that I haven't tested this yet. I need to confirm that we can access an APK's root "assets" file such as "file:///android_asset/app.json".
```java
} else if (path.startsWith("/android_asset/")) {
	Uri contentUri = TiFileProvider.createUriFrom(possibleURI).toString();
	if (contentUri != null) {
		return new TitaniumBlob(contentUri.toString());
	}
	return null;
}
```

FYI: The "file:///android_asset/Resources" path needs special handling in Titanium because the given path may reference encrypted assets such as a `.js` or `.json` file. This is where a Titanium project's "Resources" files gets dumped.

jquick-axway:
We should grab the relevant file url/path tests from the following PRs and turn them into unit tests. They exercise several different URL schemes and our relative path handling.
- https://github.com/appcelerator/titanium_mobile/pull/10881
- https://github.com/appcelerator/titanium_mobile/pull/10304

Would you like me to help write up the unit tests?

I definitely need to double check our "file:///android_asset/" handling outside of our "Resources" directory too.

android/titanium/src/java/org/appcelerator/titanium/io/TiFileFactory.java
jquick-axway: Oh bummer. The above code causes infinite recursion. I'll have to come up with a different solution.

jquick-axway: Ideally, we should wrap a path such as "file:///android_asset/app.json" with our `TiResourceFile` since it supports fast directory listing support (we can't get a directory listing from a "content://" URL which is what my above code is turning it into). Unfortunately, our `TiResourceFile` class is currently limited to files under the "file:///android_asset/Resources" directory, but I believe @garymathews's upcoming snapshot PR will change this since it now wants the "Resources/" base directory.

So, this should be a task for later.

For now, either return `null` for this case or I guess continue to return a broken `TiFile` like the old code was doing before.

sgtcoolguy:
@jquick-axway Sorry, I've been bouncing all over the place so didn't see these comments until now. Yeah feel free to hack this up, add tests, or fix it in any way.

These PRs grew out of implementing the Node-compatible `fs` stuff, so as long as they fix the underlying issues I'm ok if you tweak it to be more compatible with other APIs/future anticipated work here.

sgtcoolguy:
So I've updated this with a number of tests. The reality is **not** pretty. I added a test for the very specific issue I had fixed that was breaking the Node `fs` shim PR (#10998): basically moving/copying with `file:` URIs that din't have the `//` after the colon would break. The test now passes and that's awesome. I'd like to keep this ticket/PR limited to that issue if we can.

I had also tried to add tests around various possible URIs to refer to files, but iOS and Android are woefully different, breaking in different ways. Also, the platforms are inconsistent *internally* in that different APIs handle URIs/paths differently (see above that `Ti.Filesystem.getFile()` doesn't actually surface the fixed bug, but `Ti.Filesystem.File.move()` and `Ti.Filesystem.File.copy()` do!).
I think this is a much larger issue and should likely be dealt with in a separate ticket that tries to enforce parity in URI handling across platforms for `Ti.Filesystem.getFile()`  -and maybe also follow-on tickets to ensure the other APIs play by the same rules too (i.e. Webview references, `Ti.App.appURLToPath()`, `Ti.Platform.canOpenURL()`, `Ti.Database.open/install`, `Ti.Network.HTTPClient.file`, `Ti.Filesystem.File.move()` and `Ti.Filesystem.File.copy()`, `TiUrl` class in Android)

jquick-axway:
Personally, I wish our `File` object only supported file system paths; not URLs. But what's done is done. Changing it would probably be too huge of a breaking-change.

I will say that Android "content://" URLs are very commonly used for accessing sandboxed files like photos/videos and they might not actually reference files on the file system (they can be in-memory blobs, database blobs, etc.). In this case, there is no notion of "directory". This also means there is no means of getting a directory listing of files from a "content://". This is by design since it would break the sandboxing of files. (Google does support a database row cursoring feature via `CURSOR_DIR_BASE_TYPE`, but most apps don't implement it in their `ContentProvider`.)

One thing that bothers me is that if we receive a relative path without a scheme, it defaults to the `applicationDataDirectory` on Android. I would expect it to default to `resourcesDirectory`. I haven't had time to check what iOS' default behavior is.

PR #11106, opened 2019-08-02
Speculative db device fix

sgtcoolguy:

**Optional Description:**
This is a speculative fix to see if this helps the XT1032 Android device pass the DB test suite. Right now it does not seem to "interrupt" a series of async queries if we call `close()` on a timer.

I changed from using a `synchronized(this)` block to using an explicit `Lock` instance that we lock/unlock - with the query method that may be running off main thread acquiring via `lockInterruptibly()`. Additionally, the lock implementation is set to be "fair" so that it gives preferences to the longest waiting thread. This should ensure threads acquire the lock in-order


garymathews (approved):
CR: **PASS** 👍

sgtcoolguy:
cherry-picked to master

PR #11109, opened 2019-08-03
fix(ios): accessibility properties not returning proper value

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-24171


iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
```diff
@@ -363,7 +363,10 @@ - (void)setAccessibilityHint_:(id)accessibilityHint
 
 - (void)setAccessibilityHidden_:(id)accessibilityHidden
 {
-  self.accessibilityElementsHidden = [TiUtils boolValue:accessibilityHidden def:NO];
+  id accessibilityElement = self.accessibilityElement;
+  if (accessibilityElement != nil) {
```
janvennemann: Is the nil check necessary here? Won't it be a noop anyway sending the below message to `accessibilityElement` if it's nil.

Also use the property style accessor `accessibilityElementsHidden `.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
```diff
@@ -2838,6 +2838,31 @@ - (void)setAccessibilityHidden:(id)accessibilityHidden
...
 }
 
+- (NSObject *)accessibilityElement
+{
+  return self.view.accessibilityElement;
+}
+
+- (NSString *)accessibilityLabel
+{
+  return [self accessibilityElement].accessibilityLabel;
```
janvennemann: Use `self.accessibilityElement` here and below (like you already did in `TiUIView.m`)

janvennemann (changes requested):
Just some minor notes about code style. Rest looks good!

iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
vijaysingh-axway: Here 'accessibilityElement' is internal function. In TiUIView.m it is defined as property because it has to be used outside as well.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
vijaysingh-axway: If you see methods above this method, they use similar way. I think nil check was added to avoid to execute the other 3 statements (it is implemented previously). To make it more aligned with other methods (above), I have used method style accessor.

janvennemann: Ah ok, that's alright then

iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
janvennemann: Oh damn, didn't notice. Thanks for the clarification.

janvennemann (approved):
LGTM!

ssekhri:
FR passed.
Accessibility label, value and hint returned successfully in both classic and alloy apps.

Verified on:
Mac OS 10.14.6
Ti SDK: 8.2.0.v20190802170526
Appc CLI: 7.1.0
Node: 10.5.0
JDK: 10.0.2
Xcode: 10.3
Studio: 5.1.3.201907112159

PR #11116, opened 2019-08-06
Update WebView.yml

jawa9000:
Update for documentatio for broken Ti.UI.SIZE for WebViews per https://jira.appcelerator.org/browse/TIDOC-3355


apidoc/Titanium/UI/WebView.yml
```diff
@@ -127,6 +127,56 @@ description: |
...
+            {
+              font-family: 'Lato-Regular';
+              src: url('fonts/Lato-Regular.ttf');
+            }
+        </style>
+
+    ### Ti.UI.SIZE and WebViews
+
+    With Titanium 8.0.0+, <Titanium.UI.SIZE> does not work for WebViews. We recommend to give a **fixed height** to <Titanium.UI.WebView> (as noted in [TIDOC-3355](https://jira.appcelerator.org/browse/TIDOC-3355)).
+
```
vijaysingh-axway: Can you add  something like  --

> As a workaround you can try to get the 'document.body.scrollHeight' inside <Ti.UI.WebView.load> event of webview  and set the height to webview. See following example.

jawa9000: I've added your comment.

PR #11117, opened 2019-08-06
chore(ios): exposed new constants of blur effect style

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27310


iphone/Classes/TiUIiOSProxy.m
```diff
@@ -443,6 +443,24 @@ - (id)BLUR_EFFECT_STYLE_PROMINENT
   }
   return [NSNull null];
 }
+
+#if IS_SDK_IOS_13
+MAKE_SYSTEM_PROP(BLUR_EFFECT_STYLE_SYSTEM_ULTRA_THIN_MATERIAL, UIBlurEffectStyleSystemUltraThinMaterial);
```
janvennemann: These new properties are missing a check when running on devices with iOS < 13.. See https://jira.appcelerator.org/browse/TIMOB-23778 and the above `BLUR_EFFECT_STYLE_*` properties.

Maybe this is good chance to extend the `MAKE_SYSTEM_PROP` macro and add an additional parameter to do the above runtime check?

janvennemann (changes requested):

iphone/Classes/TiUIiOSProxy.m
vijaysingh-axway: @janvennemann 
If we are building with iOS SDK 13, these all constants will be available in in iOS 13+ and iOS  < 13. But for iOS < 13  ( [[self blurView] setEffect:[UIBlurEffect effectWithStyle:style]];) setEffect native API internally take care of it and will not add any effect.
If we are building with iOS SDK < 13, in that case these constants will not be available and will give null because constants are wrapped inside macro. 

I have tested this.
So we do not have to add any check for that.

janvennemann: Alright, thanks for clearing this up. I saw the way the other properties are implemented and was wondering if this affects the new properties as well. If you say the check isn't needed then this looks good!

janvennemann (approved):

ssjsamir (approved):
**FR Passed** Able to see that new constants are exposed for blur styling effects. Tested with the test case from:https://jira.appcelerator.org/browse/TIMOB-27310

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 11 beta 5
Node.js ^8.11.1
iPhone 8 plus sim (13.0)
```

PR #11123, opened 2019-08-07
fix(buffer): constructor and from() compatibility

janvennemann:
Fixes a few issues i've found while working with our new `Buffer` implementation.

- Expose deprecated constructor signatures. Those are widely used in the ecosystem. Not having them implemented lead so undefined behavior and deadlocks in third-party modules.
- Add support for passing `Uint8Array` to `from()`

### DIFF
--- a/common/Resources/ti.internal/extensions/node/buffer.js
+++ b/common/Resources/ti.internal/extensions/node/buffer.js
@@ -14,6 +14,8 @@
  * a Uint8Array in any of our APIs that take a Ti.Buffer and eventually deprecating/removing Ti.Buffer.
  */
 
+import { isInsideNodeModules } from './internal/util';
+
 // https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings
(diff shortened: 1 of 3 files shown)


common/Resources/ti.internal/extensions/node/buffer.js
```diff
@@ -1347,6 +1376,19 @@ class Buffer {
 				value.copy(buffer, 0, 0, length);
 				return buffer;
 			}
+			if (value instanceof Uint8Array) {
```
sgtcoolguy: combine with line 1361 above (the Array case) basically just have the if check for `Array.isArray(value) || value instanced Uint8Array` as both cases iterate through and copy the bytes. The masking with 0xFF should basically be a no-op for a Uint8Array's values.

Then we can gain the fix for empty array you did here.

sgtcoolguy (approved):
LGTM, just a minor change to combine two `Buffer.from()` cases that are effectively the same

PR #11125, opened 2019-08-07
fix(ios): console.log does not log properly if it has multiple arguments

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27313

[console.log](https://docs.appcelerator.com/platform/latest/#!/api/Global.console) has different behavior from [Titanium.API.log](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.API-method-log)


iphone/TitaniumKit/TitaniumKit/Sources/API/TiConsole.h
```diff
@@ -14,6 +14,9 @@ JSExportAs(timeLog,
            -(void)timeLog
            : (id)args withData
            : (NSArray *)logData);
+JSExportAs(log,
```
sgtcoolguy: I'm not sure this is actually required as the log method id exported by the APIModule parent class...

sgtcoolguy (approved):
LGTM

ssjsamir (approved):
**FR Passed:** console.log now logs properly if it has multiple arguments.

**Tested with the following test case:** 

```
backgroundColor: '#fff'
});
const button = Ti.UI.createButton({
title: 'Log'
});
 
button.addEventListener('click', function(e){
console.log(1, ' test', true);
});
 
win.add(button);
win.open();

```
**Test Environment**
MacOS Mojave version 10.14.4
Xcode 10.3
Node.js ^8.12.0
iPhone 8 (13).
"NPM":"4.2.13","CLI":"7.0.11

PR #11133, opened 2019-08-12
fix(ios): properly inject CODE_SIGN_IDENTITY flag for Xcode 11 compatibility

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27330

Requires https://github.com/appcelerator/ioslib/pull/91 + package.json update to 1.7.11 once merged.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -3122,10 +3122,36 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 		};
 		xobjs.PBXShellScriptBuildPhase[buildPhaseUuid + '_comment'] = '"' + name + '"';
 	} else if (this.target === 'device') {
-		// sign the application using a signing identity that contains the phrase "iPhone Developer"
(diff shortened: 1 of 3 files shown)


hansemannn:
Using the old certificates fixed it for me. Although I'd still argue that we should properly assign certs directly instead of gambling that Xcode does it for us in the future. 

Also, still not supporting automatic code signing these days is really bad and will cause more and more issues in the future as developers move to generating their certs automatically.

cb1kenobi:
@hansemannn We used to explicitly set the certificate, but then I believe we went to the generic "iPhone Developer" when we added Apple Watch support. We mix the watch app targets into the Titanium app's Xcode project and wire up the target dependencies and I believe that there were issues signing the watch apps and the iOS app when using an explicit cert.

I'm not opposed to looking at it again, we just need to remember to test watch apps. :)

+1 for the "Apple Developer" cert support.

hansemannn:
Makes sense. Can you remove the milestone? Then we should be fine :)

PR #11135, opened 2019-08-12
fix(ios): properly fire close event for iOS 13+ modal windows

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27332

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiControllerProtocols.h
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiControllerProtocols.h
@@ -40,6 +40,7 @@
 - (void)viewWillDisappear:(BOOL)animated;
 - (void)viewDidAppear:(BOOL)animated;
 - (void)viewDidDisappear:(BOOL)animated;
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController;
 
 - (void)preferredContentSizeDidChangeForChildContentContainer:(id<UIContentContainer>)container;
 - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator;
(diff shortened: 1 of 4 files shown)


hansemannn:
It seems like this crashes with the test suite but I cannot access the logs. I thought that no  Xcode 11 guards are required since it's only a new delegate function which part of an existing delegate and won't be called on earlier versions, but it may does cause issues!

vijaysingh-axway (changes requested):
@hansemannn It is crashing when opening and closing window multiple times. See PR https://github.com/appcelerator/titanium_mobile/pull/11057. In this PR, this issue is already fixed.

hansemannn:
@vijaysingh-axway Is that really the same thing? Your PR is for configuring whether or not a modal window should be swipeable, this one is a bugfix for windows that don't trigger a `close` event, causing huge glitches in depending apps. I see you introduced a new flag for it, so if it works for this one as well, feel free to cherry-pick and integrate my changes into your PR.

vijaysingh-axway:
> @vijaysingh-axway Is that really the same thing? Your PR is for configuring whether or not a modal window should be swipeable, this one is a bugfix for windows that don't trigger a `close` event, causing huge glitches in depending apps. I see you introduced a new flag for it, so if it works for this one as well, feel free to cherry-pick and integrate my changes into your PR.

@hansemannn  What I mean was, In my PR https://github.com/appcelerator/titanium_mobile/pull/11057, 'close' event handling for iOS 13+ modal windows is already there . I have verified your test case. So we do not need this PR. It would be great, if you can test at your end. Thanks!

hansemannn:
I don't see how this solves my issue. I want the window to be swipeable and without any other code changes for backwards compatibility. Explicitly disallowing modal states is different I think. But I will test your changes!

hansemannn:
Not required anymore.

PR #11137, opened 2019-08-12
feat: add basic ssl/tls support in tcp socket

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27327

**Optional Description:**
This adds basic support for SSL/TLS enabled endpoints using our existing TCP socket. To keep things simple this will automatically start a SSL/TLS session after the socket is connected.

My use case is making HTTPS/WSS request for which this is enough. Other protocols may require additional setup before SSL/TLS can be applied which is not supported by this solution. Alternatively I thought about exposing the functionality as a method to give a user more control as combination of `useTls` property (to create the required `SSLSocket` on Android) and a `startTls/startHandshake` method.


android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
```diff
@@ -220,22 +233,46 @@ public void run()
...
+				boolean startTls = TiConvert.toBoolean(getProperty("startTls"), false);
+
+				if (startTls) {
+					SocketFactory sslSocketFactory = SSLSocketFactory.getDefault();
+					SSLSocket sslSocket = (SSLSocket) sslSocketFactory.createSocket();
+					sslSocket.setUseClientMode(true);
+					SSLParameters sslParameters = new SSLParameters();
+					List sniHostNames = new ArrayList(1);
+					sniHostNames.add(new SNIHostName(host));
+					sslParameters.setServerNames(sniHostNames);
```
jquick-axway: The `SNIHostName` class and `SSLParameters.setServerNames()` method are API Level 24 only (aka: Android 7.0).
https://developer.android.com/reference/javax/net/ssl/SNIHostName
https://developer.android.com/reference/javax/net/ssl/SSLParameters#setServerNames(java.util.List%3Cjavax.net.ssl.SNIServerName%3E)

This means this code will fail on Android OS versions older than 7.0. Our min supported API Level is 19 (aka: Android 4.4). So, this code will need to be rewritten to work on older OS versions. If we can't find an alternative solution, then we'll have to guard the code as follows...
```java
if (Build.VERSION.SDK_INT >= 24) {
	SSLParameters sslParameters = new SSLParameters();
	List sniHostNames = new ArrayList(1);
	sslParameters.setServerNames(sniHostNames);
	sslSocket.setSSLParameters(sslParameters);
}
```

I suggest testing this with an Android 4.4 device/emulator to verify that your code changes works.

jquick-axway (changes requested):

jquick-axway:
So, I'm not sure how I feel about this. Adding TLS support to a TCP object seems kind of backwards to me. It really works the other way around where TLS supports/uses TCP. Most frameworks would separate the two into different classes.

But I "get" why you want to make it more convenient to use. Especially since TLS is such a commonly used protocol on top of TCP. Hmm...

janvennemann:
addressed your feedback regarding the Android compatibility issues and changed the parameter name to `useTls`.

From a layer standpoint this sure is the wrong order approach this. However, considering our current socket implementation this is a way to do it without a total rewrite how we do sockets.

On iOS we use [CocoaAsyncSocket](https://github.com/robbiehanson/CocoaAsyncSocket) which only has a `startTls` method to handle SSL/TLS. Android uses `SSLSocket` which extends the default `Socket` so it can be used interchangeably in the current implementation. That's why i choose this approach to add basic SSL/TLS features to the socket.

android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
```diff
@@ -220,25 +241,55 @@ public void run()
...
-					int timeout = TiConvert.toInt(timeoutProperty, 0);
-
+				int port = TiConvert.toInt(getProperty("port"));
+				boolean useTls = TiConvert.toBoolean(getProperty("useTls"), false);
+
+				if (useTls) {
+					SocketFactory sslSocketFactory = new TiSocketFactory(null, null, NetworkModule.TLS_DEFAULT);
+					SSLSocket sslSocket = (SSLSocket) sslSocketFactory.createSocket();
+					sslSocket.setUseClientMode(true);
+					if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
jquick-axway: `KITKAT` is API Level 19. This should be `VERSION_CODES.N` for API Level 24.
Personally, I would use the literal value `24` since there would be no chance of confusion and saves time looking it up if you don't have this memorized. I'm not opposed to using Google's version constants, but I find using literal API Level values helps avoid mistakes.

The SNI code below is to check for man-in-the-middle attacks, right? So, it's okay to run without it on older OS versions. Our [appcelerator.https](https://github.com/appcelerator-modules/appcelerator.https) module can be used in conjunction with `HTTPClient` to check for man-in-the-middle attacks. Perhaps the solution in the future is to extend our `TCP` object to support our module's returned `SecurityManager` in the future.

janvennemann: Oh, damn! Must have looked up the wrong version code. I thought using version codes would be more expressive, but my own mistake just proved me wrong i guess.

The SNI code is to support servers that serve multiple domains via vhosts. With SNI the server will provide the correct certificate for the requested host. The alternative for older devices would be to provide a way to manually verify the certificate host name, see https://developer.android.com/training/articles/security-ssl#WarningsSslSocket. I've added the `validateServerIdentity` option that allows a user to manually verify the server's identity in such cases. It's modeled after node's [tls.checkServerIdentity](https://nodejs.org/api/tls.html#tls_tls_checkserveridentity_hostname_cert)

For more sophisticated checks or SSL pinning our https module might indeed come handy. But yeah, that's something for the future.

android/modules/network/src/java/ti/modules/titanium/network/socket/TCPProxy.java
```diff
@@ -220,22 +242,74 @@ public void run()
...
+						Log.d(TAG, "checkServerIdentity");
+						X500Principal peerPrincipal = (X500Principal)sslSession.getPeerPrincipal();
+						String canonicalName = peerPrincipal.getName(X500Principal.CANONICAL);
+						int separatorIndex = canonicalName.indexOf(",");
+						String commonName = canonicalName.substring(canonicalName.indexOf("cn=") + 3, separatorIndex != -1 ? separatorIndex : canonicalName.length());
+						HashMap<String, Object> certObject = new HashMap();
+						HashMap<String, String> subjectMap = new HashMap();
+						subjectMap.put("CN", commonName);
+						certObject.put("subject", subjectMap);
+						Object result = checkServerIdentity.call(getKrollObject(), new Object[] { host, certObject });
```
janvennemann: @jquick-axway There is some strange behavior going on here. If this callback returns an `Error` object, i get an empty `HashMap` as the result. Not sure if this is related due to this running on a separate thread. Only happens on Android, iOS works fine.

```js
checkServerIdentity(hostname, cert) {
  return new Error('failed'); // will result in an empty HashMap
  // return {message: 'failed'} // will result in a properly populated HashMap
}
```

jquick-axway: I think I know what's wrong. In V8, the JavaScript `Error` object's properties are not enumerable. I had to modify the `Error` prototype below so that `JSON.stringify()` could retrieve the non-enumerable properties.
https://github.com/appcelerator/titanium_mobile/blob/master/common/Resources/ti.internal/extensions/js/Error.js

janvennemann: Ah yeah that explains it. Thanks for all the details! I wanted to have the API close to what Node.js does hence using the error as return value. I'll try a few other variations and update the PR once i settled for one.

PR #11139, opened 2019-08-12
fix(android): splash open animation sometimes stutters on cold start

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27328

**Summary:**
- This is an issue on Google's end caused by opening another activity while the 1st one (the splash) is still going through its open animation on cold start. In mid-transition, the 1st activity restarts the animation using child activity's transition which is the cause of the "stutter".
- Solved by always overriding transitions when opening an activity from root activity (the splash). Changed to use a fade-in effect, which looks better than a slide-up animation from splash.
- Note that this issue was always happening on Android P and older OS versions before. First child window displayed from splash wrongly used fade-in effect instead of default slide-up if the 1st window was opened fast enough on startup.
- This is a much more noticeable issue in Android Q beta 4.

**Test:**
1. Acquire a Pixel 3 device/emulator with Android Q beta 4 or newer.
2. Download kitchensink-v2 project from [here](https://github.com/appcelerator/kitchensink-v2).
3. Build and run on the Pixel 3.
4. Verify that splash does not "stutter" as shown in GIF in [TIMOB-27328](https://jira.appcelerator.org/browse/TIMOB-27328).
5. To re-test, "force-quit" the app and relaunch.


garymathews (approved):
CR: **PASS**

ssjsamir (approved):
**FR Passed**: No longer able to see the stutter on the splash screen using the test steps above. 

**Test environment**
```
MacOS Mojave version 10.14.4
Node.js: 8.11.3
Pixl 3 Sim (Android Q)
Java Version: 1.8.0_131
NDK: 12.1.2977051
```

PR #11149, opened 2019-08-15
fix(android): guard for tab counts limit for bottom style

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27302

**Description:**
Add a simple guard for adding more windows than the tabs controller support.
Would it make sense to add a unit test for trying to `setActiveTab()` a tab that is not added to the controller? This is the closest scenario I could think of to swiping out of bounds.

**Test case:**
There is a good code sample in the ticket.


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIAbstractTabGroup.java
```diff
@@ -126,7 +126,7 @@
 	public abstract String getTabTitle(int index);
 
 	// region protected fields
-	protected static final String TAG = "TiUITabLayoutTabGroup";
+	protected static String TAG = "";
```
jquick-axway: I don't like us having a settable static TAG variable here. If one TabGroup style ever happened to launch another TabGroup child activity, then the last one would clobber first's TAG name.

Personally, I'm okay with `TAG` being set to "TiUIAbstractTabGroup" from a traceability standpoint. We always set TAG to the Java class name and not the JS type name anyways.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIAbstractTabGroup.java
```diff
@@ -135,6 +135,9 @@
 	protected boolean tabsDisabled = false;
 	protected int numTabsWhenDisabled;
 	protected int colorPrimaryInt;
+	// Max number of tabs supported. Currently in use for BottomNavigation style only.
+	// Default value of -1 is used to represent no tab limit.
+	protected int maxTabsSupported = -1;
```
jquick-axway: Instead of a member variable, how about we have a `getMaxTabsSupported()` method instead. By default, it can return `Intent.MAX_VALUE` like so. Setting it to max int means we can easily do greater-than checks.
```java
public int getMaxTabsSupported()
{
	return Integer.MAX_VALUE;
}
```

Then "TiUIBottomNavigationTabGroup" can then override it like this...
```java
@Override
public int getMaxTabsSupported()
{
	return 5;
}
```

Ideally, we should call [BottomNavigationView.getMaxItemCount()](https://developer.android.com/reference/android/support/design/widget/BottomNavigationView#getmaxitemcount) to get the maximum, but unfortunately it is an instance method, not a static method.

jquick-axway: Hold on. Maybe we don't need a `maxTabsSupported` member variable or `getMaxTabsSupported()` method.

A simpler solution might be to have `TiUIBottomNavigationTabGroup` override the `addTab()` method like this?
```java
@Override
public void addTab(TabProxy tabProxy)
{
	if (this.mBottomNavigationView == null) {
		return;
	}

	final int MAX_TABS = this.mBottomNavigationView.getMaxItemCount();
	if (this.tabs.size() < MAX_TABS) {
		super.addTab(tabProxy);
	} else {
		Log.w(TAG, "Bottom style TabGroup cannot have more than " + MAX_TABS + " tabs.");
	}
}
```

jquick-axway (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIAbstractTabGroup.java
ypbnv: I have totally missed the `getMaxItemCount()` method for the bottom navigation view. Moved to using it and override the `addTab()` method. It is a far better and clean solution. Thanks!

jquick-axway (approved):
CR: Pass

sgtcoolguy:
The 8_3_X version of this PR was FR'd and passed by @keerthi1032 
I've closed it, though as master is the new 8.3.0 target so this should suffice for landing on 8.3.0.

PR #11154, opened 2019-08-15
chore(android)(8_3_X): compile using SDK 29

garymathews:
- Update `android/package.json` to compile using API 29
- Update Android Studio project files to target Android Q
- Update `TitaniumTest` project to target Android Q

[TIMOB-26950](https://jira.appcelerator.org/browse/TIMOB-26950)

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -23,13 +23,13 @@
 		"checksum": "7798aae19b2efff53b1f18d53345291ab16a574e0524350eed3db6d45757bc476a158303b8df6c5f359439f9184818d61eba6bd820cec404e268ab0796cc7c5a"
 	},
 	"minSDKVersion": "19",
-	"compileSDKVersion": "28",
+	"compileSDKVersion": "29",
 	"vendorDependencies": {
 		"android sdk": ">=23.x <=28.x",
-		"android build tools": ">=25.x <=28.x",
-		"android platform tools": "28.x",
+		"android build tools": ">=25.x <=29.x",
+		"android platform tools": "29.x",
 		"android tools": "<=26.x",
-		"android ndk": ">=r11c <=r17b",	
+		"android ndk": ">=r11c <=r20",	
 		"java": ">=1.8.x"
 	},
 	"repository": {
--- a/android/.idea/misc.xml
+++ b/android/.idea/misc.xml
@@ -5,26 +5,37 @@
     <option name="myDefaultNotNull" value="android.support.annotation.NonNull" />
     <option name="myNullables">
       <value>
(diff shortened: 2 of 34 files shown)


garymathews:
@sgtcoolguy We need to update Jenkins to support this

android/package.json
```diff
@@ -23,13 +23,13 @@
 		"checksum": "7798aae19b2efff53b1f18d53345291ab16a574e0524350eed3db6d45757bc476a158303b8df6c5f359439f9184818d61eba6bd820cec404e268ab0796cc7c5a"
 	},
 	"minSDKVersion": "19",
-	"compileSDKVersion": "28",
+	"compileSDKVersion": "29",
 	"vendorDependencies": {
-		"android sdk": ">=23.x <=28.x",
-		"android build tools": ">=25.x <=28.x",
-		"android platform tools": "28.x",
+		"android sdk": ">=23.x <=29.x",
```
jquick-axway: We don't want to change "android sdk" to "29.x" yet... because it'll make our build system target Android Q by default. I've made a separate ticket to do this in [TIMOB-26951](https://jira.appcelerator.org/browse/TIMOB-26951). Can you change it back please?
```
"android sdk": ">=23.x <=28.x",
```

jquick-axway (changes requested):
We need to update the `DEFAULT_API_LEVEL` constant to `29` here...
https://github.com/appcelerator/titanium_mobile/blob/8_3_X/build/lib/android/androidsdk.js

We should also update the following "AndroidManifest.xml" files to target 29 as well...
https://github.com/appcelerator/titanium_mobile/blob/8_3_X/android/modules/android/AndroidManifest.xml
https://github.com/appcelerator/titanium_mobile/blob/8_3_X/android/modules/ui/AndroidManifest.xml
https://github.com/appcelerator/titanium_mobile/blob/8_3_X/android/titanium/AndroidManifest.xml

android/package.json
garymathews: This was intentional, wouldn't we want to target by default to see what potential issues we run into? I also wanted to see what tests would fail, but Jenkins needs updating.

jquick-axway (approved):
CR: Pass

sgtcoolguy:
Android 29 (as well as NDK r20, and android 29 emulator using google play services image) is currently rolling out to all our android sdk/ndk build nodes: https://jenkins.appcelerator.org/job/jenkins/job/jenkins/job/master/158/

This PR should be good to rebuild soon...

garymathews:
Yay, it's passing!

PR #11160, opened 2019-08-20
fix: exclude all frameworks from optimizing files step (8_3_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27097

8_3_X backport of #10905


janvennemann:
Closing as this fix has already landed in 8_3_X.

PR #11166, opened 2019-08-22
feat(android): add support for TabGroup in NavigationWindow

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27316

**Description:**
Allows usage of a TabGroup as a root or a child view in NavigationWindow. As of now opening a TabGroup as a child is done through a new method `openTabGroup()`. I did it that way in order to not mess up with the currently available `openWindow()`. If that does not seem a good solution I can go for a change in the method parameter type and reuse what we currently provide. Let me know what do you think about that.
I will edit the docs once we decide if a new API is necessary.

Edit: I did not think about the iOS side. It would be a parity problem with a new method...

**Test case:**
There are two JS files in the JIRA ticket providing a good code sample. If we stick with the new method they would need to have `openWindow()` replaced by `openTabGroup`.


jquick-axway:
@ypbnv, we don't want to add a new `NavigationWindow.openTabGroup()` method. We want to use the existing `NavigationWindow.openWindow()` method which supports open both `Window` and `TabGroup` on iOS. Note that in our JavaScript API docs, our `TabGroup` type derives from `Window` making this okay from an API standpoint. (The point of this ticket was to achieve better parity with iOS.)

I think we should go ahead and change our Java `NavigationWindowProxy.openWindow()` method. We can change the 1st argument from `WindowProxy` to `Object` and do an `instanceof` check for `WindowProxy` and `TabGroupProxy`.

ypbnv:
Reworked the PR in order for the support to be added through the existing `openWidnow` method.

android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
```diff
@@ -47,7 +50,7 @@ public void popToRootWindow(@Kroll.argument(optional = true) Object arg)
 	{
 		// Keep first "root" window
 		for (int i = windows.size() - 1; i > 0; i--) {
-			WindowProxy window = windows.get(i);
+			WindowProxy window = ((WindowProxy) windows.get(i));
```
jquick-axway: Casting it to `WindowProxy` will crash if it's of type `TabGroup`. Since you've changed the `windows` member variable to be a collection of `TiWindowProxy` types, you can get rid of the cast and close the `TiWindowProxy` object.

You will also need to change the `NavigationWindowProxy.closeWindow()` method so that it can close a tab group. Perhaps change the `closeWindow()` argument from type `WindowProxy` to `Object` and then do an `instanceof TiWindowProxy` like how we're doing it in the `openWindow()` method.

android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
```diff
@@ -59,21 +62,32 @@ public void close(@Kroll.argument(optional = true) Object arg)
 	// clang-format on
 	{
 		popToRootWindow(arg);
-		closeWindow(windows.get(0), arg); // close the root window
+		closeWindow(((WindowProxy) windows.get(0)), arg); // close the root window
```
jquick-axway: Same issue mentioned above.
This will crash if `windows` member variable returns a `TabGroupProxy` object.

tests/Resources/ti.ui.navigationwindow.addontest.js
```diff
@@ -0,0 +1,49 @@
...
+			navigationWindow = Ti.UI.createNavigationWindow({
+				window: rootWin
+			}),
+			tabGroup = Ti.UI.createTabGroup({ title: 'TabGroup',
+				tabs: [ createTab('Tab 1'),
+					createTab('Tab 2'),
+					createTab('Tab 3') ]
+			});
+		navigationWindow.open();
+		navigationWindow.openTabGroup(tabGroup);
```
jquick-axway: We need to replace the `openTabGroup()` method with `openWindow()`.

jquick-axway (changes requested):

ypbnv:
@jquick-axway Updated.

android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
```diff
@@ -47,8 +50,8 @@ public void popToRootWindow(@Kroll.argument(optional = true) Object arg)
 	{
 		// Keep first "root" window
 		for (int i = windows.size() - 1; i > 0; i--) {
-			WindowProxy window = windows.get(i);
-			closeWindow(window, arg);
+			TiWindowProxy window = windows.get(i);
+			window.close(arg);
```
jquick-axway: There is a bug here.

You're not removing any of the objects from the `windows` member variables. Versus the old code would call `closeWindow()` which would call `windows.remove()`.

ypbnv: My bad. I did not see that. Updated.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
@ypbnv , I dont see the tab icons when a particular tab is selected. I am not sure if it gets hidden of it turns dark. Also, it does not show any indication on the tabs for which tab is selected.  
Screen Shot 2019-10-29 at 10 38 13 AM

jquick-axway:
@lokeshchdhry, this is actually the correct "default" behavior for bottom tabs. What's happening is the selected tab's icon is tinted dark. Since the app uses a dark theme by default, you can't see the selected tab's icon anymore.

If you changed the tab color to some other color (say "blue") then you'll be able to see the icon.
```javascript
function createTab(title) {
	var window = Ti.UI.createWindow({ title: title });
	window.add(Ti.UI.createLabel({ text: title + " View" }));
	var tab = Ti.UI.createTab({
		title: title,
		icon: "/assets/images/tab1.png",
		window: window,
	});
	return tab;
}

var tabGroupSettings = {
	tabs: [createTab("Tab 1"), createTab("Tab 2"), createTab("Tab 3")],
	tabsBackgroundColor: "blue",
	tabsBackgroundSelectedColor: "blue",
};
if (Ti.Platform.name === "android") {
	tabGroupSettings.shiftMode = false;
	tabGroupSettings.style = Ti.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION;
}
var tabGroup = Ti.UI.createTabGroup(tabGroupSettings);
tabGroup.open();
```

ssekhri:
FR Passed. TabGroup works fine in NavigationWindow on Android.
Verified on:
Mac OS: 10.14.5
SDK: 8.3.0.v20191029100019
Appc CLI: 7.1.1
JDK: 1.8.0_162
Node: 10.5.0
Studio: 5.1.4.201909061933
Device: Nexus4(v5.1.1) device, Pixel3(v10.0) emulator

PR #11172, opened 2019-08-23
style: replace grunt-appc-js with eslint

janvennemann:
We currently have two different linting setups active in this repo. Running `npm run test` will use the outdated `grunt-appc-js` task to run linting, while the git pre-commit hook is set to just run `eslint`. This caused unexpceted results for me numerous times while editing files and different linting rules were required to pass the git pre-commit hook and `npm run test`.

This will align the linting setup to only use `eslint` by replacing `grunt-appc-js` with `grunt-eslint` and applying the necessary changes to some files.


sgtcoolguy (approved):
LGTM. I'd add that way way back on my TODO list is to get rid of grunt altogether. I think we could feasibly just move over to using npm scripts to do what it's providing now.

PR #11176, opened 2019-08-26
fix(android): fix onlink callback being cretion only

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27238

**Description:**
Fix `onlink` callback property being creation-only.

**Test case:**

```js
var htmlText =
'<!DOCTYPE html>\n' +
'<html>\n' +
'	<head>\n' +
'	</head>\n' +
'	<body>\n' +
'		<p>WebView "onlink" Test</p>\n' +
'		<br/>\n' +
'		<br/>\n' +
'		<a href="mylink://show/alert">Show Alert</a>\n' +
'	</body>\n' +
'</html>\n';
 
function onLinkHandler(e) {
if (e.url === "mylink://show/alert") {
alert("'onlink' callback invoked.");
return false;
}
return true;
}
 
html: htmlText,
//	onlink: onLinkHandler,  // <- Must be set upon creation to work.
});
webView.onlink = onLinkHandler;
window.add(webView);
window.open();
```


garymathews (approved):
CR: **PASS**

PR #11177, opened 2019-08-26
fix(android): fix onlink callback being cretion only

ypbnv:
(cherry picked from commit 1f51e13d2a9947ac2106de0bd4589b556fc3da38)

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27238

**Description:**
Fix `onlink` callback property being creation-only.

**Test case:**

```js
var htmlText =
'<!DOCTYPE html>\n' +
'<html>\n' +
'	<head>\n' +
'	</head>\n' +
'	<body>\n' +
'		<p>WebView "onlink" Test</p>\n' +
'		<br/>\n' +
'		<br/>\n' +
'		<a href="mylink://show/alert">Show Alert</a>\n' +
'	</body>\n' +
'</html>\n';
 
function onLinkHandler(e) {
if (e.url === "mylink://show/alert") {
alert("'onlink' callback invoked.");
return false;
}
return true;
}
 
html: htmlText,
//	onlink: onLinkHandler,  // <- Must be set upon creation to work.
});
webView.onlink = onLinkHandler;
window.add(webView);
window.open();
```


garymathews (approved):
CR: **PASS**

sgtcoolguy:
@ypbnv I think we should add a simple unit test here to create a web view and set the online property after creation and verify it ends up getting called.

sgtcoolguy (dismissed):
I think we should add a simple unit test here to create a web view and set the onlink property after creation and verify it ends up getting called.

jquick-axway:
@sgtcoolguy, @ypbnv,

This is in regards to unit testing. I remember looking into this about 8 years ago. On Android, the native listener only works for physical taps on links. That listener won't be invoked when attempting to navigate via JavaScript, such as by setting the `window.location` property. But that said, I do remember iOS' old `UIWebView` would invoke its "onlink" equivalent when navigating via JavaScript making it easy to unit test, but I don't know if the same is true with iOS' `WKWebView`.

I believe our Appium test suite can simulate physical touches. Perhaps this is the way to go for automated testing?

ssjsamir (approved):
FR Passed: Able to see the "onlink" callback being invoked. Tested with the test case above.

**Test Environment**
```
MacOS Catalina 10.15 Beta
Node.js ^8.11.1
Android Pixel XL 7.1 Emulator 
"NPM":"4.2.15","CLI":"7.1.1"
```

PR #11184, opened 2019-08-28
fix(ios): set navigation controller’s background color

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27263


iphone/Classes/TiUINavigationWindowProxy.m
```diff
@@ -220,6 +220,7 @@ - (void)navigationController:(UINavigationController *)navigationController will
     [theWindow windowWillOpen];
     [theWindow windowDidOpen];
   }
+  navController.view.backgroundColor = theWindow.view.backgroundColor;
```
hansemannn: It feels like a workaround, since thats not necessary natively. Can you please go through the view tree and check the differences between native views / VC's?

lokeshchdhry:
FR Passed for this part of the fix. 
Likely more fixes will be added for this issue in additional PR's.

Studio Ver:       5.1.3.201907112159
SDK Ver:          8.2.0.v20190903100931
OS Ver:           10.14.5
Xcode Ver:        Xcode 11.0
Appc NPM:         4.2.15-1
Appc CLI:         7.1.1-master.6
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.1
Alloy Ver:        1.14.1
Node Ver:         10.16.1
NPM Ver:          6.9.0
Java Ver:         10.0.2
Simulator: Iphone XR ios 13, iphone XR ios 12.4

iphone/TitaniumKit/TitaniumKit/Sources/Modules/TiUIWindowProxy.m
```diff
@@ -268,6 +268,27 @@ - (void)preferredContentSizeDidChangeForChildContentContainer:(id<UIContentConta
...
+        appearance.backgroundColor = self.view.backgroundColor;
+      } else {
+        appearance.backgroundColor = newColor.color;
+      }
+      controller.navigationController.navigationBar.standardAppearance = appearance;
+      controller.navigationController.navigationBar.scrollEdgeAppearance = appearance;
+      controller.navigationController.navigationBar.backgroundColor = UIColor.clearColor;
+    }
+  }
+#endif
```
hansemannn: This looks very very hacky and I feel like it's not fixing the actual issue but only workarounding the visual glitches. I hope there can be found a proper fix for this issue.

hansemannn: In addition, the navigation bar now shows a different color than the background color of the window.

vijaysingh-axway: @hansemannn Yeah. It's kind of hack. In native apps as well people suggest same. See [here](https://stackoverflow.com/questions/56556254/in-ios13-the-status-bar-background-colour-is-different-from-the-navigation-bar-i). And this will work without any visual issue.

In context of property 'largeTitleEnabled' is  'true' in iOS 13-
A). As for as navigation bar color and transition similar to native behaviour is concerned. For that above SDK changes are not required. To make it work by default we have to make changes as per point C. If you want this to work in your app now, you have to modify your code-
1. You have to set window's property 'extendEdges' to [Ti.UI.EXTEND_EDGE_TOP]. By default it is Ti.UI.EXTEND_EDGE_NONE.
2. Set 'translucent' property to 'true'. By default it is 'true' 
3. If you are using tableview, listview or scrollview (basically view’s having scroll view) on window, it will work perfect.
4. If you are using any other view other than views in point 3, you have to set ‘top’ property of view to height of navigation bar. Otherwise it will move up and overlap with navigation bar.

B). As you have mentioned navigation bar color is different than the bar color. I guess you are using "barColor" property to set. In SDK code barColor has preference over backgroundColor of window. See the attached examples in ticket, it is working fine. You always have option to set 'barColor'.

C). In native development, property 'edgesForExtendedLayout' is by default UIRectEdgeAll and default value of 'translucent' is 'true'. So it takes the background color of controller's view. If you use tableview or scrollview it will work perfect. If you want to create any other view with position (0, 0, width, height). It'll overlap with navigation bar. To avoid this problem there is option to use autolayout constraints. 
 In titanium auto layout is in beta. So we can not use that.
I'll create a new ticket to change default behaviour of  window's property extendEdges to Ti.UI.EXTEND_EDGE_ALL, which apple recommends. And probably need to handle window's layout if there is navigation bar(which will need more work).

PR #11188, opened 2019-08-29
feat(ios): support generic apple certificates

ewanharris:
If passed in the name/team ID variant, we make a best attempt to map it back to the fullname variant of cert type/name/team ID. It's possible that we cant figure out the correct name (in the case where the name matches both the iOS cert and Apple cert), in which case we will error and prompt for the correct value.

Closes TIMOB-27358

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27358

Requires appcelerator/ioslib/pull/93

I need to fully verify that this doesn't break any of the signing of extensions.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -223,7 +223,7 @@ iOSBuilder.prototype.findCertificates = function findCertificates(name, type) {
 			for (const scope of types) {
 				if (scopes[scope]) {
 					for (const cert of scopes[scope]) {
-						if (cert.name === name) {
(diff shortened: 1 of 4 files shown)


cb1kenobi (approved):

ssjsamir (approved):
**FR Passed**, able to use generic apple certificates when building to device and when distributing to the app store. 

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 11 beta 5
Node.js ^8.11.1
iPhone 8 plus sim (13.0)
```

PR #11190, opened 2019-08-30
fix(ios): remove old UIWebView left overs

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27367

Suggesting to merge at the earliest possible time so users won't be affected by this when it becomes serious.


janvennemann (approved):

hansemannn:
@janvennemann @sgtcoolguy Please allow me to argue that this change does not require unit tests and should therefore receive the `no tests` label :P. Also, QE will likely not have much to test as the only change was a removal of an API that can't be used anymore - even before this change.

janvennemann:
Agreed on the no tests label! The only thing QE could test is to submit an app to make sure the deprecation note is gone. cc @lokeshchdhry

ssjsamir (approved):
FR passed: Not getting any deprecation errors when building to the app store. 

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 10.2.1
Xcode 11 beta 5
Node.js ^8.11.1
```

PR #11195, opened 2019-09-04
impr(android): implement TLSv3 support

garymathews:
- Implement `TLSv3` support into `Titanium.Network`

##### TEST CASE
```JS
const lbl = Ti.UI.createLabel({ text: 'Loading...' });

const client = Ti.Network.createHTTPClient({
    onload: e => {
        const html = e.source.responseText;

        if (html.includes('protocol_tls1_3">Yes')) {
            lbl.text = 'TLSv3: Supported';
            lbl.color = 'green';
        } else {
            lbl.text = 'TLSv3: Unsupported';
            lbl.color = 'red';
        }
    },
    onerror: e => {
        lbl.text = 'Could not determine TLSv3 support!';
        lbl.color = 'red';
    },
    timeout: 8000
});
client.open('GET', 'https://ssllabs.com/ssltest/viewMyClient.html');
client.send();

win.add(lbl);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26956)


sgtcoolguy (changes requested):
Please add a unit test that automates the exact test case you pasted in manually.

jquick-axway:
@garymathews, maybe we should add the `TLS_VERSION_1_3` constant to our iOS code in this PR as well. It turns out our iOS code never uses these constants; it only defines them.
https://github.com/appcelerator/titanium_mobile/blob/16f04c5a2437a948b6cbd3895421035a4c180983/iphone/Classes/NetworkModule.h#L19
https://github.com/appcelerator/titanium_mobile/blob/16f04c5a2437a948b6cbd3895421035a4c180983/iphone/Classes/NetworkModule.m#L196

The rest of your changes look fine. Thanks!

jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
The response text for httpclient request to a TLS test page returns success for TLS1.3 on Android Q  emulator. For older Android versions the TLS1.3 support is returned as false.
Also when tlsVersion is set to 1.2 for httpclient then the response text returns false on Android Q emulator as well.
Verified on:
Mac OS 10.14.6
Ti SDK: 8.3.0.v20190913134535
Appc CLI: 7.1.1
Node: 10.5.0
JDK: 10.0.2
Studio: 5.1.4.201909061933
Device: Pixel2 emulator (10.0), Pixel2 emulator (8.0.0)

PR #11198, opened 2019-09-05
fix(ios): set CODE_SIGN_IDENTITY to the cert name

ewanharris:
It seems to be that if you have an Apple cert and an iPhone cert with the same Team ID, xcode will prefer to use the Apple cert due to the ambigious certificate name we passed in, this would then error out as the provisioning profile we passed in would not match the cert xcode selected (incorrectly) for us

Fixes TIMOB-27358

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27358

I have tested this with a watch extension to dist-adhoc/dist-appstore builds and the codesigning is successful, the adhoc build can be installed to device, and xcode verifies the appstore build just fine. I would like to test some other extensions also.

Update:

Tested with signing a StickerPack and a Today extension and it signs correctly also

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -3123,7 +3123,7 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 
 	// add the post-compile build phase for dist-appstore builds
 	if (this.target === 'dist-appstore' || this.target === 'dist-adhoc') {
-		buildSettings.CODE_SIGN_IDENTITY = '"iPhone Distribution"';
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
LGTM!

ssjsamir (approved):
**FR Passed**, able to package using the correct distribution certificates. 

**Test Environment**
```
MacOS Mojave version 10.14.4
Xcode 11 beta 5
Node.js ^8.11.1
iPhone 8 plus sim (13.0)
"NPM":"4.2.15-1","CLI":"7.1.1-master.6"
```

PR #11200, opened 2019-09-06
fix(ios): update tabbar after trait colection changed

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27126


keerthi1032:
FR passed. tab group image changes concurrently  when mode changes on iOS 13.

Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.2.0.v20190906123526
Device =iPhone XR iOS 13
Simulator -iPhone 8 iOS 13

iphone/Classes/TiUITabProxy.m
```diff
@@ -54,9 +56,18 @@ - (void)_configure
...
+  [[NSNotificationCenter defaultCenter] addObserver:self
+                                           selector:@selector(didChangeTraitCollection:)
+                                               name:kTiTraitCollectionChanged
+                                             object:nil];
   [super _configure];
 }
 
+- (void)didChangeTraitCollection:(NSNotification *)info
+{
+  [self updateTabBarItem];
```
hansemannn: Shouldn't the UIImage instances use semantic images already and check automatically?

vijaysingh-axway: There seems a bug from apple side. See https://stackoverflow.com/questions/57011346/uitabbar-uitabbaritem-and-ios-13s-dark-mode

hansemannn: Okay. But `traitCollectionDidChange:` fires for iOS 8+ and also if the device is rotated or font scaling is updated. It's minor but should be guarded in a later release to prevent unnecessary UI work.

PR #11208, opened 2019-09-10
build(docgen): fixes docgen module usage failing on Windows platform

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27388

**Description:**
Reworks the calling of titanium-docgen generation and validation due to child process spawning being
broken on Windows. It ignores the shebang script and fails to initialize the called module. So instead we are manually calling the generated binary for Windows.

Thanks, @ewanharris and @janvennemann, for helping me deal with this!


ewanharris (approved):
Looks good to me!

PR #11212, opened 2019-09-11
fix(ios): include new iphone models into os extension

garymathews:
- Include new iPhone models into `os` extension
- Prevent `cpuModelAndSpeed` from failing
- Fixes ` ios.os.#cpus() returns array of objects whose length matches Ti.Platform.processorCount` test case

##### TEST CASE
```JS
const os = require('os');
console.log(JSON.stringify(os.cpus(), 1));

```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27395)


sgtcoolguy (approved):
LGTM, Thanks Gary!

PR #11213, opened 2019-09-12
fix(ios): inverted condition on baseURL construction

slash-84:
# In this PR

This PR will fix `baseURL` usage on iOS that is resolved in a `window.location` with `file://` instead of `http://` or `https://`.

# Bug description

## Code to reproduce

```
$.webview.setHtml('<html><body><script>alert(window.location);</script></body></html>', {
    baseURL: 'https://www.google.com/'
});
```

## What happens

The `alert(window.location)` outputs `https://www.google.com/` on Android, while on iOS the output is `file:///https://www.google.com/`.

## Notes

I found this strange behavior because I'm trying to insert 3rd party embeds, often without success because they internally use relative protocol URLs that are eventually resolved as `file://` instead of `https://`.

This was expected on `7.X`, because of the absence of `baseURL` param, but on `8.X` I was expected that baseURL would reflect on window.location (like it happens on Android).

[Jira ticket](https://jira.appcelerator.org/browse/TIMOB-27396)


janvennemann (approved):
Looks good to me! Thanks for the fix!

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.14.5
SDK: 8.3.0.v20191009084313
Appc CLI: 7.1.1
JDK: 1.8.0_162
Node: 10.5.0
Studio: 5.1.4.201909061933
Xcode: 11.0
Device: iOS simulator v12.2, 13.0

PR #11214, opened 2019-09-13
fix(android): exclude JS in HTML files from processing

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27398

**Optional Description:**
Prevent processing of JS files that are referenced in HTML during incremental builds and make sure that those JS file will not be removed by the automatic builder resource cleanup.

This is an Android only issue due to the way the `jsFiles` list is handled in the builder. The iOS builder actually removes entries from JS files that are used in HTML so the issue only occurred on Android. Switching to the pre-filtered `inputFiles` of the task solved the issue.

The update of `appcd-tasks` is for better error handling since it didn't properly pass the original error in some cases. This made it harder to debug this issue in the first place. With the update the a task failure will now pass down the original error.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2722,6 +2722,7 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
 						const from = jsFiles[relPath];
 						const to = path.join(this.buildBinAssetsResourcesDir, relPath);
 						copyFile.call(this, from, to, next);
+						this.unmarkBuildDirFile(to);
 					};
 				}), done);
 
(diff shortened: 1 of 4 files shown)


cb1kenobi (approved):
CR'd and FR'd. Looks good. **APPROVED**

keerthi1032:
FR passed. JS files are included when building. 
**Test Environment:**
  Version                     = 10.14.6
Node.js
Titanium CLI
Titanium SDK
  SDK Version                 = local sdk 8.2.0.v20190913101659 and local 8.3.0.v20190913100522
Device -Oneplus 5t android 9
Emulator -pixel android 9

PR #10045, opened 2018-05-14
[TIMOB-25503] Android: Fix CardView template compatibility

garymathews:
- Fix ability to use `Ti.UI.Android.CardView` in `Ti.UI.ListView` templates

##### TEST CASE
```JS
var win = Ti.UI.createWindow({
        backgroundColor: 'gray'
    }),
    listView = Ti.UI.createListView({
        templates: {
            test: {
                childTemplates: [{
                    type: 'Ti.UI.Android.CardView',
                    childTemplates: [{
                        type: 'Ti.UI.Label',
                        bindId: 'label',
                        properties: {
                            color: 'black',
                            bindId: 'label'
                        }
                    }],
                    properties: {
                        width: Ti.UI.FILL,
                        height: Ti.UI.SIZE,
                        cardUseCompatPadding: true,
                        backgroundColor: 'white',
                        layout: 'vertical'
                    }
                }]
            }
        },
        defaultItemTemplate: 'test'
    }),
    section = Ti.UI.createListSection(),
    items = [];

['A', 'B', 'C'].forEach((item) => items.push({
    label: { text: item },
    template: 'test'
}));

section.setItems(items);
listView.setSections([ section ]);
win.add(listView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25503)

PR #11219, opened 2019-09-16
fix(android): fix gradient background support for CardView

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27258

**Description:**
Fix support of Gradient background for Ti.UI.Android.CardView. Refactor the assigning of nativeView for TiUICardView in order to avoid duplicated execution of `processProperties()`. Add some "removed" properties in the docs and remove the color/image states properties. They are currently bugged, but I would not want to expand the scope of this PR.
Note: No unit tests, since this is a bug that is only visual. All the View's events are triggered but it remains invisible.

This PR's test case would be good to be ran again, since there are changes in the code it added:
https://github.com/appcelerator/titanium_mobile/pull/10045 

**Test case:**
_app.js_
```js
var win = Ti.UI.createWindow({layout: 'vertical'}),
	cardView1 = Ti.UI.Android.createCardView({
		width: 200,
		height: 200,
			backgroundGradient: {
			type: 'linear',
			colors: ['red', 'green'],
			startPoint: { x: '50%', y: 0 },
			endPoint: { x: '50%', y: '100%' },
			backFillStart: false
		},
	}),
	cardView2 = Ti.UI.Android.createCardView({
		top: 50,
		width: 200,
		height: 200,
		backgroundColor: 'red'
	});

cardView1.addEventListener('click', function() {
	cardView1.backgroundGradient = {
		type: 'linear',
		colors: ['yellow', 'magenta'],
		startPoint: { x: '50%', y: 0 },
		endPoint: { x: '50%', y: '100%' },
		backFillStart: false
	}
});

cardView2.addEventListener('click', function() {
	cardView2.backgroundColor = 'blue';
});

win.add([cardView1, cardView2]);
win.open();

```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
@@ -6,6 +6,12 @@
  */
 package ti.modules.titanium.ui.widget;
 
+import android.content.Context;
+import android.support.v7.widget.CardView;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.FrameLayout;
+
 import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.common.Log;
@@ -14,20 +20,16 @@ import org.appcelerator.titanium.TiDimension;
 import org.appcelerator.titanium.proxy.TiViewProxy;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.view.TiCompositeLayout;
-import org.appcelerator.titanium.view.TiUIView;
 import org.appcelerator.titanium.view.TiCompositeLayout.LayoutArrangement;
-
-import android.content.Context;
-import android.support.v7.widget.CardView;
-import android.view.View;
-import android.view.ViewGroup;
-import android.widget.FrameLayout;
+import org.appcelerator.titanium.view.TiGradientDrawable;
+import org.appcelerator.titanium.view.TiUIView;
 
 public class TiUICardView extends TiUIView
 {
 	public int paddingLeft, paddingTop, paddingRight, paddingBottom;
 
 	private static final String TAG = "TiUICardView";
+	private final TiCardView cardView;
 
 	public class TiUICardViewLayout extends TiCompositeLayout
 	{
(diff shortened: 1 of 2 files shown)


garymathews (changes requested):
You also need to update the statements in:
```Java
@Override
public void propertyChanged(String key, Object oldValue, Object newValue, KrollProxy proxy) {
...
```
To use `this.cardView`, add `PROPERTY_BACKGROUND_GRADIENT` handling and remove properties handled by the parent class.

ypbnv:
Updated with the changes in `propertyChanged()`. Also extracted setting the gradient background and extended the test case.

hansemannn:
@ypbnv Unfortunately, with this patch the `borderRadius` is not respected anymore.  Everything else seems to work fine. Bonus: A ripple effect would be good to work, but thats okay not to.

ypbnv:
Gradient background does not work together with CardView's border radius. From what I have found this is a problem in native Android as well, unfortunately. Setting any drawable as a background besides the internal RoundRectDrawable will break the corner radius of the native view.

Setting the PR as work in progress because I am still trying to somehow work around this limitation.

As a workaround for now I would suggest creating a "container" view with the desired background gradient and width and height set to Ti.UI.FILL. Putting it as the first child in a CardView and adding any other content to it. Not the best solution performance-wise, but it would allow usage of gradient and border radius.

hansemannn:
It works good! One note: If you also pass a `backgroundColor`, it will not work.

hansemannn:
Unfortunately it broke the `backgroundColor` usage completely:
```js

var cardView = Ti.UI.Android.createCardView({
	width: 200,
	height: 200,
	borderRadius: 100,
	backgroundColor: '#333'
});

win.add(cardView);
win.open();
```
This should render a circle, but it renders a square, since the `borderRadius` is ignored when combined with a `backgroundColor`.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
```diff
@@ -137,81 +151,65 @@ public void processProperties(KrollDict d)
...
-		}
-
-		view = new TiCardView(getProxy().getActivity(), arrangement);
-		view.setPadding(0, 0, 0, 0);
-		view.setFocusable(false);
-		TiCardView cardview = (TiCardView) view;
-
-		if (d.containsKey(TiC.PROPERTY_BACKGROUND_COLOR)) {
-			cardview.setCardBackgroundColor(TiConvert.toColor(d, TiC.PROPERTY_BACKGROUND_COLOR));
+		if (d.containsKeyAndNotNull(TiC.PROPERTY_BACKGROUND_GRADIENT)) {
```
hansemannn: @ypbnv This is what broke it I think. You should handle both background-colors and background-gradients.

hansemannn:
Closing for now due to open issues with backwards compatibility and merge conflicts.

PR #11220, opened 2019-09-16
fix(ios): sf symbol handling for application shortcut

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27394


iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
```diff
@@ -136,6 +137,23 @@ - (id)initWithImage:(UIImage *)image_
...
 }
 
+- (id)initWithSystemImage:(NSString *)imageName
+{
+  if (![TiUtils isIOSVersionOrGreater:@"13.0"]) {
+    return nil;
+  }
+
+  if (self = [super init]) {
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
```
janvennemann:
```suggestion
#if IS_SDK_IOS_13
```

I feel a little uncomfortable only wrapping the image creation in the iOS 13 condition. Sure, this code block should never be reached when not built with iOS 13 SDK but only wrapping that single line inside it calls for unexpected behavior in the following lines, e.g. when module developers use `TiBlob` and wrongfully use `initWithSystemImage:`.

janvennemann (changes requested):

iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
vijaysingh-axway: Agree. Wrapped whole function inside macro.

Macro IS_SDK_IOS_13 will not be available inside TitaniumKit framework.

janvennemann (approved):

ssekhri:
FR Passed. An image blob is shown as part fo the app shortcut. Also no error is shown when adding the shortcut.
Verified on:
Mac OS 10.14.6
Ti SDK: 8.3.0.v20190923105712
Appc CLI: 7.1.1
Node: 10.5.0
JDK: 10.0.2
XCode: 11.0
Studio: 5.1.4.201909061933
Device: iPhone X(v13.0)

PR #11222, opened 2019-09-16
fix(ios)(8_2_X): sf symbol handling for application shortcut

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27394

### DIFF
--- a/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
+++ b/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
@@ -313,6 +313,8 @@ properties:
 
         The recommended size for image files is 35dp (@2px: 70dp, @3x: 105dp). Also check the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationshortcuticon)  
         for more information on shortcut icons.
+        
+        From SDK 8.2.1 and iOS 13+ you can use system image got from <Ti.UI.iOS.systemImage>.
 
     type: [Number,String,Titanium.Contacts.Person]
     constants: Titanium.UI.iOS.SHORTCUT_ICON_TYPE_*
--- a/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
+++ b/apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
@@ -314,7 +314,7 @@ properties:
         The recommended size for image files is 35dp (@2px: 70dp, @3x: 105dp). Also check the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationshortcuticon)  
         for more information on shortcut icons.
         
-        From SDK 8.2.1 and iOS 13+ you can use system image got from <Ti.UI.iOS.systemImage>.
+        On iOS 13 and higher, with SDK 8.2.1 or higher, you can pass in the Ti.Blob instance returned from <Ti.UI.iOS.systemImage>.
 
     type: [Number,String,Titanium.Contacts.Person]
     constants: Titanium.UI.iOS.SHORTCUT_ICON_TYPE_*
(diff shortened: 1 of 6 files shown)


apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
```diff
@@ -313,6 +313,8 @@ properties:
 
         The recommended size for image files is 35dp (@2px: 70dp, @3x: 105dp). Also check the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationshortcuticon)  
         for more information on shortcut icons.
+        
+        From SDK 8.2.1 and iOS 13+ you can use system image got from <Ti.UI.iOS.systemImage>.
```
sgtcoolguy:
```suggestion
        On iOS 13 and higher, with SDK 8.2.1 or higher, you can pass in the Ti.Blob instance returned from <Ti.UI.iOS.systemImage>.
```

iphone/Classes/TiUIApplicationShortcutsProxy.m
```diff
@@ -211,6 +212,14 @@ - (UIApplicationShortcutIcon *)findIcon:(id)value
     return [UIApplicationShortcutIcon iconWithTemplateImageName:[self urlInAssetCatalog:value]];
   }
 
+#ifdef IS_SDK_IOS_13
+  if ([value isKindOfClass:[TiBlob class]] && [TiUtils isIOSVersionOrGreater:@"13.0"]) {
```
sgtcoolguy: do you need to check the iOS version here if you have the ifdef guard as well?

iphone/Classes/TiUIiOSApplicationShortcutsProxy.m
```diff
@@ -211,6 +212,15 @@ - (UIApplicationShortcutIcon *)findIcon:(id)value
     return [UIApplicationShortcutIcon iconWithTemplateImageName:[self urlInAssetCatalog:value]];
   }
 
+#ifdef IS_SDK_IOS_13
+  if ([value isKindOfClass:[TiBlob class]] && [TiUtils isIOSVersionOrGreater:@"13.0"]) {
```
sgtcoolguy: same as above...

sgtcoolguy (changes requested):
docs need a tweak

iphone/Classes/TiUIApplicationShortcutsProxy.m
```diff
@@ -211,6 +212,14 @@ - (UIApplicationShortcutIcon *)findIcon:(id)value
...
+    TiBlob *blob = (TiBlob *)value;
+    if (blob.type == TiBlobTypeSystemImage) {
+      return [UIApplicationShortcutIcon iconWithSystemImageName:blob.systemImageName];
```
hansemannn: A wild guess here: Instead of creating new blob types and all this handling, why not just introduce a new `systemIcon` String proxy property and pass it directly to the `iconWithSystemImageName` constructor? That'd probably be around 10 lines of code in total.

iphone/Classes/TiUIApplicationShortcutsProxy.m
vijaysingh-axway: Yes. Its needed. UIApplicationShortcutIcon's api 'iconWithSystemImageName' is not available for iOS < 13. Let's say I ran my app using sdk 8.2.1 with Xcode 11 and run on iOS 12, in that case it will give error at run time.

iphone/Classes/TiUIApplicationShortcutsProxy.m
vijaysingh-axway: @hansemannn 1. UIApplicationShortcutIcon has 4 type of icons, which are getting handled from same place in SDK. And developer have to pass proper value to 'icon' property and it gets internally handled on basis of type.
2. Only one type of icon can be set at a time. Having 2 property will create confusion.
3. Similar type of handling sf symbols will be better for developer.

janvennemann (approved):
LGTM!

ssekhri:
FR Passed. An image blob is shown as part fo the app shortcut. Also no error is shown when adding the shortcut.
Verified on:
Mac OS 10.14.6
Ti SDK: 8.2.0.v20190920114538
Appc CLI: 7.1.1
Node: 10.5.0
JDK: 10.0.2
XCode: 11.0
Studio: 5.1.4.201909061933
Device: iPhone X(v13.0)

PR #11223, opened 2019-09-17
- TIMOB-27406 fixed

phobeous:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27406

`TIMOB-27406` Selecting a file from downloads folder raises a Long parse exception on Android 8 or higher.

**Optional Description:**
DocumentsContract.getDocumentId on Android 8+ does not return an integer but a full path prefixed with "raw:/" what raises a parse exception.


jquick-axway:
Honestly, the code you are editing in the "TitaniumBlob.java" file is a hack. The only reason it exists is to maintain backward compatibility where some Titanium devs expect to have direct file access for the given URL. But devs can't depend on this behavior and Android 7.0 no longer allows one app to access another app's files under external storage.

The correct solution is to remove that entire code block and **only** use the `ContentResolver` code at the bottom of that method. That would work for all cases. Hmm...

sgtcoolguy:
Closing in favor of #11395. Thanks for the PR, though!

PR #10637, opened 2019-01-19
feat(android): Added Ti.App "close" event support

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26542

**Summary:**
- Fires "close" event just before JS runtime terminates.
  - We document that this must be the last event. So, this is the perfect place to do it.
  - Had to modify `V8Object` and `V8Function` to allow V8 calls while in the "released" state.
- KrollRuntime was refactored a bit:
  - Removed `RELAUNCH` state. No longer needed since 8.0.0's intent refactoring and now that Services/BroadcastReceivers won't keep JS runtime alive anymore.
  - Removed `CountDownLatch`. No longer needed since JS runtime can only be ran from UI thread.
  - Modified `V8Object` and `V8Function` to allow V8 calls while `KrollRuntime` is in the `RELEASED` state, which happens when an `OnDisposingListener` is being invoked.
- Moved `cancelTimers()` call from `KrollRuntime` to `TiApplication`.
  - Doing this via listener is a better solution. No longer tightly coupled.

**Note:**
I've noticed that our `TiApplication.softRestart()` does not follow the rules and calls the `V8Runtime`'s dispose/init methods directly. This means our `OnDisposingListener`s won't get invoked. The `Ti.App._restart()` handling still works with my PR (I've tested it), but we may want to look into this in the future.

**Close Event Test:**
1. Build and run the below code on Android.
2. Press the "Back" button.
3. Verify the following got logged: `### App 'close' event was fired.`
4. Relaunch the app.
5. Repeat steps 2 and 3.

```javascript
Ti.App.addEventListener("close", function() {
	Ti.API.info("### App 'close' event was fired.");
});
Ti.UI.createWindow().open();
```

**LiveView Test:**
_This test verifies that our runtime restart changes work okay._
1. Build and run [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) via LiveView on Android.
2. Verify that the app launches okay.
3. Edit a JS file and click save.
4. Verify that the app re-launches okay.

PR #11224, opened 2019-09-17
fix(android): fix views with border and transparency

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26394

**Description:**
Rework handling the disabling of the HW acceleration for views with transparency and borders. `isHardwareAccelerated()` returns whether the view is attached to a Window with enabled HW acceleration:
https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/core/java/android/view/View.java#20915
So it depends on the view being attached to a Window. Our call of `disableHWAcceleration()` in TiUIView.java happens before the native view is actually attached to a window, so it would require an additional check in order to determine when to disable acceleration for the TiBorderWrapperView. Currently `borderView.isHardwareAccelerated()` will always return false, resulting in skipping our method.
We had a change in 6.3.0.GA for this code in this PR:
https://github.com/appcelerator/titanium_mobile/pull/9410
Ran the test cases from it and it works fine both on Android 19 and Android 29.
This PR also fixes the case described in this JIRA ticket - https://jira.appcelerator.org/browse/TIMOB-26291
During FR I would recommend trying the test cases from all three PRs to avoid regressions.
Note: No unit test, since this is a visual fix.

**Test case:**
_app.js_
```js
     backgroundColor: '#fff'
});
 
var v1 = Titanium.UI.createView({
     backgroundColor: 'rgba(0,0,0,0.5)',
     height: 100,
     width: Ti.UI.FILL,
     borderRadius: 20
});
win.add(v1);
win.open();


```


garymathews (approved):
CR: **PASS**

ssjsamir (approved):
FR Passed: Views with borderRadius now show up semi-transparent. Tested with the test case in the description and additional tests from the mentioned PRs above were done.

**Test Environment**
```
MacOS Catalina 10.15 Beta
Node.js ^8.11.1
Android Pixel XL 7.1 Emulator 
"NPM":"4.2.15","CLI":"7.1.1"
Java Version 1.8.0_131
Android NDK:  12.1.2977051
```

hansemannn:
Either this one or https://github.com/appcelerator/titanium_mobile/pull/10637 is causing a regression with Ti.Map which now displays an empty map, after updating to latest master. You can try with our internal project, I think both @garymathews  and @jquick-axway have access to it.

jquick-axway:
@hansemannn, we'll check it out. Thanks.
Question: Do you have a border applied to your map?

jquick-axway:
Okay, I've confirmed that we have a regression on "master". If you apply a border to a VideoView or MapView (something that requires hardware acceleration), then it won't render. I can understand this happening if "borderRadius" was involved (that's buggy on Google's end), but this is happening to rectangular borders now.
```javascript
var videoView = Ti.Media.createVideoPlayer({
	url: "http://assets.appcelerator.com.s3.amazonaws.com/video/media.m4v",
	autoplay: true,
	borderWidth: "4dp",
	borderColor: "red",
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
window.add(videoView);
window.open();
```

jquick-axway:
The fix for the regression can be found here:
https://github.com/appcelerator/titanium_mobile/pull/11262

It turns out the old code (before this PR) was wrongly "trying" to disabling hardware acceleration when it didn't need to, but failing due to what this PR was fixing. So, this PR merely the old bug to the surface.

hansemannn:
Thanks Josh, that was fast!

PR #11230, opened 2019-09-18
refactor: remove grunt usage and replace with npm scripts

ewanharris:
This removes the usage of grunt and goes all in on using npm scripts in its place.

Here's a before and after of what you want to run

`grunt`/`grunt lint` -> `npm run lint`
`grunt format:TYPE` -> `npm run format:TYPE` where TYPE is one of android, ios, or js

There is now also separate linting for the each type that can be used like `npm run:TYPE` where type is one of android, docs, ios, or js

I'm filing this as a WIP for now as I'm going to start work on hooking up the existing unit tests under android/cli in this PR also. I just wanted to see how Jenkins handles these changes, and gather some opinions

Edit:

Updated PR to run any tests that match the pattern `**/cli/tests/test-*.js`, this includes the existing AndroidManifest parser tests, which have also been updated to use chai.expect (the existing should usage failed on every `should match object` test as it seems between those being written, and today should also started asserting the prototype)

Next steps is to try integrate those into CI


package.json
```diff
@@ -18,13 +18,23 @@
...
+    "lint": "npm-run-all --parallel lint:**",
+    "lint:android": "./build/scons format --android",
+    "lint:docs": "tdoc-validate ./apidoc",
+    "lint:ios": "./build/scons format --ios",
+    "lint:js": "eslint .",
+    "format": "npm-run-all --parallel format:**",
+    "format:android": "./build/scons format --android --fix",
+    "format:ios": "./build/scons format --ios --fix",
+    "format:js": "npm run lint:js -- --fix",
+    "unit-tests": "nyc --exclude **/cli/tests/test-*.js mocha **/cli/tests/test-*.js"
```
sgtcoolguy: so in keeping with this theme, I would suggest we name this "test:unit" and migrate "android-test" and "ios-test" to be "test:android" and "test:ios" respectively.

sgtcoolguy: or maybe even "test:cli"? "test:js" hopefully align the naming across the various actions.

ewanharris: Added the following `test:android`, `test:cli`, `test:ios`.

I didn't point `npm test` to run these though as I think that is probably unwanted. I also wasn't sure on whether to have `cli` or `js` in the naming, I went with the former as while it isn't inline with the format/lint steps, it is clearer as to what the task does.

Jenkinsfile
```diff
@@ -168,6 +168,27 @@ def unitTests(os, nodeVersion, npmVersion, testSuiteBranch, testOnDevices) {
 	}
 }
 
+def cliUnitTests(nodeVersion, npmVersion) {
+	return {
+		node('git && osx') { // ToDo: refactor to try and run across mac, linux, and windows?
```
ewanharris: I guess this may come at the annoyance of having to skip certain tests on certain platforms

janvennemann (approved):
Awesome, @ewanharris! Looks good to me!

sgtcoolguy:
manually rebased on master, handled conflicts and merged.

PR #11246, opened 2019-09-27
fix(ios): added xcworkspacedata file generation

cb1kenobi:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27331


iphone/cli/commands/_build.js
```diff
@@ -3825,6 +3825,24 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 			this.logger.debug(__('Writing %s', dest.cyan));
 			fs.ensureDirSync(parent);
 			fs.writeFileSync(dest, contents);
+
+			// Xcode 11 requires the workspacedata file to exist or else it doesn't properly
+			// rebuild the app if the Xcode project changes
+			const xcworkspaceDir = path.join(this.buildDir, this.tiapp.name + '.xcodeproj', 'project.xcworkspace');
```
hansemannn: It should also be added in the iOS Xcode template, so users don't need to create this file on every clean build.

hansemannn: Also, I think it's somehow causing a force-rebuild. At least, all source SDK files (*.m) are recompiled.

hansemannn: And finally, the logs don't work again after running it 2-3 more times.

cb1kenobi: Good point, I can add the xcworkspace to the Titanium.xcodeproj dir.

You are correct, the build is doing a rebuild because the Xcode project keeps changing because the log server port is defined in as a GCC define.

Gah, so the fix doesn't work. I'll have to take a peek.

Without rehashing all of the reasons that got us into this place, we have a couple options:

1. Have the iOS build attempt to connect to the app and verify the app id, before bothering to pick a random port. The build used to do this, but it only works if the app is foregrounded.

2. Move the log server port from a GCC define to a json file that is parsed at runtime. We would probably still need to run xcodebuild, but it should be super fast since no code changed.

3. Rip out the log server and have NSLog() write to a log file for simulator builds only, then the iOS build would tail that file. I believe this is how it used to work. Device builds would continue to use the log server.

4. Something else?

I personally like option 3. Simple to implement.

hansemannn: If device builds work fine now, 3. would be a good option. But wouldn't the file inside the `.xcodeproj` fix it already? And for the forced change, either the file can be unmarked (like some others already).

cb1kenobi: From what I can tell, the `project.xcworkspace` is requried by Xcode 11 for xcodebuild to detect changes to project build settings like the gcc defs. So, we need this file regardless.

When it builds your app and launches the app in the simulator, the build will have already selected the log server port. On a subsequent build, the port is in use, thus a new random port is selected and set in the gcc def. Now xcodebuild detects the setting change and builds only the affected files which unfortunately is basically all of them and that's the rebuild you are seeing.

If you simply kill the app in the simulator (not backgrounding the app or killing the sim, but kill the app), then when you rebuild, the build will use the same log server port and no rebuild is necessary.

Yeah, the `project.xcworkspace` should maybe be managed by the iOS build and unmarked for differential builds, but given this file never changes, I sorta cheated and just make sure it exists.

hansemannn:
@cb1kenobi Works fine now!

cb1kenobi:
@hansemannn Yeah, but the rebuilds are painful.

I'm currently implementing option 3 as described above and digging through the old 5.x code to see how the simulator builds used to write log files. It used to direct all NSLog() calls to `~/Library/Developer/CoreSimulator/Devices/<udid>/data/.../<app guid>.log`.

I found it in the tooling and got that updated, but can't find it in the Obj-C code. I believe it had to do with https://github.com/appcelerator/titanium_mobile/blob/5_5_X/iphone/Classes/TiBase.m#L117-L120. Sorta lost.

cb1kenobi:
@hansemannn OK, I figured it out. I reverted to file-based logging for simulator builds and now no need to invoke xcodebuild. Build times are back down to about 2 seconds.

We still use the log server for device builds.

I think there's room for improvement, but probably good enough for now.

hansemannn:
I feel like this is the wrong direction. Writing to a log file (i/o on the main thread) is a huge performance killer on dev and there must definitely be a better way. It's impolite to ask, but can you maybe investigate this a bit further? I'm happy to help and test!

cb1kenobi:
@hansemannn again, this for simulator builds only. I don't see how this is a huge performance killer. If you were doing a bunch of logging and had a 10 year old laptop with a spinning hard drive, then maybe, but if you're doing some general debug logging and you have a relatively new Mac with an SSD, I highly doubt you'd even notice.

Did you know this is how logging was performed from Titanium v1 through v5? It worked fine. The only reason we moved away from this was to standardize on the log server for both sim and device builds. Device builds never ran into the same problem because the likelihood the port is in use on your device is way less than on your local machine where the simulator has to share ports with the host including the already running version of your app from the last build.

So, I will run this design change past the team and unless there is a reasonable concern, this solution is the one we'll go with. Feel free to prove this design is bad and/or propose a better one.

hansemannn:
It just a worse solution than finding the actual issue and fix it. And it's not my job to write CLI code, sorry.

iphone/cli/commands/_build.js
```diff
@@ -4596,6 +4597,11 @@ iOSBuilder.prototype.copyTitaniumiOSFiles = function copyTitaniumiOSFiles() {
 		path.join(this.platformPath, 'iphone', 'Titanium.xcodeproj', 'xcshareddata', 'xcschemes', 'Titanium.xcscheme'),
 		path.join(this.buildDir, this.tiapp.name + '.xcodeproj', 'xcshareddata', 'xcschemes', name + '.xcscheme')
 	);
+	copyAndReplaceFile.call(
+		this,
+		path.join(this.platformPath, 'iphone', 'Titanium.xcodeproj', 'project.xcworkspace', 'contents.xcworkspacedata'),
```
ewanharris: I think we're missing this file from the PR? Jenkins is erroring with

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ENOENT: no such file or directory, stat '/Users/build/Library/Application Support/Titanium/mobilesdk/osx/8.2.1.v20190930143205/iphone/iphone/Titanium.xcodeproj/project.xcworkspace/contents.xcworkspacedata'
```

ewanharris (changes requested):
Missing a file I think?

iphone/cli/commands/_build.js
cb1kenobi: Fixed!

hansemannn:
@cb1kenobi I still don't get it, sorry. If you need to do save the logs to a physical file and read it again, why not just use the console API's Apple offers already? The console API is available since Xcode 8 I think and is widely used in other projects like [NSLogger](https://github.com/fpillet/NSLogger/) to stream logs from outside Xcode. It also fixes some other issues like not being able to receive Swift's `debugLog` logs, since they are no NSLog instances and therefore could not be caught from Titanium before.

That should be the right direction. If you need a prototype of the API, I can probably help.

cb1kenobi:
I've created https://jira.appcelerator.org/browse/TIMOB-27442 to investigate other means of relaying log messages. Being that the 8.2.1 release is right around the corner, I don't think there's time to switch to a different mechanism.

ewanharris (approved):
Copying review from other PR

The CLI code looks ok to me, I'm able to rebuild with the app open just fine and don't get a full rebuild.

I'm going to ask @appcelerator/ios to take a check of the SDK code though.

Just wanted to also note that we now get some extraneous logging in comparison to 8.2.0.GA
Before:

-- Start simulator log -------------------------------------------------------
[DEBUG] Reading stylesheet from: /Users/awam/Library/Developer/CoreSimulator/Devices/99124622-6287-4600-B62A-FE053C8B8B5E/data/Containers/Bundle/Application/1E45668E-CA30-48B2-8DE8-CBF4BD3ABEF2/ticreateapp.app/stylesheet.plist
[INFO]  ticreateapp 1.0 (Powered by Titanium 8.2.0.6bfdf6e394)
[DEBUG] Firing app event: started
[INFO]  load
With this PR:

-- Start simulator log -------------------------------------------------------
[DEBUG] Reading stylesheet from: /Users/awam/Library/Developer/CoreSimulator/Devices/99124622-6287-4600-B62A-FE053C8B8B5E/data/Containers/Bundle/Application/639726C4-D6B6-46BE-8D05-67716B56645B/ticreateapp.app/stylesheet.plist
[DEBUG] You've implemented -[<UIApplicationDelegate> application:performFetchWithCompletionHandler:], but you still need to add "fetch" to the list of your supported UIBackgroundModes in your Info.plist.
[DEBUG] You've implemented -[<UIApplicationDelegate> application:didReceiveRemoteNotification:fetchCompletionHandler:], but you still need to add "remote-notification" to the list of your supported UIBackgroundModes in your Info.plist.
[TRACE] [ticreateapp] assertion failed: 18G95 17A577a: libxpc.dylib + 83495 [9A486F3D-BBFA-36F2-9105-42A059A13A8A]: 0x7d
[DEBUG] APSAnalytics: Initialization started...
[INFO]  APSAnalyticsStore: session.start
[INFO]  ticreateapp 1.0 (Powered by Titanium 8.2.1.ad9ad91a4a)
[INFO]  Reachability Flag Status: -R ------- networkStatusForFlags
[DEBUG] Firing app event: started
[INFO]  Application started
[INFO]  load
[DEBUG] APSAnalytics: Attempting to send 1 analytics event(s)
[INFO]  APSAnalytics: Successfully sent 1 stored event(s)

lokeshchdhry:
FR Passed.

Logs show up for concurrent builds.

SDK Ver: 8.2.1 local build
OS Ver: 10.14.5
Xcode Ver: Xcode 11.1
Appc NPM: 4.2.15
Appc CLI: 7.1.2-master.3
Daemon Ver: 1.1.3
Ti CLI Ver: 5.2.2
Alloy Ver: 1.14.1
Node Ver: 10.16.1
NPM Ver: 6.9.0
Java Ver: 10.0.2
IOS Simulator: ⇨ iPhone 11 Pro (com.apple.CoreSimulator.SimRuntime.iOS-13-1)

lokeshchdhry:
Waiting for the review for the extra log mentioned above by @ewanharris .

janvennemann (approved):
iOS changes are looking good.

BTW, this also solves an issue where logs from third-party frameworks in native modules where missing. They are now showing up again. See the related ticket [TIMOB-26861](https://jira.appcelerator.org/browse/TIMOB-26861). I guess this is a result of the additional logs that are now printed, as Ewan already mentioned.

lokeshchdhry:
Is this good to merge then ?

PR #11249, opened 2019-09-30
fix(ios): fix TiBase header to order macros properly, re-use macros

sgtcoolguy:
**Description:**
While attempting to build ti.barcode-zxing against SDK 8.2.0.GA, it would fail due to multiple matches of `count` method against the `args` for capture/parse going through our TiBase.h macros. This ia a very specific bug fix for building native modules against the SDK, and seems to only come up when you have 3rd party headers that may define types with a count method.

This fixes the macros to explicitly cast to `NSArray *` (after checking for it) when calling count on the argument, which fixes the issue. I also re-ordered the macro definitions so they proceed in-order (i.e. a macro doesn't refer to another defined below it, must be above); and to have macros call other macros rather than re-implement the same code to check types/etc.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiBase.h
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiBase.h
@@ -129,6 +129,24 @@ NSString *JavascriptNameForClass(Class c);
 
 #define NULL_IF_NIL(x) ({ id xx = (x); (xx==nil)?[NSNull null]:xx; })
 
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
See https://jenkins.appcelerator.org/blue/rest/organizations/jenkins/pipelines/modules/pipelines/ti.barcode/branches/PR-112/runs/4/nodes/48/steps/82/log/?start=0 for example of how this is breaking that module's build

PR #11253, opened 2019-10-02
fix(android): performance issue with deeply nested views as of 7.5.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27434

**Summary:**
- A re-layout of 15 or more nested views can cause major performance issues as of 7.5.0.
  * Caused by window inset handling needed by our "extendSafeArea" and "safeAreaPadding" features.
  * Inset propagation to child views (which in turn trigger re-layouts) gets exponentially worse the deeper it gets.
- Resolved by storing last received insets and only dispatching insets to children if changed.

---
**Performance Test:**
1. Build and run the code attached to [TIMOB-27434](https://jira.appcelerator.org/browse/TIMOB-27434) on Android.
2. Tap-and-hold on the text within the `TextField`.
3. Verify that copy/paste clipboard immediately appears. (Used to hang for a few seconds.)
4. Rotate the app to landscape.
5. Verify that the app does not hang for a few seconds when rotated.

---
**Inset Test:**
_This verifies that insets are still correctly propagated to child views._
1. Build and run the "ExtendSafeAreaDrawerTest.js" attached to [TIMOB-26427](https://jira.appcelerator.org/browse/TIMOB-26427) on Android.
2. Verify that red title bar extends beneath the status bar and its title text is not overlapped by the status bar, as shown in [TIMOB-26427](https://jira.appcelerator.org/browse/TIMOB-26427) screenshot.
3. Tap on the hamburger button in the top-left corner to open the drawer layout.
4. Verify that drawer's blue title bar extends beneath status bar and its text is not overlapped by the status bar, as shown in [TIMOB-26427](https://jira.appcelerator.org/browse/TIMOB-26427) screenshot.
5. Rotate the app to landscape and verify both the red and blue title bars text is not overlapped as shown in the screenshots.


android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
```diff
@@ -464,6 +472,13 @@ public WindowInsets onApplyWindowInsets(WindowInsets insets)
 			return null;
 		}
 
+		// Do not propagate given insets to child views unless the insets have changed.
+		// This greatly improves performance if we have deeply nested views since each view will trigger a relayout.
+		if (insets.equals(this.previousInsets)) {
+			return insets;
+		}
+		this.previousInsets = insets;
```
sgtcoolguy: Does this also need an Android api level guard? Would accessing it on api level < 20 cause a crash or something?

sgtcoolguy (dismissed):

android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
jquick-axway: We don't need a guard inside the `onApplyWindowInsets()` method since it has argument type `WindowInsets` which is API Level 20 only. Attempting to call `onApplyWindowInsets()` on a lower API level would crash, meaning that the API Level guard needs to be done on the caller's side.

But good question. Thanks for asking.

jquick-axway: That said, I'll do a quick experiment with Google's `@RequireApi(ApiLevel)` Java annotation. Android Studio's built-in linting tool is supposed to trigger compiler errors/warnings when making unguarded calls to methods like this. When we switch over to gradle, this may help reduce crashes on older OS versions.

jquick-axway: I wrote up a ticket [TIMOB-27437](https://jira.appcelerator.org/browse/TIMOB-27437) to implement the `@RequireApi()` annotations in the future. We can't leverage it until we support the gradle linting tool and add the Google Support Annotation library.

lokeshchdhry:
FR Passed.

SDK Ver: 8.3.0 local build
OS Ver: 10.14.5
Xcode Ver: Xcode 11.0
Appc NPM: 4.2.15
Appc CLI: 7.1.1
Daemon Ver: 1.1.3
Ti CLI Ver: 5.2.1
Alloy Ver: 1.14.1
Node Ver: 10.16.1
NPM Ver: 6.9.0
Java Ver: 10.0.2
Android Devices: ⇨ google Pixel (Android 9)
Emulator: Android 6.0

PR #11262, opened 2019-10-04
fix(android): VideoPlayer with border fails to render as of 8.3.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27444

**Summary:**
- This issue was caught before release.
- Bug was hardware acceleration was being wrongly disabled on views with rectangular borders.
  * This is an issue for views that require hardware acceleration such as `VideoPlayer` and `MapView`. These 2 views will fail to render in this case.
- Modified call to setLayerType() to no longer use reflection. We can call directly since min API Level supported is 19.

---
**Video Player Test:**
1. Acquire an Android device with Internet access.
2. Build and run the below code on that Android device.
3. Verify that the video plays on startup.
4. Verify that you see red rounded border around the video.

```javascript
var videoView = Ti.Media.createVideoPlayer({
	url: "http://assets.appcelerator.com.s3.amazonaws.com/video/media.m4v",
	autoplay: true,
	borderRadius: 20,
	borderWidth: "4dp",
	borderColor: "red",
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
window.add(videoView);
window.open();
```

---
**Rounded Border Transparency Test:**
_(Verifies that [TIMOB-26394](https://jira.appcelerator.org/browse/TIMOB-26394) is still fixed.)_
2. Verify that you see a semi-transparent blue rectangle with rounded borders on top of a white background with a red plus.

```javascript
	backgroundColor: "white",
})
window.add(Ti.UI.createView({
	backgroundColor: "red",
	width: "50dp",
	height: Ti.UI.FILL,
}));
window.add(Ti.UI.createView({
	backgroundColor: "red",
	width: Ti.UI.FILL,
	height: "50dp",
}))
window.add(Ti.UI.createView({
	backgroundColor: "rgba(0.0,0.0,1.0,0.5)",
	borderRadius: 20,
	width: "85%",
	height: "50%",
}));
window.open();
```


ypbnv (approved):
Looks good to me!

jquick-axway:
@ypbnv, for your info, you did not create the bug. Your PR merely brought it to the surface. What was happening was the below code was wrongly disabling hardware acceleration if "backgroundColor" was null/undefined (which is the default) or if its alpha value was set to `0xFF`. It was an issue with the if-condition because the `0xFF` was being treated as an `int` (the default behavior) since it wasn't casted to `byte`. Which means `0xFF` as an integer would be treated as `255` while a `(byte)0xFF` would be treated as `-1`... and they never ended up equalling each other... ever.
```java
byte bgAlpha = bgColor != null ? (byte) (bgColor >> 24) : (byte) 0xFF;
if (bgAlpha != 0xFF) {
	disableHWAcceleration();
}
```

ssaddique:
FR: Passed. I can confirm that the bug has been fixed and that the video renders correctly, as of the build from this PR.

jquick-axway:
@ssaddique, I don't think that error is related to the issue this PR. Try testing it with an older Titanium SDK (such as 8.2.0) via the same emulator. Does the error still happen? If so, then let's write it up as a separate bug report. Thanks.

ssaddique:
@jquick-axway - You are correct. I have tested with various older SDK's and only API 22 (as far as I'm aware - I don't have any older emulators/device to test) is affected and is **not** related to this PR. Looks like a separate ticket will need to be created for <= API 22.

PR #11269, opened 2019-10-09
build: allow building a windows ti.main.js

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27452

Allow specifying `--all` to the `scons build` command that will ensure we create a   `ti.main.js` entry for Windows when building. Then use this in Jenkins

Verified locally by pulling down Windows artifacts from the master branch, and setting up to match a mainline branch build. The Windows zip had the windows `ti.main.js` file. I think we need to update the Windows CLI to look for it there though

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -294,7 +294,11 @@ timestamps {
 					// TODO parallelize the iOS/Android/Mobileweb/Windows portions?
 					dir('build') {
 						timeout(15) {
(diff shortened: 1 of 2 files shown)


sgtcoolguy (approved):
LGTM

infosia:
Can we test this with higher priority and merge asap because this is blocker issue to build Windows nightly artifacts.

PR #11274, opened 2019-10-10
fix(ios): remove additional gc protection once proxy is remembered (8_2_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27207

8_2_X backport of #11028


ssaddique:
FR: Failed. Please see attatched PDF of findings: 
[Findings for TIMOB-27207.pdf](https://github.com/appcelerator/titanium_mobile/files/3712401/Findings.for.TIMOB-27207.pdf)

vijaysingh-axway (approved):
Looks good.

lokeshchdhry:
@ssaddique , For me when I check it on a device < ios 13. I can see that objects are getting garbage collected & memory is freed. Make sure wait for a few minutes for the GC to happen sometimes.

ssaddique:
@lokeshchdhry - I checked on iOS 13 simulator iPhone 11 Pro and I waited 15 minutes and only garbage collects just the one time for me. Maybe I'm doing it wrong, im not sure...

janvennemann:
@lokeshchdhry @ssaddique I've backported the debug function i added to the master PR so you guys can use it here to. Just add the following snippet to the test cases to automatically force a GC after 30s:

```js
setTimeout(() => {
  Ti.App.iOS.garbageCollectForDebugging();
}, 30000);
```

Let me know once the FR passes so i can remove the debug code from this PR again.

ssaddique:
@janvennemann - I can confirm that by adding the suggested code snippet, garbage collection works **reliably**.

Unfortunately, without the code snippet, garbage collection is **unreliable**. It always carries out GC at least once. Sometimes twice. But it is unreliable and most of the time further GC's do not occur.

janvennemann:
@ssaddique This is by design. JavaScriptCore decides when to schedule a GC as necessary. Sometimes this can occur after a few seconds, sometimes it takes minutes. It depends on several factors internal to JSC so using the debug function is the only way to reliably force it.

PR #11283, opened 2019-10-16
fix(ios): update objc module template clang settings

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27410

**Optional Description:**
Updates the Objective-C module template with the most recent default clang settings from Xcode 11. This includes `CLANG_ENABLE_MODULES=YES` which is required for compatibility with TitaniumKit, since it contains module imports. The Swift template already contains these settings.

Also bumps the deployment target to 9.0, to match our app project template.


vijaysingh-axway (approved):
CR passed.

ssekhri:
@janvennemann with this PR the clang setting to Enable Modules (C and Objective-C) in xcode project for the module does gets set to Yes by default. However following the test instructions and test code mentioned in the ticket https://jira.appcelerator.org/browse/TIMOB-27410 we still do get the same error with both Xcode 10.3 and 11.1.

janvennemann:
@ssekhri did you update the sample module? Also make sure to update the modules `titanium.xcconfig`, since they use a hardcoded value for the `TITANIUM_SDK` path.

ssekhri:
FR Passed. The clang setting to Enable Modules (C and Objective-C) is set to Yes by default. The module builds successfully.
Verified on:
Mac OS: 10.14.5
SDK: 8.3.0.v20191016042241
Appc CLI: 7.1.1
JDK: 1.8.0_162
Node: 10.5.0
Studio: 5.1.4.201909061933
Xcode: 11.1

PR #11289, opened 2019-10-22
fix(ios): update to core-js 3 to fix ios 9 crash

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27484

This is to fix a bug where the toStringTag property of a TypedArray was not able to be accessed and led to a crash on startup on iOS 9. Additionally refactored the xcode-project-build script to reuse the transpile code from builder.js.

I want to try and inspect the bundle here as it looks to have increased in size on iOS (Android has no change, I need to check Windows). It's probable that the replacement suggestion on the [babel docs](https://babeljs.io/docs/en/next/babel-polyfill.html) doesn't have the exact same mapping of polyfills so we're including more, which is probably better imo.


common/Resources/ti.main.js
```diff
@@ -23,7 +23,9 @@ try {
 }
 
 // Load JS language polyfills
-import '@babel/polyfill';
+import 'core-js/stable';
```
sgtcoolguy: To avoid the web exclusion stuff, maybe change this to `import 'core-js/es';`? That includes only language polyfills, and not web ones.See https://github.com/zloirock/core-js#commonjs-api

build/lib/builder.js
```diff
@@ -10,7 +10,6 @@ const commonjs = require('rollup-plugin-commonjs');
 
 const git = require('./git');
 const utils = require('./utils');
-const copyPackageAndDependencies = utils.copyPackageAndDependencies;
```
sgtcoolguy: I think the method in utils can also be removed since you removed both usages?

build/scons-xcode-project-build.js
```diff
@@ -49,59 +45,17 @@ async function generateIndexJSON(dirToTraverse) {
...
-		],
-		external: [ './app', 'com.appcelerator.aca', '/semantic.colors.json' ]
-	});
+	await builder.transpile('ios', ios.babelOptions, path.join(outputDir, 'ti.main.js'));
 
-	// write the bundle to disk
-	console.log('Writing common SDK JS bundle to disk');
-	return bundle.write({ format: 'cjs', file: `${outputDir}/ti.main.js` });
 }
 // FIXME: Combine common code here and in packager.js!
```
sgtcoolguy: Looks like maybe this can removed, since that's what you're doing here... :)

sgtcoolguy (changes requested):

common/Resources/ti.main.js
ewanharris: That reduced the size by around 100kb. There looks to be a 60kb increase over the current size but I feel that should be acceptable

build/scons-xcode-project-build.js
ewanharris: Doh 😆

sgtcoolguy (approved):

ssjsamir (approved):
**FR Passed** Able to build to an iOS 9 device without seeing any crashes. Tested using the default Titanium application.

**Test Environment**
```
MacOS Catalina 10.15.1 beta
Xcode 11
Node.js 10.16.3
iPhone 5s (9.3.5)
{"NPM":"4.2.15-1","CLI":"7.1.1"}
```

hansemannn:
This caused a HUGE regression:
```js
(Ti.Platform.osname === 'iphone' || Ti.Platform.osname === 'ipad')
```
On an iOS device (iPhone) returns false  (!) if <transpile>true</transpile>  is enabled. I’m shocked that this wasn’t covered by the appc unit tests so far.

cc @sgtcoolguy @ewanharris @garymathews

ewanharris:
@hansemannn I don't think this is the PR that has caused that. This change landed in 8.3.0.GA and I don't see the invalid behaviour there

Here's my test (I tried simulator and device)
```js
console.log((Ti.Platform.osname === 'iphone' || Ti.Platform.osname === 'ipad'));
console.log(Ti.Platform.osname === 'iphone');
console.log(Ti.Platform.osname === 'ipad');
```

8.3.1.GA (correct)

```
[INFO]  true
[INFO]  true
[INFO]  false
```

9.0.0.v20200124022032 (incorrect)

```
[INFO]  false
[INFO]  true
[INFO]  false
```

Looking at the generated code it's the below

```js
console.log(false);
console.log(Ti.Platform.osname === 'iphone');
console.log(Ti.Platform.osname === 'ipad');
```
I believe this is down to the [inlining values improvement](appcelerator/node-titanium-sdk/pull/131) in node-titanium-sdk (included in node-titanium-sdk@4.2.0) being included before #11400 was merged (#11443 updated to node-titanium-sdk@4.2.1). I'll confirm this on Monday, and make sure we prioritise that FR over others to avoid any issues for folks trying the nightly builds.

This wouldn't have made it to the GA as that improvement is in the 9.0.0 release. Looking at it, I'm going to posit that this is the cause of the test failures the SDK build has been seeing

hansemannn:
Yeah, we're on master., Thanks @ewanharris !

hansemannn:
@ewanharris Has there been progress on this issue?

ewanharris:
@hansemannn #11400 was merged in and should be available in the latest CI build (9.0.0.v20200128095338).

I gave it a quick test on an iOS sim and I'm seeing the correct values, let me know if you have any more issues

PR #11294, opened 2019-10-22
feat(android): Update V8 to 7.8.279.23

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27178

**Description:**
This is a work in progress to update to the latest V8 builds we can. Current Android SDK is on top of V8 7.3.492.26.

This PR currently updates us to V8 7.8.279.23
Current stable android v8 is 7.8.279.23 via https://omahaproxy.appspot.com

What I did:
- Attempted to maintain backwards compatibility for APIs that modules may use in the appcelerator/v8_titanium builds
- Update any V8 API usages in titanium_mobile to use the newer APIs and drop removed/deprecated API usages (including in the module template)
- Cleaned up some of the templates/c++ code to try and remove unnecessary headers includes (to hopefully speed up builds of the SDK a little)
- Commit c4aa037 includes a minor performance patch - it moves attempting to load "core" (read: native) modules *after* handling requires starting with `'/'` or `'.'`. When I turned on some debug logging in the c++ code, I saw that it kept trying to load bindings for `'.'`. This does mean that technically we'd "lose" the ability to load native modules starting with a `'.'`, but I'm not even sure we allow that in module id naming.
(Note that I also noticed a bug in the building of the require path used for add-on tests in the test suite where it was trying to load add-on test files with no `'/'` prefix, so it hit the "core" module code path and failed)


android/runtime/v8/src/native/modules/ScriptsModule.cpp
```diff
@@ -126,13 +126,13 @@ void WrappedScript::CreateContext(const FunctionCallbackInfo<Value>& args)
...
 
-	// TODO Do we need to esnure the same context object is wrapped and returned?
+	// TODO Do we need to ensure the same context object is wrapped and returned?
 	new WrappedContext(isolate, context); // wrap the context's global prototype...
 
 	// If a sandbox is provided initial the new context's global with it.
 	if (args.Length() > 0) {
 		Local<Object> sandbox = args[0].As<Object>();
-		Local<Array> keys = sandbox->GetPropertyNames();
+		Local<Array> keys = sandbox->GetPropertyNames(context).ToLocalChecked();
```
garymathews: I think this will break compatibility with prior native modules

garymathews:
- Updated to latest V8 version `7.8.279.23` which is patched to be fully backwards compatible with Titanium SDK 7.0.0+ native modules
- Updated various V8 API usages to the latest revisions

package.json
```diff
@@ -118,6 +118,7 @@
     "p-limit": "^2.2.0",
     "pngjs": "^3.4.0",
     "request": "^2.87.0",
+    "request-promise-native": "^1.0.8",
```
sgtcoolguy: this should be a devDependency, shouldn't it? We only use it for the snapshot generation during the SDK build itself.

android/runtime/v8/src/native/TypeConverter.cpp
```diff
@@ -239,7 +240,7 @@ jobject TypeConverter::jsObjectToJavaFunction(Isolate* isolate, JNIEnv *env, Loc
 {
 	Local<Function> func = jsObject.As<Function>();
 	Persistent<Function, CopyablePersistentTraits<Function>> jsFunction(isolate, func);
-	jsFunction.MarkIndependent();
+	// jsFunction.MarkIndependent(); // Method has been removed!
```
jquick-axway: Do we need to do anything here?

If this is where we box anonymous JS callbacks, then we shouldn't make it a weak reference since it might get garbage collected before invoked on the Java side.

build/lib/android/snapshot.js
```diff
@@ -1,114 +1,75 @@
...
 
-	// Generate 'startup.js'
-	const output = await promisify(ejs.renderFile)(path.join(__dirname, 'startup.ejs'), { script: commonBundle }, {});
-	await fs.writeFile(STARTUP_PATH, output);
+		console.log('Attempting to request snapshot...');
 
-	// Set correct permissions for 'mksnapshot'
-	await fs.chmod(MKSNAPSHOT_PATH, 0o755);
+		// Obtain snapshot `id` and start new snapshot generation.
+		const id = await request.post(SNAPSHOT_URL, { body: { v8, script }, json: true });
```
jquick-axway: If you don't have Internet access, then the `await request.post()` will reject and throw an exception. Perhaps we should catch it, log the reject's error, and continue without the snapshot?

Although for Jenkins builds of our SDK (namely release builds), we don't want this behavior.

garymathews: We're catching the exception [here](https://github.com/appcelerator/titanium_mobile/blob/master/build/lib/android/index.js#L70). So this shouldn't be an issue.

android/runtime/v8/src/native/TypeConverter.cpp
garymathews: I think this is okay, it looks like we store a persistent reference to the JS function so it can't be garbage collected and release the reference when the Java function is destroyed `Java_org_appcelerator_kroll_runtime_v8_V8Function_nativeRelease`

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

Checked liveview, kitchensink & native modules 7.0.0+.

Studio Ver:       6.0.0.201911251516
SDK Ver:          9.0.0 local build
OS Ver:           10.14.5
Xcode Ver:        Xcode 11.2.1
Appc NPM:         5.0.0-2
Appc CLI:         7.1.2
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.4
Node Ver:         12.13.1
NPM Ver:          6.12.1
Java Ver:         11.0.1
Android Devices:  ⇨ google Pixel (Android 10)

lokeshchdhry:
Can someone please resolve the conflict.

PR #11295, opened 2019-10-24
fix(ios): webview onlink is now called only on link activated

slash-84:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27522

What happens:

- on iOS both the onlink function and the beforeload event are called when setting the html, while on Android this happen only for beforeload.

- on iOS both the onlink function and the beforeload event are called also for embeds (like iframe) while on Android this not happen at all.

This is causing me some troubles working on iOS because of "false positive" events that i need to distinguish as user clicks or normal page loading.

I'd like to suggest a PR where

**solution 1:**

- iOS will send onlink only when navigation type is WKNavigationTypeLinkActivated
- beforeload event will expose the `isMainFrame` indication to indicate if it was generated on the main page or an iframe.

or

**solution 2:**

- iOS will send onlink only when navigation type is WKNavigationTypeLinkActivated
- beforeload event is generated only if navigationAction.targetFrame.isMainFrame is truly

At the moment I've implemented in the PR the first solution, since iOS is already a little different from Android because it already exposes the platform specific attribute `navigationType`


jquick-axway (changes requested):
I disagree with this change. (Although I understand why you're doing it.)

We don't want to provide a `navigationType` because there isn't a good way to make it cross-platform.

Instead, we should only invoke the `onlink` callback when iOS' navigation type is `WKNavigationTypeLinkActivated`.

slash-84:
> I disagree with this change. (Although I understand why you're doing it.)
> 
> We don't want to provide a `navigationType` because there isn't a good way to make it cross-platform.
> 
> Instead, we should only invoke the `onlink` callback when iOS' navigation type is `WKNavigationTypeLinkActivated`.

@jquick-axway thanks for the feedback. I've updated the PR accordingly.

vijaysingh-axway (commented):
CR passed.

ssekhri:
FR Passed. Webview onLink is not called along with beforeLoad when setting html. It is fired when a link in the hmtl is clicked and the lis being navigated.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20191127111135
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 5.1.4.201909061933
Xcode: 11.2
Device: iPhone 7Plus(v12.3.1), iOS simulator 13.2

PR #11301, opened 2019-10-29
feat(android): semantic colors and night mode

drauggres:
- generate color resources from semantic.colors.json
- add Ti.UI.Android.nightModeStatus API
- add Ti.UI.Android.MODE_NIGHT_* constants

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27501

Sample app:
https://github.com/ewanharris/darkmode-example/pull/1

``` javascript
if (OS_ANDROID) {
￼	const mode = Ti.UI.Android.nightModeStatus;
￼	if (mode === Ti.UI.Android.MODE_NIGHT_YES) {
￼		Ti.UI.semanticColorType = Ti.UI.SEMANTIC_COLOR_TYPE_DARK;
￼	} else {
￼		Ti.UI.semanticColorType = Ti.UI.SEMANTIC_COLOR_TYPE_LIGHT;
￼	}
￼}
```

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2375,7 +2375,10 @@ AndroidBuilder.prototype.generateAppProject = async function generateAppProject(
 		this.generateI18N(),
 
 		// Generate a "res/values" styles XML file if a custom theme was assigned in app's "AndroidManifest.xml".
-		this.generateTheme()
+		this.generateTheme(),
+
+		// Generate "semantic.colors.xml" in "res/values" and "res/values-night"
+		this.generateSemanticColors()
 	]);
 
 	// Generate an "AndroidManifest.xml" for the app and copy in any custom manifest settings from "tiapp.xml".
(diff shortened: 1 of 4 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/android/AndroidModule.java
```diff
@@ -228,6 +229,13 @@
 	@Kroll.constant
 	public static final int OVER_SCROLL_NEVER = 2; //android.view.View.OVER_SCROLL_NEVER;
 
+	@Kroll.constant
+	public static final int MODE_NIGHT_NO = Configuration.UI_MODE_NIGHT_NO;
+	@Kroll.constant
+	public static final int MODE_NIGHT_YES = Configuration.UI_MODE_NIGHT_YES;
+	@Kroll.constant
+	public static final int MODE_NIGHT_UNDEFINED = Configuration.UI_MODE_NIGHT_UNDEFINED;
```
hansemannn: Can you map `nightModeStatus` to the existing [https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI-property-semanticColorType](`semanticColorType`) property and it's `Ti.UI.SEMANTIC_COLOR_TYPE_DARK` + `SEMANTIC_COLOR_TYPE_LIGHT` type?

android/cli/commands/_build.js
```diff
@@ -3554,6 +3555,96 @@ AndroidBuilder.prototype.generateI18N = function generateI18N(next) {
 	next();
 };
 
+AndroidBuilder.prototype.generateSemanticColors = function generateSemanticColors(next) {
```
hansemannn: Does this also allow the UI to "refresh" automatically? Otherwise I wonder whats the benefit over using the existing one. Thx!

android/modules/ui/src/java/ti/modules/titanium/ui/android/AndroidModule.java
drauggres: Those constants are used as values of `Ti.UI.Android.nightModeStatus` which is an Android analogue of `Titanium.App.iOS.userInterfaceStyle`.
I believe `Titanium.App.iOS.userInterfaceStyle` does not return `SEMANTIC_COLOR_TYPE_DARK` or `SEMANTIC_COLOR_TYPE_LIGHT, does it (never checked, docs says it is `String[]`)?

hansemannn: It returns a Number (USER_INTERFACE_STYLE_UNSPECIFIED = 0, USER_INTERFACE_STYLE_LIGHT = 1, USER_INTERFACE_STYLE_DARK = 2). So if you change the docs, it should be all set.

drauggres: `USER_INTERFACE_STYLE_*` constants are defined in `Ti.App.iOS` namespace.

jquick-axway:
FYI: We're feature freezing Titanium 8.3.0. The plan is to release 8.3.0 in the very near future.

Our current focus is "gradle" support in Titanium 9.0.0 and we don't really want any more changes that will cause merge conflicts at the moment.

Also, I think it's important that we line-up the dark/light mode handling between Android and iOS as much as possible. That may mean moving our `Ti.App.iOS` APIs/constants to `Ti.App`. The Day/Night feature is a different animal and we may need to play around with the `Theme.AppCompat.DayNight` style, although I'm pretty sure this can only work by having the activity destroy/recreate itself like Sergey said.

hansemannn:
@jquick-axway At least for the constants, I think it's a pretty lightweight change for parity. Right now, it's already on the parity namespace (`Ti.UI.*`) but hardcoded to light mode on Android. If it could actually get the "real" values on the getter, it would be a quick win. @ewanharris is also into the topic.

m1ga:
guess this won't make it into 9.0.0? Would be nice to have parity for the dark-mode that is already working on iOS

hansemannn:
@m1ga It works well without these changes (we're not using it and have night mode support live in production).

m1ga:
@hansemannn interesting. For me `Ti.UI.semanticColorType` is always `light` so I only get the light colors in `fetchSemanticColor`.

hansemannn:
Oh, we made an own module for that (to stay SDK independent). I'll send it to you tomorrow!

drauggres:
Conflicts resolved.
Branch rebased onto `master`.

One of the features provided by this PR is:
> generate color resources from semantic.colors.json

So you will be able to use the same colors in custom `theme.xml` without defining them again.

sgtcoolguy:
Closed in favor of now-merged #11457 

Thanks again @drauggres !

PR #11303, opened 2019-10-30
fix(ios): ipad popover crash

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27503

Prevents changing the presentation controller's delegate for iPad popovers. It's set to the popover proxy by default which does all the required setup before showing the popover.


sgtcoolguy (approved):
LGTM

sgtcoolguy:
@janvennemann Can we add sample code to the test suite that used to trigger the crash? I understand we don't **currently** run tests on an iPad, but I suspect we may eventually add that to Jenkins.

ssekhri:
FR Passed. Popover on iPad does not crash and work fine.
Verified on:
Mac OS: 10.14.5
SDK: 8.3.0.v20191107080002
Appc CLI: 7.1.1
JDK: 1.8.0_162
Node: 10.5.0
Studio: 5.1.4.201909061933
Xcode: 11.1
Device: iPad7(v13.1), iPad simulator v13.1, 12.2, 11.4

PR #11308, opened 2019-10-31
fix(ios): avoid crash when openStream fails

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27515

**Description:**
wraps in try/catch to properly set exception on JS context
if JS code is in async callback/event/timer and has no try/catch it will still crash
Also, differs from 'old' proxies in that the uncaught exception handler code isn't invoked

I'm not sure what the correct way to handle this is. This at least offers the user the ability to avoid the crash. If the code is invoked in the main body of a js file, it'll pop the error handler/controller with the stack trace.

But if it's invoked on an event listener it will:
- crash if not surrounded by try/catch
- let the user handle the error if caught by try/catch

I don't think it's specific to this PR, but we should investigate having the "unhandled" case on an event listener/callback/timer also show the error controller like the "old proxies" (the ones using the JS C API, not the Obj-C API) do.

I thought we had tried to interject an unhandled exception handler in the Obj-C API before and had to roll it back because it was affecting the mocha test suite's timer behavior (See #10862)

### DIFF
--- a/iphone/Classes/FilesystemModule.m
+++ b/iphone/Classes/FilesystemModule.m
@@ -79,10 +79,18 @@
   NSString *resolvedPath = [self pathFromComponents:[args subarrayWithRange:NSMakeRange(1, [args count] - 1)]];
   NSArray *payload = [NSArray arrayWithObjects:resolvedPath, [NSNumber numberWithInt:mode], nil];
 
-  KrollContext *context = GetKrollContext([[JSContext currentContext] JSGlobalContextRef]);
+  JSContext *jsContext = [JSContext currentContext];
+  JSGlobalContextRef globalContext = [jsContext JSGlobalContextRef];
+  KrollContext *context = GetKrollContext(globalContext);
   KrollBridge *ourBridge = (KrollBridge *)[context delegate];
-  id file = [[[TiFilesystemFileStreamProxy alloc] _initWithPageContext:ourBridge args:payload] autorelease];
-  return [self NativeToJSValue:file];
+  @try {
+    id file = [[[TiFilesystemFileStreamProxy alloc] _initWithPageContext:ourBridge args:payload] autorelease];
+    return [self NativeToJSValue:file];
+  } @catch (NSException *exception) {
+    JSValue *jsException = [self NativeToJSValue:exception];
+    [jsContext setException:jsException];
+    return nil;
+  }
 }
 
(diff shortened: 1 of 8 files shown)


sgtcoolguy:
I've pushed a couple follow-on commits to address some of the test failures once we get past the crash. Much like on Android, iOS uses a special class to handle encrypted files (TiFilesystemBlobProxy, equivalent to Android's TiResourceFile). This implementation didn't have some methods at all, also didn't have implementations of some we could feasibly do. I fixed those issues.

Haven't looked at the Blob failures yet, I assume something similar there.

iphone/Classes/FilesystemModule.m
```diff
@@ -171,16 +180,21 @@ - (JSValue *)getFile
...
   NSString *newpath = [self pathFromComponents:args];
+  TiFile *fileProxy = [self getFileProxy:newpath];
+  return [self NativeToJSValue:fileProxy];
+}
 
-  if ([newpath hasPrefix:[self resourcesDirectory]] && ([newpath hasSuffix:@".html"] || [newpath hasSuffix:@".js"] || [newpath hasSuffix:@".css"] || [newpath hasSuffix:@".json"])) {
-    NSURL *url = [NSURL fileURLWithPath:newpath];
+- (TiFile *)getFileProxy:(NSString *)path
+{
+  if ([path hasPrefix:[self resourcesDirectory]] && ([path hasSuffix:@".html"] || [path hasSuffix:@".js"] || [path hasSuffix:@".css"] || [path hasSuffix:@".json"])) {
```
garymathews: I believe we can remove the suffix checks for `html`, `css` since we don't encrypt those.

garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

`Ti.Filesystem.openStream` does not cause a crash with `deploytype` test.

OS Ver:           10.14.5
Xcode Ver:        Xcode 11.1
Appc NPM:         4.2.15
Appc CLI:         7.1.2-master.3
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.1
Node Ver:         12.13.0
NPM Ver:          6.12.0
Java Ver:         10.0.2
Simulator:        Iphone 11 pro 13.1

PR #11319, opened 2019-11-07
fix(android): do not modify original ListView proxy

garymathews:
- Do not modify original ListView item proxy, which would affect every other row

##### TEST CASE
```JS
    templates: {
        template: {
            type: 'Ti.UI.View',
            childTemplates: [
                {
                    type: 'Ti.UI.View',
                    bindId: 'view',
                    properties: {
                        height: 30,
                        backgroundColor: 'white'
                    },
                    childTemplates: [
                        {
                            type: 'Ti.UI.Label',
                            bindId: 'label',
                            properties: {
                                color: 'black'
                            }
                        }
                    ]
                }
            ]
        }
    },
});
 
const items = [];
for (let i = 0; i < 18; i++) {
    items.push({
        template: 'template',
        label: {
            text: `Item #${i}`
        }
    });
}
items.push({
    template: 'template',
    label: {
        text: `Item #19`
    },
    view: {
        backgroundColor: 'red'
    }
});
listView.sections = [ Ti.UI.createListSection({ items }) ];
 
win.add(listView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27496)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
```diff
@@ -53,6 +54,11 @@ private void fireItemClick(String event, Object data)
 		if (event.equals(TiC.EVENT_CLICK) && data instanceof HashMap) {
 			KrollDict eventData = new KrollDict((HashMap) data);
 			TiViewProxy source = (TiViewProxy) eventData.get(TiC.EVENT_PROPERTY_SOURCE);
+
+			// FIXME: We should not need to create a placeholder proxy for each item. ListView needs refactoring to remedy this.
+			source = (TiViewProxy) KrollProxy.createProxy(source.getClass(), source.getKrollObject(), new Object[0],
+														  source.getCreationUrl().url);
+			eventData.put(TiC.EVENT_PROPERTY_SOURCE, source);
```
jquick-axway: We're not checking if `source` is `null` before calling `source.getClass()` and its other methods. Should we move this code into the if-block below? (Will the `!source.equals(this)` below be an issue?)

Also, the `createProxy()` method will call the given `KrollObject`'s `setProxySupport()` method, which replaces the previous proxy object's listeners. But this "might" be okay in this case since I don't think you set "click" listeners on it anyways.
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/java/org/appcelerator/kroll/KrollObject.java#L42

lokeshchdhry:
FR Passed.

SDK Ver:          9.0.0 local build
OS Ver:           10.14.5
Xcode Ver:        Xcode 11.1
Appc NPM:         4.2.15
Appc CLI:         7.1.2-master.5
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.2
Node Ver:         12.13.0
NPM Ver:          6.12.0
Java Ver:         10.0.2
Android Devices:  ⇨ google Pixel (Android 9)
Android Emulator: ⇨ Android 4.4.2

PR #11335, opened 2019-11-14
docs(changelog): update script to include bundled module versions

sgtcoolguy:
After manually preparing the 8.3.0 SDK RC release notes, I found another spot to improve generation - the bundled modules table listing.

So now when we generate the changelog we also include the listing of pre-packaged modules and the versions for each of Android and iOS.

### DIFF
--- a/changelog/config.js
+++ b/changelog/config.js
@@ -92,6 +92,43 @@ function guessPreviousBranch(version) {
 	// ideally it should be like "v1.2.3", but we're doing like '8_1_1_GA' so far
 }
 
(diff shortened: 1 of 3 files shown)


ewanharris (commented):
Looks good! Should the `commonjs` and `hyperloop` sections also include Windows?

sgtcoolguy:
@ewanharris The release notes typically didn't mention Windows SDK in the module table, so I didn't include them here. I think eventually we may want to, but like how we treat ti.map would be hard since there ti.map is actually not separated to a module.

PR #11336, opened 2019-11-14
build(node): drop node 8, support node 10.13+, node 12 

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27605


cb1kenobi (approved):
Yes!

package.json
```diff
@@ -153,10 +153,10 @@
     "visual studio": ">=10 <=12",
     "windows phone sdk": "8.x",
     "msbuild": ">=4.0",
-    "node": "8.x || 10.x"
+    "node": "10.x || 12.x"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=10.13"
```
cb1kenobi: Just out of curiosity is there a technical reason 10.12 won't work? Could/should this be `10.x`? I know the daemon requires specifically 10.2.0 or newer, but what is it with 10.13.0?

sgtcoolguy: Spoke with @mukherjee2 about this on Teams. My understanding is that the daemon requires 10.2+, but we might as well use Node's own LTS definition. 10.13 was the first official LTS release on the Node 10.x line.

cb1kenobi: OK, fine by me.

PR #11339, opened 2019-11-14
refactor(android): build with gradle

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-19300
- https://jira.appcelerator.org/browse/TIMOB-24517
- https://jira.appcelerator.org/browse/TIMOB-25240
- https://jira.appcelerator.org/browse/TIMOB-27325

**Hyperloop Warning:**
This PR will temporarily break hyperloop. Once this PR has been merged, we can then rebuild hyperloop module with gradle via the following private hyperloop PR...
https://github.com/appcelerator/hyperloop.next/pull/329.

**App Build Time Improvements:**
Benchmarked with [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) on a 15" MacBook Pro 2015 with a solid state drive.
```
Non-encrypted emulator builds:
- Clean Build: 52s -> 20s  (2.6x faster)
- Incremental: 18s -> 7s   (2.6x faster)

Encrypted device/production builds:
- Clean Build: 53s -> 21s  (2.5x faster)
- Incremental: 50s -> 14s  (3.6x faster)

Notes:
- Benchmarks exclude gradle daemon startup of ~5s which is a one time hit.
- Titanium did not support encrypted incremental builds before. This is new.
- App gradle builds will be faster once we rebuild modules as AARs.
```

**Summary of Changes:**
- Replaced "ant" tool usage with "gradle" for SDK, app, and module builds.
- Updated min Android build-tools requirement from 25.0.0 to 28.0.3.
- Titanium SDK Library:
  * Now built as a single AAR instead of individual JARs.
  * Can now easily open `titanium_mobile/android` directory in Android Studio.
  * Replaced `titanium_mobile/android` project `TitaniumTest` with `app`.
  * Supports building debug version of C/C++ `.so` libraries for asserts and debug logging.
  * Removed all Google JARs and res files and replaced with gradle/maven references.
  * Transpile/polyfill/rollup/snapshot of "ti.main.js" common folder is now done as a pre-build step of "titanium" library. Supports incremental builds when JS files change.
  * Jenkins/node-scons builds will now fail if unable to acquire V8 snapshot.
- Apps:
  * Significantly improved build times. Now supports incremental device/production builds.
  * Now automatically digitally signs APK with signature scheme v2.
  * Now merges module/app "AndroidManifest.xml" files via gradle.
  * Now supports "AndroidManifest.xml" merge `tools:` attributes. See [link](https://developer.android.com/studio/build/manifest-merge)
  * Added support for `./platform/android/build.gradle` file for defining dependencies.
  * Build no longer post-fixes hashes to app's `res/drawable` image files. Now stored as-is.
  * Building for "development" or "test" now builds a debug version. Can be used to debug modules and SDK via Android Studio.
  * Added new build hook event `build.android.requestResourcesDirPaths`. To be used by plugins/modules (such as hyperloop) to add additional "Resources" files to app build.
- Modules:
  * Are now built as AARs and installed as local maven repos for version management.
  * Modules built with 9.0.0 are not backward compatible. Must set "minsdk" to 9.0.0.
  * Now supports "Resources" folder whose files are copied into app as-is (like iOS).
  * Added support for `./platform/android/AndroidManifest.xml` file.
- Dropped support for appc config option `android.mergeCustomAndroidManifest`.
  * Must use  XML `tools:remove` or `tools:removeAll` attributes instead which is Google's equivalent feature.

---
**KitchenSink Test:**
_Test the below on Mac and Windows._
1. Download the [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) app.
2. Change its "tiapp.xml" file to reference this PR's SDK version.
3. Build and run to the Android emulator.
4. Verify app's top and bottom bars are RED. _(Indicates custom theme XML file works.)_
5. Tap through all of the app's UI, verify that everything works. _Particular Maps._
6. Build and run to the emulator again. (Tests increment build.)
7. Tap through all of the app's UI again.
8. Run it on emulator via Appc Studio debugger. Verify break-points work.
9. Run it on emulator via Appc Studio Live-View feature. Verify JS changes re-launches app.
10. Build and run to an Android device. (Tests encrypted JS builds.)
11. Tap through all of the app's UI again.
12. Build and run to the device again. (Tests incremental encrypted builds.)
13. Tap through all of the app's UI again.
14. Run it on device via Appc Studio debugger. Verify break-points work.
15. Run it on device via Appc Studio Live-View feature. Verify JS changes re-launches app.
16. Build app for production. (This requires a keystore file for signing app.)
17. Uninstall app from device/emultor.
18. Install and run production app on device/emulator.
19. Tap through all of the app's UI again.

---
**APK Signature Scheme v2 Test:**
1. Following the test procedure attached to [TIMOB-24517](https://jira.appcelerator.org/browse/TIMOB-24517).
2. Verify command line tool returns `true` for both v1 and v2 signatures.
_(Note that v3 should be `false`. Not currently supported.)_

---
**"AndroidManifest.xml" Error Handling Test:**
1. Follow test procedure attached to [TIMOB-27325](https://jira.appcelerator.org/browse/TIMOB-27325).
2. Build the app.
3. Verify that app build fails and it logs what's wrong with the XML. _(This is a good thing.)_

---
**Custom "build.gradle" Test:**
1. Create a Classic app.
2. Copy the below `app.js` to the project.
3. Copy the below `build.gradle` file to project's `./platform/android` folder.
4. Build and run on Android.
5. Verify black "play" button is shown onscreen. _(Comes from gradle file's referenced library.)_
6. In build log, verify you see message: `@@@ [build.gradle] JavaCompile.doFirst`

./Resources/app.js
```javascript
	backgroundColor: "white",
});
window.add(Ti.UI.createImageView({
	image: "android.resource://" + Ti.App.id + "/drawable/exo_notification_small_icon"
}));
window.open();
```

./platform/android/build.gradle
```groovy
repositories {
	google()
	jcenter()
}
tasks.withType(JavaCompile) {
	doFirst {
		println('@@@ [build.gradle] JavaCompile.doFirst')
	}
}
dependencies {
	implementation 'com.google.android.exoplayer:exoplayer-ui:2.8.0'
}
```


sgtcoolguy:
So to review what we discussed on Teams/call. There's a mismatch in that the app's build.gradle file is specifying a dependency on `org.appcelerator:titanium:9.0.0` but the AAR is actually versioned something like `9.0.0v201911191234567`. The options are to:
- update the generation of the app's build.gradle to inject the proper version (using something like `this.titaniumSdkName` in place of `this.titaniumSdkVersion`). If we go down this road, the car has the very specific version, but we'd likely have to alter our release scripts to manipulate that version in the AAR/POM/etc to the short GA name (like 9.0.0.GA) since our release takes an existing built sdk zip and alters the internal folder name now.
- generate AAR with the "base" version without the tag/qualifier on it. (i.e. 9.0.0). The AAR won't have the fully qualified version number in it, but then I don't think we'd need to alter any of the AAR/POM/etc when we release.

lokeshchdhry:
@jquick-axway , When you get a chance can you please resolve the conflicts. I will start with the FR for the current changes we have.

android/titanium/prebuild.js
```diff
@@ -0,0 +1,167 @@
...
+	for (let nextPath of filePaths) {
+		commandLine += ' ' + quotePath(nextPath);
+	}
+	return exec(commandLine);
+}
+
+/** Executes the pre-build step. */
+async function main() {
+	// Run the following operations in parallel.
+	return Promise.all([
```
garymathews: Missing `await`
```JS
return await Promise.all([
```

android/titanium/prebuild.js
```diff
@@ -0,0 +1,167 @@
...
+		'*/src/js/*.js',
+		{
+			cwd: path.join(__dirname, '..', 'modules'),
+			realpath: true
+		}
+	));
+
+	// Generate a "KrollJS.h" file containing all of the above fetched JS files.
+	// TODO: Replace our usage of python to make it easier to build on Windows.
+	let commandLine = 'python';
```
garymathews: Minor, but string manipulation could be reduced by doing something similar:
```JS
let args = [
	'python',
	quotePath(path.join(runtimeV8DirPath, 'tools', 'js2c.py')),
	quotePath(path.join(runtimeV8DirPath, 'generated', 'KrollJS.h')),
	quotePath(path.join(runtimeV8DirPath, 'generated', 'bootstrap.js'))
];
filePaths.forEach(path => args.push(quotePath(path)));

return exec(args.join(' '));
```

android/titanium/prebuild.js
garymathews: Actually, this is correct. But for some reason my `KrollJS.h` wasn't being generated until I added an `await` here...?

jquick-axway: I think I'll go ahead and change it anyways. From a style standpoint, I think it makes sense for an async function to return the resulting value (or no value) instead of a promise. And if you want a function to return a promise, then the function should not use the async keyword.

android/titanium/prebuild.js
jquick-axway: I agree. Done.

android/titanium/libv8-services.js
```diff
@@ -0,0 +1,288 @@
...
+	const filePath = path.join(__dirname, '..', 'package.json');
+	const fileContent = await fs.readFile(filePath, 'utf8');
+	return JSON.parse(fileContent);
+}
+
+/**
+ * Does a transpile/polyfill/rollup of our "titanium_mobile/common/Resources" JS files.
+ * Will then generate a C++ header file providing a V8 snapshot of the rolled-up JS for fast startup times.
+ */
+async function createSnapshot() {
```
garymathews: This could be simplified a little more:
```JS
async function createSnapshot() {
	// Perform a transpile/polyfill/rollup of our common JS files.
	// This contains our "ti.main.js" which the app executes on startup before executing the "app.js" file.
	const rollupOutputDirPath = path.join(__dirname, 'build', 'outputs', 'ti-assets', 'Resources');
	const rollupOutputFilePath = path.join(rollupOutputDirPath, 'ti.main.js');
	await fs.ensureDir(rollupOutputDirPath);
	const mainBuilder = new Builder({ args: [ 'android' ] });
	const androidBuilder = new AndroidBuilder({});
	await mainBuilder.transpile('android', androidBuilder.babelOptions, rollupOutputFilePath);
	const rollupFileContent = (await fs.readFile(rollupOutputFilePath)).toString();

	// Create the C++ directory we'll be generating the snapshot header file to.
	const cppOutputDirPath = path.join(__dirname, '..', 'runtime', 'v8', 'generated');
	const v8SnapshotHeaderFilePath = path.join(cppOutputDirPath, 'V8Snapshots.h');
	await fs.ensureDir(cppOutputDirPath);

	// Requests our server to create snapshot of rolled-up "ti.main" in a C++ header file.
	try {
		const MAX_ATTEMPTS = 20; // Allow for two minutes.

		// Post rolled-up "ti.main" script to server and obtain a snapshot ID as a response.
		// We will send an HTTP request for the snapshot code later.
		console.log('Attempting to request snapshot...');
		const snapshotUrl = 'http://v8-snapshot.appcelerator.com';
		const packageJsonData = await loadPackageJson();
		const requestOptions = {
			body: {
				v8: packageJsonData.v8.version,
				script: rollupFileContent
			},
			json: true
		};

		// Request `snapshotId` from server.
		const snapshotId = await request.post(snapshotUrl, requestOptions);
		for (let attempts = 0; attempts <= MAX_ATTEMPTS; attempts++) {

			// Request generated snapshot from server, if done.
			const response = await request.get(`${snapshotUrl}/snapshot/${snapshotId}`, {
				simple: false,
				resolveWithFullResponse: true
			});

			// Validate status codes.
			if (response.statusCode === 200) {

				// Server has finished creating a C++ header file containing all V8 snapshots.
				// Write it to file and flag that we're done.
				console.log('Writing snapshot...');
				await fs.writeFile(v8SnapshotHeaderFilePath, response.body);

			} else if (statusCode === 202 && attempts < MAX_ATTEMPTS) {

				// Snapshot server is still building. We need to retry later.
				console.log('Waiting for snapshot generation...');
				await new Promise(resolve => setTimeout(resolve, 6000));
				continue;
			}

			// Give up if received an unexpected response.
			console.error('Could not generate snapshot, skipping...');
			break;
		}
	} catch(e) {
		console.error(`Failed to request snapshot: ${e}`);

		// Generate an empty C++ header if the above failed.
		// Note: The C++ build will fail if file is missing. This is because it is #included in our code.
		await fs.writeFile(v8SnapshotHeaderFilePath, '// Failed to build V8 snapshots. See build log.');
	}
}
```

Jenkinsfile
```diff
@@ -326,6 +326,14 @@ timestamps {
 
 					// TODO parallelize the iOS/Android/Windows portions?
 					ansiColor('xterm') {
+						timeout(15) {
+							def buildCommand = "npm run clean -- --android-ndk ${env.ANDROID_NDK_R16B} --android-sdk ${env.ANDROID_SDK}"
+							if (includeWindows) {
+								buildCommand += ' --all'
+							}
+							sh buildCommand
+							recordIssues(tools: [clang(), java()])
```
garymathews: I believe the build is failing due to the:
```Jenkinsfile
recordIssues(tools: [clang(), java()])
```
Removing this line should resolve the issue.

android/titanium/libv8-services.js
jquick-axway: Thanks. I agree. I've implemented this... but with some alterations.

I've added a new environment variable `TI_SDK_BUILD_REQUIRES_V8_SNAPSHOTS`, which when set to `1`, will purposely fail the build if we were unable to generate the snapshot. I'm setting this in our `titanium_mobile/build` script. The idea being we don't want Jenkins to accidentally release an SDK build without V8 snapshot support for fast app startup times. But I think our fallback mechanism is okay for local Android Studio builds; especially when building offline.

lokeshchdhry:
**Please do not merge until Win android FR is done.**

FR Passed for MAC android. 

**Ran below checks using Studio & CLI**

- Ran tests mentioned above.

- KS build & runs fine.

- Ran tests in smoke test & they look good.

- Checked creation & build of default alloy & classic apps & it looks good.

- Checked `min-sdk` & `target-sdk` in tiapp.xml & works as expected.

- TIMOB-24517 - Works as expected.

- TIMOB-27325 - Works as expected. 
If `<uses-library>` is inside manifest it will throw error now:
```
[ERROR] :  [GRADLE] FAILURE: Build failed with an exception.
[ERROR] :  [GRADLE] 
[ERROR] :  [GRADLE] * What went wrong:
[ERROR] :  [GRADLE] Execution failed for task ':app:processDebugResources'.
[ERROR] :  [GRADLE] > Android resource linking failed
[ERROR] :  [GRADLE]   /Users/lchoudhary/Desktop/workspaces/workspace_2019/ZZZgradleclassic/build/android/app/build/intermediates/merged_manifests/debug/AndroidManifest.xml:11: AAPT: error: unexpected element <uses-library> found in <manifest>.
[ERROR] :  [GRADLE]       
[ERROR] :  [GRADLE] 
[ERROR] :  [GRADLE] * Try:
[ERROR] :  [GRADLE] Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[ERROR] :  [GRADLE] 
[ERROR] :  [GRADLE] * Get more help at https://help.gradle.org
[ERROR] :  [GRADLE] 
[ERROR] :  [GRADLE] BUILD FAILED in 3s
``` 

- TIMOB-25240 - Incremental builds are significantly faster now.

- Checked if modules can be created & build successfully with gradle via CLI & Studio. Looks good via CLI found an issue with studio: https://jira.appcelerator.org/browse/TISTUD-9204.

- Checked if old modules built with non gradle SDK  i.e <9.0 work with 9.0 for backward compatibility & they work fine.

- Checked modules created with 9.0 have `min-sdk` as 9.0 & looks good.

I will add more if I remember something later .......

sgtcoolguy:
@jquick-axway When this PR is eventually ready to merge, what is your expectation for the merge itself? i.e. do you plan to do a lot of rebasing locally to clean it all up and rebase on top of tip of `master` branch? Should it all be squashed into a single merge commit? Because I think 44 commits is a bit much. 

On larger PRs I tend to push commits until the CR is done and then manually rebase interactively (locally) to squash down to a smaller more coherent set of commits that could be rebased onto master.

jquick-axway:
@sgtcoolguy, I'd rather we merge it as-is. I'm not a fan of squashing commits... unless the prior commits were horribly broken (in this case they're not). In fact, I purposely split some of my changes into separate commits so that they can be easily understood from a history/blame perspective.

ssekhri:
FR Passed on windows machine as well running both android device and emulators. 
Performed the tests as mentioned by @lokeshchdhry for mac testing.

drauggres:
@lokeshchdhry 
cc @jquick-axway

FYI `kitchensink-v2` is failed to build on linux:
> Error: ENOENT: no such file or directory, open '<TI_SDK_PATH>/modules/android/facebook/8.0.0/facebook.jar'

because the file name is actually `Facebook.jar`

jquick-axway:
@drauggres, yeah, I see the issue. Thanks. From looking at the Titanium `8_3_X` branch via blame, it looks like this has been an issue for over 2 years. I can see our build script doing a `toLowerCase()` on the JAR filename in the link below (and I blindly followed it).
https://github.com/appcelerator/titanium_mobile/blob/243afd00e0760f2060e797312942ee65d47b9f5f/android/cli/commands/_build.js#L1575

The difference was that the old build script would log the following warning and carry on. The negative impact being that the Facebook APIs won't work since it'll fail to include the JAR. (This is bad.)
```
[WARN]  Module facebook version 8.0.0 does not have a main jar file
```

Our new build script rightfully logs it as an error now. I'll right this up to be fixed. Thanks again.

PR #11343, opened 2019-11-18
chore(ios): drop xcode 8 support

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27497


sgtcoolguy:
Pretty important to note that this also bumps minimum iOS from 9 to 10.

ewanharris:
@sgtcoolguy, I thought the `minIosVersion` dictated that and the `ios sdk` entry dictated the minimum compiling SDK? (I'm sure that one might be being used incorrectly though somewhere). I'll build this PR locally and build to an iOS 9 device to see what happens

sgtcoolguy:
Ok, so we have two separate "minimums" in the package.json. The one bumped here (in `vendorDependencies`) indicates "the minimum supported iOS SDK required when building" according to our CLI code. It's also used by the node-titanium-sdk transpile target as the minimum to transpile for.

The other `minIosVersion` is also used by the Cli, but honestly it's hard to tell what the difference here is/should be. Seems to be related to code selecting xcodes/sims/target sdk version? There's a whole section of the code checking this value: https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L2115-L2160

Maybe @cb1kenobi has a better handle on the difference between the two (or why we have 2 values that more or less seem to denote min iOS sdk versions)?

ewanharris:
Just confirming that this PR works fine building to an iOS 9 device, I think the transpile target should be updated to use [this.minIosVersion](https://github.com/appcelerator/titanium_mobile/blob/33ea2fb0bc35f75a67674e8ba939538b12c7fe64/iphone/cli/commands/_build.js#L1458) which correctly calculates the minimum iOS version for an app based off `minIosVersion` from `package.json` and the `min-ios-ver` tag from the tiapp, not the iOS SDK which has no bearing on the JS Engine that will be targeted

vijaysingh-axway:
Yes. This will not change minimum iOS target. cc @cb1kenobi

janvennemann (approved):
The iOS changes are looking good, approved!

BTW, i noticed it too that we use the wrong value for the preset-env option while working on the Webpack tooling. The appcd plugin already uses the correct value:

https://github.com/appcelerator/appcd-plugin-webpack/blob/597b2981ae05bfb1c0d95f259a4293a12fecb3f8/src/utils/webpack.js#L93

// EDIT: I created https://jira.appcelerator.org/browse/TIMOB-27630 for the update in the SDK

ewanharris (approved):
Seeing as the babel setup fix was split out to TIMOB-27630, this PR looks good to me

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.1, 10.14.6
SDK: 9.0.0.v20191118123859
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3

PR #11346, opened 2019-11-19
feat: webpack support

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27429

This adds a new CLI hook that will allow the use of Webpack with Titanium.

## Motivation

The current JavaScript processing pipeline in Titanium is mainly done in [node-titanium-sdk](https://github.com/appcelerator/node-titanium-sdk). Although we improved the implementation a lot in the last couple major releases, it still lacks the ability to be customized by users. The only configuration that can be done is to include additional Babel plugins.

With our continuing effort to improve compatibility with Node.js core modules the use of third-party Node modules should also be simplified. Depending on the project type, currently they have to be installed into `Resources/node_modules` (Classic) or `app/lib/node_modules` (Alloy). This is not very intuitive, since users are used to installing Node modules directly into the project root.

Last but not least, to support Angular and Vue.js implementations a Webpack build pipeline needs to be in place. The tooling around those is depending on Webpack.

## Features

To address the issues outlined above an alternative JavaScript build pipeline using Webpack is now in place. It comes with pre-configured configurations for existing Classic and Alloy projects, as well as for the upcoming Vue.js and Angular implementations (experimental). Webpack support is implemented as a [plugin](https://github.com/appcelerator/appcd-plugin-webpack) for the [Appc Daemon](https://github.com/appcelerator/appc-daemon).

Features for the initial release:

- Bundle source code and assets of apps using Webpack
- Pre-defined Webpack configurations for all Titanium projects. These should work out of the box for most projects but are fully configurable using hooks/plugins if need be.
- Webpack builds are performed in a background process and automatically re-build changed files to minimize incremental app build times.
- Easy use of Node modules. Install NPM packages directly into the project root directory and require them in Titanium code. Webpack then takes care of the rest and makes sure to properly resolve and bundle them into the app.
- Web UI to view build stats and control the Webpack builds.

## Migration

There is a detailed [migration guide](https://github.com/appcelerator/appcd-plugin-webpack/blob/master/migration.md) for existing projects. It describes all steps required to enable Webpack in existing projects and what changes are required for the project. Changes to the code base should be minimal and only affect `require`/`import` statements in most cases.

## How it works

The hook communicates with the Appc Daemon to start new Weback builds and query the current status of running builds. The majority of the work is then done by the daemon in the background. [appcd-plugin-webpack](https://github.com/appcelerator/appcd-plugin-webpack) will automatically configure Webpack based on the project type so users don't have to deal with Webpack configuration files.

Webpack uses `Resources` as its output directory so our build can continue as usual once Webpack is done. This ensures minimal changes to the existing build pipeline.

To get the most out of the "always running" approach of the daemon, once Webpack is started it keeps running for as long as you actively work on a project. After 10min of no activity it stops to save system resources. On the the next build it will automatically be started again. It is also possible to start and stop the Webpack build via the Web UI.

The Webpack plugin will make sure to restart the build when necessary, for example when

- Build settings change, like target or deploy type.
- Files change, like `tiapp.xml` or `babel.config.js`
- The `--force` flag is explicitly set

As a result, there are two kinds of clean builds when Webpack is enabled. The first one when Webpack is not started, which is basically the same as usual. Everything starts from scratch, including JS processing through Webpack. And all builds from then on will just query the daemon for the current Webpack build status. If everything is up-to-date the whole JS handling in our build will be skipped. This means that even subsequent clean builds are faster.

## Benchmarks

I currently only have benchmarks for Alloy apps since i don't have a decently large Classic project to test with. Anyway, let's talk some numbers. You will notice that there are two passes for clean builds. The first one is with Webpack not running, the second one with Webpack already running. All times are the average of three builds.

Timings with our very own KitchenSink v2:

|Task|Standard|Webpack|
|---|---|---|
|Clean build `#1`|32s|36s|
|Clean build `#2`|32s|30s|
|Incremental build|8s|789ms|

The first clean build where Webpack is not already running takes a little longer since there is the overhead of starting Webpack, which can take a couple of seconds.

Now, the bigger the app gets, the better the build time improvements become. The following benchmarks were done with a large real-world project that contains ~100 controllers/views, ~20 widgets and NPM dependencies.

|Task|Standard|Webpack|
|---|---|---|
|Clean build `#1`|2m 6s|1m 55s|
|Clean build `#2`|2m 6s|1m 36s|
|Incremental build|20s 965ms|1s 861ms|

## Action items

- [x] ~~The hooks tries to locate the appcd daemon either globally when run via `ti build` or the shipped version included in appc-cli when run with `appc run`. Depending on whether the appc-cli daemon can be updated to v3.1.0 or not this behavior needs to be adjusted.~~ `appc-cli@8.0.0`  ships with `appcd@3.1.0`!
- [x] Evaluate Webpack generated source maps.
- [x] Update `node-titanium-sdk` once https://github.com/appcelerator/node-titanium-sdk/pull/128 is merged.
- [ ] Create new project templates that have Webpack enabled by default.

## Testing Instructions

I've created a new branch of our Kitchensink App which can be used for easy testing: https://github.com/janvennemann/kitchensink-v2/tree/webpack

To enable Webpack in any other project follow the [migration guide](https://github.com/appcelerator/appcd-plugin-webpack/blob/develop/migration.md).

## Kown issues

- Webpack builds are not yet compatible with Hyperloop. All Hyperoop related requires need to be marked as external so Webpack doesn't try to bundle them

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -1530,8 +1530,8 @@ AndroidBuilder.prototype.run = async function run(logger, config, cli, finished)
 		await this.checkIfNeedToRecompile();
 
 		// Notify plugins that we're prepping to compile.
-		await new Promise((resolve) => {
-			cli.emit('build.pre.compile', this, resolve);
+		await new Promise((resolve, reject) => {
+			cli.emit('build.pre.compile', this, e => (e ? reject(e) : resolve()));
 		});
 
 		// Make sure we have an "app.js" script. Will exit with a build failure if not found.
@@ -2874,7 +2874,16 @@ AndroidBuilder.prototype.copyResources = function copyResources(next) {
 				});
 			})
 			.then(() => {
-				this.tiSymbols = task.data.tiSymbols;
+				if (this.useWebpack) {
+					// Merge Ti symbols from Webpack with the ones from legacy js processing
+					Object.keys(task.data.tiSymbols).forEach(file => {
+						const existingSymbols = this.tiSymbols[file] || [];
+						const additionalSymbols = task.data.tiSymbols[file];
+						this.tiSymbols[file] = Array.from(new Set(existingSymbols.concat(additionalSymbols)));
+					});
+				} else {
+					this.tiSymbols = task.data.tiSymbols;
+				}
 
 				// Copy all unprocessed files to "app" project's APK "assets" directory.
 				appc.async.parallel(this, copyUnmodified.map(relPath => {
--- a/cli/hooks/webpack.js
+++ b/cli/hooks/webpack.js
@@ -1,0 +1,177 @@
+// eslint-disable-next-line security/detect-child-process
+const { execSync } = require('child_process');
+const fs = require('fs-extra');
+const path = require('path');
+const semver = require('semver');
+const util = require('util');
+require('colors');
+
+const appcd = require('../lib/webpack/appcd');
+const WebpackService = require('../lib/webpack/service');
+
+const MIN_APPCD_VERSION = '3.2.0';
+
+const createWebpackLogger = (logger) => {
+	const webpackLogger = {
+		formatWithBadge(args) {
+			return `${' WEBPACK '.bgCyan.black} ${util.format(...args)}`;
+		}
+	};
+	[ 'info', 'debug', 'warn', 'error' ].forEach(level => {
+		webpackLogger[level] = (...args) => logger[level](webpackLogger.formatWithBadge(args));
+	});
+	return webpackLogger;
+};
+
+exports.id = 'ti.webpack';
+exports.init = (logger, config, cli) => {
+	const isAppcCli = typeof process.env.APPC_ENV !== 'undefined';
+	let commandName;
+	let isWebpackEnabled = false;
+	let projectType;
+	let client;
+
+	cli.on('cli:command-loaded', (hookData) => {
+		const command = hookData.command;
+		commandName = command.name;
+	});
+
+	cli.on('cli:post-validate', () => {
+		projectType = getWebpackProjectType(cli.argv['project-dir']);
+		if (projectType === null) {
+			return;
+		}
+		isWebpackEnabled = true;
+		process.env.TI_USE_WEBPACK = true;
+
+		if (commandName !== 'build') {
+			return;
+		}
+
+		let appcdRootPath;
+		if (isAppcCli) {
+			// we were invoked by appc-cli, load bundled daemon client.
+			appcdRootPath = resolveModuleRoot('appcd');
+			if (!fs.existsSync(appcdRootPath)) {
+				throw new Error('Unable to find Appcelerator Daemon inside the appc-cli. Please make sure to use a recent appc-cli version. You can install/select it with  "appc use latest".');
+			}
+		} else {
+			// plain ti-cli, load daemon from global modules
+			const globalModulesPath = execSync('npm root -g').toString().replace(/\s+$/, '');
+			appcdRootPath = resolveModuleRoot('appcd', [ globalModulesPath ]);
+			if (!fs.existsSync(appcdRootPath)) {
+				throw new Error('Unable to find global Appcelerator Daemon install. You can install it with "npm i appcd -g".');
+			}
+		}
+		ensureAppcdVersion(appcdRootPath, MIN_APPCD_VERSION, isAppcCli);
+		const AppcdClient = resolveAppcdClient(appcdRootPath);
+		const baseClient = new AppcdClient();
+		if (isAppcCli) {
+			baseClient.appcd = path.resolve(appcdRootPath, '..', '.bin', 'appcd');
+			if (process.platform === 'win32') {
+				baseClient.appcd += '.cmd';
+			}
+		}
+		client = appcd(baseClient);
+	});
+
+	cli.on('build.pre.compile', {
+		priority: 800,
+		post(builder, callback) {
+			if (!isWebpackEnabled) {
+				return callback();
+			}
+
+			builder.useWebpack = true;
+
+			const badgedLogger = createWebpackLogger(logger);
+			const webpackService = new WebpackService({
+				client,
+				logger: badgedLogger,
+				cli,
+				builder,
+				projectType
+			});
+			Promise.resolve().then(async () => {
+				await webpackService.build();
+				return callback();
+			}).catch(e => {
+				if (e.status === 404) {
+					badgedLogger.info('Daemon was unable to find the Webpack plugin. To continue you need to:');
+					badgedLogger.info('');
+					const installCommand = 'npm i -g @appcd/webpack-plugin';
+					badgedLogger.info(`- Install it with ${installCommand.cyan}`);
+					const restartCommand = `${isAppcCli ? 'appc ' : ''}appcd restart`;
+					badgedLogger.info(`- Restart the daemon with ${restartCommand.cyan}`);
+					badgedLogger.info('');
+				}
+
+				if (cli.argv.platform === 'ios') {
+					// The iOS build shows the error message only, log the stack manually
+					// to make sure it get printed as well
+					logger.error(e.stack);
+				}
+
+				callback(e);
+			});
+		}
+	});
+};
+
+function getWebpackProjectType(projectDir) {
+	if (typeof projectDir !== 'string') {
+		return null;
+	}
+	const pkgPath = path.join(projectDir, 'package.json');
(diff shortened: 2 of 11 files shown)


cli/hooks/webpack.js
```diff
@@ -0,0 +1,129 @@
...
+		if (process.env.APPC_ENV) {
+			// we were invoked by appc-cli, load bundled daemon client.
+			appcdRootPath = resolveModuleRoot('appcd');
+			if (!fs.existsSync(appcdRootPath)) {
+				throw new Error('Unable to find Appcelerator Daemon inside the appc-cli. Please make sure to use a recent appc-cli version. You can install/select it with  "appc use latest".');
+			}
+			ensureAppcdVersion(appcdRootPath, MIN_APPCD_VERSION, true);
+		} else {
+			// plain ti-cli, load daemon from global modules
+			const globalModulesPath = execSync('npm root -g').toString().replace(/\s+$/, '');
```
sgtcoolguy: this seems potentially problematic on windows...

cli/hooks/webpack.js
```diff
@@ -0,0 +1,129 @@
...
+
+function resolveModuleRoot(name, paths) {
+	try {
+		let resolvedPath;
+		if (paths) {
+			resolvedPath = require.resolve(name, { paths });
+		} else {
+			resolvedPath = require.resolve(name);
+		}
+		return resolvedPath.substr(0, resolvedPath.indexOf('/appcd/') + 7);
```
sgtcoolguy: Do we know that `require.resolve` returns paths with '/' separates on windows?

cli/hooks/webpack.js
```diff
@@ -0,0 +1,129 @@
...
+	} catch (e) {
+		return null;
+	}
+}
+
+function ensureAppcdVersion(appcdRootPath, version, isAppcCli = false) {
+	const pkg = JSON.parse(fs.readFileSync(path.join(appcdRootPath, 'package.json')));
```
sgtcoolguy: could potentially use fs-extra's readJSONSync or plain old require to automatically do the JSON parsing of the input file for you...

cli/lib/webpack/service.js
```diff
@@ -0,0 +1,226 @@
+const assert = require('assert');
+const crypto = require('crypto');
+const EventEmitter = require('events');
+
+const WebpackProgressReporter = require('./reporter');
+
+class WebpackService {
+	constructor(options) {
+		const { AppcdClient, builder, cli, logger } = options;
```
sgtcoolguy: I think you could put this destructuring into the function param declaration: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_fields_from_objects_passed_as_function_parameter

cli/lib/webpack/service.js
```diff
@@ -0,0 +1,226 @@
...
+				if (e.state === 'ready' || e.state === 'error') {
+					this.webpackStatusEmitter.off('status', handler);
+
+					if (e.state === 'ready') {
+						clearTimeout(buildTimeout);
+						if (!this.shouldWatch) {
+							this.client.disconnect();
+						}
+						return resolve();
+					} else {
```
sgtcoolguy: shouldn't need the `else` block/wrap if the `if` block always returns.

cli/lib/webpack/service.js
```diff
@@ -0,0 +1,226 @@
...
+
+			// @todo what to do in this case? Give useful tips!
+			const buildTimeout = setTimeout(() => reject(new Error('Did not receive success message from Webpack appcd plugin within 120s, aborting...')), 120000);
+		});
+	}
+
+	startWebpack(options) {
+		this.logger.debug(`Starting Webpack with options: ${JSON.stringify(options)}`);
+		this.webpackStatus = { state: 'building' };
+		return new Promise((resolve, reject) => {
```
sgtcoolguy: really not a fan of all this hand-rolling of Promises, but I'm not sure there's a nice way to move to `async`/`await` with events like this.

package.json
```diff
@@ -110,7 +110,7 @@
     "markdown": "0.5.0",
     "moment": "^2.22.2",
     "node-appc": "^0.3.4",
-    "node-titanium-sdk": "^4.0.0",
+    "node-titanium-sdk": "^4.2.0",
```
sgtcoolguy: this is outdated now...

cli/lib/webpack/service.js
```diff
@@ -0,0 +1,226 @@
...
+			'build'
+		];
+		const pluck = (obj, props) => {
+			return props.reduce((acc, prop) => {
+				acc[prop] = obj[prop];
+				return acc;
+			}, {});
+		};
+		try {
+			assert.deepStrictEqual(
```
sgtcoolguy: to avoid the need to try/catch you may want to switch to `util.isDeepStrictEqual()` which returns true/false rather than throws
https://nodejs.org/api/util.html#util_util_isdeepstrictequal_val1_val2

cli/lib/webpack/service.js
janvennemann: Yeah, me neither, but i don't think there is a nice way around this since `appcd-client` is event based.

cc @cb1kenobi What do you think about having an alternative promise based API for simple requests like these? Always having to write request like the following can get quite repetitive.

```js
client
  .request({
    path: '/appcd/status'
  })
  .once('response', onResponse)
  .once('error', onError);
```

I could imagine a promise based API with `get(path): Promise` and `post(path, data): Promise`, for example.

cli/lib/webpack/service.js
janvennemann: Yes, that's much better, thanks!

cli/hooks/webpack.js
janvennemann: Changed to `path.sep` just to be sure.

cli/lib/webpack/service.js
```diff
@@ -0,0 +1,239 @@
...
+				this.logger.info('Starting Appcelerator Daemon (subsequent builds will be faster) ...');
+				await appcd.connect({ startDaemon: true });
+			} else {
+				throw e;
+			}
+		}
+
+		const { host, port } = appcd.client;
+		this.webUiUrl = `http://${host}:${port}${this.basePath}/web`;
+		this.logger.info(`🖥  Web UI ready on ${this.webUiUrl.cyan}`);
```
cb1kenobi, line 78: FYI: One of the topics the CLI style guide team has on the table is the use of emojis. I believe at this point, they are discouraged, but likely to not be permitted.

janvennemann: Buuhhhh, but i wanted to be one of cool kids! JK, that's the only emoji i use so not that big of a deal to remove it.

cli/hooks/webpack.js
```diff
@@ -0,0 +1,165 @@
...
+			appcdRootPath = resolveModuleRoot('appcd', [ globalModulesPath ]);
+			if (!fs.existsSync(appcdRootPath)) {
+				throw new Error('Unable to find global Appcelerator Daemon install. You can install it with "npm i appcd -g".');
+			}
+		}
+		ensureAppcdVersion(appcdRootPath, MIN_APPCD_VERSION, isAppcCli);
+		const AppcdClient = resolveAppcdClient(appcdRootPath);
+		const baseClient = new AppcdClient();
+		if (isAppcCli) {
+			baseClient.appcd = path.resolve(appcdRootPath, '..', '.bin', 'appcd');
```
ewanharris, line 70: @janvennemann This will need to be `appcd.cmd` when running on Windows as the appcd binary will fail to start (just confirmed with @ssjsamir)

android/cli/commands/_build.js
```diff
@@ -1530,8 +1530,8 @@ AndroidBuilder.prototype.run = async function run(logger, config, cli, finished)
 		await this.checkIfNeedToRecompile();
 
 		// Notify plugins that we're prepping to compile.
-		await new Promise((resolve) => {
-			cli.emit('build.pre.compile', this, resolve);
+		await new Promise((resolve, reject) => {
+			cli.emit('build.pre.compile', this, e => (e === null ? resolve() : reject(e)));
```
cb1kenobi, line 1534: Not saying you have to change this, but I always do `e => e ? reject(e) : resolve()` just in case `e` not null, but still falsey.

cli/hooks/webpack.js
janvennemann: `execSync` starts a new shell on windows and directly passes the command to that shell, so this shouldn't be an issue. I tested this on my Window 10 machine and works, but i'll ask QE to double check.

PR #11348, opened 2019-11-19
docs: fix PickerColumn parent class

drauggres:
Ti.UI.Picker:
- extends Ti.UI.View which extends Ti.Proxy and none of them have `enabled` property

Ti.UI.PickerColumn:
- is really extends Ti.UI.View (at least on [android](https://github.com/appcelerator/titanium_mobile/blob/33ea2fb0bc35f75a67674e8ba939538b12c7fe64/android/modules/ui/src/java/ti/modules/titanium/ui/PickerColumnProxy.java#L25))
- extending it from `Ti.Proxy` is [breaking typing](https://gist.github.com/drauggres/77a0bd495ca3e7d4a7dbd7e170156fb6#file-titanium-d-ts-L6995) for `Ti.UI.Picker` method `add`:

```typescript
declare namespace Titanium {

    namespace UI {

        class View extends Titanium.Proxy {
            // ...
            add(view: Titanium.UI.View|Array<Titanium.UI.View>): void;
        }

        class Picker extends Titanium.UI.View {
            // ...
            add(data: Array<Titanium.UI.PickerRow>|Titanium.UI.PickerRow|Array<Titanium.UI.PickerColumn>|Titanium.UI.PickerColumn): void;
        }

        class PickerColumn extends Titanium.Proxy {
            // ...
        }

        // ...
    }
}
```
First argument of `add` method MUST be `Ti.UI.View` or inherited from it (or `Ti.UI.View[]`)

Extending `Ti.UI.PickerColumn` from `Ti.UI.View` and later excluding all it properties, methods and events might seem stupid, but it is the real state of affairs.

P.S. TypeScript declaration file is generated with [this](https://github.com/appcelerator/docs-devkit/pull/8).

### DIFF
--- a/apidoc/Titanium/UI/Picker.yml
+++ b/apidoc/Titanium/UI/Picker.yml
@@ -49,7 +49,7 @@ excludes:
     properties: [
         anchorPoint,backgroundDisabledColor,backgroundDisabledImage,backgroundFocusedColor,children,
         backgroundFocusedImage,backgroundGradient,backgroundImage,backgroundLeftCap,backgroundRepeat,
-        backgroundSelectedColor,backgroundSelectedImage,backgroundTopCap,center,enabled,layout,tintColor,zIndex]
+        backgroundSelectedColor,backgroundSelectedImage,backgroundTopCap,center,layout,tintColor,zIndex]
     methods: [remove, removeAllChildren, replaceAt]
 since: "0.8"
 platforms: [android, iphone, ipad]
--- a/apidoc/Titanium/UI/PickerColumn.yml
+++ b/apidoc/Titanium/UI/PickerColumn.yml
@@ -10,7 +10,25 @@ description: |
     On Android, the `useSpinner` property must be enabled to support multiple columns.
 
     See <Titanium.UI.Picker> for further examples of usage.
-extends: Titanium.Proxy
+extends: Titanium.UI.View
+excludes:
+  properties: [
+    accessibilityHidden, accessibilityHint, accessibilityLabel, accessibilityValue, animatedCenter,
+    anchorPoint, backgroundColor, backgroundDisabledColor, backgroundDisabledImage,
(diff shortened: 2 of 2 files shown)


sgtcoolguy:
The first change to remove `enabled` from the exclusion list for `Picker` looks fine to me.

The second change, I'm not enough of a TypeScript user to know about. 

In reality `PickerColumn` does extend from `Ti.UI.View` on Android, but extends from `Ti.Proxy` on iOS.

@ewanharris and @janvennemann perhaps you know? Basically Picker is trying to override `Ti.Ui.View`'s `add` method - and specify that a Picker can only have PickerRow or PickerColumns as children. It looks correct to me as it is generated in the typescript, it's specifying an add method with only those as possible arguments. But does it not properly "override" the `Ti.UI.View.add()` method that is more liberal and allows any `Ti.UI.View` children?

Do we need to pretend a `PickerColumn` is a child class of `Ti.UI.View` with almost all the properties/methods stripped off?

janvennemann:
Hm, since we are extending from different types on the native side, one approach to document it is obviously always the wrong one. To me it seems like this needs fixing on the native side.

I just checked with our current typings and there seems to be something off with the `add` method in general. I get four variations but all show up as either `add(view: any)` or `add(view: any[])`, which is not pretty useful. We may have to revisit how we declare the `add` method.

drauggres:
Also worth mentioning that `PickerColumn` was extended from `View` since the beginning until [this commit]( https://github.com/appcelerator/titanium_mobile/commit/b31c2afcc62dec4f98cf698125844133dc5d528f#diff-06211ff79fbedc8057110fba4d129b61).

drauggres:
> To me it seems like this needs fixing on the native side.

It would be better to have a different method on `Ti.UI.Picker` instead of `add` (e.g. `addColumn`).
But removing `add` will be a "breaking change".

sgtcoolguy (approved):

PR #11349, opened 2019-11-19
NPM Scripts

sgtcoolguy:
**Description:**
This adds a few more npm scripts to run portions of the build/package process (and makes use of them from Jenkins). I'd prefer to "hide" the scons cli stuff behind npm scripts in case we want to change it or rename it (since really I just kept the name around from a very, very old build tool we used!).

Also, as I'm going it's pretty nice to have them chained with one another.

Open to suggestions on naming, or trying to break apart the `test:ios` and `test:android` entries. I run them the most often, but it does a full clean and rebuild of everything, which can be very annoying if you're trying to just tweak the tests. Maybe just drop the clean call and see if we can't improve our iOS/android builds of the SDK to actually work incrementally?


janvennemann:
Looks good to me! I use the same naming scheme for NPM scripts as well, so i'm fine with that. However, breaking up the test and build commands would be really appreciated, especially when testing Android. The iOS build is quite fast so it doesn't have that much of an impact on test run times, but running the Android test command takes forever due to the slow build.

PR #11352, opened 2019-11-20
docs: add missing types for events properties

drauggres:
https://jira.appcelerator.org/browse/AC-6443

Documentation for some event properties does not include types.

### DIFF
--- a/apidoc/Titanium/App/App.yml
+++ b/apidoc/Titanium/App/App.yml
@@ -256,6 +256,7 @@ events:
     properties:
       - name: state
         summary: Proximity state value.
+        type: Boolean
     platforms: [iphone, android]
     since: {android: "3.3.0"}
 
@@ -267,31 +268,41 @@ events:
     properties:
       - name: message
         summary: The error message.
+        type: String
       - name: line
         summary: The line where the error occurred.
+        type: Number
       - name: sourceId
         summary: A unique identification for the source file.
+        type: Number
         platforms: [iphone, ipad]
       - name: type
         summary: The type of error.
+        type: String
         platforms: [iphone, ipad]
       - name: sourceURL
         summary: The URL to the source file.
+        type: String
         platforms: [iphone, ipad]
       - name: backtrace
         summary: The backtrace of function calls when the error occurred.
+        type: String
         platforms: [iphone, ipad]
       - name: title
         summary: The title for the error.
+        type: String
         platforms: [android]
       - name: sourceName
         summary: The name of the source file.
+        type: String
         platforms: [android]
       - name: lineSource
         summary: The line source reference.
+        type: String
         platforms: [android]
       - name: lineOffset
         summary: The offset on the line where the error occurred.
+        type: Number
         platforms: [android]
     platforms: [android, iphone, ipad]
     since: '4.1.0'
--- a/apidoc/Titanium/Geolocation/Geolocation.yml
+++ b/apidoc/Titanium/Geolocation/Geolocation.yml
@@ -307,6 +307,7 @@ events:
     properties:
       - name: code
         summary: If `success` is `false`, the error code is available.
+        type: Number
         platforms: [android, iphone, ipad]
 
       - name: success
--- a/apidoc/Titanium/Media/AudioPlayer.yml
+++ b/apidoc/Titanium/Media/AudioPlayer.yml
@@ -217,6 +217,7 @@ events:
         constants: Titanium.Media.AUDIO_STATE_*
       - name: description
         summary: Text description of the state of playback.
+        type: Number
 
   - name: complete
     summary: Fired when the audio has finished playing.
@@ -271,6 +272,7 @@ events:
     properties:
       - name: progress
         summary: Current progress, in milliseconds.
+        type: Number
         
   - name: seek
     summary: Fired once the [seekToTime](Titanium.Media.AudioPlayer.seek) method completes.
@@ -282,6 +284,8 @@ events:
             If the new request completes without being interrupted by another seek 
             request or by any other operation this event will be invoked with 
             the `finished` parameter set to `true`. 
+        type: Boolean
+    platforms: [iphone, ipad]
 properties:
   - name: STATE_BUFFERING
     summary: Audio data is being buffered from the network.
--- a/apidoc/Titanium/Network/BonjourBrowser.yml
+++ b/apidoc/Titanium/Network/BonjourBrowser.yml
@@ -29,6 +29,7 @@ events:
     properties:
       - name: services
         summary: An array of BonjourService objects corresponding to currently available services.  If you cache this value, including using it as table data, be aware that it could become out of date at any time due to the asynchronous nature of Bonjour service discovery.
+        type: Array
     since: '3.0.0'
 
 properties:
--- a/apidoc/Titanium/Network/BonjourBrowser.yml
+++ b/apidoc/Titanium/Network/BonjourBrowser.yml
@@ -29,7 +29,7 @@ events:
     properties:
       - name: services
         summary: An array of BonjourService objects corresponding to currently available services.  If you cache this value, including using it as table data, be aware that it could become out of date at any time due to the asynchronous nature of Bonjour service discovery.
-        type: Array
+        type: Array<Titanium.Network.BonjourService>
     since: '3.0.0'
 
 properties:
--- a/apidoc/Titanium/Network/TCPSocket.yml
+++ b/apidoc/Titanium/Network/TCPSocket.yml
@@ -33,8 +33,10 @@ events:
     properties:
       - name: from
         summary: the reference for the socket that data was retrieved from
+        type: Number
       - name: data
         summary: a blob representing the data read, can be interpreted via toString
+        type: Titanium.Blob
   - name: readError
     summary: an error occured when reading
     properties:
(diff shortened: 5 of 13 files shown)


apidoc/Titanium/Media/AudioPlayer.yml
```diff
@@ -282,6 +284,8 @@ events:
             If the new request completes without being interrupted by another seek 
             request or by any other operation this event will be invoked with 
             the `finished` parameter set to `true`. 
+        type: Boolean
+    platforms: [iphone, ipad]
```
drauggres: `seek` also is iOS-only event

apidoc/Titanium/Network/BonjourBrowser.yml
```diff
@@ -29,6 +29,7 @@ events:
     properties:
       - name: services
         summary: An array of BonjourService objects corresponding to currently available services.  If you cache this value, including using it as table data, be aware that it could become out of date at any time due to the asynchronous nature of Bonjour service discovery.
+        type: Array
```
drauggres: `Array<Titanium.Network.BonjourService>` ?

apidoc/Titanium/Network/TCPSocket.yml
```diff
@@ -33,8 +33,10 @@ events:
     properties:
       - name: from
         summary: the reference for the socket that data was retrieved from
+        type: Number
```
drauggres: Is it really `Number` or am I missing something (I have no experience in obj-c)?

apidoc/Titanium/App/App.yml
```diff
@@ -267,31 +268,41 @@ events:
     properties:
       - name: message
         summary: The error message.
+        type: String
       - name: line
         summary: The line where the error occurred.
+        type: Number
       - name: sourceId
         summary: A unique identification for the source file.
+        type: Number
```
drauggres: I could not find `sourceId` in the code or Apple docs. Does it really exists?

apidoc/Titanium/App/App.yml
```diff
@@ -267,31 +268,41 @@ events:
...
         platforms: [iphone, ipad]
       - name: type
         summary: The type of error.
+        type: String
```
drauggres: I could not find `type` in the code or Apple docs. Does it really exists?

apidoc/Titanium/App/App.yml
```diff
@@ -267,31 +268,41 @@ events:
...
         platforms: [android]
       - name: sourceName
         summary: The name of the source file.
+        type: String
         platforms: [android]
       - name: lineSource
         summary: The line source reference.
+        type: String
         platforms: [android]
       - name: lineOffset
```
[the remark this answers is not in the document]
sgtcoolguy: Ugh, looks like iOS uses `column` as the property name for what looks to me to be the exact same thing here (`lineOffset`).

apidoc/Titanium/App/App.yml
```diff
@@ -267,31 +268,41 @@ events:
...
         platforms: [iphone, ipad]
       - name: type
         summary: The type of error.
+        type: String
         platforms: [iphone, ipad]
       - name: sourceURL
         summary: The URL to the source file.
+        type: String
         platforms: [iphone, ipad]
       - name: backtrace
```
sgtcoolguy: I honestly don't see this anywhere... I don't think it exists.

apidoc/Titanium/App/App.yml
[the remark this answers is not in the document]
sgtcoolguy: Also, iOS has `stack` and `nativeStack` properties which are strings and are equivalent to Android's `javascriptStack` and `javaStack`. @garymathews I think we desperately need to clean this API up here.

We no longer user `backtrace` AFAICT. We should standardize on `stack` for the JS stack trace, mark `backtrace` removed (when?), mark `javascriptStack` deprecated. Same idea for standardizing on `nativeStack` over `javaStack` (probably deprecate `javaStack` to be removed in the future like SDK 10?)

apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -307,6 +307,7 @@ events:
     properties:
       - name: code
         summary: If `success` is `false`, the error code is available.
+        type: Number
```
sgtcoolguy: I'm honestly not even sure this event can be fired with an error...

apidoc/Titanium/Network/BonjourBrowser.yml
sgtcoolguy: Yes, it should be `Array<Titanium.Network.BonjourService>`

apidoc/Titanium/Network/TCPSocket.yml
sgtcoolguy: yes, it is.

apidoc/Titanium/UI/Slider.yml
```diff
@@ -126,6 +126,7 @@ events:
         summary: |
             Dictionary with properties `width` and `height` of the size of the thumb.
             Available with custom thumb image.
+        type: Dictionary
```
sgtcoolguy: Should probably define a simple type for this. It's just an Object with `width` and `height` properties, which are both `Number` type.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -497,6 +498,7 @@ events:
     summary: Fired when the web view content is loaded.
     properties:
       - name: url
+        type: String
         summary: URL of the web document.
```
sgtcoolguy: Ti.UI.WebView's `load` event has a property `title` on iOS, which is a `String` and is "The page title."

apidoc/Titanium/UI/WebView.yml
```diff
@@ -530,6 +532,7 @@ events:
     properties:
       - name: url
         summary: The URL of the web document that is stopped.
+        type: String
```
sgtcoolguy: There's also a hard-coded `message` String property. I think it always returns `"Webview did not load blacklisted url."`

apidoc/Titanium/UI/WebView.yml
```diff
@@ -541,6 +544,7 @@ events:
...
```
sgtcoolguy: Same as above, There's also a hard-coded `message` String property. I think it always returns `"Webview did not load blacklisted url."`

sgtcoolguy (changes requested):
Mostly looks good to me, and clearly raises some API inconsistencies in our `uncaughtException` stuff.

apidoc/Titanium/UI/WebView.yml
```diff
@@ -530,6 +532,7 @@ events:
...
     deprecated:
         since: "6.1.0"
```
sgtcoolguy: @appcelerator/android we should remove the code for this for 9.0.0 (and mark this property as removed in docs here), here's the code: https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewClient.java#L152-L153

apidoc/Titanium/Geolocation/Geolocation.yml
drauggres: AFAIK on Android `code` is always `0`
(https://github.com/appcelerator/titanium_mobile/blob/b295e6321ba38e29612b51d93c1518e87e6fffee/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/TiCompass.java#L135-L136)

apidoc/Titanium/UI/Slider.yml
sgtcoolguy: Note that the most relevant is our `Dimension` type declared in Ti.Ui.View apidocs.

That type has width, height, x, and y; and it is used for the `rect` and `size` properties for `View`s (which can be access after `postlayout` is fired)

In fact, we basically even say the `size` property's `Dimension` just has `x` and `y` === 0; which says to me we should really have a `Size` type which holds `width` and `height`, that `Ti.Ui.View.size` should return that, and that `Dimension` should be a subclass that adds `x` and `y` (and is returned by `Ti.Ui.View.rect`)

(Also, Android has optional properties on Dimension of `absoluteX` and `absoluteY`, which aren't documented on that type, but in `Ti.Ui.View.rect`!)

drauggres: Don't think there is a way to describe properties of a property. The only solution I know is to define additional "pseudo" type like `Dimension` or `AnimationOption`. (btw, docs have ~230 other properties typed with `type: Dictionary`).

But I definitely want a way to describe them all.

sgtcoolguy: Yeah that's what I was getting at. Ideally we'd define the "pseudo" type `Size` with just `width` and `height`, modify the `Dimension` "pseudo" type to extend it.

Also, good point that `Dictionary` has been used as a catch-all. We should warn (not error) about it, since really it's another way of saying "generic `Object`" and over time should be cleaned up/defined.

apidoc/Titanium/App/App.yml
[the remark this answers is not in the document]
sgtcoolguy: @garymathews Opened https://jira.appcelerator.org/browse/TIMOB-27632 for this API parity issue

PR #11353, opened 2019-11-20
docs: add Ti.UI.View.id property

drauggres:
Property `id` of `Ti.UI.View` is not documented.

### DIFF
--- a/apidoc/Titanium/UI/View.yml
+++ b/apidoc/Titanium/UI/View.yml
@@ -1419,6 +1419,15 @@ properties:
     platforms: [iphone, ipad]
     since: 7.3.0
 
+  - name: id
+    summary: View's identifier.
+    description: |
+        The `id` property of the Ti.UI.View represents the view's identifier. The identificator string does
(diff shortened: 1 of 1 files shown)


drauggres:
Since we already have `Ti.UI.View.getViewById` method, I think it wouldn't be wrong to consider `id` property as a legal part of the API.

apidoc/Titanium/UI/View.yml
```diff
@@ -1419,6 +1419,15 @@ properties:
     platforms: [iphone, ipad]
     since: 7.3.0
 
+  - name: id
+    summary: View's identifier.
+    description: |
+        The `id` property of the Ti.UI.View represents the view's identifier. The identificator string does
```
sgtcoolguy:
```suggestion
        The `id` property of the Ti.UI.View represents the view's identifier. The identifier string does
```

sgtcoolguy (changes requested):
small tweak to doc text.

This looks good to me. We don't have specialized methods for this property, but as you state we have `getViewById` which makes use of the optional `id` property. It should likely be official API.

PR #11354, opened 2019-11-20
fix(ios): updated behaviour to original

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27350

### DIFF
--- a/tests/Resources/ti.blob.addontest.js
+++ b/tests/Resources/ti.blob.addontest.js
@@ -1,0 +1,30 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+const should = require('./utilities/assertions');
+
+describe('Titanium.Blob', function () {
+	describe('#toString()', function () {
+		it('binary content', function () {
+			const blob = Ti.Filesystem.getFile('SmallLogo.png').read();
+			should(blob.toString()).eql('[object TiBlob]');
+		});
+
+		it('empty string', function () {
+			const blob = Ti.createBuffer({ value: '' }).toBlob();
+			should(blob.toString()).eql('');
+		});
+
+		it('text content', function () {
+			const blob = Ti.createBuffer({ value: 'test toString()' }).toBlob();
+			should(blob.toString()).eql('test toString()');
+		});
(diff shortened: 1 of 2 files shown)


sgtcoolguy (approved):
LGTM

sgtcoolguy (changes requested):
Generally LGTM, but then I dug deeper...

I think we could fairly easily automate a test for this. Originally my understanding was that a Ti.Blob's toString() method returned it's text value. But I did not consider binary blobs.

So my guess here for the behavior is that if the Blob contains text, toString() should return that, and if it contains binary bytes, it should return `"[object TiBlob]"` . In fact, this is exactly what Android's implementation attempts to do. 

And in further looking, iOS's TiBlob has a `description` method that more or less does the same thing. That appears to be how "old" proxies handled toString() on IOS.

So I think we need to:
- Remove the TiBlob.m `description` method
- Add this JS-based fix as you have it
- add some unit tests for TiBlob that test when:
  - the blob has empty string content
  - the blob has ascii/text content
  - the blob has binary content

vijaysingh-axway:
@sgtcoolguy 
1. API description is used internally. It is not exposed to developers.  It is overridden from NSObject to log proper data while using Ti.API. See [here](https://github.com/appcelerator/titanium_mobile/blob/b295e6321ba38e29612b51d93c1518e87e6fffee/iphone/TitaniumKit/TitaniumKit/Sources/API/APIModule.m#L43). When [newArray componentsJoinedByString:@" "] get called, it looks fro description of every object inside newArray. I think we should not remove it.
2. Unit tests added.

tests/Resources/ti.blob.addontest.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe('Titanium.Blob', function () {
+	var win;
+
+	afterEach(function (done) {
```
sgtcoolguy: There's no need for this hook as none of the tests open a window...

sgtcoolguy (approved):
LGTM, pushed modified commits to have messages match convention and remove the unnecessary test hook.  Tests kind of do FR for QE, but should probably double-check with the original ticket's use case.

ssjsamir (approved):
**FR Passed:** Tested using the the test case mentioned in https://jira.appcelerator.org/browse/TIMOB-27350 able to see data received as a blon
`[INFO]   Received Data: [object TiBlob] `

**Test Environment**
```
MacOS Catalina 10.15.1 beta
Xcode 11.3
Node.js 10.16.3
iPhone 8 (ios 13.3)
```

PR #11357, opened 2019-11-20
build(ios): run tests on device

garymathews:
This runs our unit test suite on an actual iPhone XR attached to a CI build node.

- It uses [ios-deploy](https://github.com/ios-control/ios-deploy) to launch the app on the phone.
- It adds some iOS specific tests for FeedbackGenerator and CollisionBehavior because otherwise related constants on `Ti.UI.iOS` don't get defined (due to how on-device we shave down the APIs based on usage) and we get test failures checking for them.
- This exposed a bug on iOS devices where `Ti.Filesystem.getFile()` and `getAsset()` would not always work as expected.
  - Commit: c722a4fcb2e6475746a7d1e4880a97f67c1c391d
  - Relates to the resources directory path and "standardized" path not matching exactly (and for whatever reason subdirs/files wouldn't "standardize" to the same prefix) so we would strip the prefix incorrectly. (specifically, on device we get a path starting with `/private/var/containers/...` that would resolve to `/var/containers/...`, but a path to a file with the `private` prefix wouldn't also get "standardized"/resolved to drop the leading `/private` in the path; so we need to check against either prefix for stripping)
  - Also affected `__filename` and `__dirname` on iOS devices.
- We use 3rd-party npm packages to convert [CgBI extension formatted PNGs](https://iphonedevwiki.net/index.php/CgBI_file_format) to "valid" PNGs before using pngjs/pixelmatch to compare snapshot images in our tests suites.
  - This also required the buffer perf/fixes from #11810, which would need to be merged first (this PR is rebased on top of that one!)


sgtcoolguy:
@garymathews I'm working on setting up the build nodes with the certs/profiles necessary for this...

Jenkinsfile
```diff
@@ -17,7 +17,7 @@ def includeWindows = isMainlineBranch // Include Windows SDK if on a mainline br
 // Note that the `includeWindows` flag also currently toggles whether we build for all OSes/platforms, or just iOS/Android for macOS
 def runDanger = isPR // run Danger.JS if it's a PR by default. (should we also run on origin branches that aren't mainline?)
 def publishToS3 = isMainlineBranch // publish zips to S3 if on mainline branch, by default
-def testOnDevices = isMainlineBranch // run tests on devices
+def testOnDevices = true //isMainlineBranch // run tests on devices
```
sgtcoolguy: obviously would need to revert this before merge...

sgtcoolguy (commented):
The build changes look ok, with the note that we'd need to revert one line before merge.

The part that's confusing is why the binary blob Ti.Utils hash function tests fail.

Maybe the Logo.png is getting transformed/manipulated only when building for device?

sgtcoolguy:
I'm still baffled by why these specific tests are failing. They're giving the expected values on my now personal iPhone when I build, so... 🤷🏻‍♂️

sgtcoolguy:
@garymathews @ahutton1217 It looks like the iPhone that used to be connected to macOS-darwin isn't anymore? Or maybe the node got macOS updated and it needs to re-trust or something? The build is saying it can't launch iTunes (which no longer exists on the box since they broke it up into separate apps!). When I VNC'd in, I couldn't see any connected device in Finder or in Xcode's Devices and Simulators window.

sgtcoolguy:
manually fixed merge conflicts, merged to master. Cherry-picking to 9_3_X as well

PR #11358, opened 2019-11-21
docs: mark as optional property "animated" of AnimationOption

drauggres:
Property "animated" is not required by View's "hide" and "show" methods.
Also it will cause errors in TypeScript declaration file (if you'll try to build one from docs) for methods "hide" and "show" of "Titanium.UI.iPad.Popover" class.

```typescript

declare interface AnimationOption {
    animated?: boolean;  // "animated: boolean;" will cause "Error TS2416"
}

declare interface PopoverParams {
    animated?: boolean;
    rect?: Dimension;
    view: Titanium.UI.View;
}

class View extends Titanium.Proxy {
    hide(options?: AnimationOption): void;
    show(options?: AnimationOption): void;
}

class Popover extends Titanium.UI.View {
    hide(options: PopoverParams): void;
    show(params: PopoverParams): void;
}
```

P.S. TypeScript declaration file is generated with [this](https://github.com/appcelerator/docs-devkit/pull/8).

### DIFF
--- a/apidoc/Titanium/UI/View.yml
+++ b/apidoc/Titanium/UI/View.yml
@@ -2017,6 +2017,7 @@ properties:
     summary: Determines whether to enable a circular reveal animation.
     type: Boolean
     default: false
+    optional: true
     osver: {android: {min: 5.0}}
 
 ---


drauggres:
Found this in the [current `.d.ts` generator](https://github.com/appcelerator/docs-devkit/blob/4b817b46cd817c8323a7dbe5a3871d4920129abf/packages/titanium-docgen/generators/typescript_generator.js#L678):
> // Make all properties of global interfaces optional by default

Do they really have to be considered optional by default or is it just a "quick hack" (to avoid updating all documentation)?

sgtcoolguy (approved):
LGTM

sgtcoolguy:
cc @janvennemann regarding @drauggres 's question around the typescript generation comment

PR #11134, opened 2019-08-12
refactor(android): improve parity for keyboardType

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-17203

**Description:**
Refactor kyeboardType handling in order to get rid off redundant code and improve parity. Assign the TYPE_CLASS_TEXT only for text type input types.
Note: No unit tests since this is visual change.

@vijaysingh-axway Would you be able to check two things on the iOS related to the ticket - is the TYPE_NAMEPHONE_PAD constant used, because I see it is not taken into account in the Android code? Also is the expected behavior in iOS to be able to put multiple decimal points for TYPE_DECIMAL_PAD? That would not be a native android behavior. Setting the decimal tab flag only allows one decimal point.

**Test case:**
_app.js_
```js
var win = Ti.UI.createWindow({layout:'vertical'}),
	textASCII = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_ASCII}),
	textDecimalPad = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_DECIMAL_PAD}),
	textDefault = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_DEFAULT}),
	textEmail = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_EMAIL}),
	textNamephonePad = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_NAMEPHONE_PAD}),
	textNumbersPunctuation = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_NUMBERS_PUNCTUATION}),
	textNumberPad = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_NUMBER_PAD}),
	textPhonePad = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_PHONE_PAD}),
	textURL = Ti.UI.createTextField({width: 300, keyboardType: Titanium.UI.KEYBOARD_TYPE_URL});
win.add([textASCII,
		textDecimalPad,
		textDefault,
		textEmail,
		textNamephonePad,
		textNumbersPunctuation,
		textNumberPad,
		textPhonePad,
		textURL]);
win.open();

```

[...]

PR #11360, opened 2019-11-21
feat(android): match for decimal separator depending on locale

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27616

**Description:**
This fixes a bunch of disparities for input types, mainly showing a numpad for decimal numbers input
type that accepts separator based on the locale of the device. Slight changes for other constants to get as close as possible to the results for different constants on iOS.
No unit tests as this PR changes what software keyboard is shown.
This is a breaking changes since it changes what keyboard is shown for KEYBOARD_TYPE_NUMBERS_PUNCTUATION constant to match the iOS counterpart. Unfortunately I was not able to force Android to open the soft keyboard directly on numbers and punctuation.

@jquick-axway Thanks for the pointer! After digging up some source trying to get the new constructor I figured out what the problem is. Android is properly accepting the decimal symbol based on the locale, but the `DigitsKeyListener` enters "needsAdvancedInput" mode due to the character set not matching any of these:
https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/core/java/android/text/method/DigitsKeyListener.java#79
As a result of this we receive the inputType for Text:
https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/core/java/android/text/method/DigitsKeyListener.java#324
So we end up with a keyboard that shows everything, but accepts only digits and the decimal separator for the current locale. This is the disparity with iOS. I am setting a DigitsListener that returns an `inputType` that will show a numpad, showin both comma and dot as a separator, but accepting only the one for the locale.

@Topener Thank for helping me with the behavior on iOS for different constants! 
  

**Test case:**
_app.js_
```js
var hintsArray = [
    'KEYBOARD_TYPE_ASCII',
    'KEYBOARD_TYPE_NUMBERS_PUNCTUATION',
    'KEYBOARD_TYPE_URL',
    'KEYBOARD_TYPE_NUMBER_PAD',
    'KEYBOARD_TYPE_PHONE_PAD',
    'KEYBOARD_TYPE_EMAIL',
    'KEYBOARD_TYPE_NAMEPHONE_PAD',
    'KEYBOARD_TYPE_DEFAULT',
    'KEYBOARD_TYPE_DECIMAL_PAD'
    ],
    win = Ti.UI.createWindow({ layout: 'vertical'});
for (var i=0; i < 9; i++) {
    win.add(Ti.UI.createTextField({ width: 300,
        top: 10,
        keyboardType: i,
        hintText: hintsArray[i]
    }));
}
win.open();
```
TextField with TYPE_DECIMAL_PAD `keyboardType` should accept decimal separator based on locale. "." for English(US) and "," for French (or any locale that uses dot as separator). 

Warning: I am expecting conflicts this and https://github.com/appcelerator/titanium_mobile/pull/11134 get merged.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
@@ -17,6 +17,8 @@ import android.text.Editable;
 import android.text.InputFilter;
 import android.text.InputType;
 import android.text.Spannable;
+import android.text.SpannableStringBuilder;
+import android.text.Spanned;
 import android.text.TextUtils.TruncateAt;
 import android.text.TextWatcher;
 import android.text.method.DialerKeyListener;
@@ -46,6 +48,7 @@ import org.appcelerator.titanium.util.TiRHelper.ResourceNotFoundException;
 import org.appcelerator.titanium.util.TiUIHelper;
 import org.appcelerator.titanium.view.TiUIView;
 
+import java.text.DecimalFormatSymbols;
 import java.util.HashMap;
 import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -736,21 +731,31 @@ public int getInputType()
...
+					customKeyListener = true;
 					break;
 				case KEYBOARD_URL:
 					Log.d(TAG, "Setting keyboard type URL-3", Log.DEBUG_MODE);
 					tv.setImeOptions(EditorInfo.IME_ACTION_GO);
 					textTypeAndClass |= InputType.TYPE_TEXT_VARIATION_URI;
 					break;
 				case KEYBOARD_DECIMAL_PAD:
-					textTypeAndClass = (InputType.TYPE_NUMBER_FLAG_DECIMAL | InputType.TYPE_NUMBER_FLAG_SIGNED);
+					tv.setKeyListener(new DigitsKeyListener(Locale.getDefault(), true, true) {
```
jquick-axway: The `new DigitskeyListener()` with the `Locale` argument is API Level 26 only. So, this will likely crash on an OS older than Android 8.0. But we definitely need to call this constructor on Android 8.0 and higher since it's the only official way to get it to support a localized decimal point as can be seen in Google's code below...
[DigitsKeyListener.java#L202](https://github.com/aosp-mirror/platform_frameworks_base/blob/d87727f7d9350fd94831c5c5e138f9400e1823fd/core/java/android/text/method/DigitsKeyListener.java#L202)

Do you have any ideas on how to get this working on older Android OS versions?

Since Google's source code is Apache licensed, we can copy their current code into our `TiDigitsKeyListener` and use that for the older OS versions only. Effectively, we'd be backporting it like how Google backports it via their support libraries. Ideally, we should avoid this unless we have no other choice.

ypbnv:
@jquick-axway Totally forgot about that... Unfortunately the only way I was able to get backwards compatibility back to API 19 was to add a custom `filter` override for the DigitsKeyListener. Not really a fan of this method, but it does the work. I found out that they Google has improved that for API 24, but I am not sure if it is worth it to make 3 separate cases.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -815,6 +967,11 @@ public int getInputType()
 		disableChangeEvent = false;
 	}
 
+	private boolean isSignChar(char value)
+	{
+		return value == '-' || value == '+';
```
jquick-axway, line 972: We also need to check for unicode minus signs `\u2212` and `\u2013 `.
This is based on what Google has coded here...
https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/text/method/DigitsKeyListener.java

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -707,50 +711,196 @@ public void handleKeyboard(KrollDict d)
...
+					if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+						tv.setKeyListener(new DigitsKeyListener(Locale.getDefault(), true, true) {
+							@Override
+							public int getInputType()
+							{
+								return decimalInputType;
+							}
+						});
+					} else {
+						// For API levels lower than 26 we override the filter method of the DigitsKeyListener too
```
jquick-axway, line 775: We need to comment that we've copied Google's code; which is how we're back-porting it.
https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/text/method/DigitsKeyListener.java

jquick-axway (commented):
I'll make the final edits to this PR.

jquick-axway:
Closing in favor of PR #11652 

Android's keyboard handling requires some additional refactoring. Google's decimal separator fixes in Android 8.0 does not fully fix the problem and a custom key listener on our end. Also, we still had major parity differences between Android and iOS that need to be resolved.

PR #11364, opened 2019-11-22
feat(ios): support font scaling for custom fonts

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-25847


apidoc/Titanium/UI/Font.yml
```diff
@@ -100,6 +100,10 @@ properties:
     description: |
         Use one of the TEXT_STYLE constants from <Titanium.UI> to set the font to a predefined system font.
         When this property is set to a valid value, all other font properties are ignored.
+
+        Notes:
+         Since SDK 9.0.0 property <Font.fontFamily> will not be ignored for iOS 11+. 
+         Custom fonts can be used with this property for dynamic font management.
```
janvennemann: This reads like a bug fix release note so i don't think it should be part of the documentation.

janvennemann (changes requested):

apidoc/Titanium/UI/Font.yml
vijaysingh-axway: We have to tell developer that this behavior is available to use for iOS 11+. So I think there should be something in doc to tell this otherwise how developer will know. That's why I put in Notes.

janvennemann: Yes, i agree that we have to make this changes public somewhere. And for that i think the release notes are a better place. So users now right away, ok that was fixed in 9.0.0.

A note in the docs saying that the property works as expected with the current SDK seems unnecessary. There should have been a note in the docs for previous SDK versions saying that the property does **not** work as expected, which we would have removed now.

janvennemann: After reading this in context again i'm fine with the message how it is. IMHO still not ideal though, but that's an issue of our docs not being versioned.

janvennemann (approved):

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20191216133819
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516
Xcode: 11.2
Device: iPhone Xr(v13.2.2)

PR #11366, opened 2019-11-25
chore: add npm script to report deprecated but unremoved api

sgtcoolguy:
**Optional Description:**
This adds an npm script: `npm run deprecations` which will parse the apidocs and find types/properties/methods we've marked as deprecated but have not yet marked as removed.

Lists the APIs in order from oldest deprecated version to newest. This should give a good listing for things we need to get rid of in breaking change releases.

Gives me:
Screen Shot 2019-11-25 at 4 11 10 PM

### DIFF
--- a/build/scons-deprecations.js
+++ b/build/scons-deprecations.js
@@ -1,0 +1,124 @@
+#!/usr/bin/env node
+'use strict';
+
+const path = require('path');
+const glob = require('glob');
+const yaml = require('js-yaml');
+const fs = require('fs-extra');
+const promisify = require('util').promisify;
+const semver = require('semver');
+const chalk = require('chalk');
+
+/**
+ * @param {object} thing type, property or method from docs
+ * @param {string} thingPath full API path of the object we're checking
+ */
+async function checkDeprecatedButNotRemoved(thing, thingPath) {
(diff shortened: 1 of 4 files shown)


sgtcoolguy:
Relates to #11362 

Note that I didn't have it check method parameters or events.

sgtcoolguy:
cc @janvennemann this probably should go into appcelerator/docs-devkit with a parameter to filter out results with a new-enough version (i.e. when we lint modules or sdk api docs, we can passing's and say, show me all results so long as the deprecated since version is < 8.0.0)

sgtcoolguy:
Also possibly useful? another variant that finds deprecated and removed APIs older than version X for potential removal from the apidocs altogether (i.e. we probably don't need to document APIs that have been removed 2/3+ years ago any more)

sgtcoolguy:
OK, followed it up with another script that can show the "old" removed apis to give us a listing of things we should consider dropping from the YML docs altogether. For now it's set to consider any API removed prior to 7.0.0 release.

janvennemann (approved):
Awesome! This will be really helpful in cleaning up our code base and docs. Code looks good, approved!

sgtcoolguy:
merged manually

PR #11367, opened 2019-11-25
fix(android): prevent snapshots from failing build

garymathews:
- Implement fallback for when a snapshot fails to generate for a platform.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27627)


android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -769,6 +769,32 @@ public boolean isCoverageEnabled()
 		return false;
 	}
 
+	public static void launch()
+	{
+		final TiRootActivity rootActivity = (TiRootActivity) TiApplication.getAppRootOrCurrentActivity();
```
jquick-axway: We need to call `getRootActivity()` instead.
Calling `getAppRootOrCurrentActivity()` can return a `TiActivity` object when running in LiveView.

We may want to check if `rootActivity` is `null` just in case Android developer option "Don't keep activities" is enabled... although our `softRestart()` method checks for this already.

jquick-axway (changes requested):

PR #11369, opened 2019-11-26
fix(android): fix setting DatePicker's minDate\maxDate

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27625

**Description:**
Fix for setting min\max Date for Date type of Ti.UI.Picker. Some code optimization. Adding accessors
for the properties. There was a scenario where the `value` was not matching what the Picker was showing, so in order to avoid that now the getter is returning what the native component is showing.
Thus the unit tests that would cover such an attempt.

About DatePickerProxy - would that be a leftover class ? I don't see support for `Ti.UI.createDatePicker` on Android side and I think we can remove the file.

**Test case:**
_app.js_
```js
var maxDate = new Date();
 
    layout: "vertical",
    backgroundColor: "#454441"
});
     type: Ti.UI.PICKER_TYPE_DATE,
     maxDate: maxDate
});
window.add(picker);
    title: "Set Max Date"
});
    picker.maxDate = new Date(maxDate.getFullYear() + 1, maxDate.getMonth(), maxDate.getDate());
});
window.add(button);
window.open();

```


ypbnv:
I believe the `minDate` test fails due to a typo in it. It was undetected so far because minDate did not have a getter providing the value directly from the native picker component.

android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
```diff
@@ -778,4 +780,45 @@ public String getApiName()
...
+																	   : datePicker.getMaxDate());
+			}
+		}
+		return null;
+	}
+
+	@Kroll.getProperty
+	public Date getValue()
+	{
+		if (type == UIModule.PICKER_TYPE_DATE) {
```
sgtcoolguy: This doesn't account for Time pickers. I know Android doesn't support Date + Time pickers for now, but for a Time picker, I think this should likely return the Date object equivalent to "today" at the selected time of the picker.

sgtcoolguy: Also, not the point of this PR, but why *don't* we have a Date + Time picker on Android? I mean we support each individually. Why can't we just create a component that includes both (and merges the date and time value into a single Date object?). Maybe this is a good job to do at the JS level, overriding the createPicker method to handle that specific case and have it silently delegate to both pickers under the hood?

tests/Resources/ti.ui.picker.addontest.js
```diff
@@ -0,0 +1,74 @@
...
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe('Titanium.UI.Picker', function () {
+
+	var win;
```
sgtcoolguy: might want to "clean up" (close) the window after the test like the other UI test suites do.

sgtcoolguy (changes requested):

sgtcoolguy:
Related, the `useSpinner` property has been deprecated since 5.2.1, I think we can remove the property and the extra code we have around it here. (`TiUITimeSpinner`, `TiUIDateSpinner`, `TiUISpinner`)

jquick-axway:
@sgtcoolguy, I don't think we should remove the "useSpinner" feature and perhaps consider un-deprecating it. This is because some app devs like using it.

If you follow the link below, the reason it was deprecated was because there used to be a bug where the spinner's content would not line-up with the other columns correctly. Instead of fixing it, the feature was deprecated and we told devs to use the drop-down picker instead.
https://jira.appcelerator.org/browse/TIMOB-17998

But we fixed the spinner's bug in 7.3.0. So, I think we should keep it now. :)
https://jira.appcelerator.org/browse/TIMOB-19822

android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
jquick-axway: I agree that we should our own custom Date+Time picker.
I've written a feature request for it here: [TIMOB-27840](https://jira.appcelerator.org/browse/TIMOB-27840)

jquick-axway:
I'm closing this PR in favor of PR #11621 which is focused more on the bug itself.
We can refactor things later.

PR #11197, opened 2019-09-05
impr(android): implement foregroundServiceType parameter

garymathews:
- Implement `foregroundServiceType` parameter into `Ti.Android.Service.foregroundNotify`

##### TEST CASE
**tiapp.xml**
```XML
<android>
  <manifest>
    <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
  </manifest>
  <services>
    <service url="locationService.js" android:foregroundServiceType="location"/>
  </services>
</android>
```
**locationService.js**
```JS
console.log(`Background location service started.`);

// Create foreground service.
Ti.Android.currentService.foregroundNotify(
	123,
	Ti.Android.createNotification({
		contentTitle: "Background Location Service",
		contentText: "Obtaining location data...",
		contentIntent: Ti.Android.createPendingIntent({
			intent: Ti.App.Android.launchIntent || Ti.Android.createIntent(),
		})
	}),
	Ti.Android.FOREGROUND_SERVICE_TYPE_LOCATION
);
```
**app.js**
```JS
const win = Ti.UI.createWindow({ backgroundColor: 'gray', layout: 'vertical' });
const backgroundServiceBtn = Ti.UI.createButton({ title: 'START BACKGROUND SERVICE' });
const listView = Ti.UI.createListView();

let count = 1;
let service = null;

// Set accuracy to high
Ti.Geolocation.accuracy = Ti.Geolocation.ACCURACY_HIGH;
 
// Enable manual configuration via location providers.
Ti.Geolocation.Android.manualMode = true;
 
// Define a location provider.
Ti.Geolocation.Android.addLocationProvider(
    Ti.Geolocation.Android.createLocationProvider({
        name: Ti.Geolocation.Android.PROVIDER_GPS,
        minUpdateDistance: 0,
        minUpdateTime: 5000
    })
);
 
function getLocation () {

    // Create location event listener.
    Ti.Geolocation.addEventListener('location', e => {

        // Create new section for location data.
        let section = Ti.UI.createListSection({ headerTitle: `#${count++} - ${new Date().toTimeString()}` });
        if (e.success) {
            if (e.coords) {
                e = e.coords;
            }

            // Insert location data.
            section.setItems([
                { properties: { title: 'LOCATION:\n' + e.latitude + ', ' + e.longitude, color: 'green' } },
[...]

PR #11378, opened 2019-12-06
fix(android): ignore additional require argument

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27633

@garymathews This is my naive solution, which just drops the additional parameter we currently support in our `require` on Android. I couldn't find any `require`s in our codebase that make use of this second parameter.

I'm note sure if this is just a relic from older versions or if we actually still use this. We could change this PR to something like this to avoid breaking changes:

```js
  function require(path, context) {
    if (typeof context !== 'object' || context === null) {
      context = {}
    }
    return self.require(path, context);
  }
```

### DIFF
--- a/android/runtime/common/src/js/module.js
+++ b/android/runtime/common/src/js/module.js
@@ -562,8 +562,8 @@ Module.prototype._runScript = function (source, filename) {
 	var self = this,
 		url = 'app://' + filename;
 
-	function require(path, context) {
-		return self.require(path, context);
+	function require(path) {
+		return self.require(path);
 	}
 	require.main = Module.main;
 


sgtcoolguy:
Looks to me like we pass around this stuff in the context:
- currentTab
- currentService
- currentTabGroup
- global
- module
- sourceUrl

The module/sourceUrl are set by the module/js require/load code anyways.

The interesting ones are `currentTab`, `currentTabGroup` and `currentService`:
- I don't see where we're actually setting a value for `currentTab` or `currentTabGroup`. I suspect they may be broken?
- `currentTabGroup` isn't even documented.
- `currentService` appears to be passed in via Module.runModule via V8Runtime java/cpp layer

I suspect that the require should probably not take a second argument at all and should inherit the parent module's context with overridden module/sourceUrl values.

I'll see if I can come up with something....

sgtcoolguy:
Note that currentTab doesn't appear to be implemented on iOS. I suspect it may have been removed long ago but the docs never deprecated/removed it.

sgtcoolguy:
...and that `Ti.Android.currentActivity` and `rootActivity` are implemented in java code.

jquick-axway:
This will **break** the `Ti.Android.currentService` property.

This property is not defined in Java code. It's defined in JavaScript.
How the code currently works is every required-in JS file gets its own **copy** of the `Ti.Android` module. A service JS file referenced in the "tiapp.xml" (ex: `<service url="MyService.js"/>`) will have its copy of the `Ti.Android` module assigned a Java `ServiceProxy` to the `Ti.Android.currentService` property. Each service JS will reference a different Java `ServiceProxy` instance since they can be ran in parallel. And the `currentService` property will be `undefined` for non-service scripts.

Pretty much everyone who uses Android services use the `currentService` property. It is a popular feature. Especially for devs who want to collect location data in the background. Example code which leverages this property can be found here...
https://github.com/appcelerator/titanium_mobile/pull/11197

I agree that this is not a great design. Changing it isn't trivial though.

jquick-axway (changes requested):
This will break the `Ti.Android.currentService` property.

sgtcoolguy:
I'm going to kill this one off in favor of #11380, since this naive approach seems like it'll break `Ti.Android.currentService` and there's a lot of collateral changes I think that should be made around this stuff.

jquick-axway:
Hmm... maybe I'm wrong. We do have a mocha unit test which tests the `currentService` property below.
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/ti.android.service.normal.js

I think it's not failing because we only apply the `ServiceProxy` as a 2nd argument when we load the service JS from the Java side. No one requires-in a service JS directly. It's been a year since I've played around with this code, so, my memory is a bit fuzzy on the exact details. I would still want to run various tests to be sure though.

PR #11380, opened 2019-12-06
fix(android): ignore additional require argument

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27633

**Description:**

Related to #11378

If users pass in a non-object argument as a second argument to require, that is used as the internal `context` object that is shuttled around and fails with a crash (since we tried to hang properties it).

This doesn't expose the context as an argument at all, but instead uses parent module's context with a changed `sourceUrl` value.

Note that the context object being passed around seems almost entirely unnecessary. The only values I truly see used are the `sourceUrl` which we pass in to proxies and help resolve relative paths in something like a `Ti.UI.View` - and I'm not even sure it's passing in the correct value (maybe be passing in the parent module's url rather than the "current"?).

This exposed that a lot of the context values are not being used properly:
- `currentTab` is not actually used anywhere so `Ti.UI.currentTab` is always undefined on Android. It doesn't appear to be implemented on iOS either. I would suggest we mark the property removed (since it's been broken for some time) - or we should likely hack something in the common JS code to access the currentTab (by hijacking TabGroup open/close and it's `activeTab`).
- `currentTabGroup` wasn't used anywhere or documented.
- `Ti.Android.currentActivity` is implemented in Java
- `Ti.Android.currentService` is passed through this context via V8Runtime.runModule and appears to be working correctly
  - additional note: iOS has a background service and sets a reference in `Ti.App.currentService`, while Android uses `Ti.Android.currentService`. They seem to me to be conceptually equivalent, so why? And the iOS variant isn't in the Ti.App docs!


sgtcoolguy:
Something strange is going on, as the test I added to verify that Ti.UI.currentTab is broken is not actually running on iOS. No idea why.

sgtcoolguy:
Turns out that when encrypting JS, iOS can't then use the Filesystem API to try and gather the add-on tests: https://jira.appcelerator.org/browse/TIMOB-27646

janvennemann (approved):
Looks good to me, approved!

I guess we need a follow up ticket to address the possible parity improvements regarding `Ti.App.currentService` and `Ti.Android.currentService`. If they can be unified we should definitely do that.

ssjsamir (approved):
FR Passed: Application no longer crashes when using Additional parameters in 'require'.

**Test Steps:**

1. Create a new Titanium application 
2. Add the following to the `app.js`
`require('./test.js', true);
`
3. Create a `test.js` file
4. Add the following to the `test.js` file
`console.log("****** I should be visible, No crash should be present ****** ")`
5. Run the application 
6. Above log should be visible in the console and application should not crash

**Test Environment**
```
MacOS Catalina 10.15.1 beta
Xcode 11.3
Node.js 10.16.3
Pixel xl (Android 10)
```

PR #11383, opened 2019-12-09
Add progress event to Ti.UI.WebView on Android

arifje:
Add progress event for Ti.UI.WebView on Android.

Ticket: https://jira.appcelerator.org/browse/AC-6456

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -12,8 +12,6 @@ import android.os.Handler;
 import android.os.Message;
 import android.webkit.ValueCallback;
 import android.webkit.WebView;
-
-import java.util.Map;
 import java.util.HashMap;
 
 import org.appcelerator.kroll.KrollDict;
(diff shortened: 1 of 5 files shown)


jquick-axway:
We need to double check how this works when the webpage contains "iframes" between Android and iOS. The "progress" event should either only apply to (1) the main page or (2) represent all of the web content (main page + iframes). I'm leaning towards option 1 since an iframe can timeout.

skoften:
@jquick-axway agree on option 1, as iframes can been seen as separate/nested webviews with there kwn loading time (and timeouts indeed, because of third party domains), right?

garymathews (approved):
CR: **PASS**

Can confirm `progress` reaches `100` once page and iframe content has loaded.

ssjsamir (approved):
**FR Passed** able to see the progress event go from 0-100: 
Using the following test case:
```
    url: 'http://www.google.com'
});
webview.addEventListener('progress', function(e) {
  Ti.API.info(e.progress);
});
win.add(webview);
win.open();
```

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta 
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
iphone 8 Sim (13.4)
Pixel Xl (10.0)
```

PR #11384, opened 2019-12-09
chore(android): include x86_64 in module template

garymathews:
###### NOTE: REQUIRES https://github.com/appcelerator/titanium_mobile/pull/11294

- Include `x86_64` architecture in module builds by default

##### TEST CASE \#1
- Attempt to rebuild existing module such as [ti.map](https://github.com/appcelerator-modules/ti.map)
- Built module should contain `x86_64` libraries

##### TEST CASE \#2
- Attempt to build an application with Titanium 9.0.0 including modules built with Titanium 7.0.0+
- Application should build successfully, but warn `x86_64` has been excluded:
```
[INFO]  Module ti.map does not contain x86_64 ABI. Application will build without x86_64 ABI support!
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27487)


lokeshchdhry:
Found some issues. Have updated @garymathews.

sgtcoolguy:
@garymathews I think this is having issues with the native modules used in the test suite now. Maybe because of the addition of `x86_64` architecture, it's requiring that arch in the modules (which obviously isn't there yet)?

jquick-axway:
Ideally, we want to support including `9.x.x` modules and older modules for backward compatibility. But the issue with this is that the older built modules don't have `x86_64` built libraries, which means you'll crash when running on an `x86_64` device/emulator (it's all or nothing).

**Proposal 1 (Best Solution):**
In our `_build.js` script's module validation code, we can check the `minsdk` of all included modules. If we find at least 1 module that has a `minsdk` older than `9.0.0`, then we should remove the `x86_64` string from the architectures array. We should log a warning about if both an old and new module are included (mixed mode) and we can't support `x86_64` in this case.

And perhaps we should not bump up the `apiversion` because I believe this prevents us from including older modules for backward compatibility.

**Proposal 2:**
If the app includes at least 1 module whose `mindsdk` is `9.0.0` or higher, then all modules need to be `9.0.0` or higher. That is, older modules cannot be included and we trigger a build failure. It's all or nothing.

I don't think this is a good solution though. We're going to need to rebuild `ti.playservices` for `9.0.0`, so, odds are most devs are going to get bit by this.

jquick-axway:
Also, our `_buildModule.js` script's manifest migration should probably force devs to add `x86_64` to the `architectures:` property in the module's `manifest` file.

Should it even be configurable at all? Most of the time no... unless you're including a 3rd party C/C++ `*.so` library. Hmm...

jquick-axway (approved):
CR: Pass

jquick-axway:
@lokeshchdhry, I suggest that you test with our [ti.imagefactory](https://github.com/appcelerator-modules/ti.imagefactory) module instead. Now that our AndroidX PR has been merged, the `ti.map` module won't build without code changes since it has  Google Support library class references (it needs to be migrated).

lokeshchdhry:
FR Passed.

1. Module built with SDK 9.0.0 builds & works as expected with an app. It has the x86_64 arch.
2. Module built with 7.0.0+ also works when used in an app built with SDK 9.0.0. As it does not have the x86_84 arch we see the warning as defined in this PR.

SDK Ver:          9.0.0
OS Ver:           10.14.5
Xcode Ver:        Xcode 11.3
Appc NPM:         4.2.15
Appc CLI:         7.1.2
Daemon Ver:       1.1.3
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.4
Node Ver:         12.13.1
NPM Ver:          6.12.1
Java Ver:         11.0.1
Android Devices:  ⇨ google Pixel (Android 10)

PR #11385, opened 2019-12-09
chore(ios): drop iOS 9 support

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27641


hansemannn:
Is this really *necessary*? I see many apps that want to support this version, as well as competitive frameworks that still maintain it. And since it's compiled out during build time anyways, it doesn't hurt anyone.

iphone/Classes/TiAppiOSUserNotificationActionProxy.m
```diff
@@ -79,38 +75,22 @@ - (id)notificationAction
...
 }
 
 - (void)setAuthenticationRequired:(id)value
 {
-  if ([TiUtils isIOSVersionOrGreater:@"10.0"]) {
-    return; // Not available on iOS 10+
-  }
-
-  [(UIMutableUserNotificationAction *)[self notificationAction] setAuthenticationRequired:[TiUtils boolValue:value]];
+  DebugLog(@"[WARN] This property is not available in iOS 10+.");
```
sgtcoolguy: mark as deprecated/removed in 9.0.0 in our docs

iphone/Classes/TiAppiOSUserNotificationActionProxy.m
```diff
@@ -79,38 +75,22 @@ - (id)notificationAction
...
 }
 
 - (void)setDestructive:(id)value
 {
-  if ([TiUtils isIOSVersionOrGreater:@"10.0"]) {
-    return; // Not available on iOS 10+
-  }
-
-  [(UIMutableUserNotificationAction *)[self notificationAction] setDestructive:[TiUtils boolValue:value]];
+  DebugLog(@"[WARN] This property is not available in iOS 10+.");
```
sgtcoolguy: mark as deprecated/removed in 9.0.0 in our docs

iphone/Classes/TiAppiOSUserNotificationActionProxy.m
```diff
@@ -79,38 +75,22 @@ - (id)notificationAction
...
 }
 
 - (void)setBehavior:(id)value
 {
-  if ([TiUtils isIOSVersionOrGreater:@"10.0"]) {
-    return; // Not available on iOS 10+
-  }
-
-  [(UIMutableUserNotificationAction *)[self notificationAction] setBehavior:[TiUtils intValue:value]];
+  DebugLog(@"[WARN] This property is not available in iOS 10+.");
```
sgtcoolguy: mark as deprecated/removed in 9.0.0 in our docs

iphone/Classes/TiAppiOSUserNotificationActionProxy.m
```diff
@@ -79,38 +75,22 @@ - (id)notificationAction
 
 - (void)setActivationMode:(id)value
 {
-  if ([TiUtils isIOSVersionOrGreater:@"10.0"]) {
-    return; // Not available on iOS 10+
-  }
-
-  [(UIMutableUserNotificationAction *)[self notificationAction] setActivationMode:[TiUtils intValue:value]];
+  DebugLog(@"[WARN] This property is not available in iOS 10+.");
```
sgtcoolguy: mark as deprecated/removed in 9.0.0 in our docs

sgtcoolguy (changes requested):

sgtcoolguy:
@hansemannn Is it *necessary*, I suppose not we: could try to maintain backwards compatibility with every version of every OS. But we kept iOS 9 around for at least one (possibly 2) full breaking release cycles past our own internal usage thresholds before.

This cleans up some of our code, and according to analytics we have near-zero usage on iOS 9 devices (and looking at Apple/3rd party stats also confirms this - there's really only the interesting case of old iPads "stuck" at 9 due to Apple's OS upgrade limit on them). Additionally, what's not really shown/discussed here is that the minimum iOS version also plays role in the JS polyfilling layer we have. Moving up one major version should help drop a significant number of shims/polyfills for ES6 support. (See https://caniuse.com/#compare=safari+9,safari+10&compare_cats=JS )

iphone/Classes/TiAppiOSUserNotificationActionProxy.m
vijaysingh-axway: Actually these properties are available but at creation time only for iOS 10+. Manual setters are not available. I think I should delete these setters.

vijaysingh-axway: @sgtcoolguy Deleted setters as these are creation-only. These properties are already used while calling - (void)_initWithProperties:(NSDictionary *)properties.

hansemannn:
Despite the fact that probably only 20 % of the Titanium actually use the Appcelerator Analytics, I feel like the benefits are not enough to do this right now. iOS 8 was dropped just in SDK 8.0.0, but if no one complains now, it's probably okay.

sgtcoolguy (approved):
code changes look good.

QE, please ask me to merge this after FR so that I can squash it to one commit and tweak the commit message. Thanks!

iphone/cli/commands/_build.js
```diff
@@ -2114,23 +2114,20 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
...
-					logger.info(__('Building for iOS %s; using %s as minimum iOS version', version.format(this.iosSdkVersion, 2).cyan, defaultMinIosSdk.cyan));
-					this.minIosVer = defaultMinIosSdk;
-				} else if (version.lt(this.minIosVer, defaultMinIosSdk)) {
-					logger.info(__('The %s of the iOS section in the tiapp.xml is lower than minimum supported version: Using %s as minimum', 'min-ios-ver'.cyan, version.format(defaultMinIosSdk, 2).cyan));
-					this.minIosVer = defaultMinIosSdk;
+				let defaultMinIosVersion = this.packageJson.minIosVersion;
+
+				this.minIosVer = this.tiapp.ios['min-ios-ver'] || defaultMinIosVersion;
+
+				if (version.gte(this.iosSdkVersion, '10.0') && version.lt(this.minIosVer, '10.0')) {
```
ewanharris: I think maybe this check and the check on 2125 _used_ to be useful (this line looks to have been to force people down the right path #8198) but now our minimum is 10 maybe it's redundant?

`this.iosSdkVersion >= 10` and `this.iosSdkVersion >= 6` will always be true nowadays, and `this.minIosVer < 10` being true seems to duplicate the check on 2128 where we force the minimum to 10 if someone sets it to 9??

ewanharris (approved):
Just a minor comment around maybe removing some (what appear to me at least) dupe logic checks, but looks good to me!

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20191213104900
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516
Xcode: 11.2

PR #11390, opened 2019-12-12
chore(modules): remove deprecated modules

ewanharris:
BREAKING CHANGE: Removes ti.touchid and ti.safaridialog from the SDK distribution

Closes [TIMOB-27650](https://jira.appcelerator.org/browse/TIMOB-27650)


sgtcoolguy (approved):
LGTM

These modules have been deprecated and replace by newer modules we ship. If users need to continue using them, they can keep using the globally installed versions from previous SDK installs - or grab the zips from the modules' Github releases.

PR #11386, opened 2019-12-10
refactor(ios, android): remove deprecated APIs

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27617

**Description:**
Follow-on to #11368 

This is an attempt to include more removals, updated docs for long-deprecated APIs

- `Ti.Android.Activity` events `create`, `destroy`, `restart`, `resume`, `start` and `stop` are removed here for 9.0.0 (we have properties to register callback function for these lifecycle hooks)
- `Ti.UI.TabGroup` `selected`/`unselected` events are removed here in 9.0.0 (use `blur`/`focus`)
- `Ti.UI.Tab` `blur`/`focus` events are removed here for 9.0.0 (use `selected`/`unselected`)
- `Ti.Contacts` `#getGroupByID()`, `#getPersonByID()`, `#requestAuthorization()` methods removed here for 9.0.0
- `Ti.Calendar` `#requestEventsAuthorization()` and `#requestAuthorization()` methods were removed in 8.1.0
- `Ti.Media` `#requestAudioPermissions()`, `#requestCameraAccess()`, `#requestAuthorization()` methods removed here for 9.0.0
- `Ti.UI.Android.CardView` `card*` properties were removed in 6.0.0
- `Ti.UI.TextField` and `Ti.UI.TextArea` `appearance` property removed here for 9.0.0 (use `keyboardAppearance`)
- `Ti.Ui.TextField` `paddingRight` and `paddingLeft` properties removed here for 9.0.0 (use `padding.right` and `padding.left`)
- `Ti.Gesture` `#isPortrait()` and `#isLandscape()` were removed in 7.0.0 on Android, 8.1.0 on iOS

PR #11391, opened 2019-12-12
docs: extract out common types

sgtcoolguy:
**JIRA**
https://jira.appcelerator.org/browse/TIMOB-27688

**Description:**
- Updates to titanium-docgen 4.1.0 which now warns about using generic `Dictionary`/`Object` types
- Extracted out `Size` type: `width` and `height`
- Extracted `Dimension` type from View's yml file, have it extend `Size` and add `x` and `y` properties
  - `Ti.Blob.#imagesAsCropped()` now refers to `Dimension` rather than a custom type that had the same properties.
  - `Ti.Media CameraMediaItemType` now refers to `Dimension` for `cropRect` and `Size` for `previewRect` rather the defining custom types with same properties.
- Extracted `Padding` type from `ViewPadding`. Attempted to clean up any special types that had the same concept (basically any padding/inset properties/types) to refer to it.
    - Kept `TabIconInsets`, had it extend `Padding` and noted right/bottom are excluded.
    - Kept `TextFieldPadding`, had it extend `Padding` and noted top/bottom are Android-only.
- Fix `Ti.DB.ResultSet` `#field()` and `#fieldByName()` returns to be a single entry with type as an array of possible return types.

### DIFF
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -95,7 +95,7 @@ methods:
                 * `UIApplicationOpenURLOptionsOpenInPlaceKey` (Boolean value)
                 * `UIApplicationOpenURLOptionUniversalLinksOnly` (Boolean value)
             Read more about the available keys in the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionskey?language=objc).
-        type: Dictionary
+        type: OpenURLOptions
         optional: true
       - name: callback
         summary: The optional callback that is called once the URL is opened (iOS 10+).
@@ -407,3 +407,34 @@ properties:
   - name: irq
     summary: The number of milliseconds the CPU has spent in irq mode.
     type: Number
+
+---
+name: OpenURLOptions
+summary: |
+    The optional options to pass to the URL handling (iOS 10+). Pass a
+    dictionary with one or more of the following string-keys:
+        * `UIApplicationOpenURLOptionsSourceApplicationKey` (String value)
+        * `UIApplicationOpenURLOptionsAnnotationKey` (Array value)
+        * `UIApplicationOpenURLOptionsOpenInPlaceKey` (Boolean value)
+        * `UIApplicationOpenURLOptionUniversalLinksOnly` (Boolean value)
+    Read more about the available keys in the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionskey?language=objc).
+properties:
+  - name: UIApplicationOpenURLOptionsSourceApplicationKey
+    type: String
+    summary: The value of this key is an String containing the bundle ID of the app that made the request. If the request originated from another app belonging to your team, UIKit sets the value of this key to the ID of that app. If the team identifier of the originating app is different than the team identifier of the current app, the value of the key is nil.
+    optional: true
+
+  - name: UIApplicationOpenURLOptionsAnnotationKey
+    type: Array
+    summary: contains the information passed to a document interaction controller object’s annotation property.
+    optional: true
+
+  - name: UIApplicationOpenURLOptionsOpenInPlaceKey
+    type: Boolean
+    summary: When the value of this property is `false`, you must copy the document to maintain access to it. If the flag is not set, you also must copy the document before you can use it.
+    optional: true
+
+  - name: UIApplicationOpenURLOptionUniversalLinksOnly
+    type: Boolean
+    summary: When you include this key in the options dictionary, the method opens the URL only if the URL is a valid universal link and there is an installed app capable of opening that URL.
+    optional: true
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -426,7 +426,7 @@ properties:
 
   - name: UIApplicationOpenURLOptionsAnnotationKey
     type: Array
-    summary: contains the information passed to a document interaction controller object’s annotation property.
+    summary: contains the information passed to a document interaction controller object's annotation property.
     optional: true
 
   - name: UIApplicationOpenURLOptionsOpenInPlaceKey
--- a/apidoc/Titanium/UI/2DMatrix.yml
+++ b/apidoc/Titanium/UI/2DMatrix.yml
@@ -123,7 +123,7 @@ methods:
 
       - name: angle
         summary: |
-            Angle to rotate to, in degrees. On Android, if `angleTo` is specified, this specifies
+            Angle to rotate to, in degrees. On Android, if `toAngle` is specified, this specifies
             the starting angle for a rotation animation.
         type: Number
 
--- a/apidoc/Titanium/UI/2DMatrix.yml
+++ b/apidoc/Titanium/UI/2DMatrix.yml
@@ -291,7 +291,7 @@ properties:
         Point to rotate around, specified as a dictionary object with `x` and `y`
         properties, where { x: 0.5, y: 0.5 } represents the center of whatever is being
         rotated.
-    type: Dictionary
+    type: Point
     optional: true
     default: (0.5, 0.5)
     platforms: [android]
--- a/apidoc/Titanium/UI/Padding.yml
+++ b/apidoc/Titanium/UI/Padding.yml
@@ -1,6 +1,6 @@
 ---
-name: Padding
-summary: Dictionary object of parameters for the padding/insets applied to all kinds of views.
+name: HorizontalInsets
+summary: Dictionary object of parameters for horizontal-only insets applied to [Table](Titanium.UI.TableView) and [List](Titanium.UI.ListView) views. Only `left` and `right` properties are used (see <Padding>).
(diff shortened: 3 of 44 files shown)


sgtcoolguy:
cc @drauggres You might be interested in these changes

sgtcoolguy:
Relates to #11386

@janvennemann Perhaps you'd be best for review? I attempted to try and basically "refactor" our docs to use some commonly defined types for many of the "simple object which has a few properties" arguments/callbacks. The goal was to clean the docs, try and re-use common types when defining them, and hopefully to improve TypeScript definitions and IDE content assist indirectly.

apidoc/Titanium/UI/2DMatrix.yml
```diff
@@ -291,7 +291,7 @@ properties:
         Point to rotate around, specified as a dictionary object with `x` and `y`
         properties, where { x: 0.5, y: 0.5 } represents the center of whatever is being
         rotated.
-    type: Dictionary
+    type: Point
```
drauggres: Also should be changed in Matrix2D.

sgtcoolguy: good catch, just fixed this in next commit

apidoc/Titanium/UI/iOS/MenuPopup.yml
```diff
@@ -33,8 +33,8 @@ methods:
     summary: Hides the menu popup.
     parameters:
       - name: params
-        summary: Includes options how the menu popup should be hidden.
-        type: MenuPopupHideParams
+        summary: Includes options how the menu popup should be hidden. Introduced in SDK 5.2.0
+        type: AnimatedOptions
```
drauggres: Default was `{animated: true}`

apidoc/Titanium/UI/iPad/Popover.yml
```diff
@@ -39,14 +39,14 @@ methods:
     parameters:
       - name: options
         summary: Display properties to use when hiding the popover.
-        type: PopoverParams
+        type: AnimatedOptions
```
drauggres: Default was `{animated: true}`

apidoc/Titanium/UI/iPad/Popover.yml
```diff
@@ -39,14 +39,14 @@ methods:
...
 
   - name: show
     summary: Displays the popover.
     parameters:
       - name: params
         summary: Display properties to use when displaying the popover.
-        type: PopoverParams
+        type: ShowPopoverParams
```
drauggres: Default was `{animated: true}`

apidoc/Titanium/UI/Window.yml
```diff
@@ -1147,7 +1149,7 @@ properties:
     description: |
         This was separated from the <Titanium.UI.Window.androidback> event. You need to define this
         callback if you explicitly want to override the back button behavior.
-    type: Callback<Object>
+    type: Callback<void>
```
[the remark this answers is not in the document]
sgtcoolguy: yeah I believe it is in our doc format. `Callback` is synonymous with `Function`. As to the ` void` arg, well that is seen as ok in the validation. Otherwise, I'm not sure how to define that the function takes no arguments. I think the generators are ok with with `void` arg

apidoc/Titanium/UI/ScrollView.yml
```diff
@@ -54,11 +56,12 @@ methods:
       - name: contentOffsetXY
         summary: |
             X and Y coordinates to which to reposition the top-left point of the scrollable region.
-        type: Dictionary
+        type: Point
       - name: animated
-        summary: Determines whether the scrollable region reposition should be animated
-        type: contentOffsetOption
+        summary: A JS object with an `animated` property which determines whether the scrollable region reposition should be animated
+        type: AnimatedOptions
```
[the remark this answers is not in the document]
sgtcoolguy: that's true, but this method is iOS-only so the method and any arguments are therefore iOS-only.

apidoc/Titanium/UI/ScrollView.yml
```diff
@@ -67,9 +70,9 @@ methods:
         summary: Scaling factor of the scroll view's content.
         type: Number
 
-      - name: animated
-        summary: Determines whether the scrollable region reposition should be animated
-        type: zoomScaleOption
+      - name: options
+        summary: A JS object with an `animated` property which determines whether the scrollable region reposition should be animated
+        type: AnimatedOptions
```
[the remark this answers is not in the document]
sgtcoolguy: that's true, but this method is iOS-only so the method and any arguments are therefore iOS-only.

apidoc/Titanium/UI/iOS/MenuPopup.yml
sgtcoolguy: Yeah, I struggled with how to handle the cases where the default `animated` value is `true` vs `false`. Do we just mention it in the summary? Do I define a new sub-type of `AnimatedOptions` that just overrides the property's default value to be `true`?

apidoc/Titanium/UI/iPad/Popover.yml
sgtcoolguy: same as above

apidoc/Titanium/UI/iPad/Popover.yml
sgtcoolguy: same

apidoc/Titanium/UI/iOS/MenuPopup.yml
sgtcoolguy: To do some follow-up: These special "types" I've been extracting are really a better match as TypeScript interfaces, not classes. Because there's no actual way to instantiate an `AnimatedOptions` type - no such thing exists. It's more about the shape of a generic JS object we expect to be passed in to the methods. As such, the conversion to TypeScript would drop any "default" values on objects like this. It's more about documenting the property names, whether they're optional/required, and if they're "readonly". So trying to set the default value on the type may actually be moot - and I may need to just be thorough in mentioning the assumed defaults on all the methods that accept it. 😢

sgtcoolguy: Just thought of another possible way to document this - set the default value on the method argument to the raw JS object that it "assumes". Old docs already do this for things like Point (i.e. `{ x: 0, y: 0 }`) We've been pretty loosey-goosey with how we handle `default` though. Not sure if we'd need more rules around it or just assume that the string value is the raw JS code/value that is assumed to be default. (How would we validate that, since I know we do have cases where we have sentences there?)

drauggres: At least we should keep this information in `summary` or `description` (same for iOS-only properties).

apidoc/Titanium/UI/Padding.yml
```diff
@@ -0,0 +1,31 @@
...
+    summary: Top padding/inset
+    default: 0
+
+  - name: bottom
+    type: Number
+    summary: Bottom padding/inset
+    default: 0
+
+---
+# TODO: Should we invert the relationship here? (i.e. Padding extends from HorizontalInsets?)
```
janvennemann: I'd go this way around. The concept of `excluding` things is just not how inheritance works IMHO.

apidoc/Titanium/UI/View.yml
```diff
@@ -1645,7 +1660,7 @@ properties:
         The correct values will only be available when layout is complete.
         To determine when layout is complete, add a listener for the
         [postlayout](Titanium.UI.View.postlayout) event.
-    type: Dimension
+    type: Dimension # TODO: Define an extension of Dimension with the android-only absolute properties!
```
janvennemann: +1

janvennemann (changes requested):
Only two minor notes, the rest looks good.

PS: Man this looks like some tedious work, awesome job!

apidoc/Titanium/Platform/Platform.yml
```diff
@@ -407,3 +407,34 @@ properties:
...
+        * `UIApplicationOpenURLOptionUniversalLinksOnly` (Boolean value)
+    Read more about the available keys in the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionskey?language=objc).
+properties:
+  - name: UIApplicationOpenURLOptionsSourceApplicationKey
+    type: String
+    summary: The value of this key is an String containing the bundle ID of the app that made the request. If the request originated from another app belonging to your team, UIKit sets the value of this key to the ID of that app. If the team identifier of the originating app is different than the team identifier of the current app, the value of the key is nil.
+    optional: true
+
+  - name: UIApplicationOpenURLOptionsAnnotationKey
+    type: Array
```
drauggres: Here must be a type of items stored in the array.

sgtcoolguy: Yeah... I couldn't find what the heck goes in here.

drauggres: FYI: with `type: Array` current version of the "typescript" generator will produce incorrect output.

PR #11393, opened 2019-12-13
docs: fix mistakes in apidocs

drauggres:


apidoc/Titanium/UI/iOS/ApplicationShortcuts.yml
```diff
@@ -97,6 +97,8 @@ methods:
         summary: |
             Use the `itemtype` property to determine which shortcut should be returned.
         type: String
+    returns:
+      type: ShortcutParams
```
sgtcoolguy: This looks good, but looking at the code, I think there's some discrepancy in the property names for the `ShortcutParams` type itself.

I'll try and address that in my doc PR here: #11391

PR #11408, opened 2019-12-22
feat(android): extend gradle.properties / local.properties

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27696

Also adds the generated Android Studio files to the .gitignore


android/cli/lib/gradle-wrapper.js
```diff
@@ -379,6 +406,14 @@ async function writeJavaPropertiesFile(filePath, properties) {
 
 	// Create the properties files with the text lines generated above.
 	await fs.writeFile(filePath, fileLines.join('\n') + '\n');
+
+	// Add own configuration is provided. Following a proper format is on own risk!
+	const fileName = filePath.split('/').pop();
+	const gradlePropertiesOverwriteFile = path.join(platformPath, fileName);
+
+	if (fs.existsSync(gradlePropertiesOverwriteFile)) {
```
sgtcoolguy: if we're being consistent about using async/await:
`if (await fs.exists(gradlePropertiesOverwriteFile)) {`

android/cli/lib/gradle-wrapper.js
```diff
@@ -379,6 +406,14 @@ async function writeJavaPropertiesFile(filePath, properties) {
...
+		await fs.appendFile(filePath, fs.readFileSync(gradlePropertiesOverwriteFile, 'utf-8').toString() + '\n');
```
sgtcoolguy: Again, for consistency:
```js
const overrideContents = await fs.readFile(gradlePropertiesOverwriteFile, 'utf-8'); // no need for toString() as specifying encoding enforces it returns a String
return fs.appendFile(filePath, `${overrideContents}\n`); // no need to await, can just return the Promise since this function is async
```

hansemannn:
@sgtcoolguy Updated

jquick-axway:
@hansemannn, I'm closing this PR in favor of PR #11428 

It's implemented in a similar way. It still appends the app project's `gradle.properties` to the generated one. But I would prefer to keep the `gradle-wrapper.js` script as generic as possible so that we can use it on non-Titanium app projects too.

PR #11411, opened 2019-12-28
feat(android): Add rotate event

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6462

Adding the `rotate` event to a view:
```JS
	backgroundColor: "#fff"
});
win.addEventListener("pinch", onPinch);
win.addEventListener("rotate", onRotate);

var ldi = Ti.Platform.displayCaps.logicalDensityFactor;
var v = Ti.UI.createView({
	width: 100,
	height: 100,
	backgroundColor: "#f00",
	touchEnabled: false
});

win.add(v);
win.open();


function onPinch(e) {
	v.scaleX = e.scale;
	v.scaleY = e.scale;
	v.left = e.focusX / ldi - 50;
	v.top = e.focusY / ldi - 50;
}

function onRotate(e) {
	console.log("rotate", e.rotate);
	v.rotation = e.rotate;
}
```
you can pinch and rotate the view with two fingers

https://user-images.githubusercontent.com/4334997/160287718-8af400d0-a26c-4bb8-904f-d2a4697c915a.mp4

Help from someone else please:
* parity for iOS


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1722,6 +1719,34 @@ public void onLongPress(MotionEvent e)
...
+			private boolean doRotationEvent(MotionEvent event)
+			{
+				// Calculate the angle between the two fingers
+				float deltaX = event.getX(0) - event.getX(1);
+				float deltaY = event.getY(0) - event.getY(1);
+				double radians = Math.atan(deltaY / deltaX);
+				// Convert to degrees
+				int degrees = (int) (radians * 180 / Math.PI);
+
+				switch (event.getActionMasked()) {
```
garymathews: Maybe simplify the switch statement a little?
```Java
switch (event.getActionMasked()) {
    case MotionEvent.ACTION_MOVE:
        data.put(TiC.PROPERTY_ROTATE, degrees);
        data.put(TiC.EVENT_PROPERTY_SOURCE, proxy);
        fireEvent(TiC.EVENT_ROTATE, data);
        break;
    default:
        break;
}
```

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1722,6 +1719,34 @@ public void onLongPress(MotionEvent e)
...
+						// rotate
+						KrollDict data = new KrollDict();
+						data.put(TiC.PROPERTY_ROTATE, degrees);
+						data.put(TiC.EVENT_PROPERTY_SOURCE, proxy);
+						fireEvent(TiC.EVENT_ROTATE, data);
+
+						break;
+				}
+
+				return true;
```
garymathews: Since this method never returns anything other than `true`, this should probably be a `void`

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1722,6 +1719,34 @@ public void onLongPress(MotionEvent e)
...
 			int pointersDown = 0;
 
+			private boolean doRotationEvent(MotionEvent event)
+			{
+				// Calculate the angle between the two fingers
+				float deltaX = event.getX(0) - event.getX(1);
+				float deltaY = event.getY(0) - event.getY(1);
+				double radians = Math.atan(deltaY / deltaX);
+				// Convert to degrees
+				int degrees = (int) (radians * 180 / Math.PI);
```
garymathews: Could use [Math.toDegrees()](https://docs.oracle.com/javase/7/docs/api/java/lang/Math.html#toDegrees(double))

garymathews (changes requested):
Thanks for the PR @m1ga! Minor changes requested. I'll make a ticket for parity with iOS too.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
m1ga: i think a simple if would even be better. I've left the other parts in in case we'll need a rotate start event but I think it's not needed for this issue

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
m1ga: :+1: will do these changes tomorrow. Thanks for the feedback

m1ga:
closing in favor of https://github.com/tidev/titanium-sdk/pull/13882

PR #11412, opened 2020-01-02
fix(android): improve memory handling of TiBlob image methods

garymathews:
- Allow GC to occur after memory intensive TiBlob image manipulation methods

##### TEST CASE
```JS
const img = Ti.UI.createImageView();
 
win.addEventListener('open', async () => {
 
    // Obtain large image blob. (8205px, 11750px)
    let blob = await getImage('https://tinyurl.com/rxob7ju');
 
    // Keep re-sizing the image down by 10%
    for (let i = 0; i < 10; i++) {
 
        // De-reference original blob so it can be freed.
        blob = blob.imageAsResized(blob.width / 1.1, blob.height / 1.1);
 
        // Idle for 100ms, allowing V8 GC to collect.
        // QE: Should not need to do this to display image.
        // await new Promise(resolve => setTimeout(resolve, 100));
    }
 
    // Display re-sized image.
    img.image = blob;
});
 
win.add(img);
win.open();
 
function getImage(url) {
    return new Promise((resolve, reject) => {
       const client = Ti.Network.createHTTPClient({
            onload(e) {
                resolve(e.source.responseData);
            },
            onerror: reject,
            timeout : 10000
        });
        client.open('GET', url);
        client.send();
    });
}
```
- Should see re-sized image displayed

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27695)


android/runtime/common/src/java/org/appcelerator/kroll/KrollRuntime.java
```diff
@@ -114,13 +114,38 @@ public static KrollRuntime getInstance()
...
 	{
 		if (instance != null) {
 			instance.setGCFlag();
 		}
 	}
 
+	/**
+	 * Force V8 garbage collection.
+	 */
+	public static void softGC()
```
sgtcoolguy: I'm not sure the naming here is very apt. This looks to be more like `forceJsGC()` and `forceJsAndJavaGC()`

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -786,14 +791,18 @@ public TiBlob imageAsResized(Number width, Number height)
...
 		} catch (IllegalArgumentException e) {
 			Log.e(TAG, "Unable to resize the image. Illegal Argument: " + e.getMessage(), e);
-			return null;
 		} catch (Throwable t) {
 			Log.e(TAG, "Unable to resize the image. Unknown exception: " + t.getMessage(), t);
-			return null;
+		} finally {
+			// Perform soft garbage collection to reclaim memory.
+			KrollRuntime instance = KrollRuntime.getInstance();
+			if (instance != null) {
```
sgtcoolguy: All these null checks remind me of the null object pattern. I know it's super common to check for null in Java and we have null checks everywhere, but does KrollRuntime.getInstance() ever realistically return null? If so, maybe have it return a dummy NullRuntime that has no-ops for things like `softGC()`? If not, mark the method as @NonNull and avoid the checks? Looks to me the only way it could be null is if somehow it got called before `init()` (though certainly it looks like there's som minimal code paths we hit before that happens, like the asset helper/cache stuff.

tests/ti.blob.addontest.js
```diff
@@ -0,0 +1,69 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2020 by Axway, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
```
sgtcoolguy: I think we can officially stop writing ES5 style code now in our test suite. Please use `let`/`const`

tests/ti.blob.addontest.js
```diff
@@ -0,0 +1,69 @@
...
+
+describe('Titanium.Blob', function () {
+	var win;
+
+	afterEach(function (done) {
+		if (win) {
+			// If `win` is already closed, we're done.
```
sgtcoolguy: Totally unrelated to this PR, but every time I see this in our test suite it makes me shake my head: Why do we not have any property/method to inquire about the current state of a window/view? We should be able to know if a window is closed or not.

tests/ti.blob.addontest.js
```diff
@@ -0,0 +1,69 @@
...
+	it('resize very large image', function (finish) {
+
+		win = Ti.UI.createWindow({ backgroundColor: 'gray' });
+		const img = Ti.UI.createImageView();
+
+		// Obtain large image blob. (8000px, 8000px)
+		let blob = Ti.Filesystem.getFile('large.jpg').read();
+		should(blob).be.an.Object;
+
+		win.addEventListener('open', async () => {
```
sgtcoolguy: no need for `async` - there's no `await` usage and event listeners have no expected return type (of `Promise` or anything else)

tests/ti.blob.addontest.js
```diff
@@ -0,0 +1,69 @@
...
+			win.addEventListener('close', function listener () {
+				clearTimeout(t);
+
+				if (win) {
+					win.removeEventListener('close', listener);
+				}
+				win = null;
+				done();
+			});
+			win.close();
```
sgtcoolguy: It also makes me think - I know traditionally our APIs are very event-based for async operations, but we might want to consider starting to add optional callback args at the end of methods that fire these async events as a different, more Node-style, means of interacting with them. i.e.
```javascript
win.close(() => done());
```
It's certainly less verbose for simple operations and avoids the need to remove listeners. (Note that Node's EventEmitter does have one-shot listener APIs, but we never really make use of them in our examples so who knows if they actually work as intended on our proxies - `win.once('close', () => {});`)

sgtcoolguy:
cc @brentonhouse this looks to be a potential "fix"/improvement for the workflow/use case you mentioned in the internal app.

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
garymathews: To be honest, if `getInstance()` were `null` then something bad happened. But technically, it is possible for it to be `null`.

I'm going to set it to `@Nullable` and leave a TODO for future reference so we can clean up our codebase.

tests/ti.blob.addontest.js
garymathews: I agree, but I'm going to leave this as is since this is just the _addon_ template. The test case is ES6+, which will be merge into our test suite.

tests/ti.blob.addontest.js
garymathews: That would be nice! Or even better:
```JS
win.close().then(_ => {
    // Window closed.
});
```

ssaddique:
FR: **Pass**

PR #11417, opened 2020-01-08
fix(android): app builds fail with uppercase module JARs on case-sensitive systems

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27706

**Note:**
This is mostly an issue on Linux, but can be reproduced on macOS. Titanium's only module that has an uppercase letter is "ti.facebook", which contains a "Facebook.jar".

This is not an issue with the module id. It is an issue with module "name" set in its "manifest" file.

**Test:**
Following the test procedure in [TIMOB-27706](https://jira.appcelerator.org/browse/TIMOB-27706) and verify that the app no longer crashes when accessing Facebook in kitchensink-v2.


sgtcoolguy (approved):
fix looks good. 

May want to explicitly add comment/note that when building modules we are generating the json file with a forced lowercase name from the manifest, but when we generate the module/JAR/project we retain the case from the manifest. (or something to link exactly why this change/fix was required and case sensitivity is important here, maybe even a pointer to the code in _buildModule.js where we're passing the name as-is and converting to lowercase - or even better yet, find a way to extract out a common function to enforce the proper name/case is used in both building and consumption of modules!).

ssekhri:
FR passed.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20200107215936
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516

ssekhri:
Waiting for pending CR before merge

jquick-axway:
@sgtcoolguy, I'll change the code to not lowercase the JSON files too. The externalized JSON is a new thing in 9.0.0 so it's okay. Older built modules embedded the JSON within the JAR, which I got rid of because it was bloating the built APK (typically by a couple megs).

jquick-axway:
Updated PR to not lowercase binding JSON file.

Note that none of our released modules produce this JSON file yet. This is something we're introducing for 9.0.0 built modules. So, there's nothing to test yet (but I've tested it on my end already).

ssekhri:
FR passed with latest changes.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20200108134222
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516

PR #11419, opened 2020-01-09
fix(ios): expose TiApp singleton accessor to swift

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27622

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
@@ -137,7 +137,7 @@ TI_INLINE void waitForMemoryPanicCleared() //WARNING: This must never be run on 
 /**
  Returns singleton instance of TiApp application object.
  */
-+ (TiApp *)app;
++ (TiApp *)app NS_SWIFT_NAME(sharedApp());
 
 /**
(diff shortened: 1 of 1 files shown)


hansemannn:
Using methods for shared swift objects is dicouraged. ✌️

janvennemann:
@hansemannn Any recommendation how to achieve this, since we are talking about accessing Objective-C class in Swift here? If this would be a Swift class to begin with we would obviously just have a `static let shared = TiApp()` to have the `shared` property. However, since in Objective-C the singleton accessor is `+ (TiApp *)app` is there an easy way to map this to a Swift property?

The underlying issue here was that the Swift compiler mistakingly identified the `+ (TiApp *)app` method as a class factory method, resulting in a bunch of compiler errors when trying to get the singleton instance of `TiApp` in Swift.

vijaysingh-axway:
I don't think there is any issue with this implementation. What Hans has suggested is best practice for swift. As our SDK is written in Objective C, so APIs are written considering the Objective C standard.

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.1
SDK: 8.3.1.v20200110080957
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Xcode: 11.2

PR #11425, opened 2020-01-09
feat: extend global console API to be more Node-compatible

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26572

**Description:**
This moves the `global.console` definition entirely into the `common` JS-layer so that it is defined in a cross-platform way. The removes the JS definition baked into Android's runtime JS, and removes the native Obj-c console on iOS.

Our global console object receives the `Ti.API` module (so that we can call `info`, `warn`, `error` and `debug` on it). Note that native logging and Ti.API have the concept of log levels, whereas node console really treats log/info/debug as just going to `stdout` and warn/error as going to `stderr`. So we have to try and retain the log level notion when possible (we sniff for being constructed with an object that has a property `apiName === 'Ti.API'`, which allows "hijacking" the actual underlying writes in JS tests to verify the strings passed along). It also supports NodeJs stream-like objects in the constructor (but then loses the log level concept - instead passing log/info/debug to `stdout` and warn/error to `stderr`)

Note that we don't have any real shim API for NodeJS streams yet, though (so I couldn't say test a console that actually wrapped a file stream to write to files).

Oh, an important thing to note as well: *I did not try to implement `console.table()`. It seemed... extensive.*


common/Resources/ti.internal/extensions/js/console.js
```diff
@@ -1,30 +1,156 @@
...
-console.debug = function (...args) {
-	nativeDebug.call(console, formatWithOptions(kColorInspectOptions, ...args));
-};
+class Console {
+	constructor(options) {
+		// Allow passing in a fake Ti.API that we can hijack the info/warn/error/debug calls!
+		if (options && options.apiName === 'Ti.API') {
+			this._apiModule = options;
+		} else {
+			// if (!options || typeof options.write === 'function') {
```
janvennemann: It might be good to spit out a warning / error here since this will practically create a broken `console` object which will lead to a crash when used due to `_apiModule` being undefined.

janvennemann (changes requested):

common/Resources/ti.internal/extensions/js/console.js
```diff
@@ -1,30 +1,160 @@
...
+		const warned = logTime(this, label);
+		if (!warned) {
+			this._times.delete(label);
+		}
+	}
+
+	timeLog(label = 'default', ...logData) {
+		logTime(this, label, logData);
+	}
+	// TODO: console.table()
```
janvennemann: I was wondering if we should Implement `table()` as a no-op, printing a one-time warning that it is not implemented yet. Just to be API compatible and reduce possible crashes with third-party modules that actually try to use it.


Something like this:
```js
table() {
  if (!table.warned) {
    table.warned = true;
    process.emitWarning('"console.table" is not yet implemented in Titanium!');
  }
}
```

common/Resources/ti.internal/extensions/js/console.js
```diff
@@ -1,30 +1,160 @@
...
-console.log = function (...args) {
-	nativeLog.call(console, formatWithOptions(kColorInspectOptions, ...args));
-};
+		// Support wrapping Ti.API (which retains log level)
+		if (this._apiModule) {
+			this._apiModule[level](string);
+		} else {
+			// Support Node.JS streams like stdout/stderr which don't have log levels
+			const stream = (level === 'warn' || level === 'error') ? this._stderr : this._stdout;
+			// TODO: Handle this._ignoreErrors by doing try/catch/finally, hanging error handlers
```
janvennemann: This sounds rather trivial and i think should be added since the default value for `ignoreErrors` is `true`.

janvennemann (changes requested):
Just two notes about TODO items that i think can be quickly added.

PR #11427, opened 2020-01-09
fix(ios): server is receiving two consecutive calls for the same url and cookies updated while reloading webview

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27623


ssekhri:
FR Passed.
The beforeLoad is called only once at the time of webview being open. Also the cookies are updated during reload.
Verified on:
Mac OS: 10.15.1
SDK: 8.3.1.v20200109101955
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516
Xcode: 11.2
Device: iOS simulator 13.2

janvennemann (approved):
LGTM!

PR #11434, opened 2020-01-15
feat(android): target Java8 and Kotlin support for native modules

ypbnv:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-25896
- https://jira.appcelerator.org/browse/TIMOB-27472

**Description:**
Targets Java 8 and adds support for Kotlin for developing native android modules.

**Test case:**
1. Download `ti.kotlin.zip` attached to [TIMOB-25896](https://jira.appcelerator.org/browse/TIMOB-25896).
2. Unzip it.
3. Open the terminal.
4. `CD` to directory: `./ti.kotlin/android`
5. Run: `appc run -p android`
6. Above builds the module and builds/runs its "example" app.
7. Verify example app shows a white background with a red square.
8. Verify that you see the following messages in the log.
```
[INFO]   TiKotlinModule: (main) [0,106] @@@ Java8Interface.printMessage1() called.
[INFO]   Java8: (main) [0,106] @@@ Java8Interface.printMessage2() called.
[INFO]   ExampleProxy: (main) [3,109] @@@ Example creation property 'message' set to: Creating an example Proxy
[INFO]   ExampleProxy: (main) [0,109] @@@ Example.printMessage() called: Hello world!
[INFO]   ExampleProxy: (main) [0,109] @@@ Example.setMessage() called: Hi world!.  It's me again.
[INFO]   ExampleProxy: (main) [0,109] @@@ Example.printMessage() called: Hello world!
```

### DIFF
--- a/android/templates/module/generated/build.gradle
+++ b/android/templates/module/generated/build.gradle
@@ -120,7 +120,7 @@ android {
 			}
 		}
 	}
-	// TODO: Add Java 8 support in the future.
+
 	compileOptions {
 		sourceCompatibility JavaVersion.VERSION_1_8
 		targetCompatibility JavaVersion.VERSION_1_8
@@ -130,6 +130,10 @@ android {
 	packagingOptions {
 		exclude '/lib/**/libc++_shared.so'
 		exclude '/lib/**/libkroll-v8.so'
+	}
+	// Sets the JVM target in order to allow usage of Java 8 features in Kotlin native modules.
+	kotlinOptions {
+		jvmTarget = '1.8'
 	}
 }
 
--- a/android/templates/module/generated/build.gradle
+++ b/android/templates/module/generated/build.gradle
@@ -1,6 +1,8 @@
 
 apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
 
 repositories {
 	maven {
@@ -120,8 +122,8 @@ android {
 	}
 	// TODO: Add Java 8 support in the future.
 	compileOptions {
-		sourceCompatibility JavaVersion.VERSION_1_7
-		targetCompatibility JavaVersion.VERSION_1_7
+		sourceCompatibility JavaVersion.VERSION_1_8
+		targetCompatibility JavaVersion.VERSION_1_8
 	}
 	// Exclude the C++ runtime and Titanium kroll libraries for all architectures when packaging the AAR file.
 	// Avoids app build failure due to library file name collision. (Apps will use "titanium.aar" bundled *.so files.)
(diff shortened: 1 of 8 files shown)


android/templates/module/generated/build.gradle
```diff
@@ -120,8 +122,8 @@ android {
 	}
 	// TODO: Add Java 8 support in the future.
 	compileOptions {
-		sourceCompatibility JavaVersion.VERSION_1_7
-		targetCompatibility JavaVersion.VERSION_1_7
+		sourceCompatibility JavaVersion.VERSION_1_8
+		targetCompatibility JavaVersion.VERSION_1_8
```
jquick-axway: Can you remove the `// TODO:` comment please?

android/titanium/build.gradle
```diff
@@ -105,8 +105,8 @@ android {
 	}
 	// TODO: Add Java 8 support in the future.
 	compileOptions {
-		sourceCompatibility JavaVersion.VERSION_1_7
-		targetCompatibility JavaVersion.VERSION_1_7
+		sourceCompatibility JavaVersion.VERSION_1_8
+		targetCompatibility JavaVersion.VERSION_1_8
```
jquick-axway: Can you remove the `// TODO:` comment please?

android/templates/module/generated/build.gradle
jquick-axway: Do we need to do the below so that Kotlin can support Java 8?
Google's docs suggest that we do. Would you mind confirming please?
   https://developer.android.com/studio/write/java8-support

```groovy
android {
	kotlinOptions {
		jvmTarget = '1.8'
	}
}
```

jquick-axway (changes requested):

android/templates/module/generated/build.gradle
ypbnv: I have totally missed to add that. I ran a double check and yes - it is required if we want to use Java 8 features in Kotlin files. I did a test with defining a static method with a body in an interface class and access it from a native module. Without this option it resulted in the following exception:
`Kotlin: Calls to static methods in Java interfaces are prohibited in JVM target 1.6. Recompile with '-jvm-target 1.8'` 
After adding them the module compiled fine and the method call was successful.

ypbnv:
@jquick-axway Updated.

jquick-axway (approved):
CR: Pass

android/templates/module/generated/build.gradle
```diff
@@ -158,12 +164,15 @@ dependencies {
 	def krollAptJarPath = '<%- krollAptJarPath.replace(/\\/g, '\\\\') %>'
 	annotationProcessor files(krollAptJarPath)
 	compileOnly files(krollAptJarPath)
+	kapt files(krollAptJarPath)
```
jquick-axway: @ypbnv, can you add the following to the module's "build.gradle" file above the `dependencies{}` section please?
```groovy
// Titanium does not support incremental @Kroll annotation processing. Disable it in kotlin to avoid build warnings.
project.ext['kapt.incremental.apt'] = false
```

This gets rid of the following build warning...
```
w: [kapt] Incremental annotation processing requested, but support is disabled because the following processors are not incremental: org.appcelerator.kroll.annotations.generator.KrollJSONGenerator (NON_INCREMENTAL).
```

Alternatively, you can set `kapt.incremental.apt` via the "gradle.properties" file using our "_buildModule.js", but doing it via a "build.gradle" file like I'm doing above works too.

ypbnv: Updated

ssekhri:
FR Passed. The test case for Java8 and Kotlin work as expected on both mac and windows OS.
Verified on:
Mac OS: 10.15.1, Win 10 Pro
SDK: 9.0.0.v20200122065826
Appc CLI: 7.1.2
JDK: 11.0.4, 1.8.0_162
Node: 10.16.3

jquick-axway:
@ypbnv, is the "Allow edits from maintainers" unchecked for this PR?
We're unable to "Update" it with "master".
(I don't have permission to update this PR from the command line either.)

PR #11440, opened 2020-01-17
fix(ios): added proper condtion to import MediaPlayer

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27106


janvennemann (approved):
LGTM!

PR #11441, opened 2020-01-19
fix(android): properly set ImageView tintColor

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27721


ypbnv (approved):
Thanks for the PR!
Looks good to me!

jquick-axway:
I've confirmed that this PR's code change makes Android's tint behave exactly like iOS' for both masked images and completely opaque images (such as camera photos).

---
**Side Note:**
The only thing I don't like is iOS and Android's tinting behavior of opaque images (no alpha; such as photos). I would expect a "tint" to apply a color blend effect like how it works in Photoshop. But instead you see a solid color (the tint color). But the only way to achieve this color blending effect is to grayscale the image first, then apply the tint color.
_(I don't think this should stop this PR. This is how it works on iOS. I just don't like the behavior.)_

hansemannn:
@jquick-axway I made [titanium-image-filters](https://github.com/hansemannn/titanium-image-filters) exactly for that :). Or you can used masked image filters that are available in the SDK as well.

hansemannn:
@garymathews The unit test failures seem unrelated, but maybe critical: `Error: expected Object {} to be '#ff1f1f'`

ssjsamir (approved):
**FR Passed:** Able to see the tint applied to the following test case (https://jira.appcelerator.org/browse/TIMOB-27721) : 

```
const icon = Ti.UI.createImageView({
tintColor: 'red',
width: 30,
image: 'example.png'
});
window.add(icon);
window.open();
```

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
Pixel Xl 7.1.1 Sim
```

PR #11446, opened 2020-01-24
fix(ios): behaviour of toString function of TiBlob fixed

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27707


janvennemann (commented):
I'd really like to know why this hack is necessary (and why it only happens on iOS 10), but i don't have the time to look into it right now. 

@sgtcoolguy you did the TiBlob rewrite to the new ObjC JavaScriptCore API and added this extension, any idea why the prototype override added here vanishes after some time? I guess it has to do with the way our KrollObjects are created and that the `constructor.prototype` is getting GC'd at some point. Or maybe it's just a bug in JSC on iOS 10?

vijaysingh-axway:
> I'd really like to know why this hack is necessary (and why it only happens on iOS 10), but i don't have the time to look into it right now.
> 
> @sgtcoolguy you did the TiBlob rewrite to the new ObjC JavaScriptCore API and added this extension, any idea why the prototype override added here vanishes after some time? I guess it has to do with the way our KrollObjects are created and that the `constructor.prototype` is getting GC'd at some point. Or maybe it's just a bug in JSC on iOS 10?

@janvennemann This problem is happening on iOS 10.x only. iOS 11+ it is working fine. The hack ensures the overwritten constructor be alive. I guess it may be something to do with JavaScriptCore in iOS 10. Otherwise this problem should be on iOS 11+ as well.

janvennemann (approved):
@vijaysingh-axway Yeah, that's indeed really strange. Let's go with this workaround for now and address this later should we gain any more insight.

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.1
SDK: 9.0.0.v20200131101019
Appc CLI: 7.1.2
JDK: 11.0.4
Node: 10.16.3
Studio: 6.0.0.201911251516
Xcode: 11
Device: iPhone XR(v13.2.2), iPhone 5C(v10.3.3),

PR #11447, opened 2020-01-24
feat: include app root node_modules folder

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26352

**Description:**
This PR modifies our CLI build process to include/handle a project's root `node_modules` folder into the app. 

This PR adds support for including production npm packages in the app's root `node_modules` folder. Note that this doesn't attempt to manage that folder, it's expected a user will run `npm install` or `npm ci` themselves outside the course of the `ti` commands.

As part of this PR I am also trying to refactor both iOS/Android build codebases to extract out commonalities, move to `async`/`await` and use the incremental state management built into `appc-tasks` - in particular the iOS build is littered with cache related code that obscures the underlying workflow. Moving to the appc-tasks and having that more or less transparently handled makes life simpler.


brentonhouse:
@sgtcoolguy  - A couple of things to note.  I looked through the code and might just be missing them.

- Excluding native modules from being copied (https://github.com/brentonhouse/titanium-module-copier/blob/1521d6421295c7fbc21244d12550d1e6e13392d3/index.js#L126-L130)
- Exclude anything with `titanium.ignore` from being copied . (https://github.com/brentonhouse/titanium-module-copier/blob/1521d6421295c7fbc21244d12550d1e6e13392d3/index.js#L122-L124)
- Exclude anything with titanium.type === 'widget' .  (https://github.com/brentonhouse/titanium-module-copier/blob/1521d6421295c7fbc21244d12550d1e6e13392d3/index.js#L131-L142)
- Including peerDependencies . (https://github.com/brentonhouse/titanium-module-copier/blob/1521d6421295c7fbc21244d12550d1e6e13392d3/index.js#L117-L119)

brentonhouse:
Also, I took a look at the node_modules in the build directory and it looks like everything copied using `@titanium/module-copier` gets minified, source mapped, etc.

sgtcoolguy:
@brentonhouse The issue with the `@titanium/module-copier` w/ hook approach right now is that it leads to extra file I/O (and you have to make sure you're hooking at the right point in the build cycle so that it does get processed). In all likelihood, I think it makes sense to do a PR on it to export more than just the one sync copying method so that you can use the library to give you the directory set back (which is what I did in the local copy here) and then the SDK CLI build can use that to get the input set.

sgtcoolguy:
Sounds like #11346 may make this moot (at least going forward if people transition to using Webpack)

brentonhouse:
> Sounds like #11346 may make this moot (at least going forward if people transition to using Webpack)

While webpack is cool, there are a lot of breaking changes in it and you can probably expect a good percentage of enterprises and devs that won't be able to or choose not to use webpack.

sgtcoolguy:
This PR has a change to force it to use a specific branch of the test suite that verifies it's behavior. The change to the Jenkinsfile will need to be reverted/removed before merge. Additionally, after this is merged, the change to the test suite should likely be merged over on master to ensure we continue to test root-level package.json usage.

sgtcoolguy:
Open questions/tasks:
  - handling of alloy widgets
  - turn off most/some of this processing for webpack projects?
  - Should `devDependencies` sometimes be included (i.e. if deploy type is test/development)?
    - Specifically, what if a developer wants to use something like mocha/should/jest/chai to test their app via sim/device, but not include in production builds?

garymathews (changes requested):

garymathews (approved):
Android CR: **PASS**

vijaysingh-axway (approved):
iOS changes looks good.

PR #11448, opened 2020-01-25
chore: update ti.identity module

garymathews:
- Update `ti.identity`
  - Rebuild for Titanium SDK 9.0.0

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27726)


jquick-axway (approved):
CR: Pass

sgtcoolguy:
@garymathews I know there's a bit of a chicken and egg problem here, but PRs that need to be built against yet-unreleased versions of the SDK can point to builds off master in the Jenkinsfile to do so. When we're updating all these modules to 9.0.0/AndroidX/gradle we should also be updating the sdkVersion used to build/test them to a build off master. Then once 9.0.0.GA goes out we can point to that.

jquick-axway (dismissed):
We need to update "modules.json" to reference v3.0.1 of this module.

PR #11462, opened 2020-02-04
fix(android): strip xmlns definitions from child elements in AndroidManifest.xml

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27746

**Note:**
The damage has already been done to the 9.0.0 built modules "ti.map" and "ti.playservices" since this issue is already in their "AndroidManfiest.xml" files packaged within their AARs. We will need to rebuild the ticket's listed modules after merging this PR.

### DIFF
--- a/android/cli/lib/android-manifest.js
+++ b/android/cli/lib/android-manifest.js
@@ -75,8 +75,8 @@ class AndroidManifest {
 			let startReplaceIndex = text.indexOf('<manifest');
 			if (startReplaceIndex >= 0) {
 				startReplaceIndex = text.indexOf('>', startReplaceIndex);
-				text = text.replace(/xmlns:android=".*?"/g, (match, p1) => {
-					return (p1 <= startReplaceIndex) ? match : '';
+				text = text.replace(/\sxmlns:android=".*?"/g, (match, offset) => {
+					return (offset <= startReplaceIndex) ? match : '';
 				});
 			}
 		}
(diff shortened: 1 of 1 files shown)


android/cli/lib/android-manifest.js
```diff
@@ -67,7 +67,18 @@ class AndroidManifest {
...
 		if (this._xmlDomDocument && this._xmlDomDocument.documentElement) {
+			// Write XML content to string.
 			text += this._xmlDomDocument.documentElement + os.EOL;
+
+			// Remove all "xmlns:android=<url>"" namespace definitions from child elements.
+			// Google only allows it in root <manifest/> element or else a build/validation error will occur.
+			let startReplaceIndex = text.indexOf('<manifest');
+			if (startReplaceIndex >= 0) {
+				startReplaceIndex = text.indexOf('>', startReplaceIndex);
+				text = text.replace(/xmlns:android=".*?"/g, (match, p1) => {
```
garymathews: Minor, but since there are no capture groups the arguments should be `(match, offset) => {`

garymathews: Minor, the regex could be amended to remove the duplicate space:
```
/\sxmlns:android=".*?"/g
```

jquick-axway: Good idea. I'll go ahead and do it. Thanks.

garymathews (approved):
CR: **PASS** 👍

PR #11464, opened 2020-02-05
docs: fix global methods and properties

drauggres:
In this PR:
 - `Console` and `JSON` describe interfaces, whereas `console` and `JSON` (yeah, same name) are properties of the global object (this will allow to define them in the same way as in TypeScript's [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts)).
 - second parameter of `setTimeout`, `setInterval` and first parameter of `clearTimeout`, `clearInterval` were made optional.

Also worth to mention, that parameters of `console` methods (`log`/`error` etc.) actually are "varargs", but there is no (established) way to describe this in docs.

### DIFF
--- a/apidoc/Global/Console/Console.yml
+++ b/apidoc/Global/Console/Console.yml
@@ -1,5 +1,5 @@
 ---
-name: Global.console
+name: Console
 summary: Console logging facilities.
 description: |
     The toplevel `console` support is intended to supplement <Titanium.API>
@@ -88,7 +88,7 @@ methods:
   - name: timeEnd
     summary: Stop a timer that was previously started.
     description: |
-          Stop a timer that was started by calling [console.time](Global.console.time), and
+          Stop a timer that was started by calling [console.time](Console.time), and
           output the time since the timer was begun to the console in milliseconds.
           If no timer exists a warning will be logged to the console.
     parameters:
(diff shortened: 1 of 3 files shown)


apidoc/Global/Console/Console.yml
```diff
@@ -88,7 +88,7 @@ methods:
   - name: timeEnd
     summary: Stop a timer that was previously started.
     description: |
-          Stop a timer that was started by calling [console.time](Global.console.time), and
+          Stop a timer that was started by calling [console.time](Console.time), and
```
drauggres: I am not sure about this links and don't know how to check if they are valid.

sgtcoolguy (approved):
LGTM

sgtcoolguy:
As for varargs, I'm actually not entirely sure how we're supposed to handle them in our yml format. My best guess is the use of the `repeatable` property on a parameter? https://github.com/appcelerator/docs-devkit/blob/3b1c537d29e8c6ac90eaf711e74e52dd225ec4fe/packages/titanium-docgen/validate.js#L134

That's used in some of the generators, and typically results in the "name" shown in whatever html is generated as `name...` rather than the typical `name`. Notably, it doesn't appear to be consulted in the jsduck generator that's actually used to generate our current doc site (indirectly).

cc @janvennemann @drauggres

drauggres:
> As for varargs, I'm actually not entirely sure how we're supposed to handle them in our yml format. My best guess is the use of the `repeatable` property on a parameter? 

Found https://jira.appcelerator.org/browse/TIMOB-9799 and [TDoc+Specification](https://wiki.appcelerator.org/display/guides2/TDoc Specification#TDocSpecification-MethodParameterSpecification)

drauggres:
Something went wrong with the html-docs: https://docs.appcelerator.com/platform/latest/#!/api/Global.console
(maybe because the windows platform is still there)

PR #11468, opened 2020-02-10
docs: add "repeatable" property on methods parameters

drauggres:
https://github.com/appcelerator/titanium_mobile/pull/11464#issuecomment-583501478

The only method I known that still has undocumented "vararg" is `Ti.Database.DB.executeAsync`.


sgtcoolguy (approved):
LGTM

PR #11469, opened 2020-02-10
fix(ios): skip compiled plist lookup for watchOS framework compatibility

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27756


cb1kenobi (approved):
This makes sense. We've never seen an example with an extension with a framework, so I can see the problem. **APPROVED**!

PR #11479, opened 2020-02-17
docs: remove JSON from documentation

drauggres:
The global `JSON` object is part of ECMA-262 specification ([5th edition](https://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205th%20edition%20December%202009.pdf)).

@janvennemann https://github.com/appcelerator/docs-devkit/pull/25#issuecomment-586366525:
> Agreed, we should remove `JSON` from our docs since it's part of the ECMA spec. Users can refer to [JSON](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/JSON) on MDN for example.

Related ticket: https://jira.appcelerator.org/browse/TIDOC-211 (15/Apr/2011)


janvennemann (approved):

janvennemann (changes requested):
Ah damn, totally missed this. The `description` property in `Global.yml` still contains a reference to the JSON docs. Please remove that section as well.

janvennemann (approved):

PR #11486, opened 2020-02-19
fix(android): hyperloop build fails if NDK not installed

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27771

**Test:**
1. Use a machine that does not have the Android NDK set up or configured.
2. Download the [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples) project.
3. Update "tiapp.xml" hyperloop module version to `5.0.1`.
4. Build and run on Android.
5. Verify it builds and runs successfully. _(It used to fail to build.)_

### DIFF
--- a/support/module/packaged/modules.json
+++ b/support/module/packaged/modules.json
@@ -63,8 +63,8 @@
 	},
 	"hyperloop": {
 		"hyperloop": {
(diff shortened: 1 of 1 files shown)


lokeshchdhry:
FR Passed.

Hyperloop build is successful with no NDK installed.

Module Ver: 5.0.1

PR #11489, opened 2020-02-20
feat(ios): add search bar token API

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27773


iphone/Classes/TiUISearchBarProxy.m
```diff
@@ -184,9 +184,52 @@ - (id)barColor
...
+
+  searchToken.representedObject = [TiUtils stringValue:@"identifier" properties:token def:[TiUtils createUUID]];
+
+  [[[self searchBar] searchTextField] insertToken:searchToken atIndex:index];
+}
+
+- (void)removeTokenAtIndex:(id)index
+{
+  ENSURE_SINGLE_ARG(index, NSNumber);
+  [[[self searchBar] searchTextField] removeTokenAtIndex:[TiUtils intValue:index]];
```
vijaysingh-axway: It will crash for iOS < 13. Apply check.
```suggestion
  if (![TiUtils isIOSVersionOrGreater:@"13.0"]) {
     return;
   }
  [[[self searchBar] searchTextField] removeTokenAtIndex:[TiUtils intValue:index]];
```

iphone/Classes/TiUISearchBarProxy.m
```diff
@@ -184,9 +184,52 @@ - (id)barColor
...
+}
+
+- (NSArray<NSDictionary<NSString *, NSString *> *> *)tokens
+{
+  NSArray<UISearchToken *> *tokens = [[[self searchBar] searchTextField] tokens];
```
vijaysingh-axway: Same  as above. Will crash for iOS < 13. Apply check.
```suggestion
  if (![TiUtils isIOSVersionOrGreater:@"13.0"]) {
     return;
   }
  NSArray<UISearchToken *> *tokens = [[[self searchBar] searchTextField] tokens];
```

apidoc/Titanium/UI/SearchBar.yml
```diff
@@ -323,3 +351,24 @@ examples:
...
+---
+name: SearchBarToken
+summary: The search bar token for the <Titanium.UI.SearchBar.insertTokenAtIndex> method.
+since: 9.1.0
+osver: { ios: { min: "13.0" } }
+platforms: [iphone, ipad]
+properties:
+  - name: identifier
+    summary: The identifier of the search bar token.
+    type: Titanium.UI.ListSection
```
vijaysingh-axway: It should not be 'Titanium.UI.ListSection' .

iphone/Classes/TiUISearchBarProxy.m
```diff
@@ -184,9 +184,52 @@ - (id)barColor
...
+  int index = [TiUtils intValue:params[1]];
+
+  UISearchToken *searchToken = [UISearchToken tokenWithIcon:[TiUtils toImage:token[@"image"] proxy:self]
+                                                       text:[TiUtils stringValue:@"text" properties:token]];
+
+  if (token[@"identifier"] == nil) {
+    NSLog(@"[WARN] Missing search token identifier! Using a generated UUID …");
+  }
+
+  searchToken.representedObject = [TiUtils stringValue:@"identifier" properties:token def:[TiUtils createUUID]];
```
vijaysingh-axway: Property 'representedObject' is nullable. I think there is no need to assign any default value if it is not passed.

vijaysingh-axway (changes requested):
@hansemannn Please address review comments. Thanks!

iphone/Classes/TiUISearchBarProxy.m
hansemannn: Is this really required? The docs already state that it's iOS 13+. If someone calls an unavailable method in other frameworks / languages, it crashes as well. I am trying to keep the code as clean as possible.

iphone/Classes/TiUISearchBarProxy.m
hansemannn: It's required to be able to identify them when updating / deleting them.

iphone/Classes/TiUISearchBarProxy.m
hansemannn: Same as above - shouldn't be necessary

iphone/Classes/TiUISearchBarProxy.m
vijaysingh-axway: I think it is required. It may be we can log a message to know developer that this api is not available rather crashing app. If I see other implementations, it has check for APIs with iOS versions. 
You have also added check in function  "- (void)insertTokenAtIndex:(id)params" :)

vijaysingh-axway (approved):
CR passed.

ssaddique:
**FR Passed**

**Test Environment**

SDK Ver: 9.1.0 (this PR)
OS Ver: 10.14.6
Xcode Ver: Xcode 11.3.1
Appc CLI: 7.1.2
Daemon Ver: 1.1.3
Ti CLI Ver: 5.2.2
Alloy Ver: 1.14.4
Node Ver: 10.17.0
Emulator: All iOS 13 simulators.

Tested with < iOS 13 simulators, no crashes or anything unexpected.

PR #11493, opened 2020-02-21
fix(ios): convert function to use arrow function

ssjsamir:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27768


ewanharris (approved):
CR: Looks good!

PR #11498, opened 2020-02-25
fix(android): obtain holder for module references

garymathews:
- Obtain module instance when holder cannot be found.

##### TEST CASE
```JS
const permissions = Ti.Filesystem.requestStoragePermissions;
 
// Attempt to call 'requestStoragePermissions'.
permissions(e => {
    if (e.success) {
        console.log('Storage permissions granted.');
        return;
    }
    console.warn('Storage permissions denied.');
});
```
- Should see a console log regarding storage permissions.
- Should not see
```
Couldn't obtain argument holder
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27777)


android/kroll-apt/src/main/resources/org/appcelerator/kroll/annotations/generator/ProxyBinding.fm
```diff
@@ -755,8 +755,12 @@ beprovided to handle the result.
 		holder = holder->FindInstanceInPrototypeChain(getProxyTemplate(isolate));
 	}
 	if (holder.IsEmpty() || holder->IsNull()) {
-		LOGE(TAG, "Couldn't obtain argument holder");
-		args.GetReturnValue().Set(v8::Undefined(isolate));
-		return;
+		if (!moduleInstance.IsEmpty()) {
+			holder = moduleInstance.Get(isolate);
```
sgtcoolguy, line 759: seems like you'd need to do another `if` check like above here after trying the module instance? I think it *shouldn't* ever be empty/null, but who knows. We don't want to simply hard crash if it ever is...

sgtcoolguy (changes requested):

jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
Don't see the error "Couldn't obtain argument holder"
The permission is requested on the device and log message are seen based on allow/reject of permission.
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200413125029
Appc CLI: 8.0.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202003181504

PR #11457, opened 2020-01-28
feat: cross-platform light/dark mode API

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27501

**Description:**
This incorporates the code from @drauggres in his PR #11301

This attempts to make a more comprehensive cross-platform API for dealing with determining the current theme/mode, and consistency in the return value of `Ti.UI.fetchSemanticColor`.

**API**
- `Ti.UI.userInterfaceStyle` reports the current theme via `Ti.UI` constants: `USER_INTERFACE_STYLE_DARK`, `USER_INTERFACE_STYLE_LIGHT`, and `USER_INTERFACE_STYLE_UNSPECIFIED`
- `Ti.UI.fetchSemanticColor(String name)` returns a `Ti.UI.Color` proxy for the named color (and you can get the underlying "color" by calling `toHex()` on it to get a hex string) on iOS 13+, a string (representing the hex value) on Android and iOS < 13. The mismatched return types are necessary to make it easy to just pass the returned value to the UI properties without conversion right now.
  - Note that we support usage of the "standard" named colors baked into iOS that are adaptive to theme changes: [Standard Colors](https://developer.apple.com/documentation/uikit/uicolor/standard_colors?language=objc) and [UI Element Colors](https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors?language=objc)
- `Ti.UI` now has a new `userInterfaceStyle` event which gets fired when the device theme changes. The object received in the event listener has a `value` property whose value is the constant for the new `USER_INTERFACE_STYLE_*` that was set.

**Deprecations**
- `Ti.App.iOS.userInterfaceStyle` -> use `Ti.UI.userInterfaceStyle`
- `Ti.App.iOS.USER_INTERFACE_STYLE_DARK` -> use `Ti.UI.USER_INTERFACE_STYLE_DARK`
- `Ti.App.iOS.USER_INTERFACE_STYLE_LIGHT` -> use `Ti.UI.USER_INTERFACE_STYLE_LIGHT`
- `Ti.App.iOS.USER_INTERFACE_STYLE_UNSPECIFIED` -> use `Ti.UI.USER_INTERFACE_STYLE_UNSPECIFIED`
- Undocumented `Ti.UI.iOS.fetchSemanticColor()` (use `Ti.UI.fetchSemanticColor()`)

**Doc updates**
[...]

PR #11499, opened 2020-02-25
docs: fix Attribute.value

drauggres:
Attribute.value is optional:
```javascript
var text =  'Bacon ipsum dolor Appcelerator Titanium rocks! sit amet fatback leberkas salami sausage tongue strip steak.';

    text: text
});

// Underlines text
attr.addAttribute({
    range: [0, text.length]
});
```

Atttibute.value can be of `string` type:
```javascript
// Sets a background color
{
    value: "red",
    range: [text.indexOf('Appcelerator'), ('Appcelerator').length]
},
```

Examples from `apidoc/Titanium/UI/AttributedString.yml`

### DIFF
--- a/apidoc/Titanium/UI/Attribute.yml
+++ b/apidoc/Titanium/UI/Attribute.yml
@@ -112,7 +112,7 @@ properties:
           * <Titanium.UI.ATTRIBUTE_LINE_BREAK_BY_TRUNCATING_MIDDLE>
 
         These can also be combined the same way as the underline styles.
-    type: Number
+    type: [String, Number]
     constants: [ Titanium.UI.ATTRIBUTE_UNDERLINE_STYLE_*,
                  Titanium.UI.ATTRIBUTE_WRITING_DIRECTION_*,
                  Titanium.UI.ATTRIBUTE_LETTERPRESS_STYLE ]


jquick-axway:
The "value" property is not optional on iOS...
https://github.com/appcelerator/titanium_mobile/blob/73c73b6d352ada5918ec30e008c95bf12d7ed633/iphone/Classes/TiUIAttributedStringProxy.m#L52-L56

So, we need to keep the docs as-is. At least until we make it optional on iOS like Android.

drauggres:
> The "value" property is not optional on iOS...
> https://github.com/appcelerator/titanium_mobile/blob/73c73b6d352ada5918ec30e008c95bf12d7ed633/iphone/Classes/TiUIAttributedStringProxy.m#L52-L56
> 
> So, we need to keep the docs as-is. At least until we make it optional on iOS like Android.

This is not how it works.
When a property is not optional it means that you can't do:
```
attr.addAttribute({
    range: [0, text.length]
});
```
because "required" property "value" will be missing.
If is it not required in one platform it should be documented as optional.

jquick-axway:
> If is it not required in one platform it should be documented as optional.

That would just make it confusing from a cross-platform point of view. Especially for Titanium app devs who work on the iOS side first.

drauggres:
OK. Then we fix only type.

sgtcoolguy (commented):
Some notes after looking at the iOS codebase:
- This relates to #11457, because...
- On iOS the value may be a `Color` object for some attributes (useful with semantic colors) - note that the linked PR adds that doc type
- For attribute `type` of `Titanium.UI.ATTRIBUTE_FONT`, the value may be a String (font name, size is then hard-coded to 17), or the simple object defined as type `Font` in our docs.
- I don't see it listed as a valid `type` constant, but `Ti.UI.ATTRIBUTE_PARAGRAPH_STYLE` is - and it's `value` has to be a special object type already defined as `ParagraphAttribute`

I struggle with the `value` property being optional or not. It's clearly optional on Android for some attribute `type`s, but we enforce it as required on iOS. It very much feels like it may or may not be optional based on the `type` of attribute and may make sense to have useful defaults when it's not provided.

I'm not sure if there's a nice way to document all these variants, as the `type` of the attribute kind of toggles whether the `value` is required/optional and what possible `value` types we can use for them.

sgtcoolguy:
Closing as this was superseded by #11685 (and another commit I just made to master to add `ParagraphAttribute` as another supported type)

PR #11505, opened 2020-02-28
fix(android): reuse old TableViewRow when applicable

garymathews:
- Reuse old `TableViewRow` when applicable

##### TEST CASE
- Project on JIRA

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27769)


jquick-axway:
Your code change is more likely to create a new view container instead of re-using the existing one (like the old code was doing). This means the `TextField`'s parent view will be swapped-out, causing the virtual keyboard to disconnect and re-connect to the field. This has been proven to cause a huge performance issues with virtual keyboards that resize to show a suggestion bar... which in turn resizes the window height-wise every time the virtual keyboard changes. The `getView()` method will be called for every row every time a `TableView`/`ListView` gets resized.

I tried your PR with an Amazon Fire Tablet. The issue this PR is trying to fix is still there. Tapping on a `TextField` causes it to gain and immediately lose the focus, because the virtual keyboard resizes when tapping on a `TextField`. Something else is causing the problem.

jquick-axway:
It looks like this has always been an issue on an Amazon Fire Tablet. So, it's no worse than before.

So, we'll go with your change. It won't break any thing. It might bring back that performance issue, but I can't reproduce it.

We need to transition TableView and ListView to use Google's native RecyclerView anyways. So, we can relook at these things when we do that. Thanks!

jquick-axway (approved):
CR: Pass

ssaddique:
**FR Passed.** See JIRA ticket for details.

PR #11508, opened 2020-02-29
fix(android): module "clean" shouldn't error if missing "libs" folder

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27784

**Summary:**
When doing module development, doing an `appc ti clean` in a module directory would error out if directory is missing a "libs" subdirectory. This can happen if you've never built the module. It will also alway happen for Titanium 9.0.0 built modules since they never produce a "libs" folder anymore. This is a minor issue that can be safely ignored since all other folders/files are correctly deleted before the task fails (it just looks bad).

**Test:**
1. Follow the steps-to-reproduce in [TIMOB-27784](https://jira.appcelerator.org/browse/TIMOB-27784).
2. Verify that a clean does not throw an error.


sgtcoolguy (approved):

ssekhri:
FR Passed. Module clean works fine without error.
Verified on:
Mac OS: 10.15.1
SDK: 9.1.0.v20200228205731
Appc CLI: 8.0.0
JDK: 11.0.4
Node: 10.17.0

PR #11512, opened 2020-03-02
fix(buffer): allow setting falsy values through proxy

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27785

`setAdjustedIndex` was returning the value it did set on the buffer. this value was used as the return value for the proxy's  set handler. returning falsy values from there means that the handler failed setting the value which results in error messages like `Proxy object's 'set' trap returned falsy value for property '0'`

This fixes the handler by always returning true. I also removed the return statement from `setAdjustedIndex` as it is not used anywhere else.


common/Resources/ti.internal/extensions/node/buffer.js
```diff
@@ -1615,7 +1615,8 @@ const arrayIndexHandler = {
 		if (typeof propKey === 'string') {
 			const num = Number(propKey);
 			if (Number.isSafeInteger(num)) {
-				return setAdjustedIndex(target, num, value);
+				setAdjustedIndex(target, num, value);
```
sgtcoolguy: I think we should still return the value of setAdjustedIndex here, but that implementation is wrong...

common/Resources/ti.internal/extensions/node/buffer.js
```diff
@@ -1644,7 +1645,6 @@ function setAdjustedIndex(buf, index, value) {
 	if (index >= 0 || index < buf._tiBuffer.length) {
 		buf._tiBuffer[index + buf.byteOffset] = value;
```
sgtcoolguy: We should return true in this case to indicate success setting the value, and return false below (outside the if) to indicate failure (out of bounds)

sgtcoolguy (changes requested):

common/Resources/ti.internal/extensions/node/buffer.js
janvennemann: I compared against Node's behavior and setting a value on the buffer that is out of bounds is a noop.

```js 
const b = Buffer.alloc(1);
b[0] = 1;
b[1] = 2;
console.log(b);  // prints <Buffer 01>
```

That's why I removed the return value from `setAdjustedIndex` and always return `true`.

sgtcoolguy (approved):

ssaddique:
**FR Passed**

**Test Environment**
Studio Ver:     5.1.4.201909061933
SDK Ver:        9.0.0.v20200304064118
OS Ver:         10.14.6
Xcode Ver:      Xcode 11.3.1
Appc CLI:       7.1.2
Daemon Ver:     1.1.3
Alloy Ver:      1.14.4
Node Ver:       10.17.0
Emulators:       All iOS 13 iPhone simulators.

My testing has found that this bug affects only iOS 13 devices. < iOS 13 there is no error and result displays as expected. 
PR ready to be merged, waiting for Jenkins.

PR #11461, opened 2020-02-04
feat(android): add "google-services.json" support

hansemannn:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27745

**Test:**
1. Download the [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) app.
2. Build and run on Android.
3. Verify that the app builds and runs without issue.
4. Download [google-services.json](https://axwaysoftware-my.sharepoint.com/:u:/g/personal/jquick_axway_com/EU-gKkYY4wRFpVN4mrOcr3UBI8Ci9-trkF6V8ZdzmAfiVA?e=gMH5Ut). _(Only Axway employees can access this.)_
5. Copy the `google-services.json` file to: `./app/platform/android`
6. Build and run on Android.
7. In the log, verify that you see the following message.
```
[INFO]  [GRADLE] Parsing json file: /Users/<YourFolder>/kitchensink-v2/build/android/app/google-services.json
```

PR #11514, opened 2020-03-03
feat(android)(9_0_X): add "google-services.json" support

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27745

**Cherry-pick of PR:** #11461


m1ga:
@hansemannn with this PR firebase-auth works.

hansemannn:
Jep, thx! I really hope this makes it into 9.0.1. Can't update to 9.x before this.

jquick-axway:
@hansemannn, @m1ga, are you unable to update your firebase modules until this change makes it in?

m1ga:
@jquick-axway in my tests I can't use the latest versions of the modules because of `string not found` errors. I was able to use firebase-core + firebase-cloud-messaging with a lower version e.g. fcm 20.0.1 so it will work. 
But firebase-auth won't work without the patch. Here is a test version: http://migaweb.de/firebase.auth-android-3.0.0.zip if you just include it inside the tiapp.xml it will crash with a `string not found` error. As soon as I use this PR and put the google-service.json in the right place it works fine.

I didn't try to update the other modules to the latest module versions yet since I have a setup now that works and I don't want to change it :)

jquick-axway:
That's what I was afraid of. For your info, the google-services gradle tool turns the JSON into string xml entry under the "res" folder. So, it's failing to find these string resources.

hansemannn:
Yep, that was exactly our issue. We used a local fork that includes the patch now, so once this is merged, we can go official again!

ssekhri:
FR Passed
Mac OS: 10.15.4
SDK: 9.0.1.v20200413124532
Appc CLI: 8.0.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202003181504

PR #11516, opened 2020-03-06
feat(android): replace clang-format with gradle checkstyle

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27473

**Summary:**
Replaced "clang-format" tool used to enforce Java coding standards in the Titanium SDK with gradle "checkstyle" plugin.

---
**Pros:**
- Enforces coding standards (like clang) but also checks for common Java coding mistakes.
- Now also checks for:
  * Unused imports.
  * Wildcard imports. ex: `import java.io.*;`
  * String reference equality. _(Should use value equality instead.)_
  * Classes that override `equals()` also must override `hashCode()`.
  * etc...
- Tool runs before Java compile. Will trigger build failure if rules are violated.
  * This is new. Encourages devs to be more pro-active regarding style.
  * Old way was to run clang-format manually or have Jenkins do it.
- Violations are logged as human readable errors.
  * Android Studio logs clickable filename/line-number links of violations.

**Cons:**
- Adds about 5 seconds to build time.
- Cannot re-format code with this tool. Developer must do it manually.
- Checkstyle can't support all rules that clang-format does.
  * Cannot enforce Xcode style horizontal alignment of method arguments.
  * Will log violation if you attempt to horizontally align values in an array initialization block. Must init in 1 line or indent 1 tab over on next line.
  * Cannot enforce space within curly braces `{}`. _(Might be able to do this via regex, but it would have to ignore curly braces within strings.)_


sgtcoolguy:
So I'm all for using something like checkstyle since it can catch actual coding issues, unused code, etc. The issue this raises is enforcement/auto-formatting of Java code. The clang-format hook via husky would forcibly auto-format the Java code on `git commit`. Moving this way would basically enforce the format at build-time of the SDK - but we wouldn't have any CLI/hook to re-format code to make it happy, it'd be on developers to do that (and we wouldn't have a quick pre-build way to check the formatting). It also meant that the feedback (at least on Jenkins builds) would be later in the process.

So a couple things:
- Is there any way to use a check stye config to actually re-format the code, or can it only report violations?
- In either case, can we retain a husky/lint-staged hook to incrementally check only modified files on `git commit` to catch the issue earlier in the process?
- Jenkins has infrastructure to be able to process warnings/errors from Checkstyle (and android lint, findbugs, tons of tools). We should alter the Jenkinsfile to grab the resulting checkstyle output and attach that to the build. (I don't know if this would run on our clean command or our build command, but it's likely as simple as editing Jenkinsfile#302 to be: `recordIssues(tools: [clang(), java(), checkstyle()])`)

sgtcoolguy:
I see there's a command line way to invoke checkstyle, so maybe we could hook this up to the lint-staged git hook: https://checkstyle.org/cmdline.html

That way it'd just run checkstyle on the modified files at git commit time.

jquick-axway:
> Is there any way to use a check stye config to actually re-format the code, or can it only report violations?

Unfortunately, no. It only logs violations. The developer has to fix them manually.

The good news is that the logged violations are human readable. And also Android Studio turns the logged file paths and line numbers into clickable links which takes you directly where the issue occurs within the IDE (I love this). Here is some example output after I purposely added some violations to the code...
```
> Task :titanium:checkJavaStyle
[ant:checkstyle] [ERROR] /Users/jquick2/Development/titanium_mobile.jquick/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java:66: 'member def modifier' has incorrect indentation level 0, expected level should be 4. [Indentation]
[ant:checkstyle] [ERROR] /Users/jquick2/Development/titanium_mobile.jquick/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java:67: Must use tabs for indentation. [RegexpSinglelineJava]
[ant:checkstyle] [ERROR] /Users/jquick2/Development/titanium_mobile.jquick/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java:68: Line cannot have whitespace at the end. [RegexpSinglelineJava]

> Task :titanium:checkJavaStyle FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':titanium:checkJavaStyle'.
> Checkstyle rule violations were found. See the report at: file:///Users/jquick2/Development/titanium_mobile.jquick/android/titanium/build/reports/checkstyle/checkJavaStyle.html
  Checkstyle files with violations: 1
  Checkstyle violations by severity: [error:3]
```

> In either case, can we retain a husky/lint-staged hook to incrementally check only modified files on git commit to catch the issue earlier in the process?

We could, but checkstyle is expected to be given the entire source tree in order for it to check for "design" issues. I've already created custom `checkJavaStyle` gradle tasks in our "titanium", "kroll-apt", and "app" gradle sub-projects that already does all of this. We can run checkstyles for all projects (in parallel) via the below command line, but we would have to do a `node scons clean` first to finish setting up the gradle environment.
```
./gradlew checkJavaStyle
```

> Jenkins has infrastructure to be able to process warnings/errors from Checkstyle (and android lint, findbugs, tons of tools). We should alter the Jenkinsfile to grab the resulting checkstyle output and attach that to the build.

Checkstyle currently outputs an HTML file reporting which Java files it scanned and the errors it found. Is this something you want?
https://docs.gradle.org/current/userguide/checkstyle_plugin.html#sec:checkstyle_customize_xsl

Currently, violations are also logged out to stderr just like compiler errors. This is how I'm triggering the Java compiler errors (checkstyle is a prebuild step) and allows us to be more pro-active about this while we're writing code. (Versus doing it via clang-format is more reactive.)

Jenkinsfile
```diff
@@ -4,6 +4,12 @@ library 'pipeline-library'
 // Keep logs/reports/etc of last 30 builds, only keep build artifacts of last 3 builds
 properties([buildDiscarder(logRotator(numToKeepStr: '30', artifactNumToKeepStr: '3'))])
 
+// Copy Android SDK/NDK directory paths from Jenkins' environment variables to Google's environment variables.
+// Allows us to run gradle without having to create a "local.properties" file with these paths.
+env.ANDROID_SDK_ROOT = env.ANDROID_SDK
```
sgtcoolguy: This likely does not work as you might expect. You're running this outside of the context of any node, so I think it'll basically get/set values from the master Jenkins build machine (while the actual build/test will run on different nodes with likely different paths for the SDK/NDK!)
I'd suggest retaining passing the values in as command line arguments at the build/clean command time.

Jenkinsfile
```diff
@@ -293,19 +299,19 @@ timestamps {
...
 						} // timeout
 						timeout(15) {
-							def buildCommand = "npm run build -- --android-ndk ${env.ANDROID_NDK_R16B} --android-sdk ${env.ANDROID_SDK}"
+							def buildCommand = "npm run build --"
 							if (isMainlineBranch) {
 								buildCommand += ' --all'
 							}
 							sh label: 'build', script: buildCommand
-							recordIssues(tools: [clang(), java()])
+							recordIssues(tools: [checkstyle(), clang(), java()])
```
sgtcoolguy: So this assumes that build/clean will write xml files with the results somewhere. (The default is to assume finding them via the pattern `**/checkstyle-result.xml`) It's unclear if the gradle build does that already (you mentioned an html report)

sgtcoolguy: So I checked out this PR and ran it, it appears we'll need to tweak this to be:
```groovy
recordIssues(tools: [checkstyle(pattern: 'android/**/build/reports/checkstyle/checkJavaStyle.xml'), clang(), java()])
```

The issue is that checkstyle is also being run during the `npm test` call earlier as well (and I assume that's why you're also trying to just set the NDK/SDK values in environment variables). So I think it may make sense to recordIssues after the npm test call above and just try and record java/clang issues after build?

sgtcoolguy (approved):
LGTM and now builds fine and is integrated with Jenkins. @jquick-axway you'll likely want to rebase/fix our series of commits to make Jenkins happy down to one.

jquick-axway:
@sgtcoolguy, the Jenkins commits are now squashed.

There's nothing here for QE to test.
@garymathews, you can give this PR a go if you want.

PR #11517, opened 2020-03-06
docs(api): make animation parameter optional

janvennemann:
Makes the `animation` parameter in `replaceSectionAt` optional.


sgtcoolguy (approved):
yup, this is right. (Sucks that we can't tag the parameter as iOS-only in the docs)

PR #11537, opened 2020-03-11
fix(iOS) : check applied to avoid crash if video player url is null

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27799


tests/Resources/ti.media.videoplayer.addontest.js
```diff
@@ -0,0 +1,72 @@
...
+			videoPlayer.play();
+
+			setTimeout(function () {
+				videoPlayer.stop();
+				videoPlayer.url = null;
+			}, 2000);
+		});
+
+		videoPlayer.addEventListener('playing', function () {
+			finish();
```
sgtcoolguy, line 67: I'm not sure this test actually will work to reproduce the crash/fix. When the window opens it immediately tries to play the original video. Won't that then fire the `playing` event and the test will finish before the 2s timer fires to stop and set the url to null?

I believe the test should be something like:
```js
it.ios('App should not crash when setting video player url to null (TIMOB-27799)', function (finish) {
 		this.timeout(10000);

  		win = Ti.UI.createWindow();
 		const videoPlayer = Ti.Media.createVideoPlayer({
 			height: '72%',
 			url: '/movie.mp4'
 		});
 		win.add(videoPlayer);
 			videoPlayer.play();

  			setTimeout(function () {
 				videoPlayer.stop();
 				videoPlayer.addEventListener('playing', e => {
 				  try {
 				    should(e).have.a.property('url').which.is.null();
 				    finish();
 				  } catch (err) {
 				    finish(err);
 				  } 				  
 				});
 				videoPlayer.url = null;
 			}, 50);
 		});
  		win.open();
 	});
 });
```

sgtcoolguy (changes requested):

tests/Resources/ti.media.videoplayer.addontest.js
vijaysingh-axway: @sgtcoolguy  '[playing'](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Media.VideoPlayer-event-playing) event fired when currently playing movie changes. So it will not fired first. After changing url 'playing' event  will fire. 
I retested and it is crashing without the changes and working fine with changes.

sgtcoolguy (approved):

PR #11568, opened 2020-03-25
fix(android): assume application/javascript is not binary 

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27823

**Optional Description:**
Android 29+ reports js content mime type as `application/javascript`. Used to report `text/javascript`. As a result our blob code that sniffed if it's contents were binary were assuming js files were binary.


garymathews (approved):
CR: **PASS**

PR #11571, opened 2020-03-25
fix(android): prevent high cpu usage in Android Studio

garymathews:
- Disable specific _Inspections_ causing high CPU utilization

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27825)


jquick-axway:
@garymathews, Android Studio still slams my CPU for about 15-20 minutes with this change.

I sent you an updated inspections XML which makes it super fast for me. Turns out the most expensive inspections were under the "Probably bugs" section. I disabled most of those.

sgtcoolguy:
Given this is an internal dev thing, can we merge it in? Does this work well for both @jquick-axway  and @garymathews ?

jquick-axway:
@garymathews, we should add the following to our "titanium_mobile/android/.gitignore" file so that we can easily commit changes to this XML file in the future.
```
!/.idea/inspectionProfiles/
```

jquick-axway (approved):

PR #11575, opened 2020-03-26
fix(android): allow TiDownloadManager to fire fail listener

garymathews:
- Allow `TiDownloadManager` to fire fail listener, which prevented `error` events for some components

##### TEST CASE
```JS
	image: 'https://invalid.host.com/image.jpg'
});

img.addEventListener('error', e => {
	console.error(JSON.stringify(e, null, 4));
});

win.add(img);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27797)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
@@ -6,7 +6,6 @@
  */
 package org.appcelerator.titanium.util;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.lang.ref.SoftReference;
 import java.lang.reflect.Constructor;
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
@@ -6,6 +6,9 @@
  */
 package org.appcelerator.titanium.util;
 
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Message;
 import java.io.InputStream;
 import java.lang.ref.SoftReference;
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
@garymathews Can you please turn that manual test into a mocha test? It looks like it should be pretty straightforward to convert...

android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
```diff
@@ -88,7 +88,7 @@ public void download(URI uri, TiDownloadListener listener)
 	 * <p>
 	 * Returns null if failed to download content or if given an invalid argument.
 	 */
-	public InputStream blockingDownload(final URI uri)
+	public InputStream blockingDownload(final URI uri) throws Exception
```
jquick-axway, line 91: We need to update the `@return` comment since it no longer returns `null` if failed to download.

android/titanium/src/java/org/appcelerator/titanium/util/TiDownloadManager.java
```diff
@@ -146,13 +146,7 @@ public void run()
```
jquick-axway, line 146: The above `if (TiApplication.isUIThread()) { ... }` code block should be changed to throw an exception too for consistency. This code block can be tested when setting a Titanium view's `backgroundImage` property.

How about something like this?
```java
// If running on main UI thread, then block this thread while doing download on another thread.
// Note: Using "HttpUrlConnection" on UI thread will cause a "NetworkOnMainThreadException" to be thrown.
if (TiApplication.isUIThread()) {
	// Provides the error result of the threaded-out runnable below.
	class RunnableResult
	{
		Exception exception;
	}

	// Perform the blocking download on another thread.
	// Downloaded content will be made available via Titanium's "TiResponseCache".
	final RunnableResult runnableResult = new RunnableResult();
	Thread thread = new Thread(new Runnable() {
		@Override
		public void run()
		{
			try (InputStream stream = blockingDownload(uri)) {
				if (stream != null) {
					KrollStreamHelper.pump(stream, null);
				}
			} catch (Exception ex) {
				runnableResult.exception = ex;
			}
		}
	});
	thread.start();
	thread.join();
	if (runnableResult.exception != null) {
		throw runnableResult.exception;
	}

	// Return a stream to the downloaded file/content via our response cache.
	URI cachedUri = TiResponseCache.fetchEndpointFollowingRedirects(uri);
	if (cachedUri != null) {
		inputStream = TiResponseCache.openCachedStream(cachedUri);
	}
	return inputStream;
}
```

jquick-axway (changes requested):

garymathews:
@jquick-axway Updated PR

jquick-axway (approved):
CR: Pass

jquick-axway:
@garymathews, using `AtomicReference` to box the exception was a smart idea. I like it. And you're the first person to add a Java lambda to our codebase. 👍

ssjsamir (approved):
FR passed: Able to see the correct events being fired, tested with the test case above and the following test case: 
```
Ti.UI.setBackgroundColor('#000');

var tabGroup = Ti.UI.createTabGroup();

    title:'Tab 1',
    backgroundColor:'#fff',
});

	image:'http://files.daddyhunt.com/sites/default/files/imagecache/mangrid-212/photos/z/z/z/277606/Photo_on_2010-09-14_at_13.14_3.jpg'
});
win1.add(img);
img.addEventListener('error', function(){
    Ti.API.info('--------------fired error event----------');
});

tabGroup.addTab(Ti.UI.createTab({
    title:'Tab 1',
    window:win1
}));
tabGroup.open();
```

**Test Environment**

**MacOS Catalina: 10.15.5 Beta 
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
Pixel Xl 7.1.1 Sim**

PR #11621, opened 2020-04-11
fix(android): changing picker minDate/maxDate after open

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27625

**Summary:**
Changing picker's `minDate` or `maxDate` after window open would not let user choose value between that range. Would still use the old min/max range. Would also crash if the new range is completely outside of the old range.

**Test:**
2. Verify that you can only select a date between May 1st and July 31st.
3. Tap the "Change Min/Max" button.
4. Verify that you can only select a date between January 1st and March 31st.
_(This used to crash before this PR.)_

```javascript
	type: Ti.UI.PICKER_TYPE_DATE,
	top: "50dp",
	minDate: new Date(2020, 4, 1),
	maxDate: new Date(2020, 6, 31),
});
window.add(picker);
var button = Ti.UI.createButton({ title: "Change Min/Max Date" });
	picker.minDate = new Date(2018, 0, 1);
	picker.maxDate = new Date(2018, 2, 31);
});
window.add(button);
window.open();
```


garymathews (approved):
LGTM!

ssjsamir (approved):
FR Passed: Application no longer crashes and able to see correct date ranges with the code posted above. 

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
Pixel Xl 7.1.1 Sim
```

PR #11644, opened 2020-04-17
fix(android): production builds no longer copy AAB to distribution folder as of 9.0.1

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27852

**Summary:**
- Regression introduced in 9.0.1.
- Caused by using newest Android gradle tool. Google renamed built AAB file and Titanium was failing to find it.
- Added APK and AAB file validation. Will now trigger build failure if files are not found.

**Test:**
1. Delete the last built APK and AAB files from the distribution folder.
2. Build for "production".
3. Go to the selected distribution folder.
4. Verify that APK and AAB files were generated.

PR #11645, opened 2020-04-17
fix(android)(9_0_X): production builds no longer copy AAB to distribution folder as of 9.0.1

jquick-axway:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27852

**Cherry-pick of PR:** #11644


garymathews (approved):
CR: **PASS**

android/cli/commands/_build.js
```diff
@@ -3689,7 +3689,17 @@ AndroidBuilder.prototype.buildAppProject = async function buildAppProject() {
...
 		// Set path to the app-bundle file that was built up above.
 		// Our "package.js" event hook will later copy it to the developer's chosen destination directory.
-		this.aabFile = path.join(this.buildDir, 'app', 'build', 'outputs', 'bundle', 'release', 'app.aab');
+		this.aabFile = path.join(this.buildDir, 'app', 'build', 'outputs', 'bundle', 'release', 'app-release.aab');
+	}
+
+	// Verify that we can find the above built file(s).
+	if (!await fs.exists(this.apkFile)) {
+		this.logger.error(`Failed to find built APK file: ${this.apkFile}`);
+		process.exit(1);
```
sgtcoolguy, line 3698: Just an FYI - I am not fond of our use of exiting the process immediately in these functions like this. Ideally we simple return an error to the callback/throw an Error (if in an async function or sync function with no callback) - and let the top-level CLI do the exiting of the process with a failure exit code. We tend to have these calls sprinkled throughout CLI related code making re-use harder if we ever want to re-use in another context.

jquick-axway: I'll change it to throw an exception instead, which I think is fine in this case.

That said, I don't think we should throw an exception for say end-user configuration issues, because the logged stack trace would make it look like an issue on our end when it's not.

lokeshchdhry:
FR Passed.
`.aab` file is successfully generated & copied to the specified location.

Studio Ver:       6.0.0.202003181504
SDK Ver:          9.0.2 local build
OS Ver:           10.15.4
Xcode Ver:        Xcode 11.4
Appc NPM:         5.0.0
Appc CLI:         8.0.0
Daemon Ver:       3.2.0
Ti CLI Ver:       5.2.2
Alloy Ver:        1.14.6
Node Ver:         12.13.1
NPM Ver:          6.12.1
Java Ver:         11.0.6

PR #11646, opened 2020-04-19
feat(android): animate elevation

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6531

Being able to animate the `elevation` value. 
Also removed some of the pre-Honeycomb stuff.

```javascript
	layout: "vertical",
	backgroundColor: "#fff"
});

var v = Ti.UI.createView({
	elevation: 20,
	width: 200,
	height: 200,
	top: 50,
	backgroundColor:"#fff"
});
var v2 = Ti.UI.createView({
	width: 200,
	height: 200,
	top: 50,
	backgroundColor:"#fff"
});

win.add([v,v2]);

setTimeout(function(){
	v.animate({
		duration: 2000,
		elevation: 0,
		autoreverse:true,
		repeat: 10
	});
	v2.animate({
		duration: 2000,
		elevation: 20,
		autoreverse:true,
		repeat: 10
	});
}, 1000)

win.open();
```


android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
```diff
@@ -238,6 +232,10 @@ public void applyOptions(HashMap options)
 			backgroundColor = TiConvert.toColor(options, TiC.PROPERTY_BACKGROUND_COLOR);
 		}
 
+		if (options.containsKey(TiC.PROPERTY_ELEVATION)) {
+			elevation = TiConvert.toFloat(options, TiC.PROPERTY_ELEVATION);
```
garymathews, line 236: Also specify default in case of invalid value
```Java
elevation = TiConvert.toFloat(options, TiC.PROPERTY_ELEVATION, -1);
```

android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
```diff
@@ -329,6 +295,10 @@ private AnimatorSet buildPropertyAnimators(int x, int y, int w, int h, int paren
 			addAnimator(animators, ObjectAnimator.ofFloat(view, "alpha", toOpacity.floatValue()));
 		}
 
+		if (elevation != -1) {
```
jquick-axway, line 298: I think we should avoid doing equality checks with floats. I'd prefer that we change it to the following instead.
```java
if (elevation >= 0) {
	addAnimator(animators, ObjectAnimator.ofFloat(view, "elevation", elevation));
}
```

Also, would you mind updating the `Animation.yml` file with this new property please?
We can add this feature to 9.1.0.

The rest looks great. Thanks!

jquick-axway (changes requested):

android/titanium/src/java/org/appcelerator/titanium/util/TiAnimationBuilder.java
m1ga: you're right, that looks better! Fixed + docu added

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
FR Passed: Able to see elevation in the animation from the test case provided above

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta 
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
Pixel Xl 7.1.1 Sim
```

PR #11648, opened 2020-04-20
fix(ios): use jsvalue instead of jsmanagedvalue to handle events

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27839


sgtcoolguy (approved):
Might want @janvennemann to also chime in, because I've made mistakes overusing JSManagedValues before?

My understanding was that if you're storing a JSValue in an object exported to the JS engine (ObjcProxy here) that you need to use JSManagedValues to avoid a retain cycle: https://developer.apple.com/documentation/javascriptcore/jsmanagedvalue?language=objc

So is the issue that it's basically setting the value to nil prematurely? If so, is that because the proxy that fires the events is not sticking around somehow? Basically: is this fixing the issue because it's forcing a retain cycle for the proxy and callback to live forever - vs.  - do we need to somehow make the proxy live longer by retaining it elsewhere.

I suspect that the actual root cause here is that the modules implemented in the newer Obj-C API are not being kept alive as singletons forever. But maybe causing a retain cycle between the module and it's event listeners is the right solution anyways (i.e. the module instance stays alive so long as it's reachable via JS *or* it has at least one event listener?)

janvennemann:
@sgtcoolguy I think you are on the right track. I took another look at our core and found this:

https://github.com/appcelerator/titanium_mobile/blob/cedb733fba82c98b4c1fe37790b30605b537cc75/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollObject.m#L226-L234

Notice how the newer style proxies get returned straight away, whereas the old style proxies go through `noteObject:`, which makes sure to store the proxy in a hidden property on its owning object. So it looks like the issue is indeed that the object of new style ObjC modules gets GC'ed.

That the module instance stays alive so long as it's reachable via JS or it has at least one event listener sounds right to me. The question is do we have more things that depend on the module being held alive, except for event listeners?

sgtcoolguy:
I'm kind of stuck here. I don't know that there's any other reason to keep them around outside of firing off events to listeners? If we did want to explicitly keep them around anyways, how would we?

I think this PR is fine so long as we make an explicit note/comment in the code here that we made the choice to force a retain cycle between the module/proxy and the event listeners to keep both alive so long as there were any listeners (so that in the future we remember that this is causing one and we made the explicit decision to do so for this reason).

vijaysingh-axway:
@sgtcoolguy  Added comment in code as per your suggestion.

ssjsamir (approved):
FR Passed: Tested using the test cases mentioned in: https://jira.appcelerator.org/browse/TIMOB-27839

**Test environment**
```
MacOS Catalina: 10.15.5 Beta
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
ipad pro 12.9 inch (ios 13.4)
ipad pro 9.7 inch (ios 12.2)
```

PR #11654, opened 2020-04-22
fix(ios): callback not invoked if missing options dictionary in Ti.Platform.openURL

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27861


iphone/Classes/PlatformModule.m
```diff
@@ -297,17 +297,17 @@ - (NSNumber *)totalMemory
...
 
-  // iOS 10+
   NSMutableDictionary *optionsDict = [NSMutableDictionary dictionary];
-  if ([options isKindOfClass:[NSDictionary class]]) {
-    optionsDict = (NSMutableDictionary *)options;
-  } else if ([options isKindOfClass:[JSValue class]]) {
-    callback = (JSValue *)options;
+  NSDictionary *dictionary = [options toDictionary];
+  if (dictionary.allKeys.count > 0) {
+    optionsDict = (NSMutableDictionary *)dictionary;
```
janvennemann, line 308: I know it was like this before, but since you are already touching this piece of code: Casting to `NSMutableDictionary` is unnecessary here. Should someone ever try to write to the `optionsDict` here it will crash, since the underlying type is still `NSDictionary` (the proper way would have been using `mutableCopy`). `optionsDict`  never gets mutated anyway, so it should be a `NSDictionary`.

iphone/Classes/PlatformModule.m
```diff
@@ -297,17 +297,17 @@ - (NSNumber *)totalMemory
...
   BOOL result = NO;
 
-  // iOS 10+
   NSMutableDictionary *optionsDict = [NSMutableDictionary dictionary];
-  if ([options isKindOfClass:[NSDictionary class]]) {
-    optionsDict = (NSMutableDictionary *)options;
-  } else if ([options isKindOfClass:[JSValue class]]) {
-    callback = (JSValue *)options;
+  NSDictionary *dictionary = [options toDictionary];
+  if (dictionary.allKeys.count > 0) {
```
janvennemann, line 307: Although highly unlikely, but something like the following would break this check:

```js
const callback = () => {};
callback.foo = 'bar';
Ti.Platform.openURL('https://www.google.com', {}, callback);
```

Unfortunately, JSValue has no built-in functionality to test whether a value is a function or not. However, the C-API has [JSObjectIsFunction](https://developer.apple.com/documentation/javascriptcore/1451769-jsobjectisfunction?language=objc), which we could use for an extension:

```objc
@interface JSValue (TitaniumAdditions)

- (BOOL)isFunction;

@end

@implementation JSValue (TitaniumAdditions)

- (BOOL)isFunction {
  if (!self.isObject) {
    return false;
  }
  
  JSContextRef context = self.context.JSGlobalContextRef;
  JSObjectRef object = JSValueToObject(context, self.JSValueRef, NULL);
  if (object == NULL) {
    return NO;
  }
  return JSObjectIsFunction(context, object);
}

@end
```

janvennemann (changes requested):

iphone/Classes/PlatformModule.m
vijaysingh-axway: Good idea. Updated PR according to suggestion.

janvennemann (approved):

ssjsamir (approved):
FR Passed: callback is now invoked if it is missing the options dictionary in Ti.Platform.openURL.
Tested using the test cases from https://jira.appcelerator.org/browse/TIMOB-27861 and https://jira.appcelerator.org/browse/TIMOB-23888

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
iphone 8 sim (ios 13.4)
```

PR #11655, opened 2020-04-22
fix(ios): callback not invoked if missing options dictionary in Ti.Platform.openURL

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27861


iphone/Classes/PlatformModule.m
```diff
@@ -295,17 +295,17 @@ - (NSNumber *)totalMemory
...
   NSURL *newUrl = [TiUtils toURL:url proxy:self];
   BOOL result = NO;
 
-  // iOS 10+
   NSMutableDictionary *optionsDict = [NSMutableDictionary dictionary];
-  if ([options isKindOfClass:[NSDictionary class]]) {
-    optionsDict = (NSMutableDictionary *)options;
-  } else if ([options isKindOfClass:[JSValue class]]) {
-    callback = (JSValue *)options;
+  NSDictionary *dictionary = [options toDictionary];
```
sgtcoolguy, line 304: this call is ok if the second arg is a callback function?

sgtcoolguy (changes requested):
Can we write a quick unit test to verify the fix - basically ensure the callback gets fired when passing in no options?

```js
Ti.Platform.openURL('http://www.google.com', e => finish());
```

Or will it exit the app to Safari or something? Maybe there's a URL that can be handled by the test app itself or in a way that lets the test app keep going?

jquick-axway:
@sgtcoolguy, I think a good way to unit test this is to set up the mocha app to use a custom URL scheme and have `openURL()` open itself. This way you won't leave the app... and we can also verify that `Ti.App.iOS` event "handleurl" works as well (along with Android's intent handling). But we would have to make a PR in the mocha suite to set this up. I'm willing to do this if you'd like.

iphone/Classes/PlatformModule.m
vijaysingh-axway: Yes. This will return blank dictionary.
But there is another issue in PR. I'll update PR.

vijaysingh-axway: There is an 

> Can we write a quick unit test to verify the fix - basically ensure the callback gets fired when passing in no options?
> 
> ```js
> Ti.Platform.openURL('http://www.google.com', e => finish());
> ```
> 
> Or will it exit the app to Safari or something? Maybe there's a URL that can be handled by the test app itself or in a way that lets the test app keep going?

Yeah. For this particular issue a simple test case can be written. Open a custom url scheme which does not exist. e.g  
 Ti.Platform.openURL("randomapp://", e => finish());
I'll add this.

sgtcoolguy:
@jquick-axway Yeah that sounds like a good idea - I'd love it if you could add that.

jquick-axway:
@sgtcoolguy, here is the mocha suite PR which tests via app's custom URL scheme.
https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/250

sgtcoolguy (approved):

ssjsamir (approved):
FR Passed: callback is now invoked if it is missing the options dictionary in Ti.Platform.openURL.
Tested using the test cases from https://jira.appcelerator.org/browse/TIMOB-27861 and https://jira.appcelerator.org/browse/TIMOB-23888

**Test Environment**
```
MacOS Catalina: 10.15.5 Beta
Xcode: 11.4
Java Version: 1.8.0_131
Node.js: 10.16.3
iphone 8 sim (ios 13.4)
```

PR #11658, opened 2020-04-25
feat(android): animate color

m1ga:
**merge after https://github.com/appcelerator/titanium_mobile/pull/11646**

Since the other PR #11646  is already approved I've used it as the master to this PR since it changing the same file.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27855

Being able to animate the `color` value.

```
Ti.UI.setBackgroundColor('white');
var win = Ti.UI.createWindow({
	backgroundColor: 'white'
});
var l = Ti.UI.createLabel({
	text: "test",
	color: "#000"
});
win.add(l);

setTimeout(function() {
	l.animate({
		color: "#ff0",
		duration: 1000,
		repeat: 10,
		autoreverse: true
	});
}, 1000);
win.open();
```

PR #11666, opened 2020-04-28
feat: update view properties after animation

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-13764

**Description:**
After running `Ti.UI.View#animate()`, the animated properties should reflect their new values. This PR attempts to do just that.

Just prior to firing the `complete` event and calling the optional `callback` function, we apply the properties held by the animation to the underlying `Ti.UI.View`.

**Note:** This is a behavioral change! I would argue it is not a breaking change (unless I broke the implementation here somehow) - it's actually a sort of bug fix of "bad" behavior we've had forever.

I've updated a handful of tests we had that did animations and checked the end values - updating them to verify we use the target value from the animation. I *did not* try transition animations (one view to another).

**Another Note:** This was triggered by reviewing #11658 - where a `Ti.UI.Label`'s `color` property was animated. Note that iOS has an issue in animating a label color where it just immediately changes color rather than animating over time. It's an OS limitation. Annoying.


apidoc/Titanium/UI/Animation.yml
```diff
@@ -21,10 +21,12 @@ description: |
 
     Use the <Titanium.UI.createAnimation> method to create an animation object.
 
-    Note that when you animate a view's size or position, the actual layout properties (such as
+    Note that on SDKs older than 9.1.0, when you animate a view's size or position the actual layout properties (such as
     `top`, `left`, `width`, `height`) are not changed by the animation. See the description of the
     [animate](Titanium.UI.View.animate) method for more information.
 
+    As of 9.1.0, the animated properties shoudl reflect their final values just before the `complete` event and/or the <Ti.UI.View.animate> callback is fired.
```
vijaysingh-axway, line 28:
```suggestion
    As of 9.1.0, the animated properties should reflect their final values just before the `complete` event and/or the <Ti.UI.View.animate> callback is fired.
```

vijaysingh-axway (approved):
iOS changes looks good.

jquick-axway:
@sgtcoolguy, this PR crashes on iOS at the end of the animation if you set "autoreverse" to `true`.

```javascript
	backgroundColor: "white",
	fullscreen: true,
});
var squareView = Ti.UI.createView({
	backgroundColor: "blue",
	borderColor: "black",
	borderWidth: "2dp",
	width: "80dp",
	height: "80dp",
});
window.add(squareView);
window.addEventListener("postlayout", function onLayout() {
	window.removeEventListener("postlayout", onLayout);
	var windowSize = window.size;
	var minDimension = Math.min(windowSize.width, windowSize.height);
	var letterboxScale = minDimension / squareView.size.width;
	var animation = Ti.UI.createAnimation({
		transform: Ti.UI.createMatrix2D().scale(letterboxScale, letterboxScale),
		curve: Ti.UI.ANIMATION_CURVE_EASE_IN,
		duration: 2000,
		autoreverse: true,
//		repeat: 2,
	});
	squareView.animate(animation);
});
window.open();
```

Here is the stack trace...
```
[WARN]  New layout set while view [object TiUIView] animating: Will relayout after animation.
[WARN]  *** Terminating app due to uncaught exception 'org.appcelerator.TiUIViewProxy', reason: 'Invalid type passed to function'
[WARN]  *** First throw call stack:
[WARN]  (
[WARN]          0   CoreFoundation                      0x00000001081ca8db __exceptionPreprocess + 331
[WARN]          1   libobjc.A.dylib                     0x0000000106fd7ac5 objc_exception_throw + 48
[WARN]          2   TitaniumKit                         0x0000000106ea6574 JavascriptNameForClass + 0
[WARN]          3   TitaniumKit                         0x0000000106efb442 -[TiProxy applyProperties:] + 258
[WARN]          4   TitaniumKit                         0x0000000106ee7fe3 -[TiAnimation animationCompleted:finished:context:] + 327
[WARN]          5   UIKitCore                           0x00000001145b53e3 -[UIViewAnimationBlockDelegate _didEndBlockAnimation:finished:context:] + 847
[WARN]          6   UIKitCore                           0x0000000114588e47 -[UIViewAnimationState sendDelegateAnimationDidStop:finished:] + 343
[WARN]          7   UIKitCore                           0x0000000114589469 -[UIViewAnimationState animationDidStop:finished:] + 293
[WARN]          8   QuartzCore                          0x000000010595bc5b _ZN2CA5Layer23run_animation_callbacksEPv + 321
[WARN]          9   libdispatch.dylib                   0x000000010a600d02 _dispatch_client_callout + 8
[WARN]          10  libdispatch.dylib                   0x000000010a60ca50 _dispatch_main_queue_callback_4CF + 1276
[WARN]          11  CoreFoundation                      0x0000000108131a79 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
[WARN]          12  CoreFoundation                      0x000000010812c126 __CFRunLoopRun + 2310
[WARN]          13  CoreFoundation                      0x000000010812b4d2 CFRunLoopRunSpecific + 626
[WARN]          14  GraphicsServices                    0x000000010c70c2fe GSEventRunModal + 65
[WARN]          15  UIKitCore                           0x00000001140e2fc2 UIApplicationMain + 140
[WARN]          16  ClassicAppTest                      0x0000000101b1902c main + 1228
[WARN]          17  libdyld.dylib                       0x000000010a66e541 start + 1
[WARN]  )
[WARN]  libc++abi.dylib: terminating with uncaught exception of type NSException
```

jquick-axway (changes requested):
I don't think we should update the properties if "autoreverse" property is set to `true`, because the view will return back to its original settings at the end of the animation.

But other than that, yes, I agree with this change.

sgtcoolguy:
ok pushed changes to guard against the crash @jquick-axway found on iOS, avoid applying the properties if the animation is autoreversed, and added a unit test for an autoreversed animation (which did surface the crash before I fixed it locally).

jquick-axway:
I tested the newest updates to this PR with the below ticket's attached JS test files... along with some 3D matrix tests I have. Looks good!
https://jira.appcelerator.org/browse/TIMOB-16789

jquick-axway (approved):
CR: Pass

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200430133928
Appc CLI: 8.0.0
JDK: 11.0.6
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 11.4
Device: iPhone X(v13.2), iPhone 11 simulator (v13.4), iPad Pro 12.9 inch simulator (v13.4)

PR #11677, opened 2020-05-01
fix(doc): updated doc

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIDOC-3397


jquick-axway (approved):

sgtcoolguy:
I tweaked this slightly to set two of the dict's properties as optional and provide the defaults they use when omitted - and then merged the PR manually.

PR #11679, opened 2020-05-02
fix(android): Blob imageAsX() methods ignore EXIF orientation from non-file types

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27872

**Note 1:**
Resolved by fetching EXIF orientation via `InputStream` instead of by file path.

**Note 2:**
The blob's "width" and "height" properties will still return the un-rotated size if referencing a JPEG with EXIF orientation. Unfortunately, it needs to stay this way since these are used to size drawables in views. The width/height need to reflect the actual size of the uncompressed bitmap which Google does not pre-rotate. In the future, the solution would be for us to load JPEGs ourselves pre-rotated and pre-flipped (for mirrored images) via libjpeg.

**Test:**
1. Build and run `DownloadImageBlobTest.js` attached to [TIMOB-27872](https://jira.appcelerator.org/browse/TIMOB-27872) on Android.
2. Make sure device has Internet access.
3. Tap on the "Download" button.
4. Select "JPEG - EXIF Rotate 90" in dialog.
5. Verify displayed image is upright.
6. Tap on the "Download" button.
7. Select "JPEG - EXIF Rotate 180" in dialog.
8. Verify displayed image is upright.
9. Tap on the "Download" button.
10. Select "JPEG - EXIF Rotate 270" in dialog.
11. Verify displayed image is upright.
12. Tap on the "Download" button.

_(Note that the EXIF Flip, Transpose, and Transverse tests now correctly show images upright too, but they will also be mirrored. Fixing the mirroring effect would require a larger refactoring of our code in the future.)_


garymathews (approved):
CR: **PASS**

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200512154029
Appc CLI: 8.0.0
JDK: 11.0.6
Node: 10.17.0
Studio: 6.0.0.202005141803
Device: Pixel 3(v10.0), Mi A1(v9.0), Nexus4(v5.1.1), Pixel3(v10.0) emulator, Pixel 3XL(v9.0) emulator

PR #11680, opened 2020-05-03
feat(android): add getScaledTouchSlop() to ListView

m1ga:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27879

**Optional Description:**

Prevents the `scrolling` direction change to quickly. The current implementation will trigger an up/down change when you stop scrolling and your finger moves one pixel.

```
var listView = Ti.UI.createListView({});
var sections = [];

var fruitDataSet = [
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Apple'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
    {properties: { height: 100,title: 'Banana'}},
];
sections.push(fruitSection);

listView.appendSection(sections);
listView.addEventListener("scrolling",function(e){
	console.log(e.direction);
})
win.add(listView);
win.open();
```


jquick-axway:
@m1ga, I don't know if I like the idea of this. This effectively dampens all "scroll" events where it'll only fire if it has moved 5 pixels away from its last position. I'm curious what your use-case is.

m1ga:
@jquick-axway check the JIRA description. The current implementation will fire up/down changes for every pixel and it will create some noise if you stop moving your finger but the the phone will recognize tiny changes and e.g. it will already show/hide your views because it thinks you've change direction. Just putting the finger on the screen will most of the time already fire a up/down event.
Default value can always be 1 so it behaves like before but with 5 the user scroll experience is better

jquick-axway:
@m1ga, oh I see. Sorry about misunderstanding this. What you're doing makes sense.

My only feedback then is:
- We shouldn't hard-code it to 5 pixels. Instead, use Google's [ViewConfiguration.getScaledTouchSlop()](https://developer.android.com/reference/android/view/ViewConfiguration#getScaledTouchSlop()) to get the pixel tolerance needed which is scaled based on device DPI.
- I don't think we need a new `scrollDamping` property. We can just hard-code it to use the touch-slop value like how Google does it internally with scroll views and drag-and-drop.

FYI: We use the touch-slop API to handle nested scrolling for Titanium's `TextFields` and `TextAreas` (link below). It works well and I haven't heard of any complaints.
[TiUIEditText.java#L90-L93](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIEditText.java#L90-L93)

What do you think?

m1ga:
@jquick-axway didn't know about that method! Nice, updated the PR, remove the custom value. Works fine here

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
**FR Passed** Tested using the test case mentioned above, able to see scrolling direction being more stable and not changing constantly (perviously just placing your finger on the screen was enough to trigger the scroll event) 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
API29 Pixel XL emulator
```

ssaddique:

**Test environment**
```
OS Ver:         10.15.3
Xcode Ver:      Xcode 11.6
Appc NPM:       5.0.0
Appc CLI:       8.0.0
Node Ver:       10.17.0
Java Ver:       1.8.0_162
Device:      Pixel 3a (API 30 - Android 11)
```

PR #11683, opened 2020-05-05
feat: add Ti.Local.parseDecimal() method

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27874

**Summary:**
- Parses a localized numeric string to a number.
- Uses current locale or can provide optional locale for 2nd argument.
- Returns `NaN` (not a number) if given an invalid string.
- Can parse localized string returned by `String.formatDecimal()`.
- Expected to be used to parse value from numeric TextFields.


iphone/Classes/LocaleModule.h
```diff
@@ -22,6 +22,10 @@ JSExportAs(getString,
            -(NSString *)getString
            : (NSString *)key withHint
            : (id)hint);
+JSExportAs(parseDecimal,
+           -(NSNumber *)parseDecimal
+           : (NSString *)text withLocaleId
+           : (id)localeId);
```
sgtcoolguy, line 28: So I'm a bit conflicted over the use of `id` for the second optional argument. The JSC API doesn't give us a straight-forward way to handle optional arguments. I just tested, and if you declare it as an `NSString*` (which it should be), but you don't provide the arg - it converts the `undefined` value to `"undefined"` as a String. Which is clearly an annoyance (and probably why I did the same above for `getString`'s `hint` arg).

So I guess we have to either declare as `id` like this so undefined becomes `nil` - or not even define the optional argument and instead pull it out of `[JSContext.currentArguments]`.

tests/Resources/ti.locale.addontest.js
```diff
@@ -0,0 +1,118 @@
...
+	describe('#parseDecimal()', () => {
+		it('compared with String.formatDecimal()', () => {
+			should(Ti.Locale.parseDecimal).be.a.Function();
+
+			const numericValue = 1234567.8;
+
+			let numericString = String.formatDecimal(numericValue);
+			let parsedValue = Ti.Locale.parseDecimal(numericString);
+			should(parsedValue).be.a.Number();
+			should(Math.abs(parsedValue - numericValue)).be.lessThan(Number.EPSILON);
```
sgtcoolguy, line 22: Neat trick - never knew about the use of EPSILON to compare floats.

May be easier to write this as:
``` js
should(parsedValue).be.approximately(numericValue, Number.EPSILON);
```
See https://shouldjs.github.io/#assertion-approximately

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -82,6 +83,39 @@ public String formatTelephoneNumber(String telephoneNumber)
...
+			} else {
+				numberFormat = NumberFormat.getInstance();
+			}
+
+			// Enable thousands separator parsing support. (ex: "1,234,567")
+			numberFormat.setGroupingUsed(true);
+
+			// Remove leading spaces and plus sign. Number format will fail to parse if there.
+			text = text.trim();
+			if ((text != null) && text.startsWith("+")) {
```
sgtcoolguy, line 105: Shouldn't need the null check since you're already trimming without checking above. I assume our bridge layer enforces non-null if the argument is not marked optional? If not, the null check needs to move to the very top.

sgtcoolguy (approved):

ssjsamir (approved):
**FR Passed:** Tested using the following test case below. 

**Test Case:** 

```
var oldNumericValue = 123.456;
var stringValue = String.formatDecimal(oldNumericValue);
var newNumericValue = Ti.Locale.parseDecimal(stringValue);
if (Math.abs(newNumericValue - oldNumericValue) < Number.EPSILON) {
 	console.log('Parsed number matches original number.');
}

var number1 = Ti.Locale.parseDecimal('1,234,567.8', 'en-US');
var number2 = Ti.Locale.parseDecimal('1.234.567,8', 'de-DE');
if (Math.abs(number1 - number2) < Number.EPSILON) {
	console.log('Parsed numbers match.');
}

var result = Ti.Locale.parseDecimal('Invalid');
if (Number.isNaN(result)) {
	console.log('Invalid string returned NaN.');
}

```
**Result**
```
[INFO]   Parsed number matches original number.

[INFO]   Parsed numbers match.

[INFO]   Invalid string returned NaN.
```

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
iphone 8 Sim (14.0 Beta)
API230 Pixel XL2 emulator
```

ssaddique:

**Test environment**
```
OS Ver:         10.15.3
Xcode Ver:      Xcode 11.6
Appc NPM:       5.0.0
Appc CLI:       8.0.0
Node Ver:       10.17.0
Java Ver:       1.8.0_162
Emulator:      Pixel 4 (API 30)
```

PR #11693, opened 2020-05-07
feat(ios): support circular require references

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27888

**Description:**
This is a PR to unify our require implementation between iOS and Android. Android had a mostly JS-based implementation based off Node.js. The native layer uses bindings for native/core modules and loading those. But the bulk of loading JS code is actually done in JS. On iOS, the entire flow was native.

This PR has been carefully rebased into meaningful commits to help Code Reviews (look at the individual commits, the full PR is overwhelming!).

The commits generally go as follows:
- Copy the Node.js require implementation into `common` and hijack iOS `require` to be replaced by it during `ti.main.js` load (and then add a test to verify it fixes the circular require issue in an app)
- Rip out the native implementation of `require` on iOS and do a refactoring of the code to make the underlying bindings cleaner.
- Create a new separate bundle in `common` for `ti.kernel.js` - equivalent to Android's `kroll.js`. This consists of the very core kernel JS scripts to set up the JS environment before `ti.main.js` is loaded.
- Refactor iOS startup to more closely match Android by executing `ti.kernel.js` bootstrap function - the most important/relevant thing it defines is `Module` - the class that does `require`. Then when we're asked to run `ti.main.js` or a service entry point, we run it though `Module.runModule()`.
- Then refactor Android to use the `common/ti.kernel.js` in place of it's old `kroll.js`. The extensions to titanium APIs are moved out into the `common/ti.internal/extensions/ti` folder to occur during `ti.main.js`.


sgtcoolguy:
For some reason, this PR appears to make it so we cannot permanently alter Ti.Blob's prototype from JS to add the `arrayBuffer` method. I *was* able to implement the method natively after cherry-picking the Promise implementation from #10554 and defining `arrayBuffer` in Obj-C.

See:
- d92f05d
- 7c6c9d3

vijaysingh-axway:
@sgtcoolguy 
1. I can see iOS KrollPromise  is included in this PR and PR #10554 as well. I guess it should not be in this PR or is it here for any purpose?
2. Can you resolve conflicts please?

iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -94,6 +94,7 @@ - (void)reportScriptError:(JSValue *)error inJSContext:(JSContext *)context
 
 - (void)showScriptError:(TiScriptError *)error
 {
+  // FIXME: If nativeStack is a string already this fails!
```
janvennemann, line 97: This should be fixed in [#12155](https://github.com/appcelerator/titanium_mobile/pull/12155/files#diff-4aacf67647acf3dcda586104901c0f41add90eb30a18fa7cde15d8add6c460b6R1347)

sgtcoolguy:
@vijaysingh-axway I needed native Promise support for the arrayBuffer Ti.Blob stuff, so #10554 needs to be merged first and then this needs to be rebased over top (and conflicts resolved)

common/Resources/ti.internal/extensions/ti/ui/activityindicator.js
```diff
@@ -0,0 +1,9 @@
+/* globals OS_ANDROID */
+
+if (OS_ANDROID) {
+	// Define constants for ActivityIndicator here for now.
+	Titanium.UI.ActivityIndicator.STATUS_BAR = 0;
```
garymathews, line 5: This doesn't look to be documented. It appears to move `Ti.UI.Android.PROGRESS_INDICATOR_*` constants to `Ti.UI.ActivityIndicator`. Should this be `Ti.UI.ProgressIndicator` instead?

common/Resources/ti.internal/extensions/ti/ui/view.js
```diff
@@ -4,17 +4,15 @@
...
 
-	var View = Titanium.UI.View;
-	var TiWindow = Titanium.TiWindow;
-
-	var _add = View.prototype.add;
+	const _add = View.prototype.add;
 	View.prototype.add = function (child) {
 
-		if ((child instanceof TiWindow)) {
+		if ((child instanceof Titanium.TiWindow)) {
```
garymathews, line 15: Unnecessary double brackets

common/Resources/ti.internal/kernel/android/invoker.js
```diff
@@ -36,49 +35,52 @@
...
+		const names = namespace.split('.');
+		for (const name of names) {
+			let api;
+			// Create a module wrapper only if it hasn't been wrapped already.
+			if (Object.prototype.hasOwnProperty.call(apiNamespace, name)) {
+				api = apiNamespace[name];
+			} else {
+				function SandboxAPI() {
+					// FIXME: Use non-deprecated way to get prototype!
+					var proto = this.__proto__; // eslint-disable-line no-proto
```
garymathews, line 62: Use `const` and `Object.getPrototypeOf(this)` ?

garymathews (approved):
Android changes look good, just very minor things noted above.

common/Resources/ti.internal/kernel/module.js
```diff
@@ -0,0 +1,635 @@
...
+				if (loaded) {
+					return loaded.exports;
+				}
+
+				// Fallback to old Titanium behavior of assuming it's actually an absolute path
+
+				// We'd like to warn users about legacy style require syntax so they can update, but the new syntax is not backwards compatible.
+				// So for now, let's just be quite about it. In future versions of the SDK (7.0?) we should warn (once 5.x is end of life so backwards compat is not necessary)
+				// eslint-disable-next-line max-len
+				console.warn(`require called with un-prefixed module id: ${request}, should be a core or CommonJS module. Falling back to old Ti behavior and assuming it's an absolute path: /${request}`);
```
jquick-axway, line 264: I assume you uncommented this on purpose? If so, then we should say something in the release notes.

Note that we see this warning on every app launch with "com.appcelerator.aca" if that module is not included. Most app developers are not using this module, so this is causing noise in the log.
https://github.com/appcelerator/titanium_mobile/blob/master/common/Resources/ti.internal/aca.js

tests/Resources/require.addontest.js
```diff
@@ -0,0 +1,17 @@
...
+/* eslint no-unused-expressions: "off" */
+'use strict';
+const should = require('./utilities/assertions');
+
+describe('require()', () => {
+	it('can handle circular references', () => {
+		const value = require('./circular.entry');
+		should.exist(value);
+	});
+});
```
jquick-axway, line 17: We can add this unit test directly into the `./tests/Resources/require.js` file now.

build/lib/builder.js
```diff
@@ -148,6 +170,17 @@ class Builder {
...
+		if (!outDir) {
+			outDir = path.join(TMP_DIR, 'common', platform);
+		}
+		await fs.emptyDir(outDir);
+		return Promise.all([
+			this.generateTiMain(platform, babelOptions, outDir),
+			this.generateKernelBundle(platform, babelOptions, outDir),
+		]);
+	}
+
```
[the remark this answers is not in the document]
jquick-axway: Also, we don't really need "/common/Resources/android/ti.kernel.js" to be packaged with the SDK for Android since this JS file is embedded in the C/C++ side. I verified this "ti.kernel.js" is being added as an asset within the APK. It's only about 200kb and not a big deal if bundled within the Android app, but it isn't necessary either.

vijaysingh-axway:
@sgtcoolguy While running the PR using Xcode, I am getting following error - 
`Titanium[33147:305062] [ERROR] /ti.kernel.js:1214
	        fileIndex = JSON.parse(json);
                                ^
SyntaxError: JSON Parse error: Unexpected identifier "undefined"
    at parse ([native code])
    at filenameExists (/ti.kernel.js:1214:32)
    at loadAsFile (/ti.kernel.js:1085:31)
    at loadAsFileOrDirectory (/ti.kernel.js:1017:36)
    at require (/ti.kernel.js:828:49)
    at Object.<anonymous> (file:///ti.main.js:14814:10)
    at loadAsync (file:///ti.main.js:14760:13)
    at Object.<anonymous> (file:///ti.main.js:14811:10)
    at load (/ti.kernel.js:708:23)
    at runModule (/ti.kernel.js:1258:17)

    Foundation 0x0000000104dfeef0 _NSDescriptionWithLocaleFunc + 93
    CoreFoundation 0x000000010adb489c __CFStringAppendFormatCore + 11852`

common/Resources/ti.internal/extensions/ti/ti.blob.js
```diff
@@ -4,22 +4,8 @@
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
-/* globals OS_IOS, OS_VERSION_MAJOR */
 const buffer = Ti.createBuffer({ value: '' });
 const blob = buffer.toBlob();
-
-if (OS_IOS) {
```
vijaysingh-axway, line 11: I think it should be here. There were 2 issues - https://jira.appcelerator.org/browse/TIMOB-27707 and https://jira.appcelerator.org/browse/TIMOB-27350 . Though unit test is passing. I'll check for lower iOS versions tomorrow.

sgtcoolguy: `toString()` is now implemented natively. As for the hack, I think it just works now?

common/Resources/ti.internal/extensions/ti/ui/activityindicator.js
sgtcoolguy: This was copied over (and split out) from he old ui.js in Android's Kroll.js code: https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/js/ui.js#L109-L113

I don't know why it was there int he first place, so maybe we can drop it?

jquick-axway: @sgtcoolguy, I think you're right. The constants match 1-to-1. We should probably delete these constants.

common/Resources/ti.internal/extensions/ti/ti.blob.js
vijaysingh-axway: I tested it. It looks for iOS 13+, it is working fine. For iOS < 13, it is not working. We need hack for iOS < 13.

vijaysingh-axway:
> @sgtcoolguy While running the PR using Xcode, I am getting following error -
> `Titanium[33147:305062] [ERROR] /ti.kernel.js:1214
> fileIndex = JSON.parse(json);
> ^
> SyntaxError: JSON Parse error: Unexpected identifier "undefined"
> at parse ([native code])
> at filenameExists (/ti.kernel.js:1214:32)
> at loadAsFile (/ti.kernel.js:1085:31)
> at loadAsFileOrDirectory (/ti.kernel.js:1017:36)
> at require (/ti.kernel.js:828:49)
> at Object. (file:///ti.main.js:14814:10)
> at loadAsync (file:///ti.main.js:14760:13)
> at Object. (file:///ti.main.js:14811:10)
> at load (/ti.kernel.js:708:23)
> at runModule (/ti.kernel.js:1258:17)
> 
> ```
> Foundation 0x0000000104dfeef0 _NSDescriptionWithLocaleFunc + 93
> CoreFoundation 0x000000010adb489c __CFStringAppendFormatCore + 11852` 
> ```

After new commits, it is working fine. Thanks!

PR #11315, opened 2019-11-05
fix(android): semantic colors with alpha value

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27519

Test case:

<details>
  <summary>Resources/app.js</summary>

```javascript
const image = 'https://upload.wikimedia.org/wikipedia/commons/thumb/2/21/64_365_Color_Macro_%285498808099%29.jpg/1280px-64_365_Color_Macro_%285498808099%29.jpg';
const CFG = {
    color: {}
};
const fileName = 'semantic.colors.json';
const colorsetFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, fileName);
if (colorsetFile.exists()) {
    const colorset = JSON.parse(colorsetFile.read().text);
    Object.keys(colorset).forEach(name => {
        CFG.color[name] = Ti.UI.fetchSemanticColor(name);
    });
}

const win = Ti.UI.createWindow({
    backgroundImage: image
});

const container = Ti.UI.createView({
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    backgroundColor: 'transparent',
    layout: 'vertical',
    center: {
        y: '50%'
    }
});

Object.keys(CFG.color).forEach(name => {
    appendView(name, CFG.color[name])
});

appendView('Not semantic', 'gray');
appendView('Not semantic', 'rgba(255, 0, 255, 0.33)');

function appendView(name, colorValue) {
    const view = Ti.UI.createView({
        backgroundColor: colorValue,
        width: Ti.UI.FILL,
        height: 30
    });
    const label = Ti.UI.createLabel({
        color: 'black',
        text: `${name}: ${colorValue}`,
        center: {x: '50%'}
    });
    view.add(label);
    container.add(view);
}

win.add(container);

win.open();
```

</details>

<details>
  <summary>Resources/semantic.colors.json</summary>
```javascript
{
  "green_100.0": {
    "light": {
      "color": "#00ff00",
      "alpha": "100.0"
    },
    "dark": {
      "color": "#008000",
      "alpha": "100.0"
    }
  },
  "blue_75.0": {
    "light": {
      "color": "#0000ff",
      "alpha": "75.0"
    },
    "dark": {
      "color": "#000080",
      "alpha": "75.0"
    }
  },
  "cyan_50.0": {
    "light": {
      "color": "#00ffff",
      "alpha": "50.0"
    },
    "dark": {
      "color": "#008080",
      "alpha": "50.0"
    }
  },
[...]

PR #11699, opened 2020-05-10
feat: support semantic/named colors for color properties

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27895

Notes:

1. The color lookup is fully backwards compatible for iOS < 11 as well
2. Android does not have a version limitation, since named colors are available since Android API level 1 

This adds the following:
- a new `Ti.UI.Color` proxy on Android
  - this brings *some* parity with iOS (Android doesn't support passing these to color properties directly like iOS does)
  - right now the only real API here is the `toHex()` method that returns the color's hex string as *ARRGGBB* format
- `Ti.UI.Android.getColorResource()` which can take a numeric argument (for the resource id for the color) or a string argument (for the short color name).
  - i.e. `Ti.UI.Android.getColorResource(Ti.Android.R.color.darker_gray)` or `Ti.UI.Android.getColorResource('holo_blue_bright')`
  - it will return a `Ti.UI.Color` proxy instance that wraps the underlying `android.graphics.Color` instance
- color properties can have named colors passed in to them (in addition to our existing support for hex strings, `rgb()`, `rgba()` and some of the standard web color names like `'cyan'`, `'magenta'`, etc)
- `Ti.UI.fetchSemanticColor()` will fall back to using `Ti.UI.Android.getColorResource()` under the hood if the name isn't in the `semantic.colors.json` file but there **is** a `Ti.Android.R.color[name]` entry (and will still return a valid string to be passed to a color property, should be a hex string in this case).


iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
```diff
@@ -30,11 +32,45 @@ + (UIColor *)checkmarkColor
...
+
++ (UIColor *)semanticColorNamed:(NSString *)colorName
+{
+  if ([TiUtils isIOSVersionOrGreater:@"11.0"]) {
+    return [UIColor colorNamed:colorName];
+  }
+
+  if (semanticColors == nil) {
+    NSString *colorsPath = [NSBundle.mainBundle pathForResource:@"semantic.colors" ofType:@"json"];
+    if (![NSFileManager.defaultManager fileExistsAtPath:colorsPath]) {
```
sgtcoolguy, line 44: Do we need to worry about an encrypted version of the file?

iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
```diff
@@ -30,11 +32,45 @@ + (UIColor *)checkmarkColor
...
+    NSData *colors = [NSData dataWithContentsOfFile:colorsPath options:NSDataReadingMappedIfSafe error:nil];
+    semanticColors = [NSJSONSerialization JSONObjectWithData:colors options:NSJSONReadingMutableContainers error:nil];
+  }
+
+  NSDictionary *colorMap = semanticColors[colorName];
+  if (colorMap == nil) {
+    return nil;
+  }
+
+  NSString *currentTraitCollection = TiApp.controller.traitCollection.userInterfaceStyle == UIUserInterfaceStyleDark ? @"dark" : @"light";
```
sgtcoolguy, line 57: This new chunk of code feels like it should live in the `common` JS area, but I don't see how we could do that and have it effectively hijack all the possible places where colors are set (i.e. all the proxies with various color properties).

sgtcoolguy:
@hansemannn If you keep the SDK's npm packages up to date while developing, it should use a husky hook to auto-format the source on IOS when you commit.

Otherwise, try `npm run format:ios` to manually force it to fix formatting on IOS source.

iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
hansemannn: Are JSON resource files encrypted as well? Looking at a recent production build, it doesn't seem to.

iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
hansemannn: Yep, that was the reason to do it here. And since the JSON only needs to be loaded once, it doesn't cause an overhead to process the array (which wouldn't be a problem on the JS side as well though).

hansemannn:
I updated it, but running `npm run format:ios` attempted to format 15+ files that I haven't even touched, so I only commited the related ones. Also, for unit tests, I could add it to the mocha-suite, but since I need some resource files for it as well, I would like to consider adding `no_tests` and perform a pull request to that repo for it.

sgtcoolguy:
@hansemannn You should be able to plop the tests and Resources files into the `tests/Resources` dir and it'll get looped in.

sgtcoolguy:
Note also, the semantic.colors.json file seems to have broken the actual iOS app build?
```
[2020-05-11T15:28:31.980Z] [ERROR] :  ** BUILD FAILED **
[2020-05-11T15:28:31.980Z] [ERROR] :  The following build commands failed:
[2020-05-11T15:28:31.980Z] [ERROR] :  	CpResource /Users/build/jenkins/workspace/ium-sdk_titanium_mobile_PR-11699/titanium-mobile-mocha-suite/scripts/mocha/build/iphone/semantic.colors.json /Users/build/jenkins/workspace/ium-sdk_titanium_mobile_PR-11699/titanium-mobile-mocha-suite/scripts/mocha/build/iphone/build/Products/Debug-iphonesimulator/mocha.app/semantic.colors.json
[2020-05-11T15:28:31.980Z] [ERROR] :  (1 failure)
[2020-05-11T15:28:31.980Z]  
[2020-05-11T15:28:31.980Z] Error: Exited unexpectedly with exit code: 1
[2020-05-11T15:28:31.980Z]     at ChildProcess.prc.on.code (/Users/build/jenkins/workspace/ium-sdk_titanium_mobile_PR-11699/titanium-mobile-mocha-suite/scripts/test.js:467:19)
[2020-05-11T15:28:31.980Z]     at ChildProcess.emit (events.js:198:13)
[2020-05-11T15:28:31.980Z]     at maybeClose (internal/child_process.js:982:16)
[2020-05-11T15:28:31.980Z]     at Socket.stream.socket.on (internal/child_process.js:389:11)
[2020-05-11T15:28:31.980Z]     at Socket.emit (events.js:198:13)
[2020-05-11T15:28:31.980Z]     at Pipe._handle.close (net.js:607:12)
script returned exit code 1
```

iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
jquick-axway: We don't encrypt JSON files under "Resources". So, we should be okay, but it doesn't hurt to test with a device build just in case.

(We do encrypt our internal bootstrap JSON and app properties JSON though. Our build system is capable of doing this for all JSON files if we wanted to.)

hansemannn:
@sgtcoolguy I will check that one! I thought it might be cool to have a semantic.colors.json in development as well, but it seems like some refs of it are taken over to the app template and cause duplicate resource issues there.

hansemannn:
I'm afraid that I won't be able to update the sample app and remove the demo files from the project before next week, so we may need to close this or hand it over!

hansemannn:
@sgtcoolguy I removed the file again and added a (very basic) mocha test: https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/255

sgtcoolguy:
Relates to #11315

iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m
```diff
@@ -30,11 +32,44 @@ + (UIColor *)checkmarkColor
...
+
+  return [Webcolor webColorNamed:colorMap[currentTraitCollection]];
```
sgtcoolguy, line 58: I think this will cause an infinite loop...

sgtcoolguy:
I hijacked this PR, but added some pretty neat stuff. Once @hansemannn mentioned we can't do UI tests I shrugged and moved on... and then thought again and decided to hack it into our test suite. I added a test which sets a `10px` by `10px` view to a named background color (what this PR adds). Then it calls `Ti.UI.View.toImage()` and compares the generated PNG image against a known good PNG using [pixelmatch](mapbox/pixelmatch). Pretty neat! It also surfaced some bugs in our Node Buffer shim and forced me to make a very bare-bones stream shim - so that [pngjs](lukeapage/pngjs), [pixelmatch](mapbox/pixelmatch) and [browserify-zlib](browserify/browserify-zlib) all worked at runtime. If the image differs we can generate a diff PNG and place the expected/actual side-by-side in a folder that Jenkins will archive. If no reference image exists, it can grab that off the simulator and stick it where Jenkins can archive it too (or if you run locally, it'll copy it where it needs it to be).

So, there's a fair number of changes here to support adding image comparisons in our test suite.

hansemannn:
Thats great news @sgtcoolguy! Since we will immediately cherry pick the changes once merged (because we cannot wait for 9.1.0), can you maybe split it into two pull requests via two squash commits? Not sure where to start cherry picking here 😄 Of course only if it doesn't take too much time.

sgtcoolguy:
I do not plan to squash all these commits down. I spent a good deal of time cleaning up the commits around the actual feature vs test vs docs and breaking it apart by platform. Those are contained in the first 12 commits or so. Once merged, the test override stuff will get moved over to the test suite repo in a separate commit. Those commits (and fixes that were required in build scripts and shims) are in the 18 or so commits at the end here. Again, I'm trying to use good conventional commits and not just squash everything down for the sake of having one or two massive commits that are easier for cherry-picking 😉 

If you only care about cherry-picking the feature, I think these are the commits that are solely the fixes/feature required:
- bb397745597433d7623c7574452161de206e7e56
- d86730d4276e1899620d4021461323a89793b8cf
- ab25eef02ee20d9a9a591c4e9e4f2292523fdf93
- 257f67b8ea522905a4b3b125d55d8f6c6dc539db
- 3b8c219683fe7de8bb23c9ad05d8a5d5cc96f3bc

vijaysingh-axway (approved):
iOS changes looks good.

android/titanium/src/java/org/appcelerator/titanium/proxy/ColorProxy.java
```diff
@@ -0,0 +1,61 @@
...
+		if (this.color == null) {
+			return "#000000";
+		}
+		// Padding! If length is less than 8, pad with leading 0s
+		String hex = Integer.toHexString(this.color.toArgb());
+		while (hex.length() < 8) {
+			hex = "0" + hex;
+		}
+		return "#" + hex;
+	}
```
jquick-axway, line 47: I think we should return RGBA like how iOS used to do it.

```java
// Get ARGB color value.
int intValue = this.color.toArgb();

// Convert to RGBA hex string.
String hexString = String.format(
	"#%02X%02X%02X%02X",
	(intValue >>> 16) & 0xFF,
	(intValue >>> 8) & 0xFF,
	intValue & 0xFF,
	intValue >>> 24);
return hexString;
```

android/titanium/src/java/org/appcelerator/titanium/proxy/ColorProxy.java
```diff
@@ -0,0 +1,61 @@
...
+
```
jquick-axway, line 48: Perhaps we should add a `toString()` method in case the color object is appended to a string in JavaScript.
```
@Kroll.method
@Override
public String toString()
{
	return toHex();
}
```

jquick-axway (commented):
There is 1 typo in the doc.

Also, is there a particular reason why this PR is changing from RGBA to ARGB hex strings? Personally, I'd prefer that we stick to RGBA.

android/titanium/src/java/org/appcelerator/titanium/proxy/ColorProxy.java
jquick-axway: Oh hold on. We can't use the `Color.toArgb()` method. It's API Level 26. It will crash on OS versions older than Android 8.0.
https://developer.android.com/reference/android/graphics/Color#toArgb()

So that leaves the `red()`, `green()`, `blue()`, and `alpha()` methods which unfortunately returns the components values as normalized `0.0-1.0` float values. Hmm... using a `Color` object member variable might not be worth it. Pretty much all Android APIs take a raw color `int` value anyways.

jquick-axway (dismissed):

apidoc/Titanium/UI/Color.yml
```diff
@@ -9,18 +9,25 @@ description: |
...
+    See [R.color](https://developer.android.com/reference/android/R.color) for a list of system color names to use to pass to
+    <Titanium.Android.UI.getColorResource> on Android.
 extends: Titanium.Proxy
 since: "9.1.0"
-platforms: [iphone, ipad]
+platforms: [iphone, ipad, android]
 methods:
   - name: toHex
-    summary: Returns a hex string representation of the color
+    summary: Returns a hex string representation of the color (in the RRGGBB or AARRGGBB formats)
```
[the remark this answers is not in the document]
jquick-axway: I can see that our `TiColorHelper.parseColor()` method has a `shortHexPattern` which accepts a `#AARRGGBB` or `#RRGGBB` formatted string. Is this why you switched it over to ARGB from RGBA? (I'm curious what iOS supports now.)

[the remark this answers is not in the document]
sgtcoolguy: Yes, I originally was trying to use RGBA, but then found out that Android natively uses ARGB, and that iOS code we already had used ARGB as well.

So there's clearly a mismatch with CSS here, but I did not want to break existing code.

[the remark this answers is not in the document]
sgtcoolguy: At first, I saw that iOS returned RGBA in TiUtils.hexColorValue - so I aimed for RGBA. But then I noticed under the hood the color parsing code actually expected ARGB hex values: https://github.com/appcelerator/titanium_mobile/blob/master/iphone/TitaniumKit/TitaniumKit/Sources/API/Webcolor.m#L163-L198

At that point since iOS code already had expected ARGB inputs, and Android natively uses ARGB (and expected it in TiColorHelper.parseCOlor, vbecaus eat defers to Color.parseCOlor which expects ARGB) - I opted to just line it all up as ARGB. To move over to RGBA makes sense to align with CSS (and was why I approached it first), but would be a breaking change on both platforms.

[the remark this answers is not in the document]
jquick-axway: Okay. Makes sense. 👍

jquick-axway (approved):
CR: Pass

PR #11708, opened 2020-05-17
fix(ios): fix watchOS targets that include frameworks

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27757

We are using this fix locally for around 4 months now but never had the chance to push the change. I added some inline comments to describe parts of these changes.


iphone/cli/commands/_build.js
```diff
@@ -3409,6 +3408,13 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 								extFrameworkReference = key.split('_')[0];
 								xobjs.PBXFileReference[extFrameworkReference] = extObjs.PBXFileReference[extFrameworkReference];
 								xobjs.PBXFileReference[extFrameworkReference + '_comment'] = child.comment;
+
+								const currentFileRef = Object.assign({}, xobjs.PBXFileReference[extFrameworkReference]);
+
+								if (!currentFileRef.path.includes('extensions/') && currentFileRef.sourceTree === '"<group>"') {
+									xobjs.PBXFileReference[extFrameworkReference].path = `${ext.relPath}/${currentFileRef.path}`;
+									xobjs.PBXFileReference[extFrameworkReference].name = currentFileRef.path;
+								}
```
hansemannn, line 3417: Before, the build failed because the frameworks embedded into the targets would only have the relative path to the build directory, e.g. `MyTarget/MyFramework.framework` instead of the relative Titanium project path.

iphone/cli/commands/_build.js
```diff
@@ -3430,7 +3436,7 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 					});
 				}
 
-				const handledBuildPhases = [ 'PBXSourcesBuildPhase', 'PBXFrameworksBuildPhase', 'PBXResourcesBuildPhase', 'PBXCopyFilesBuildPhase' ];
+				const handledBuildPhases = [ 'PBXSourcesBuildPhase', 'PBXFrameworksBuildPhase', 'PBXResourcesBuildPhase', 'PBXCopyFilesBuildPhase', 'PBXShellScriptBuildPhase' ];
```
hansemannn, line 3439: The `PBXShellScriptBuildPhase` wasn't handled so far, causing optional framework scripts to be skipped.

iphone/cli/commands/_build.js
```diff
@@ -3643,8 +3654,6 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 
 						if (legacySwift) {
 							extBuildSettings.EMBEDDED_CONTENT_CONTAINS_SWIFT = 'YES';
-						} else {
-							extBuildSettings.ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = 'YES';
 						}
```
hansemannn, line 3657: This was an old left over from the Swift 2 times. Nowadays, swift support us enabled by default and the build manages the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` flag on a module level.

iphone/cli/commands/_build.js
```diff
@@ -3448,6 +3454,11 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 						return;
 					}
 
+					if (type === 'PBXShellScriptBuildPhase' && this.deployType !== 'production') {
+						this.logger.warn(`Only include ${type} in "${targetName}" for production, skipping now …`);
```
cb1kenobi, line 3458: This doesn't seem like a Titanium thing. Nevertheless, I think the wording could be less confusing.

Assuming the watch app project has a `PBXShellScriptBuildPhase`, then it will always display this warning for simulator/device builds. By saying "Only include" implies "Remove it" for non-production builds and that doesn't seem like a good recommendation.

Wouldn't it be better to just say "Excluding PBXShellScriptBuildPhase for non-production build"?

hansemannn:
(ping)

ssekhri:
FR Passed

PR #11723, opened 2020-05-20
refactor(android): rewrite old python scripts in Node.js

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27719

**Description:**
In working on PR #11693, I was doing a lot of work on our build/startup for iOS/Android and ran across our usage of these very old scripts to generate under the hood files for Android. We already had some code for doing the C string byte array baking stuff from titanium_mobile_windows, so I copied it over and the work grew to get rid of these last python scripts.

The generation of the bootstrap.js was partially ported from the old python script, but I removed some logic that was unused in the SDK's modern/existing bindings (i.e. having to merge multiple definitions of the same module/proxy together).

Note that there's an nearly-identical process in the native module template: generate-cpp-header-files.js. That implementation intentionally doesn't use npm packages to do it's work. Long-term it should simply point at the SDK to do the work for it (i.e. these new Node scripts should be "formalized" and shipped in the built SDK and that file should just mostly defer to them)


jquick-axway (commented):
CR: Pass

jquick-axway:
We should test this PR with the last released encrypted DB module for backward compatibility just-in-case.

This is because that module's generated "bootstrap.js" also pushes an entry into to the core `invocationAPIs` array. From looking at the module's generated bootstrap, it "should" be fine since it pushes to the array directly via its own `addInvocationAPI()` function.

PR #11725, opened 2020-05-21
Remove python dependency in ios build, remove some unused dependencies

ewanharris:
* Removes a couple unused dependencies
* Moves the localecompiler functionality into the JS build scripts and uses `node-titanium-sdk` to do the heavy lifting. I verified that the `plutil` isn't required for device builds also and made the `app.js` file use the `L` function to test this

Closes [TIMOB-27792](https://jira.appcelerator.org/browse/TIMOB-27792)


sgtcoolguy (approved):
LGTM

PR #11726, opened 2020-05-21
feat(webdialog/ios): use ASWebAuthenticationSession when possible

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/MOD-2621

An FR was done at the module (by me), so I'm not sure if we want this do go in to `in-qe-testing`?


vijaysingh-axway (approved):
CR passed.

PR #11728, opened 2020-05-22
feat: add ability to run in background without UI

jquick-axway:
Work-in-Progress
---

**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27896

**Alloy:**
This PR requires the below Alloy PR to be merged into `appc` CLI 8.1.0.
https://github.com/appcelerator/alloy/pull/959

**Play Services:**
This PR requires the below "ti.playservices" PR.
https://github.com/appcelerator-modules/ti.playservices/pull/39

**Summary:**
- Added new "tiapp.xml" boolean property "run-in-background".
  * When set `true`, JS runtime is bound to the lifetime of the app process. Will be started as soon as app process starts, before UI.
  * When set `false` (the default), JS runtime is bound to the lifetime of the UI.
- Added new read-only `Ti.UI.hasSession` boolean property.
  * Will be `true` if app can host UI and create windows.
  * Will be `false` if app is unable to host any UI yet and is in the background.
- Added new `Ti.UI` event "sessionbegin".
  * Fired when app should create its windows.
  * Only fired when backgrounding feature is enabled.
- Bootstrap changes:
  * Added support for optional `showUI()` async method. Will be invoked on every "sessionbegin" event. (To be used by "ti.playservices" module.)
  * Added promises support to bootstrap `execute()` and `showUI()` functions.
- Added `Ti.Android.startForegroundServices()` method.
  * Needed to start a foreground service while app is in the background.
- Modified `Ti.App.Android.launchIntent` to never return `null`.
  * Now returns main launcher intent as a fallback.
  * Needed when creating an intent in the background without any UI.
- Add ability to start Android app in background after device bootup.
  * Feature automatically works if you add `RECEIVE_BOOT_COMPLETED` to "AndroidManifest.xml".
- Updated "ti.playservices" module:
  * Now uses Google Play Services library version `17.3.0`.
  * Now uses native update dialog when calling `makeGooglePlayServicesAvailable()`.
  * Updated bootstrap script to use this PR's async `showUI()` method.

---
**To-Do:**
- Test with "remote" policy.
- Add unit tests.
- Updates docs.

---
**Background Test:**
1. Create a new Classic App project.
2. Add the below "tiapp.xml" settings to the project.
3. Replace project's "app.js" file with the below.
4. Copy below "MyServices.js" file to project's "Resources" directory.
5. Build and run on Android.
6. Verify that the window shows "Launch Count: 1".
7. Verify that a "Foreground Service" notification appeared in status bar.
7. Back out of the app.
8. In the log, verify timestamps are posted once a second.
9. Re-launch the app.
10. Verify that the window shows "Launch Count: 2".
11. Reboot the Android device.
12. Wait for the "Foreground Service" notification to re-appear in status bar.
13. Tap on the notification.
14. Verify that it launches the app and it shows "Launch Count: 1".

tiapp.xml
```xml
<ti:app>
	<property name="run-in-background" type="bool">true</property>
		<manifest>
			<uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
			<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
		</manifest>
		<services>
			<service url="MyService.js"/>
		</services>
	</android>
</ti:app>
```

app.js
```javascript
var count = 0;
function openWindow() {
	count++;
	window.add(Ti.UI.createLabel({ text: "Launch Count: " + count }));
	window.open();
}

Ti.UI.addEventListener("sessionbegin", function() {
	openWindow();
});
if (Ti.UI.hasSession) {
	openWindow();
}

Ti.Android.startForegroundService(Ti.Android.createServiceIntent({ url: "MyService.js" }));

setInterval(function() {
	var date = new Date();
	Ti.API.info(String.format("@@@ %02d:%02d:%02d", date.getHours(), date.getMinutes(), date.getSeconds()));
}, 1000);
```

MyService.js
```javascript
Ti.API.info("@@@ Posting notification.");
Ti.Android.currentService.foregroundNotify(123, Ti.Android.createNotification({
	contentTitle: "Foreground Service",
	contentText: "Content Text",
		intent: Ti.App.Android.launchIntent,
	}),
}));
```

---
**Test Restore After OS Force Quit:**
1. On Android, go to its "Developer Options" screen.
2. Tap on the "Background process limit" row. (It's near the bottom of the list.)
3. Tap on "No background processes". _(This tells the OS to force quit the last displayed app when launching another app.)_
4. Build and run the below code on the same Android device.
5. In the app, tap on the "Show Child Window" button.
6. Press the Android "Home" button.
7. Tap on another app such as "Gmail".
8. Press the Android "Home" button.
9. Resume the Titanium app.
10. Verify that the app restarts, starting with the splash screen. _(It will launch with a blank window, but quickly restart. Unfortunately, we are stuck with this behavior.)_
11. Change "tiapp.xml" property "run-in-background" to `false`.
12. Repeat steps 4-10.
13. Change device's "Background process limit" setting back to "Standard".

tiapp.xml
```xml
<ti:app>
	<property name="run-in-background" type="bool">true</property>
</ti:app>
```

app.js
```javascript
function openWindow() {
	var openButton = Ti.UI.createButton({ title: "Show Child Window" });
	openButton.addEventListener("click", function(e) {
		var childWindow = Ti.UI.createWindow({ backgroundColor: "orange" });
		var closeButton = Ti.UI.createButton({ title: "Close Me" });
		closeButton.addEventListener("click", function(e) {
			childWindow.close();
		});
		childWindow.add(closeButton);
		childWindow.open();
	});
	window.add(openButton);
	window.open();
}

Ti.UI.addEventListener("sessionbegin", function() {
	openWindow();
});
if (Ti.UI.hasSession) {
	openWindow();
}
```


sgtcoolguy:
Relates to #11689

My understanding on the iOS side is:
- iOS < 13 we could add a "background service" that ran after the app was backgrounded for some short period of time (~30 seconds now?)
- On iOS 13 we can use the new background task Api to schedule longer-heavier duty background work.
I think in either case we can't really show UI - but can do networking (using urlsession), and can pop a local notification.

cc @janvennemann

jquick-axway:
@sgtcoolguy,

Our iOS background service is our own thing and not a built-in iOS feature. When the app's UI is put into the background by pressing the Home button, we execute the service's JS file in its own JS runtime. We terminate the service's JS runtime when the end-user bring the app back into the foreground. So, in this case, the UI is always present. I also question the usefulness of this feature since whatever you can do in that service script's JS runtime you can also do in the main JS runtime.
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L1058-L1088
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiAppiOSBackgroundServiceProxy.m#L29-L37

We also support the following API on all iOS versions. If the app's UI is suspended in the background, this is supposed to resume it in the background. Apple's example use-case for this is for social media apps to update themselves in the background. I haven't figured out how to get this feature working yet, but supposedly the UI is always present in this case as well.
```javascript
Ti.App.iOS.setMinimumBackgroundFetchInterval(Ti.App.iOS.BACKGROUNDFETCHINTERVAL_MIN);
```

I **think** iOS 13's new `BackgroundTask` feature is the only thing that starts up the app in the background without UI and we're supposed to look at Apple's new `UISceneSession` related APIs to see if a new session has been created. (By the way, I named our new Titanium `hasSession` property and `sessionbegin` event after Apple's new `UISceneSession` feature.)

janvennemann:
@jquick-axway Our current background service on iOS is not our own thing. It uses [`beginBackgroundTaskWithExpirationHandler`](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtaskwithexpiratio), which is how you did background work before the new BackgroundTasks framework was introduced in iOS 13. It's not really comparable to the new BackgroundTasks though, since you could only tell the system that the app is going to continue to do some work after it has been backgrounded. And you are probably right that spinning up a new JS runtime is not required since you can do all the work on the main thread (and practically, it was done anyway since our Kroll core always executes JS on the main thread).

The `setMinimumBackgroundFetchInterval` is for the [`backgroundfetch`](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.App.iOS-event-backgroundfetch) event. It is supposed to be used together with the [urlsession](https://github.com/appcelerator-modules/ti.urlsession/blob/master/ios/example/app.js) module, to periodically update the app in background. This is the equivalent of the new [BGAppRefreshTask](https://developer.apple.com/documentation/backgroundtasks/bgapprefreshtask) in iOS 13.

Now, I'm not entirely sure that we even have to prevent the UI/app.js from being loaded on iOS. The app delegate has a [`window`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623056-window) property, which is used to present the app’s visual content on the device’s main screen. We create this programmatically in [`didFinishLaunchingWithOptions`](https://github.com/appcelerator/titanium_mobile/blob/65200b36f467cac70b6d93f7c257ce0f06629f8c/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L340) since we don't have a Storyboard, which would initialize this automatically for us. AFAIK the new iOS 13 scenes only allow you to have multiple windows with different interfaces to your app. I think that UIKit makes sure not actually render any UI when the app is started in the background. So i was wondering why we should prevent the UI/app.js from loading?

jquick-axway:
@janvennemann,

> So i was wondering why we should prevent the UI/app.js from loading?

We don't want to do this. The original question was asking is it possible for an iOS app to start up in the background without UI. So far, the answer is **no**. Our `didFinishLaunchingWithOptions()` method will always create the main window. In regards to this PR, this means my new `Ti.UI.hasSession` property on iOS will always return `true`; it will never return `false`. (That was the whole point of the iOS backgrounding discussion.)

Regarding Apple's new `UISceneSession` related APIs, they don't make it clear if all sessions can be dismissed on iOS. Although, I suspect this can happen in a Mac app via catalyst... since a Mac app can run in the dock without any windows. Regardless, I'm going to change my iOS code for `hasSession` to check if at least 1 session is available to future proof it.

**Side Note:**
@janvennemann, for your info, when you enable this PR's backgrounding feature, then the "app.js" on Android will pretty much always be executed without UI. Meaning the new `Ti.UI.hasSession` property will return `false` on startup. This is actually the normal native behavior on Android. And Android can destroy the UI at any time and the UI will have to be re-created for the same app process next time the end-user taps on the app.

janvennemann:
@jquick-axway Ah ok, i think i might have miss-understood a few things. I thought it was still up to discussion how to get the same working on iOS. But yeah, as it currently stands, i don't see how that would be possible. Thanks for the quick heads up on the Android UI behavior, i didn't know that.

I haven't played around with `UISceneSession` too much yet so i'm curious how you want to check if a session is available? Is that even possible when the app isn't using scenes?

jquick-axway:
> i'm curious how you want to check if a session is available?

I just wrote it now. It looks like the below. I've already tested it on iOS 13 and iOS 10.

```objective-c
- (BOOL)hasSession
{
  BOOL result = YES;
#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
  if ([TiUtils isIOSVersionOrGreater:@"13.0"]) {
    NSSet<UIScene *> *scenes = [[UIApplication sharedApplication] connectedScenes];
    if (!scenes || (scenes.count <= 0)) {
      result = NO;
    }
  }
#endif
  return result;
}
```

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -769,40 +796,68 @@ public boolean isCoverageEnabled()
...
+		ActivityProxy activityProxy = null;
 		final TiRootActivity rootActivity = TiApplication.getInstance().getRootActivity();
-		if (rootActivity == null) {
+		if (rootActivity != null) {
+			activityProxy = rootActivity.getActivityProxy();
+		}
+
+		// If runtime's lifetime is bound to the UI, then do not continue if failed to find root activity.
+		// We only allow this if "run-in-background" setting is enabled in "tiapp.xml".
+		if (KrollRuntime.isBoundToUI() && (rootActivity == null)) {
```
sgtcoolguy, line 827: Looks like you bail before loading the snapshot/ti.main.js here if the app is "running on background". The issue is that may cause some confusion from devs who may expect polyfills/node shims/whatever. I assume we'd have to modify our `common` js code to be more aware of whether there's a UI session to avoid doing some of it's work when in the BG?

android/app/src/main/assets/Resources/app.js
```diff
@@ -1,60 +1,48 @@
...
+		font: { fontSize: 20, fontFamily: 'Helvetica Neue' },
+		textAlign: 'center',
+		width: 'auto'
+	}));
+	tabGroup.addTab(tab1);
+	tabGroup.addTab(tab2);
+	tabGroup.open();
+}
+
+Ti.UI.addEventListener('sessionbegin', function () {
```
sgtcoolguy, line 43: Note sure how to avoid this, but it feels repetitive to hang an event listener and check a property to open the UI.

android/runtime/common/src/js/module.js
```diff
@@ -551,12 +551,13 @@ Module.prototype._runScript = function (source, filename) {
...
-	// an activity. If app was restarted for Service only, we don't want
-	// to go this route. So added currentActivity check. (bill)
-	if (self.id === '.' && self.context.currentActivity) {
+	// Make above require() function global, if not done already.
+	if (!global.require) {
 		global.require = require;
+	}
 
+	// If this is the first script we're running, then initialize the following.
+	if (self.id === '.') {
```
sgtcoolguy, line 560: I see you removed the check to see if we're an activity or service. The toggled behavior here is about two things:
- trying to hook the debugger to pause on app start
- running the main entry point in the global context without the typical wrapping function does for every other JS file.

I'm not sure if the first would work in say a service entry point JS file?

As to the latter, we've been warning devs for a number of versions that the app.js global scope stuff would go away. so I don't know that it matters here anymore if we execute in the global context or in an injected wrapper function. (in other words, I think that's probably fine)

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
jquick-axway: This if-condition maintains our old behavior when backgrounding is disabled where we only support loading JS files once the UI is available. The root activity is responsible for loading "ti.main" and guarantees that our extensions, polyfills, etc. are available. So, we're okay here. (I'll double check our JS-Service handling, which had broken handing before but works fine with our new backgrounding feature.)

When backgrounding is enabled, we load "ti.main" during `Application.onCreate()` which gets called before any activities are created and leave the JS runtime running for the duration of the app process.

android/runtime/common/src/js/module.js
jquick-axway: With the ways things are now, we'll always launch from "ti.main" first. Even if a JS service has been started. We always load "ti.main" from `Application.onCreate()` before anything else.

If backgrounding is disabled, we really shouldn't allow a JS-service to be loaded in the background because launching and backing-out of the UI would terminate the JS runtime and kill all JS-services. It was already a broken concept before I touched the code.

I suppose I could add a safe-guard here and add an if-check for "ti.main" before hooking up the debugger?

jquick-axway:
@sgtcoolguy, setting the Window's `exitOnClose` property to `true` (the default) tells the system to destroy the window and all of its child windows upon close. It does not kill the app process. So, if backgrounding is enabled, the app process keeps running and the next time the end-user launches the app then another "sessionbegin" event will be fired for the same JS runtime.

If you set it to `false`, then it does not destroy/close the window and homes-out instead. The UI stack is preserved. This is the ideal setting when backgrounding is disabled and makes it work more like iOS.

garymathews (commented):
I thought it would be nice if developers could omit the `Ti.UI.hasSession` check (the Window proxy could do it itself?) and also have the ability to re-use Windows. Here's a test case, similar to the one in the PR:
```JS
let count = 1;

const label = Ti.UI.createLabel({ text: 'Launch Count: ' + count });

Ti.UI.addEventListener('sessionbegin', _ => {

    // Resumed from background.
    console.log('sessionbegin');

    // Update label.
    label.text = 'Launch Count: ' + count++;

    // Re-open window.
    window.open();
});

console.log('normal launch');

window.add(label);
window.open();

Ti.Android.startForegroundService(Ti.Android.createServiceIntent({ url: 'MyService.js' }));

setInterval(_ => {
    const date = new Date();
    console.log(String.format('@@@ %02d:%02d:%02d', date.getHours(), date.getMinutes(), date.getSeconds()));
}, 1000);
```
It's nearly functional, just needs validation for task/session on `Ti.UI.Window.open()`

android/app/src/main/assets/Resources/app.js
jquick-axway: So, instead of doing this...
```javascript
//app.js
Ti.UI.addEventListener('sessionbegin', function () {
	openWindow();
}
if (Ti.UI.hasSession) {
	openWindow();
}
```

We could do the below instead and have our "ti.main.js" always fire a "sessionbegin" event after requiring-in the "app.js"
```javascript
//app.js
Ti.UI.addEventListener('sessionbegin', function () {
	openWindow();
}
```
```javascript
//ti.main.js
require("app.js");
if (Ti.UI.hasSession) {
	Ti.UI.fireEvent('sessionbegin');
}
```

However, I don't really like this. I only wanted the "sessionbegin" event to be fired when our new "tiapp.xml" property "run-in-background" is `true`. And when "run-in-background" is `false`, we only start-up the JS runtime **after** the UI session begins natively, meaning this event isn't relevant in this mode... unless... we want to support multiple UI sessions in parallel in the future.

Having he `Ti.UI.hasSession` if-check in the "app.js" (the way I currently have it) means we can easily enable/disable the "run-in-background" property without having to change the "app.js" code. It's compatible with both modes, which means we can stick to 1 app code template for both.

PR #11731, opened 2020-05-24
feat(android): use kotlin in native modules templates

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27906


jquick-axway:
I don't think we should "replace" our default Java module template with Kotlin.

I think we should still default to Java, but offer Kotlin as an option similar to what you did on iOS with Swift and the `--code-base` command line argument.

My personal opinion...
I don't consider Kotlin to be more modern. The 2 languages have their pros/cons. Java has the advantage of being a well known language, easy to hire developers for, and easy to find code examples. Kotlin's advantage is they can easily add new language features that is backward compatible on older Android OS versions.

hansemannn:
Kotlin is the clear way moving forward on Android. All new Android API's in Titanium should be written in Kotlin, just like iOS should (and does) with Swift. Google actively moves away from promoting Java since last year and all Android samples are now Kotlin-first. Titanium can be early here and move to a future proof core and module ecosystem.

While I'm fine with keeping both (for now), we should deprecate Java modules and select Kotlin by default. I think it's super important to act progressively on this topic.

jquick-axway:
> All new Android API's in Titanium should be written in Kotlin

I don't agree. There is no point in using a mix of languages in our core code. Java is not going away and it's easier to hire/find experienced Java developers than Kotlin developers.

> Google actively moves away from promoting Java since last year and all Android samples are now Kotlin-first.

Yes, they do actively promote Kotlin, but Google's usage of Java/Kotlin is still mixed in their newest projects. For example, Google's new covid-19 contact exposure/tracing sample app is written in Java.
https://github.com/google/exposure-notifications-android

> We should deprecate Java modules and select Kotlin by default.

I'm 100% against this.
Java is way too popular of a language to even consider this. Besides, developers should have the freedom of "choice" of which language to use anyways. It's not for us to decide. Also, you can mix Java and Kotlin code in the same module "today", so, this really isn't a big deal either.

hansemannn:
I had this discussions with ObjC guys 5 years ago as well and the language is close to death now. I'm extremely sad to read this feedback, but I will close this PR then. If Titanium makes it another 5 years, we talk again.

m1ga:
I still find the `--code-base` option to be attractive. Just to have the choice between both would be nice to have. Doesn't have to be default

jquick-axway:
I'm thinking when you do an `appc new`, it should automatically prompt the end-user if there is more than 1 template folder... unless you set the `--code-base` argument for the platform. That might be a good way of handling this in the future.

Would we ever support a 3rd template on Android? Perhaps for C/C++. Who knows what the future might hold. :)

PR #11734, opened 2020-05-26
fix(ios): start initial timer id from one

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27898

**Optional Description:**
We use simple number values as timer identifiers, which used to start from `0`. This could cause unexpected behavior when checking for a timer id since it evaluates to a falsy value for the first timer. Always starting and incrementing the timer ids from `1` solves the issue.


vijaysingh-axway (approved):
CR passed.

vijaysingh-axway:
@janvennemann Is it possible to add unit test for this? I think it can be achieved by  appending the log message in a string  and compare it after completion.

janvennemann:
@vijaysingh-axway I thought about that, but the issue is actually pretty hard to test with our test suite. This issue only occurs for the very first timer started after the app launched. However, we use timers all over the place so i don't even get a chance to test for this particular bug.

PR #11735, opened 2020-05-26
fix(ios): start initial timer id from one (9_0_X)

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27898

9_0_X backport of #11734


vijaysingh-axway (approved):
CR passed.

PR #11748, opened 2020-06-03
feat(webdialog/iOS)(9_0_X): use ASWebAuthenticationSession when possible

vijaysingh-axway:
https://jira.appcelerator.org/browse/MOD-2621


sgtcoolguy (approved):
Looks to be a simple module upgrade. So that looks fine. I assume QE has tested/will test the new module version for web dialog...

PR #11753, opened 2020-06-04
fix(ios): guard against nil proxy in TiViewController 

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27930

**Description:**
We're seeing crashes on the test suite in CI. Mainly on iPad runs, but also occasionally iPhone. The crash logs seem to indicate that `TiViewController`'s underlying `_proxy` field is nil. This PR simply guards for that case on all the delegate methods that access `_proxy` (and effectively try to forward delegate method calls).

Note that this should avoid the crash but there may be a deeper root cause that should be fixed as well. My best guess at this point is some odd edge case scenario with `Ti.UI.iOS.NavigationWindow`.


vijaysingh-axway (approved):
I don't think this will have any effect on crash. Message passing on nil object is valid in Objective C.
e.g 
id _proxy = nil;
BOOL isTrue = [_proxy conformsToProtocol:@protocol(TiWindowProtocol)];  will not crash and return 'false'.

sgtcoolguy:
I don't think the issue is that the _proxy is nil, it's that it's a reference to a deallocated proxy instance (i.e. a "zombie"). So I assume the nil check will guard against actually trying to call the conformsToProtocol message and avoid the crash. But the bigger issue is that we have a "zombie" pointer to a deallocated proxy instance still hung onto by the `TiViewController`

ssjsamir (approved):
FR Passed, Unit test suite no longer crashes on iPad tests.

PR #11759, opened 2020-06-07
feat(android): implement Shortcut proxy

garymathews:
- Implement `Ti.UI.createShortcut` for cross-platform shortcuts

##### TEST CASE
```JS
const window = Ti.UI.createWindow({ backgroundColor: 'grey', layout: 'vertical' });
const shortcut = Ti.UI.createShortcut();

let lastShortcutItem;

function test(title, callback) {
	const button = Ti.UI.createButton({ title, top: 50 });
	button.addEventListener('click', callback);
	window.add(button);
}

test('ADD RANDOM SHORTCUT', _ => {
	const rand = Math.floor(Math.random() * (999 - 1)) + 1;
	const item = Ti.UI.createShortcutItem({
    	id: `test_shortcut_${rand}`,
    	title: `SHORTCUT_${rand}`,
    	description: `DESCRIPTION_${rand}`
    });
	lastShortcutId = item;
	shortcut.add(item);
});

test('REMOVE LAST SHORTCUT', _ => {
	try {
		shortcut.remove(lastShortcutItem);
	} catch { }
});

test('REMOVE ALL SHORTCUTS', _ => {
	shortcut.removeAll();
});

test('NUMBER OF SHORTCUTS', _ => {
	alert(`staticItems: ${shortcut.staticItems.length}`);
	alert(`items: ${shortcut.items.length}`);
});

Ti.UI.Shortcut.addEventListener('click', e => {
	console.log('shortcut: ' + JSON.stringify(e, null, 1))
});

window.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27889)


android/modules/ui/src/java/ti/modules/titanium/ui/ShortcutProxy.java
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		return items.toArray(new ShortcutItemProxy[items.size()]);
+	}
+
+	/**
+	 * Obtain array of current active shortcuts.
+	 * @return Array of `ShortcutItemProxy` objects.
+	 */
+	@Kroll.getProperty
```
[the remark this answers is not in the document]
sgtcoolguy: That wasn't my concern, it was whether our annotation processor/binding layer handled when a method had no set/get prefix. I just double-checked and it does (it strips get/set/is prefixes if they exist).

apidoc/Titanium/UI/ShortcutItem.yml
```diff
@@ -30,54 +36,41 @@ properties:
...
+    availability: creation
 
-            win.add(btn);
-            win.open();
+  - name: visible
+    summary: Shortcut visibility.
+    description: Determines if the shortcut is visible.
+    type: Boolean
+    platforms: [android, iphone, ipad]
+    availability: creation
```
sgtcoolguy, line 73: This is not a creation/construction property, this line should be removed.

apidoc/Titanium/UI/ShortcutItem.yml
```diff
@@ -30,54 +36,41 @@ properties:
...
+    deprecated:
+        since: "9.1.0"
+        notes: Use [Titanium.UI.Shortcut.items](Titanium.UI.Shortcut.items) to determine if shortcut is active.
```
sgtcoolguy, line 76: missing newline at end of file

tests/Resources/ti.ui.shortcut.addontest.js
```diff
@@ -0,0 +1,51 @@
...
+	it.android('Ti.UI.Shortcut', () => {
+		should(Ti.UI.Shortcut).not.be.undefined();
+	});
+
+	it.android('createShortcut', () => {
+		should(Ti.UI.createShortcutItem).not.be.undefined();
+		should(Ti.UI.createShortcut).be.a.Function();
+
+		// ONLY compatible with Android 7.1+, end test early.
+		if (Ti.Platform.osname === 'android') {
```
sgtcoolguy, line 24: FYI, You can now use `OS_ANDROID` as a simple guard and the app/sdk build fills it in with a boolean (like with Alloy):
```js
if (OS_ANDROID) {
  // android-specific code
}
```

tests/Resources/ti.ui.shortcut.addontest.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+
+const should = require('./utilities/assertions');
+
+describe('Titanium.UI.Shortcut', () => {
+
+	it.android('Ti.UI.Shortcut', () => {
```
sgtcoolguy, line 15: misleading title? maybe "namespace exists"?

tests/Resources/ti.ui.shortcut.addontest.js
```diff
@@ -0,0 +1,51 @@
...
+		should(Ti.UI.Shortcut).not.be.undefined();
+	});
+
+	it.android('createShortcut', () => {
+		should(Ti.UI.createShortcutItem).not.be.undefined();
+		should(Ti.UI.createShortcut).be.a.Function();
+
+		// ONLY compatible with Android 7.1+, end test early.
+		if (Ti.Platform.osname === 'android') {
+			const version = Ti.Platform.version.split('.');
```
sgtcoolguy, line 25: I'd refactor this to extract it to a top-level Boolean variable, because...

tests/Resources/ti.ui.shortcut.addontest.js
```diff
@@ -0,0 +1,51 @@
...
+			if (parseInt(`${version[0]}${version[1]}`) < 71) {
+				return;
+			}
+		}
+
+		// Create shortcut.
```
sgtcoolguy, line 31: I think you should break out separate tests/sub0suites for each property/method here. Right now you're just checking existence and that's fine and it doesn't look like too much, but eventually we'll ant to expand the tests for each to verify existence, default values, return types, basic workflow, etc.

tests/Resources/ti.ui.shortcutitem.test.js
```diff
@@ -0,0 +1,56 @@
...
+		// ONLY compatible with Android 7.1+, end test early.
+		if (Ti.Platform.osname === 'android') {
+			const version = Ti.Platform.version.split('.');
+			if (parseInt(`${version[0]}${version[1]}`) < 71) {
+				return;
+			}
+		}
+
+		// Create shortcut item.
+		const item = Ti.UI.createShortcutItem({
```
sgtcoolguy, line 32: We often only test the "golden path" and that's a good first start. May also want to consider - what if we don't supply some of these properties? Are some required? If so, what do we expect it to do then, throw an Error? Do we expect them to adopt default values? What if we pass in values with incorrect types - do we expect an Error? implicit conversion?

android/modules/ui/src/java/ti/modules/titanium/ui/ShortcutModule.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.appcelerator.kroll.KrollModule;
+import org.appcelerator.kroll.annotations.Kroll;
+
+/**
+ * This module exist so `Ti.UI.Shortcut.addEventListener()` is defined.
+ * `Ti.UI.createShortcut` is manually defined in `UIModule`.
+ */
+@Kroll.module(parentModule = UIModule.class)
+public class ShortcutModule extends KrollModule
```
sgtcoolguy, line 17: Feels a bit odd to have this module stub so we can hang event listeners. Looks like we also always fire the event at the same time to Ti.App `shortcutitemclick` event - so you could probably hack this pretty easily in a js extension (basically add a dumb Ti.UI.Shortcut add/removeEventListener that registers for Ti.App event and forwards it)

sgtcoolguy (approved):

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200706133155
Appc CLI: 8.0.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Device: MiA1(v9.0) device, Pixel3(v10.0) emulator

PR #11114, opened 2019-08-04
feat: add generic border API (Android)

m1ga:
Android parity for https://github.com/appcelerator/titanium_mobile/pull/11113

**Optional Description:**

```javascript
var win = Ti.UI.createWindow({
	backgroundColor: '#333',
	layout: "horizontal"
});

var view1 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 20,
	width: 200,
	left: 10,
	top: 10,
	borderRadiusEdges: [Ti.UI.BORDER_EDGE_TOP_RIGHT, Ti.UI.BORDER_EDGE_TOP_LEFT]
});
var view2 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 20,
	width: 200,
	left: 10,
	top: 10,
	borderRadiusEdges: [Ti.UI.BORDER_EDGE_TOP_RIGHT, Ti.UI.BORDER_EDGE_BOTTOM_LEFT]
});
var view3 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 20,
	width: 200,
	left: 10,
	top: 10,
	borderRadiusEdges: [Ti.UI.BORDER_EDGE_TOP_LEFT, Ti.UI.BORDER_EDGE_BOTTOM_RIGHT]
});
var view4 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 20,
	width: 200,
	left: 10,
	top: 10,
	borderRadiusEdges: [Ti.UI.BORDER_EDGE_BOTTOM_LEFT, Ti.UI.BORDER_EDGE_BOTTOM_RIGHT]
});
var view5 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 100,
	width: 200,
	left: 10,
	top: 10,
	borderRadiusEdges: [Ti.UI.BORDER_EDGE_TOP_LEFT, Ti.UI.BORDER_EDGE_TOP_RIGHT]
});
var view6 = Ti.UI.createView({
	backgroundColor: '#fff',
	height: 200,
	borderRadius: 20,
	width: 200,
	left: 10,
	top: 10
});

win.add([view1, view2, view3, view4, view5, view6]);
win.open();
```

image

PR #11769, opened 2020-06-11
feat(android): implement Ti.UI.View.borderRadiusCorners

garymathews:
- Continuation of https://github.com/appcelerator/titanium_mobile/pull/11114 from @m1ga
- Implement `Ti.UI.View.borderRadiusCorners`

##### TEST CASE
```JS
const view_a = Ti.UI.createView({
	backgroundColor: 'red',
	width: 100,
	height: 50,
	top: 100,
	borderRadius: 20,
	borderRadiusCorners: [
		Ti.UI.BORDER_CORNER_TOP_LEFT,
		Ti.UI.BORDER_CORNER_TOP_RIGHT
	]
});
const view_b = Ti.UI.createView({
	backgroundColor: 'blue',
	width: 100,
	height: 50,
	top: 150,
	borderRadius: 20,
	borderRadiusCorners: [
		Ti.UI.BORDER_CORNER_BOTTOM_LEFT,
		Ti.UI.BORDER_CORNER_BOTTOM_RIGHT
	]
});

window.add([ view_a, view_b ]);
window.open();
```
- Should see a blue and red square with rounded corners

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27946)

### DIFF
--- a/apidoc/Titanium/UI/View.yml
+++ b/apidoc/Titanium/UI/View.yml
@@ -1323,6 +1323,13 @@ properties:
     type: Number
     default: 0
 
+  - name: borderRadiusCorners
+    summary: An array of corners to use specified <Titanium.UI.View.borderRadius> on.
+    type: Array<Number>
+    constants: Titanium.UI.BORDER_CORNER_*
+    platforms: [android, iphone, ipad]
+    since: "9.1.0"
+
   - name: borderWidth
     summary: Border width of the view.
     description: |
--- a/tests/Resources/ti.ui.view.addontest.js
+++ b/tests/Resources/ti.ui.view.addontest.js
@@ -11,7 +11,7 @@ const should = require('./utilities/assertions');
 
 describe('Titanium.UI.View', () => {
 
-	it.android('borderRadiusEdges', () => {
+	it.android('borderRadiusCorners', () => {
 		const view = Ti.UI.createView({
 			width: 100,
 			height: 100,
--- a/tests/Resources/ti.ui.view.addontest.js
+++ b/tests/Resources/ti.ui.view.addontest.js
@@ -1,0 +1,28 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2020 by Axway, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+const should = require('./utilities/assertions');
+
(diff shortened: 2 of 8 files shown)


tests/Resources/ti.ui.view.addontest.js
```diff
@@ -0,0 +1,28 @@
...
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+const should = require('./utilities/assertions');
+
+describe('Titanium.UI.View', () => {
+
+	it.android('borderRadiusEdges', () => {
```
vijaysingh-axway, line 14:
```suggestion
	it.android('borderRadiusCorners', () => {
```

tests/Resources/ti.ui.view.addontest.js
```diff
@@ -0,0 +1,28 @@
...
+		const view = Ti.UI.createView({
+			width: 100,
+			height: 100,
+			borderRadiusCorners: [
```
vijaysingh-axway, line 18: Shouldn't this give an exception as this property is valid with 'Titanium.UI.View.borderRadius'.

garymathews: I don't think it should, we don't when setting `borderWidth`; this is just a test to see if the properties get set.

apidoc/Titanium/UI/View.yml
```diff
@@ -1323,6 +1323,13 @@ properties:
     type: Number
     default: 0
 
+  - name: borderRadiusCorners
+    summary: An array of corners to use specified <Titanium.UI.View.borderRadius> on.
+    type: Array<Number>
+    constants: Titanium.UI.BORDER_CORNER_*
+    platforms: [android, iphone, ipad]
+    since: "9.1.0"
+
```
jquick-axway, line 1332: This feature is only supported on Android 5.0 and above, right?
In which case we need to add the following...
`osver: {android: {min: "5.0"}}`

jquick-axway: We should also document that if this property is "undefined", then `borderRadius` is applied to all corners. If an array is assigned to this property, then it must specify each corner to apply to radius to. If assigned an empty array, then `borderRadius` is not applied to any other corners.

vijaysingh-axway:
@garymathews Need to update implementation as per given suggestion https://github.com/appcelerator/titanium_mobile/pull/11770#issuecomment-648510542 .

garymathews:
Closing, superseded by https://github.com/appcelerator/titanium_mobile/pull/11796

PR #11770, opened 2020-06-11
feat(iOS): implement Ti.UI.View.borderRadiusCorners

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27305

Note - Merge with android  PR #11769 , which have doc changes .


iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
```diff
@@ -385,9 +385,36 @@ - (void)frameSizeChanged:(CGRect)frame bounds:(CGRect)bounds
...
     [self renderRepeatedBackground:backgroundImage];
   }
+
+  if ([self hasCustomBorderRadii]) {
+    __block UIRectCorner corners;
+    NSArray<NSNumber *> *edges = [proxy valueForKey:@"borderRadiusCorners"];
+    TiDimension borderRadius = TiDimensionFromObject([proxy valueForKey:@"borderRadius"]);
+
+    if (TiDimensionIsUndefined(borderRadius)) {
+      [self throwException:@"Cannot create custom border radius" subreason:@"The border radius is invalid" location:CODELOCATION];
```
vijaysingh-axway, line 395: @garymathews  In iOS , exception is thrown if property `borderRadiusCorners ` is set without setting `borderRadius `. How is behavior in Android?
@sgtcoolguy Should we throw exception in this case?
Or Log a warning message and ignore it? 
Or Ignore it silently?
I am in favor of logging a warning message.

janvennemann:
@garymathews  @vijaysingh-axway i think we should make this more in line with how CSS border radius works, i.e. one can define the border radius for each corner individually instead of just having one radius and set the edges that it applies to. This would make this more future proof in terms of a possible new layout engine and CSS support.

vijaysingh-axway:
> @garymathews @vijaysingh-axway i think we should make this more in line with how CSS border radius works, i.e. one can define the border radius for each corner individually instead of just having one radius and set the edges that it applies to. This would make this more future proof in terms of a possible new layout engine and CSS support.

@sgtcoolguy Need your guidance here.
Do we want to have what Jan has suggested? I guess yes.
Both iOS and android PRs are ready, do we want to modify these PRs or want to create new ticket for that?
 cc @jquick-axway @garymathews

jquick-axway:
@janvennemann, @vijaysingh-axway, @garymathews,

The way it works now, we're **enabling** which corner a border radius is applied to via flags. Those flags are applied via an array... and we can change the property to accept a number where the flags are OR'ed together too.

Alternatively, we can implement this via a dictionary like the below where a different border radius can be applied per corner. I'm okay with the below too. My only question is, is there any practical use-case where you would want to apply different borders radius values to different corners other than turning them off?
```javascript
view.borderRadius = {
	topLeft: 20,
	topRight: 20,
	bottomLeft: 0,  // Can also leave it undefined.
	bottomRight: 0, // Can also leave it undefined.
};
```

sgtcoolguy:
Given [how CSS handles this](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius), we should match. In practical terms this means that:
- There shouldn't be a separate `borderRadiusCorners` property, it should just fall into `borderRadius`
- You should be able to set a single value which gets applied to all 4 corners
- You should be able to provide 2 values, they get applied to `top-left-and-bottom-right` and `top-right-and-bottom-left`
- You can provide 4 values and they apply to: `top-left`, `top-right`, `bottom-right`, `bottom-left`

I think 3 values use case is probably pretty unusual, and if it's not super-easy to implement, I wouldn't bother.
I also think it's probably a bit hard to handle the elliptical radii, so I'd ignore that use case for now.

vijaysingh-axway:
> Given [how CSS handles this](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius), we should match. In practical terms this means that:
> 
> * There shouldn't be a separate `borderRadiusCorners` property, it should just fall into `borderRadius`
> * You should be able to set a single value which gets applied to all 4 corners
> * You should be able to provide 2 values, they get applied to `top-left-and-bottom-right` and `top-right-and-bottom-left`
> * You can provide 4 values and they apply to: `top-left`, `top-right`, `bottom-right`, `bottom-left`
> 
> I think 3 values use case is probably pretty unusual, and if it's not super-easy to implement, I wouldn't bother.
> I also think it's probably a bit hard to handle the elliptical radii, so I'd ignore that use case for now.

@sgtcoolguy Thanks for clarification! @garymathews We need to update implementation.

vijaysingh-axway:
Closing in favor of PR #11801 .

PR #11771, opened 2020-06-11
feat(ios): add tabsTransparent and tabsBackgroundView properties

AndreaVitale:
**JIRA:** https://jira.appcelerator.org/browse/AC-6533

In an app that I'm developing I need to create a custom shape for the TabGroup. 
So I need to add a custom view to use as background. There no way with the current TabGroup properties to set the it with transparent background. 
So I added both `tabsBackgroundView` and `tabsTransparent` properties in order to achieve a result like this:

Schermata 2020-06-11 alle 15 09 51

### DIFF
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -391,6 +391,40 @@ DEFINE_EXCEPTIONS
   controller.tabBar.selectionIndicatorImage = [self loadImage:value];
 }
 
+- (void)setTabsTransparent_:(id)value
+{
+  UITabBar *tabBar = [controller tabBar];
+
+  if ([TiUtils boolValue:value def:NO]) {
+    if (@available(iOS 13, *)) {
+      UITabBarAppearance *appearance = [tabBar.standardAppearance copy];
+
+      if ([TiUtils boolValue:value def:YES]) {
+        [appearance configureWithTransparentBackground];
+      } else {
+        [appearance configureWithDefaultBackground];
+      }
+
+      tabBar.standardAppearance = appearance;
+    } else {
+      UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), NO, 0.0);
+      UIImage *blank = UIGraphicsGetImageFromCurrentImageContext();
+      UIGraphicsEndImageContext();
+
+      tabBar.backgroundImage = blank;
+      tabBar.shadowImage = blank;
+    }
+  }
+}
+
+- (void)setTabsBackgroundView_:(id)value
+{
+  TiViewProxy *child = (TiViewProxy *)value;
+
+  [controller.tabBar addSubview:child.view];
+  [controller.tabBar sendSubviewToBack:child.view];
+}
+
 - (void)setShadowImage_:(id)value
 {
   //Because we still support XCode 4.3, we cannot use the shadowImage property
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
@AndreaVitale I can see that the underlying implementation is different here, but what exactly are the differences between the new properties and `tabsTranslucent` and `tabsBackgroundImage`? It's not immediately/obviously clear to me when you'd use one set of properties versus the others. Do we have a simple code sample to compare?

iphone/Classes/TiUITabGroup.m
```diff
@@ -391,6 +391,40 @@ - (void)setActiveTabBackgroundImage_:(id)value
   controller.tabBar.selectionIndicatorImage = [self loadImage:value];
 }
 
+- (void)setTabsTransparent_:(id)value
+{
+  UITabBar *tabBar = [controller tabBar];
+
+  if ([TiUtils boolValue:value def:NO]) {
```
sgtcoolguy, line 398: You're already checking the value inside the iOS 13 specific block. I think this guard needs to be removed here...

iphone/Classes/TiUITabGroup.m
```diff
@@ -391,6 +391,40 @@ - (void)setActiveTabBackgroundImage_:(id)value
...
+
+      if ([TiUtils boolValue:value def:YES]) {
+        [appearance configureWithTransparentBackground];
+      } else {
+        [appearance configureWithDefaultBackground];
+      }
+
+      tabBar.standardAppearance = appearance;
+    } else {
+      UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), NO, 0.0);
```
sgtcoolguy, line 410: This looks like the code for iOS < 13 where you've set to `true`. Need to handle the `false` case.

iphone/Classes/TiUITabGroup.m
```diff
@@ -391,6 +391,40 @@ - (void)setActiveTabBackgroundImage_:(id)value
...
+      tabBar.shadowImage = blank;
+    }
+  }
+}
+
+- (void)setTabsBackgroundView_:(id)value
+{
+  TiViewProxy *child = (TiViewProxy *)value;
+
+  [controller.tabBar addSubview:child.view];
```
sgtcoolguy, line 424: Just place the subview/child at the back to begin with to avoid the two calls?
```suggestion
  [controller.tabBar insertSubview:child.view atIndex:0];
```

iphone/Classes/TiUITabGroup.m
```diff
@@ -391,6 +391,40 @@ - (void)setActiveTabBackgroundImage_:(id)value
...
+
+      tabBar.backgroundImage = blank;
+      tabBar.shadowImage = blank;
+    }
+  }
+}
+
+- (void)setTabsBackgroundView_:(id)value
+{
+  TiViewProxy *child = (TiViewProxy *)value;
```
sgtcoolguy, line 422: What if this was already set/called before? You'd need to remove the pre-existing background view before inserting the new

sgtcoolguy (changes requested):
cc @vijaysingh-axway I'm still unsure the relation between these new properties and the existing `tabsTrunslucent`/`tabsBackgroundImage`. I would think you'd be able to achieve the intended effect with those, but perhaps not?

PR #11782, opened 2020-06-18
fix(ios): event sessioncompleted does not get all specified values

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27821


apidoc/Titanium/App/iOS/iOS.yml
```diff
@@ -1252,15 +1252,22 @@ events:
 
       - name: message
         summary: A string containing the localized description of the error.
+        description: |
```
sgtcoolguy, line 1255: Our doc validation doesn't accept `description` values for event properties.

This is because our doc scripts do not read/use/look for descriptions of event properties or method parameters. You'll need to fold the details into the summary string. (Or we need to alter our doc scripts to incorporate descriptions for these when converting to jsduck/json/html/etc)

janvennemann (approved):
CR passed, looks good to me! I'll move it to QE review since the notes from @sgtcoolguy are only docs related and can be fixed while QE does the FR.

apidoc/Titanium/App/iOS/iOS.yml
vijaysingh-axway: Updated doc!

ssjsamir (approved):
FR Passed, tested using the test cases and instructions mentioned in: https://jira.appcelerator.org/browse/TIMOB-27821?focusedCommentId=455888&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-455888

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
iphone 8 Sim (14.0 Beta)
```

PR #11795, opened 2020-06-24
feat(android): support touch feedback for all backgrounds or no background

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26315

**Summary:**
- The `Ti.UI.View` properties `touchFeedback` and `touchFeedbackColor` used to only show a tap ripple effect when a `backgroundColor` was applied to the view. This was a documented limitation.
- Now shows touch feedback ripple to the following as well:
  * backgroundImage
  * backgroundGradient
  * backgroundColor set to TRANSPARENT
  * no background
- Note that touch feedback is only applied to a view's background.

**Test:**
1. Build and run the below on Android 5.0 or higher.
2. Tap on "Transparent Background" and verify a yellow ripple appears.
3. Tap on "White Background" and verify a yellow ripple appears.
4. Tap on "Gradient Background" and verify a yellow ripple appears.
5. Tap on "Image Background" and verify a yellow ripple appears only on the image.
6. Tap on "Custom Button" and verify a yellow ripple appears within the blue button.

```javascript
	layout: "vertical",
	scrollType: "vertical",
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
});
scrollView.add(Ti.UI.createLabel({
	text: "\nTransparent\nBackground\n",
	touchFeedback: true,
	touchFeedbackColor: "yellow",
	top: "10dp",
	width: "75%",
}));
scrollView.add(Ti.UI.createLabel({
	text: "\nWhite\nBackground\n",
	color: "black",
	backgroundColor: "white",
	touchFeedback: true,
	touchFeedbackColor: "yellow",
	top: "10dp",
	width: "75%",
}));
scrollView.add(Ti.UI.createLabel({
	text: "\nGradient\nBackground\n",
	color: "white",
	backgroundGradient: {
		type: "linear",
		startPoint: { x: "0%", y: "50%" },
		endPoint: { x: "100%", y: "50%" },
		colors: [ { color: "white", offset: 0.0}, { color: "blue", offset: 1.0 } ],
	},
	touchFeedback: true,
	touchFeedbackColor: "yellow",
	top: "10dp",
	width: "75%",
}));
scrollView.add(Ti.UI.createLabel({
	text: "\nImage\nBackground\n",
	color: "green",
	backgroundImage: "/appicon.png",
	touchFeedback: true,
	touchFeedbackColor: "yellow",
	top: "10dp",
	width: "100dp",
	height: "100dp",
}));
scrollView.add(Ti.UI.createButton({
	title: "Custom Button",
	backgroundImage: "android.resource://android/drawable/panel_picture_frame_bg_focus_blue",
	touchFeedback: true,
	touchFeedbackColor: "yellow",
	top: "10dp",
}));
window.add(scrollView);
window.open();
```


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1190,23 +1183,51 @@ protected boolean canApplyTouchFeedback(@NonNull KrollDict props)
 
 	/**
 	 * Applies touch feedback. Should check canApplyTouchFeedback() before calling this.
-	 * @param backgroundColor The background color of the view.
-	 * @param rippleColor The ripple color.
+	 * @param rippleColor The ripple color to use. Set to null to use system's default ripple color.
 	 */
-	private void applyTouchFeedback(@NonNull Integer backgroundColor, @Nullable Integer rippleColor)
+	private void applyTouchFeedback(Integer rippleColor)
```
sgtcoolguy, line 1188: Any particular reason to drop the `@Nullable`? I kind of like having it be explicit in our APIs so we can use tools to verify checking for null pointers or not...

tests/Resources/ti.ui.view.addontest.js
```diff
@@ -0,0 +1,114 @@
...
+				done();
+			});
+			win.close();
+		} else {
+			win = null;
+			done();
+		}
+	});
+
+	it.android('touchFeedback', (finish) => {
```
sgtcoolguy, line 39: I appreciate the effort in creating a test, but I don't think it's really verifying anything since this is an interactive UI change, correct?

sgtcoolguy (approved):
The changes LGTM. I don't think the test here would achieve anything, unfortunately so likely this just should be added to QE's UI integration suite if possible.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
jquick-axway: Personally, I question the value of the `@Nullable` annotation. All reference types are implicitly nullable. The only benefit it provides is more for IDE intellisense purposes since it reveals the annotations and effectively documents the API which we can already do via JavaDoc comments.

The `@NonNull` annotation is far more valuable since we can use it to trigger compiler/linter errors. So, if it doesn't have the `@NonNull` annotation, then it's assumed nullable.

Anyways, that's my 2 cents. Unless you can think of something else that might benefit from this?

tests/Resources/ti.ui.view.addontest.js
jquick-axway: Right. It doesn't verify anything, but it does exercise the native API to make sure it doesn't crash. So, I guess it's better than nothing? :man_shrugging:

Ideally, QE should test this by hand via a functional/smoke test since this is a visual feature.

ssjsamir (approved):
**FR Passed**, tested using the test case in description and following the mentioned test steps. 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
API30 Pixel XL2 emulator
```

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
sgtcoolguy: I get that things are implicitly Nullable (so Lon as they're not primitives). I saw it more as a marker of what APIs we've "done" since we've got this large legacy codebase that we didn't use the annotations on. If we don't use `@Nullable`, it's unclear if we've already reviewed the API and decided they're all implicitly nullable, or if we simply haven't gotten around to adding annotations.

sgtcoolguy:
Fixed conflict and manually merged via rebase on master.

PR #11800, opened 2020-06-30
Add Kotlin template, replace --code-base with --ios-code-base

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27906

* Adds a new Kotlin template that can be selected using `--android-code-base kotlin` or via the question prompting. If not provided this defaults to `java`
* Adds a new `--ios-code-base` flag in place of `--code-base`. `--code-base` is now hidden and kept for backwards compatibility and only works for iOS projects, we should remove this in SDK 10.
* Changes the create command to copy over common assets from `templates/module/default` first and then copy over the language specific assets. Assets that exist in both the default template and the language template will be overriden.


android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/ExampleProxy.kt.ejs
```diff
@@ -0,0 +1,86 @@
...
+
+	@get:method
+	@get:getProperty
+	@set:method
+	@set:setProperty
+	var message: String
+		get() = "Hello World from my module"
+		set(message) {
+			Log.d(LCAT, "Tried setting module message to: $message")
+		}
```
jquick-axway, line 85: The "message" property should be rewritten like this.
```kotlin
var message: String
	@Kroll.getProperty
	get() = "Hello World from my module"
	@Kroll.setProperty
	set(message) {
		Log.d(LCAT, "Tried setting module message to: $message")
	}
```

FYI: The kotlin compiler will turn the above into 2 methods `getMessage()` and `setMessage()`. You can't annotate the literal property declaration since it's not a method (will cause a compiler error). The compiler turns properties to Java methods in the end. So, we have to annotate the getters/setters instead.

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
```diff
@@ -0,0 +1,80 @@
...
+package <%- moduleId %>;
+
+import org.appcelerator.kroll.KrollModule
+import org.appcelerator.kroll.KrollDict
+import org.appcelerator.kroll.annotations.Kroll.*
+import org.appcelerator.kroll.common.Log
+import org.appcelerator.kroll.common.TiConfig
+import org.appcelerator.titanium.TiApplication
+
+@module(name = "<%- moduleNameCamel %>", id = "ti.<%- moduleId %>")
```
jquick-axway, line 19: Need to prefix it as `@Kroll.`
```kotlin
@Kroll.module(name = "<%- moduleNameCamel %>", id = "ti.<%- moduleId %>")
```

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
```diff
@@ -0,0 +1,80 @@
...
+		@onAppCreate
+		fun onAppCreate(app: TiApplication?) {
+			Log.d(LCAT, "inside onAppCreate")
+			// put module init code that needs to run when the application is created
+		}
+	}
+
+	// Methods
+
+	@method
```
jquick-axway, line 44: Needs to be `@Kroll.method`

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
```diff
@@ -0,0 +1,80 @@
...
+	fun example(): String {
+		Log.d(LCAT, "example() called")
+		return "hello world"
+	}
+	
+	@method
```
jquick-axway, line 50: Needs to be `@Kroll.method`

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
```diff
@@ -0,0 +1,80 @@
...
+	@set:setProperty
+	@set:method
+	var exampleProp: String
+		get() {
+			Log.d(LCAT, "get example property")
+			return "hello world"
+		}
+		set(value) {
+			Log.d(LCAT, "set example property: $value")
+		}
```
jquick-axway, line 79: We have to annotate the getters/setters or else a compiler error will occur.
```kotlin
var exampleProp: String
	@Kroll.getProperty
	get() {
		Log.d(LCAT, "get example property")
		return "hello world"
	}
	@Kroll.setProperty
	set(value) {
		Log.d(LCAT, "set example property: $value")
	}
```

jquick-axway (changes requested):

jquick-axway:
Thanks for taking this on @ewanharris!
I only have an issue with the annotations in the Kotlin source files. They will cause compiler errors.

I assume entering `appc new -t timodule` it will now prompt me by **default** which languages I want for both Android and iOS? (From your code changes, it looks like the answer is yes.)

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/ExampleProxy.kt.ejs
```diff
@@ -0,0 +1,86 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ *
+ */
+
+package <%- moduleId %>
+
+import android.app.Activity
+import org.appcelerator.kroll.KrollDict
+import org.appcelerator.kroll.annotations.Kroll.*
```
jquick-axway, line 14: I don't think we should do a wildcard import, which means prefixing everything with `Kroll.` down below.

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/ExampleProxy.kt.ejs
jquick-axway: Oh hold on. I wasn't paying attention. We **can** annotate the property declaration since you prefixed it with `@get:` and `@set:`. I still think we should prefix it with `@Kroll.` and ditch the `@Kroll.method` part.
```kotlin
@get:Kroll.getProperty
@set:Kroll.setProperty
var message: String
	get() = "Hello World from my module"
	set(message) {
		Log.d(LCAT, "Tried setting module message to: $message")
	}
```

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
```diff
@@ -0,0 +1,80 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ *
+ */
+
+package <%- moduleId %>;
+
+import org.appcelerator.kroll.KrollModule
+import org.appcelerator.kroll.KrollDict
+import org.appcelerator.kroll.annotations.Kroll.*
```
jquick-axway, line 14: We should get rid of the wildcard import.
```kotlin
import org.appcelerator.kroll.annotations.Kroll
```

android/templates/module/kotlin/template/android/src/{{ModuleIdAsFolder}}/{{ModuleNameCamel}}Module.kt.ejs
jquick-axway: My bad. We can annotate the property like this.
```kotlin
@get:Kroll.getProperty
@set:Kroll.setProperty
var exampleProp: String
	get() {
		Log.d(LCAT, "get example property")
		return "hello world"
	}
	set(value) {
		Log.d(LCAT, "set example property: $value")
	}
```

jquick-axway (changes requested):

jquick-axway:
Sorry about all of the review comments @ewanharris.
My comments pretty much boil down to "don't use wildcard imports". I probably could have reviewed this better.
I think this is overall a great change. Thanks!

ewanharris:
@jquick-axway No bother at all! I'm glad you were thorough as this is a pretty accurate picture of me reviewing the Kotlin code I have no idea what I'm doing

I think I've fixed all the annotations and something I'd missed initially where the module ID was being prefixed with `ti`. One question around semicolons (bike shedding commences), it seems they're optional it Kotlin but we have a couple scattered around in the template. Do we want to use them everywhere, or prefer not to?

jquick-axway:
@ewanharris, I agree about **not** using semicolons in our kotlin code templates. They're optional and they're not typically used. The same is true with gradle/groovy scripts. Android Studio's kotlin app templates don't have semicolons either. I didn't even notice them in this PR's template, so, good catch!

jquick-axway (approved):
CR: Pass

ssjsamir (approved):
**FR Passed** Able to:

1. Get prompts for android and ios code bases
2. Able to use various CLI options (--android-code-base with a value of java or kotlin and --ios-code-base with a value of objc or swift) 
3. Able to still use --code-base prompt for the iOS code base
4. Able to build and run a Android kotlin module 

@ewanharris One thing I noticed was when using the --android-code-base or --ios-code-base flags one of the prompts specifies an option to use the other platform only:

```
samir@Samirs-MacBook-Pro desktop % ti create -t module --android-code-base kotlin

Titanium Command-Line Interface, CLI version 5.2.2, Titanium SDK version 9.1.0
Copyright (c) 2012-2017, Appcelerator, Inc.  All Rights Reserved.

Please report bugs to http://jira.appcelerator.org/

Target platform (all|android|ios) [all]:
```

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
API29 Pixel XL emulator
```

ewanharris:
@sgtcoolguy @cb1kenobi, any thoughts on Samir's observation above? I don't think it should be an error case, but maybe we should log a warning if a platform specific codebase is passed in and that platform is not selected?

sgtcoolguy:
@ewanharris Honestly, I'm kind of torn. How do we handle other cases where we have platform-specific options and you then don't select that platform? I assume we just silently ignore them.

ewanharris:
@sgtcoolguy I guess some similar examples would be:

* providing something like `--pp-uuid` (iOS provisioning profile) to a `ti build -p android` command
* providing `--url` (app only) to a `ti create --type module` command

In both those instances we just silently ignore the unknown option with no logging

cb1kenobi:
It's worth noting that platform specific options such as `--pp-uuid` are not defined until the `--platform` has been parsed. Unknown options are naively parsed.

Project creation does a similar thing using the `--type` option, however the create module command doesn't have anything specific with `--platforms`. In theory, the `--platforms` option would add the `--android-code-base` and `--ios-code-base` options.

One solution would be to have `--code-base` list both iOS and Android options and allow multi-select, though `fields` doesn't currently support multiselect.

Another option is to not prompt for code base at all and just always generate both the Java and Kotlin Android versions along with the Obj-C and Swift iOS versions. There would be 2 folders for each platform instead of two. If someone doesn't want the Java version, then can just delete it.

PR #11803, opened 2020-07-02
fix(ios): report width/height of Blob as pixels, not points

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27997

**Optional Description:**
With some sizes and scales, it is impossible to reconstruct the true number of pixels
in JS just with screen density/scale. We need to multiple image scale vs pts natively.
i.e. 10px square image on a 3x device reported width/height of 3, device scale of 3.
Multiplying in JS naively we'd get image size of 9px x 9px, which was incorrect.

Found when an image comparison test failed on #11699 which was the result of using a 10px x 10px view/image on a 3x simulator. (I knew about the points vs pixels issue, but had to work around it by forcing the view to use a multiple of 3 for it's size, since `blob.width * Ti.Platform.displayCaps.logicalDensityFactor` gave us `9`)

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
@@ -68,7 +68,7 @@ static NSString *const MIMETYPE_JPEG = @"image/jpeg";
 {
   [self ensureImageLoaded];
   if (image != nil) {
-    return image.size.width;
+    return image.size.width * image.scale;
   }
   return 0;
 }
@@ -78,7 +78,7 @@ GETTER_IMPL(NSUInteger, width, Width);
 {
   [self ensureImageLoaded];
   if (image != nil) {
-    return image.size.height;
+    return image.size.height * image.scale;
   }
   return 0;
 }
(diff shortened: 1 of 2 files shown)


janvennemann (approved):

ssjsamir (approved):
**FR Passed:** Width/Height now returns the expected values, tested using the test case below. 

**Test case:**
```
win = Ti.UI.createWindow();
  backgroundColor: 'green',
  width: '11px',
  height: '13px'
});
win.add(view);
win.addEventListener('postlayout', function postlayout() {
  win.removeEventListener('postlayout', postlayout);
  const blob = view.toImage();
  console.log('Value should = 11, Value = ' + blob.width);
  console.log('Value should = 13, Value = ' + blob.height);
});
win.open();
```

**Result:**
```
[INFO]   Value should = 11, Value = 11

[INFO]   Value should = 13, Value = 13
```
**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
iphone 8 Sim (14.0 Beta)
```

ssaddique:

**Test environment**
OS Ver:         10.15.3
Xcode Ver:      Xcode 11.6
Appc NPM:       5.0.0
Appc CLI:       8.0.0
Node Ver:       10.17.0
Java Ver:       1.8.0_162
Simulator:       iPhone 11 Pro (13.6)

PR #11809, opened 2020-07-08
fix(ios): setting TableView row layout to "horizontal" or "vertical" crashes

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28001


tests/Resources/ti.ui.tableview.addontest.js
```diff
@@ -0,0 +1,59 @@
...
+				// After adding table, app should not crash
+				win.add(table);
+				finish();
+			} catch (err) {
+				finish(err);
+			}
+		}
+
+		win.addEventListener('postlayout', addTableView);
+		win.add(table);
```
sgtcoolguy, line 56: You're adding the table before the open *and* after post layout. Which one used to cause it to crash? Also we should likely test horizontal since that caused it as well?

vijaysingh-axway: Typo. Removed one before open. Table added on widow anytime causes crash. 
Added rows with 'horizontal' layout as well.

sgtcoolguy (approved):

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 9.1.0.v20200709124407
Appc CLI: 8.0.0
JDK: 11.0.6
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0 beta
iOS simulator: 12.2, 13.5, 14.0 beta

PR #11813, opened 2020-07-10
feat(ios): expose new APIs to use location AccuracyAuthorization

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27976

### DIFF
--- a/apidoc/Titanium/Geolocation/Geolocation.yml
+++ b/apidoc/Titanium/Geolocation/Geolocation.yml
@@ -282,6 +282,63 @@ methods:
     platforms: [iphone, ipad, android]
     since: "5.1.0"
 
+  - name: requestTemporaryFullAccuracyAuthorization
+    summary: Requests the user's permission to temporarily use location services with full accuracy.
+    description: |
+        If your app doesn't have permission to access accurate location (see [locationAccuracyAuthorization](Ti.Geolocation.locationAccuracyAuthorization), 
+        you can use this method to request temporary access to accurate location.  This access will expire automatically, 
+        but it won't expire while the user is still engaged with your app.  So, for example, while your app is in 
+        the foreground your app will retain the temporary access it was granted. Similarly, if your app starts a 
+        Continuous Background Location session with the background location indicator enabled (see [showBackgroundLocationIndicator](Ti.Geolocation.showBackgroundLocationIndicator), 
+        your access to accurate location will remain as long as the background location indicator remains enabled.  
+        This allows your app to provide session-oriented experiences which require accurate location (e.g. fitness or navigation), 
+        even if the user has decided not to grant your app persistent access to accurate location. 
+
+        You need to add key `NSLocationTemporaryUsageDescriptionDictionary` in `tiapp.xml`. See following example -
+
+        ``` xml
+        <ti:app>
+            <ios>
+                <plist>
+                    <dict>
+                        <key>NSLocationTemporaryUsageDescriptionDictionary</key>
+                        <dict>
+                            <key>PurposeKey1</key>
+                            <string>
+                                Specify the reason for accessing the user's location data with full accuracy.
+                                This appears in the alert dialog when asking the user for permission to
+                                access their location.
+                            </string>
+                            <key>PurposeKey2</key>
+                            <string>
+                                Specify the reason for accessing the user's location data with full accuracy.
+                                This appears in the alert dialog when asking the user for permission to
+                                access their location.
+                            </string>
+                        </dict>
+                    </dict>
+                </plist>
+            </ios>
+        </ti:app>
+        ```
+
+        When CoreLocation prepares the prompt for display, it will look at the NSLocationTemporaryUsageDescriptionDictionary key 
+        in your tiapp.xml.  The value should be a dictionary containing usage descriptions.  The purposeKey you provide to 
+        this method must correspond to an entry in that dictionary.
+    parameters:
+      - name: purposeKey
+        summary: |
+            A key in the NSLocationTemporaryUsageDescriptionDictionary dictionary of the app's tiapp.xml file. 
+            The value for this key is an app-provided string that describes the reason for accessing 
+            location data with full accuracy.
+        type: String
+      - name: callback
+        summary: Function to call upon user decision to grant authorization.
+        type: Callback<LocationAccuracyAuthorizationResponse>
+    platforms: [iphone, ipad]
+    osver: {ios: {min: "14.0"}}
+    since: "9.2.0"
+
   - name: reverseGeocoder
     summary: Tries to resolve a location to an address.
     description: |
@@ -485,6 +542,23 @@ properties:
     platforms: [android, iphone, ipad]
     since: "2.0.0"
 
+  - name: ACCURACY_REDUCED
+    summary: |
+        The level of accuracy used when an app isn’t authorized for full accuracy location data.
+    description: |
+        Use with [accuracy](Titanium.Geolocation.accuracy).
+        The accuracy of location data is reduced in both space and time using approaches like 
+        selecting a nearby point of interest and updating the location at most a few times per hour. 
+        The approximate location preserves the user’s country, typically preserves the city, 
+        and is usually within 1–20 kilometers of the actual location.
+        If your app is authorized to access location information with full accuracy, you can use 
+        this constant to access location data as if the app didn’t have that authorization.
+    type: Number
+    permission: read-only
+    platforms: [iphone, ipad]
+    osver: {ios: {min: "14.0"}}
+    since: "9.2.0"
+
   - name: AUTHORIZATION_AUTHORIZED
     summary: |
         A [locationServicesAuthorization](Titanium.Geolocation.locationServicesAuthorization) value
@@ -629,6 +703,22 @@ properties:
     permission: read-only
     platforms: [iphone,ipad]
     since: "3.1.0"
+
+  - name: ACCURACY_AUTHORIZATION_FULL
+    summary: The user authorized the app to access location data with full accuracy.
+    type: Number
+    permission: read-only
+    osver: {ios: {min: "14.0"}}
+    platforms: [iphone,ipad]
+    since: "9.2.0"
+
+  - name: ACCURACY_AUTHORIZATION_REDUCED
+    summary: The user authorized the app to access location data with reduced accuracy.
+    type: Number
+    permission: read-only
+    osver: {ios: {min: "14.0"}}
+    platforms: [iphone,ipad]
+    since: "9.2.0"
 
   - name: accuracy
     summary: Specifies the requested accuracy for location updates.
@@ -808,6 +898,21 @@ properties:
     default: <Titanium.Geolocation.ACTIVITYTYPE_OTHER>
     since: "3.1.0"
     platforms: [iphone,ipad]
+
+  - name: locationAccuracyAuthorization
+    summary: A value that indicates the level of location accuracy the app has permission to use.
+    description: |
+        If the value of this property is <Titanium.Geolocation.ACCURACY_AUTHORIZATION_FULL>, 
+        you can set any constant to [accuracy](Titanium.Geolocation.accuracy). If value is 
+        <Titanium.Geolocation.ACCURACY_AUTHORIZATION_REDUCED> setting [accuracy](Titanium.Geolocation.accuracy)
+        to a value other than <Titanium.Geolocation.ACCURACY_REDUCED> has no effect on the location information.
+    type: Number
+    constants: [Titanium.Geolocation.ACCURACY_AUTHORIZATION_FULL, Titanium.Geolocation.ACCURACY_AUTHORIZATION_REDUCED]
+    default: <Titanium.Geolocation.ACCURACY_AUTHORIZATION_REDUCED>
+    osver: {ios: {min: "14.0"}}
+    since: "9.2.0"
+    platforms: [iphone,ipad]
+    permission: read-only
 
   - name: pauseLocationUpdateAutomatically
     summary: Indicates whether the location updates may be paused.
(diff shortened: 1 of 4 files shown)


tests/Resources/ti.geolocation.addontest.js
```diff
@@ -0,0 +1,38 @@
...
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions');
+
+describe.windowsBroken('Titanium.Geolocation', function () {
+	const isiOS14 =  (parseInt(Ti.Platform.version.split('.')[0]) >= 14);
```
sgtcoolguy, line 13: cc @vijaysingh-axway @jquick-axway @garymathews Seem to me like over time we tend to eventually write some code like this to guard tests (and I assume devs do too to guard codepaths). Any suggestions on an API to help not have to do this ugly parsing? Maybe just shrug and say use the `semver` npm package in your app for version comparisons?

iphone/Classes/GeolocationModule.m
```diff
@@ -848,6 +854,35 @@ - (void)requestLocationPermissions:(CLAuthorizationStatus)authorizationType with
   }
 }
 
+#if IS_SDK_IOS_14
+- (void)requestTemporaryFullAccuracyAuthorization:(NSString *)purposeKey withCallback:(JSValue *)callback
+{
+  if (![TiUtils isIOSVersionOrGreater:@"14.0"]) {
+    DebugLog(@"[WARN] This is available in iOS 14.0+ only");
```
sgtcoolguy, line 861: fire the callback with an error-style event? i.e. code of 1, message stating "Supported on iOS 14+"?

tests/Resources/ti.geolocation.addontest.js
vijaysingh-axway: Can we expose API something like Ti.Utils.isIOSVersionOrGreater('14.0') inside SDK?  Internally in iOS we use https://github.com/appcelerator/titanium_mobile/blob/adf38f33dbaa7dee244d3ef89154dab97c705271/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUtils.m#L215

jquick-axway: On Android, we already have an `API_LEVEL` property. This is typically used over Android OS version major/minor numbers. But there is no cross-platform equivalent of this.
https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Platform.Android-property-API_LEVEL

I think exposing an OS `versionMajor` and `versionMinor` as numeric properties would work well. All platforms can support this, including Android. If we were to re-add Windows support in the future, then we would also need to add a `versionBuild` property.

sgtcoolguy: So a `Ti.Platform.versionMajor` and `Ti.Platform.versionMinor` represented as numbers (integers)?

tests/Resources/ti.geolocation.test.js
```diff
@@ -244,6 +246,30 @@ describe.windowsBroken('Titanium.Geolocation', function () {
...
+		}
+	});
+
+	it.ios('.ACCURACY_AUTHORIZATION_REDUCED', function () {
+		if (isiOS14) {
+			should(Ti.Geolocation).have.constant('ACCURACY_AUTHORIZATION_REDUCED').which.is.a.Number();
+		}
+	});
+
+	it.ios('.locationAccuracyAuthorization', function () {
```
sgtcoolguy, line 267: Add test that verifies `requestTemporaryFullAccuracyAuthorization` method exists (and maybe additionally that on iOS < 14 it immediately calls the callback with an error)

iphone/Classes/GeolocationModule.m
```diff
@@ -848,6 +854,40 @@ - (void)requestLocationPermissions:(CLAuthorizationStatus)authorizationType with
   }
 }
 
+#if IS_SDK_IOS_14
+- (void)requestTemporaryFullAccuracyAuthorization:(NSString *)purposeKey withCallback:(JSValue *)callback
+{
+  if (![TiUtils isIOSVersionOrGreater:@"14.0"]) {
+    NSMutableDictionary *propertiesDict = [TiUtils dictionaryWithCode:1 message:@"Supported on iOS 14+"];
+    NSArray *invocationArray = [[NSArray alloc] initWithObjects:&propertiesDict count:1];
+    [callback callWithArguments:invocationArray];
```
sgtcoolguy, line 863: Why not just use the short-hand array syntax (and drop the alloc/init/release for a one-short one-element array)?
```suggestion
    [callback callWithArguments:@[ propertiesDict ]];
```

iphone/Classes/GeolocationModule.m
```diff
@@ -848,6 +854,40 @@ - (void)requestLocationPermissions:(CLAuthorizationStatus)authorizationType with
...
+                                                                                   NSMutableDictionary *propertiesDict = [TiUtils dictionaryWithCode:0 message:nil];
+                                                                                   if (error != nil) {
+                                                                                     propertiesDict = [TiUtils dictionaryWithCode:1 message:error.description];
+                                                                                   } else {
+                                                                                     propertiesDict[@"accuracyAuthorization"] = @([[self locationPermissionManager] accuracyAuthorization]);
+                                                                                   }
+                                                                                   NSArray *invocationArray = [[NSArray alloc] initWithObjects:&propertiesDict count:1];
+                                                                                   [callback callWithArguments:invocationArray];
```
sgtcoolguy, line 880: same as above - why not use shorthand array syntax?

tests/Resources/ti.geolocation.addontest.js
jquick-axway: Yeah, just like that. That would make it look nice and simple like this...
```javascript
if (OS_IOS && (Ti.Platform.versionMajor >= 14)) {
	// Uses newest iOS feature.
}
```

Since we're pushing devs to use the `OS_IOS` and `OS_ANDROID` constants, do you think we should add new `OS_VERSION_MAJOR` and `OS_VERSION_MINOR` constants too? We can define them in our "ti.main.js" on app startup like this...
```javascript
//ti.main.js
global.OS_VERSION_MAJOR = Ti.Platform.versionMajor;
global.OS_VERSION_MINOR = Ti.Platform.versionMinor;
```

Then the app code can look this...
```javascript
if (OS_IOS && (OS_VERSION_MAJOR >= 14)) {
	// Uses newest iOS feature.
}
```

tests/Resources/ti.geolocation.test.js
vijaysingh-axway: Added unit test for `requestTemporaryFullAccuracyAuthorization `.

iphone/Classes/GeolocationModule.m
vijaysingh-axway: Done. Thanks!

iphone/Classes/GeolocationModule.m
vijaysingh-axway: Updated!

tests/Resources/ti.geolocation.test.js
```diff
@@ -244,7 +246,52 @@ describe.windowsBroken('Titanium.Geolocation', function () {
 		should(Ti.Geolocation.trackSignificantLocationChange).be.true();
 	});
 
+	it.ios('.ACCURACY_REDUCED', function () {
+		if (OS_VERSION_MAJOR >= 14) {
```
sgtcoolguy, line 250: So we define these special filters like `it.ios`, where it skips on other platforms. It may make sense to expand to add ones like `it.ios14Plus` because technically we'd want to skip on older versions, not have a no-op test that always passes. (Alternatively, once we upgrade from ti-mocha to the real mocha package, I think you can explicitly call `this.skip()` inside a test to mark it at runtime)

tests/Resources/ti.geolocation.test.js
```diff
@@ -244,7 +246,52 @@ describe.windowsBroken('Titanium.Geolocation', function () {
...
+		if (OS_VERSION_MAJOR < 14) {
+			return finish();
+		}
+
+		should(Ti.Geolocation.requestTemporaryFullAccuracyAuthorization).be.a.Function();
+		Ti.Geolocation.requestTemporaryFullAccuracyAuthorization('purposekey', function (e) {
+			try {
+				// It will always give error because 'purposekey' is not in tiapp.xml.
+				should(e).have.property('success').which.is.a.Boolean();
+				should(e.success).be.eql(false);
```
sgtcoolguy, line 285:
```suggestion
				should(e.success).be.false();
```

tests/Resources/ti.geolocation.test.js
```diff
@@ -244,7 +246,52 @@ describe.windowsBroken('Titanium.Geolocation', function () {
...
+				should(e).have.property('code').which.is.a.Number();
+				should(e.code).be.eql(1);
+				should(e).have.property('error').which.is.a.String();
+				finish();
+			} catch (err) {
+				finish(err);
```
sgtcoolguy, line 291:
```suggestion
				return finish(err);
```

tests/Resources/ti.geolocation.test.js
```diff
@@ -244,7 +246,52 @@ describe.windowsBroken('Titanium.Geolocation', function () {
...
+		should(Ti.Geolocation.requestTemporaryFullAccuracyAuthorization).be.a.Function();
+		Ti.Geolocation.requestTemporaryFullAccuracyAuthorization('purposekey', function (e) {
+			try {
+				// It will always give error because 'purposekey' is not in tiapp.xml.
+				should(e).have.property('success').which.is.a.Boolean();
+				should(e.success).be.eql(false);
+				should(e).have.property('code').which.is.a.Number();
+				should(e.code).be.eql(1);
+				should(e).have.property('error').which.is.a.String();
+				finish();
```
sgtcoolguy, line 289: typically I'd move this outside the try/catch block, because if an `after` test hook fails this could cause the catch block to fire and call finish again.

sgtcoolguy (approved):
LGTM. My review comments are just suggestions, not required.

apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -282,6 +282,63 @@ methods:
     platforms: [iphone, ipad, android]
     since: "5.1.0"
 
+  - name: requestTemporaryFullAccuracyAuthorization
```
jquick-axway, line 285: Do we have to create a new method? From a cross-platform standpoint, it might be nice to re-use our existing [requestLocationPermissions()](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Geolocation-method-requestLocationPermissions) and [hasLocationPermissions()](https://docs.appcelerator.com/platform/latest/#!/api/Titanium.Geolocation-method-hasLocationPermissions) methods and add the new "purpose" string key as a parameter... or even better add support for a settings dictionary to these methods with a new "purpose" field.

On Android, location permission can be granted temporarily by the end-user too. We can still leverage the `hasLocationPermission()` method to check if it's still granted. And the `Ti.Location.accuracy` property defines the accuracy of the permission we need to look for. On Android, the `ACCESS_FINE_LOCATION` sounds like the equivalent of the "accurate" permission on iOS (which means read from the GPS).
https://developer.android.com/preview/privacy/permissions#one-time

So, can we do something like this?
```javascript
const permissionSettings = {
	authorization: Ti.Geolocation.AUTHORIZATION_ALWAYS,
	purpose: 'PurposeKey1',
};
if (Ti.Geolocation.hasLocationPermissions(permissionSettings)) {
	fetchLocationInfo();
} else {
	Ti.Geolocation.requestLocationPermissions(permissionSettings, (e) => {
		if (e.success) {
			fetchLocationInfo();
		}
	});
}
```

**Question:**
On iOS 14, is the "purpose" key required? Are people going to have location accessing problem if they don't update their location permission handling code?

vijaysingh-axway: @jquick-axway This new API is for requesting for full accuracy if user requires it in app. And it get expires. It is not mandatory. e.g your app do not need  full accurate location, you should simply need location access which will not expire. let's say at some screen you need full accuracy to show use'r location, you have to go for it. See [here](https://developer.apple.com/documentation/corelocation/cllocationmanager/3600217-requesttemporaryfullaccuracyauth?language=objc) for more details

'purpose' key is only required if one is going to access location with full accuracy. It will not break existing implementation.

jquick-axway: So my only beef with this is if the app developer sets `Ti.Location.accuracy` to `ACCURACY_HIGH`, then I think `hasLocationPermissions()` should return `false` if it has high accuracy has expired, because that's what the app desires. That's why I'm question the necessity of these new APIs... that and I think we're making this more complicated for app developers than it needs to be. A portable API that works between all iOS versions and Android would be the ideal.

vijaysingh-axway: If user has granted location permission `hasLocationPermissions()` will be `true`. ACURRACY_HIGH (kCLLocationAccuracyBest) is desirable not guaranteed.
In iOS 14+ when app request for location permission there comes a popup (see attached Image1) which have a button mentioning 'Precise: On'. If user sets it off, app will not get location data with full accuracy. If at some point it is required to get 'full accuracy', with this new API app can ask for permission to get full accuracy in location data with a valid reason (see image2 ). And it will be temporary.
Image1- IMG_1632

image2 - 
IMG_1634

apidoc/Titanium/Geolocation/Geolocation.yml
```diff
@@ -282,6 +282,63 @@ methods:
...
+        You need to add key `NSLocationTemporaryUsageDescriptionDictionary` in `tiapp.xml`. See following example -
+
+        ``` xml
+        <ti:app>
+            <ios>
+                <plist>
+                    <dict>
+                        <key>NSLocationTemporaryUsageDescriptionDictionary</key>
+                        <dict>
+                            <key>PurposeKey1</key>
```
ssekhri, line 306: The key should be "purposekey1" instead of "PurposeKey1"

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 9.2.0.v20200831113302
Appc CLI: 8.1.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0 Beta6
iPhone 7Plus(v14.0 Beta6)

sgtcoolguy:
Holding for @vijaysingh-axway and @jquick-axway to continue discussion about the APIs before we merge this.

sgtcoolguy:
@jquick-axway Opened up https://jira.appcelerator.org/browse/TIMOB-28114 to track investigating a better cross-platform API to hide the platform specific and OS version specific codepaths.

PR #11819, opened 2020-07-14
chore(android): update gradle tools and dependencies

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27938

**Summary:**
- Updated gradle from `6.3.0` to [6.5.1](https://docs.gradle.org/6.5.1/release-notes.html).
- Updated Android gradle build tool from `3.6.2` to [4.0.1](https://developer.android.com/studio/releases/gradle-plugin#4-0-0).
  * Can no longer use Android Studio versions older than 4.0.
- Updated kotlin gradle plugin from `1.3.61` to [1.3.72](https://github.com/JetBrains/kotlin/blob/1.3.70/ChangeLog.md#1372).
- Updated AndroidX dependencies to newest versions

---
**KitchenSink Test:**
1. Build and run [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) on Android.
2. Navigate to every Ui window and verify that everything still works.

**Hyperloop Test:**
1. Build and run [hyperloop-example](https://github.com/appcelerator/hyperloop-examples) on Android.
2. Navigate to every Ui window and verify that everything still works.

**Module Test:**
1. Download the source for the [ti.imagefactory](https://github.com/appcelerator-modules/ti.imagefactory) module.
2. Open the Terminal.
3. `CD` to the `./ti.imagefactory/android` folder.
4. Enter: `appc run -p android`
5. Verify that the module successfully builds and runs.


android/package.json
```diff
@@ -28,10 +28,10 @@
 	"compileSDKVersion": "29",
 	"vendorDependencies": {
 		"android sdk": ">=23.x <=29.x",
-		"android build tools": ">=28.0.3 <=29.x",
+		"android build tools": ">=29.0.2 <=29.x",
```
sgtcoolguy, line 31: feels like a bit of an odd version range to use here...

sgtcoolguy (approved):
file changes LGTM

android/package.json
jquick-axway: To clarify, Android gradle tool v4.0.0 requires SDK build tool v29.0.2 or higher.
https://developer.android.com/studio/releases/gradle-plugin?buildsystem=ndk-build#4-0-0

There isn't a release version of SDK build v30.0.0 yet. It's currently an RC version.
https://developer.android.com/studio/releases/build-tools#notes

Also, we don't use these build tools directly anymore. Our Titanium SDK build scripts do pretty much everything via gradle (except for `adb` which is okay). But I'm trying to keep up with the version ranges here just in case something else depends on them that I'm not aware of.

jquick-axway: Or would you prefer we "uncap" the build tools version and not set a max?
I think that might be okay.

android/build.gradle
```diff
@@ -6,14 +6,14 @@
...
-	ext.kotlin_version = '1.3.61'
+	ext.kotlin_version = '1.3.72'
 
 	repositories {
 		google()
 		jcenter()
 	}
 	dependencies {
-		classpath 'com.android.tools.build:gradle:3.6.2'
+		classpath 'com.android.tools.build:gradle:4.0.0'
```
garymathews, line 16: It looks like `4.0.1` is available

jquick-axway: Google literally released 4.0.1 yesterday. :sweat_smile: 
https://mvnrepository.com/artifact/com.android.tools.build/gradle/4.0.1

I'll update it now. Thanks.

garymathews (approved):

ssjsamir (approved):
**FR Passed:**
Able to build the apps mentioned in the test steps above using the updated gradle tools and dependencies. 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.0.0""
API30 Pixel XL emulator
android23-x86 emulator (nexus 5x)
```

ssaddique:
Ticket verified on build 9.1.0.v20200804082025. 

I can confirm everything builds as normal.

**Test environment**
```
OS Ver:         10.15.3
Xcode Ver:      Xcode 11.6
Appc NPM:       5.0.0
Appc CLI:       8.0.0
Node Ver:       10.17.0
Java Ver:       1.8.0_162
Emulator:      Pixel 4 (API 30)
```

PR #11822, opened 2020-07-15
fix(ios): properly handle zero-length stream writes

sgtcoolguy:
**Description:**
Due to fixing some tests assertions some existing bugs cropped up on iOS. The tests are around "truncating" a file to 0 bytes.

Our stream handling code was broken for writing a 0-byte buffer to a file and would throw an error to the callback. But it should be treated as a success 0-byte write. Additionally I found a possible case using Ti.Filesystem.FileStream where we *did* handle it as a no-op, but wouldn't fire to the callback if called async.

### DIFF
--- a/iphone/Classes/StreamModule.m
+++ b/iphone/Classes/StreamModule.m
@@ -56,6 +56,17 @@
     lengthValue = [[buffer data] length];
   }
 
+  if (lengthValue == 0) {
+    // NO-OP
+    NSMutableDictionary *event = [TiUtils dictionaryWithCode:0 message:nil];
+    if (stream != nil) {
+      [event setObject:stream forKey:@"source"];
+    }
+    [event setObject:NUMINT(0) forKey:@"bytesProcessed"];
+    [self _fireEventToListener:@"io" withObject:event listener:callback thisObject:nil];
+    return;
+  }
+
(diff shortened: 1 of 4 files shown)


iphone/Classes/LocaleModule.m
```diff
@@ -84,6 +84,12 @@ - (NSNumber *)parseDecimal:(NSString *)text withLocaleId:(id)localeId
 
   // Remove localized thousands separators from string if they exist. NSScanner fails to parse them.
   NSString *thousandsSeparator = [locale objectForKey:NSLocaleGroupingSeparator];
+  // They changed France's separator character!
+  // Treat \u00a0 and \u202f interchangeably
+  // https://haacked.com/archive/2020/05/17/french-thousand-separator-mystery/
+  if ([thousandsSeparator isEqualToString:@"\u202f"]) {
+    text = [text stringByReplacingOccurrencesOfString:@"\u00a0" withString:@"\u202f"];
+  }
```
jquick-axway, line 92: I have an alternative solution below. If the localized thousands separator is whitespace, then I strip out all whitespace character types, include ASCII spaces. This allows it to correctly parse the **reverse** issue with a string containing `\u202F` on iOS12 which defaults to `\u00A0`.
https://github.com/appcelerator/titanium_mobile/pull/11824

PR #11826, opened 2020-07-19
fix: fix crashing Ti.Platform.openURL

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27846

This is a severe regression from the SDK 9 changes related to the refactoring of several core module like "Ti.Platform" and "Ti.Geolocation". `JSValue` here are never nil (in this case, by default they are "JS-undefined"). It can probably also have caused other regressions, so please check on those as well. An example test case is added in JIRA, which could be added to the test suite, if it doesn't cause trouble for other test cases because the app is backgrounded.


sgtcoolguy:
Fixed in 31424c6c2e85ad5d18c5d45c156221545e8a0e22 (and cherry-picked to 9_1_X and 9_3_X)

PR #11847, opened 2020-07-27
feat(android): add passcode fallback to "ti.identity" module

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/MOD-2588

**Summary:**
When the module has been configured with `AUTHENTICATION_POLICY_PASSCODE`, then it should fallback to using password/pin/touch-pattern if fingerprint scanner is not available.

**Note:**
This is only supported on Android 6.0 and higher. This module's `isSupported()` method will always return `false` on older Android OS versions, even if a passcode is enabled.


sgtcoolguy (approved):
LGTM

PR #11850, opened 2020-07-28
chore(android)(9_3_X): compile with API Level 30

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28046

**Summary:**
- Compiles against API Level 30 (aka: Android 11).
- Does not "target" API Level 30 yet until all breaking-changes have been resolved.
  * Will update "package.json" file's "andoid sdk" to target `30.x` in a future PR.


jquick-axway:
@sgtcoolguy, there appears to be an Android SDK folder permission issue on the Jenkins node. The gradle build system is attempting to auto-install SDK Platform 30 (aka: API Level 30; aka: Android 11), but it's failing.

sgtcoolguy:
For our build noes, traditionally we have a special admin user with sudo rights that installs most things, and the actual builds run as a separate "normal" build user. So in this case the SDK/NDK installs as done by the admin and the build user doesn't have write permissions on the folders. @garymathews has altered our chef scripts to install android 30 and that is rolling out to the nodes now. But perhaps long-term since gradle can auto-install SDK components, I just alter the chef scripts to make the SDK root dir world-writable?

jquick-axway:
> But perhaps long-term since gradle can auto-install SDK components, I just alter the chef scripts to make the SDK root dir world-writable?

Yes, that should work

Note that gradle auto-downloads/installs library dependencies to the Android SDK folder too. I'm curious why I wasn't seeing this after updating our dependency versions. Or were you updating the installed libraries in parallel?

garymathews (approved):

jquick-axway:
Would anyone object to skipping QE testing and just merging it?
It already builds/runs our test suit and my own apps locally just fine.
(We need this merged to start working on our other Android 11 tickets.)

sgtcoolguy (approved):

PR #11852, opened 2020-07-28
feat: update template for angular 9

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27856

**Optional Description:**

Updated template for Titanium Angular with support for the new Webpack build pipeline and Angular 9.

**Test case:**

1. Create a new Titanium Angular app with `ti create --template angular-default` or `appc new --ng`
2. Build and run the app
3. Verify that the example works. You should see a very simple app with a welcome screen shown on first launch. Tapping on the button opens a second screen, which is the main view of the app. The main screen shows a simple data binding example with a button and a label. Clicking the button will update the label text, and reset after 5 taps.


android/templates/app/angular-default/template/platform/android/res/values/colors.xml
```diff
@@ -5,4 +5,4 @@
     <color name="angularLightBlue">#63a4ff</color>
     <color name="angularRed">#de0032</color>
     <color name="windowBackground">#fafafa</color>
-</resources>
+</resources>
```
sgtcoolguy, line 8: Remove this change/add back EOF newline

lint-staged.config.js
```diff
@@ -10,6 +21,15 @@ module.exports = {
 	'iphone/TitaniumKit/TitaniumKit/Sources/API/TopTiModule.m': [
 		'npm run ios-sanity-check --'
 	],
-	'!(**/locales/**/*).js': 'eslint',
+	'*.js': async files => {
+		const filtered = await asyncFilter(files, async file => {
```
sgtcoolguy, line 25: Curious why this was added. Won't eslint ignore the passed in file if it's in the `.eslintignore` file anyways?

templates/app/angular-default/template/README.md
```diff
@@ -2,6 +2,10 @@
```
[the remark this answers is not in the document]
sgtcoolguy: Same with `templates/app/angular-default/template/src/assets/android/appicon.png` below and `android/templates/app/angular-default/template/src/assets/android/appicon.png` above. So I assume this is intentional. Just curious why you wouldn't align the launch icons and the app icons.

sgtcoolguy (changes requested):

templates/app/angular-default/hooks/post-create.js
```diff
@@ -1,41 +1,56 @@
...
+		await updatePackageJson(projectPath, { name: projectName });
 		logger.info('Installing Angular project dependencies');
-		let errorOutput = '';
-		let npmExecutable = 'npm';
-		const spawnOptions = { cwd: path.dirname(packageJsonPath) };
-		if (process.platform === 'win32') {
-			spawnOptions.shell = true;
-			npmExecutable += '.cmd';
-		}
+		await installDependencies(projectPath);
```
ewanharris, line 16: Before I setup the SDK to point to my local titanium-angular, this would error and cause an unhandled promise rejection. I guess we either need to make sure the SDK handles these gracefully or maybe try/catch these?

ewanharris (changes requested):

templates/app/angular-default/template/README.md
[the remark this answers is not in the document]
janvennemann: I wasn't really happy with having only one logo. Previously it was only the Angular logo, but then our Titanium logo is completely missing which i didn't like. I then only used the Titanium logo, but then nothing indicated that this was an Angular app, which i also wasn't happy with. So i settled for the Titanium logo as the app icon, and the Angular logo as the launch logo.

Now that i'm thinking about it again, it maybe useful to have kind of a mashup logo that has both our Titanium Logo and the Angular logo in it. I'll ask marketing if they can come up with something.

janvennemann:
@ewanharris @sgtcoolguy addressed review comments and updated the PR to use a shared post-create hook from #11858

ewanharris (approved):
Changes look good to me. Should this land after #11858 now that it has a dependency on the post-create hook from that PR?

janvennemann:
@ewanharris Yup, that's the plan!

PR #11856, opened 2020-07-31
fix(android)(9_1_X): amend Ti.UI.Shortcut implementation

garymathews:
Note to QE
---
- We need to merge PR #11857 first.
- Afterwards, we need to update this PR to re-enable the Android unit tests.

---
**Summary:**
- Amend `Ti.UI.Shortcut` implementation so `Ti.UI.createShortcut()` is not necessary

##### TEST CASE
```JS
	backgroundColor: 'grey',
	layout: 'vertical'
});

let lastShortcutItem;

function test(title, callback) {
	const button = Ti.UI.createButton({ title, top: 50 });
	button.addEventListener('click', callback);
	window.add(button);
}

test('ADD RANDOM SHORTCUT', _ => {
	const rand = Math.floor(Math.random() * (999 - 1)) + 1;
	const item = Ti.UI.createShortcutItem({
    	id: `test_shortcut_${rand}`,
    	title: `SHORTCUT_${rand}`,
    	description: `DESCRIPTION_${rand}`
    });
	lastShortcutItem = item;
	Ti.UI.Shortcut.add(item);
});

test('REMOVE LAST SHORTCUT', _ => {
	try {
		Ti.UI.Shortcut.remove(lastShortcutItem);
	} catch (e) {
		console.error(e);
	}
});

test('REMOVE ALL SHORTCUTS', _ => {
	Ti.UI.Shortcut.removeAll();
});

test('NUMBER OF SHORTCUTS', _ => {
	alert(`staticItems: ${Ti.UI.Shortcut.staticItems.length}`);
	alert(`items: ${Ti.UI.Shortcut.items.length}`);
});

Ti.UI.Shortcut.addEventListener('click', e => {

	// e should contain 'item' property.
	console.log('shortcut: ' + JSON.stringify(e, null, 1))
});

window.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27889)


android/modules/ui/src/java/ti/modules/titanium/ui/ShortcutModule.java
```diff
@@ -6,21 +6,239 @@
...
+	 * @return Array of `ShortcutItemProxy` objects.
+	 */
+	@Kroll.getProperty
+	public ShortcutItemProxy[] staticItems()
+	{
+		if (!shortcutCapability || shortcutManager == null) {
+			return null;
+		}
+
+		return toShortcutItems(shortcutManager.getPinnedShortcuts());
```
jquick-axway, line 129: I think we need to call `ShortcutManager.getManifestShortcuts()` instead.
https://developer.android.com/reference/android/content/pm/ShortcutManager#getManifestShortcuts()

Google's docs suggest that pinned shortcut list can contain both static and dynamic shortcuts.
https://developer.android.com/guide/topics/ui/shortcuts/creating-shortcuts#pinned

jquick-axway (dismissed):
@garymathews ,

We need to re-enable the Android unit test here...
https://github.com/garymathews/titanium_mobile/blob/TIMOB-27889_91X/tests/Resources/ti.ui.shortcut.test.js#L15-L20

And here...
https://github.com/garymathews/titanium_mobile/blob/TIMOB-27889_91X/tests/Resources/ti.ui.shortcutitem.test.js#L15-L20

Then I think we're done. Thanks!

PR #11858, opened 2020-07-31
feat: webpack template for classic apps

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27716

**Optional Description:**

This is a revamped version of the default two-tabbed app template for Webpack enabled projects. It adds a bunch of additional stuff that isn't in the original default template, but where i felt that users will probably add it anyway sooner or later.

- Two-tabbed UI example with a very basic layout using labels on the first tab, and a more complex layout example on the second tab.
- New app icon and launch image. They are placed in separate files so they can easily be changed individually.
  - Android: The app icons were generated using the Image Asset Studio in Android Studio and simply dropped into the app instead of using the `appicon.png`, enabling rounded and adaptive icons. The launch image is a single 9-patch PNG file. 
  - iOS: The app icons are generated from `DefaultIcons.png` in the project root, while the launch image comes from `LaunchLogo.png`.
  - **Note:** We really need to get some structure into this. The sources for these are scattered around the project, which makes it quite confusing where to put all the files.
- Custom theme for Android so users can easily change it. I figured that most users need this anyway at some point, so why not give them a minimal example to start with. I used it to align the look and feel of the app so it looks similar on both Android and iOS.
- FontAwesome icon font to demonstrate custom font usage.
- README file with basic build instructions and further links for how-tos and API docs
- Source linting using eslint
- VSCode extension suggestion. When users open the newly created project with VSCode and they don't have our extension, it will be suggested by VSCode.

**Note**

~~The updated Angular template and this template both contain a very similar hook to do various post project creation tasks like installing NPM dependencies. I plan to move this hook out of the templates to make it reusable. I'll do this next week in one of these two new template PRs.~~ This PR now contains the shared post-create hook for Webpack enabled project and needs to be merged before #11852 can be tested.


android/templates/app/webpack-default/template/platform/android/res/values/colors.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sgtcoolguy, line 1: Is this file something that may get generated/overridden by the `semantic.colors.json` file for light/dark themes?

templates/app/webpack-default/hooks/post-create.js
```diff
@@ -0,0 +1,73 @@
...
+		const projectName = cli.argv.name;
+		const projectPath = path.join(cli.argv['workspace-dir'], projectName);
+		try {
+			await Promise.all([
+				copyGitIgnore(projectPath),
+				updatePackageJson(projectPath, { name: projectName }),
+				renderReadme(projectPath, { name: projectName }),
+				installDependencies(projectPath, logger)
+			])
+			next();
```
sgtcoolguy, line 21: This pattern comes up a lot in our unit tests, and after some weirdness I have moved to the pattern like so:
``` js
try {
  // do something
} catch (err) {
  return next(err);
}
next();
```

Otherwise technically the call to `next()` in the try block could throw an `Error` and trigger the `catch` block calling `next(err);` again.

templates/app/webpack-default/template/src/home.js
```diff
@@ -0,0 +1,34 @@
...
+const wrapper = Ti.UI.createView({ layout: 'vertical', height: Ti.UI.SIZE });
+const searchIcon = faIcon('edit', 96, { color: '#ccc' });
+wrapper.add(searchIcon);
+const hint = Ti.UI.createLabel({
+	attributedString: Ti.UI.createAttributedString({
+		text: 'Edit home.js to change this view!',
+		attributes: [ {
+			type: Ti.UI.ATTRIBUTE_FOREGROUND_COLOR,
+			value: '#E82C2A',
+			range: [ 5, 7 ]
```
sgtcoolguy, line 15: totally unrelated to this PR, but this syntax really confuses me for range. I guess since I'm used to mathematical ranges being declared like `(0, 5]` or `[1, 6]` or `[1, 6)` where `(` is exclusive and `[` is inclusive. Obviously this is JS syntax and an array, but my brain immediately thinks: "oh, he's selecting text indices 5 to 7 inclusive" when actually you're selecting text starting at index 5, but for the next 7 characters. Not sure if anyone has any thoughts on making this more obvious without breaking backwards compatibility....
cc @jquick-axway @ewanharris @garymathews @vijaysingh-axway

templates/app/webpack-default/template/src/profile.js
```diff
@@ -0,0 +1,159 @@
...
+	top: 30,
+	width: 120,
+	height: 120,
+	borderRadius: 60
+});
+scrollView.add(profilePic);
+const username = Ti.UI.createLabel({
+	text: 'Verona Blair',
+	font: {
+		fontSize: 28
```
sgtcoolguy, line 32: Another nitpick unrelated to this PR: why do we name the property `fontSize` if it's going to be inside a `font` object. Seems redundant.

templates/app/webpack-default/template/src/utils.js
```diff
@@ -0,0 +1,90 @@
...
+		options = style;
+		style = null;
+	}
+	if (style === null) {
+		style = 'solid';
+	}
+	const fontFamily = fontMap[style];
+	return Ti.UI.createLabel(Object.assign({}, {
+		font: {
+			fontFamily,
```
sgtcoolguy, line 85: same deal as above, feels very redundant to me to preface these properties with `font` in the name if they'll be under a font object/property.

templates/app/webpack-default/template/src/utils.js
```diff
@@ -0,0 +1,90 @@
...
+	return Ti.UI.createTab(mergedOptions);
+}
+
+/**
+ * Checks if the current device runs the specified platform.
+ *
+ * @param {string} platform Platform name to check
+ * @return {boolean} True if current device is running on the specified platform, false if not.
+ */
+export function runs(platform) {
```
sgtcoolguy, line 35: So [appcelerator/babel-plugin-transform-titanium](https://github.com/appcelerator/babel-plugin-transform-titanium#special-defines) actually has `OS_IOS` and `OS_ANDROID` "constants" that it will pre-populate to guard platform-specific code. (It also tries to sniff some common ways of detecting platform like the `Ti.Platform.osname` checks against constants, and injecting values for `Ti.Platform.osname/name`)

It may be more straight-forward and lead to reduced dead-code to use those directly in the app rather than extract a method like this.

android/templates/app/webpack-default/template/platform/android/README.md
```diff
@@ -0,0 +1,13 @@
+# Android Platform Folder
+
+This folder contains various files used für the native Android build.
```
sgtcoolguy, line 3:
```suggestion
This folder contains various files used for the native Android build.
```

templates/app/webpack-default/template/src/home.js
janvennemann: I can totally relate. While writing this i was like, what the hell why is this range not doing what i want? This does not make any sense! Until i eventually looked up the docs and found out that the second value is the length, and not the end of the range.

templates/app/webpack-default/template/src/utils.js
janvennemann: Ah crap, i totally forgot those additions to our transform plugin. I'll switch it out to use the constants!

android/templates/app/webpack-default/template/platform/android/res/values/colors.xml
janvennemann: AFAIK we generate a `ti.semantic.colors.xml` for that, so we don't touch the default `colors.xml`.

templates/app/webpack-default/template/src/home.js
jquick-axway: We could modify the "range" property to accept either a dictionary or array. With a dictionary, you effectively have named parameters, making it more obvious how it works.
```javascript
var attributedString = Ti.UI.createAttributedString({
	text: "Hello World",
	attributes: [{
		type: Ti.UI.ATTRIBUTE_UNDERLINES_STYLE,
		range: {  // Add support for dictionary.
			index: 0,
			length: 5,
		},
	}, {
		type: Ti.UI.ATTRIBUTE_BACKGROUND_COLOR,
		value: "red",
		range: [0, 5],  // <- The old way.
	}],
});
```

templates/app/webpack-default/template/src/profile.js
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+/* global tabGroup */
+
+import { createWindow, createTab, faIcon } from './utils';
+
+const title = 'Profile';
+const window = createWindow({ title, layout: 'vertical' });
+
+const scrollView = Ti.UI.createScrollView({ layout: 'vertical' });
```
jquick-axway, line 17: You need to set creation property `scrollType` to `"vertical"` or else a warning will be logged on Android. This is because Android's native `ScrollView` does not support scrolling both vertically and horizontally like iOS. It needs to be told upon creation which direction to support. (It defaults to vertical, but will log a warning if the property is missing.)
```javascript
const scrollView = Ti.UI.createScrollView({
	layout: 'vertical',
	scrollType: 'vertical'
});
```

templates/app/webpack-default/template/src/profile.js
```diff
@@ -0,0 +1,162 @@
...
+	iconTint: '#E82C2A',
+	iconBgTint: '#FFC5C3'
+}));
+const tableView = Ti.UI.createTableView({
+	data: [ settingsSection, logoutSection ],
+	top: 40,
+	height: Ti.UI.SIZE,
+	scrollable: false,
+	rowSeparatorInsets: { left: 20, right: 20 },
+	separatorStyle: Ti.UI.TABLE_VIEW_SEPARATOR_STYLE_NONE
```
jquick-axway, line 64: You're setting the `rowSeparatorInsets` property to shorten the line width, but you're also removing the line by setting `separatorStyle` to NONE.

templates/app/webpack-default/template/tiapp.xml
```diff
@@ -0,0 +1,55 @@
...
+	<url>not specified</url>
+	<description></description>
+	<copyright>not specified</copyright>
+	<icon>appicon.png</icon>
+	<fullscreen>false</fullscreen>
+	<navbar-hidden>false</navbar-hidden>
+	<analytics>true</analytics>
+	<guid></guid>
+	<property name="ti.ui.defaultunit">dp</property>
+	<property name="run-on-main-thread" type="bool">true</property>
```
jquick-axway, line 16: Let's remove the "run-on-main-thread" property.
It's no longer supported and is currently ignored.

templates/app/webpack-default/template/src/home.js
ewanharris: Just another +1 on "wtf does that syntax mean". I like Josh's suggestion, it makes it a lot clearer to me

janvennemann: The dictionary seems like a great idea. I would prefer `{ start: 6, length: 5 }` though ;)

templates/app/webpack-default/template/src/profile.js
janvennemann: Would it make sense to have a default value of `vertical` for that? Most of the time when i use scroll views i want them to be vertical anyway.

templates/app/webpack-default/template/src/profile.js
janvennemann: Ah good catch. I probably missed that since i initially had separators but then removed them since i liked it better without.

templates/app/webpack-default/template/.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+	"recommendations": [
+		"axway.vscode-titanium"
```
ewanharris, line 3: Love it

templates/app/webpack-default/template/src/profile.js
jquick-axway: It already defaults to "vertical" now, but it logs a warning stating that you should explicitly set it. This warning comes from our code and it's been this way for years.

ewanharris (approved):
No comments from me, this is awesome @janvennemann huge kudos for going above and beyond here

templates/app/webpack-default/template/.vscode/extensions.json
jquick-axway: I didn't know we can do this. We should do the same in all of our templates.

jquick-axway: @janvennemann, @ewanharris, maybe we should add a `/.vscode/settings.json` file which tells the editor to ignore our generated directories. What do you think?
```json
{
	"files.exclude": {
		"build/": true,
		"i18n/": true,
		"platform/": true,
		"Resources/": true
	},
	"search.exclude": {
		"build/": true,
		"i18n/": true,
		"platform/": true,
		"Resources/": true
	},
	"files.watcherExclude": {
		"build/": true,
		"i18n/": true,
		"platform/": true,
		"Resources/": true
	}
}
```

jquick-axway: I wrote up a ticket to add these vscode files to our other app templates here...
https://jira.appcelerator.org/browse/TIMOB-28070

lokeshchdhry:
FR Passed.

Webpack template for classic apps looks good.

This being the base for alloy & angular template, alloy template works as expected. Apps with this template build & run fine on Android & IOS on MAC & Windows.

Angular template will be tested in this PR https://github.com/appcelerator/titanium_mobile/pull/11852

PR #11859, opened 2020-08-01
feat: add OS version major/minor integer constants

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28061

**Summary:**
Added new integer constants...
- `Ti.Platform.versionMajor`
- `Ti.Platform.versionMinor`
- `OS_VERSION_MAJOR`
- `OS_VERSION_MINOR`

**Reason:**
Makes it easier to implement new OS features. For example...
```javascript
if (OS_IOS && (OS_VERSION_MAJOR >= 13)) {
	// Do something on iOS 13 or higher only.
}
```

---
**Test:**
_Run this test on different Android/iOS versions._
1. Build and run the below on Android or iOS.
2. Verify `versionMajor` and `versionMinor` match `Ti.Platform.version` string.
3. Verify `OS_VERSION_MAJOR` and `OS_VERSION_MINOR` match `Ti.Platform.version` string.

```javascript
window.add(Ti.UI.createView({ height: OS_IOS ? "20dp" : "10dp" }));
window.add(Ti.UI.createLabel({
	text: "Ti.Platform.version: " + Ti.Platform.version,
	top: "10dp",
}));
window.add(Ti.UI.createLabel({
	text: "Ti.Platform.versionMajor: " + Ti.Platform.versionMajor,
	top: "10dp",
}));
window.add(Ti.UI.createLabel({
	text: "Ti.Platform.versionMinor: " + Ti.Platform.versionMinor,
	top: "10dp",
}));
window.add(Ti.UI.createLabel({
	text: "OS_VERSION_MAJOR: " + OS_VERSION_MAJOR,
	top: "10dp",
}));
window.add(Ti.UI.createLabel({
	text: "OS_VERSION_MINOR: " + OS_VERSION_MINOR,
	top: "10dp",
}));
window.open();
```

### DIFF
--- a/apidoc/Global/Global.yml
+++ b/apidoc/Global/Global.yml
@@ -323,6 +323,8 @@ properties:
   - name: OS_VERSION_MAJOR
     summary: The operation system's major version number.
     description: |
+        This returns the same value as the <Titanium.Platform.versionMajor> property.
+
         ``` js
         if (OS_IOS && (OS_VERSION_MAJOR >= 13)) {
             // Do something on iOS 13 or higher only.
@@ -334,8 +336,12 @@ properties:
 
   - name: OS_VERSION_MINOR
     summary: The operating system's minor version number.
-    description: Will return zero if OS does not have a minor version.
+    description: |
+        This returns the same value as the <Titanium.Platform.versionMinor> property.
+
+        Will return zero if the OS does not have a minor version.
     type: Number
+    default: 0
     permission: read-only
     since: "9.2.0"
 
--- a/apidoc/Global/Global.yml
+++ b/apidoc/Global/Global.yml
@@ -320,6 +320,25 @@ properties:
     permission: read-only
     since: "9.0.0"
 
+  - name: OS_VERSION_MAJOR
+    summary: The operation system's major version number.
+    description: |
+        ``` js
+        if (OS_IOS && (OS_VERSION_MAJOR >= 13)) {
+            // Do something on iOS 13 or higher only.
+        }
+        ```
+    type: Number
+    permission: read-only
+    since: "9.2.0"
+
+  - name: OS_VERSION_MINOR
+    summary: The operating system's minor version number.
+    description: Will return zero if OS does not have a minor version.
+    type: Number
+    permission: read-only
+    since: "9.2.0"
+
   - name: ENV_DEV
     summary: Alias for <ENV_DEVELOPMENT>
     type: Boolean
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -363,14 +363,20 @@ properties:
 
   - name: versionMajor
     summary: The operating system's major version number.
+    description: |
+      This returns the same value as the [OS_VERSION_MAJOR](Global.OS_VERSION_MAJOR) constant.
     type: Number
     permission: read-only
     since: "9.2.0"
 
   - name: versionMinor
     summary: The operating system's minor version number.
-    description: Will return zero if OS does not have a minor version.
+    description: |
+      This returns the same value as the [OS_VERSION_MINOR](Global.OS_VERSION_MINOR) constant.
+
+      Will return zero if the OS does not have a minor version.
     type: Number
+    default: 0
     permission: read-only
     since: "9.2.0"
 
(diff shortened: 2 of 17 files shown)


apidoc/Titanium/Platform/Platform.yml
```diff
@@ -357,10 +357,23 @@ properties:
...
+    type: Number
+    permission: read-only
+    since: "9.2.0"
+
+  - name: versionMinor
+    summary: The operating system's minor version number.
+    description: Will return zero if OS does not have a minor version.
+    type: Number
+    permission: read-only
+    since: "9.2.0"
```
sgtcoolguy, line 375:
``` yml
default: `0`
```

Note that we haven't really tightened the definition for what is a valid "default" value. So it's really just for doc consumption on the site - but ideally we'd be able to more tightly specify this in some way that we could auto-generate tests or something from it.

iphone/Resources/app.js
```diff
@@ -20,3 +20,6 @@ btn.addEventListener('click', function() {
 
 win.add(btn);
 win.open();
+
+console.log("@@@ OS_VERSION_MAJOR: " + OS_VERSION_MAJOR);
```
sgtcoolguy, line 24: Need to revert these changes

apidoc/Global/Global.yml
```diff
@@ -320,6 +320,25 @@ properties:
...
+            // Do something on iOS 13 or higher only.
+        }
+        ```
+    type: Number
+    permission: read-only
+    since: "9.2.0"
+
+  - name: OS_VERSION_MINOR
+    summary: The operating system's minor version number.
+    description: Will return zero if OS does not have a minor version.
```
sgtcoolguy, line 337: link/refer to `Ti.Platform.versionMinor`?

apidoc/Global/Global.yml
```diff
@@ -320,6 +320,25 @@ properties:
     permission: read-only
     since: "9.0.0"
 
+  - name: OS_VERSION_MAJOR
+    summary: The operation system's major version number.
```
sgtcoolguy, line 324: link/refer to `Ti.Platform.versionMajor`?

sgtcoolguy (changes requested):
This is awesome! It really shows how we're making it easier to add features cross-platform more easily too - the global defines in JS, the simplified JSC API on iOS, the already simple way to declare the property on Android.

My feedback is all nitpicks and none of it is required **except** the rollback of the change to the iOS app.js used by Xcode.

jquick-axway:
Updated PR:
- Updated API docs.
- Updated unit tests to verify `versionMajor`/`versionMinor` match components in `version` string.
- Removed test log message from iOS test app. _(My bad.)_

apidoc/Titanium/Platform/Platform.yml
jquick-axway: Good idea. I'll do the same for `OS_VERSION_MINOR`.

jquick-axway:
@sgtcoolguy, what do you think of us adding our `OS_ANDROID` and `OS_IOS` constants to the new "globals.js" file?
```javascript
Object.defineProperties(global, {
	'OS_ANDROID': { value: OS_ANDROID, writable: false },
	'OS_IOS': { value: OS_IOS, writable: false },
	OS_VERSION_MAJOR: { value: Ti.Platform.versionMajor, writable: false },
	OS_VERSION_MINOR: { value: Ti.Platform.versionMinor, writable: false }
});
```

The reason to do this is because these constants are **undefined** if we're **NOT** running babel on the app project. This can happen:
- In our `titanium_mobile/android/app` test project under Android Studio.
- In our `titanium_mobile/iphone/Resources` test project under Xcode.
- In a Titanium app project if "tiapp.xml" has `<transpile>false</transplie>` set.

sgtcoolguy:
> @sgtcoolguy, what do you think of us adding our `OS_ANDROID` and `OS_IOS` constants to the new "globals.js" file?
> 
> ```js
> Object.defineProperties(global, {
> 	'OS_ANDROID': { value: OS_ANDROID, writable: false },
> 	'OS_IOS': { value: OS_IOS, writable: false },
> 	OS_VERSION_MAJOR: { value: Ti.Platform.versionMajor, writable: false },
> 	OS_VERSION_MINOR: { value: Ti.Platform.versionMinor, writable: false }
> });
> ```

Yeah, I'm totally fine with that. Makes sense to me.

sgtcoolguy (approved):

sgtcoolguy:
This feels very much like a PR that we should be able to merge without a formal QE FR, since we have unit tests to cover usage.

jquick-axway:
@sgtcoolguy , I agree.

PR #11866, opened 2020-08-04
chore(android)(9_3_X): use latest emulator and ndk

garymathews:
- Use latest Android 11.0 (30) emulator
- Build using latest `r21d` NDK
- Fix running of `npm install` on Windows, when launching test suite
- Fix running of `adb` commands on Windows
- Fix device identifier used for ADB commands
- Sets auto-rotation if disabled, when launching test suite


garymathews:
@sgtcoolguy do we need to allow rotation on the emulator?

sgtcoolguy:
> @sgtcoolguy do we need to allow rotation on the emulator?

I don't think so? If so that'd be a new thing...

tests/Resources/ti.app.test.js
```diff
@@ -320,6 +320,13 @@ describe('Titanium.App', function () {
 
 	it.android('pause/resume events', function (finish) {
 		this.timeout(5000);
+
+		// Cannot resume activity on Android 10.0+
```
sgtcoolguy, line 324: Seems like something we should add in our docs at the very least. Are we sure this is a hard OS limitation, or is there a fix we need?

tests/plugins/mocha.test.support/hooks/hook.js
```diff
@@ -78,3 +79,28 @@ async function wakeDevices(logger, builder) {
...
+
+	async function autoRotate(device) {
+		const deviceId = device !== 'emulator' ? [ '-s', device ] : [];
+
+		// Enable auto-rotation if currently disabled.
+		const rotationStatus = await adb([ ...deviceId, 'shell', 'settings', 'get', 'system', 'accelerometer_rotation' ]);
+		if (rotationStatus && rotationStatus === '0') {
+			logger.info(`Enabling auto-rotation for ${device}`);
+
+			await adb([ ...deviceId, 'shell', 'settings', 'set', 'system', 'accelerometer_rotation', '1' ]);
```
sgtcoolguy, line 93:
```suggestion
			await adb([ ...deviceId, 'shell', 'settings', 'put', 'system', 'accelerometer_rotation', '1' ]);
```

I tried this on the actual box and did find the setting was `'0'` and then tried this command and it said:
`Invalid command: set`
After some googling, I realized it should be `put`, which worked manually.

sgtcoolguy:
FWIW, I just set up to run this locally and this single test case is also failing locally on an emulator for me.

tests/Resources/ti.app.test.js
jquick-axway: Yes, this is a well known limitation Google introduced as of Android 10. The idea is you don't want some rogue backgrounded app to display a window without user interaction. Pretty much the equivalent of browser popups. iOS has never supported this and I'm not aware of customer apps that do this. (It's considered extremely naughty.)
https://developer.android.com/guide/components/activities/background-starts

Perhaps a test like this is best done by QE's UI integration suite, where I would hope we can run the same test for iOS.

Our "ti.android.test.js" script's "activity callback" unit test will fail on Android 10 and above for the same reason. Although, we should be able to keep that unit test by having it open/close a child window instead of navigating to the home screen.
https://github.com/appcelerator/titanium_mobile/blob/b21c5d79d00e6efad06fcea441712f57c3bbaa01/tests/Resources/ti.android.test.js#L65

garymathews:
@sgtcoolguy I've updated this PR to address a number of issues when running `adb` commands and also spawning processes on Windows

Jenkinsfile
```diff
@@ -95,7 +95,7 @@ def androidUnitTests(nodeVersion, npmVersion, testOnDevices) {
 								sh label: 'Run Test Suite on device(s)', script: "npm run test:integration -- android -T device -C all"
 							// run PR tests on emulator
 							} else {
-								sh label: 'Run Test Suite on emulator', script: "npm run test:integration -- android -T emulator -D test -C android-28-playstore-x86"
+								sh label: 'Run Test Suite on emulator', script: "npm run test:integration -- android -T emulator -D test -C android-30-playstore-x86"
```
jquick-axway, line 98: Can we use an x86_64 emulator image instead?

garymathews: We only install `x86` on the nodes atm

garymathews:
Looks like `EmulatorManager.isRunning()` isn't firing the callback on the Jenkins node, even though it works fine locally

https://github.com/appcelerator/node-titanium-sdk/blob/master/lib/emulator.js#L180

jquick-axway:
@garymathews, our "ti.ui.window.test.js" script's "orientationModes" unit test fails when I run it in the Android 11 emulator. It's because the OS initially opens the activity/window at the device's current orientation and then does a transition animation to the expected orientation (so 2 layouts are occurring). Unfortunately, there is nothing we can do about this. I've seen this behavior before years ago on 3rd party Android devices. And you can't disable this animation either.

I resolved it by adding a `setTimeout()`. I also re-enabled the `LANDSCAPE_RIGHT` test because it was working fine for me after adding this timer. (It might have been failing before for the same reason?)
```javascript
describe.android('.orientationModes', () => {
	this.slow(5000);
	this.timeout(20000);

	function doOrientationModeTest(orientation, finish) {
		win = Ti.UI.createWindow({
			orientationModes: [ orientation ]
		});
		win.addEventListener('open', () => {
			setTimeout(() => {
				try {
					win.orientationModes.should.have.length(1);
					win.orientationModes[0].should.eql(orientation);
					win.orientation.should.eql(orientation);
				} catch (e) {
					return finish(e);
				}
				finish();
			}, 500);
		});
		win.open();
	}

	it('PORTRAIT', finish => {
		doOrientationModeTest(Ti.UI.PORTRAIT, finish);
	});

	it('LANDSCAPE_LEFT', finish => {
		doOrientationModeTest(Ti.UI.LANDSCAPE_LEFT, finish);
	});

	it('LANDSCAPE_RIGHT', finish => {
		doOrientationModeTest(Ti.UI.LANDSCAPE_RIGHT, finish);
	});
});
```

Jenkinsfile
sgtcoolguy: I can certainly modify our chef scripts to set up x86_64 emulators. I'll see if I can't do that now...

tests/plugins/mocha.test.support/hooks/hook.js
```diff
@@ -32,24 +33,66 @@ exports.init = (logger, config, cli) => {
...
 }
 
-async function wakeDevices(logger, builder) {
+async function wakeDevices(logger, builder, emulatorManager) {
 
 	async function wake(device) {
 		logger.info(`Waking up ${device}`);
 
-		const deviceId = device !== 'emulator' ? [ '-s', device ] : [];
+		const deviceId = device !== 'emulator' ? [ '-s', await getDeviceId(device, emulatorManager) ] : [];
```
sgtcoolguy, line 95: This change appears to have broken the part of the build where it wakes the emulator. It's just sitting there at this point until the build times out.

tests/plugins/mocha.test.support/hooks/hook.js
```diff
@@ -32,24 +33,66 @@ exports.init = (logger, config, cli) => {
...
 		}
 		done();
 	});
 };
 
-async function adb(argumentArray) {
-	return new Promise(resolve => appc.subprocess.run(ADB_PATH, argumentArray, { shell: false, windowsHide: true }, (error, stdout, _stderr) => {
-		resolve(stdout);
-	}));
+async function adb(args) {
```
sgtcoolguy, line 47: I don't which exact change broke the build, but the CI machine is waiting after spitting out that it's going to wake the emulator and then not proceeding and the build eventually times out. I suspect perhaps the changes to running `adb` might be involved?

tests/plugins/mocha.test.support/hooks/hook.js
sgtcoolguy, line 95: Note that on the CI box it's spitting out `"Waking up android-30-playstore-x86"` - so presumably at this point `device === 'android-30-playstore-x86'`, which means it's actually going to trigger the code to `await getDeviceId(device, emulatorManager)` - which doesn't seem right to me. 

Looking at https://github.com/appcelerator/titanium_mobile/blob/master/android/cli/hooks/run.js#L85-L94

I think if `builder.target === 'emulator'` then `builder.deviceId` can be trusted?
So should this condition be testing if `builder.target === 'device'`? I'm not sure what you were trying to fix here by trying to get the deviceId in another way.

garymathews: You're right it should have been comparing `builder.target`, fixed that now.

The issue I was trying to fix is that it would use the incorrect device identifier for the emulator, which is an issue when multiple emulators are open.
```
[INFO]  Waking up Android_R_11.0_30_
[WARN]  Could not wake Android_R_11.0_30_:
adb.exe: more than one device/emulator
```
In this case, it would attempt to use the AVD identifier `Android_R_11.0_30_` instead of the device identifier `emulator-5554` to execute adb commands.

I guess this doesn't really matter on our test nodes, unless we want to test on multiple OS versions in the future. So I'll remove the handling of that for now.

jquick-axway:
@sgtcoolguy , should we merge this PR now?

PR #11867, opened 2020-08-05
feat(ios): allow multiple photo selection

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27984


iphone/Classes/MediaModule.m
```diff
@@ -1105,8 +1106,20 @@ - (void)openPhotoGallery:(id)args
 {
   ENSURE_SINGLE_ARG_OR_NIL(args, NSDictionary);
   ENSURE_UI_THREAD(openPhotoGallery, args);
-  [self showPicker:args isCamera:NO];
+
+  NSArray *types = (NSArray *)[args objectForKey:@"mediaTypes"];
+#if IS_SDK_IOS_14
+  if ([TiUtils isIOSVersionOrGreater:@"14.0"] && ([args objectForKey:@"selectionLimit"] || [TiUtils boolValue:[args objectForKey:@"allowMultiple"] def:NO])) {
```
hansemannn, line 1112: Can we maybe always use it on iOS 14? The old image picker is now deprecated and should be replaced for iOS 14+.

iphone/Classes/MediaModule.m
```diff
@@ -1755,6 +1775,150 @@ - (void)handleTrimmedVideo:(NSURL *)theURL withDictionary:(NSDictionary *)dictio
...
+                                       if ([media mimeType] == nil) {
+                                         [media setMimeType:@"video/mpeg" type:TiBlobTypeFile];
+                                       }
+                                       [videoArray addObject:@{ @"media" : media, @"mediaType" : (NSString *)kUTTypeMovie, @"success" : @(YES), @"code" : @(0) }];
+                                     } else {
+                                       [videoArray addObject:@{ @"error" : error.description, @"code" : @(error.code), @"success" : @(NO), @"mediaType" : (NSString *)kUTTypeMovie }];
+                                       DebugLog(@"[ERROR] Failed to load video- %@ .", error.description);
+                                     }
+                                     dispatch_group_leave(group);
+                                   }];
```
hansemannn, line 1894: There seem to be a lot of duplicate code logic here. Maybe move to own handler class / methods to clean up the business logic around this.

iphone/Classes/MediaModule.m
```diff
@@ -1755,6 +1775,150 @@ - (void)handleTrimmedVideo:(NSURL *)theURL withDictionary:(NSDictionary *)dictio
...
+    if (videoArray != nil) {
+      [dictionary setObject:videoArray forKey:@"videos"];
+    }
+    [self sendPickerSuccess:dictionary];
+  });
+
+  if (autoHidePicker) {
+    [self closeModalPicker:picker];
+  }
+}
```
hansemannn, line 1919: You are on a background thread here. Maybe validate if `closeModalPicker` and `sendPickerSuccess` really are performed on the main thread. In any case, there doesn't seem to be much processing here, so is a global queue / background queue necessary?

sgtcoolguy:
iOS formatting is failing lint. Run 'npm run format:ios'

iphone/Classes/MediaModule.m
vijaysingh-axway: In new PHPickerViewController, there is no option for image editing(may be included in future releases). I think this is important feature and we can not remove it in minor release.

iphone/Classes/MediaModule.m
vijaysingh-axway: There is no special handling in this delegate method for `closeModalPicker` and  `sendPickerSuccess`. It is same as with old picker. In `dispatch_group_notify` queue is mandatory to pass.

vijaysingh-axway:
@jquick-axway Can you verify doc for android multiple selection callback?

apidoc/Titanium/Media/Media.yml
```diff
@@ -2115,6 +2124,39 @@ properties:
     osver: {ios: {min: "11.0"}}
 
 ---
+name: CameraMediaMultipleItemsType
+summary: A media object from photo gallery when <Titanium.Media.allowMultiple> is `true` or in iOS <Titanium.Media.selectionLimit> is set.
+extends: SuccessResponse
+properties:
+  - name: success
```
sgtcoolguy, line 2131: shouldn't have to document the properties already on the parent `SuccessResponse` type: success, error, code

iphone/Classes/MediaModule.h
```diff
@@ -9,6 +9,11 @@
 #if defined(USE_TI_MEDIAGETAPPMUSICPLAYER) || defined(USE_TI_MEDIAOPENMUSICLIBRARY) || defined(USE_TI_MEDIAAPPMUSICPLAYER) || defined(USE_TI_MEDIAGETSYSTEMMUSICPLAYER) || defined(USE_TI_MEDIASYSTEMMUSICPLAYER) || defined(USE_TI_MEDIAHASMUSICLIBRARYPERMISSIONS)
 #import <MediaPlayer/MediaPlayer.h>
 #endif
+#if IS_SDK_IOS_14
```
sgtcoolguy, line 12: combine the nested if/endif?
```suggestion
#if IS_SDK_IOS_14 && defined(USE_TI_MEDIAOPENPHOTOGALLERY)
#import <PhotosUI/PHPicker.h>
#endif
```

iphone/Classes/MediaModule.h
```diff
@@ -17,12 +22,16 @@
 #import <TitaniumKit/TiViewProxy.h>
 
 @class AVAudioRecorder;
-
 @interface MediaModule : TiModule <
                              UINavigationControllerDelegate,
 #if defined(USE_TI_MEDIASHOWCAMERA) || defined(USE_TI_MEDIAOPENPHOTOGALLERY) || defined(USE_TI_MEDIASTARTVIDEOEDITING)
                              UIImagePickerControllerDelegate,
 #endif
+#if IS_SDK_IOS_14
```
sgtcoolguy, line 30: combine, same as above

iphone/Classes/MediaModule.h
```diff
@@ -36,6 +45,11 @@
 // Camera picker
 #if defined(USE_TI_MEDIASHOWCAMERA) || defined(USE_TI_MEDIAOPENPHOTOGALLERY) || defined(USE_TI_MEDIASTARTVIDEOEDITING)
   UIImagePickerController *picker;
+#endif
+#if IS_SDK_IOS_14
```
sgtcoolguy, line 49: combine if/endif

iphone/Classes/MediaModule.m
```diff
@@ -1376,6 +1389,14 @@ - (void)destroyPicker
   }
   RELEASE_TO_NIL(picker);
 #endif
+
+#if IS_SDK_IOS_14
+#if defined(USE_TI_MEDIAOPENPHOTOGALLERY)
```
sgtcoolguy, line 1394: combine if/endif

apidoc/Titanium/Media/Media.yml
```diff
@@ -2104,8 +2106,15 @@ properties:
...
-    since: "6.0.0"
+    platforms: [android, iphone, ipad]
+    osver: {ios: {min: "14.0"}}
+    since: { android: "6.0.0", iphone: "9.2.0", ipad: "9.2.0" }
+  - name: selectionLimit
+    summary: Specifies number of assets that can be selected.
+    type: Boolean
+    platforms: [iphone, ipad]
+    osver: {ios: {min: "14.0"}}
+    since: "9.2.0"
```
jquick-axway, line 2117: I think "selectionLimit" should be ignored unless "allowMultiple" is set `true`. Especially since Android does not support this property (not even natively). Requiring the app developer to set "allowMultiple" to `true` would then make the code portable.

We should also document that setting this to zero allows you to select an unlimited number of media.

iphone/Classes/MediaModule.m
jquick-axway, line 1112: I think the `selectionLimit` property should be **ignored** unless `allowMultiple` is set `true`. The reason is because `selectionLimit` is not supported on Android... so you can't just set that property alone. It wouldn't be portable.

So, I think the "if" condition should be this...
```objc
if ([TiUtils isIOSVersionOrGreater:@"14.0"] && [TiUtils boolValue:[args objectForKey:@"allowMultiple"] def:NO]) {
   // Show multiple selection window.
} else {
   // Show single selection window.
}
```

apidoc/Titanium/Media/Media.yml
```diff
@@ -2054,7 +2054,9 @@ summary: |
 properties:
   - name: success
     summary: Function to call when the photo gallery is closed after a successful selection.
-    type: Callback<CameraMediaItemType>
+    description: |
+        If <Titanium.Media.allowMultiple> is `true` or <Titanium.Media.selectionLimit> is set, Callback<CameraMediaMultipleItemsType> is called otherwise Callback<CameraMediaItemType>.
```
jquick-axway, line 2058: This sentence is a bit awkward. Let's change it to this...
```
If <Titanium.Media.allowMultiple> is `true`, then Callback<CameraMediaMultipleItemsType> will be invoked. Otherwise Callback<CameraMediaItemType> will be invoked for single selection.
```

iphone/Classes/MediaModule.m
vijaysingh-axway: Agree! Updated.

apidoc/Titanium/Media/Media.yml
vijaysingh-axway: Yes. Updated "selectionLimit" to work when "allowMultiple" is `true` only.

iphone/Classes/MediaModule.m
```diff
@@ -1109,7 +1109,7 @@ - (void)openPhotoGallery:(id)args
 
   NSArray *types = (NSArray *)[args objectForKey:@"mediaTypes"];
 #if IS_SDK_IOS_14
-  if ([TiUtils isIOSVersionOrGreater:@"14.0"] && ([args objectForKey:@"selectionLimit"] || [TiUtils boolValue:[args objectForKey:@"allowMultiple"] def:NO])) {
+  if ([TiUtils isIOSVersionOrGreater:@"14.0"] && [TiUtils boolValue:[args objectForKey:@"allowMultiple"] def:NO]) {
```
hansemannn, line 1112: Any reason why not to use the PHPicker by default on iOS 14 all the time? I'd love to use it from now on - the UX is amazing, also for single selections!

vijaysingh-axway: 1. In new PHPickerViewController, there is no option for image editing. I think this is important feature and we can not remove it in minor release. (I'll create a ticket to deprecate editing property, in case it is not supported by Apple in future releases).
2. In PHPickerViewController, video file need to copy locally as per discussion https://developer.apple.com/forums/thread/652695. I think this is not the best way. May be its handling can be changed by Apple. 
So I have limited this API to multiple selection only. It will be updated in future release.

In case you want to use the API with single selection, set  `Ti.Media.allowMultiple = true` and Ti.Media.selectionLimit = 1. But yes, a minor change required in your callback.

hansemannn: Ohh, thats sad! Didn't know they missed parity when updating the API. Alrighty, thanks for the clarification Vijay! 🙂

ssekhri:
FR Passed
On iOS 14 - Able to select multiple photo, live photo or video or a combination of these media types upto the defined selection limit.
On iOS 13.x - The selection of single item of a media type works fine as before.

Verified on:
Mac OS: 10.15.4
SDK: 9.2.0.v20200910075418
Appc CLI: 8.1.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0 Beta6
iPhone 7Plus(v14.0 Beta6), iPhone X(v13.6.1)

PR #11870, opened 2020-08-05
feat: add vscode config files to app templates

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28070

**Summary:**
- Adds config files under the `.vscode/` folder.
- Hides/excludes generated folder from editor such as the `build` folder.
- Recommends our Titanium extension when opened in VS Code.

**Note:**
We need to update our Alloy repo's app template to get it to ignore the project's root `i18n`, `platform`, and `Resources` folders. Titanium's app template has to assume it's for a Classic app project.


ewanharris (approved):
I like this! I'll file a separate ticket to have alloy update apps with its own settings.json file

templates/app/angular-default/template/.vscode/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+	"files.exclude": {
+		"build/": true,
+		"i18n/": true,
+		"platform/": true,
+		"Resources/": true
```
janvennemann, line 6: These changed with the latest Angular template. I don't touch `i18n` and `platform` anymore so they basically work like in a classic app.

```suggestion
		"build/": true,
		"Resources/": true
```

janvennemann (commented):
Left a note on the Angular settings. If you don't mind i would take the settings from this PR and directly apply them to #11852 since that PR is still open.

Also, we should definitely document this somewhere. I worked with a sample app recently that had this and i was confused at first why the hell `build` and `Resources` were missing until i remembered the VSCode settings. Might be confusing for users that are not aware of this and they create a new app using the updated templates and suddenly folders are missing/hidden.

jquick-axway:
@janvennemann , so are you suggesting we should close this PR and you'll update all of the app templates in your PR? I'm okay with that.

Regarding hiding directories in VS Code, a common mistake I've seen people make in Alloy is to modify the root `i18n` and `platform` folders... and be confused as to why their changes didn't work. By hiding them, it would hopefully give them the clue to make changes under the `app` folder instead.

Also, we think file locking issues might occur on Windows under the generated folders if you leave VS Code's "indexing" turned on. Might not be true anymore, but I've seen a few complaints in the past where builds would fail on Windows while using VS Code.

janvennemann:
> @janvennemann , so are you suggesting we should close this PR and you'll update all of the app templates in your PR? I'm okay with that.

We can keep this PR for the changes to the default template. But i would like like to make the changes to the Angular template in #11852 since the whole Angular template received a complete overhaul there anyway.

> Regarding hiding directories in VS Code, a common mistake I've seen people make in Alloy is to modify the root `i18n` and `platform` folders... and be confused as to why their changes didn't work. By hiding them, it would hopefully give them the clue to make changes under the `app` folder instead.

Yeah, that should make things clearer. 

> Also, we think file locking issues might occur on Windows under the generated folders if you leave VS Code's "indexing" turned on. Might not be true anymore, but I've seen a few complaints in the past where builds would fail on Windows while using VS Code.

In any case, it should be definitely good performance wise to disable indexing in our generated folders. Every now and then i see VSCode go nuts on CPU usage during our builds. Maybe that's related.

jquick-axway:
Maybe we should hide and disable indexing on the `node_modules` directory too. That's supposed to be npm managed after all... and it could be an indexing performance nightmare.

jquick-axway:
@janvennemann, I've updated the Angular app template's `.vscode/settings.json` to exclude `node_modules` folder and `package-lock.json` like we did for our other app templates.

janvennemann:
@jquick-axway `node_modules` is excluded by default from search and files watcher so IMHO we don't need to specifically set that in our settings. Also, i'm not fully convinced that hiding `package-lock.json` and `node_modules` is a good idea. Especially `package-lock.json`, since that file is frequently the source of merge conflicts and needs to be edited.

jquick-axway:
Since `node_modules` and `package-lock.json` is generated by `npm`, I think it makes sense to exclude them. People won't normally tinker with these files directly unless something has gone horribly wrong. And I don't think it hurts to explicitly exclude `node_modules` from search/watcher. We definitely do have to add `node_modules` to the vscode `files.exclude` to hide it from the Explorer panel.

The way I see it, if a small number of developers want to see the excluded folders/files, then they can change the `settings.json` after app creation. Let's not over-think this.

janvennemann:
I'm fine with explicitly excluding `node_modules` and also hiding it from explorer since that folder is generally excluded from version control. Just wanted to mention that it's already part of VSCodes' defaults.

package-lock.json` on the other hand is usually not excluded from version control, and IMHO should not be hidden since it can contain changes that require the developer's attention. And hiding it makes this interaction unnecessarily complicated.

ssjsamir (approved):
**FR Passed** 
+ Able to see the recommendation to install the Titanium plugin
Screenshot 2020-08-19 at 15 29 43

+ Able to see the following files when adding a titanium application into Vscode. 
**extensions.json**
```
{
	"recommendations": [
		"axway.vscode-titanium"
	]
}
```
**settings.json**
```
{
	"files.exclude": {
		"build/": true,
		"node_modules/": true,
		"package-lock.json": true
	},
	"search.exclude": {
		"build/": true,
		"node_modules/": true,
		"package-lock.json": true
	},
	"files.watcherExclude": {
		"build/": true,
		"node_modules/": true,
		"package-lock.json": true
	}
}
```
+ Able to see the build folder not being shown when running the app
+ Can see the build folder once settings _settings.json_ have been removed

**Test Environment**
```
MacOS Big Sur: 11.0 Betam 4
Xcode: 12.0 Beta 5
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
iphone 8 Sim (14.0 Beta)
VS code 1.48.0
Titanium plugin 0.9.0
```

PR #11872, opened 2020-08-06
Test suite fix

sgtcoolguy:
**Description:**
Sometimes when test failed on actual android devices on master, 9_1_X and 9_3_X, the test suite never though it had finished. It was getting into a bad state when a test result JSON output went over a single line *and* the output had both log level *and* a device name prefix. I've cleaned the code up a bit, grabbed a build log from Jenkins for an occurrence, verified the bug occurred and then fixed it (and added a unit test for our test handling code!).

Now if a device goes funky and fails tests in some way that produces large output, we should get the test results properly rather than get confused, hit the build timeout and mark the build failed.

**NOTE**: This has a commit to force running on devices to test the PR. This needs to be reverted/removed before merge!


sgtcoolguy:
merged manually

garymathews (approved):
Other than the minor JSDoc warnings this LGTM!

PR #11897, opened 2020-08-12
feat(iOS): support new APIs timeoutForResource and waitsForConnectivity

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28078

I don't think we can add unit test for property `waitsForConnectivity` as it requires user interaction. E.g. To test this, off wifi first and the on it, is required. Which is not possible with public APIs.


apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -536,6 +564,26 @@ properties:
         On iOS, the default is 15000 (15 seconds).
     type: Number
 
+  - name: timeoutForResource
```
sgtcoolguy, line 567: add `since: '9.2.0'`

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -536,6 +564,26 @@ properties:
...
+    summary: |
+        The maximum amount of time (in milliseconds) that a resource request should be allowed to take.
+    default: |
+        The default is 7*24*60*60*1000 milliseconds (7 days).
+    type: Number
+    platforms: [iphone, ipad]
+
+  - name: waitsForConnectivity
```
sgtcoolguy, line 575: add `since: '9.2.0'`

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -138,6 +138,34 @@ description: |
     <Titanium.Network.getHTTPCookies>, <Titanium.Network.getHTTPCookiesForDomain>, <Titanium.Network.getSystemCookies>,
     <Titanium.Network.removeHTTPCookie>, <Titanium.Network.removeHTTPCookiesForDomain>, <Titanium.Network.removeAllHTTPCookies>,
     <Titanium.Network.removeSystemCookie>, <Titanium.Network.removeAllSystemCookies>.
+
+    #### Connect to local network in iOS
+
+    On iOS 14 and later, while connecting to local network a prrompt will be shown to request user's permission.
```
sgtcoolguy, line 144: typo:
```suggestion
    On iOS 14 and later, while connecting to local network a prompt will be shown to request user's permission.
```

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -138,6 +138,34 @@ description: |
...
+    Add key `NSLocalNetworkUsageDescription` to the `ios plist` section of the tiapp.xml file to show the message on prompt.
+    If this key is not added deafult message will be shown in prompt.
```
sgtcoolguy, line 146: typo:
```suggestion
    If this key is not added default message will be shown in prompt.
```

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -536,6 +564,26 @@ properties:
         On iOS, the default is 15000 (15 seconds).
     type: Number
 
+  - name: timeoutForResource
+    summary: |
```
sgtcoolguy, line 568: Link to the apple docs for the underlying property we're surfacing? https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408153-timeoutintervalforresource?language=objc

iphone/TitaniumKit/TitaniumKit/Libraries/APSHTTPClient/APSHTTPRequest.h
```diff
@@ -61,14 +62,14 @@ typedef NS_ENUM(NSInteger, APSRequestError) {
...
  @discussion An optional NSOperationQueue for delegate callbacks.
  The default value is nil, which means delegate callbakcs occur on
- the caller's thread if the synchronous property is NO. If the
+ the caller's thread if the synchronous property is NO. If the 
  synchronous property is YES then this property is ignored.
  */
 @property (nonatomic, strong, readwrite) NSOperationQueue *theQueue;
 
 /*!
  @discussion An optioanl array of run loop modes for delegate calllbacks
```
sgtcoolguy, line 71: typo:
```suggestion
 @discussion An optional array of run loop modes for delegate callbacks
```

apidoc/Titanium/Network/HTTPClient.yml
sgtcoolguy: Also does this property only matter if `waitsForConnectivity` is `true`? If so, may want to link to that property in the description here or note it in some way.

sgtcoolguy (changes requested):

apidoc/Titanium/Network/HTTPClient.yml
vijaysingh-axway: 1. Added apple doc link. 
2. This property is not dependent on `waitsForConnectivity`  is `true`. Vice versa is true. If `waitsForConnectivity` is `true` then it matters. It is documented in `waitsForConnectivity`

sgtcoolguy: So, it *does* only matter if `waitsForConnectivity === true`? Please state so in the summary or description.

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -536,6 +564,31 @@ properties:
...
+        The maximum amount of time (in milliseconds) that a resource request should be allowed to take.
+    description: |
+        For more detail see [timeoutintervalforresource](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408153-timeoutintervalforresource?language=objc)
+        from Apple documentation.
+    default: |
+        The default is 7*24*60*60*1000 milliseconds (7 days).
```
sgtcoolguy, line 574:
```suggestion
        The default is `7*24*60*60*1000` milliseconds (7 days).
```

sgtcoolguy (changes requested):

apidoc/Titanium/Network/HTTPClient.yml
vijaysingh-axway: @sgtcoolguy  Sorry for confusion.
1. This property ensures how long to wait for whole resource to transfer before giving up. 
2. If  waitsForConnectivity === true, in that case also this property matters.

Updated doc to explain above.

tests/Resources/ti.network.httpclient.test.js
```diff
@@ -735,4 +735,26 @@ describe('Titanium.Network.HTTPClient', function () {
 		xhr.open('GET', 'https://www.google .com/'); // URL with space
 		xhr.send();
 	});
+
+	it.ios('#timeoutForResource', function (finish) {
+		const xhr = Ti.Network.createHTTPClient({
+			timeout: 6e4,
+			timeoutForResource: 50
```
sgtcoolguy, line 742: 50 ms is very short, and unlikely it'll ever pass, but maybe just in case we make it like 5ms?

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -536,6 +564,34 @@ properties:
...
 
+  - name: timeoutForResource
+    summary: |
+        The maximum amount of time (in milliseconds) that a resource request should be allowed to take.
+    description: |
+        The resource timeout controls how long (in milliseconds) to wait for an entire resource to transfer before giving up.
+        The resource timer starts when the request is initiated and counts until either the request completes or this timeout is reached, whichever comes first.
+        For more detail see [timeoutintervalforresource](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408153-timeoutintervalforresource?language=objc)
+        from Apple documentation.
+        This property also applies if <Ti.Network.HttpClient.waitsForConnectivity> is set to `true`.
```
sgtcoolguy, line 575: I have to say that this property is still fairly confusing to me. I suppose the unit test example helps explain it a little better to me - so maybe we should add an example in the docs?

It looks like a request would timeout if the minimum value of this property *OR* the `timeout` property were reached? (i.e. if `timeout` is 10 seconds and this is 5 seconds, we'd time out after 5 seconds. If this is 10 seconds and `timeout` is 7 seconds, we'd time out after 7 seconds?)

And the big difference here is when used in conjunction with `waitsForConnectivity` where `timeout` would "start" once connectivity/request begins, but `timeoutForResource` would "start" counting once you scheduled the request (i.e. if we lose connectivity and fire off a request, this timer starts immediately while `timeout` wouldn't start until we regained connectivity and actually sent off the request?)

Again, all of this is to say it may be useful to try and do some more explaining in the docs here to help people understand the difference between the properties and how they interact - because I'm just guessing at the above.

sgtcoolguy (approved):
The code changes look fine to me so this is ready for QE test.

I do think we could use some help on the docs here though as the new properties aren't obvious how they are used/interact.

ssekhri:
FR Passed.
Test with both test cases in the ticket comments.
Verified on:
Mac OS: 10.15.4
SDK: 9.2.0.v20200827065243
Appc CLI: 8.1.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0 Beta 6
Device: iPhone 7 Plus(v14.0 Beta6)

PR #11901, opened 2020-08-13
test(android): grab generated images from devices too

sgtcoolguy:
Attempts to pull generated images from android devices (not just sims)


sgtcoolguy:
merged manually after cleanup

PR #11902, opened 2020-08-14
chore(deps): update liveview to 1.5.3

ewanharris:
Note: we should also test the built SDK against a different major node version than it was built with (10.17.0). This is to make sure that FSEvents fully works as we don't recompile native modules on an SDK install (it's built using N-API, so I _think_ it should "just work")


janvennemann (approved):

ssjsamir (commented):
LGTM, When testing liveview and making changes to the `app.ts` and `index.xml` the reload to webpack enabled applications were working as intended. 

*Test Environment*
{code:java}
MacOS Big Sur: 11.0 Beta
Xcode: 12.0 Beta 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
iphone 8 Sim (14.0 Beta)
API29 Pixel XL emulator
{code}

ssjsamir (approved):

PR #11911, opened 2020-08-15
feat(android): no_limits layout

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6589

**Optional Description:**

Make the statusbar background fully transparent but still show it (instead of going fullscreen) 

left window:
```
"Window" : {
	extendSafeArea: true
}
```

right window:
```
"Window" : {
	extendSafeArea: true,
	noLimits: true
}
```

### DIFF
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -1025,11 +1025,11 @@ properties:
   - name: hidesSearchBarWhenScrolling
     summary: A Boolean value indicating whether the integrated search bar is hidden when scrolling any underlying content.
     description: |
-        When the value of this property is true, the search bar is visible only when the scroll position 
-        equals the top of your content view. When the user scrolls down, the search bar collapses into 
-        the navigation bar. Scrolling back to the top reveals the search bar again. When the value of 
-        this property is false, the search bar remains regardless of the current scroll position. 
-        You must set <Titanium.UI.ListView.showSearchBarInNavBar> or <Titanium.UI.TableView.showSearchBarInNavBar> 
+        When the value of this property is true, the search bar is visible only when the scroll position
+        equals the top of your content view. When the user scrolls down, the search bar collapses into
+        the navigation bar. Scrolling back to the top reveals the search bar again. When the value of
+        this property is false, the search bar remains regardless of the current scroll position.
+        You must set <Titanium.UI.ListView.showSearchBarInNavBar> or <Titanium.UI.TableView.showSearchBarInNavBar>
         property for this property to have any effect.
     default: true
     type: Boolean
@@ -1151,6 +1151,17 @@ properties:
     type: Titanium.UI.NavigationWindow
     permission: read-only
     platforms: [iphone, ipad]
+
+  - name: noLimits
+    summary: Adds the FLAG_LAYOUT_NO_LIMITS to the window.
+    description: |
+        In combination with `extendSafeArea` it will extend the window below the statusbar but also
+        set the statusbar to completly transparent.
+    type: Boolean
+    default: {android: false}
+    since: {android: "9.2.0"}
+    availability: creation
+    osver: {android: {min: "4.4"}}
 
   - name: opacity
     summary: The opacity from 0.0-1.0.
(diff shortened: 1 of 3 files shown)


apidoc/Titanium/UI/Window.yml
```diff
@@ -1152,6 +1152,17 @@ properties:
     permission: read-only
     platforms: [iphone, ipad]
 
+  - name: noLimits
+    summary: Adds the FLAG_LAYOUT_NO_LIMITS to the window.
+    description: |
+        In combination with `extendSafeArea` it will extend the window below the statusbar but also
+        set the statusbar to completly transparent.
```
sgtcoolguy, line 1159:
```suggestion
        set the statusbar to be completely transparent.
```

apidoc/Titanium/UI/Window.yml
```diff
@@ -1152,6 +1152,17 @@ properties:
...
+    type: Boolean
+    default: {android: false}
```
sgtcoolguy, line 1161: We don't have a set doc standard for what `default` values should be, but generally we try to use a literal default value, or some explanatory text if it can't be a literal, or differs by platform.
```suggestion
    default: `false`
```

sgtcoolguy (changes requested):

apidoc/Titanium/UI/Window.yml
m1ga: I've used it from https://github.com/appcelerator/titanium_mobile/blob/5989a607c4dc8adee1d2d888e5427de4d52b055d/apidoc/Titanium/UI/Window.yml#L926

the other `default: false` values are without ` ` `, so shoud I use it here with or without?

garymathews: Ah, we need to fix this as it shows incorrectly https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.Window-property-extendSafeArea

m1ga: https://github.com/appcelerator/titanium_mobile/pull/11924 like this?

sgtcoolguy (approved):

sgtcoolguy:
@garymathews Can you please give CR? LGTM, but you or Josh should verify as well. Also, we'll need to re-target this to 9_3_X since master/9.2.0 is iOS only right now.

jquick-axway:
@m1ga , I disagree with this change.

You're making a bad assumption that `FLAG_LAYOUT_NO_LIMITS` will make the top status bar and bottom navigation bar transparent. This is undocumented. While this appears to be true between Android 5.0 - 10.0 in the emulator, it's not true in the Android 4.4 emulator (the status bar is completely opaque) and this flag is ignored on Android 11 (I tested this in the emulator and a real device). Also, we shouldn't assume 3rd party devices will make the status bar transparent as well.

The `FLAG_LAYOUT_NO_LIMITS` comes with other trade-offs that developers might not want as well. It breaks our "safeAreaPadding" feature, preventing use from detecting system insets anymore like status bar height. It also breaks virtual keyboard "adjustPan" support.

Side Note: You can already set this flag "today" by doing the following...
```javascript
const FLAG_LAYOUT_NO_LIMITS = 512;
	extendSafeArea: true,
	windowFlags: FLAG_LAYOUT_NO_LIMITS,
});
```

The **BEST** solution that you can do today is to use a theme that sets the status bar transparent.

Alternately, we can add an API that allows the developer to set that status bar color via the Java [Window.setStatusBarColor()](https://developer.android.com/reference/android/view/Window#setStatusBarColor(int)) method. This Java method does allow you to set it transparent. We also get to keep are "safeAreaPadding" feature so that you can correctly detect inset height of the status bar. Note that this Java method is API Level 21 (aka: Android 5.0). I think this approach is probably the most convenient to use.

m1ga:
@jquick-axway oh wow! Thanks for having such a detailed look at it. Didn't know that it is different on all platforms. I'll check that again

jquick-axway:
@m1ga , if you don't mind, I'm going to close this PR in favor of PR #12809 which implements this via a "windowFlags" constant.

PR #11924, opened 2020-08-18
docs: fix extendSafeArea text

m1ga:
Follow-up from https://github.com/appcelerator/titanium_mobile/pull/11911#discussion_r471842834

**Optional Description:**

Using the syntax from https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.ImageView-property-autorotate

### DIFF
--- a/apidoc/Titanium/UI/Window.yml
+++ b/apidoc/Titanium/UI/Window.yml
@@ -923,7 +923,7 @@ properties:
         Read more about the safe-area layout-guide in the [Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/overview/iphone-x/).
     platforms: [android, iphone, ipad]
     type: Boolean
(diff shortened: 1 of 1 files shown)


apidoc/Titanium/UI/Window.yml
```diff
@@ -923,7 +923,7 @@ properties:
         Read more about the safe-area layout-guide in the [Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/overview/iphone-x/).
     platforms: [android, iphone, ipad]
     type: Boolean
-    default: {android: false, ios: true}
+    default: false on Android, true on iOS.
```
sgtcoolguy, line 926: To get it to show the raw js value literal as monospace/code, wrap in backticks
```suggestion
    default: `false` on Android, `true` on iOS.
```

PR #11931, opened 2020-08-19
fix(android): remove unnecessary v8 refs

garymathews:
- Remove references to `IsNearDeath()`, which allows Titanium to be used without a modified V8 library
  - This means the V8 library can be compiled without our compatibility patches and work on 9.3.0+


sgtcoolguy (approved):
LGTM

PR #11935, opened 2020-08-20
fix(android): native modules should use min sdk for dependency

garymathews:
- Prevent native Titanium Android modules from depending on current `org.appcelerator.titanium` version and use `minsdk` specified in module manifest

##### TEST CASE
1. Build a native Android module (i.e: https://github.com/appcelerator-modules/titanium-identity)
2. Include module in project that uses an older Titanium SDK
3. Build should not fail with
```
[ERROR] : [GRADLE] Could not determine the dependencies of task ':app:compileDebugJavaWithJavac'.
[ERROR] : [GRADLE] > Could not resolve all task dependencies for configuration ':app:debugCompileClasspath'.
[ERROR] : [GRADLE] > Could not find org.appcelerator:titanium:9.0.3.
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28092)


android/templates/module/generated/build.gradle
```diff
@@ -247,10 +247,14 @@ publishing {
...
-						childNode.appendNode('artifactId', it.name)
-						childNode.appendNode('version', it.version)
+
+						// Skip Titanium SDK dependency.
+						if (it.group != 'org.appcelerator') {
+							def childNode = dependenciesNode.appendNode('dependency')
+							childNode.appendNode('groupId', it.group)
+							childNode.appendNode('artifactId', it.name)
+							childNode.appendNode('version', it.version)
+						}
```
jquick-axway, line 257: Nice catch. I don't think we should remove our Titanium library as a dependency though. We should be able to set the version to the minsdk like this.
```groovy
if ((it.group != null) && (it.name != null)) {
	def childNode = dependenciesNode.appendNode('dependency')
	childNode.appendNode('groupId', it.group)
	childNode.appendNode('artifactId', it.name)
	if ((it.group == 'org.appcelerator') && (it.name == 'titanium')) {
		childNode.appendNode('version', <%- minSdkVersion %>)
	} else {
		childNode.appendNode('version', it.version)
	}
}
```

I coded the above without testing it. So, you'll need to double check it.

jquick-axway: Hold on... my above code is wrong. Don't use `minSdkVersion` because that's the min API Level.

In our "_buildModule.js", we need to pass `manifest.minsdk` as `tiMinSdkVersion` and use that instead.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

PR #11936, opened 2020-08-21
feat: add "uprightWidth" and "uprightHeight" properties to Ti.Blob

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28093

**Summary:**
- These new properties provide image width/height when factoring in JPEG EXIF orientation.
- Needed since Android does not auto-rotate the image when loading into memory. Will only auto-rotate when displaying on-screen. Meaning "width" and "height" represent the image's unrotated dimensions.
- In the future, we should refactor our Android image loading code to load it upright like iOS, but this would be a large undertaking that is best done in a major version of Titanium

**Note:**
If we're not comfortable with adding these properties to the SDK, then alternatively we can add a method to our "ti.imagefactory" to fetch an image's upright dimensions instead.

**Test:**
1. Build and run `DownloadImageBlobTest.js` attached to [TIMOB-28093](https://jira.appcelerator.org/browse/TIMOB-28093) on Android.
2. Make sure device has Internet access.
3. Tap on the "Download" button.
4. Select "JPEG - EXIF Rotate 90" in dialog.
5. Verify image is upright and scaled proportionally (not stretched) to fit the window.
6. Repeat steps 3-5 for "JPEG - EXIF Rotate 180".
7. Repeat steps 3-5 for "JPEG - EXIF Rotate 270".
8. Repeat steps 3-5 for "JPEG - EXIF Flip Horizontal".
9. Repeat steps 3-5 for "JPEG - EXIF Flip Vertical".
10. Repeat steps 3-5 for "JPEG - EXIF Transpose".
11. Repeat steps 3-5 for "JPEG - EXIF Transverse".
12. Repeat all of the above on iOS.

_(Note that the EXIF Flip, Transpose, and Transverse tests will appear as mirrored. Fixing the mirroring effect would require a larger refactoring of our code in the future.)_


garymathews (approved):
CR: **PASS**

Although I'm okay with this change, I think it would have been nice to add `Ti.Blob.exif` where some EXIF properties could be exposed. We could start with the corrected `width`, `height` and `rotation` then add things like `latitude`, `longitude` and `model` in the future. This means we wouldn't need to deprecate these properties in the future when we eventually refactor our image loading code and they wouldn't be redundant on iOS (as width and height is already corrected for rotation).

ssjsamir (approved):
**FR Passed**, Tested using the instructions mentioned in the description. 

_**Note:**_
Android shows mirrored images for `EXIF Flip, Transpose, and Transverse` while on iOS in is upright. 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta 5
Xcode: 12.0 Beta 5
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
iphone 8 Sim (14.0 Beta)
iphone 11 (14.0 Beta 5)
Pixel XL (Android 10)

```

PR #11938, opened 2020-08-21
fix(android): allow node to clear event loop

ewanharris:
As the iterate function here is sync with large numbers of files (say a hyperloop project) we could
run into an issue where node hasn't had a chance to clear the event loop and we'd bog it down with
the constant calls. By putting the test callback into a nextTick call it allows node the chance to
clear the event loop before we choke it

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28090


jquick-axway (approved):
LGTM

ssjsamir (approved):
**FR Passed**: Able to build Hyperloop-examples app to android. 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta 5
Xcode: 12.0 Beta 5
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
Android Pixel XL (10)
```

PR #11967, opened 2020-08-26
feat: add Ti.Platform.versionPatch and global.OS_VERSION_PATCH

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28098

**Description:**
Follow-on to #11859

This may not be quite as useful generally as major/minor - but in working on the macOS support, the third version segment can be fairly useful to know.


jquick-axway (approved):
CR: Pass

PR #11969, opened 2020-08-26
fix(ios): use arc to create corner radius instead of qudratic curve

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28101


tests/Resources/ti.ui.view.test.js
```diff
@@ -1125,6 +1125,37 @@ describe('Titanium.UI.View', function () {
...
+				borderRadius: '30px',
+				backgroundColor: 'yellow'
+			});
+
+			win.addEventListener('postlayout', function postlayout() {
+				win.removeEventListener('postlayout', postlayout); // only run once
+				try {
+					should(view.borderRadius).be.a.String();
+					should(view.borderRadius).eql('30px');
+					should(outerView).matchImage(`snapshots/borderRadius30px_30px_${density}x.png`);
```
sgtcoolguy, line 1148: I would have just named the file: `borderRadius30px_${density}x.png` since you're just setting one value of 30px

tests/Resources/ti.ui.view.test.js
```diff
@@ -1125,6 +1125,37 @@ describe('Titanium.UI.View', function () {
 			win.add(outerView);
 			win.open();
 		});
+
+		it.ios('1 value to create circle', finish => {
```
sgtcoolguy, line 1129: why limit this to iOS? Presumably we want cross-platform consistency. I ran this locally on an emulator with 3.5 density and it appears to line up with iOS.

borderRadius30px_30px_3 5x

sgtcoolguy (changes requested):
Please make the test run cross-platform and use the images I included for android. We want to ensure the UI looks/behaves highly consistently cross-platform.

tests/Resources/ti.ui.view.test.js
vijaysingh-axway: Updated!

tests/Resources/ti.ui.view.test.js
vijaysingh-axway: Added.

tests/Resources/ti.ui.view.test.js
```diff
@@ -1125,6 +1125,37 @@ describe('Titanium.UI.View', function () {
...
+				borderRadius: '30px',
+				backgroundColor: 'yellow'
+			});
+
+			win.addEventListener('postlayout', function postlayout() {
+				win.removeEventListener('postlayout', postlayout); // only run once
+				try {
+					should(view.borderRadius).be.a.String();
+					should(view.borderRadius).eql('30px');
+					should(outerView).matchImage(`snapshots/borderRadius30px_${density}x.png`);
```
sgtcoolguy, line 1148: I just realized - this image should not differ by density, because it's all declared in px, so density here shouldn't matter. Density matters for images when the size of any part is related to the density (i.e. something is using `dp`, `dpi`, `pt` units). So you shouldn't need to inject the density into the filename, and should only need one version of the image per platform.

vijaysingh-axway: You are right. Updated!

garymathews:
@vijaysingh-axway could we add a test with `viewShadow`?

vijaysingh-axway:
> @vijaysingh-axway could we add a test with `viewShadow`?

viewShadow issue is not fixed this PR. I am working on this as part of https://jira.appcelerator.org/browse/TIMOB-28103

sgtcoolguy (approved):

ssjsamir (approved):
**FR Passed**, able to use arc to create corner radius instead of quadratic curve. Tested using the following test cases: 

```
backgroundColor: 'red',
width: 100,
height: 100,
borderRadius: '50 50 50 50'
});
 
window.add(view);
window.open();
```

```
var _window = Ti.UI.createWindow({

backgroundColor : 'blue'
});
 
var view = Ti.UI.createView({
left : 10,
height : 100,
width : 100,
top : 10,
backgroundColor : 'white',
borderWidth : 20,
borderColor : 'red',
borderRadius : 10
});
 
var b3 = Ti.UI.createButton({
top : 250,
title : "borderRadius=30/20"
});
b3.addEventListener("click", function() {
view.borderRadius = ['30', '20', '30', '20'];
});
 
var b4 = Ti.UI.createButton({
top : 300,
title : "borderRadius=30 All"
});
b4.addEventListener("click", function() {
view.borderRadius = 30;
});
 
var b5 = Ti.UI.createButton({
top : 350,
title : "borderRadius=0"
});
b5.addEventListener("click", function() {
view.borderRadius = 0;
});
 
_window.add(view);
_window.add(b3);
_window.add(b4);
_window.add(b5);
 
_window.open();
```


```
backgroundColor: '#333'
});
 
var view = Ti.UI.createView({
backgroundColor: '#fff',
height: 350,
bottom: 0,
borderRadius: '20 20dp 20px 12px',
});
 
win.add(view);
win.open();
```
**Test Environment**
```
MacOS Big Sur: 11.0 Beta 5
Xcode: 12.0 Beta 5
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
iphone 8 Sim (14.0 Beta)
```

PR #11977, opened 2020-08-28
feat(ios): support new date picker styles

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28104

Note: Without these, default pickers will look off on iOS 14+, so developers should very carefully check their apps for pickers.

### DIFF
--- a/apidoc/Titanium/UI/Picker.yml
+++ b/apidoc/Titanium/UI/Picker.yml
@@ -364,6 +364,17 @@ properties:
     platforms: [android]
     since: "5.0.0"
 
+  - name: datePickerStyle
+    summary: Request a style if the picker is a date picker (`PICKER_TYPE_DATE`).
+    description: |
+        If the style changed, then the date picker may need to be resized and will
+        generate a layout pass to display correctly.
+    type: Number
+    platforms: [iphone, ipad]
+    constants: Titanium.UI.iOS.DATE_PICKER_STYLE_*
+    default: <Titanium.UI.iOS.DATE_PICKER_STYLE_AUTOMATIC>
+    osver: { ios: { min: "13.4" } }
+
   - name: type
     summary: Determines the type of picker displayed
     description: |
--- a/apidoc/Titanium/UI/iOS/iOS.yml
+++ b/apidoc/Titanium/UI/iOS/iOS.yml
@@ -1348,6 +1348,41 @@ properties:
     since: "3.2.0"
     osver: {ios: {min: "7.0"}}
 
+  - name: DATE_PICKER_STYLE_AUTOMATIC
+    summary: |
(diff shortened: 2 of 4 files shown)


sgtcoolguy (approved):
code changes LGTM, we should add some unit tests that verify the new constants exist in JS, maybe a UI test that validates once set we get the sort of UI we expect?

ssekhri:
The datePickerStyle of DATE_PICKER_STYLE_COMPACT is mentioned as min iOS 13.4. Should it not be min iOS 14.0?

hansemannn:
No, it's iOS 13.4 already: https://developer.apple.com/documentation/uikit/uidatepickerstyle/uidatepickerstylecompact?language=objc

ssekhri:
@hansemannn thanks for the reference. However with the current PR the DATE_PICKER_STYLE_COMPACT does not show up as compact style on iOS 13.4+ device/simulator. It works fine on iOS 14. On iOS 13.4+ it still shows as wheel style.

hansemannn:
It is primarily used for iOS 13.4 apps that use Catalyst (macOS support), thats why I added it here already. In iOS 14, it becomes available on the "normal" / non-Catalyst iOS apps as well. Should I change it?

vijaysingh-axway:
> It is primarily used for iOS 13.4 apps that use Catalyst (macOS support), thats why I added it here already. In iOS 14, it becomes available on the "normal" / non-Catalyst iOS apps as well. Should I change it?

I don't think we should change it. As it is marked as iOS 13.4+ in apple doc, we should also have the same. Not working properly on iOS 13.4, might be apple bug. @ssekhri May be you can try on iOS 13.5, 13.6 and 13.7.

hansemannn:
I mean, it will work on iOS 13.4 combined with Catalyst. So I think it should be fine, especially because #11955 is rolling in soon.

ssekhri:
I agree that we should be in sync with Apple docs on the version support. However it would be good to have it mentioned in the docs that DATE_PICKER_STYLE_COMPACT style would not show as expected on iOS <14 until used with Mac Catalyst.

hansemannn:
@ssekhri I agree - and added the notes.

ssekhri:
FR Passed.
The datePickerStyle property working fine on iOS 13.4+ and iOS 14.
The property can be successfully set to DATE_PICKER_STYLE_AUTOMATIC, DATE_PICKER_STYLE_WHEELS, DATE_PICKER_STYLE_COMPACT, DATE_PICKER_STYLE_INLINE.

Verified on:
Mac OS: 10.15.4
SDK: 9.2.0.v20200831152744
Appc CLI: 8.1.0
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0 Beta6
iPhone 7Plus(v14.0 Beta6), iOS simulator 13.5

PR #12020, opened 2020-09-08
fix(ios): guard new picker types

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28112

Keeps compatibility with the stated xcode versions, from what I can tell this seems safe to use? It fixes the issue on 11.3.1 for me and I assume it's backwards compatible

### DIFF
--- a/iphone/Classes/TiUIiOSProxy.m
+++ b/iphone/Classes/TiUIiOSProxy.m
@@ -192,7 +192,7 @@
 
 #ifdef USE_TI_UIPICKER
 
-#if __IPHONE_13_4
+#if IS_SDK_IOS_134
 - (NSNumber *)DATE_PICKER_STYLE_AUTOMATIC
 {
   if (![TiUtils isIOSVersionOrGreater:@"13.4"]) {
--- a/iphone/Classes/TiUIiOSProxy.m
+++ b/iphone/Classes/TiUIiOSProxy.m
@@ -192,6 +192,7 @@
 
 #ifdef USE_TI_UIPICKER
 
+#if __IPHONE_13_4
 - (NSNumber *)DATE_PICKER_STYLE_AUTOMATIC
 {
   if (![TiUtils isIOSVersionOrGreater:@"13.4"]) {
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIiOSProxy.m
```diff
@@ -192,6 +192,7 @@ - (NSNumber *)ROW_ACTION_STYLE_NORMAL
 
 #ifdef USE_TI_UIPICKER
 
+#if __IPHONE_13_4
```
sgtcoolguy, line 195: I'm not sure of the typical guards used on the iOS codebase. Looks like we tend to use `__IPHONE_OS_VERSION_MAX_ALLOWED`? Also, I think e may want the guards to be placed inside each method so if it's not defined we return the -1 value that we return in the runtime guard case where iOS < 13.4

ewanharris: I tried to see if there was any form of existing `IS_SDK_IOS_XYZ` and the only similar thing I found was this in TiUIWebViewProxy https://github.com/appcelerator/titanium_mobile/blob/840b0d279f79248d1511fc518fa28fda9573be73/iphone/Classes/TiUIWebViewProxy.m#L619

I wasn't sure which form is right, a new `IS_SDK_IOS_134` or this one that seems to be built in (or the difference of them).

vijaysingh-axway: @ewanharris  1. __IPHONE_OS_VERSION_MAX_ALLOWED is preferred way. 
    2. Positioning is fine.
    3. Need to guard in function https://github.com/appcelerator/titanium_mobile/blob/00f4b58d03678d1a266fa9a75713f938556f105a/iphone/Classes/TiUIPicker.m#L182 as well.

ewanharris: Updated as requested!

iphone/iphone/Titanium_Prefix.pch
```diff
@@ -17,6 +17,12 @@
 #define IS_SDK_IOS_14 false
 #endif
 
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130400
+#define IS_SDK_IOS_134 true
```
vijaysingh-axway, line 21: We might want to change this macro  to  IS_SDK_IOS_13_4.  Convention should be IS_SDK_IOS_major_minor.

PR #12026, opened 2020-09-09
feat(ios): use/support/build xcframeworks

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28116

**Description:**
This is required for #11955 

These are the changes to move to using xcframeworks for APS, tiverify, TitaniumKit and for native modules.

This includes changes to the libraries:

- appcelerator/aps_sdk#347
- appcelerator/APSHTTPClient#54
- appcelerator/titanium_verify#33

Xcode minimum is bumped to 11 as that is required for xcframework support.

This PR does build/include the built maccatalyst binaries in the xcframeworks. It *does not* enable macOS as a target platform. #11955 will be used for that.

**Note** that to avoid breaking existing native modules we will selectively exclude arm64 from simulator builds of apps if we detect a native module that is not an xcframework - or is built with Xcode 11 and is missing arm64 simulator binaries.

If the developer is running an arm64 Mac build of Node (there are none yet officially!), and a native module is assumed to not have arm64 sim support, we'll error out with a message about it. (the "fix" there is for us/them to rebuild the native modules using Xcode 12 w/ xcframework support)

**TODO**
 - [x] Address @vijaysingh-axway review feedback around use of Info.plist from xcframework to determine arch/folder name support.
 - [x] make the arm64 sim exclusion smarter by handling if an xcframework has no arm64 sim support.


sgtcoolguy:
@vijaysingh-axway This is the xcframeworks only changes. I addressed your initial feedback re: Xcode 11 built xcframeworks in a lazy/dumb way - checking for the Xcode 12+ dir name and falling back to the Xcode 11 folder name (rather than try and get fancy with parsing Info.plist).

If we see an xcframework built on Xcode 11 missing arm64, we'll exclude arm64 on app simulator builds just as we would for a non-xcframework based native module.

iphone/Classes/ContactsModule.h
```diff
@@ -6,7 +6,7 @@
  */
 #import <TitaniumKit/TiModule.h>
 
-#ifdef USE_TI_CONTACTS
+#if defined(USE_TI_CONTACTS) && (!TARGET_OS_MACCATALYST || IS_SDK_IOS_14)
```
vijaysingh-axway, line 9: I see that TARGET_OS_MACCATALYST is getting used. I guess it is not removed intentionally to save time. I think it is fine.

iphone/cli/commands/_buildModule.js
```diff
@@ -460,156 +459,202 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
...
+		const macOsVersion = osInfo.osver;
+
+		// 1. Create a build for the simulator
+		xcodebuildHook(xcBuild, xcodeBuildArgumentsForTarget('iphonesimulator'), opts, 'xcode-sim', () => {
+			// 2. Create a build for device
+			xcodebuildHook(xcBuild, xcodeBuildArgumentsForTarget('iphoneos'), opts, 'xcode-dist', () => {
+				const osVersionParts = macOsVersion.split('.').map(n => parseInt(n));
+				if (osVersionParts[0] > 10 || (osVersionParts[0] === 10 && osVersionParts[1] >= 15)) {
+					// 3. Create a build for the maccatalyst
+					xcodebuildHook(xcBuild, xcodeBuildArgumentsForTarget('macosx'), opts, 'xcode-macos', next);
```
vijaysingh-axway, line 492: building module for macOS is intentional?

iphone/iphone/Titanium.entitlements
```diff
@@ -1,5 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-<dict/>
+<dict>
```
vijaysingh-axway, line 4: These should be removed as we are excluding mac target in this PR.

iphone/iphone/Titanium.xcodeproj/project.pbxproj
```diff
@@ -2406,9 +2417,15 @@
...
 				PRODUCT_NAME = Titanium;
+				PROVISIONING_PROFILE_SPECIFIER = "";
+				"PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*]" = "";
+				SUPPORTS_MACCATALYST = YES;
 				SWIFT_OBJC_BRIDGING_HEADER = "Titanium-Bridging-Header.h";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_VERSION = 5.0;
+				PROVISIONING_PROFILE_SPECIFIER = "";
+				"PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*]" = "";
+				SUPPORTS_MACCATALYST = YES;
```
vijaysingh-axway, line 2428: Is it left intentional? Or we should remove this?

iphone/package.json
```diff
@@ -19,7 +19,7 @@
 	"minIosVersion": "10.0",
 	"minWatchosVersion": "2.0",
 	"vendorDependencies": {
-		"xcode": ">=9.0 <=12.x",
+		"xcode": ">=11.0 <=12.x",
 		"ios sdk": ">=10.0 <=14.x"
```
vijaysingh-axway, line 23:
```suggestion
		"ios sdk": ">=13.0 <=14.x"
```

vijaysingh-axway (commented):
@sgtcoolguy  Changes looks good. I have left few comments. Basically questions. Can you look in that? Thanks!

iphone/cli/commands/_buildModule.js
sgtcoolguy: Yes, I wanted to future proof. We do still plan on landing macOS if we can. But if we don't this would avoid having to rebuild the module multiple times.

iphone/Classes/ContactsModule.h
sgtcoolguy: Contacts API is available in Mac Catalyst 14+. So I intentionally guarded here like this.

iphone/iphone/Titanium.xcodeproj/project.pbxproj
sgtcoolguy: Intentional. Wanted to guard for Mac catalyst so modules could build for catalyst and just be ready without having to make them rebuild multiple times.

iphone/package.json
sgtcoolguy: Yeah, I think the min iOS bump can easily be done separately and iOS a bigger issue.

iphone/iphone/Titanium.entitlements
sgtcoolguy: Yeah, this should probably be in the macOS support PR.

sgtcoolguy:
@ewanharris I've incorporated your feedback. Updated the line for the framework search path sniffing, and right before we create-xcframework we'll wipe the destination if it exists to avoid the "couldn't be copied" error.

PR #12043, opened 2020-09-12
fix: add version to new "datePickerStyle" property

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28104

Also, the `default` value does not seem to be picked up correctly?!

Bildschirmfoto 2020-09-12 um 16 10 27


sgtcoolguy:
@hansemannn We have no real "spec" for default values. They should be treated more like strings than markdown, so using the special custom markdown-esque link to a constant won't render as expected right now.

PR #12046, opened 2020-09-14
Fix incremental builds

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28130

* Unmarks the Product directory to prevent us removing required files like the tiverify lib and TitaniumKit framework
* Small change to the buildModule file to call `Array.from` on the architectures set when logging it to  avoid it being logged as `[object Set]`

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -6687,13 +6687,19 @@ iOSBuilder.prototype.removeFiles = function removeFiles(next) {
 	this.unmarkBuildDirFiles(path.join(this.buildDir, 'export_options.plist'));
 	this.unmarkBuildDirFiles(path.join(this.buildDir, this.tiapp.name + '.xcarchive'));
 
+	const productsDir = path.join(this.buildDir, 'build', 'Products');
 	try {
-		const releaseDir = path.join(this.buildDir, 'build', 'Products', 'Release-iphoneos');
+		const releaseDir = path.join(productsDir, 'Release-iphoneos');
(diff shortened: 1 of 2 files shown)


iphone/cli/commands/_build.js
```diff
@@ -4638,7 +4638,7 @@ iOSBuilder.prototype.copyTitaniumiOSFiles = function copyTitaniumiOSFiles() {
 						// The content of Frameworks directory only change if we change SDK version. So it is safe to copy whole directory.
 						// Copy whole 'Frameworks' directory from SDK to build directory, to preserve symlink available in Titaniumkit.xcframework.
 						// TODO: Is there any better way to do this?
-
+						fs.emptyDirSync(path.join(this.buildDir, dir));
```
sgtcoolguy, line 4641: may want to stick a comment in the code why we're doing this so it doesn't get reverted way down the road.

sgtcoolguy (approved):
LGTM

cb1kenobi (approved):
LGTM

ssjsamir (approved):
FR Passed: Able to build incrementally without any errors (iOS and MacOS)

**Test Environment**
```
MacOS Big Sur: 11.0 Beta 6
Xcode: 12.0 Beta 5
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1-master.5""
iphone 8 Sim (14.0 Beta)
```

iphone/cli/commands/_build.js
ewanharris: Had missed this comment, just pushed up a commit now that explains why we nuke before copying

PR #12062, opened 2020-09-14
chore(deps): upgrade ti.map for ios to 4.0.0

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28124

**Description:**
includes arm64 sim support, macos support

PR #12063, opened 2020-09-14
chore(deps): upgrade ti.identity for ios to 2.0.0

sgtcoolguy:

**Optional Description:**
includes arm64 sim support, macOS support

Related to #12062


ewanharris (approved):
FR passed across simulator, device, and mac targets. Noted that on mac the keychain example currently throws an entitlement error, as noted in the CLI build code I think we'll need to figure out how to allow injecting entitlements

PR #12065, opened 2020-09-14
chore(deps): upgrade ti.applesignin to 2.0.0

sgtcoolguy:

**Description:**
includes arm64 sim support, macOS support

Relates to #12063


ewanharris (approved):
FR passed.

Module functions ok on iOS 14 device, similarly to ti.identity I'm struggling to get the signing/entitlements configured correctly for macos target to allow this but I don't think that should block this from landing

PR #12090, opened 2020-09-16
fix(ios): use TITANIUM_SDK variable to point at xcframework path in xcode project

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28143

**Description:**
This was informally discussed on Teams, so there is no JIRA ticket yet. If you create a module with a 9.2.x SDK, the module template baked in the path of that SDK's version of the TitaniumKit.xcframework. This change makes the file path reference use the `$(TITANIUM_SDK)` variable in the path so that when users/developers update the `ios/titanium.xcconfig` file to target a specific SDK version, that will get reflected into the xcframework reference.

I did this manually in the apple sign in module when I added xcframework/Mac support: https://github.com/appcelerator-modules/titanium-apple-sign-in/commit/010324234d37dc778222c8d5333929a397e4dd58#diff-ab2479047e1642c6a102069cd6069bb2R24


vijaysingh-axway (approved):
CR passed. In my test it is working fine!

lokeshchdhry:
FR Passed.

PR #12092, opened 2020-09-16
 feat(ios): support xcframeworks in modules/platform folders

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-27986

**Description:**
This adds support for integrating `.xcframework`s in the same manner we do `.framework`s for app projects.

I tested this while developing appcelerator-modules/ti.barcode#134

It seems to properly integrate the ti.barcode module and the `ZXingObjC.xcframework` that module now includes (in it's `platform` folder on that PR). I fumbled through this a bit, so it may be doing some things wrong: 
- such as adding wrong/too many search paths for frameworks? 
- Assuming all xcframeworks are dynamic (tiverify.xcframework in the SDK itself isn't. I don't know the correct way to "check" for static vs dynamic xcframeworks. Just look for .framework subdirs?)
- I didn't attempt to tackle the general issue that we (and Apple!) have long conflated architectures with arch/target combinations. This is present in our module code as well (i.e. the manifest). We can't use 'arm64' as shorthand for device arm64 anymore - there will be arm64 Apple Silicon simulators! And arm64 Apple Silicon maccatalyst targets! So I don't know what to do with the collection of architectures on xcframeworks and stripping them. I assume Xcode does that itself with xcframewokrs (they're basically pre-sliced target frameworks/binaries with multiple architectures for that target).

### DIFF
--- a/iphone/cli/hooks/frameworks.js
+++ b/iphone/cli/hooks/frameworks.js
@@ -9,11 +9,18 @@ const exec = require('child_process').exec; // eslint-disable-line security/dete
 const fs = require('fs-extra');
 const IncrementalFileTask = require('appc-tasks').IncrementalFileTask;
 const path = require('path');
+const plist = require('simple-plist');
+const { promisify } = require('util');
 
-const frameworkPattern = /([^/]+)\.framework$/;
+const parsePlist = promisify(plist.readFile);
+
+const frameworkPattern = /([^/]+)\.(xc)?framework$/;
 
 const FRAMEWORK_TYPE_STATIC = 'static';
 const FRAMEWORK_TYPE_DYNAMIC = 'dynamic';
+
+const MAC_CATALYST_PLATFORM_VARIANT = 'maccatalyst';
+const IOS_PLATFORM = 'ios';
 
 exports.cliVersion = '>=3.2.1';
 
@@ -57,7 +64,12 @@ class FrameworkManager {
 			post: (builder, callback) => {
 				this._logger.trace('Starting third-party framework detection');
 				this._builder = builder;
-				this.detectFrameworks().then(callback, callback);
+				this.detectFrameworks().then(callback, e => {
+					if (this._cli.argv.platform === 'ios') {
+						this._logger.error(e.stack);
+					}
+					callback(e);
+				});
 			}
 		});
 
@@ -69,54 +81,52 @@ class FrameworkManager {
 	/**
 	 * Detects all available frameworks.
 	 *
-	 * @return {Promise}
+	 * @return {Promise<void>}
 	 * @access private
 	 */
-	detectFrameworks() {
-		return this.findFrameworkPaths()
-			.then(frameworkPaths => {
-				if (frameworkPaths.length === 0) {
-					return Promise.resolve();
-				}
+	async detectFrameworks() {
+		const frameworkPaths = await this.findFrameworkPaths();
+		if (frameworkPaths.length === 0) {
+			return;
+		}
 
-				let incrementalDirectory = path.join(this._builder.projectDir, 'build', 'incremental');
-				let outputDirectory = path.join(this._builder.projectDir, 'build', 'inspectFrameworks');
-				let task = new InspectFrameworksTask({
-					name: 'ti:inspectFrameworks',
-					logger: this._logger,
-					incrementalDirectory: incrementalDirectory
-				});
-				task.outputDirectory = outputDirectory;
-				frameworkPaths.forEach(frameworkPath => {
-					task.addFrameworkPath(frameworkPath);
-				});
-				task.postTaskRun = () => {
-					this._frameworks = task.frameworks;
+		const incrementalDirectory = path.join(this._builder.projectDir, 'build', 'incremental');
+		const outputDirectory = path.join(this._builder.projectDir, 'build', 'inspectFrameworks');
+		const task = new InspectFrameworksTask({
+			name: 'ti:inspectFrameworks',
+			logger: this._logger,
+			incrementalDirectory
+		});
+		task.outputDirectory = outputDirectory;
+		frameworkPaths.forEach(frameworkPath => {
+			task.addFrameworkPath(frameworkPath);
+		});
+		task.postTaskRun = () => {
+			this._frameworks = task.frameworks;
 
-					// Convert the internal ES6 map to an object to avoid ES6 in the builder
-					let frameworkObject = {};
-					this._frameworks.forEach(frameworkInfo => {
-						frameworkObject[frameworkInfo.name] = {
-							name: frameworkInfo.name,
-							path: frameworkInfo.path,
-							type: frameworkInfo.type,
-							architectures: Array.from(frameworkInfo.architectures)
-						};
-					});
-					this._builder.frameworks = frameworkObject;
+			// Convert the internal ES6 map to an object to avoid ES6 in the builder
+			const frameworkObject = {};
+			this._frameworks.forEach(frameworkInfo => {
+				frameworkObject[frameworkInfo.name] = {
+					name: frameworkInfo.name,
+					path: frameworkInfo.path,
+					type: frameworkInfo.type,
+					architectures: Array.from(frameworkInfo.architectures)
 				};
-				return task.run();
 			});
+			this._builder.frameworks = frameworkObject;
+		};
+		return task.run();
 	}
 
 	/**
-	 * Finds any .framework directories inside the project and its modules.
+	 * Finds any .framework/.xcframework directories inside the project and its modules.
 	 *
-	 * @return {Promise} Promise resolving to array of available third-party framework directory paths
+	 * @return {Promise<string[]>} Promise resolving to array of available third-party framework directory paths
 	 * @access private
 	 */
-	findFrameworkPaths() {
-		let pathsToScan = [
+	async findFrameworkPaths() {
+		const pathsToScan = [
 			path.join(this._builder.projectDir, 'platform', 'ios'),
 			path.join(this._builder.projectDir, 'platform', 'iphone')
 		];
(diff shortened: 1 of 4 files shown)


iphone/cli/hooks/frameworks.js
```diff
@@ -526,11 +533,11 @@ class FrameworkIntegrator {
...
 	 */
 	addFrameworkFileReference(frameworkInfo) {
-		let frameworkPackageName = frameworkInfo.name + '.framework';
-		let fileRefUuid = this._builder.generateXcodeUuid();
+		const frameworkPackageName = frameworkInfo.packageName;
+		const fileRefUuid = this._builder.generateXcodeUuid();
 		this._xobjs.PBXFileReference[fileRefUuid] = {
 			isa: 'PBXFileReference',
-			lastKnownFileType: 'wrapper.framework',
+			lastKnownFileType: frameworkInfo.wrapperName, // FIXME: What if framework is an xcframework?
```
vijaysingh-axway, line 540: In case of xcframework it should be "wrapper.xcframework" .

iphone/cli/hooks/frameworks.js
```diff
@@ -552,13 +559,13 @@ class FrameworkIntegrator {
...
-		let frameworkPackageName = frameworkInfo.name + '.framework';
-		var buildFileUuid = this._builder.generateXcodeUuid();
+		const frameworkPackageName = frameworkInfo.packageName;
+		const buildFileUuid = this._builder.generateXcodeUuid();
 		this._xobjs.PBXBuildFile[buildFileUuid] = {
 			isa: 'PBXBuildFile',
 			fileRef: fileRefUuid,
 			fileRef_comment: frameworkPackageName,
-			platformFilter: 'ios'
+			// platformFilter: 'ios' // FIXME: Is this needed?
```
vijaysingh-axway, line 568: 'platformFilter'  ensures that if included framework/xcframework is for 'ios' only or 'macos' only or iOS+macOS both (default).

So if a .framework supports ios only (which is general case), above line should be there.
In case of .xcframework, we need to check if it contains mac library or not. If not, above line should be there.

iphone/cli/hooks/frameworks.js
```diff
@@ -568,21 +575,21 @@ class FrameworkIntegrator {
...
-		let embeddedBuildFileUuid = this._builder.generateXcodeUuid();
+		const frameworkPackageName = frameworkInfo.packageName;
+		const embeddedBuildFileUuid = this._builder.generateXcodeUuid();
 		this._xobjs.PBXBuildFile[embeddedBuildFileUuid] = {
 			isa: 'PBXBuildFile',
 			fileRef: fileRefUuid,
 			fileRef_comment: frameworkPackageName,
 			settings: { ATTRIBUTES: [ 'CodeSignOnCopy', 'RemoveHeadersOnCopy' ] },
-			platformFilter: 'ios'
+			// platformFilter: 'ios' // FIXME: Is this necessary?
```
vijaysingh-axway, line 592: Same as above

iphone/cli/hooks/frameworks.js
janvennemann: I think the fixme comment can be removed, the correct file type will be returned [here](https://github.com/appcelerator/titanium_mobile/pull/12092/files#diff-db24295025b1208c3a957f26d2358510R903)

iphone/cli/hooks/frameworks.js
```diff
@@ -808,15 +821,20 @@ class FrameworkInfo {
...
 
-		if (!(architectures instanceof Set)) {
-			throw new TypeError('Framework architectures must be a set of valid architectures');
+			if (!(architectures instanceof Set)) {
+				throw new TypeError('Framework architectures must be a set of valid architectures');
+			}
+			this._architectures = architectures;
+		} else {
+			this._type = FRAMEWORK_TYPE_DYNAMIC; // FIXME: Determine if this is dynamic. We shoudl be able to look under the folders and see somehow...
+			this._architectures = new Set(); // just ignore arches?
```
janvennemann, line 836: `.xcframework` packages contain a `Info.plist` we could parse to get a list of all included libraries. This can then be used to determine whether the binaries are dynamic or static. I'll implement this and then update the PR!

vijaysingh-axway: @janvennemann Parsing Info.plist would be great. We need it in [_build.js](https://github.com/appcelerator/titanium_mobile/blob/eca0416a255b1cac57c6526d257860225465b0cd/iphone/cli/commands/_build.js#L2230) and [_buildModule.js](https://github.com/appcelerator/titanium_mobile/blob/193bbc3b5d8b80fc1b6f8cf4fb1dedbf12a64986/iphone/cli/commands/_buildModule.js#L631)

janvennemann:
@sgtcoolguy @vijaysingh-axway i added a couple of things to this PR:

- Parsing supported archs and platforms from `Info.plist`
- Determine mach-o type from one library inside the `.xcframework` package. I assumed that all libraries will have the same type so i only check this one time.
- Set `platformFilter` based on the parsed platforms as mentioned above by Vijay.

iphone/cli/hooks/frameworks.js
```diff
@@ -201,7 +210,7 @@ class FrameworkManager {
 		let simulatorArchitectures = new Set([ 'i386', 'x86_64' ]);
 
 		for (let frameworkInfo of this._frameworks.values()) {
-			if (frameworkInfo.type !== FRAMEWORK_TYPE_DYNAMIC) {
+			if (frameworkInfo.type !== FRAMEWORK_TYPE_DYNAMIC || frameworkInfo.wrapperName.endsWith('xcframework')) {
```
sgtcoolguy, line 213: we should probably just extract a property/method on FrameworkInfo that tells you if it's an xcframework. We have conditionals like this all over the code now. (In fact, design-wise it may make sense to have a subclass called `XCFrameworkInfo` that handles those small changes as method/getter overrides)

iphone/cli/hooks/frameworks.js
```diff
@@ -796,6 +827,14 @@ class FrameworkInspector {
 	}
 }
 
+/**
+ * @typedef FrameworkMetadata
+ * @property {Object} name Framework metadata
+ * @property {String} path Path to the framework
+ * @property {String} type Framework's binary type (static or dynamic)
+ * @property {Set<string>} architectures - Set of supported architectures
```
sgtcoolguy, line 835: missing optional supportedPlatforms property on this typedef

iphone/cli/hooks/frameworks.js
```diff
@@ -740,20 +754,37 @@ class FrameworkInspector {
...
+			const result = await this.detectBinaryTypeAndArchitectures(binaryPathAndFilename);
+			meta.type = type = result.type;
+			meta.architectures = result.architectures;
+			frameworkInfo = new FrameworkInfo(meta);
+		} else {
+			const archs = new Set();
+			const supportedPlatforms = new Set();
+			const xcFrameworkInfo = await parsePlist(path.join(frameworkPath, 'Info.plist'));
+			for (const libInfo of xcFrameworkInfo.AvailableLibraries) {
+				const libraryPath = path.join(frameworkPath, libInfo.LibraryPath);
```
sgtcoolguy, line 773: move this under the if guard to avoid creating if not used

iphone/cli/hooks/frameworks.js
```diff
@@ -740,20 +754,37 @@ class FrameworkInspector {
...
+		let type;
+		const meta = { name: frameworkName, path: frameworkPath };
+		if (packageExtension === '.framework') {
+			const binaryPathAndFilename = path.join(frameworkPath, frameworkName);
+			const result = await this.detectBinaryTypeAndArchitectures(binaryPathAndFilename);
+			meta.type = type = result.type;
+			meta.architectures = result.architectures;
+			frameworkInfo = new FrameworkInfo(meta);
+		} else {
+			const archs = new Set();
```
sgtcoolguy, line 769: I still think this code and most of our CLI code should move away from bare architecture sets like this, as they're rendered ambiguous by the new Apple Silicon stuff. `'arm64'` is now ambiguous and is not implicitly iOS device. So I'm not sure collecting this info is actually useful unless we're combining arch with "platform".

In fact, the move to xcframework makes our use of architectures in module `manifest`s redundant and useless since the xcframework spells out exactly the set of targets and architectures it holds.

iphone/cli/hooks/frameworks.js
```diff
@@ -55,7 +61,12 @@ class FrameworkManager {
 			post: (builder, callback) => {
 				this._logger.trace('Starting third-party framework detection');
 				this._builder = builder;
-				this.detectFrameworks().then(callback, callback);
+				this.detectFrameworks().then(callback, e => {
+					if (this._cli.argv.platform === 'ios') {
```
sgtcoolguy, line 65: why spit out the error stack trace only on iOS? Won't this always be true? Wouldn't we spit out the full error before?

sgtcoolguy (commented):
I can't approve, only "comment", since I'm the original author 😆 

Anyways, feedback is optional below. Looks good to me.

iphone/cli/hooks/frameworks.js
```diff
@@ -728,28 +749,51 @@ class FrameworkInspector {
...
+			const supportedPlatforms = new Set();
+			const xcFrameworkInfo = await parsePlist(path.join(frameworkPath, 'Info.plist'));
+			for (const libInfo of xcFrameworkInfo.AvailableLibraries) {
+				const libraryPath = path.join(frameworkPath, libInfo.LibraryPath);
+				if (!type) {
+					const result = await this.detectBinaryTypeAndArchitectures(libraryPath);
+					type = result.type;
+				}
+				libInfo.SupportedArchitectures.forEach(a => archs.add(a));
+				supportedPlatforms.add(libInfo.SupportedPlatform);
```
vijaysingh-axway, line 779: 'SupportedPlatform' is always 'ios'. I think you want to use 'SupportedPlatformVariant' .

iphone/cli/hooks/frameworks.js
janvennemann: The collection of archs here is only done to determine if a third-party framework uses a fat binary that needs to be stripped of unused archs when doing a dist build. It's true that XCFramework packages don't need this anymore since Xcode makes sure to only link with the correct library archs, but i still decided to collect all archs present in the package to have the complete metadata, just in case.

iphone/cli/hooks/frameworks.js
janvennemann: Well, it's kinda messy (surprise!). This has been an issue on both platforms until Josh refactored the whole Android build for the Gradle support. Now the Android build manually spits out the error [here](https://github.com/appcelerator/titanium_mobile/blob/41ebeecedf3661685bff8ad6626d30295aa488e6/android/cli/commands/_build.js#L1575). The iOS build on the other hand has no special handling and relies on the default error handling in the common builder [here](https://github.com/appcelerator/titanium_mobile/blob/master/cli/commands/build.js#L332) which only prints out the error message.

iphone/cli/hooks/frameworks.js
janvennemann: I'm not sure i follow here. A XCFramework package can contain binaries for MacOS as well. I read the `SupportedPlatform` here to set the appropriate `PlatformFilter` in the Xcode build file entry. Why should `SupportedPlatform` always be `ios`?

sgtcoolguy: @janvennemann Vijay is referring to the info.plist value for `SupportedPlatform` always `=== 'ios'`. So you can't use that value to determine the filter later. You need to look at the optional `SupportedPlatformVariant` key's value, which `=== 'maccatalyst'` for macOS, and `=== 'simulator'` for the sim binaries.

janvennemann: Ah yeah, i used a test framework to check possible values of the Info.plist and forgot to build the MacOS binary with the `-destination 'platform=macOS,variant=Mac Catalyst'` flag. Now that i rebuilt if with this flag i sea what you mean.

vijaysingh-axway: Yes. @sgtcoolguy  is right. I was referring Info.plist's contents.

janvennemann (approved):

PR #12093, opened 2020-09-16
fix(ios): ignore the mac build if it failed and exclude sim arm64 if thirdparty .framework is there

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28138

-  Silently ignore the mac build, if it failed for any reason( e.g included third party framework does not support mac). Log a warning.
- If in  module there is any third party .framework file, exclude arm 64 from simulator build and give a warning. .xcframework is only way to support simulator arm64.
- Build fb module using PR https://github.com/appcelerator-modules/ti.facebook/pull/128. It is building but obviously no mac-target support as Facebook framework do not have mac support. 
- It will ensure backward compatibility of module build.

### DIFF
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -13,7 +13,7 @@
 
 'use strict';
 
-const { exec, spawn } = require('child_process'); // eslint-disable-line security/detect-child-process
+const { spawn } = require('child_process'); // eslint-disable-line security/detect-child-process
 
 const appc = require('node-appc'),
 	archiver = require('archiver'),
@@ -29,7 +29,11 @@ const appc = require('node-appc'),
 	temp = require('temp'),
 	util = require('util'),
 	__ = appc.i18n(__dirname).__,
-	series = appc.async.series;
+	series = appc.async.series,
+	xcode = require('xcode');
+
+const plist = require('simple-plist');
+const parsePlist = util.promisify(plist.readFile);
 
 function iOSModuleBuilder() {
 	Builder.apply(this, arguments);
@@ -47,6 +51,7 @@ iOSModuleBuilder.prototype.validate = function validate(logger, config, cli) {
 	this.moduleName    = cli.manifest.name;
 	this.moduleIdAsIdentifier = this.scrubbedName(this.moduleId);
 	this.moduleVersion = cli.manifest.version;
+	this.isMacOSEnabled  = this.detectMacOSTarget();
 	this.moduleGuid    = cli.manifest.guid;
 	this.isFramework   = fs.existsSync(path.join(this.projectDir, 'Info.plist')); // TODO: There MUST be a better way to determine if it's a framework (Swift)
 
@@ -86,6 +91,32 @@ iOSModuleBuilder.prototype.validate = function validate(logger, config, cli) {
 	}.bind(this);
 };
 
+iOSModuleBuilder.prototype.detectMacOSTarget = function detectMacOSTarget() {
+	if (this.manifest.mac !== undefined) {
+		return this.manifest.mac === 'true';
+	}
+	const pbxFilePath = path.join(this.projectDir, `${this.moduleName}.xcodeproj`, 'project.pbxproj');
+
+	if (!fs.existsSync(pbxFilePath)) {
+		this.logger.warn(__(`The Xcode project does not contain the default naming scheme. This is required to detect macOS targets in your module.\nPlease rename your Xcode project to "${this.moduleName}.xcodeproj"`));
+		return true;
+	}
+
+	let isMacOSEnabled = true;
+	const proj = xcode.project(pbxFilePath).parseSync();
+	const configurations = proj.hash.project.objects.XCBuildConfiguration;
+
+	for (const key of Object.keys(proj.hash.project.objects.XCBuildConfiguration)) {
+		const configuration = configurations[key];
+		if (typeof configuration === 'object' && configuration.buildSettings.SUPPORTS_MACCATALYST) {
+			isMacOSEnabled = configuration.buildSettings.SUPPORTS_MACCATALYST === 'YES';
+			break;
+		}
+	}
+
+	return isMacOSEnabled;
+};
+
 iOSModuleBuilder.prototype.run = function run(logger, config, cli, finished) {
 	Builder.prototype.run.apply(this, arguments);
 
@@ -107,7 +138,9 @@ iOSModuleBuilder.prototype.run = function run(logger, config, cli, finished) {
 		'compileJS',
 		'buildModule',
 		'createUniversalBinary',
-		'verifyBuildArch',
+		function (next) {
+			this.verifyBuildArch().then(next, next);
+		},
 		'packageModule',
 		'runModule',
 
@@ -447,6 +480,9 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 					this.logger.error('[' + type + '] ' + line);
 				}, this);
 				this.logger.log();
+				if (type === 'xcode-macos') {
+					this.logger.info(__('To exclude mac target, set/add mac: false in manifest file.'));
+				}
 				process.exit(1);
 			}
 
@@ -476,10 +512,43 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 			args.push('SUPPORTS_MACCATALYST=YES');
 		}
 
+		if (target === 'iphonesimulator') {
+			// Search for third party framewrok included in module. If found any, exclude arm64 from simulator build.
+			// Assumption is that simulator arm64  architecture can only be supported via .xcframework.
+			const frameworksPath = path.join(this.projectDir, 'platform');
+			const legacyFrameworks = new Set();
+
+			fs.readdirSync(frameworksPath).forEach(filename => {
+				if (filename.endsWith('.framework')) {
+					legacyFrameworks.add(filename);
+				}
+			});
+			if (legacyFrameworks.size > 0) {
+				const pbxFilePath = path.join(this.projectDir, `${this.moduleName}.xcodeproj`, 'project.pbxproj');
+				const proj = xcode.project(pbxFilePath).parseSync();
+				const configurations = proj.hash.project.objects.XCBuildConfiguration;
+				let excludeArchs = 'EXCLUDED_ARCHS=arm64 ';
+
+				// Merge with excluded archs in xcode setting
+				for (const key of Object.keys(proj.hash.project.objects.XCBuildConfiguration)) {
+					const configuration = configurations[key];
+					if (typeof configuration === 'object' && configuration.name === 'Release' && configuration.buildSettings.EXCLUDED_ARCHS) {
+						let archs = configuration.buildSettings.EXCLUDED_ARCHS;  // e.g. "i386 arm64 x86_64"
+						archs = archs.replace(/["]/g, '');
+						excludeArchs = excludeArchs.concat(archs);
+						break;
+					}
+				}
+
+				args.push(excludeArchs);
+				this.logger.warn(__(`The module is using frameworks (${Array.from(legacyFrameworks)}) that do not support simulator arm64. Excluding simulator arm64. The app using this module may fail if you're on an arm64 Apple Silicon device.`));
+			}
+		}
+
 		return args;
 	}.bind(this);
 
-	this.cli.env.getOSInfo(function (osInfo) {
+	this.cli.env.getOSInfo(osInfo => {
 		const macOsVersion = osInfo.osver;
 
 		// 1. Create a build for the simulator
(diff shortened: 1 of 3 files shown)


iphone/cli/commands/_buildModule.js
```diff
@@ -476,6 +483,25 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
...
+		if (target === 'iphonesimulator') {
+			// Search for third party framewrok included in module. If found any, exclude arm64 fromsimulator build.
+			// Assumption is that simulator arm64  architecture can only be supported via .xcframework.
+			const frameworkPattern = /([^/]+)\.framework$/;
+			let frameworksPath = path.join(this.projectDir, 'platform');
+			let legacyFrameworks = new Set();
+
+			fs.readdirSync(frameworksPath).forEach(function (filename) {
+				let possibleFrameworkPath = path.join(frameworksPath, filename);
+				if (frameworkPattern.test(possibleFrameworkPath)) {
```
sgtcoolguy, line 495: This was probably stolen from `frameworks.js` code, but I don't think we need a regex here or to build the path, just check for `filename.endsWith('.framework')`

iphone/cli/commands/_buildModule.js
```diff
@@ -476,6 +483,25 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 			args.push('SUPPORTS_MACCATALYST=YES');
 		}
 
+		if (target === 'iphonesimulator') {
+			// Search for third party framewrok included in module. If found any, exclude arm64 fromsimulator build.
+			// Assumption is that simulator arm64  architecture can only be supported via .xcframework.
+			const frameworkPattern = /([^/]+)\.framework$/;
+			let frameworksPath = path.join(this.projectDir, 'platform');
```
sgtcoolguy, line 490: this value isn't re-assigned, use `const`

iphone/cli/commands/_buildModule.js
```diff
@@ -476,6 +483,25 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
...
+			let legacyFrameworks = new Set();
```
sgtcoolguy, line 491: same, use `const`

iphone/cli/commands/_buildModule.js
```diff
@@ -476,6 +483,25 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 			args.push('SUPPORTS_MACCATALYST=YES');
 		}
 
+		if (target === 'iphonesimulator') {
+			// Search for third party framewrok included in module. If found any, exclude arm64 fromsimulator build.
+			// Assumption is that simulator arm64  architecture can only be supported via .xcframework.
+			const frameworkPattern = /([^/]+)\.framework$/;
```
sgtcoolguy, line 489: unnecessary, see below

iphone/cli/commands/_buildModule.js
```diff
@@ -447,7 +447,14 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 					this.logger.error('[' + type + '] ' + line);
 				}, this);
 				this.logger.log();
-				process.exit(1);
+				if (type === 'xcode-macos') {
```
sgtcoolguy, line 450: Feels like this should be handled by the caller (code at bottom of the file that launches these xcodebuilds). I'm fine with the change here since that code would get messy if we just bubbled up the exit code and had to do error handling there as it stands. But I think long-term we should move towards async/await in place of callbacks and do the handling there (so just bubble up the code as return value here, don't call process.exit here, etc and then it's a matter of try/catch down there)

sgtcoolguy (approved):
suggestions aren't required, but I think would clean the code up a little/make it simpler.

sgtcoolguy:
As per @hansemannn 's comments in the linked ticket, we may just want to use the Xcode build setting as the "flag" to determine if the module is intended to target macOS. We've basically been forcing `SUPPORTS_MACCATALYST` to `YES`. It may make more sense to just let them toggle it on/off themselves. Maybe we can default it to `YES` in our module template for new modules going forward?

Or are we worried that module developers just won't toggle it on, even though it may literally require no other work for them to get macOS support included?

vijaysingh-axway:
1. Regarding Han's solution,  The modules which are not generating frameworks (Objc based modules), do not have option to out out from mac-target build in Xcode project. This option is available for frameworks only. I think his PR will not work for all modules.

2. Regarding exclusion of target or arch - In Xcode 12, Apple has given EXCLUDED_ARCHS for excluding the architectures from build. Can we add new keys in manifest file to exclude architectures?
e.g.
exclude_mac_archs: arm64, x86_64, all
exclude_device_archs: arm64, armv7, all
exclude_sim_archs: arm64, i386, x86_64, all

all = do not build for particular target .
Default will be build for all architectures and targets.

Or will it be more complex ? cc @sgtcoolguy @janvennemann .

hansemannn:
Maybe just a `mac: true` flag in the manifest? Defaulting to `true`, easy to parse by the module script and highly backwards compatible.

hansemannn:
@vijaysingh-axway I tried your update and it still doesn't work:
```
[ERROR] There is discrepancy between the architectures specified in module manifest and compiled binary.
[ERROR] Architectures in manifest: armv7,arm64,i386,x86_64
[ERROR] Compiled binary architectures:
[ERROR] Please update manifest to match module binary architectures.
```
Did you try with Ti.WidgetKit?

janvennemann:
Just using the setting from the Xcode project would be ideal of course. But as Vijay already pointed out this is not possible in ObjC modules since the project is still setup to build static libraries and now framework package, so Xcode doesn't offer the same options as for Swift based native modules.

Maybe we can use the `mac: true` flag as suggested by Hans as an option that works for both, and fallback to check the Xcode project setting for modern Swift based modules. That way ObjC modules can still control it, but Swift modules don't need to hassle with the manifest file, especially since there already is an equivalent option in Xcode available.

vijaysingh-axway:
> @vijaysingh-axway I tried your update and it still doesn't work:
> 
> ```
> [ERROR] There is discrepancy between the architectures specified in module manifest and compiled binary.
> [ERROR] Architectures in manifest: armv7,arm64,i386,x86_64
> [ERROR] Compiled binary architectures:
> [ERROR] Please update manifest to match module binary architectures.
> ```
> 
> Did you try with Ti.WidgetKit?

@hansemannn It is failing because the architectures mentioned in manifest file is not same as in created XCFramework. Reason is, in _buildModule.js we are using hardcoded way to read architecture. Jan Vennemann is working on parsing of Info.plist available inside .xcframework as part of PR https://github.com/appcelerator/titanium_mobile/pull/12092 . After that we can update it to use architectures from there.

It reminds me, should we fail the module build if manifest architectures do not match the architectures available in xcframework? Any how arm64 is in device and simulator both and x86_64 is in mac and simulator both.  We can not verify exact architectures.
 I think in [verifyBuildArch](https://github.com/appcelerator/titanium_mobile/blob/41ebeecedf3661685bff8ad6626d30295aa488e6/iphone/cli/commands/_buildModule.js#L595) we should verify if it is xcframework and log the supported targets and corresponding architectures in module and proceed to package module. cc @janvennemann

hansemannn:
@vijaysingh-axway Thanks for the insight! So there is no workaround to compile the module? It's blocking our next update right now. Amy advice to work around it would help already!

hansemannn:
@vijaysingh-axway With the latest change, the module compiles successfully already! The app build then fails with the following error:
```
Error: ENOENT: no such file or directory, open '/Users/user/Documents/dev/apps/myapp/modules/iphone/ti.widgetkit/1.0.0/TiWidgetkit.xcframework/ios-i386_x86_64-simulator/TiWidgetkit.framework/TiWidgetkit'
    at Object.openSync (fs.js:462:3)
    at Object.readFileSync (fs.js:364:35)
    at iOSBuilder.<anonymous> (/Users/user/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200915121722/iphone/cli/commands/_build.js:2252:54)
    at Array.forEach (<anonymous>)
    at iOSBuilder.<anonymous> (/Users/user/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200915121722/iphone/cli/commands/_build.js:2176:20)
    at iOSBuilder.<anonymous> (/Users/user/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200915121722/node_modules/node-titanium-sdk/lib/builder.js:387:3)
    at Object.callback (/Users/user/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200915121722/node_modules/node-appc/lib/timodule.js:349:52)
    at /Users/user/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200915121722/node_modules/node-appc/lib/timodule.js:415:52 {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/Users/user/Documents/dev/apps/myapp/modules/iphone/ti.widgetkit/1.0.0/TiWidgetkit.xcframework/ios-i386_x86_64-simulator/TiWidgetkit.framework/TiWidgetkit'
}
```

I suppose that will be fixed by the changes from Jan then.

vijaysingh-axway:
@hansemannn  It should work now.

hansemannn:
@vijaysingh-axway The module build works, but the containing app cannot use the module:
```
Couldn't find module: ti.widgetkit for architecture: x86_64
```

iphone/cli/commands/_buildModule.js
```diff
@@ -894,8 +917,6 @@ iOSModuleBuilder.prototype.runModule = function runModule(next) {
 
 			// 5. unzip module to the tmp dir. Use native binary on macOS, as AdmZip doesn't support symlinks used in mac catalyst frameworks
 			const proc = spawn('unzip', [ '-o', this.moduleZipPath, '-d', tmpProjectDir ]);
-			proc.stdout.on('data', data => this.logger.info(data.toString().trimEnd()));
```
sgtcoolguy, line 897: without these lines, the unzip never finished for me. We have to "consume" the stdout/stderr for it to finish.

iphone/cli/commands/_buildModule.js
```diff
@@ -558,15 +628,17 @@ iOSModuleBuilder.prototype.createUniversalBinary = function createUniversalBinar
...
-	} else {
-		args.push(buildType);
-		args.push(lib);
-		if (!this.isFramework) {
-			args.push('-headers');
-			args.push(headerPath);
+	if (this.isMacOSEnabled) {
+		lib = findLib('macosx');
+		if (lib instanceof Error) {
+			this.logger.warn(__('The module is missing 64-bit support of macos. Ignoring mac target for this module...'));
```
sgtcoolguy, line 634: The warning is misleading. It's missing macOS support entirely, not 64-bit support.

iphone/cli/commands/_buildModule.js
```diff
@@ -488,8 +553,13 @@ iOSModuleBuilder.prototype.buildModule = function buildModule(next) {
 			xcodebuildHook(xcBuild, xcodeBuildArgumentsForTarget('iphoneos'), opts, 'xcode-dist', () => {
 				const osVersionParts = macOsVersion.split('.').map(n => parseInt(n));
 				if (osVersionParts[0] > 10 || (osVersionParts[0] === 10 && osVersionParts[1] >= 15)) {
-					// 3. Create a build for the maccatalyst
-					xcodebuildHook(xcBuild, xcodeBuildArgumentsForTarget('macosx'), opts, 'xcode-macos', next);
+					// 3. Create a build for the mac-catalyst if enabled
+					if (this.isMacOSEnabled) {
```
sgtcoolguy, line 557: Kind of a dumb edge-case, but if they have no indication of Mac support in either a positive or negative value, what should we do? I think this will assume no macOS support, so will not try to build for it.

 But if they haven't set `SUPPORTS_MACCATALYST` in Xcode and they have no `mac` key/value pair in the `manifest`, then technically they haven't indicated one way or the other. Would we *try* to build for Mac and just be ok with it failing in that case?

iphone/cli/commands/_buildModule.js
sgtcoolguy: Also, they've "turned on" Mac support, so shouldn't this be an Error, not a warning log and move on?

sgtcoolguy:
@vijaysingh-axway  I'm testing locally with WidgetKit and Facebook and cleaning up the commits. I'll force push and merge (if build is successful) afterwards.

sgtcoolguy:
OK, I've force-pushed a rebased set of commits that worked for me locally for both @hansemannn WidgetKit module and our Facebook Mac support PR. Note that on Hans' module I had to edit the manifest to remove `armv7` and `i386` from the architectures (as it does not build for them).

For Facebook, I basically just had to point to the locally built SDK and set the manifest's `mac: false`

hansemannn:
@sgtcoolguy I wonder what's wrong with my config. I pulled the latest 9.2.0 build, updated the architectures and `$(TITANIUM_SDK)` path, but still get (a now different) error:
```
[TRACE] [xcode-macos] error: While building for macOS, no library for this platform was found in '~/Library/Application Support/Titanium/mobilesdk/osx/9.2.0.v20200922084018/iphone/Frameworks/TitaniumKit.xcframework'. (in target 'TiWidgetkit' from project 'titanium-widget-kit')
```

hansemannn:
Update: It seems like the latest 9.2.0 build is still not available, but building from source finally fixes the issue. Thanks everyone!

hansemannn:
Are there plans to finally fix all iOS 14 module issues and therefore make Titanium compatible with iOS 14+ modules? I filed [this ticket](https://jira.appcelerator.org/browse/TIMOB-28155) to track it, but it may have been overseen so far :)

PR #12094, opened 2020-09-17
fix(ios): remove old tiverify.xcframework references from project

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28099

**Optional Description:**
The node `xcode` module does not properly handle `.xcframework` packages yet. Through a specific set of options it can be tricked into choosing the proper internal options so that `removeFramework` still works. This works fine for `TitaniumKit.xcframework` but for `tiverify.xcframework` this left on orphaned entry in the frameworks build-phase. Hyperloop bails on this orphaned references so i had to update the options to properly remove `tiverify.xcframework` as well.


iphone/cli/commands/_build.js
```diff
@@ -3946,7 +3946,9 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 	});
 
 	// add tiverify.framework
-	xcodeProject.removeFramework('tiverify.xcframework', { customFramework: true, embed: true });
+	// technically the `lastKNownFileType` option is wrong, but the xcode module
+	// does not properly handle `.xcframework` extensions without it yet
+	xcodeProject.removeFramework('tiverify.xcframework', { lastKnownFileType: 'wrapper.framework' });
```
sgtcoolguy, line 3951: So the value for `lastKnownFileType` looks incorrect to me, wasn't sure if the comment above is what you meant? Shouldn't it be `'wrapper.xcframework'`?

Here's the line in the pbxproj file that it pertains to:
```
B112F935244E69B1005901DA /* tiverify.xcframework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcframework; name = tiverify.xcframework; path = ../lib/tiverify.xcframework; sourceTree = "<group>"; };
```

janvennemann: Yes, the comment is exactly for that. Node `xcode` doesn't know how to handle `.xcframework` and fails to determine the correct parameters that are required to remove all bits in the Xcode project. Incorrectly specifying the `lastKnownFileType` fixes that.

sgtcoolguy (approved):

ssjsamir (approved):
FR Passed, Able to build Hyperloop examples Project. 

**Test Environment**
```
MacOS Big Sur: 11.0 Beta 6
Xcode: 12.0 
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
iphone 8 Sim (14.0 Beta)
```

PR #12107, opened 2020-09-20
feat(ios): support iOS 14+ App Clips

hansemannn:
**JIRA:**  https://jira.appcelerator.org/browse/AC-6578


sgtcoolguy:
LGTM but @vijaysingh-axway  and @janvennemann  should review/weigh in

vijaysingh-axway:
In my initial testing I found- 
1. In cli we do not have any command from which developer can  launch AppClip in device, to verify the changes in app clip target. But they can run app clip from generated Xcode project. (I think this should be okey. We can create a cli ticket for same which can be done later).

2. Build failed due to Bundle version mismatch from app clip. For this use same version (available in tiapp.xml ) in created native app clip(go to app clip target->general- > Build) cc QE.

vijaysingh-axway (approved):
Changes looks good. 
The added comments in PR are basically to ease the testing and may be add in release doc somewhere.

lokeshchdhry:
FR Passed.

PR #12116, opened 2020-09-22
chore(deps): update facebook module to 9.0.0 for ios, 10.0.0 for android

sgtcoolguy:

**Description:**
See appcelerator-modules/ti.facebook#128

Updates the Facebook module for iOS and Android. Biggest change is iOS goes from Facebook SDK 6.0.0 -> 7.1.1; Android goes from 5.13.0 -> 7.1.0. iOS now built as an xcframework (does not include arm64 sim support or macOS support due to limitations in facebooks SDK framework)


lokeshchdhry:
FR Passed.
Both modules seem to work fine on IOS & Android.

PR #11758, opened 2020-06-05
fix(ios): TableViewRow does not return getRect methods

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-27935

PR #12128, opened 2020-09-23
fix(ios): check for selector availability

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28148


iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewProxy.m
```diff
@@ -213,7 +213,9 @@ - (void)add:(id)arg
 
   if (windowOpened) {
     // Turn on clipping because I have children
-    [[self view] updateClipping];
+    if ([[self view] respondsToSelector:@selector(updateClipping)]) {
+      [[self view] updateClipping];
+    }
```
hansemannn, line 218: I'm pretty sure this does not fix the real issue. Primarily, because this used to work before and the error suggests that a wrong instance is bubbled up in the event

janvennemann: @vijaysingh-axway I think Hans is right here. I did a quick debug session and it appears that `self.view` is holding a wrong value. Xcode debugger says it's `nil` so it may have been accidentally released and now holds an invalid pointer. Can you please double check what's going on?

janvennemann (changes requested):
See comment above.

sgtcoolguy:
Looking at diffs between 9.0.3 and 9.1.0, this change related to TIMOB-27935 looks suspicious: https://github.com/appcelerator/titanium_mobile/commit/b15d1840c9be3cbf7cc74b10381e3656846f87b8

I think this stems because now a row proxy returns the `TiUITableViewCell` whereas before it returned `nil`.

vijaysingh-axway:
To fix an issue (kind of parity)  https://jira.appcelerator.org/browse/TIMOB-27935, as Chris has mentioned, I have returned the cell view associated with row proxy which is not TiUIView. This is causing the crash. Initially it was nil so there was not issue. So-
In current fix I have checked for availability of selector then only call it. I think there should not be any related issue after this fix.

sgtcoolguy:
So, this may be a quick and dirty fix for this issue, but it exposes the fact that the earlier fix broke the `[TiViewProxy view]` return type contract of returning a `TiUIView*`. It no longer returns a `TiUIView*`, it returns a subclass of `UIView*`. The better long-term "fix" here would be if we could separate callers that needed a `TiUIView*` from those needing a `UIView*`. i.e. the code around calculating the `rect` property actually only needs a `UIView*`, but the `size` property needs a `TiUIView*` (which shows that the earlier fix only fixed the rect property, not size!) - and then have different methods to return either?

Or maybe we can extract some common protocol used by callers and ensure `TiUITableViewCell` and `TiUIView` adhere to it?

vijaysingh-axway:
I am thinking to revert the changes of PR https://github.com/appcelerator/titanium_mobile/pull/11758  and reopen https://jira.appcelerator.org/browse/TIMOB-27935 to properly fix.

vijaysingh-axway:
Reopened https://jira.appcelerator.org/browse/TIMOB-27935 to look in this again.

janvennemann (approved):

ssjsamir (approved):
FR Passed: Tested with the following test case: 

```
var tableData = [];
   
var win = Ti.UI.createWindow({ backgroundColor: 'white' });
 
var table = Ti.UI.createTableView({ objName: 'table' });
 
for (var i = 0; i <= 20; i++) {
var row = Ti.UI.createTableViewRow({ title: 'click me' });
 
var view = Ti.UI.createView({
layout: 'vertical',
height: Ti.UI.SIZE,
touchEnabled: false,
width: "50%",
left: 0
});
row.add(view)
tableData.push(row);
}
 
table.setData(tableData);
 
table.addEventListener('click', function (e) {
if (e.source && e.source.objName !== 'table') {
e.source.add(Ti.UI.createLabel({
text: 'REQUIRED',
left: '60%',
color: 'blue',
horizontalWrap: true,
width: "40%",
height: Ti.UI.SIZE
}));
 
}
});
 
win.add(table);
win.open();
```



**Test Environment**
```
MacOS Big Sur: 11.0 Beta 9
Xcode: 12.2 beta 2
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
iphone 11 (ios 14)
```

sgtcoolguy (dismissed):
I get that the rollback off the other fix "fixes" this issue. But it also re-opens a different crash that affected a customer. We need to address that as well - we cannot merge this in and be done with it, because we're just trading one crash for another

sgtcoolguy:
OK, I was mistaken. The fix for TIMOB-27935 caused this bug and TIMOB-28001 to occur - not that the fix for TIMOB-27935 also fixed TIMOB-28001
So, rolling back really only re-opens https://jira.appcelerator.org/browse/TIMOB-27935 which is  an issue found internally and does not seem to be as severe as this bug or TIMOB-28001 (both crashes).

PR #12129, opened 2020-09-24
fix(ios): guard PHPicker usage with openPhotoGallery

ewanharris:
Fixes TIMOB-28151

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28151

Guard the usage of PHPicker with `defined(USE_TI_MEDIAOPENPHOTOGALLERY)` otherwise 💥

### DIFF
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -1784,7 +1784,7 @@ MAKE_SYSTEM_PROP(VIDEO_REPEAT_MODE_ONE, VideoRepeatModeOne);
 }
 #endif
 
-#if IS_SDK_IOS_14
+#if IS_SDK_IOS_14 && defined(USE_TI_MEDIAOPENPHOTOGALLERY)
 - (void)showPHPicker:(NSDictionary *)args
 {
   if (_phPicker != nil) {


iphone/Classes/MediaModule.m
```diff
@@ -1784,7 +1784,7 @@ - (void)handleTrimmedVideo:(NSURL *)theURL withDictionary:(NSDictionary *)dictio
 }
 #endif
 
-#if IS_SDK_IOS_14
+#if IS_SDK_IOS_14 && defined(USE_TI_MEDIAOPENPHOTOGALLERY)
```
hansemannn, line 1787: You _may_ also need to check it's delegate methods, but not sure if they're also affected.

sgtcoolguy: delegate methods appear to already be guarded

sgtcoolguy (approved):
LGTM, was able to reproduce locally and confirm the fix.

sgtcoolguy:
To reproduce:
``` js
Ti.Media.beep();
window.open();
```

Then you *must* build for device: `ti build -p ios -T device`

PR #12146, opened 2020-10-01
fix(ios): ignore duplicate and unnecessary framework search paths

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28158

This addresses two issues with framework search paths:

We used `appc.util.mix` to assign default build settings to the `Debug` and `Release` build configurations in the Xcode project. This would assign the `FRAMEWORK_SEARCH_PATHS` by references so they both used the same array which caused duplicate entries. Switching to `lodash.merge` fixes this since it merges arrays recursively by value.

In #12026 recursive search paths were added to handle the nested structure of XCFrameworks. This causes Xcode to construct huge framework search path arguments during the build since it would expand to every subfolder inside native modules. This [article](https://github.com/msaps/XCFramework-Link-Issue#explicit-architecture-framework-search-paths) suggests that this is unnecessary since Xcode has a special build phase that automatically scans XCFrameworks and automatically selects the appropriate .framework for the current architecture without consulting `FRAMEWORK_SEARCH_PATHS`. This PR excludes adding recursive framework search paths to the build settings. I've tested it with `ti.map` and `hyperloop` and i was still able to build without issues so it indeed seems that the recursive search paths are not required.


janvennemann:
@sgtcoolguy I think this might also be worth considering for 9.2.1

sgtcoolguy (approved):
LGTM

lokeshchdhry:
FR Passed.

ssaddique:
FR #2 passed. KitchenSink project builds as normal.

**Test environment**
SDK Ver:        9.2.1.v20201002104158
OS Ver:         10.15.3
Xcode Ver:      Xcode 11.6
Appc NPM:       5.0.0
Appc CLI:       8.1.1
Ti CLI Ver:     5.2.5
Node Ver:       10.17.0
NPM Ver:        6.13.6
Java Ver:       1.8.0_162
Emulator:       iPhone 11 Pro (13.6)

PR #12153, opened 2020-10-02
fix(ios): append TitaniumKit.xcframework if module is created with sdk < 9.2.0

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28159

### DIFF
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -28,6 +28,7 @@ const appc = require('node-appc'),
 	path = require('path'),
 	temp = require('temp'),
 	util = require('util'),
+	xcodeParser = require('xcode/lib/parser/pbxproj'),
 	__ = appc.i18n(__dirname).__,
 	series = appc.async.series,
 	xcode = require('xcode');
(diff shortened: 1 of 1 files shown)


lokeshchdhry:
Works for me. Want some more eyes on this from other QE's.

iphone/cli/commands/_buildModule.js
```diff
@@ -237,6 +252,60 @@ iOSModuleBuilder.prototype.processLicense = function processLicense() {
 };
 
 iOSModuleBuilder.prototype.processTiXcconfig = function processTiXcconfig(next) {
+	const srcFile = path.join(this.projectDir, this.moduleName + '.xcodeproj', 'project.pbxproj');
+	const xcodeProject = xcode.project(path.join(this.projectDir, this.moduleName + '.xcodeproj', 'project.pbxproj'));
+	const contents = fs.readFileSync(srcFile).toString();
```
sgtcoolguy, line 257: FYI, when reading a file, you can avoid converting from `Buffer` to `String` (calling `toString()`), by passing in an encoding argument to the readFile call: `fs.readFileSync(srcFile, 'utf8');`

iphone/cli/commands/_buildModule.js
```diff
@@ -237,6 +252,60 @@ iOSModuleBuilder.prototype.processLicense = function processLicense() {
...
+	const appName = this.moduleIdAsIdentifier;
+
+	if (this.isFramework && !contents.includes('TitaniumKit.xcframework')) {
+		this.logger.warn(`Module created with sdk < 9.2.0, need to add TitaniumKit.xcframework. The ${this.moduleName}.xcodeproj has been updated.`);
+
+		xcodeProject.hash = xcodeParser.parse(contents);
+		const xobjs = xcodeProject.hash.project.objects,
```
sgtcoolguy, line 264: Just a style thing in JS - in the old days you should use `var` and declare all variables at the top of the scope, because JS hoisted them to the top of the scope anyways, so we wanted to match in source to avoid inconsistency. But with `let`/`const`, they're more like variables in any other language, they're more or less treated as declared/initialized where you define them, so you don't need to stick them at the top of the block/scope. Ideally we'd break up this big block of declared variables and move them to be declared where they're needed to make it easier to follow the logic and which grouping of variables applies to what lines of code used to edit/process them.

This is a long way of saying, most of these variables aren't needed until lines 294/299 below.

sgtcoolguy (approved):
Generally LGTM.

For older modules, do they have a TitaniumKit.framework reference that needs to be removed/replaced by the xcframework reference?

PR #12155, opened 2020-10-05
fix(ios): convert native stack to array and print native error reason

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28160

This addresses a couple more issues i found in our error reporting on iOS.

**Include compete native exception message**

Adds `nativeReason` to error message from native exceptions. For example, consider the following test case:

```js
Ti.UI.createView({
  top: 20,
  backgroundGradient: {
      type: 'invalid'
  }
});
```

This used to only generate the message `Invalid type passed to function`. Now it also prints the native reason making it a little more clear: `Invalid type passed to function. Must be either 'linear' or 'radial'`.

**Include complete native stack**

This brings us directly to the next issue. By default we used to omit the first 4 entries of the native stack of being included in the error. In the above example that would remove the important hint of `-[TiGradient setType:]` being the method that caused the exception. I guess this was done to prevent printing of unrelated code paths in our error handling, but i found that we cannot reliably determine how many lines exactly are not relevant and it appears to change depending on where the exception was thrown. I changed this to include the complete native stack trace to be sure that everything that may be relevant to the crash is included.

**Ensure correct type for `nativeStack`**

The last issue addressed here is that we have different types for `nativeStack` being used when creating `TiScriptError`. Most times it's an array of strings (which is the type declared for the property), but when converting a native exception to it's JS representation we manually convert the array to a string which then caused a follow up exception while handling the original error. I wasn't sure if anything else like ACA depends on it being a string, so i only adjusted the method in `TiUtils` which converts the `JSValueRef` into a `TiScriptError`. It seems there was already a fix in ACA which also addressed this in [MOD-2552](https://jira.appcelerator.org/browse/MOD-2552) cc @garymathews


iphone/TitaniumKit/TitaniumKit/Sources/API/TiBindingTiValue.m
```diff
@@ -292,7 +292,7 @@ JSValueRef TiBindingTiValueFromNSObject(JSContextRef jsContext, NSObject *obj)
 
     // Add "nativeStack" key
     NSArray<NSString *> *nativeStack = [(NSException *)obj callStackSymbols];
-    NSInteger startIndex = 3; // Starting at index = 4 to not include the script-error API's
+    NSInteger startIndex = 0;
```
sgtcoolguy, line 295: Why not just drop this variable then?

sgtcoolguy:
@janvennemann Can we add the snippet shown above as a unit tests and verify some of these fixes officially?

iphone/TitaniumKit/TitaniumKit/Sources/API/TiBindingTiValue.m
janvennemann: Because in the if guard below we manually retrieve the native call stack, and in that case we know for sure that we can drop the first two entries.

janvennemann:
@sgtcoolguy Sure, I'll add some unit tests!

janvennemann:
Added two new tests to check for the newly constructed error message that includes the native error reason. The test to check for the correct native error stack is currently failing, since the stack contains a wrong line. I need to investigate what's going on there.

This is how stack looks like when running Titanium directly from Xcode project inside the SDK:

```js
0 CoreFoundation 0x00007fff2041ca5e __exceptionPreprocess + 242
1 libobjc.A.dylib 0x00007fff20177e78 objc_exception_throw + 48
2 TitaniumKit 0x000000010a2041ea TiExceptionThrowWithNameAndReason + 138
3 TitaniumKit 0x000000010a2932bb -[TiProxy throwException:subreason:location:] + 107
4 TitaniumKit 0x000000010a1f85a9 -[TiGradient setType:] + 745
5 Foundation 0x00007fff207a4f9a -[NSObject(NSKeyValueCoding) setValue:forKey:] + 325
6 TitaniumKit 0x000000010a2931a1 -[TiProxy setValuesForKeysWithDictionary:] + 2817
...
```

And this when run via CLI after building the SDK.

```
0 CoreFoundation 0x00007fff2041ca5e __exceptionPreprocess + 242
1 libobjc.A.dylib 0x00007fff20177e78 objc_exception_throw + 48
2 TitaniumKit 0x0000000108fe0817 JavascriptNameForClass + 0
3 Foundation 0x00007fff207a4f9a -[NSObject(NSKeyValueCoding) setValue:forKey:] + 325
4 TitaniumKit 0x0000000109034c22 -[TiProxy setValuesForKeysWithDictionary:] + 1886
...
```

sgtcoolguy:
ping @janvennemann - This ticket is linked to 9.3.0 release in JIRA and it looks like the unit tests here are failing...

sgtcoolguy:
@janvennemann I don't know why, but the call stack embedded in `nativeStack` differs for me depending on whether I run under Xcode or via the SDK test suite. If I run via Xcode, I do see the `[TiGradient setType:]`, but not when I run via the SDK test suite, even if I change the code not to trip any of the call stack at all. I assume it may be a debug vs non-debug symbolication thing?

garymathews (approved):
I've set the strip style in _Titanium.xcodeproj_ to `debugging` (default was `all`) in an attempt to prevent inconsistencies with stack traces.

Also updated the test cases, which are now passing.

PR #12163, opened 2020-10-07
feat(ios): redesigned error view

janvennemann:
This started out as a quick look at Swift UI, but i really liked the look and feel and additional information this error screen provides so i thought this might be something interesting for a future release. It requires a minimum deployments target of iOS 11, so probably something for 10.0

On devices running iOS 13 and up this will use a new redesigned error view to display all relevant error details in a clean and structured UI. This should help users to get a better grip of the error and immediately see what went wrong.

Here are two examples for the following code snippet:

```js
Ti.UI.createView({
    top: 20,
    backgroundGradient: {
        type: 'invalid'
    }
})
```

Light

Dark

cc @jquick-axway and @garymathews for possible Android parity


vijaysingh-axway:
Looks nice. We can consider it for 10.0.0.

hansemannn:
@janvennemann Would you mind rebasing this PR? Happy to take it for 10.2.0!

hansemannn (approved):

PR #12167, opened 2020-10-08
fix(ios): fix "dateTimeColor“ for iOS 14+

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28181


vijaysingh-axway (changes requested):
'DatePicker dateTimeColor (invalid "type" - TIMOB-28181)' unit test is still crashing.

hansemannn:
@vijaysingh-axway Actually, we have to always disallow the `dateTimeColor` for non-date-pickers, not just on iOS 14+. I updated the test case to reflect that

iphone/Classes/TiUIPicker.m
```diff
@@ -232,6 +232,11 @@ - (void)setBackgroundColor_:(id)value
 
 - (void)setDateTimeColor_:(id)value
 {
+  // Guard date picker and iOS 14+ date picker style
+  if (![self isDatePicker] || [(UIDatePicker *)[self picker] preferredDatePickerStyle] != UIDatePickerStyleWheels) {
+    return;
+  }
```
vijaysingh-axway, line 238: @hansemannn  One minor change required, otherwise it will crash for Xcode < 11.4
```suggestion
  if (![self isDatePicker]) {
    return;
  }
#if IS_SDK_IOS_13_4
  if ((UIDatePicker *)[self picker].preferredDatePickerStyle != UIDatePickerStyleWheels) {
    return;
  }
#endif
```

vijaysingh-axway (approved):
CR passed.

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 9.3.0.v20201012172519
Appc CLI: 8.1.1
JDK: 11.0.4
Node: 10.17.0
Studio: 6.0.0.202005141803
Xcode: 12.0.1
Device: iPhone 7Plus(v14.0), iOS simulator 13.4

PR #12181, opened 2020-10-15
fix(template): update dependencies of angular template

janvennemann:
JIRA: https://jira.appcelerator.org/browse/TIMOB-28200

- Updated `zone.js` to fix type error
- Updated eslint parser / plugins and eslint config


sgtcoolguy (approved):

templates/app/angular-default/template/.eslintrc.json
```diff
@@ -0,0 +1,104 @@
+/**
```
sgtcoolguy, line 1: I assume eslint can handle json with comments like this file? Any reason why not just stick with JS extension so comments are "officially" supported?

templates/app/angular-default/template/.eslintrc.json
```diff
@@ -0,0 +1,104 @@
...
+     */
+    {
+      "files": ["*.ts"],
+      "parserOptions": {
+        "project": [
+          "tsconfig.*?.json"
+        ],
+        "createDefaultProgram": true
+      },
+      "extends": ["plugin:@angular-eslint/recommended"],
```
sgtcoolguy, line 31: should this be top-level, or does this only apply for typescript files (i.e. what rules should apply for JS files - or are they not supported in angular projects?)

templates/app/angular-default/template/.eslintrc.json
janvennemann: I wanted to stick with the behavior in a web based Angular project where the [angular-eslint](https://github.com/angular-eslint/angular-eslint#step-2---add-new-eslint-related-configuration) command also adds this exact file.

templates/app/angular-default/template/.eslintrc.json
janvennemann: Generally you don't use JS in an Angular project which is why everything is tailored around TS. I didn't wrote this configuration myself, it is the recommended default config from [angular-eslint](https://github.com/angular-eslint/angular-eslint). I'd say if linting of JS files is required users can always add their own config targeted for `* .js` files.

PR #12182, opened 2020-10-15
fix(android): use of right property in child of horizontal layout

garymathews:
- Correct behavior of `right` property when applied to last child in `horizontal` layout for parity with iOS

##### TEST CASE
```JS
	backgroundColor: 'gray'
});
const row = Ti.UI.createView({
	layout: 'horizontal',
	backgroundColor: 'white',
	width: Ti.UI.FILL,
	height: 50
});
const view_a = Ti.UI.createView({
	backgroundColor: 'green',
	left: 50,
	// right: 50,
	width: 50,
	height: 50
});
const view_b = Ti.UI.createView({
	backgroundColor: 'red',
	// left: 50,
	right: 50,
	width: 50,
	height: 50
});

row.add([ view_a, view_b ]);
win.add(row);
win.open();
```
- Last child of horizontal layout should align to the right.
- In all other cases (where both `left` and `right` properties are defined), view should stack from left with expected padding.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28197)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
@@ -1004,13 +1004,17 @@ public class TiCompositeLayout extends ViewGroup implements OnHierarchyChangeLis
 
 		TiDimension optionLeft = params.optionLeft;
 		TiDimension optionRight = params.optionRight;
+		boolean lastChild = currentIndex == getChildCount() - 1;
 		int left = horizontalLayoutCurrentLeft + horiztonalLayoutPreviousRight;
+		horiztonalLayoutPreviousRight = (optionRight == null) ? 0 : optionRight.getAsPixels(this);
 		int optionLeftValue = 0;
 		if (optionLeft != null) {
 			optionLeftValue = optionLeft.getAsPixels(this);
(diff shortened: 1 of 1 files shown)


sgtcoolguy (changes requested):
I think this calls for a simple snapshot test using the snippet from the PR body (although to avoid having to generate different variants by density, you should use pixel sizes (instead of numbers without units).

jquick-axway:
So, I don't think we should make this change.

It turns out Android and iOS' behavior (before this PR) is already consistent. In a "horizontal" layout, setting the "right" property to right-align a view within a container is not supported on Android or iOS. Likewise, you cannot do the same with a "vertical" layout and use the "bottom" property to bottom align the view on Android and iOS.

See the following scripts I attached to [TIMOB-28197](https://jira.appcelerator.org/browse/TIMOB-28197).
- `HorizontalPinAlignTest.js`
- `VerticalPinAlignTest.js`

Neither of the above scripts will do what the ticket expects on Android or iOS, but that's not necessarily wrong either. You can easily do this with a "composite" layout. I'm leaning towards not changing the layout behavior.

_(I know this PR is trying to resolve a ListView issue I found last week. We might need to take a different approach.)_

PR #12183, opened 2020-10-15
build: remove fake titanium-sdk node_module used by titanium CLI tisdk3fixes hook

sgtcoolguy:
**Description:**
Relates to this hook (which likely should be removed!): https://github.com/appcelerator/titanium/blob/master/hooks/tisdk3fixes.js

As mentioned in Teams chat:
We've long injected this fake `titanium-sdk` node module that basically redirects `node-titanium-sdk`'s `lib/titanium.js` - and specifically the `validateCorrectSDK` method.

I think we no longer need this, even though the hook for super old SDK versions is still around. Because it appears when @cb1kenobi re-added that hook back to fix forking ti build, he added checking for node-titanium-sdk first.


ewanharris (approved):
Tested locally and this looks good, forking still works ok for build and clean commands

PR #12187, opened 2020-10-17
fix(ios): fix MediaModule compile error

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28202

### DIFF
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -2016,7 +2016,7 @@ MAKE_SYSTEM_PROP(VIDEO_REPEAT_MODE_ONE, VideoRepeatModeOne);
 - (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController
 {
 #if defined(USE_TI_MEDIASHOWCAMERA) || defined(USE_TI_MEDIAOPENPHOTOGALLERY) || defined(USE_TI_MEDIASTARTVIDEOEDITING)
-#if IS_SDK_IOS_14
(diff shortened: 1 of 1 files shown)


iphone/Classes/MediaModule.m
```diff
@@ -2016,7 +2016,7 @@ - (void)popoverPresentationControllerDidDismissPopover:(UIPopoverPresentationCon
 - (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController
 {
 #if defined(USE_TI_MEDIASHOWCAMERA) || defined(USE_TI_MEDIAOPENPHOTOGALLERY) || defined(USE_TI_MEDIASTARTVIDEOEDITING)
-#if IS_SDK_IOS_14
+#if IS_SDK_IOS_14 && defined(USE_TI_MEDIAOPENPHOTOGALLERY)
```
sgtcoolguy, line 2019: Isn't this covered by the line above?

sgtcoolguy (changes requested):

iphone/Classes/MediaModule.m
hansemannn: Yes, but it also goes into the statement if I only use `Ti.Media.showCamera` or `Ti.Media.startVideoEditing`. So it throws the compile error once we only the above, but not `Ti.Media.openPhotoGallery`

sgtcoolguy: Ah, gotcha. If one of the other OR segments passes, we don't guard properly. OK!

sgtcoolguy (approved):

PR #12029, opened 2020-09-10
impr(android): Use RecyclerView for ListView and TableView

garymathews:
- Re-write`Ti.UI.ListView` to use `RecyclerView`
- Re-write `Ti.UI.TableView` to use `RecyclerView`

###### Updated 03/12/21 to include more test cases.
<details>
<summary>SHOW</summary>
<p>

```JS
const win = Ti.UI.createWindow({
    backgroundColor: 'white'
});
const tableViewSection = Ti.UI.createListSection({
    headerTitle: 'TableView'
});
const listViewSection = Ti.UI.createListSection({
    headerTitle: 'ListView'
});
const listView = Ti.UI.createListView({
    separatorStyle: Ti.UI.TABLE_VIEW_SEPARATOR_STYLE_SINGLE_LINE,
    separatorColor: 'black',
    templates: {
        template: {
            properties: {
                layout: 'vertical'
            },
            childTemplates: [{
                    type: 'Ti.UI.Label',
                    bindId: 'title',
                    properties: {
                        top: 5,
                        left: 5,
                        right: 5,
                        font: {
                            fontSize: 18
                        },
                        color: 'black'
                    }
                },
                {
                    type: 'Ti.UI.Label',
                    bindId: 'description',
                    properties: {
                        bottom: 5,
                        left: 5,
                        right: 5,
                        font: {
                            fontSize: 14
                        },
                        color: 'black'
                    }
                },
            ]
        }
    },
    defaultItemTemplate: 'template',
    sections: [tableViewSection, listViewSection]
});
listView.addEventListener('itemclick', e => {
    const item = e.section.items[e.itemIndex];
    if (item.properties.test instanceof Function) {
        item.properties.test();
    }
});
[
    // TableView
    {
        section: tableViewSection,
        title: 'Basic Test #1',
        description: 'headerView and footerView validation.',
        test() {
            const win = Ti.UI.createWindow({
                backgroundColor: 'gray'
            });
            const headerView = Ti.UI.createView({
                backgroundColor: 'red',
                height: '80px'
            });
            const footerView = Ti.UI.createView({
                backgroundColor: 'blue',
                height: '80px'
            });
            const tableView = Ti.UI.createTableView({
                headerView,
                footerView,
                backgroundColor: 'white'
            });
            tableView.setData([{
                title: 'Row',
                color: 'white'
            }]);
            win.add(tableView);
            win.open();
        }
    },
    {
        section: tableViewSection,
        title: 'Basic Test #2',
[...]

PR #12193, opened 2020-10-20
fix(ios): TableView "sectionCount" property crash

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-13903

**Unit Test Note:**
RecylerView PR #12029 is making heavy changes to the `ti.ui.tableview.test.js` script. In order to avoid a huge amount of conflicts, we should re-enable all `iosBroken` unit tests via that PR instead of this one.
_(I've already verified this PR's change fixes the unit tests.)_

**Test:**
2. Verify an alert displays "sectionCount: 2" on app startup.

```javascript
var section1 = Ti.UI.createTableViewSection({ headerTitle: "Colors" });
section1.add(Ti.UI.createTableViewRow({ title: "Red" }));
section1.add(Ti.UI.createTableViewRow({ title: "Green" }));
section1.add(Ti.UI.createTableViewRow({ title: "Blue" }));

var section2 = Ti.UI.createTableViewSection({ headerTitle: "Fruit" });
section2.add(Ti.UI.createTableViewRow({ title: "Apples" }));
section2.add(Ti.UI.createTableViewRow({ title: "Bananas" }));
section2.add(Ti.UI.createTableViewRow({ title: "Peaches" }));

	data: [section1, section2],
});
window.add(tableView);
window.addEventListener("open", function() {
	alert("sectionCount: " + tableView.sectionCount);
});
window.open();
```


iphone/Classes/TiUITableViewProxy.m
```diff
@@ -882,9 +882,9 @@ - (void)willShow
   [(TiUITableView *)[self view] refreshSearchControllerUsingReload:YES];
 }
 
-- (NSUInteger)sectionCount
+- (NSNumber *)sectionCount
 { //TODO: Shouldn't this be in the main thread, too?
-  return [sections count];
+  return NUMUINTEGER((sections != nil) ? sections.count : 0);
```
sgtcoolguy, line 887: I think this is the only required change, and the rest of the lines/changes are not necessary. Why not just leave this method as returning `NSUInteger` as before and avoid the need to explicitly call `[NSNumber unsignedIntegerValue]` to get one in the above changes?

sgtcoolguy (changes requested):

iphone/Classes/TiUITableViewProxy.m
jquick-axway: It wasn't crashing because `sections` was `nil`.
_(That's just a safety mechanism I added just in case.)_

It was crashing because the calling code [here](https://github.com/appcelerator/titanium_mobile/blob/a292c8619be9904081712003e3872d446afb02a3/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollObject.m#L850) expected the property's return type to be an `NSObject` derived type, but it received an `NSUInteger` which is a typedef'ed unsigned int. That's why I changed it to `NSNumber`, which is the same return type our `TiUIListViewProxy` uses for `sectionCount` [here](https://github.com/appcelerator/titanium_mobile/blob/a292c8619be9904081712003e3872d446afb02a3/iphone/Classes/TiUIListViewProxy.h#L15).

sgtcoolguy (approved):

ssaddique:
**FR: Pass**

**Test Environment**
SDK Ver:        this PR
OS Ver:         10.15.6
Xcode Ver:      Xcode 12.0.1
Appc NPM:       5.0.0
Appc CLI:       8.1.1
Ti CLI Ver:     5.2.5
Alloy Ver:      1.15.2
Node Ver:       10.19.0
NPM Ver:        6.13.6
Emulator:       iPhone 8 (13.5), iPhone 11 Pro (14.0)

PR #12194, opened 2020-10-20
fix(cli/clean): set prompt on project dir option

ewanharris:
This means that the prompt now gets completions for the filesystem and the titanium cli can handle
the graceful shutdown correctly


cb1kenobi (approved):
Looks good! **APPROVED**!

PR #12201, opened 2020-10-21
refactor: remove adm-zip

sgtcoolguy:
**Description:**
This removes `adm-zip` as a direct dependency. It's still hidden in the dependency stack right now, but this removes direct usage in favor of `node-appc`'s `zip.unzip` method that's backed by `yauzl`. Also the one case of directly manipulating a JAR file to read an entry now uses `yauzl`. Note that this then removes our internal usage of native `unzip` binaries to unzip during SDK build/packaging as well.

Eventually we can hopefully get the newer node-appc usage throughout the full dependency tree here and help slim our `node_modules`

Lastly, I removed `css-parse` as a dependent. I couldn't find any usage.

Relates to:
- appcelerator/titanium#392
- appcelerator/node-appc#283


jquick-axway:
FYI: We can test the Android app build script's `getNativeModuleBindings()` changes using a module built with Titanium 8.x.x or older. _(Titanium 9.0.0 built modules never embed the JSON and makes the file externally available on the filesystem.)_

Test:
1. Add module "ti.map" version `4.5.0` to "tiapp.xml".
2. Add `require('t.map');` to "app.js".
3. Build and run on Android.
4. Verify that the `require` does not throw an exception.

ewanharris (approved):
CR/FR LGTM!

Tested Android and iOS module builds to simulator/emulator, Android builds (including Josh's comment), and general SDK building.

Pushed a commit to fix an issue found during testing Android builds

PR #12218, opened 2020-10-26
NSNumber * vs NSInteger

sgtcoolguy:
**Description:**
This grew out of taking a second look at @jquick-axway 's #12193 and @garymathews #12029

I couldn't cause a crash by re-enabling the related tests for the properties, so the fixes here are sort of speculative. But it looks like maybe they'd be useful to Gary's PR since he's intentionally disabling the tests that hit these properties again.


garymathews (approved):
LGTM!

All tests are passing on iOS, I think we can merge this?

PR #12231, opened 2020-10-29
test: generated unit tests and fixes

sgtcoolguy:
**Description:**
I used a new build script to generate tests for some types and then manually merged with our existing suites. It uncovered a fair number of issues!

I used this to generate some tests for:
- `Ti.Codec` extracted out to #12237
- `Ti.UI.Clipboard` extracted out to #12236
- `Ti.Media`
  - lots of "dead" constants/properties removed from the API long ago but still in docs
  - mismatches in types, platform support
  - `availableCameras` documented as `Array`, but if none Android returned `null` rather than empty array
- `Ti.Media.MusicPlayer`
  - typo in constant defined on `Ti.Media` used by type, so constant was exposed with bad name (using documented value was `undefined`!)
- `Ti.Media.VideoPlayer`  
  - incorrect default values
  - property whose enumerated return values weren't defined (`mediaTypes`)
  - code examples used removed/deprecated properties/constants


jquick-axway:
@sgtcoolguy, our Android `TiCameraActivity` Java class duplicates the `"public.video"` string constant below. We should probably remove that class' `MEDIA_TYPE_PHOTO` and `MEDIA_TYPE_VIDEO` constants and have it use the `MediaModule` constants instead.
https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java#L69

The rest looks good to me.

jquick-axway (approved):
CR: Pass

apidoc/Titanium/Media/VideoPlayer.yml
```diff
@@ -598,47 +556,35 @@ properties:
...
     summary: The style of the playback controls.
-    description: |
-        Mobile web only supports [VIDEO_CONTROL_NONE](Titanium.Media.VIDEO_CONTROL_NONE) and
-        [VIDEO_CONTROL_DEFAULT](Titanium.Media.VIDEO_CONTROL_DEFAULT).
+    description: Removed from iOS in SDK 7.0.0
     type: Number
     constants: Titanium.Media.VIDEO_CONTROL_*
     default: System default video controls (<Titanium.Media.VIDEO_CONTROL_DEFAULT>).
 
   - name: mediaTypes
```
vijaysingh-axway, line 575: As per code, I can see this is read-only property.  Need to add `permission: read-only`.

apidoc/Titanium/Media/VideoPlayer.yml
```diff
@@ -679,7 +625,7 @@ properties:
     summary: Whether or not the receiver allows Picture in Picture playback.
     platforms: [iphone, ipad, macos]
     type: Boolean
-    default: true
+    default: false
```
vijaysingh-axway, line 628: By default it is ‘true’.  If this property is not a valid boolean it will set false otherwise it is true. So need to change a bit change in code [here](https://github.com/appcelerator/titanium_mobile/blob/cccfb261292444e1fbda02d7ad00ba0703e6b711/iphone/Classes/TiMediaVideoPlayerProxy.m#L454)

vijaysingh-axway (commented):
There are 2 minor iOS/doc related issues. As this PR is already merged.  I'll create PR to fix that.

vijaysingh-axway:
Here is the PR https://github.com/appcelerator/titanium_mobile/pull/12239.

PR #12237, opened 2020-11-02
fix(ios): correct Ti.Codec.CHARSET_ISO_LATIN_1 to be 'latin1'

sgtcoolguy:

**Description:**
Extracted from #12231 

There was a mismatch in the static value for CHARSET_ISO_LATIN_1 between Android/iOS. The value had a typo on iOS, but also was using an unofficial name/string to refer to it. This changes to static value to be `'latin1'` to match Android and official alternative naming used for it.


vijaysingh-axway (approved):
LGTM!

PR #12239, opened 2020-11-02
fix(ios):  fix default value of property pictureInPictureEnabled

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28221

This PR is to fix iOS related issue found in PR #12231 .


apidoc/Titanium/Media/VideoPlayer.yml
```diff
@@ -576,6 +576,7 @@ properties:
     summary: The type of media in the player's current item first track.
     exclude-platforms: [android]
     type: String
+    permission: read-only
```
sgtcoolguy, line 579: the related unit test should be modified to assert the property is read-only

sgtcoolguy (approved):

lokeshchdhry:
FR Passed.

PR #12245, opened 2020-11-04
fix(ios): set tintcolor of UIBarButtonItem

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28211


tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
...
+			rightNavButtons: [ rightButton1, rightButton2 ],
+		});
+
+		win = Ti.UI.createNavigationWindow({
+			window: rootWindow
+		});
+
+		win.open();
+
+		rootWindow.addEventListener('focus', function () {
```
sgtcoolguy, line 127: remove the focus listener so it doesn't fire more than once

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
...
+			title: 'Right2',
+			color: 'green',
+		});
+
+		var leftButton = Ti.UI.createButton({
+			title: 'Left',
+			color: 'blue'
+		});
+
+		var rootWindow = Ti.UI.createWindow({
```
sgtcoolguy, line 115:
```suggestion
		const rootWindow = Ti.UI.createWindow({
```

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
...
+			title: 'Right1',
+			color: 'green',
+		});
+
+		var rightButton2 = Ti.UI.createButton({
+			title: 'Right2',
+			color: 'green',
+		});
+
+		var leftButton = Ti.UI.createButton({
```
sgtcoolguy, line 110:
```suggestion
		const leftButton = Ti.UI.createButton({
```

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
 		});
 	});
 
+	it.ios('#leftNavButtons and #rightNavButtons', finish => {
+		var rightButton1 = Ti.UI.createButton({
+			title: 'Right1',
+			color: 'green',
+		});
+
+		var rightButton2 = Ti.UI.createButton({
```
sgtcoolguy, line 105:
```suggestion
		const rightButton2 = Ti.UI.createButton({
```

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
 		});
 	});
 
+	it.ios('#leftNavButtons and #rightNavButtons', finish => {
+		var rightButton1 = Ti.UI.createButton({
```
sgtcoolguy, line 100:
```suggestion
		const rightButton1 = Ti.UI.createButton({
```

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
 		});
 	});
 
+	it.ios('#leftNavButtons and #rightNavButtons', finish => {
```
sgtcoolguy, line 99:
```suggestion
	it.ios('.leftNavButtons and .rightNavButtons', finish => {
```

I use the `#` prefix for methods, `.` for properties.

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,86 @@ describe('Titanium.UI.Window', function () {
...
+		win = Ti.UI.createNavigationWindow({
+			window: rootWindow
+		});
+
+		win.open();
+
+		rootWindow.addEventListener('postlayout', function postlayout() {
+			rootWindow.removeEventListener('postlayout', postlayout);
+			try {
+				should(rootWindow.leftNavButton).be.an.Object();
```
sgtcoolguy, line 169: Given the fix here, we should have a "snapshot" test that verifies the buttons take on the color assigned to verify the fix. These tests don't verify that at all, they just verify the nav button properties.

iphone/Classes/TiUINavBarButton.m
```diff
@@ -119,6 +119,7 @@ - (id)initWithProxy:(TiUIButtonProxy *)proxy_
       self = [super initWithImage:theimage style:[self style:proxy_] target:self action:@selector(clicked:)];
     } else {
       self = [super initWithTitle:[self title:proxy_] style:[self style:proxy_] target:self action:@selector(clicked:)];
+      self.tintColor = [TiUtils colorValue:[proxy_ valueForKey:@"color"]].color;
```
sgtcoolguy, line 122: What if no `color` value is given?

sgtcoolguy (changes requested):

iphone/Classes/TiUINavBarButton.m
hansemannn: Again, as mentioned in the ticket: Buttons are tinted via the `tintColor` property, not `color`. Not sure about Android (likely a parity issue there).

vijaysingh-axway: @hansemannn  'color' property is used to set title color in UIButton. If we set 'tintColor', it will set the tintColor of UIButton but effect is   shown on title of button (meaning color of button's title is changed). If we set 'tintColor' and 'color' both, 'color' will override the effect of 'tintColor'. 
I think we should set 'color' if available otherwise 'tintColor'  for self.tintColor here.

vijaysingh-axway: @sgtcoolguy In case of no color or tintColor, it will set nil which is default.

hansemannn: If you are sure that that's the cross platform behavior, I'm fine with that :-) But in our app, we only use tintColor and it works very well - thats why I wondered.

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,146 @@ describe('Titanium.UI.Window', function () {
...
+					try {
+						should(rootWindow.leftNavButton).be.an.Object();
+						should(rootWindow.rightNavButton).be.an.Object();
+						should(win).not.matchImage(nonColorButtonImage, 0); //  Navbutton without color should be different with a colored one
+					} catch (e) {
+						return finish(e);
+					}
+					finish();
+				}, 1000);
+			}, 1000);
```
sgtcoolguy, line 188: why the 1 second delay? Why not use a much shorter delay? (same in the line above)

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,146 @@ describe('Titanium.UI.Window', function () {
...
+				should(rootWindow.leftNavButtons).be.an.Array();
+				should(rootWindow.leftNavButtons.length).be.eql(1);
+			} catch (e) {
+				return finish(e);
+			}
+			finish();
+		});
+	});
+
+	it.ios('.leftNavButton and .rightNavButton  with color', finish => {
```
sgtcoolguy, line 144:
```suggestion
	it.ios('.leftNavButton and .rightNavButton  with color', function (finish) {
```
details below...

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,146 @@ describe('Titanium.UI.Window', function () {
...
+		this.timeout(10000);
```
sgtcoolguy, line 145: `this.timeout()` can't be used in an arrow function, as `this` isn't this single test function (arrow functions don't change `this` in context)

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,146 @@ describe('Titanium.UI.Window', function () {
...
+						return finish(e);
+					}
+					finish();
+				}, 1000);
+			}, 1000);
+		});
+	});
+
+	it.ios('.leftNavButton and .rightNavButton with tintColor', finish => {
+		this.timeout(10000);
```
sgtcoolguy, line 193: same as above test, need to make the arrow function a normal function

tests/Resources/ti.ui.window.test.js
```diff
@@ -96,6 +96,146 @@ describe('Titanium.UI.Window', function () {
...
+		rootWindow.addEventListener('postlayout', function postlayout() {
+			rootWindow.removeEventListener('postlayout', postlayout);
+			setTimeout(function () {
+				const nonColorButtonImage = win.toImage();
+				rootWindow.rightNavButton = greenColorButton;
+				setTimeout(function () {
+					try {
+						should(rootWindow.leftNavButton).be.an.Object();
+						should(rootWindow.rightNavButton).be.an.Object();
+						should(win).not.matchImage(nonColorButtonImage, 0); //  Navbutton without color should be different with a colored one
```
sgtcoolguy, line 182: The test I had in mind, was to simply change the color for a navbutton and verify the image matches expectations for this PR (namely that the text color is the one you assigned). I suppose it can be useful to verify that one given `color: 'green'` doesn't look the same as one without `color` property specified, but you can get to the same place by simply having one test verify the "default" no-color image, and another verify the green color button. Or simply a single test with multiple buttons: one assigned a color, one assigned a tintColor, one with neither assigned and verify that image.

tests/Resources/ti.ui.window.test.js
vijaysingh-axway: Let me try with shorter delay.

tests/Resources/ti.ui.window.test.js
vijaysingh-axway: Noted. I think it is not required. Let me remove it.

tests/Resources/ti.ui.window.test.js
vijaysingh-axway: @sgtcoolguy   If you are talking about matching against  hardcoded image, snapshot does not match on iPhone SE (2nd generation) simulator and iPhone 11 simulator even we use 'px' (e.g. 400px) to set width and height of navigation window. Reason looks nav button titles are a bit different in both simulator. 
So I compared screenshots with colored and non colored nav bar button and it should not match.
I had tried  -  
`	it.ios('.leftNavButton and .rightNavButton  with color', finish => {
		const rightButton = Ti.UI.createButton({
			title: 'Right',
			tintColor: 'red',
			color: 'green',
		});

		const leftButton = Ti.UI.createButton({
			title: 'Left',
			color: 'red'
		});

		const rootWindow = Ti.UI.createWindow({
			backgroundColor: 'white',
			leftNavButton: leftButton,
			rightNavButton: rightButton,
		});

		win = Ti.UI.createNavigationWindow({
			height: '400px',
			width: '400px',
			window: rootWindow
		});

		win.open();

		rootWindow.addEventListener('postlayout', function postlayout() {
			rootWindow.removeEventListener('postlayout', postlayout);
			setTimeout(function () {
				try {
					should(rootWindow.leftNavButton).be.an.Object();
					should(rootWindow.rightNavButton).be.an.Object();
					should(win).matchImage('snapshots/navButtonWithColor.png');
				} catch (e) {
					return finish(e);
				}
				finish();
			}, 1000);
		});
	});
`
But screenshots from iPhone 11 Pro max and iPhone SE (2nd generation) simulator didn't match. Am I missing anything?

sgtcoolguy: I don't know since I don't have the images to look at manually. I assume that despite the use of pixel dimensions some things may matter based on screen dpi, so you'd have to use the density in the filename of the image and produce copies for varying densities?

sgtcoolguy: I tested this locally on this PR, the issue is indeed a scale/density issue whereby the button text is sized differently on different density screens (since text/fonts are always in typographical points on iOS: https://appcelerator.github.io/titanium-docs/api/structs/font.html#fontsize)

vijaysingh-axway: Yeah. You are right. Different scale images should be added. Added 2x and 3x images. I think no need of scale 1x image as we do not support those devices.

sgtcoolguy (approved):

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 9.3.0.v20201116174003
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Xcode: 12.0.1
iPhone 7Plus(v14.0), iPhone X(v13.6.1), iOS sim 14.0

PR #12254, opened 2020-11-07
fix(ios): fix a couple of memory leaks

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28232

There are roughly 15 more. Would be great if someone from the core team could assist here.


iphone/Classes/TiUIiOSProxy.m
```diff
@@ -912,9 +912,9 @@ - (TiColor *)fetchSemanticColor:(id)color
   DEPRECATED_REPLACED(@"UI.iOS.fetchSemanticColor", @"9.1.0", @"UI.fetchSemanticColor");
 
   if ([TiUtils isIOSVersionOrGreater:@"11.0"]) {
-    return [[TiColor alloc] initWithColor:[UIColor colorNamed:color] name:nil];
+    return [[[TiColor alloc] initWithColor:[UIColor colorNamed:color] name:nil] autorelease];
   }
-  return [[TiColor alloc] initWithColor:UIColor.blackColor name:@"black"];
+  return [[[TiColor alloc] initWithColor:UIColor.blackColor name:@"black"] autorelease];
```
hansemannn, line 917: This actually causes some huge leaks, because it's used like a few hundred times by larger apps

vijaysingh-axway (approved):
CR passed!

PR #12258, opened 2020-11-09
refactor(ios): Ti.UI.Clipboard

sgtcoolguy:

**Description:**
This is a refactor that moves the iOS implementation of `Ti.UI.Clipboard` to the newer Obj-C JavaScriptCore API. The only real snag I found here was trying to handle `Ti.Filesystem.File`s because they're still using the "old proxy" API, so it doesn't automagically convert them for us and I had to add a little hack in `setData()` to handle them.

In the process I also:
- fleshed out a lot more unit tests
- made the implementation able to consume [Universal Type Identifiers](https://developer.apple.com/library/archive/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html) for the data type arguments
  - We documented using "mimeTypes", but we really only support some very specific string values that were semi-documented.
  - `getItems()` would (and still does) always report back UTIs, so us not accepting them was a big mismatch.
- Fixed iOS `getItems()` to convert `NSURL` to `String`s; `UIImage` to `Ti.Blob`s; `UIColor` to hex `String` values.
- found and fixed a bug on Android where `Ti.UI.Clipboard#hasData()` documented the `type` argument as optional but was enforcing it was supplied.

### DIFF
--- a/iphone/Classes/TiUIClipboardProxy.m
+++ b/iphone/Classes/TiUIClipboardProxy.m
@@ -1,16 +1,15 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2009-2010 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2009-Present by Axway, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 
 #ifdef USE_TI_UICLIPBOARD
 #import "TiUIClipboardProxy.h"
-#import <TitaniumKit/TiApp.h>
-#import <TitaniumKit/TiBlob.h>
-#import <TitaniumKit/TiFile.h>
-#import <TitaniumKit/TiUtils.h>
+@import TitaniumKit.TiBlob;
+@import TitaniumKit.TiFile;
+@import TitaniumKit.TiUtils;
 
 #import <MobileCoreServices/UTCoreTypes.h>
 #import <MobileCoreServices/UTType.h>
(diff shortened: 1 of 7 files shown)


iphone/Classes/UIModule.m
```diff
@@ -495,11 +494,20 @@ - (id)create3DMatrix:(id)args
 - (id)Clipboard
 {
   if (clipboard == nil) {
-    clipboard = [[TiUIClipboardProxy alloc] _initWithPageContext:[self executionContext]];
-    [self rememberProxy:clipboard];
+    clipboard = [[[TiUIClipboardProxy alloc] _initWithPageContext:[self executionContext]] retain];
```
vijaysingh-axway, line 497: There is memory leak. It should be -
```suggestion
    clipboard = [[TiUIClipboardProxy alloc] _initWithPageContext:[self executionContext]];
```

iphone/Classes/TiUIClipboardProxy.m
```diff
@@ -339,49 +328,58 @@ - (void)setItems:(id)args
...
         }
       },
       YES);
 
-  return [items autorelease];
+  return [result autorelease];
 }
 
-- (void)setData:(id)args
+- (void)setData:(NSString *)mimeType withData:(id)data
```
vijaysingh-axway, line 361: Probably better method name can be -
```suggestion
- (void)setData:(id)data withDataType:(NSString *) mimeType
```

iphone/Classes/TiUIClipboardProxy.m
```diff
@@ -284,40 +275,37 @@ - (id)hasData:(id)args
...
-- (void)setItems:(id)args
+- (void)setItems:(NSDictionary<NSString *, id> *)args
 {
-  NSArray *items = [args objectForKey:@"items"];
-  NSDictionary *options = [args objectForKey:@"options"];
-
-  __block NSMutableArray *result = [[[NSMutableArray alloc] init] retain];
-
+  NSArray<NSDictionary<NSString *, id> *> *items = args[@"items"];
+  __block NSMutableArray<NSDictionary<NSString *, id> *> *result = [[[NSMutableArray alloc] init] retain];
```
vijaysingh-axway, line 304: No need to retain it. Otherwise memory leak.
```suggestion
  __block NSMutableArray<NSDictionary<NSString *, id> *> *result = [[NSMutableArray alloc] init];
```

General rule -
1. alloc and retain increases retain count by 1.
2. release and autorelease decreases retain count by 1.
3. Object will get release completely if retain count is 0.

iphone/Classes/TiUIClipboardProxy.m
```diff
@@ -339,49 +328,58 @@ - (void)setItems:(id)args
...
   TiThreadPerformOnMainThread(
       ^{
-        items = [[[self pasteboard] items] retain];
+        NSArray<NSDictionary<NSString *, id> *> *items = [self pasteboard].items;
 
         // Check for invalid UTI's / mime-types to prevent a runtime-crash
-        for (NSDictionary *item in items) {
-          for (NSString *key in [item allKeys]) {
-            if ([key hasPrefix:@"dyn."]) {
```
vijaysingh-axway, line 353: Does this check not required?

iphone/Classes/TiUIClipboardProxy.m
```diff
@@ -339,49 +328,58 @@ - (void)setItems:(id)args
...
+              newItem[key] = [TiUtils stringValue:item[key]];
+            } else if (UTTypeConformsTo((CFStringRef)key, kUTTypeImage)) {
+              // Convert UIImage to TiBlob!
+              newItem[key] = [[[TiBlob alloc] initWithImage:(UIImage *)item[key]] autorelease];
             }
           }
-          if ([items count] == 0) {
-            break;
-          }
+          [result addObject:newItem];
```
vijaysingh-axway, line 353: newItem should be released after this line.

vijaysingh-axway (changes requested):
There are couple of memory leaks. After resolving leaks, it should be good.

iphone/Classes/TiUIClipboardProxy.m
sgtcoolguy: It likely would be, but I'm constrained by the argument order we expose to JS (datatype, then data)

iphone/Classes/TiUIClipboardProxy.m
sgtcoolguy: I didn't like this check because it "hid" the underlying problems we had sticking data into and pull it back out of the clipboard. If it's an unknown data type, it would get this `'dyn.'` prefix. And that meant if anything in the clipboard had it, we'd basically not return anything. I'll add it back to skip `'dyn.'` entries only (but still turn valid/known ones)

vijaysingh-axway (approved):
CR passed!

PR #12262, opened 2020-11-12
fix: animation default values

janvennemann:
This adds a basic test suite for `Ti.UI.Animation` based on the `generate-test` command.

The tests uncovered two issues which i've fixed in this PR as well:

- Fixes the default values of `repeat` and `autoreverse` to return the correct values instead of `undefined`.
- iOS: Store the current animation complete `callback` from `view.animate(animation, callback)` before firing the `complete` event. This could cause a race condition in certain scenarios when running multiple animation after each other causing the `callback` from the last animation to be called multiple times.

Also updates the examples in the docs to use up-to-date coding styles and consistent formatting.


apidoc/Titanium/UI/Animation.yml
```diff
@@ -228,14 +228,14 @@ examples:
...
+        const view = Titanium.UI.createView({
           backgroundColor:'red'
         });
-        var animation = Titanium.UI.createAnimation();
+        const animation = Titanium.UI.createAnimation();
         animation.backgroundColor = 'black';
         animation.duration = 1000;
-        var animationHandler = function() {
+        const animationHandler = () => {
           animation.removeEventListener('complete',animationHandler);
```
sgtcoolguy, line 239:
```suggestion
          animation.removeEventListener('complete', animationHandler);
```

iphone/TitaniumKit/TitaniumKit/Sources/API/TiAnimation.m
```diff
@@ -232,6 +240,28 @@ - (TiPoint *)center
   return center;
 }
 
+- (void)setColor:(id)color_
+{
+  [color release];
```
sgtcoolguy, line 245:
```suggestion
  RELEASE_TO_NIL(color);
```

Does it matter if we call release on a nil pointer? We tend use the above everywhere to guard against it.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiAnimation.m
```diff
@@ -232,6 +240,28 @@ - (TiPoint *)center
...
+}
+
+- (id)color
+{
+  return color.name;
+}
+
+- (void)setBackgroundColor:(id)backgroundColor_
+{
+  [backgroundColor release];
```
sgtcoolguy, line 256:
```suggestion
  RELEASE_TO_NIL(backgroundColor);
```

iphone/TitaniumKit/TitaniumKit/Sources/API/TiAnimation.m
```diff
@@ -232,6 +240,28 @@ - (TiPoint *)center
...
+  backgroundColor = [[TiUtils colorValue:backgroundColor_] retain];
+}
+
+- (id)backgroundColor
+{
+  return backgroundColor.name;
```
sgtcoolguy, line 262: shouldn't this be returning a `TiColor`, not the name? What if the backgroundColor is `nil`?

iphone/TitaniumKit/TitaniumKit/Sources/API/TiAnimation.m
```diff
@@ -596,14 +628,14 @@ - (void)animate:(id)args
       }
 
       if (backgroundColor != nil) {
-        [reverseAnimation setBackgroundColor:[TiUtils colorValue:[(TiViewProxy *)[(TiUIView *)view_ proxy] valueForKey:@"backgroundColor"]]];
-
+        [reverseAnimation setBackgroundColor:[TiUtils colorValue:[[(TiUIView *)view_ proxy] valueForKey:@"backgroundColor"]]];
         [[view_ proxy] replaceValue:backgroundColor.name forKey:@"backgroundColor" notification:NO];
```
sgtcoolguy, line 632: Is this line necessary if the next line sets the color?

tests/Resources/ti.ui.animation.test.js
```diff
@@ -0,0 +1,420 @@
...
+/* eslint no-undef: "off" */
+'use strict';
+
+const should = require('./utilities/assertions');
+
+describe('Titanium.UI.Animation', () => {
+	let instance;
+
+	beforeEach(() => {
+		// TODO: Create an instance of the type!
```
sgtcoolguy, line 18: remove the line...

tests/Resources/ti.ui.animation.test.js
```diff
@@ -0,0 +1,420 @@
...
+		instance = Ti.UI.createAnimation();
+	});
+
+	afterEach(() => {
+		// TODO: Do any cleanup of the instance we need to do!
```
sgtcoolguy, line 23: remove the line

tests/Resources/ti.ui.animation.test.js
```diff
@@ -0,0 +1,420 @@
...
+		});
+
+		describe('.opacity', () => {
+			it('is a Number', () => {
+				instance.opacity = 0.5;
+				should(instance).have.a.property('opacity').which.is.a.Number();
+			});
+		});
+
+		describe.ios('.opaque', () => {
```
sgtcoolguy, line 151: feels like this property should have a known/documented default? Or does it depend on the view it's applied to?

tests/Resources/ti.ui.animation.test.js
```diff
@@ -0,0 +1,420 @@
...
+
+		describe.ios('.view', () => {
+			it('is a Titanium.UI.View', () => {
+				instance.view = Ti.UI.createView();
+				should(instance).have.a.property('view').which.is.an.Object();
+				should(instance.view).have.a.property('apiName').which.is.eql('Ti.UI.View');
+			});
+		});
+
+		describe.ios('.visible', () => {
```
sgtcoolguy, line 229: another property which feels like it should have a known.documented default value - but perhaps it depends on the view?

tests/Resources/ti.ui.animation.test.js
```diff
@@ -0,0 +1,420 @@
...
+			win.add(container);
+		});
+
+		it('Simple Animation Applied to a View', done => {
+			var view = Titanium.UI.createView({
+				backgroundColor: 'red'
+			});
+			container.add(view);
+			once(win, 'postlayout', () => {
+				should(container).matchImage('snapshots/animation_simple_start.png');
```
sgtcoolguy, line 295: I wrap the async/callback aerations in try/catch blocks with the catch calling `finish(err);`, otherwise we don't behave as expected and the test times out rather than surfaces the Error that occurred async. So basically anything in any event callback/timer needs to be wrapped din `try`/`catch` to manually call `finish(err)` to bubble up the true error. (I assume that's a behavioral difference/bug we have versus Node!)

sgtcoolguy (approved):
I think the only real worry I have here is the implementation of `backgroundColor` on the Animation proxy - I think it needs to return the `TiColor`, not the name.

PR #12263, opened 2020-11-13
chore(android): remove hidden API usage

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27743

**Test:**
1. Create a default Classic Titanium app project.
2. Build and run on Android 11.
3. Verify the following 2 warning messages are **NOT** logged.
```
Accessing hidden method Landroid/content/res/Resources;->getCompatibilityInfo()Landroid/content/res/CompatibilityInfo; (light greylist, reflection)
Accessing hidden field Landroid/content/res/CompatibilityInfo;->applicationScale:F (greylist, reflection, allowed)
```

_**Note**: Other `Accessing hidden method...` warnings may appear, but we can't prevent them since they come from Google's libraries and not our code._


android/titanium/src/java/org/appcelerator/titanium/util/TiPlatformHelper.java
```diff
@@ -62,13 +61,10 @@ public synchronized void intializeDisplayMetrics(Activity activity)
...
 		activity.getWindowManager().getDefaultDisplay().getMetrics(dm);
 
-		// Note: this isn't public API, so there should be lots of error checking here
-		try {
-			Method gciMethod = Resources.class.getMethod("getCompatibilityInfo");
-			Object compatInfo = gciMethod.invoke(activity.getResources());
-			applicationScaleFactor = (Float) compatInfo.getClass().getField("applicationScale").get(compatInfo);
-		} catch (Exception e) {
-			Log.w(TAG, "Unable to get application scale factor, using reported density and its factor", Log.DEBUG_MODE);
+		if ((activity.getApplicationInfo().flags & ApplicationInfo.FLAG_SUPPORTS_SCREEN_DENSITIES) != 0) {
```
sgtcoolguy, line 64: I assume this is more or less translated from this? https://android.googlesource.com/platform/frameworks/base/+/android-7.0.0_r33/core/java/android/content/res/CompatibilityInfo.java#228

sgtcoolguy: Looking deeper at this. This really only appears to be used in one specific place - to hack around for Galaxy Tab when playing videos. Should we just move this logic/field there and/or consider dropping it?

https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/media/src/java/android/widget/TiVideoView8.java#L249-L272

From Googling around this appears to be a Samsung Galaxy Tab 7.0 released on 2010, and that Android 4.1.2 was the last "official" Android release supported on it.

jquick-axway: You're right. I can remove the `applicationScaleFactor` member variable and it's usage in `TiVideoView8`.

And in @garymathews' PR #12029 , he removed the usage of the `TiRHelper.getImageRessource()` method which in turn uses `TiPlatformHelper.applicationLogicalDensity`. So, once we merge his PR, I can get rid of this code completely.

garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.
Don't see the mentioned 2 warnings in the console.

PR #12270, opened 2020-11-18
fix(android): amend chevron vector icon color and size

garymathews:
- For some reason #12029 missed 5a28140
  - Set both `titanium_icon_disclosure.xml` and `titanium_icon_more.xml` to `#858585`
  - Reduce size of `titanium_icon_disclosure.xml`
  - Update `titanium_icon_more.xml` to down chevron

##### TEST CASE
```JS
	backgroundColor: 'white'
});

tableView.setData([
	{ hasCheck: true },
	{ hasChild: true },
	{ hasDetail: true }
]);

win.add(tableView);
win.open();
```


android/modules/ui/res/drawable/titanium_icon_more.xml
```diff
@@ -3,7 +3,7 @@
     android:height="24dp"
     android:viewportWidth="24.0"
     android:viewportHeight="24.0"
```
hansemannn, line 5: The other icons were not affected? Not sure actually.

garymathews: The chevron (and other icons) are actually the correct size (they weren't scaled correctly before), but for parity with previous SDKs I've made the chevron a little smaller here.

jquick-axway (approved):

jquick-axway:
**Question:** Do you think we should change our "more" down arrow icon to a down chevron like what Google usually uses on Android? Such as the one provided by Android Studio as shown in the screenshot below? The icons we're using matches the standard drop-down icon, but I see the down chevron more commonly used. Would this be a more radical change?
AndroidStudioIconSelect

lokeshchdhry:
FR Passed.
The size & down icon changes are visible as expected.

Issue #12265, opened 2020-11-13
TiFileHelper.GetTempFile no longer creates file (breaking behaviour)


MattWard-Ideagen:
Issue
TiFileHelper.GetTempFile() no longer creates temporary file, only returns a file handle.   I think the culprit is https://github.com/appcelerator/titanium_mobile/commit/314b2bb7c71bfcacbfc044e21faa01f635f88b32 where result = File.createTempFile("tia", suffix, dir); was removed. I believe a result.createNewFile() after the file handle has been established in GetTempFile will restore behaviour.

Note:  TiFileHelper.getTempFileFromInputStream in the same class uses this function prior to v9.1.0 of titanium could expect the result of calling TiFileHelper.GetTempFile() to result in a temporary file to be created and return the file handle and is now currently broken on Android (the platform i'm testing on) as 'if (tempFile.exists())' returns false it won't populate the file with the contents.

**To Reproduce**
Steps to reproduce the behavior, e.g.:
1) tempFile =  TiFileHelper.GetTempFile() 
2) Verify Files.exists(tempFile.toPath(), LinkOption.NOFOLLOW_LINKS));

**Expected behavior**
Files.exists should return true to maintain previous behaviour.

**Environment**
Titanium SDK version: v9.2.1
CLI version: 8.1.1

**Logs**:
Not applicable.

(Attach by dragging into this issue)

jquick-axway:
I assume you're using the `TiFileHelper` Java class within your own module?

As of Titanium 9.3.0, I recommend that you do the below instead. The `getTiTempDir()` method was added to 9.3.0 and that entire directory tree will be auto-deleted on every app launch/exit via another thread... versus `TiFileHelper` blocks the UI thread when doing its deletes. The new approach is the superior solution. See PR #12161 for more details.
```java
File file = File.createTempFile("tifile", ".tmp", TiApplication.getInstance().getTiTempDir());
```

But I hear you. It no longer creates the file like it used to. The breaking-change was made on purpose as of Titanium 9.1.0 via PR #11490 . Internally, the thinking was that a getter method shouldn't create a file. Just like how our JavaScript `Ti.Filesystem.getFile()` method won't create a file either. That said, this change completely broke the Java `getTempFileFromInputStream()` method and that needs to be fixed.

jquick-axway:
[...]

PR #12271, opened 2020-11-19
fix(android): Ti.Media.previewImage() fails to display in-memory blobs as of 9.1.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28246

**Github Issue:**
https://github.com/appcelerator/titanium_mobile/issues/12265

**Summary:**
- Changed `TiFileHelper.getTempFile()` to create file like it did before 9.1.0.
- Fixed Java `TiFileHelper.getTempFileFromInputStream()` method which always failed to create file.
  * Prevented our `Ti.Media.previewImage()` method from working.
  * Also caused issues in 3rd party modules.
- Fixed issue where Java `TiFileHelper` would not delete files if app was force-quit.

**Test:**
1. Run the test shown in ticket [TIMOB-28246](https://jira.appcelerator.org/browse/TIMOB-28246).
2. Verify a screenshot of the app is now shown in preview window.


android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -1038,7 +992,6 @@ public void dispose()
 	{
 		TiActivityWindows.dispose();
 		TiActivitySupportHelpers.dispose();
-		TiFileHelper.getInstance().destroyTempFiles();
```
sgtcoolguy, line 1041: why not destroy temp files here now? I assume because this no longer an active API and this just happens automatically implicitly?

android/titanium/src/java/org/appcelerator/titanium/util/TiFileHelper.java
```diff
@@ -535,91 +528,79 @@ public void deployFromZip(File fname, File dest) throws IOException
...
-
-		wipeDirectoryTree(path, dirs);
-
-		Iterator<String> d = dirs.iterator();
-		while (d.hasNext()) {
-			String fn = d.next();
-			File f = new File(fn);
-			Log.d(TAG, "Deleting Dir: " + f.getAbsolutePath(), Log.DEBUG_MODE);
-			f.delete();
+		boolean wasSuccessful = false;
```
sgtcoolguy, line 541: This is entirely a style thing, but personally I'd just do:
``` java
try {
  return deleteTree(file);
} catch (Throwable ex) {
  Log.e(TAG, "Failed to delete directory tree: " + file, ex);
}
return false;
```
I am not a fan of having temp result variables when you can directly return a result or literal.

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
jquick-axway: I've changed `TiFileHelper` to use the same temp directory we expose in JavaScript. This temp directory gets deleted on app startup and every time the JS runtime terminates. See the `deleteTiTempFiles()` method calls below...
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java#L363-L384

Also, the old `TiFileHelper` code wasn't deleting temp files if the app was forced-quit, which happens all the time while the app is backgrounded by the Android OS. This needed to be fixed.

garymathews (approved):
CR: **PASS**

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 9.3.0.v20201119151119
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Device: Pixel 3 XL(v11.0),

PR #12280, opened 2020-11-20
test: account for density in snapshots

garymathews:
- Take into account device density when creating and comparing snapshots


sgtcoolguy:
Covered in 63c4522dc03d846e1bc49acf7ab89de23c2b198c

PR #12281, opened 2020-11-20
fix(android): java deployFromZip() and deployFromAssets()

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28246

**Summary:**
- Fixes regression caused by PR #12271 .
- Caused Java `TiFileHelper.deployFromAssets()` and `TiFileHelper.deployFromZip()` to fail.
- Note that nobody uses these Java methods. It is dead code. (But could be used in the future.)
- Added new `TiFileHelper.emptyDirectory()` which behaves the same as Node's `fs.emptyDir()`.

**Test:**
I've already tested the Java code by hand via the Android Studio debugger. I think we should skip QE and merge it once code reviewed.


garymathews (approved):
CR: **PASS**

PR #12282, opened 2020-11-21
feat(android): add missing options to Intl.DateTimeFormat.resolvedOptions()

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28251

**Summary:**
- Adds missing properties returned by `Intl.DateTimeFormat.resolvedOptions()`.
- Note that this returns "timeZone" which is the only API providing this info on both Android and iOS.

**Test:**
2. Verify an alert dialog appears displaying the device's assigned time zone.
  2.1. Must match a "TZ database name" shown [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).
  2.2. Ex: Will be "America/Los_Angeles" for all of California USA.
  2.3. Ex: Will be "Europe/Dublin" for Dublin Ireland.

```javascript
alert("TimeZone: " + Intl.DateTimeFormat().resolvedOptions().timeZone);
```


android/modules/locale/src/java/ti/modules/titanium/locale/DateTimeFormatProxy.java
```diff
@@ -183,6 +183,7 @@ public void handleCreationDict(KrollDict properties)
 		this.resolvedOptions = new KrollDict();
 		this.resolvedOptions.putAll(options);
 		this.resolvedOptions.put(TiC.PROPERTY_LOCALE, locale.toString().replace("_", "-"));
+		this.resolvedOptions.put("timeZone", this.dateFormat.getTimeZone().getID());
```
hansemannn, line 186: It seems like the locale should be a two-digit number. Also, maybe add the remaining properties now as well:

Bildschirmfoto 2020-11-21 um 09 01 10

jquick-axway: Right. I know resolved options is still incomplete. This PR is only adding `timeZone` since it was easy to do.

Adding the rest of the missing properties may require a bit of a "hack". I would have to make a giant assumption about the Java `getDateInstance()`, `getTimeInstance()`, and `getDateTimeInstance()` methods returning a `SimpleDateFormat` type (which seems to be true) and parse its string pattern to determine what the rest of the properties should be. I'd prefer to avoid doing it this way, but I haven't found a better solution yet.

jquick-axway: Updated PR to add all of the other missing options.

garymathews (approved):
CR: **PASS**

ssjsamir (approved):
FR Passed: Able to get correct time zone for the device using the test case above.  

**Test Environment**
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.3
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
NEXUS 5X API 29 Emulator

PR #12291, opened 2020-11-24
Update rollup plugins, correct specified corejs version

ewanharris:
* Updates the rollups plugins from the deprecated non-scoped ones to the newer shiny scoped ones
* Updates the `corejs` property in the babel options to be the core-js version we depend on, this is required to ensure that the correct polyfills are installed for the environment. For example, our import `core-js/es` ensures that stable ES features are imported, but `Promise.allSettled` was missing. It appears that "stable" is determined by what `core-js-compat` advertises and that requires a specific version for the `corejs` setting rather than just "3". 

The Android ti.main file has 0 file size difference (308KB), and the iOS ti.main file has grown by ~8KB (436KB->444KB) due to the added polyfills

I've put a warning log in if the version/setting get out of sync, reading in the package-lock and setting it directly just didn't seem right to me.


build/lib/builder.js
```diff
@@ -37,16 +37,27 @@ function determineBabelOptions(babelOptions) {
...
-		corejs: 3
+		corejs: '3.7.0' // this should match the dependency version exactly
 	};
 	// extract out options.transform used by babel-plugin-transform-titanium
 	const transform = options.transform || {};
 	delete options.transform;
 
+	// Validate that the core-js dependency and version are matched, we use require here to make
+	// use of the cache if this is called multiple times
+	const packageLock = require('../../package-lock');
```
sgtcoolguy, line 48: So I think we should just read the package-lock and use the value. We're already reading the package-lock to check they're in sync - but we're not failing the build if they aren't, just warning. It's better than nothing, but if we can automatically enforce they're in sync we should; just do it. Otherwise I suspect they'll get out of sync and the warning will get lost in our mountains of output and no one will notice.

sgtcoolguy (changes requested):

build/lib/builder.js
ewanharris: Updated!

sgtcoolguy (approved):

PR #12298, opened 2020-11-25
feat(android): implement list and table editable and moveable functionality

garymathews:
- Implement `editing` support for ListView
  - ListItem, `canEdit` and `canMove` support
- Implement `editing` and `moving` support for TableView
  - TableViewRow, `editable` and `moveable` support

##### TEST CASE
```JS
    tabs: [
        Ti.UI.createTab({
            window: createListViewWindow(),
            title: 'ListView'
        }),
        Ti.UI.createTab({
            window: createTableViewWindow(),
            title: 'TableView'
        })
    ]
});
tabGroup.open();

function createListViewWindow() {
        backgroundColor: 'white'
    });
        items: [{
                properties: {
                    title: 'Item #1 (canEdit: true)',
                    canEdit: true
                }
            },
            {
                properties: {
                    title: 'Item #2 (canMove: true)',
                    canMove: true
                }
            },
            {
                properties: {
                    title: 'Item #3 (default)',
                }
            },
        ]
    });
        sections: [section],
        templates: {
            template: {
                properties: {
                    height: 50,
                    borderRadius: 10,
                    backgroundColor: 'gray'
                },
                childTemplates: [{
                    type: 'Ti.UI.Label',
                    bindId: 'title',
                    properties: {
                        left: '5dp',
                        color: 'white'
                    }
                }]
            }
        },
    });
    const isEditing = _ => `EDITING: ${listView.editing ? 'ON' : 'OFF'}`;
    const btn_editing = Ti.UI.createButton({
        title: isEditing(),
        bottom: 5,
        right: 5
    });
    btn_editing.addEventListener('click', _ => {
        listView.editing = !listView.editing;
        btn_editing.title = isEditing();
    });
    win.add([listView, btn_editing]);
    return win;
}

function createTableViewWindow() {
        backgroundColor: 'white'
    });
        backgroundColor: 'white',
        data: [{
                title: 'Item #1 (editable: true)',
                editable: true,
                color: 'white',
                backgroundColor: 'gray',
                borderRadius: 10,
                height: 50
            },
            {
                title: 'Item #2 (movable: true)',
                movable: true,
                color: 'white',
                backgroundColor: 'gray',
                borderRadius: 10,
                height: 50
            },
            {
                title: 'Item #3 (default)',
                color: 'white',
                backgroundColor: 'gray',
                borderRadius: 10,
                height: 50
            }
        ]
    });
    const isEditing = _ => `EDITING: ${tableView.editing ? 'ON' : 'OFF'}`;
    const btn_editing = Ti.UI.createButton({
        title: isEditing(),
        bottom: 5,
        right: 5
    });
    btn_editing.addEventListener('click', _ => {
        tableView.editing = !tableView.editing;
        btn_editing.title = isEditing();
    });
    const isMoving = _ => `MOVING: ${tableView.moving ? 'ON' : 'OFF'}`;
    const btn_moving = Ti.UI.createButton({
        title: isMoving(),
        bottom: 5,
        left: 5
    });
    btn_moving.addEventListener('click', _ => {
        tableView.moving = !tableView.moving;
        btn_moving.title = isMoving();
    });
    win.add([tableView, btn_editing, btn_moving]);
    return win;
}
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26074)


hansemannn:
Really amazing change, thank you Gary! One question if you allow: Does Android commonly also use the "drag bars" like iOS does or do users attempt to drag directly?

garymathews:
@hansemannn Thanks for the feedback Hans, I was thinking of adding a drag handle and still might. It would override the `rightImage` of the row when in `editing` mode.

And `editActions` is definitely on my _TODO_ list 😃

hansemannn:
There are 4 possible issues with this PR right now:

- After moving a cell inside a list view and closing the containing window, the list view is not visible anymore
- There is sometimes a red background color visible when moving a cell (might be an Emulator issue though)
- The `onMove` event already fires when being inside an active drag, not the drag end (like iOS does), causing performance issues for us (since it includes network requests and sync)
- It feels like there is a space between the items now. Might be unrelated to this PR, but a possible regression of #12029

jquick-axway:
@garymathews,

Regarding moving rows, I think a good example of how Google handles this on Android can be seen under `Settings/System/Languages&Input/Lanugages`. To better match what Google does, I think we should do the following...
- Add "elevation" to the row as it's being dragged so that it will show a drop shadow. When the move is done, set the elevation back to zero.
- With your PR, I have to tap-and-hold the row's drag bars before I can move it, but I don't have to do this in Google's list where it drags immediately. We should do the same.

Regarding swipe-to-edit, by default row backgrounds are transparent. When swiping a row to the "right", the red rectangle beneath it overlaps the entire row's title width making it hard to read (looks kind of ugly). The red rectangle width should end to the left of the title instead. Swiping to the "left" works a little better because the red rectangle does not overlap the title, although the red rectangle starts where the row title ends. Can we calculate the width of the row based on the width of the RecyclerView instead?

jquick-axway:
@garymathews, I tested your PR's newest changes.

When I swipe-to-edit with a row that has a transparent background (the default), the red background no longer shows through which is good. However, the trash icon appears above the row. You can see this if you get rid of the window's white "backgroundColor" since it defaults to dark gray. Here's an idea... maybe you should set the elevation on the row when doing a swipe-to-edit as well because the elevation will make it render on top for you (might be a simpler solution).

Dragging a ListView/TableView row logs the following warning every time the dragged row changes position in the list.
`[WARN]  EventsRelays: Item change notification received for unknown item: [object Object]`

And if the row's background is transparent (which is the default), then the elevation looks funky as shown below. If I apply a background color to the row, then it looks fine.
TableViewTransparentRowMove

garymathews:
**Updated PR**

- Rows will resort to using parents background to workaround transparency issues
- `move` event now fires upon release of dragging an item

As for the warning, I think it's due to the list being updated while a movement is happening. I don't think I can do anything about this.
```
W/EventsRelays: Item change notification received for unknown item: [object Object]
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ItemTouchHandler.java
```diff
@@ -81,6 +82,37 @@ private boolean canSwipe(TiViewProxy holderProxy)
...
+			final TiViewProxy parentProxy = parentView.getParent();
+			if (parentProxy == null) {
+				break;
+			}
+			parentView = parentProxy.getOrCreateView();
+		}
+
+		return parentBackground;
+	}
+
```
jquick-axway, line 144: I'm still seeing this row transparency issue if I don't set "backgroundColor" on any of the proxies (Window, ListView, ListItem), which is the default and completely uses themes.

Maybe we should look at all of the parent views and use the first non-transparent color drawable like the below? Odds are the activity's root decor view will be assigned a color drawable.
```java
private ColorDrawable getBackgroundColorDrawable(TiUIView uiView)
{
	if (uiView == null) {
		return null;
	}

	View view = uiView.getNativeView();
	while (view != null) {
		Drawable drawable = view.getBackground();
		if (drawable instanceof ColorDrawable) {
			ColorDrawable colorDrawable = (ColorDrawable) drawable;
			if (colorDrawable.getColor() != Color.TRANSPARENT) {
				return colorDrawable;
			}
		}

		ViewParent viewParent = view.getParent();
		view = (ViewParent instanceof View) ? (View) viewParent : null;
	}
	return null;
}
```

jquick-axway: Alternatively, we can add a fallback which fetches the activity's "windowBackground" color.
```java
Activity activity = this.recyclerViewProxy.getActivity();
if (activity != null) {
	TypedValue attribute = new TypedValue();
	if (activity.getTheme().resolveAttribute(android.R.attr.windowBackground, attribute, true)) {
		if ((attribute.type >= TypedValue.TYPE_FIRST_COLOR_INT) && (attribute.type <= TypedValue.TYPE_LAST_COLOR_INT)) {
			return new ColorDrawable(attribute.data);
		}
	}
}
```

jquick-axway (approved):
CR/FR: Pass

ssjsamir (approved):
**FR Passed** Tested using the test case in the description able to edit and move rows.

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.3
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Nexus 5X API 29 
```

PR #12307, opened 2020-12-01
chore(ios): replace deprecated OSSpinLock with os_unfair_lock

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28259


hansemannn:
I think it's a bit more complex, because it involves a different kind of locking than before. The test suite also crashes, which leads to a bigger threading issue

vijaysingh-axway:
The crash is happening on Mac app only. For iPhone and iPad it is working fine. As per warning and doc this is the replacement API. And OSSpinLock api has corresponding os_unfair_lock API. I think it is fine to replace. Probably we can test more.

I have rerun the Jenkins.

hansemannn:
Please just test this very carefully, as there have been regressions related to threading a few times in the past (long ago)

sgtcoolguy (approved):
Seems like a relatively straight-forward replacement.

PR #12310, opened 2020-12-02
Remove Accessors

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26119

**Description:**
This PR removes the getter/setter methods for properties where we can directly get/set the value for that property. If a get/set method is explicitly defined, it should not be removed. We've been threatening to remove these for years now. Time to rip off the band-aid.

Note that a lot of work is in the test suite. I needed to clean them up and explicitly test that many properties *don't* have accessors anymore to help me find what needed to be changed.

So the basic guideline I tried to follow here was:
- If an underlying get/set method was tagged as both `@Kroll.method` and `@Kroll.getProperty` (or `@Kroll.setProperty`) I'd remove the `@Kroll.method`.
- There are still get/set methods exposed to JS, but they should be "unusual" in the sense that they should accept additional arguments (i.e. a getter accepts some args, even optionally; a setter accepts two+ args)
  - For example `Ti.Network.HTTPClient.setRequestHeader('headerName', 'value');` 
- In a couple cases I left the getter/setter if we explicitly documented them and hadn't yet deprecated them (and they had properties you could access). I added deprecation notes in the api docs so we can eventually remove them.
- For `QuickSettingsService` it had getter/setter pairs for underlying properties *but did not expose the properties*. In that case, I exposed the properties, documented them and left the getter/setters but marked them as deprecated in api docs.
- **I did not remove the deprecation messages baked into Android's proxy code**
  - I assume we'll want to. I think at this point it's likely to just spit out false reports?

### DIFF
--- a/apidoc/Titanium/Network/HTTPClient.yml
+++ b/apidoc/Titanium/Network/HTTPClient.yml
@@ -233,6 +233,18 @@ methods:
         type: String
     platforms: [android, iphone, ipad, macos]
 
+  - name: getAllResponseHeaders
+    summary: All of the response headers.
+    description: |
+        Contains a single string, or an empty string if no headers are available.
+
+        See also: [getResponseHeader](Titanium.Network.HTTPClient.getResponseHeader).
+
+        Intended to match the `XMLHTTPRequest` [getAllResponseHeaders API](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders).
+    returns:
+        type: String
+    since: {iphone: "10.0.0", ipad: "10.0.0", macos: "10.0.0"}
+
   - name: getResponseHeader
     summary: Returns the value of the specified response header.
     returns:
--- a/apidoc/Titanium/Network/HTTPClient.yml
+++ b/apidoc/Titanium/Network/HTTPClient.yml
@@ -243,7 +243,7 @@ methods:
         Intended to match the `XMLHTTPRequest` [getAllResponseHeaders API](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders).
     returns:
         type: String
-    since: {iphone: "10.0.0", ipad: "10.0.0", macos: "10.0.0"}
+    since: {iphone: "10.0.0", ipad: "10.0.0", macos: "10.0.0", android: "0.8"}
 
   - name: getResponseHeader
     summary: Returns the value of the specified response header.
--- a/android/kroll-apt/src/main/resources/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
+++ b/android/kroll-apt/src/main/resources/org/appcelerator/kroll/annotations/generator/ProxyBindingV8.cpp.fm
@@ -276,22 +276,14 @@ void ${className}::${method.apiName}(const FunctionCallbackInfo<Value>& args)
 	jvalue* jArguments = 0;
 	</#if>
 
-	<#if method.args?size == 0 && name[0..2] == "get" && isDynamic>
+	<#if method.args?size == 0 && name[0..2] == "get" && isDynamic && !isTitaniumSdk>
 		<#assign propertyName = name[3]?lower_case + name[4..]>
 	const char *deprecationMessage = "Getter method deprecated, please use \"obj.${propertyName};\" or \"obj['${propertyName}'];\" instead.";
-		<#if isTitaniumSdk>
-	Proxy::logDeprecation(isolate, deprecationMessage);
-		<#else>
 	LOGW(TAG, deprecationMessage);
-		</#if>
-	<#elseif method.args?size == 1 && name[0..2] == "set" && isDynamic>
+	<#elseif method.args?size == 1 && name[0..2] == "set" && isDynamic && !isTitaniumSdk>
 		<#assign propertyName = name[3]?lower_case + name[4..]>
 	const char *deprecationMessage = "Setter method deprecated, please use \"obj.${propertyName} = val;\" or \"obj['${propertyName}'] = val;\" instead.";
-		<#if isTitaniumSdk>
-	Proxy::logDeprecation(isolate, deprecationMessage);
-		<#else>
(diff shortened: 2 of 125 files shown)


garymathews:
There's still some `@Kroll.method` entries in:
- [Ti.Calendar.RecurrenceRule](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java#L278)

We may also want to deprecate these methods:
- [Ti.Database.ResultSet.isValidRow()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/database/src/java/ti/modules/titanium/database/TiResultSetProxy.java#L263)
- [Ti.Filesystem.isExternalStoragePresent()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java#L85)
- [Ti.Network.HTTPClient.getAllResponseHeaders()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/network/src/java/ti/modules/titanium/network/HTTPClientProxy.java#L104) _NOTE: This is not documented._
- [Ti.UI.Clipboard.getText()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/clipboard/ClipboardModule.java#L72) _NOTE: Expose properties for other data types? (e.g: Ti.UI.Clipboard.data)_
- [Ti.Codec.getNativeByteOrder()](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/ti/modules/titanium/codec/CodecModule.java#L256)

apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -233,6 +233,18 @@ methods:
...
+    description: |
+        Contains a single string, or an empty string if no headers are available.
+
+        See also: [getResponseHeader](Titanium.Network.HTTPClient.getResponseHeader).
+
+        Intended to match the `XMLHTTPRequest` [getAllResponseHeaders API](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders).
+    returns:
+        type: String
+    since: {iphone: "10.0.0", ipad: "10.0.0", macos: "10.0.0"}
+
```
jquick-axway, line 247: Android already has an `allResponseHeaders` property and `getAllResponseHeaders()` method. It looks like you didn't remove the getter for this on Android. So, you should document this for Android too. Note that Android has had this API for years.

FYI: This used to be a parity issue between Android and iOS. We recently added `responseHeaders` property support to Android in 9.3.0 to match iOS, which returns the headers as a dictionary.
https://jira.appcelerator.org/browse/TIMOB-24983

jquick-axway (changes requested):

apidoc/Titanium/Network/HTTPClient.yml
jquick-axway: We document that `allResponseHeaders` was added to Android as of v0.8.

sgtcoolguy: Ok, I'll add android since 0.8 here as well. I explicitly didn't remove this because it's intended to mirror the XMLHTTPRequest API from browsers (which is what HTTPClient is modeled on).

sgtcoolguy:
> There's still some `@Kroll.method` entries in:
> 
> * [Ti.Calendar.RecurrenceRule](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java#L278)
> 

- OK, I will push removal of the RecurrenceRule getters

> We may also want to deprecate these methods:
> 
> * [Ti.Database.ResultSet.isValidRow()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/database/src/java/ti/modules/titanium/database/TiResultSetProxy.java#L263)
> * [Ti.Filesystem.isExternalStoragePresent()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/filesystem/src/java/ti/modules/titanium/filesystem/FilesystemModule.java#L85)

I think these fall into the category of being existing documented methods. We may want to consider exposing as properties and deprecating long-term?

> * [Ti.Network.HTTPClient.getAllResponseHeaders()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/network/src/java/ti/modules/titanium/network/HTTPClientProxy.java#L104) _NOTE: This is not documented. The property also need documenting, maybe we can outright remove the getter method?_

As noted in comments w/ Josh, this method should stay. I just missed documenting that Android has had it around since forever. I'll push that change to the docs.

> * [Ti.Media.AudioPlayer.getAudioSessionId()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/media/src/java/ti/modules/titanium/media/AudioPlayerProxy.java#L232)

We didn't expose this as a property before, so I am deprecating it for now but not removing it.

> * [Ti.UI.Clipboard.getText()](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/clipboard/ClipboardModule.java#L72) _NOTE: Expose properties for other data types? (e.g: Ti.UI.Clipboard.data)_

Given the nature of the API it felt odd to me to try and make it property only for text (but you'd need to use get/set for "data").

> * [Ti.Codec.getNativeByteOrder()](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/ti/modules/titanium/codec/CodecModule.java#L256)

This method has been around forever and is documented. We could consider deprecating and pushing towards a property access usage. But we haven't documented it as a property ever - and I'm trying to only remove "auto-generated" getter/setters for documented properties but retain existing documented methods when possible.

So I think it may make sense to expose as a property here and eventually deprecate the method to prefer the property.

> 
> I see you deprecated some of them in our docs, but we can also add the `@Deprecated` annotation.

vijaysingh-axway (approved):
iOS looks good.

PR #12318, opened 2020-12-04
fix(android): accelerometer logs warnings after app exit

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28272

**Summary:**
- After setting up an "update" event listener with `Ti.Accelerometer`, a lot of warnings will be logged after app exit if you don't remove the listener.
- Will now automatically unregister the sensor listener when JS runtime is terminated.

**Test:**
1. Build and run the below code on a real/physical device.
2. Verify the following messages are logged when you move the device.
`[INFO]  @@@ (x,y,z) = (<Number>,<Number>,<Number>)`
3. Back out of the app.
4. Verify that the above messages stop logging. _(Might take a couple seconds.)_
5. Verify that the below warning is NOT logged. _(This is the fix.)_
`[WARN] V8Object: (main) [xx,xx] Runtime disposed, cannot fire event 'update'`
6. Re-launch the app.
7. Verify that accelerometer data is being logged again.

```javascript
Ti.Accelerometer.addEventListener("update", (e) => {
	Ti.API.info(`@@@ (x,y,z) = (${e.x},${e.y},${e.z})`);
});
```


sgtcoolguy (approved):
LGTM. We could potentially add this listener only when the user adds a listener, but I imagine the code to toggle it on and off is more complicated and wouldn't gain us much here.

ssjsamir (commented):
@jquick-axway When I back out of the app I get the warning logged 3 times but unlike before it is no longer logged continuously, Is this okay?  

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 2
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Nexus 6P Android 8.1.0
```

jquick-axway:
@ssjsamir , yes, I think this is okay... for now.

I'm thinking about changing our Android code to fire events "synchronously" like how it works on iOS, which would get rid of the remaining logged warnings here... and it would also fix that drag-touch wobble effect you saw in my other PR. But a change like that would be best be done in a separate PR. But thanks for brining this up!

ssjsamir (approved):
@jquick-axway In that case FR Passed: 

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.2 Beta
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Nexus 6P (Android 8.1.0)
```

PR #12320, opened 2020-12-05
fix: convertPointToView() to use "ti.ui.defaultunit"

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27807

**Summary:**
- Fixes bug where `convertPointToView()` always returns pixels on Android.
- Fixes bug where `convertPointToView()` always returns dips on iOS.
- This method now returns the units set by "tiapp.xml" file's ti.ui.defaultunit". _(Dips by default.)_
- Fixed `convertPointToView()` to accept x/y point string values with post-fixed units.

**Test:**
2. Attempt to drag one of the squares on screen.
3. Verify square follows your finger. _(Square may wobble on Android due to a separate issue.)_
4. Build for iOS and repeat steps 2 - 3.
5. In the "tiapp.xml", set the following property to pixels.
`<property name="ti.ui.defaultunit" type="string">px</property>`
6. Build for Android and repeat steps 2 -3.
7. Build for iOS and repeat steps 2 -3.

app.js
```javascript
function createDragSquare(top, left, color) {
	var view = Ti.UI.createView({
		touchEnabled: true,
		backgroundColor: color,
		center: { x: left + 50, y: top + 50 },
		width: 100,
		height: 100,
	});
	view.addEventListener("touchmove", function(event) {
		view.center = view.convertPointToView({ x: event.x, y: event.y }, window);
		event.cancelBubble = true;
	});
	return view;
}

var window = Ti.UI.createWindow({ backgroundColor: "white" });
window.add(createDragSquare(40, 40, "red"));
window.add(createDragSquare(120, 120, "green"));
window.add(createDragSquare(200, 200, "blue"));
window.open();
```


garymathews (commented):
This PR brings current behaviour inline with iOS. However, I brought up that [Point](https://appcelerator.github.io/titanium-docs/api/structs/point.html) can accept `String` parameters allowing units to be specified. @jquick-axway found that iOS also did not support this. So either we drop `String` support, or allow units to be specified.

jquick-axway:
Updated PR. Now supports x/y string value with units postfixed to it.

garymathews (approved):
CR: **PASS**

ssjsamir (approved):
FR Passed: Tested using the test case mentioned above. Note* As mentioned in the description the Android test case wobbles but this is a separate issue. 

**Test Environment** 
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.2 Beta
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Pixel XL (10.0) Sim
```

jquick-axway:
@ssjsamir , the following PR will address that touch drag wobble issue.
https://github.com/appcelerator/titanium_mobile/pull/12340

PR #12328, opened 2020-12-08
feat(ios): support storyboards in module

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-20473

- This implementation is similar to framework.js
- It scans Resources and platform folder of modules for .storyboard files. If found, will inject storyboard in generated xcode project.


iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+'use strict';
+
+const fs = require('fs-extra');
+const IncrementalFileTask = require('appc-tasks').IncrementalFileTask;
+const path = require('path');
+
+exports.cliVersion = '>=3.2.1';
+
+exports.init = function (logger, config, cli) {
+	let storyboardManager = new StoryboardManager(cli, config, logger);
```
sgtcoolguy, line 17: why use `let` here? it doesn't get re-assigned/changed, so could be `const` (or not even assigned to a variable)

sgtcoolguy: Also, why re-order the passed in arguments (`logger, config, cli -> cli, config, logger`)? Ideally we try and retain arg order just purely for developer sanity sake (not having to look up the order for every call)

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+	/**
+	 * Initializes the storyboard manager by hooking into the required build steps.
+	 *
+	 * @access public
+	 */
+	initialize() {
+		this._cli.on('build.pre.compile', {
+			priority: 1200,
+			post: (builder, callback) => {
+				this._logger.trace('Starting storyboard detection');
```
sgtcoolguy, line 54: Please guard all of this to only happen on iOS

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+				this._builder = builder;
+				this.detectStoryboards().then(callback, e => {
+					if (this._cli.argv.platform === 'ios') {
```
sgtcoolguy, line 57: Once guard is at top-level of the callback, this isn't necessary. Also, I'm not sure why specifically we need to spit out to stack trace here? Does the higher-level handling code not give enough details when an Error occurs?

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+			files = await fs.readdir(storyboardsPath);
+		} catch (err) {
+			// ignore non-existing directories
+			return;
+		}
+
+		this._logger.trace(`Scanning ${storyboardsPath.cyan} for storyboards`);
+		const foundStoryboardPaths = [];
+		for (const filename of files) {
+			const possibleStoryPath = path.join(storyboardsPath, filename);
```
sgtcoolguy, line 145: Technically you could delay constructing the full path here to inside the if body and have the `if` just check the filename ends with storyboard. Not sure it makes a huge perf difference...

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+	/**
+	 * Finds any .storyboard file inside modules.
+	 *
+	 * @return {Promise<string[]>} Promise resolving to array of available storyboard paths
+	 * @access private
+	 */
+	async findStoryboardPaths() {
+		const pathsToScan = [];
+		for (let module of this._builder.modules) {
+			pathsToScan.push(path.join(module.modulePath, 'platform'));
```
sgtcoolguy, line 118: Do you also want to check `platform/ios`, `platform/iphone`, `Resources/ios`, `Resources/iphone`? Because `this.scanPathForStoryboards()` isn't recursive, so you'd have to add those subdirs here to the array to check.

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+	}
+
+	/**
+	 * Inspects each storyboard for their type and supported architectures.
+	 *
+	 * @param {Set.<String>} storyboardPaths - List of storyboard paths to inspect
+	 * @return {Promise}
+	 * @access private
+	 */
+	inspectStoryboards(storyboardPaths) {
```
sgtcoolguy, line 364: why not make `async` so you can `await` rather than use the `then` call in the loop?

iphone/cli/hooks/storyboard.js
```diff
@@ -0,0 +1,591 @@
...
+	 * instance for it.
+	 *
+	 * @param {String} storyboardPath - Path to the storyboard to inspect
+	 * @return {Promise<StoryboardInfo>}
+	 * @access public
+	 */
+	async inspect(storyboardPath) {
+		const packageExtension = path.extname(storyboardPath);
+		const storyboardName = path.basename(storyboardPath, packageExtension);
+		let storyboardInfo;
```
sgtcoolguy, line 524: why define `storyboardInfo` variable here? Why not just do a `return new StoryboardInfo(meta);` at the bottom?

sgtcoolguy (changes requested):

iphone/cli/hooks/storyboard.js
vijaysingh-axway: Done.

iphone/cli/hooks/storyboard.js
vijaysingh-axway: Is it really required to guard for iOS? This hook is already inside iphone/cli/hooks. I believe it will there only when iOS is building.

iphone/cli/hooks/storyboard.js
vijaysingh-axway: I think it is not required.

iphone/cli/hooks/storyboard.js
vijaysingh-axway: The storyboard is only getting fetched from modules. In modules there should not be platform/ios, platform/iphone, Resources/ios, Resources/iPhone as modules are platform specific.

iphone/cli/hooks/storyboard.js
vijaysingh-axway: Done

iphone/cli/hooks/storyboard.js
vijaysingh-axway: done.

sgtcoolguy (approved):

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 10.0.0.v20210210150622
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Xcode: 12.3

PR #12335, opened 2020-12-11
fix(ios): use supportedInterfaceOrientations of UINavigationController.topViewController instead of UINavigationController

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28282

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewController.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewController.m
@@ -104,8 +104,11 @@
      If we are in a navigation controller, let us match so it doesn't get freaked 
      out in when pushing/popping. We are going to force orientation anyways.
      */
-  if ([self navigationController] != nil) {
-    return [[self navigationController] supportedInterfaceOrientations];
+  /*
+   TIMOB-28282. Shouldn't UINavigationController.topViewController decide the supported orientation?
+   */
+  if ([self navigationController] != nil && [[self navigationController] topViewController] != self) {
+    return [[[self navigationController] topViewController] supportedInterfaceOrientations];
   }
   //This would be for modal.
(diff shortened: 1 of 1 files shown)


ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 9.2.3.v20201210165845
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Xcode: 12.0.1
iPhone 7Plus(v14.2)

iphone/TitaniumKit/TitaniumKit/Sources/API/TiViewController.m
```diff
@@ -104,8 +104,11 @@ - (UIInterfaceOrientationMask)supportedInterfaceOrientations
      If we are in a navigation controller, let us match so it doesn't get freaked 
      out in when pushing/popping. We are going to force orientation anyways.
      */
-  if ([self navigationController] != nil) {
-    return [[self navigationController] supportedInterfaceOrientations];
+  /*
+   TIMOB-28282. Shouldn't UINavigationController.topViewController decide the supported orientation?
+   */
+  if ([self navigationController] != nil && [[self navigationController] topViewController] != self) {
```
sgtcoolguy, line 110: I do not know enough about this to give useful feedback. The code looks ok style-wise, but that's about all I can offer. 

I think for a navigation controller, you're supposed to ask it's delegate for the supported orientations:
- https://developer.apple.com/documentation/uikit/uinavigationcontroller?language=objc
- https://developer.apple.com/documentation/uikit/uinavigationcontrollerdelegate/1621884-navigationcontrollersupportedint?language=objc

But I don't see us implementing the expected method for the delegate protocol in any classes we declare conforming to the protocol.

PR #12336, opened 2020-12-11
fix(android): file list/existence checks fail with remote encryption policy

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28283

**Summary:**
When property "appc-sourcecode-encryption-policy" is set to "remote"...
- Prevents bootstrap scripts from being loaded on startup.
- Causes apps using hyperloop to crash on startup as of Titanium 9.0.0.
- `File.exists()` always returns `false` for encrypted files.
- `File.getDirectoryListing()` never lists encrypted files.

**Cause:**
- Remote encrypted asset paths were missing the "Remote/" folder prefix.
- Fetched asset file list will always be empty for 1st launch since appc-verify hasn't download the key/info yet. Need to keep calling getAssetPaths() until it's been acquired.

**Test:**
1. Download the [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples) project.
2. Add the following property to the "tiapp.xml" file.
`<property name="appc-sourcecode-encryption-policy">remote</property>`
3. Build and run on Android.
4. Verify the app starts up without issue.

### DIFF
--- a/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
+++ b/android/runtime/common/src/java/org/appcelerator/kroll/util/KrollAssetHelper.java
@@ -32,6 +32,7 @@ public class KrollAssetHelper
 	private static String packageName;
 	private static String cacheDir;
 	private static AssetCrypt assetCrypt;
+	private static Collection<String> encryptedAssetFilePaths;
 	private static HashSet<String> apkAssetFilePathSet = new HashSet<>(256);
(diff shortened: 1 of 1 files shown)


ewanharris (approved):
Tested hyperloop-examples and also validated the alloy bootstrap file functions correctly, changes look good to me!

lokeshchdhry:
FR Passed.

PR #12347, opened 2020-12-15
feat(android): implement top-level-await and native import support

garymathews:
- Implement support for top-level-await
  - Requires changes to [node-titanium-sdk/lib/jsanalyze.js](https://github.com/appcelerator/node-titanium-sdk/blob/master/lib/jsanalyze.js#L79) - _In progress_
- Javascript is now parsed and executed as a module, including imported modules
  - This provides native `export` support
- Removed wrapping of modules as workaround for module execution context
- `import` has been implemented natively, although it currently still piggybacks off our `require` implementation as it parses paths and handles native modules well
- Remove lots of redundant code from previous module loading implementation
- Updated V8 to `8.7.220.29`
  - Compile using _C++14_ as a requirement for new changes in `v8.h`

##### TODO
- [x] Fix core JS methods (_console.log_, _setTimeout_, etc...)
- [x] ~Modernize our `runtime/common/src/js/*.js` files~
- [x] Create destructor to clean-up `moduleContexts`
- [x] Remove unused `runAsScript`
- [x] Tidy exceptions from module
- [x] Fix reference to internal data types (_Object_, _String_, etc...)
- [ ] Update `node-titanium-sdk` to allow for native `import/export` and top-level-await support

[TIMOB-28286](https://jira.appcelerator.org/browse/TIMOB-28286)


android/runtime/v8/src/native/modules/EvaluateModule.cpp
```diff
@@ -0,0 +1,351 @@
...
+			result = module->GetModuleNamespace().As<Object>();
+		}
+
+		// Return result.
+		args.GetReturnValue().Set(result);
+	}
+
+	void EvaluateModule::RunAsScript(const FunctionCallbackInfo<Value>& args)
+	{
+		Isolate* isolate = args.GetIsolate();
```
sgtcoolguy, line 276: Most of the body in the script/modules codepaths is the same, curious if you could extract and re-use some functions/code to avoid the duplication. Looks to me that the ScriptOrigin and the actual methods/types called for compiling/executing are the main differences.

android/runtime/common/src/js/module.js
```diff
@@ -571,20 +569,25 @@ Module.prototype._runScript = function (source, filename) {
...
 
 	const ti = new Titanium.Wrapper(context);
 
-	// In V8, we treat external modules the same as native modules.  First, we wrap the
-	// module code and then run it in the current context.  This will allow external modules to
-	// access globals as mentioned in TIMOB-11752. This will also help resolve startup slowness that
-	// occurs as a result of creating a new context during startup in TIMOB-12286.
-	source = Module.wrap(source);
-
-	const f = Script.runInThisContext(source, filename, true);
```
sgtcoolguy, line 586: This is the underlying reason why the `setTimeout` is not defined Errors are happening, AFAIK. Basically before we ran in "this" context, i.e. we passed along our main app global context and that's what the "module"/script ran under. Basically it shared the current context, so changes were reflected in the current (global) context.

With your new code there's no way to run in "this" context, so everything is isolated. In both codepaths (runAsModule, runAsScript) you're always creating a new context under which the module executes. So while we can now pass in a context object as the 3rd arg, basically it's a one-way copy from this context to the created one under the hood - i.e. we can tell it to pass along require/Ti/global, etc to be set on the global/context of the module, but that doesn't get "copied" back. Note also that just copying a reference to the global likely won't behave as expected, since contexts define new globals (see more below).

Now, I don't know if we need to retain some way to specify to run in "this" context still to fix this, or maybe when you generate your new context you can pass in the pre-existing global object to it (in `Context::New`) - I don't know if changes made to the global in that new context will be "wiped away" once that new context is "done", or if they'd persist outside the module.

I suspect this may also be why in my own PR around require I was seeing issues in extending Ti.Blob from JS suddenly - it's possible the extensions done in the bootstrapping (`kroll.js`) vs app startup (`ti.main.js`) may affect the global differently due to the context the module/script is executed under.

sgtcoolguy:
So, this is very much related to #11693 - we're hitting the same basic code path.

My PR attempts to move iOS to match Android's require/Module implementation and move all JS code into a new ti.kernel.js that we load on both platforms. (So startup becomes more like Android's 2-pass system where you load the kernel/kroll.js equivalent to bootstrap the `Module` code and then use that to load `ti.main.js`).

sgtcoolguy:
@garymathews Can we break out the V8 update/C++14 change to a separate PR we can land earlier? I assume that'd be fairly straight-forward to do and CR to get landed. Also, I'd expect that this will keep shifting as a target since we're going to want to ensure the absolute latest stable V8 release prior to feature freeze (which I'd anticipate would like end up 8.8 for this cycle?)

PR #12348, opened 2020-12-16
fix(android): implement missing scroll events for ListView and TableView

garymathews:
- Implement missing `scroll` and `scrollend` events for TableView
- Implement missing `scrolling`, `scrollstart` and `scrollend` events for ListView
- Add `type` property to events
- Fix exposure of `bindId` for child templates in events

##### TEST CASE
```JS
	sections: [ Ti.UI.createListSection() ]
});

const items = [];
	items.push({ properties: { title: `Item #${i}` } });
}
listView.sections[0].items = items;

function callback(e) {
	console.log(`${e.type} properties: ${JSON.stringify(Object.keys(e))}`);
}

listView.addEventListener('scrolling', callback);
listView.addEventListener('scrollstart', callback);
listView.addEventListener('scrollend', callback);

win.add(listView);
win.open();
```
```JS

const rows = [];
  rows.push({ title: `Row #${i}` });
}

const tableView = Ti.UI.createTableView({ data: rows });

function callback(e) {
	console.log(`${e.type} properties: ${JSON.stringify(Object.keys(e))}`);
}

tableView.addEventListener('scrolling', callback);
tableView.addEventListener('scroll', callback);
tableView.addEventListener('scrollend', callback);

win.add(tableView);
win.open();
```

[TIMOB-28290](https://jira.appcelerator.org/browse/TIMOB-28290)
[TIMOB-28291](https://jira.appcelerator.org/browse/TIMOB-28291)
[TIMOB-28292](https://jira.appcelerator.org/browse/TIMOB-28292)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -65,6 +66,49 @@ public TiListView(ListViewProxy proxy)
...
+				} else if (velocityY < 0) {
+					payload.put(TiC.PROPERTY_DIRECTION, "up");
+				}
+
+				// Set scroll velocity.
+				payload.put(TiC.EVENT_PROPERTY_VELOCITY, velocityY);
+
+				proxy.fireSyncEvent(TiC.EVENT_SCROLLING, payload);
+				return true;
+			}
```
jquick-axway, line 109: Does doing this via `onFling()` cause the "scrolling" event to fire as you drag your finger? Because that's how it used to work.

And in Titanium 9.1.0, we did dampen the event firing based on the device's "touch slop" setting. This prevents a lot of up/down events from being spammed when you touch and hold on the ListView on a sensitive screen. You have to travel a minimum distance before firing the event.
https://github.com/appcelerator/titanium_mobile/pull/11680

garymathews: It only fires after lifting from a drag, where it can then calculate the velocity/fling of the gesture

jquick-axway: I re-tested this on Android with an older version of the Titanium SDK. The old Android behavior was that the "scrolling" event was only fired when you change vertical direction. You do not have to release your finger. Everytime you drag up/down with your finger on the screen you'll get this event. And the event is not spammed as you scroll in the same direction.

However, the iOS behavior is the "scrolling" event is only fired after you release your finger from the screen. So if you fling it and release your finger, you will get a scrolling event indication the direction it as flinged in. If you drag your finger, stop scrolling (don't fling), and then release your finger then you'll get a "scrolling" event with an "undefined" direction. This matches what we document below.
https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.ListView-event-scrolling

I personally prefer Android's old behavior compared to iOS because at least you can use it hide/show the top toolbar like how Google does it... but whatever.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -65,6 +66,49 @@ public TiListView(ListViewProxy proxy)
...
+					payload.put(TiC.PROPERTY_DIRECTION, "down");
+				} else if (velocityY < 0) {
+					payload.put(TiC.PROPERTY_DIRECTION, "up");
+				}
+
+				// Set scroll velocity.
+				payload.put(TiC.EVENT_PROPERTY_VELOCITY, velocityY);
+
+				proxy.fireSyncEvent(TiC.EVENT_SCROLLING, payload);
+				return true;
```
jquick-axway, line 108: I can't scroll the ListView with a fling anymore. I think it's because you're returning `true` here. You should probably return `false` instead.

jquick-axway:
@garymathews, I tested this PR out with my `ListViewCanScrollTest.js` attached to [TIMOB-28290](https://jira.appcelerator.org/browse/TIMOB-28290) and the behavior isn't quite right yet.

I'm getting multiple "scrollstart" events when I drag my finger up and down. I should only get this even once while dragging at the start. Also the "scrollend" event never gets fired.

Also the "scrolling" event's `direction` is flipped. When I scroll up I'm getting "down" and when I scroll down I'm getting "up".

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -65,6 +68,48 @@ public TiListView(ListViewProxy proxy)
...
+		{
+			@Override
+			public void onScrollStateChanged(@NonNull RecyclerView recyclerView, int newState)
+			{
+				super.onScrollStateChanged(recyclerView, newState);
+
+				if (newState == RecyclerView.SCROLL_STATE_IDLE) {
+					isScrolling = false;
+					proxy.fireSyncEvent(TiC.EVENT_SCROLLEND, generateScrollPayload());
+				}
```
jquick-axway, line 82: If you set `ListView.canScroll` to `false` and slide your finger on the `ListView`, it'll fire a "scrollend" event when it shouldn't. We probably should do the below...
```javascript
if (isScrolling && (newState == RecyclerView.SCROLL_STATE_IDLE)) {
	isScrolling = false;
	proxy.fireSyncEvent(TiC.EVENT_SCROLLEND, generateScrollPayload());
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -68,6 +72,45 @@ public TiTableView(TableViewProxy proxy)
...
+					// Obtain last touch position for `scrollend` event.
+					final TiDimension xDimension =
+						new TiDimension(nestedRecyclerView.getLastTouchX(), TiDimension.TYPE_WIDTH);
+					final TiDimension yDimension =
+						new TiDimension(nestedRecyclerView.getLastTouchY(), TiDimension.TYPE_HEIGHT);
+					payload.put(TiC.EVENT_PROPERTY_X, xDimension.getAsDefault(nestedRecyclerView));
+					payload.put(TiC.EVENT_PROPERTY_Y, yDimension.getAsDefault(nestedRecyclerView));
+
+					proxy.fireSyncEvent(TiC.EVENT_SCROLLEND, payload);
+				}
```
jquick-axway, line 96: If I set `TableView.scrollable` to `false`, then the "scrollend" fires when it shouldn't when I drag my finger over the `TableView`. We should probably add an `isScrolling` member variable and guard against this like in `ListView`.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -68,6 +72,45 @@ public TiTableView(TableViewProxy proxy)
...
+				// Update scroll offsets.
+				scrollOffsetX += dx;
+				scrollOffsetY += dy;
+
+				final KrollDict payload = generateScrollPayload();
+
+				proxy.fireSyncEvent(TiC.EVENT_SCROLL, payload);
+			}
+		});
+
```
jquick-axway, line 113: We document that `TableView`'s "scroll" event provides properties `firstVisibleItem`, `totalItemCount`, and `visibleItemCount`. All 3 of these event properties return values in Titanium 9.2.2, but with this PR only `firstVisibleItem` is supported. The `totalItemCount` and `visibleItemCount` event properties are undefined.
https://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.TableView-event-scroll

Side Note:
iOS does not support these 3 event properties. They're undefined on iOS.

jquick-axway (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -65,6 +68,48 @@ public TiListView(ListViewProxy proxy)
...
+			}
+
+			@Override
+			public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy)
+			{
+				super.onScrolled(recyclerView, dx, dy);
+
```
jquick-axway, line 89: I noticed that `onScrolled()` gets called on its own when the window opens. So I'm seeing "scrollstart" and "scrolling" events with "direction" set to `undefined` every time I open a window hosting the `ListView`.

I looked it up and it turns out `RecyclerView.onScroleld()` will be called when add/removing rows that are visibly onscreen. Maybe we should check if arguments `dx` and `dy` are zero before proceeding like the below? I haven't tried this so I'm not sure if this is right way to avoid it.
```java
if ((dx == 0) && (dy == 0)) {
    return;
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -68,6 +73,49 @@ public TiTableView(TableViewProxy proxy)
...
+					proxy.fireSyncEvent(TiC.EVENT_SCROLLEND, payload);
+				}
+			}
+
+			@Override
+			public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy)
+			{
+				super.onScrolled(recyclerView, dx, dy);
+
```
jquick-axway, line 106: Same comment as above. A "scrolling" event always happens when a window opens hosting a `TableView`. We may want to return out if `dx` and `dy` are zero.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -186,6 +234,59 @@ public void filterBy(String query)
...
+			new TiDimension(recyclerView.getMeasuredWidth(), TiDimension.TYPE_WIDTH);
+		final TiDimension heightDimension =
+			new TiDimension(recyclerView.getMeasuredHeight(), TiDimension.TYPE_HEIGHT);
+		size.put(TiC.PROPERTY_WIDTH, widthDimension.getAsDefault(recyclerView));
+		size.put(TiC.PROPERTY_HEIGHT, heightDimension.getAsDefault(recyclerView));
+		payload.put(TiC.PROPERTY_SIZE, size);
+
+		return payload;
+	}
+
```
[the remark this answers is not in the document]
garymathews: Yep, `visibleItemCount` isn't trivial with RecyclerView still trying to figure out an optimal way to get this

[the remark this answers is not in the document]
jquick-axway: If you think implementing `visibleItemCount` would cause performance issues (especially since scroll events will likely be spammed), then perhaps we should drop support for it. This might be okay since iOS doesn't support it either. If it's expensive, it might be better to provide a method to fetch it instead and for devs to call when a "scrollend" fires.

jquick-axway (approved):
CR/FR: Pass

ssjsamir (approved):
**FR Passed**, Tested using the code in the description and able to see the mentioned events, also tested that `bindid` is visible with the following code. 

```
	backgroundColor:'white'
});


var myTemplate = {
    properties: {
        height: Ti.UI.SIZE,
        width: Ti.UI.FILL
    },
    childTemplates: [
        {
            type: 'Ti.UI.View',
             bindId: 'outerView',
            properties: {
                backgroundColor: '#fcf',
                height: Ti.UI.SIZE,
                //layout: 'vertical',
                left: 15,
                right: 15,
                width: Ti.UI.FILL
            },
            childTemplates: [
                {
                    type: 'Ti.UI.Label',
                    bindId: 'idClick',
                    properties: {
                        color: 'black',
                        height: "150dp",
                        font: { fontFamily:'Arial', fontSize: '20dp', fontWeight:'bold' },
                        //left: 0
                    }
                },

            ]
        }
    ]
};

// Create the list view
    templates: { 'template': myTemplate },
    top: Ti.Platform.osname === 'iphone' ? 20 : 0   // Allow for iOS status bar
});

// Create the list data set
var fruitSection = Ti.UI.createListSection({ });
var fruitDataSet = [
    { idClick: {text: 'Click me!!!'}},
];

// Add the data set to the list view
var sections = [fruitSection];
listView.setSections(sections);

listView.addEventListener('itemclick', function(e) {
	Ti.API.info("BindId: " + e.bindId +".."+JSON.stringify(e))
	alert(
        "BindId: " + e.bindId
    );


});

// Add the list view to the main window and open it
win.add(listView);
win.open();

```

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.3
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Nexus 5X API 29
```

PR #12358, opened 2020-12-27
feat(android): create WebP images

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6657

**Optional Description:**

Create WebP images on Android (lower filesize, needed for e.g. Whatsapp Sticker packs,...)

**Example**
```javascript

// Place an image named "img.jpg" in the root folder

	title: "img",
	backgroundColor: '#fff',
	layout: "vertical"
});

var file = Ti.Filesystem.getFile("img.jpg");

var img1 = Ti.UI.createImageView({
	image: file,
	width: 400,
	height: Ti.UI.SIZE,
	top: 10
});
win.add(img1);
console.log("JPG filesize: ", file.size);

// create webp
var blob = file.read();
var webp = blob.imageAsWebP(0.5);
console.log("WebP filesize: ", webp.size);
	image: webp,
	width: 400,
	height: Ti.UI.SIZE,
	top: 10
});
win.add(img);

win.open();
```

Screenshot_20201227-151342
[INFO]  JPG filesize:  871399
[INFO]  WebP filesize:  176528

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
@@ -873,6 +873,60 @@ public class TiBlob extends KrollProxy
 	}
 
 	@Kroll.method
+	public TiBlob imageAsWebp(Number compressionQuality)
+	{
+		Bitmap img = getImage();
+		if (img == null) {
+			return null;
+		}
+
+		float quality = 1f;
+		if (compressionQuality != null) {
+			quality = compressionQuality.floatValue();
+		}
+
+		TiBlob result = null;
+		ByteArrayOutputStream bos;
+
+		try {
+			bos = new ByteArrayOutputStream();
+			if (img.compress(
+				quality == 1f ? CompressFormat.WEBP_LOSSLESS : CompressFormat.WEBP_LOSSY, (int) (quality * 100), bos)
+			) {
+				byte[] data = bos.toByteArray();
+
+				BitmapFactory.Options bfOptions = new BitmapFactory.Options();
+				bfOptions.inPurgeable = true;
+				bfOptions.inInputShareable = true;
+
+				result = TiBlob.blobFromData(data, "image/webp");
+			}
+		} catch (OutOfMemoryError e) {
+			TiBlobLruCache.getInstance().evictAll();
+			Log.e(TAG, "Unable to get the thumbnail image. Not enough memory: " + e.getMessage(), e);
+		} catch (IllegalArgumentException e) {
+			Log.e(TAG, "Unable to get the thumbnail image. Illegal Argument: " + e.getMessage(), e);
+		} catch (Throwable t) {
+			Log.e(TAG, "Unable to get the thumbnail image. Unknown exception: " + t.getMessage(), t);
+		} finally {
+			// [MOD-309] Free up memory to work around issue in Android
+			if (img != null) {
+				img.recycle();
+				img = null;
+			}
+			bos = null;
+
+			// Perform soft garbage collection to reclaim memory.
+			KrollRuntime instance = KrollRuntime.getInstance();
+			if (instance != null) {
+				instance.softGC();
+			}
+		}
+
+		return result;
+	}
+
+	@Kroll.method
 	public TiBlob imageAsThumbnail(Number size, @Kroll.argument(optional = true) Number borderSize,
 								   @Kroll.argument(optional = true) Number cornerRadius)
 	{
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
@@ -873,7 +873,7 @@ public class TiBlob extends KrollProxy
 	}
 
 	@Kroll.method
-	public TiBlob imageAsWebp(Number compressionQuality)
+	public TiBlob imageAsWebP(Number compressionQuality)
 	{
 		Bitmap img = getImage();
 		if (img == null) {
--- a/apidoc/Titanium/Blob.yml
+++ b/apidoc/Titanium/Blob.yml
@@ -241,6 +241,22 @@ methods:
         optional: true
         default: 0
 
+  - name: imageAsWebp
+    summary: Creates a new blob by compressing the underlying image to the specified quality.
+    description: |
+        Returns the compressed or uncompressed WebP image as a blob.
+
+        If this blob doesn't represent an image, returns `null`.
+    platforms: [android]
+    since: {andorid: "10.0.0"}
+    parameters:
+      - name: quality
+        type: Number
+        summary: Quality to compress this image to. From 0.0 (lowest quality) to 1.0 (highest quality, lossless WebP).
+    returns:
+        type: Titanium.Blob
+        summary: Compressed WebP image as a blob.
+
   - name: imageWithAlpha
     summary: Returns a copy of the underlying image with an added alpha channel.
     description: |
--- a/apidoc/Titanium/Blob.yml
+++ b/apidoc/Titanium/Blob.yml
@@ -248,7 +248,7 @@ methods:
 
         If this blob doesn't represent an image, returns `null`.
     platforms: [android]
-    since: {andorid: "10.0.0"}
+    since: {android: "10.0.0"}
     parameters:
       - name: quality
         type: Number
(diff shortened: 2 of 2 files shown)


apidoc/Titanium/Blob.yml
```diff
@@ -241,6 +241,22 @@ methods:
...
         default: 0
 
+  - name: imageAsWebp
+    summary: Creates a new blob by compressing the underlying image to the specified quality.
+    description: |
+        Returns the compressed or uncompressed WebP image as a blob.
+
+        If this blob doesn't represent an image, returns `null`.
+    platforms: [android]
+    since: {andorid: "10.0.0"}
```
hansemannn, line 251: > android

apidoc/Titanium/Blob.yml
```diff
@@ -241,6 +241,22 @@ methods:
         optional: true
         default: 0
 
+  - name: imageAsWebp
```
hansemannn, line 244: imageAsWebP ?

m1ga: yeah, you are right. Looks better

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -872,6 +872,60 @@ public TiBlob imageAsCompressed(Number compressionQuality)
 		return result;
 	}
 
+	@Kroll.method
+	public TiBlob imageAsWebP(Number compressionQuality)
```
sgtcoolguy, line 876:
```suggestion
	public TiBlob imageAsWebP(@Kroll.argument(optional = true) Number compressionQuality)
```

apidoc/Titanium/Blob.yml
```diff
@@ -241,6 +241,22 @@ methods:
         optional: true
         default: 0
 
+  - name: imageAsWebP
+    summary: Creates a new blob by compressing the underlying image to the specified quality.
+    description: |
+        Returns the compressed or uncompressed WebP image as a blob.
```
sgtcoolguy, line 247:
```suggestion
        Returns the compressed or uncompressed WebP image as a [blob](Titanium.Blob).
```

apidoc/Titanium/Blob.yml
```diff
@@ -241,6 +241,22 @@ methods:
...
+
+        If this blob doesn't represent an image, returns `null`.
+    platforms: [android]
+    since: {android: "10.0.0"}
+    parameters:
+      - name: quality
+        type: Number
```
sgtcoolguy, line 254:
```suggestion
        type: Number
        optional: true
        default: 1
```

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -872,6 +872,60 @@ public TiBlob imageAsCompressed(Number compressionQuality)
...
+
+		TiBlob result = null;
+		ByteArrayOutputStream bos;
+
+		try {
+			bos = new ByteArrayOutputStream();
+			if (img.compress(
+				quality == 1f ? CompressFormat.WEBP_LOSSLESS : CompressFormat.WEBP_LOSSY, (int) (quality * 100), bos)
+			) {
+				byte[] data = bos.toByteArray();
```
sgtcoolguy, line 896: I would immediately try and free the image memory here before taking the byte array and generating a blob. Also, the whole segment feels very much like it should be part of `TiBlob.blobFromImage`. That's currently hard-coded to assume PNG/JPEG based on alpha, but probably should be refactored to have another variant where you explicitly specify the format and quality.

Otherwise, I'm not sure `width`/`height`/`uprightWidth`/`uprightHeight` will be set properly on the blob.

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -872,6 +872,60 @@ public TiBlob imageAsCompressed(Number compressionQuality)
...
+
+				BitmapFactory.Options bfOptions = new BitmapFactory.Options();
```
sgtcoolguy, line 898: This object does not appear to be used at all.

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
sgtcoolguy: Additionally, though our current APIs are sync around image cropping/modifications, the Android docs suggest that the compression/conversion of a bitmap can be quite slow. I think this method should explicitly be async (i.e. take a callback and/or return a Promise)

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -872,6 +872,60 @@ public TiBlob imageAsCompressed(Number compressionQuality)
...
+				bfOptions.inPurgeable = true;
+				bfOptions.inInputShareable = true;
+
+				result = TiBlob.blobFromData(data, "image/webp");
+			}
+		} catch (OutOfMemoryError e) {
+			TiBlobLruCache.getInstance().evictAll();
+			Log.e(TAG, "Unable to get the thumbnail image. Not enough memory: " + e.getMessage(), e);
```
sgtcoolguy, line 906: The error messages are misleading, you're not creating a thumbnail image - you're converting to a WebP image.

sgtcoolguy (changes requested):
Interesting idea. I'm worried about a few things:
- We made the mistake int he past of making these blob image manipulations sync, but they can be memory and CPU intensive and should be done async.
- This looks very similar to `imageAsCompressed`, which effectively operates as an equivalent `imageAsJPEG`. Perhaps we need a more general API where you can specify the image format *and* quality?
- I suspect this and `imageAsCompressed` should be making use of `TiBlob.blobFromImage` to ensure the objects properties are all set properly. That likely involves refactoring that method to support explicit mimetype/format and quality arguments.
- No iOS compatibility.

cc @jquick-axway

m1ga:
@sgtcoolguy I basically copied `imageAsCompressed` and just change the WebP parts (e.g. bfOptions is not used inside `imageAsCompressed` and the parameters are deprecated) :) 
You are right, a basic "convertTo" method with the image type as a parameter would be way better and more extendable in the future. 
I could move it to something like `convertImage()` and call it using the other functions so existing apps would still work

jquick-axway:
I agree with @sgtcoolguy . It makes sense to do this via the existing `imageAsCompressed()` method and have it support a format type argument.

I'm thinking we should add the following constants...
```javascript
Ti.Media.IMAGE_FORMAT_JPEG
Ti.Media.IMAGE_FORMAT_PNG
Ti.Media.IMAGE_FORMAT_WEBP
```

And set up `imageAsCompressed()` to support a dictionary argument so that we can add other options to it. And perhaps add a method indicating if the compression format is supported.
```javascript
if (!myBlob.canCompressToImage(Ti.Media.IMAGE_FORMAT_JPEG)) {
	return;
}

const newBlob = myBlob.imageAsCompressed({
	async: false, // Set true to return a promise. Defaults to false.
	quality: 0.8,
	format: Ti.Media.IMAGE_FORMAT_JPEG,
	exif: { /* Can add this in the future. */ },
});
```

I'm not sure if I dig the "async" option or not. I'm just tossing in the idea. I do prefer us making this a dictionary. If we were to ever support `IMAGE_FORMAT_BMP` in the future (not technically a compressed format), then we may want an option to set the RGBA byte order... but this may be overthinking it... and I would probably want to stream a BMP directly to file anyways to avoid out-of-memory issues.

@m1ga , we're not asking you to do any of this. You did bring up a good discussion point. :smile:
Especially since we plan on adding promise support to our existing async APIs.

sgtcoolguy:
@jquick-axway To think out loud here, I also don't love the `async` property in the dictionary. I think if we receive a dictionary we should return a `Promise` (as the "new" API call format). If we receive no args or simple `Number`, assume the old legacy sync call (which we should deprecate). But that assumes we're overloading the existing API method/name, which doesn't necessarily make sense here: `blob.imageAsCompressed()`. I know that Android calls the underlying method `compress` but this is really a conversion method that can alter quality and/or file format, so a name like `imageConvert()` (keeps image first like our existing image manipulation methods) or `convertImage()` (reads better) may make more sense.

m1ga:
close this one since it is included in ti.imagefactory (https://github.com/tidev/ti.imagefactory/pull/192)

PR #12360, opened 2021-01-01
fix(ios): fix possible NSNotification crashes

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-0000 (TBA)


iphone/Classes/TiAppiOSProxy.m
```diff
@@ -223,6 +223,10 @@ - (void)didTakeScreenshot:(NSNotification *)info
 
 - (void)didReceiveApplicationShortcutNotification:(NSNotification *)info
 {
+  if (!info.userInfo) {
```
vijaysingh-axway, line 226: I think it is not required to check. The [place](https://github.com/appcelerator/titanium_mobile/blob/2c2d7008f837da6b19c106414441cf7fe5ea9055/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L1262)   from where notification is fired, userInfo is always set and is required. I think it is intentional to not check for info.userInfo to nil. It should never nil. In case it is nil it means there is issue in the Titanium SDK and it need to be fixed.

iphone/Classes/TiAppiOSProxy.m
```diff
@@ -253,7 +257,7 @@ - (id)UserNotificationCenter
 
 - (void)didHandleURL:(NSNotification *)info
 {
-  if (![self _hasListeners:@"handleurl"]) {
+  if (![self _hasListeners:@"handleurl"] || !info.userInfo) {
```
vijaysingh-axway, line 260: Same as above. In this case also userInfo always set. See [here](https://github.com/appcelerator/titanium_mobile/blob/2c2d7008f837da6b19c106414441cf7fe5ea9055/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L437).

vijaysingh-axway (changes requested):
The changes are not required. See the comments.

iphone/Classes/TiAppiOSProxy.m
hansemannn: The issue is happening when the app is not booted. See the crash logs that definitely crash here. Using the fix it runs stable.

vijaysingh-axway: As per code, if app is not booted launchOption is added in queuedBootEvents https://github.com/appcelerator/titanium_mobile/blob/2c2d7008f837da6b19c106414441cf7fe5ea9055/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L439 . And after app get booted this is used as userInfo in posting NSNotification as mentioned https://github.com/appcelerator/titanium_mobile/blob/2c2d7008f837da6b19c106414441cf7fe5ea9055/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L243 . So it should have userInfo always.

Can you attach crash log please?

Though adding extra condition will not create any problem. I am just thinking if there is any other issue.

PR #12363, opened 2021-01-04
fix(android): optimize list view updates

garymathews:
- ~Process list/table items on a background thread~
- Optimize pre-loading of list/table items
- Only update list/table items when necessary
- Implement `noresults` for TableView
- Amend `marker` event to fire on partially visible rows for consistency

##### TEST CASE
- Project on JIRA

[TIMOB-28294](https://jira.appcelerator.org/browse/TIMOB-28294)
[TIMOB-28308](https://jira.appcelerator.org/browse/TIMOB-28308)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
@@ -7,10 +7,7 @@
 package ti.modules.titanium.ui.widget.listview;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.Queue;
-import java.util.concurrent.LinkedBlockingQueue;
 
 import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.titanium.TiApplication;
@@ -49,7 +46,7 @@ public class TiListView extends TiSwipeRefreshLayout implements OnSearchChangeLi
 
 	private final ListViewAdapter adapter;
 	private final DividerItemDecoration decoration;
-	private final List<ListItemProxy> items = Collections.synchronizedList(new ArrayList<>(CACHE_SIZE));
+	private final List<ListItemProxy> items = new ArrayList<>(CACHE_SIZE);
 	private final ListViewProxy proxy;
 	private final TiNestedRecyclerView recyclerView;
 	private final SelectionTracker tracker;
@@ -58,8 +55,6 @@ public class TiListView extends TiSwipeRefreshLayout implements OnSearchChangeLi
 	private boolean isScrolling = false;
 	private int lastScrollDeltaY;
 	private String filterQuery;
-
-	private final Queue<Thread> updateQueue = new LinkedBlockingQueue<>(2);
 
 	public TiListView(ListViewProxy proxy)
 	{
@@ -308,12 +303,9 @@ public class TiListView extends TiSwipeRefreshLayout implements OnSearchChangeLi
 	 */
 	public ListItemProxy getRowByIndex(int index)
 	{
-		synchronized (this.items)
-		{
-			for (ListItemProxy item : this.items) {
-				if (item.index == index) {
-					return item;
-				}
+		for (ListItemProxy item : this.items) {
+			if (item.index == index) {
+				return item;
 			}
 		}
 		return null;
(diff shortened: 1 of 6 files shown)


jquick-axway:
Updating the `items` member variable from another thread is going to cause race conditions since other parts of the code has unsynchronized access to it. Particularly the `filterBy()` method and the adapter.

It would probably be safer to create a new items collection instance on another thread and then replace the `items` member variable with it once done, but this would mean the adapter would have allow items to be replaced too. If we were to do it this way, then we should call `trimToSize()` on the old items list to compact it in memory in case it's huge.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -492,57 +485,45 @@ private void update(String query)
...
+		// FIXME: This is not an ideal workaround for an issue where recycled rows that were in focus
 		//        lose their focus when the data set changes. There are improvements to be made here.
 		//        This can be reproduced when setting a Ti.UI.TextField in the Ti.UI.ListView.headerView for search.
-		new Handler().post(new Runnable()
-		{
-			public void run()
+		final Activity activity = TiApplication.getAppCurrentActivity();
+		final View previousFocus = activity.getCurrentFocus();
+
+		if (activity != null && previousFocus != null) {
```
jquick-axway, line 512: We should check if `activity` is `null` before calling `activity.getCurrentFocus()` above.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -516,53 +514,45 @@ public void update(String query)
...
-		// FIXME: This is not an ideal workaround for an issue where recycled rows that were in focus
+		// FIXME: This is not an ideal workaround for an issue where recycled items that were in focus
 		//        lose their focus when the data set changes. There are improvements to be made here.
-		//        This can be reproduced when setting a Ti.UI.TextField in the Ti.UI.TableView.headerView for search.
-		new Handler().post(new Runnable()
-		{
-			public void run()
+		//        This can be reproduced when setting a Ti.UI.TextField in the Ti.UI.ListView.headerView for search.
+		final Activity activity = TiApplication.getAppCurrentActivity();
+		final View previousFocus = activity.getCurrentFocus();
```
jquick-axway, line 539: We should check if `activity` is `null` first.
```java
final View previousFocus = (activity != null) ? activity.getCurrentFous() : null;
```

jquick-axway (approved):
CR: Pass

ssjsamir (commented):
@garymathews The test case mentioned in https://jira.appcelerator.org/browse/TIMOB-28294 seems to now work fine and be more 'snappy' but with the following test case I am still not seeing '@@@ noresults' being logged in the console. 

```
function createListViewSectionItems() {
var items = [];
for (var index = 1; index <= 100; index++) {
//		items.push({ properties: { title: "Row " + index } });
items.push({ properties: { title: "Row " + index, searchableText: "Row " + index } });
//		items.push({ properties: { title: "Row " + index, searchableText: "Test " + index } });
}
return items;
}
 
headerTitle: "ListView",
items: createListViewSectionItems(),
})],
searchView: Ti.UI.createSearchBar({ showCancel: true }),
//	searchView: Ti.UI.Android.createSearchView(),
});
listView.addEventListener("noresults", () => {
Ti.API.info("@@@ noresults");
});
window.add(listView);
window.open();
```

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.3
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Pixel XL API 30 
Pixel XL API 29
```

ssjsamir (approved):
FR Passed: Able to see correct logging now when a search result is not found using the test case mentioned above also still able to the run the test case mentioned in https://jira.appcelerator.org/browse/TIMOB-28294

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.3
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Pixel XL API 30 
Pixel XL API 29
```

PR #12368, opened 2021-01-05
Refactor Android clipboard

sgtcoolguy:

**Optional Description:**
This modifies our `Ti.UI.Clipboard` to use the newer `android.content.ClipboardManager` API under the hood. The old `android.text.ClipbaordManager` is text-only and was deprecated in API level 11.

This *does not* try to add support for non-text data (i.e. uri list, intent, html), but does pave the way to using the newer API that is capable of doing some of that.

This was prompted by the occasional unit tests failures on the newer API level 30 emulator. I'm not sure if the newer API would help "fix" that, or if it's due to some app focus issue or something intermittently.


m1ga:
Superseded by https://github.com/tidev/titanium_mobile/pull/13105

PR #12379, opened 2021-01-08
chore(iOS): support ti symbols via xcconfig file to work with swift

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28297

This PR adds generated ti symbol macros in build settings of project for Objc and swift both and removes defines.h, which is not required anymore.


iphone/cli/commands/_build.js
```diff
@@ -3132,33 +3137,30 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 		},
 		legacySwift = version.lt(this.xcodeEnv.version, '8.0.0');
 
+	this.gccDefs = 'DEPLOYTYPE=' + this.deployType;
```
sgtcoolguy, line 3140: Why move from an array to a string here? Feels odd to have to remember to add leading spaces to it. Also, this seems like exactly what a `Map` is for: key/value pairs. Why not hold a `Map` with the pairs of key/values and then render it as a string only when needed to write the file?

iphone/cli/commands/_build.js
```diff
@@ -3132,33 +3137,30 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
...
 	// set additional build settings
 	if (this.target === 'simulator' || this.target === 'macos') {
-		gccDefs.push('__LOG__ID__=' + this.tiapp.guid);
-		gccDefs.push('DEBUG=1');
-		gccDefs.push('TI_VERSION=' + this.titaniumSdkVersion);
+		this.gccDefs += ' __LOG__ID__=' + this.tiapp.guid;
```
sgtcoolguy, line 3143: Use templated strings, or as suggested above use key/value pairs to stick in a `Map` instead.

iphone/cli/commands/_build.js
```diff
@@ -3132,33 +3137,30 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 		},
 		legacySwift = version.lt(this.xcodeEnv.version, '8.0.0');
 
+	this.gccDefs = 'DEPLOYTYPE=' + this.deployType;
 	// set additional build settings
 	if (this.target === 'simulator' || this.target === 'macos') {
```
sgtcoolguy, line 3142: Are these really build settings tied to the target or the deploy type? Seems like `DEBUG` should be true for `'development'` deploy type. The log id is used for writing our log file, so that seems tied to the target here. Not sure about `TI_VERSION`, but it looks to me like we set it explicitly in `writeXcodeConfigFiles()` anyways.

iphone/cli/commands/_build.js
```diff
@@ -4500,6 +4502,8 @@ iOSBuilder.prototype.writeXcodeConfigFiles = function writeXcodeConfigFiles() {
 			'OTHER_LDFLAGS[sdk=iphonesimulator*]=$(inherited) $(JSCORE_LD_FLAGS)',
 			'OTHER_LDFLAGS[sdk=iphoneos9.*]=$(inherited) -weak_framework Contacts -weak_framework ContactsUI -weak_framework WatchConnectivity -weak_framework CoreSpotlight',
 			'OTHER_LDFLAGS[sdk=iphonesimulator9.*]=$(inherited) -weak_framework Contacts -weak_framework ContactsUI -weak_framework WatchConnectivity -weak_framework CoreSpotlight',
+			'GCC_DEFINITIONS=' + this.gccDefs,
```
sgtcoolguy, line 4505: Again, here's where I'd turn the `Map` to a `string`, basically just joining key/value pairs with `'='` and entries with `' '` as a naive approach. (I assume we may eventually need to quote some values if they have spaces or something).

iphone/cli/commands/_build.js
vijaysingh-axway: These settings were already there. TI_VERSION is getting duplicate. So I removed it in my latest commit.

sgtcoolguy (approved):

lokeshchdhry:
FR Passed.

PR #12349, opened 2020-12-17
feat: return Promises for async Ti.Geolocation methods

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28279

**Description:**
This relates to #10920 and #10554

This modifies some of our async methods to also return a `Promise`:
- `Ti.Geolocation.getCurrentHeading()`
- `Ti.Geolocation.getCurrentPosition()`

In addition, this adds a number of bug fixes under the covers:
- fix(ios): report error async to callers of Ti.Geolocation#requestTemporaryFullAccuracyAuthorization if missing purpose in plist
- fix(ios): if user supplies no specific auth constant for hasLocationPermissions, assume WHEN_IN_USE
- fix(ios): shape of error object and reporting errors for getCurrentPosition() or has/requestLocationPermissions() calls
- fix(android): calling `getCurrentPosition()` without any `location` event listeners would mess up our state leading to various bugs, but most obviously if you called `getCurrentPosition()` twice, the 2nd+ calls would never fire the callback Promise (because they wouldn't do anything)

Some other changes:
- add more Info.plist key/value pairs int he tests app for geolocation related permissions
- use a binary from Lyft to preemptively set a specific gps location on the simulator for iOS (so we'd get a location/position!)

PR #12407, opened 2021-01-22
feat: return Promises for Ti.Window open()/close() methods

sgtcoolguy:
**JIRA:** https://github.com/appcelerator/titanium_mobile/pull/12407

**Description:**
Relates to #12349, #12407, #12417

This modifies some of our async methods to also return a Promise:

* `Ti.UI.Window.open()`
* `Ti.UI.Window.close()`
* `Ti.UI.NavigationWindow.open()`
* `Ti.UI.NavigationWindow.close()`
* `Ti.UI.TabGroup.open()`
* `Ti.UI.TabGroup.close()`

Unit tests are included!

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
@@ -1,11 +1,12 @@
 /**
  * Appcelerator Titanium Mobile
- * Copyright (c) 2018 by Axway, Inc. All Rights Reserved.
+ * Copyright (c) 2018-2021 by Axway, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 package ti.modules.titanium.ui;
 
+import org.appcelerator.kroll.KrollPromise;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.proxy.TiWindowProxy;
@@ -27,7 +28,7 @@ public class NavigationWindowProxy extends WindowProxy
 
 	@Override
 	@Kroll.method
-	public void open(@Kroll.argument(optional = true) Object arg)
+	public KrollPromise<Void> open(@Kroll.argument(optional = true) Object arg)
 	{
 		// FIXME: Shouldn't this complain/blow up if window isn't specified?
 		if (!opened && getProperties().containsKeyAndNotNull(TiC.PROPERTY_WINDOW)) {
@@ -37,9 +38,11 @@ public class NavigationWindowProxy extends WindowProxy
 				openWindow(rootView, arg);
 				fireEvent(TiC.EVENT_OPEN, null);
 			}
-			return;
+			return KrollPromise.create((promise) -> {
+				promise.resolve(null);
+			});
 		}
-		super.open(arg);
+		return super.open(arg);
 	}
 
 	@Kroll.method
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
@@ -6,6 +6,7 @@
  */
 package ti.modules.titanium.ui;
 
+import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollPromise;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.titanium.TiC;
@@ -13,6 +14,8 @@ import org.appcelerator.titanium.proxy.TiWindowProxy;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import androidx.annotation.NonNull;
 
 @Kroll.proxy(creatableInModule = UIModule.class)
 public class NavigationWindowProxy extends WindowProxy
@@ -55,17 +58,19 @@ public class NavigationWindowProxy extends WindowProxy
 		}
 	}
 
-	@Override
-	@Kroll.method
-	public void close(@Kroll.argument(optional = true) Object arg)
+	protected void handleClose(@NonNull KrollDict options)
 	{
 		if (opened) {
 			opened = false;
-			popToRootWindow(arg);
-			closeWindow(windows.get(0), arg); // close the root window
+			popToRootWindow(options);
+			closeWindow(windows.get(0), options); // close the root window
 			fireEvent(TiC.EVENT_CLOSE, null);
+			if (closePromise != null) {
+				closePromise.resolve(null);
+				closePromise = null;
+			}
 		}
-		super.close(arg);
+		super.handleClose(options);
 	}
 
 	@Kroll.method
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/TiWindowProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/TiWindowProxy.java
@@ -73,7 +73,8 @@ public abstract class TiWindowProxy extends TiViewProxy
 	protected boolean windowActivityCreated = false;
 	protected List<Pair<View, String>> sharedElementPairs;
 	public TiWindowProxy navigationWindow;
-	private KrollPromise openPromise;
+	protected KrollPromise openPromise;
+	protected KrollPromise closePromise;
 
 	public interface PostOpenListener {
 		void onPostOpen(TiWindowProxy window);
@@ -107,10 +108,10 @@ public abstract class TiWindowProxy extends TiViewProxy
 				promise.reject(new Throwable("Window is already opened or opening."));
 			});
 		}
+		opening = true;
+		waitingForOpen = new WeakReference<TiWindowProxy>(this);
 
 		openPromise = KrollPromise.create((promise) -> {
-			waitingForOpen = new WeakReference<TiWindowProxy>(this);
-			opening = true;
 			KrollDict options = null;
 			TiAnimation animation = null;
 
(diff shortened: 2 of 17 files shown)


sgtcoolguy:
The relevant section of the crash report produced is:
```
Exception Type:        EXC_BREAKPOINT (SIGTRAP)
Exception Codes:       EXC_ARM_BREAKPOINT at 0x00000001902e7c04 (brk 0)
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Trace/BPT trap: 5
Termination Reason:    Namespace SIGNAL, Code 0x5
Terminating Process:   exc handler [28642]

Application Specific Information:
CoreSimulator 732.18.6 - Device: iPhone 12 Pro Max (33F48821-7C57-4268-9B0F-74C71EDCC6E9) - Runtime: iOS 14.3 (18C61) - DeviceType: iPhone 12 Pro Max

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.apple.JavaScriptCore      	0x00000001902e7c04 WTFCrashWithInfo(int, char const*, char const*, int) + 20
1   com.apple.JavaScriptCore      	0x00000001907a2e6c JSC::LocalAllocator::allocateSlowCase(JSC::Heap&, JSC::GCDeferralContext*, JSC::AllocationFailureMode) + 560
2   com.apple.JavaScriptCore      	0x0000000190a9f12c void* JSC::allocateCell<JSC::ErrorInstance>(JSC::Heap&, unsigned long) + 172
3   com.apple.JavaScriptCore      	0x0000000190a9a7e4 JSC::createTypeError(JSC::JSGlobalObject*, WTF::String const&, WTF::String (*)(WTF::String const&, WTF::String const&, JSC::RuntimeType, JSC::ErrorInstance::SourceTextWhereErrorOccurred), JSC::RuntimeType) + 72
4   com.apple.JavaScriptCore      	0x0000000190aa121c JSC::createError(JSC::JSGlobalObject*, JSC::JSValue, WTF::String const&, WTF::String (*)(WTF::String const&, WTF::String const&, JSC::RuntimeType, JSC::ErrorInstance::SourceTextWhereErrorOccurred)) + 332
5   com.apple.JavaScriptCore      	0x0000000190aa2018 JSC::createNotAnObjectError(JSC::JSGlobalObject*, JSC::JSValue) + 64
6   com.apple.JavaScriptCore      	0x0000000190aecfa0 JSC::JSValue::synthesizePrototype(JSC::JSGlobalObject*) const + 216
7   com.apple.JavaScriptCore      	0x0000000190b65b04 JSC::JSPromise::createDeferredData(JSC::JSGlobalObject*, JSC::JSPromiseConstructor*) + 2064
8   com.apple.JavaScriptCore      	0x00000001902f48a8 JSObjectMakeDeferredPromise + 76
9   com.appcelerator.TitaniumKit  	0x0000000104bc784c -[KrollPromise initInContext:] + 172 (KrollPromise.m:22)
10  com.appcelerator.TitaniumKit  	0x0000000104bbece4 -[TiWindowProxy close:] + 480 (TiWindowProxy.m:330)
11  com.appcelerator.testApp.testing	0x00000001046d99ac -[TiUITabProxy closeWindowProxy:animated:] + 408 (TiUITabProxy.m:226)
12  com.appcelerator.testApp.testing	0x00000001046d8d4c __30-[TiUITabProxy cleanNavStack:]_block_invoke + 992 (TiUITabProxy.m:103)
13  com.appcelerator.testApp.testing	0x00000001046d8960 -[TiUITabProxy cleanNavStack:] + 116 (TiUITabProxy.m:85)
14  com.appcelerator.testApp.testing	0x00000001046d9808 -[TiUITabProxy removeFromTabGroup] + 136 (TiUITabProxy.m:201)
15  com.appcelerator.testApp.testing	0x00000001047168f0 -[TiUITabGroupProxy dealloc] + 272 (TiUITabGroupProxy.m:31)
16  com.appcelerator.TitaniumKit  	0x0000000104bca260 -[KrollObject dealloc] + 48 (KrollObject.m:460)
17  com.apple.JavaScriptCore      	0x00000001902e3bf0 JSC::JSCallbackObject<JSC::JSNonFinalObject>::destroy(JSC::JSCell*) + 112
```

sgtcoolguy:
After a lot of messing around my best guess here is that internal code is calling window close, and is not "consuming" the promise generated (because it's not actually surfaced to the JS engine/user) and JSC is freaking out because we have unhandled promise rejections/fulfillments. So... I'm not sure what the "fix" here would be. Perhaps to explicitly delineate the JS-facing method from the method to be used internally? Flag in some way so the `KrollPromise` creates a dummy variant that doesn't generate a real JS Promise when `init` from an internal API call and not the JS facing API surface?

hansemannn:
@sgtcoolguy Why would you add a tab group to a navigation window? They are both top-level containers and not meant to be added to each other, which would explain the crash.

sgtcoolguy:
@hansemannn developers do crazy things. It's in our test suite as a regression check because we have customers who have that code!

Anyways, I narrowed this down eventually. The crash is specifically coming from when the old style `TiProxy` hierarchy gets `KrollFinalizer` called on some object. At that point trying to use the JSC API to create a new `Promise` or `Error` crashes. So I basically have to toggle a global static `BOOL` while `KrollFinalizer` is happening and be sure not to try and actually create a "real" Promise while it is (basically make the `Promise` a no-op under the hood - create nothing, do nothing when rsolve/reject are called).

The second thing which was my initial hunch is that if I create a `Promise` under the hood in obj-c that doesn't get exposed out to the developer (because say we call `open`/`close` on a window from internal Tab code) and we do not handle the rejection/fulfillment of the Promise the engine gets wonky. I don't think this causes a crash, but it does cause an unhandled promise rejection error to get set on the context and there's no official API to actually handle that yet (they have a new method on `main` branch now, but it's unreleased). So I needed to add a way to "flush" internal Promises so that we handle them internally with noop handlers. I assume with a lot of playing around perhaps we could hack a way to detect this and fire off an `unhandledpromiserejection` event ala Node/V8. I'm not entirely sure yet (like when do we check for the error, how do we expose it, how do we "reset" it).

hansemannn:
Customers … 😄

android/titanium/src/java/org/appcelerator/titanium/proxy/TiWindowProxy.java
```diff
@@ -189,6 +201,10 @@ public void closeFromActivity(boolean activityIsFinishing)
 		// And it will dispose the handler of the window in the JS if the activity
 		// is not forced to destroy.
 		fireSyncEvent(TiC.EVENT_CLOSE, data);
+		if (closePromise != null) {
+			closePromise.resolve(null);
+			closePromise = null; // FIXME: call release() first?
```
garymathews, line 206: I think this cleanup could be moved to `release()`

```Java
@Override
public void release()
{
	if (openPromise != null) {
		openPromise = null;
	}
	if (closePromise != null) {
		closePromise = null;
	}

	super.release();
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
```diff
@@ -54,15 +57,16 @@ public void popToRootWindow(@Kroll.argument(optional = true) Object arg)
...
-	public void close(@Kroll.argument(optional = true) Object arg)
+	public KrollPromise<Void> close(@Kroll.argument(optional = true) Object arg)
 	{
+		// FIXME: don't do this stuff sync first!
 		if (opened) {
 			opened = false;
 			popToRootWindow(arg);
 			closeWindow(windows.get(0), arg); // close the root window
-			fireEvent(TiC.EVENT_CLOSE, null);
+			fireEvent(TiC.EVENT_CLOSE, null); // FIXME: what about close promise?
```
garymathews, line 67: This might be a good oppertunity to move the `fireEvent(TiC.EVENT_CLOSE, ...` to [TiWindowProxy.java](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/proxy/TiWindowProxy.java#L165)

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -202,32 +211,39 @@ - (BOOL)isRootViewAttached
...
+    return [KrollPromise rejectedWithErrorMessage:@"Window is already opened or opening." inContext:context];
   }
 
-  //Lets keep ourselves safe
+  // Lets keep ourselves safe
   [self rememberSelf];
 
+  // Don't recreate if we're calling ourselves again because root view is not loaded/attached!
+  if (openPromise == nil) {
+    openPromise = [[[KrollPromise alloc] initInContext:context] retain];
```
vijaysingh-axway, line 234:
```suggestion
    openPromise = [[KrollPromise alloc] initInContext:context];
```
No need of retain.  For safer side openPromise should be release inside -(void)dealloc as well.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -303,22 +323,26 @@ - (void)close:(id)args
     } else {
       args = [NSArray arrayWithObject:self];
     }
-    [tab closeWindow:args];
-    return;
+    return [tab closeWindow:args];
+  }
+
+  if (closePromise == nil) {
+    closePromise = [[[KrollPromise alloc] initInContext:context] retain];
```
vijaysingh-axway, line 330:
```suggestion
    closePromise = [[KrollPromise alloc] initInContext:context];
```
Same here. No need of retain. And release it in dealloc as well.

android/modules/ui/src/java/ti/modules/titanium/ui/NavigationWindowProxy.java
sgtcoolguy: I forgot I left this loose thread here. The tests prove it's not working right either. The issue here is we synchronously close the window and then call up to `super`, which sees that `opened == false` and returns a rejected `Promise`.

We really want to "skip" up the hierarchy to call `TiWindowProxy.open()` (not `WindowProxy.open()`). Maybe I can use `handleOpen()` overriding to fix this?

drauggres:
I don't think this is a good idea overall. Even though I like `Promise`s, they don't fit here.
And I think this will lead to problems with "unhandled promises" in v8 too.

sgtcoolguy:
@drauggres Why specifically do you disagree here?

In doing this implementation, it struck me by how often we actually have hidden cases of failures or "ignoring" the open/close calls we do where we simply spit something to the logs (or not even that!) and there's no means of determining concretely whether or not the open or close call worked. In the past developers are expected to hang event listeners to get notified if they do get opened/closed - but there was really no mechanism for handling failures specifically.

I do think this will force us to deal with the overall problem of generic unhandled promise rejections - which we'll run into regardless of this PR, but obviously with lots of pre-existing code out there this would quickly trigger for existing apps. But presuming we can hook a global event and default listener to log unhandled promises, this should actually help developers pinpoint where they may have open/close calls occurring in bad states or with timing issues.

hansemannn:
This may have caused [TIMOB-28394](https://jira.appcelerator.org/browse/TIMOB-28394) (still investigating)

PR #12418, opened 2021-01-27
Modernize some ListView examples

Topener:
The examples were still using "setItems" while this setter has been deprecated. A more modern way of doing this is setting the items directly in the creation of the section as opposed to adding them after.


ewanharris (approved):
LGTM!

PR #12419, opened 2021-01-27
fix(ios): calculate proper row width for UITableViewStyleInsetGrouped

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28325

For table view style UITableViewStyleInsetGrouped, it has left and right margin.


sgtcoolguy (commented):
code changes look fine. Can we add a test that reproduces/verifies the fix? I assume we can set up a view and do a snapshot image test to verify the text is visible?

vijaysingh-axway:
@sgtcoolguy  I added the image matching test case. But if I take @2x image from 'iPad Pro 12.9 inch' and run it against iPhone SE (2nd generation), which is 2x, it has approx 14000 and ~18000 pixel difference in tableview and listview, which I think is a lot. So I allowed that much pixel difference in maxPixelMismatch.

sgtcoolguy (approved):

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 10.0.0.v20210202155554
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Xcode: 12.3, 12.2
Device: iPhone 7Plus(v14.0)
Simulator: iPhone 11(v14.3), iPad Pro(v14.3), iPhone 5S(v12.2)

PR #12435, opened 2021-02-02
fix(android): rgba(int,int,int) parity

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6672

iOS will use rgb if the alpha value of rgba is missing. Android will ignore the color
```
var win = Ti.UI.createWindow({ backgroundColor: 'white' });
win.open();
var view1 = Ti.UI.createView({backgroundColor:"rgb(255,0,0)", width: 100,height: 100, top:0});
win.add(view1);
var view2 = Ti.UI.createView({backgroundColor:"rgba(255,0,0)", width: 100,height: 100});
win.add(view2);
```
should display 2 red squares.


sgtcoolguy (approved):

sgtcoolguy:
I do think this PR is ripe for a UI snapshot test to verify the fix here...

m1ga:
how do I create the snapshot images?
Something like this in `/tests/Resources/ti.ui.view.test.js` should work:

```it('rgba fallback', finish => {
		win = Ti.UI.createWindow({ backgroundColor:"#fff" });
		const rgbaView = Ti.UI.createView({
			width: 100,
			height: 100,
			backgroundColor: 'rgba(255,0,0)',
			left: 0
		});
		const rgbView = Ti.UI.createView({
			width: 100,
			height: 100,
			backgroundColor: 'rgb(0,255,0)',
			left: 100
		});

		win.addEventListener('postlayout', function postlayout() {
			win.removeEventListener('postlayout', postlayout);
			try {
				should(rgbaView).matchImage('snapshots/rgbaView_red.png');
				should(rgbView).matchImage('snapshots/rgbView_green.png');
			} catch (err) {
				return finish(err);
			}
			finish();
		});
		win.add(rgbaView);
		win.add(rgbView);
		win.open();
	});
```

sgtcoolguy:
Yeah, that looks good. When you run the test suite the first time it'll fail because it has nothing to match, but should write the generated images to disk under `tests/generated`, and if running locally I think I have it auto-copy them into the test suite itself (under `test/Resources/<platform>/snapshots`). Our CI builds should also pop them out as build artifacts that we can then check into the suite. For android we typically have emulators/devices with  2x/2.75x/3x density images, and iOS we typically check in the 2x/3x images. I can do that stuff for you since I know it can be a pain.

m1ga:
ah I see. The `generated` folder is still empty in my test run but it threw an error. I've pushed the test so perhaps the CI will generate them

sgtcoolguy:
Yeah, it makes some assumptions about how to grab the image it writes into the app. It tries to use `adb` to grab them off android device/sim, but if `adb` is not on the `PATH` it likely would fail under the hood. Or sometimes it has issues with using say `adb pull` vs `adb shell` w/ `cat`. 
https://github.com/appcelerator/titanium_mobile/blob/master/build/lib/test/test.js#L664-L709

PR #12436, opened 2021-02-02
After ignoring warning for 8 years, move log to debug

Topener:
According to git blame, this line has been throwing warnings for 8 years (at least) while having no direct impact on actual functionality. 

Let's lower the log level to debug instead so it doesn't gobble up console space


vijaysingh-axway (approved):
CR passed.

hansemannn:
It does have an impact though. It affects the frame rate if you compare it closer.

hansemannn (approved):
Okay to take it.

PR #12437, opened 2021-02-03
chore(ios): drop iOS 10 support

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28011


sgtcoolguy (approved):
LGTM

sgtcoolguy:
I *think* we also need to check our build/cli code for references to armv7 architectures now? It looks like we're doing a packaging sanity check to find `ios-arm64_armv7` headers for TitaniumKit and they don't exist - because I think with iOS 11+ there is no more armv7 (32-bit) arch necessary/built. If that is the case, then this has up/downstream changes necessary (i.e. we already build armv7 xcframeworks for tiverify/analytics/httpclient - we'll likely want to rebuild them without eventually; we specify armv7 arch in the module template; our iOS cli code for hyperloop/frameworks looks for armv7 arch)

vijaysingh-axway:
> I _think_ we also need to check our build/cli code for references to armv7 architectures now? It looks like we're doing a packaging sanity check to find `ios-arm64_armv7` headers for TitaniumKit and they don't exist - because I think with iOS 11+ there is no more armv7 (32-bit) arch necessary/built. If that is the case, then this has up/downstream changes necessary (i.e. we already build armv7 xcframeworks for tiverify/analytics/httpclient - we'll likely want to rebuild them without eventually; we specify armv7 arch in the module template; our iOS cli code for hyperloop/frameworks looks for armv7 arch)

Yes. With iOS 11+,  armv7 and i386 are no more. I'll update PR.

vijaysingh-axway:
tiverify PR (to update minimum iOS 11) - https://github.com/appcelerator/titanium_verify/pull/34
APSHTTPClient PR (to update minimum iOS 11) - https://github.com/appcelerator/APSHTTPClient/pull/56
APSAnalytics PR (to update minimum iOS 11) - https://github.com/appcelerator/aps_sdk/pull/350

vijaysingh-axway:
@sgtcoolguy Updated PR to fix build issue, module template and updated xcframeworks.
For modules update I have created new ticket https://jira.appcelerator.org/browse/TIMOB-28334

PR #12441, opened 2021-02-04
fix(tests): move to getter/setter

m1ga:

**Optional Description:**

Using getter/setter instead of methods to avoid deprecation warnings.


m1ga:
looks like iOS is missing some getter/setters

sgtcoolguy:
Relates to #12310

m1ga:
@hansemannn yes, should still be fine here to merge.

won't affect builds at the moment since the tests won't run automatically.

PR #12445, opened 2021-02-05
macOS Ti.Blob/Image scale incorrect

sgtcoolguy:
**Optional Description:**
This is a work in progress to try and debug why sometimes macOS image blobs report bad/unexpected width/height values for rendered View images.

I suspect this relates to the known issues around macOS Catalyst apps `nativeScale` being incorrect sometimes on Catalina (54009624): https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes

I *cannot* find a reliable way to detect when this occurs to hack a fix for it. The various scale values all report the correct value in the same code where the view reports a 2x density equivalent `CGRect` size (i.e. 5.5 for an 11px width). I also strongly suspect it may have something to do with then the OS has a modal dialog and/or notifications open. Once I close them all, the issue didn't seem to re-occur.

So, to sum up:
  - This affects macOS Catalyst apps only
  - It's "intermittent" in our testing
  - It may be due to a known issue with Catalina, and therefore may just "go away" on Big Sur


sgtcoolguy:
Didn't find any reliable way to detect and "fix" this issue. Sometimes macOS catalyst apps (maybe only on Catalina) simply return a view's bounds/size in its that assume 2x density while the screen/device itself reports a 1x density. So ultimately when we render the view to image the image is half the size it should be.

PR #12457, opened 2021-02-09
chore(ios): drop iOS 11 support

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28342


hansemannn:
From 10 to 12 already? Please make this in the next major if thats possible. Other frameworks are at 10/11 right now.

iphone/Classes/TiAppiOSProxy.m
```diff
@@ -451,25 +451,23 @@ - (void)registerUserNotificationSettings:(id)args
...
-          break;
-        }
-        case UNAuthorizationOptionProvidesAppNotificationSettings: // USER_NOTIFICATION_TYPE_PROVIDES_APP_NOTIFICATION_SETTINGS
-        {
-          types |= UNAuthorizationOptionProvidesAppNotificationSettings;
-          break;
-        }
-        }
+
+      switch ([TiUtils intValue:thisTypeRequested]) {
```
sgtcoolguy, line 455: merge into the switch statement above - no point having two switch statements for the same value now

sgtcoolguy (changes requested):

iphone/Classes/TiAppiOSProxy.m
vijaysingh-axway: Done. Thanks!

sgtcoolguy (approved):

PR #12458, opened 2021-02-09
fix(ios): use supportedInterfaceOrientations of UINavigationController.topViewController instead of UINavigationController

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28282

Relates to PR #12335


ssekhri:
FR Passed

PR #11432, opened 2020-01-14
impr(android): allow gradle to automatically download missing dependencies

garymathews:
- Allow Gradle to automatically download and install missing dependencies (Android SDK, Build-tools etc..)
- These changes will offload SDK validation to Gradle

##### TEST CASE \#1
- Remove `Android SDK Platform 29`
- Attempt to build Titanium application
- Gradle should download and install `Android SDK Platform 29` and build successfully

##### TEST CASE \#2
- Remove `Android SDK Build-Tools`
- Attempt to build Titanium application
- Gradle should download and install `Android SDK Build-Tools` and build successfully

##### TEST CASE \#3
- Remove `Android SDK Platform-Tools`
- Attempt to build Titanium application
- Gradle should download and install `Android SDK Platform-Tools` and build successfully

##### TEST CASE \#4
- Remove `Android SDK Tools`
- Attempt to build Titanium application
- Gradle should download and install `Android SDK Tools` and build successfully

#### Repeat test cases for Titanium native module

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-27686)

PR #12461, opened 2021-02-10
fix(android): move buildOnly assignment into pre-validate hook

ewanharris:
The parsing of aliases has not been performed by the time config is called so by waiting until
pre-validate we know that the build-only option has been parsed and handled. Changing the previous
check to also check for b will not work because the cli has not determined it as an alias of
--build-only so the value is still undefined

Closes TIMOB-28345

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28345

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -77,8 +77,6 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 
 	const _t = this;
 
-	this.buildOnly = cli.argv['build-only'] !== undefined;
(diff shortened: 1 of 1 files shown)


sgtcoolguy (approved):
seems fine to me?

garymathews (approved):
CR: **PASS**

I also re-tested the test cases in https://github.com/appcelerator/titanium_mobile/pull/11432

cb1kenobi (approved):
LGTM

PR #12481, opened 2021-02-18
fix(android): use C++14 when building native modules

garymathews:
- Use `c++14` when building native modules
- Update V8 to `8.8.278.17` that includes compatibility patch to revert https://github.com/v8/v8/commit/50ddb12d2c3ca29c0f906efe78c4299fdf6ef490

###### NOTE: We should check modules built with TiSDK 10.0 work on TiSDK 9.3.X


jquick-axway:
We still have a V8 backward compatibility issue in 10.0.0 with previously built modules. We currently fail the `invokeCallbackSync` and `invokeCallbackAsync` unit tests in the following PR. It's an issue with callbacks. The 1st argument is getting through to the callback fine, but the rest of the callback arguments are not.
_(If I can find a solution, then I'll open my open PR and close this one.)_
https://github.com/appcelerator/titanium_mobile/pull/12311/files#diff-353594100a924d948f537a44e9b97d19934cf70e5c2435303732abff7d2df6d7R222-R248

Also, 10.0.0 built modules will have their min API Level set to 21 (matching Titanium 10.0.0's min API Level), which is a problem if you want to use that module in 9.x.x built apps. It will cause an app build failure. This has nothing to do with our V8 change, but we should probably work-around it anyways. Simple solution is to hard-code the module's "build.gradle" `minSdkVersion` to `19`.

jquick-axway:
I've isolated the issue. As of V8 version 8.8, Google has made a breaking-change where method arguments are now stored in reverse order.

If you look in the "v8.h" file, look for the `Local<Value> FunctionCallbackInfo<T>::operator[](int i)` function. Note that this is an "inlined" function.

In version 8.8, it looks like this...
```cpp
template<typename T>
Local<Value> FunctionCallbackInfo<T>::operator[](int i) const {
  // values_ points to the first argument (not the receiver).
  if (i < 0 || length_ <= i) return Local<Value>(*Undefined(GetIsolate()));
  return Local<Value>(reinterpret_cast<Value*>(values_ + i));
}
```

In version 8.7, it looks like the below. And we're not defining the `V8_REVERSE_JSARGS` definition so we're ending up in the `#else` block.
```cpp
template<typename T>
Local<Value> FunctionCallbackInfo<T>::operator[](int i) const {
  // values_ points to the first argument (not the receiver).
  if (i < 0 || length_ <= i) return Local<Value>(*Undefined(GetIsolate()));
#ifdef V8_REVERSE_JSARGS
  return Local<Value>(reinterpret_cast<Value*>(values_ + i));
#else
  return Local<Value>(reinterpret_cast<Value*>(values_ - i));
#endif
}
```

Unfortunately, this method call is inlined. This explains why the method's 2nd argument returns the "this" object for older modules running in Titanium 10.0.0. And I would expect 10.0.0 built modules to have the reverse problem when running in a Titanium 9.x.x app.

I can't think of any work-around to maintain backward compatibility other than to revert V8 back to version 8.7.

garymathews:
I made a patch to disable `v8_enable_reverse_jsargs` in 8.7, but they removed the parameter altogether in 8.8 😞 
https://github.com/appcelerator/v8_titanium/commit/2971598f39074f4d1f2bc10edf32119b30fbcb3c

I'll try to revert their commit https://github.com/v8/v8/commit/50ddb12d2c3ca29c0f906efe78c4299fdf6ef490

cb1kenobi:
I literally just read about this the other day: https://v8.dev/blog/v8-release-89. They reverse the arguments to optimize for when the number of arguments mismatches the function signature.

jquick-axway (approved):
CR/FR: Pass

I ran the following tests:
- Exercised all JS types with 9.2.0 built module from PR #12311 with 10.0.0 built app.
- Built "ti.imagefactory" module with 10.0.0 and ran its example app.
- Used 10.0.0 built "ti.imagefactory" module in 10.0.0 built app on x86, x86_64, ARM32, ARM64.
- Used 10.0.0 built "ti.imagefactory" module in 9.0.0 built app on x86, x86_64, ARM32, ARM64.
- Built kitchensink-v2 with 10.0.0 and ran on x86_64 and ARM64. (Verified maps worked.)

PR #12482, opened 2021-02-18
chore(ios): update modules in SDK

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28334


iphone/cli/hooks/hyperloop.js
```diff
@@ -17,7 +17,7 @@ exports.init = function (logger, config, cli) {
 	let cmds,
 		libfile;
 
-	const archs = [ 'armv7', 'arm64', 'x86_64', 'i386' ];
+	const archs = [ 'arm64', 'x86_64' ];
```
sgtcoolguy, line 20: Do we want to separate out "required" architectures vs "optional"?  Like, will this break usage of older iOS native modules that have 32-bit libraries embedded?

vijaysingh-axway: I don't think it will break with older 32-bit libraries. From SDK 10 only arm64 and x86_64 arch will be required.

PR #12496, opened 2021-02-24
fix(ios): documentViewer setAnnotation method causing a crash

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28360


sgtcoolguy (commented):
I think a test case (or several, for various annotation types) would be helpful here.

I had thought "old-style" proxies always got `NSArray` args, but I guess not?

tests/Resources/ti.ui.ios.documentviewer.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+	afterEach(done => {
+		documentViewer = null;
+		done();
+	});
+
+	it('.name', () => {
+		// On macOS < 11, it is failing. https://developer.apple.com/forums/thread/125819
+		if (isCI && utilities.isMacOS() && OS_VERSION_MAJOR < 11) {
+			this.skip();
```
sgtcoolguy, line 40: There is no `this` context in an arrow function. You'll need to change the `it('.name', () => {` line above to:
``` js
it('.name', function () {
```

tests/Resources/ti.ui.ios.documentviewer.test.js
```diff
@@ -0,0 +1,70 @@
...
+			return;
+		}
+		documentViewer.annotation = 'annotation';
+		should(documentViewer.name).eql('example.html');
+	});
+
+	it('.annotation', () => {
+		// On macOS < 11, it is failing. https://developer.apple.com/forums/thread/125819
+		if (isCI && utilities.isMacOS() && OS_VERSION_MAJOR < 11) {
+			this.skip();
```
sgtcoolguy, line 50: There is no `this` context in an arrow function. You'll need to change the `it('.annotation', () => {` line above to:
``` js
it('. annotation', function () {
```

tests/Resources/ti.ui.ios.documentviewer.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+	let documentViewer;
+
+	beforeEach(() => {
+		documentViewer = Ti.UI.iOS.createDocumentViewer({
+			url: 'example.html'
+		});
+	});
+
+	afterEach(done => {
```
sgtcoolguy, line 32: No need to pass in a callback function to this hook or call it, it runs sync.
```suggestion
	afterEach(() => {
```

In fact, I don't think you even need this hook since you're not doing any work on the proxy like calling a `release` method - so assigning null shouldn't really help/matter.

tests/Resources/ti.ui.ios.documentviewer.test.js
vijaysingh-axway: updated

sgtcoolguy (approved):

ssekhri:
FR Passed.
Verified on:
Mac OS: 10.15.4
SDK: 10.1.0.v20210315123356
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Xcode: 12.3
iOS simulator v14.3

PR #12499, opened 2021-02-25
feat(android): sync tab and actionbar title

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28371

Will update the Actionbar title automatically when you change the tab.

**example**
```js
Ti.UI.createTabGroup({
	autoTabTitle: true,
    // style: Titanium.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION,
	tabs: [

		Ti.UI.createTab({
			window: Ti.UI.createWindow({
				title: 'Window 1'
			}),
			title: 'Page 1'
		}),

		Ti.UI.createTab({
			window: Ti.UI.createWindow({
				title: 'Window 2'
			}),
			title: 'Page 2'
		})

	]

}).open();
```


hansemannn:
Thanks @m1ga! But I think this is not the correct behavior. I would expect to see the window title at the top, not the tab title. For parity with iOS, it also would not need an own property but become the default behavior in SDK 10

m1ga:
@hansemannn ah I see. Will update the PR.

Keep in mind it is not the default behavior in other apps. e.g. in Whatsapp the actionbar title is always the app name (or they have all window tab names set to "Whatsapp") and it doesn't change:
Screenshot_20210225-202616
same for the Play Store or Youtube Music. The Actionbar title is not changing when changing the tabs

hansemannn:
Mhh, thats odd. Even with bottom navigation? I thought Instagram and some other apps moved to that behavior, maybe I missed something here.

m1ga:
ah, Instagram is changing the actionbar! But the right hamburger menu items moves between the shop and profile page :-) looks like they set a whole new actionbar

m1ga:
it is using the window title now but the first title is not set correctly ("Windo..." instead of "Window 1")

```javascript
Ti.UI.createTabGroup({
	autoTabTitle: true,
    // style: Titanium.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION,
	tabs: [

		Ti.UI.createTab({
			window: Ti.UI.createWindow({
				title: 'Window 1'
			}),
			title: 'Page 1'
		}),

		Ti.UI.createTab({
			window: Ti.UI.createWindow({
				title: 'Window 2'
			}),
			title: 'Page 2'
		})

	]

}).open();
```

but will look at that tomorrow

m1ga:
fixed the truncated title at app start. Apparently you have to expand the title as a workaround (will set it to twice the title and set it to blank).

android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
```diff
@@ -381,6 +396,15 @@ public void onWindowActivityCreated()
 		// Finish open handling by loading proxy settings.
 		handlePostOpen();
 
+		TiUIAbstractTabGroup tg = (TiUIAbstractTabGroup) view;
+		if (autoTabTitle) {
```
hansemannn, line 400: Rename to `tabGroup` instead of `tg`. Also, it can be inside the if-statement.

apidoc/Titanium/UI/TabGroup.yml
```diff
@@ -324,6 +324,13 @@ properties:
     default: true
     platforms: [iphone, ipad, macos]
 
+  - name: autoTabTitle
+    summary: If set to `true` it will automatically set the actionbar title to the current tabs window title.
+    type: Boolean
+    default: false
+    since: 10.1.0
```
hansemannn, line 331: Change to 10.2.0

hansemannn (changes requested):
A few minor issues

hansemannn (approved):

PR #12511, opened 2021-03-02
refactor(ios): change Ti.UI.TabbedBar to use UITabBar

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28373

**Summary:**
- This changes `Ti.UI.TabbedBar` to use a [tab bar](https://developer.apple.com/design/human-interface-guidelines/ios/bars/tab-bars) instead of a [segmented control](https://developer.apple.com/design/human-interface-guidelines/ios/controls/segmented-controls/).
- Titanium 10.0.0's new `Ti.UI.OptionBar` is now used to display a segmented control on iOS. See: [TIMOB-28317](https://jira.appcelerator.org/browse/TIMOB-28317)
- Intention is that a widget named `TabbedBar` should show tabs, like it does on Android, and allow app devs to layout their own tab group like user interface.

TabbedBar-iOS

**Test:**
1. Build and run `TabbedBarTest.js` attached to [TIMOB-28373](https://jira.appcelerator.org/browse/TIMOB-28373) on iOS.
2. Verify that the app shows tabs as shown in the screenshot above.
3. Tap on the 2nd tab of every bar and verify the following gets logged.
`@@@ button click index: 1`

### DIFF
--- a/apidoc/Titanium/UI/ButtonBar.yml
+++ b/apidoc/Titanium/UI/ButtonBar.yml
@@ -1,16 +1,12 @@
 ---
 name: Titanium.UI.ButtonBar
-summary: An iOS button bar component.
+summary: A horizontal bar of joined buttons.
 description: |
     The button bar is a set of buttons joined into a single control.
     On iOS, you can set up the buttons with either a title or image, but not both.
     On Android, you can set up the buttons with a title, image, or both.
 
     Use the <Titanium.UI.createButtonBar> method or **`<ButtonBar>`** Alloy element to create a button bar.
-
-    The [TabbedBar](Titanium.UI.iOS.TabbedBar) control is a button bar where the 
-    last selected button mantains a pressed or selected state. The following discussion
-    applies to both button bar and tabbed bar.
 extends: Titanium.UI.View
 since: {android: "10.0.0", iphone: "0.8", ipad: "0.8", macos: "9.2.0"}
 platforms: [android, iphone, ipad, macos]
--- a/apidoc/Titanium/UI/TabbedBar.yml
+++ b/apidoc/Titanium/UI/TabbedBar.yml
@@ -1,14 +1,17 @@
 ---
 name: Titanium.UI.TabbedBar
 
-summary: A button bar that maintains a selected state.
+summary: A stand-alone bar of tabs.
 description: |
-    A tabbed bar is a button bar that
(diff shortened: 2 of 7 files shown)


hansemannn:
Please reconsider if this is really a good idea. The UITabBar is a top-level element on iOS and should not be added to a view directly. It's probably even causing App Store rejections and can be easily overseen by developers that have larger apps and upgrade to SDK 10 (which, in terms of the recent news) will not really help developers.

vijaysingh-axway:
1. As per Apple doc [here](https://developer.apple.com/documentation/uikit/uitabbar), UITabBar is typically used in conjunction with UITabBarController. In titanium Ti.UI.TabGroup is used to implement UITabBarController.
2. I don't think adding UITabBar as standalone component will cause Appstore rejection.
3. Do we want to make breaking change for this, which people are using from long and I think no one is asking for  this?

hansemannn:
@vijaysingh-axway See [this link](https://developer.apple.com/design/human-interface-guidelines/ios/bars/tab-bars/) for reference, especially:
> **Use a tab bar strictly for navigation.** Don’t use tab bar buttons to enable actions. If you need to provide controls that act on elements in the current view, use a toolbar instead.

As well as the fact that button-only tabbed bars (which is the only default right now) will cause major display issues, because UITab's are meant to have an icon and without it, the title will be displayed at the bottom (see screenshot above).

jquick-axway:
No one here is suggesting that `UITabBar` should be used as a toolbar or as a segmented control replacement. The main point here is that `Ti.UI.TabbedBar` is misnamed for the native component currently used on iOS. The **solution** is to create a new `Ti.UI.OptionBar` to show a segmented control (already merged into Titanium 10.0.0) and change `Ti.UI.TabbedBar` to use `UITabBar` so app developers can implement their own custom tab group like interface.

If you feel this is controversial, then I can close this PR for 10.0.0. This means iOS will use a native `UISegmentedControl` for both `Ti.UI.TabbedBar` and `Ti.UI.OptionsBar`.

On Android, `Ti.UI.TabbedBar` uses the native `TabLayout` and `BottomNavigationView`. The new `Ti.UI.OptionBar` uses the native `MaterialToggleButtonGroup`. We're not changing the Android side. The native widgets being used on Android make sense.

PR #12515, opened 2021-03-03
fix(android): expose builder property for hyperloop to mark js files not to process

sgtcoolguy:
**Optional Description:**
This adds back the builder property `htmlJsFiles` for android hyperloop and then shuttles the file listing into the CLI gather code to use as a file listing to explicitly not process.


ewanharris:
Backported to 10_0_X in https://github.com/appcelerator/titanium_mobile/commit/a6ad3ed4383d159302e45b667b21083a724e0e97

PR #12516, opened 2021-03-03
Support Android drawable resource images and splash screens properly

sgtcoolguy:
**Description:**
This basically is an attempt to re-introduce the code from https://github.com/appcelerator/titanium_mobile/blob/9_3_X/android/cli/commands/_build.js#L2525-L2593 into the new walk/categorize/process workflow of the build.

We walk/gather as before, but now we categorize android splash screens into the `launchImages` map and drawable resources into `imageAssets`. We then "process" those groups specially - basically just finessing the output filenames before doing the same copying as normal `CopyResourcesTask`.

@janvennemann I'm modifying the output file paths in the task constructors, but I'm assuming we could defer to some other overridden method so it ran as part of the task run lifecycle properly (but early enough that the destination file paths are correct when it checks state).


jquick-axway:
When you do an encrypted build (ie: a device/production build), both unencrypted and encrypted JS file are being copied into the app. That is, both an unencrypted "app.js" and an encrypted "app.js.bin" can be found in the APK. It's because of the following code block which copies over the processed JS files, but these files are not encrypted. The encryption task will copy the processed files over for us.
https://github.com/appcelerator/titanium_mobile/blob/c3a68e08b9d340a4759d75b6c40996396d738fd2/android/cli/commands/_build.js#L2610-L2626

We need to add an if-check to only copy over the JS files if not encrypted. I've tested this and it solves the problem.
```javascript
// Copy all unencrypted files processed by ProcessJsTask to "app" project's APK "assets" directory.
// Note: Encrypted files are copied to app project by "ti.cloak" instead.
if (!this.encryptJS) {
	const resourcesToCopy = new Map();
	for (let [ key, value ] of jsFilesMap) {
		resourcesToCopy.set(key, {
			src: path.join(this.buildAssetsDir, key),
			dest: value.dest
		});
		this.unmarkBuildDirFile(value.dest);
	}
	const copyTask = new CopyResourcesTask({
		incrementalDirectory: path.join(this.buildTiIncrementalDir, 'copy-processed-js'),
		name: 'copy-processed-js',
		logger: this.logger,
		builder: this,
		files: resourcesToCopy
	});
	await copyTask.run();
}
```

cli/lib/gather.js
```diff
@@ -259,30 +263,48 @@ class Categorizer {
...
-				// if we are using app thinning, then don't copy the image, instead mark the
-				// image to be injected into the asset catalog. Also, exclude images that are
-				// managed by their bundles.
-				} else if (this.useAppThinning && !relPath.match(BUNDLE_FILE_REGEXP)) {
-					results.imageAssets.set(relPath, info);
-				} else {
-					results.resourcesToCopy.set(relPath, info);
+				// Toss Android splash screens into launchImages
+				if (this.platform === 'android') {
+					if (info.name === 'default.9' || info.name === 'default') {
```
jquick-axway, line 277: 9-patch images have to be PNGs. So, it should look like this...
```javascript
if ((info.name === 'default') || (info.name === 'default.9' && info.ext === 'png')) {
    // ...
}
```

android/cli/lib/process-drawables-task.js
```diff
@@ -0,0 +1,64 @@
...
+			const foldername = drawableDpiRegExp.test(origFoldername) ? 'drawable-' + origFoldername[0] + 'dpi' : 'drawable-' + origFoldername.substring(4);
+
+			const destFilename = `${info.name}.${info.ext}`;
+
+			// We have a drawable image file. (Rename it if it contains invalid characters.)
+			const warningMessages = [];
+			if (parts.length > 3) {
+				warningMessages.push(__('- Files cannot be put into subdirectories.'));
+			}
+			let destFilteredFilename = `${info.name.toLowerCase().replace(/[^a-z0-9_]/g, '_')}.${info.ext}`;
```
jquick-axway, line 43: This breaks 9-patch image handling by turning "my.9.png" to "my_9.png". We need to keep the ".9.png" part. This is the only case where a 2nd period '.' in the file name is okay.

So perhaps do the following instead?
```javascript
let destFilteredFilename = info.name.toLowerCase();
let destFilteredExtension = info.ext;
if (destFilteredFilename.endsWith('.9') && (info.ext === 'png')) {
	destFilteredFilename = destFilteredFilename.substring(0, destFilteredFilename.length - 2);
	destFilteredExtension = '9.png';
}
destFilteredFilename = `${destFilteredFilename.replace(/[^a-z0-9_]/g, '_')}.${destFilteredExtension}`;
```

jquick-axway (changes requested):

android/cli/lib/process-drawables-task.js
garymathews: Playing with the regex I was able to get this, which should acheive the same result:
```JS
let destFilteredFilename = `${info.name.toLowerCase().replace(/(?!\.9$)[^a-z0-9_]/g, '_')}.${info.ext}`;
```

garymathews (approved):
CR: **PASS**
##### Suggestions
  - Josh's suggestion for validating 9-patch extensions could be added
  - _Missing semicolon_ lint error should be addressed 🔴
  - `destFilteredFilename` can be simplified
https://github.com/appcelerator/titanium_mobile/blob/9f7bbb469d95e2e40d253e6bbd07a121f11f3a01/android/cli/lib/process-drawables-task.js#L48-L52
into
```JS
let destFilteredFilename = `${info.name.toLowerCase().replace(/(?!\.9$)[^a-z0-9_]/g, '_')}.${info.ext}`;
```
[Here's some tests for the RegEx](https://regex101.com/r/tQEB02/1)

FT: **PASS** - Tested **kitchensink-v2** on Windows

cli/lib/gather.js
jquick-axway: There's another issue here.

We only accept "default.png"/"default.9.png" as a splash screen image if it's in the below directories. That is, it should look for the default PNG in the root "src"/"origSrc" given to the `walk()` method.
- `./Resources`
- `./Resources/android`
- `./Resources/images/res-*`
- `./Resources/android/images/res-*`

The issue we have now is that it's wrongly grabbing a "default.png" from any folder which is causing a `EEXISTS: file already exists` build error since it collides with the "default.png" that I have in the correct folder.

jquick-axway: Also the old handling of "default.png" file collision provided a more meaningful error message. If i add the following files to my project:
- `./Resources/default.png`
- `./Resources/android/default.png`

The I would get the following build error.
```
[ERROR] Found conflicting resources:
[ERROR]    Resources/default.9.png
[ERROR]    Resources/android/default.9.png
[ERROR] You cannot have resources that resolve to the same resource entry name
```

Although I would be okay with us ditching the error and letting the last one win. I think files under the "android" subdirectory should take priority over files in the root directory. Similar to how Android handles "res" file name collision where the **app's** "res" files are used over the **library's** res files of the same name, allowing the app to "override" the library's resources (this is intentional).

android/cli/lib/process-drawables-task.js
```diff
@@ -0,0 +1,74 @@
...
+			if (parts.length > 3) {
+				warningMessages.push(__('- Files cannot be put into subdirectories.'));
+			}
+			// info.name may have .9 suffix, if so, we do not want to convert that .9 to _9
+			let base = info.name;
+			let suffix = '';
+			if (info.name.endsWith('.9')) {
+				base = info.name.substring(0, info.name.length - 2);
+				suffix = '.9';
+			}
```
jquick-axway, line 51: We have a major regression here.

We're not correctly renaming "res" files put under "subdirectories" like the below.
`./Resources/android/images/res-mdpi/logos/app.png`

The above file should be renamed to the below. Effectively, path separators should turn into underscores.
`logos_app.png`

But currently the code is stripping away the subdirectory name like the below. This prevents the Titanium app from finding the image. _(Note that many noteworthy Titanium apps put res files under subdirectories, even though we log warnings about it.)_
`app.png`

We should probably do something like the below.
```javascript
let base = info.name;
let suffix = '';
if (info.name.endsWith('.9')) {
	base = info.name.substring(0, info.name.length - 2);
	suffix = '.9';
}
if (parts.length > 3) {
	warningMessages.push(__('- Files cannot be put into subdirectories.'));
	base = parts.slice(3, parts.length - 1).join('_') + '_' + base;
}
let destFilteredFilename = `${base.toLowerCase().replace(/[^a-z0-9_]/g, '_')}${suffix}.${info.ext}`;
```

jquick-axway (changes requested):

sgtcoolguy:
OK, I'm incorporating Gary's feedback, and tweaking the two to match the launch images by matching the relative path (from the dir passed to the walker, i.e. `<project>/Resources`, `<project>/Resources/android`). The regex basically guards that 9-patch can only be PNG and that it can only live under the specific directories Josh mentioned, as shown in this regex tester page: https://regex101.com/r/Z8jW3D/1

jquick-axway (approved):
CR/FR: Pass

PR #12518, opened 2021-03-03
chore(android): build snapshots in CI

garymathews:
- Remove need for snapshot build server
  - Include snapshot generation in Titanium Android SDK build
- Split CI to seperate platform build jobs and packaging


.github/workflows/build.yml
```diff
@@ -76,6 +114,86 @@ jobs:
...
+      name: Install dependencies
+      if: steps.node-cache.outputs.cache-hit != 'true'
+
+    - run: npm run lint:ios
+      name: Lint
+
+    - run: npm run build:ios
+      name: Build
+
+    - name: Cache Native Modules
```
sgtcoolguy, line 142: This should no longer be necessary if we're only building the iOS source in this job and should be moved to the `package` job to run before the `npm run package` step

sgtcoolguy (approved):
This is awesome!

Some notes:
- There's no concept of transient artifacts in Github's build system, so you have to upload/download artifacts (whereas Jenkins has `stash`/`unstash` to mark files we want to transfer between nodes but not retain). So a hack around this is to add a final step to delete your transient artifacts you didn't want to retain, see https://github.com/appcelerator/v8_titanium/blob/8.8-lkgr/.github/workflows/build.yml#L155-L161
- We should still use the snapshot server on Windows/Mac platforms while we still have it, otherwise we'll all be stuck with no snapshots in local builds while developing. It obviously will need to be ripped out once w're truly EOL, but it's a nice convenience for us for now.
- The magical set of files to stash and move between build and packaging feels like we should bake into some script or something to me, but I guess it's fine for now?

build/lib/builder.js
```diff
@@ -17,7 +17,7 @@ const DIST_DIR = path.join(ROOT_DIR, 'dist');
 const TMP_DIR = path.join(DIST_DIR, 'tmp');
 
 // platforms/OS mappings
-const ALL_OSES = [ 'win32', 'linux', 'osx' ];
+const ALL_OSES = [ 'win32', 'osx' ]; // NOTE: 'linux' could be added, but is not officially supported.
```
garymathews, line 20: Provides slightly faster packaging time as _mobilesdk-linux.zip_ is not generated.

m1ga: is this just for the CI or will that also stop me from building an SDK for linux at the end?

garymathews: No it shouldn't, however I updated this PR so specifying `--all` will only omit linux on non-linux hosts.

cb1kenobi:
Rebuild the snapshot server! https://github.com/tidev/v8-snapshot-www

SDK PR: https://github.com/tidev/titanium-sdk/pull/13907

PR #12519, opened 2021-03-04
fix(android): build handling of drawables and processed JS

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-26352

**Note:**
This is an alternative solution to PR #12516.

**Summary:**
- Resolves regression caused by PR #11447. (Issue was caught before release.)
- Restores `/images/res-*` drawable handling.
- Restores `default.png` and `default.9.png` splash screen handling.
- Fixes issue where encrypted build wrongly added unencrypted JS files within APK.


android/cli/commands/_build.js
```diff
@@ -2379,24 +2379,94 @@ AndroidBuilder.prototype.generateAppProject = async function generateAppProject(
 	await fs.writeFile(path.join(this.buildAppDir, 'build.gradle'), buildGradleContent);
 };
 
+/**
+ * Function to be passed to the Walker object as an "entryCallback" option.
+ * Invoked for each entry found by the walk() method to override res drawable image handling.
+ * @param {object} params Provides to/from entry information.
+ */
+AndroidBuilder.prototype.onWalkCallback = async function onWalkCallback(params) {
```
sgtcoolguy, line 2387: We discussed this on a short call already - but I'm ok with this approach but would caution that the whole reason I had refactored was that both platforms were heavily co-mingling the gathering of the input files with the processing/handling of them which made it very difficult to parallelize and we ended up having lots of minor differences between platforms as to how each type of file was handled even when they should be the same (i.e. copying a file). This moves us back that way again for drawables/splashes on Android.

My goal was to explicitly try and separate 3 steps here: walking/gathering of input files; categorizing them into groups/buckets that get processed equivalently; and then processing each bucket  (in parallel if possible) using `appc-tasks` so that it could handle dealing with saving/comparing state and doing incremental builds. A secondary goal was to try and centralize common cli build code to a common place (i.e. why did we need separate file walking implementations between platforms; or file copying/symlinking logic?)

PR #12524, opened 2021-03-07
fix(android): fragments in children of header/footerview

hansemannn:
All credits to @drauggres!

**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28376

**MapView Test:**
1. Build and run the code in [TIMOB-28376](https://jira.appcelerator.org/browse/TIMOB-28376)'s "description" on Android.
2. Verify it does not crash and shows a map in the ListView.

**ListView Dark/Light Switch Test:**
1. Build and run `ListViewHeaderViewTest.js` attached to [TIMOB-28376](https://jira.appcelerator.org/browse/TIMOB-28376) on Android 10 or higher.
2. Once the app starts, tap on the "Home" button. _(Do not back out.)_
3. Change Android's Dark theme under "Settings/Display".
4. Resume the Titanium app.
5. Verify header/footer text color has changed to the same color as the row text.

**TableView Dark/Light Switch Test:**
1. Build and run `TableViewHeaderLabelTest.js` attached to [TIMOB-28376](https://jira.appcelerator.org/browse/TIMOB-28376) on Android 10 or higher.
2. Once the app starts, tap on the "Home" button. _(Do not back out.)_
3. Change Android's Dark theme under "Settings/Display".
4. Resume the Titanium app.
5. Verify header/footer text color has changed to the same color as the row text.


m1ga:
:+1:  for the community!

should be the same here I guess: https://github.com/appcelerator/titanium_mobile/blob/b6436261f78d53f37c6ddcb461c612345f96e493/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TableViewHolder.java#L538
at least the crash is the same when you exchange listview <-> tableview

hansemannn:
@m1ga Updated!

garymathews (approved):

garymathews (approved):
LGTM!

ssjsamir (approved):
FR Passed, tested using the test cases above.

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.2 Beta
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
iPhone 8 14.2
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListSectionProxy.java
```diff
@@ -357,11 +357,20 @@ public void release()
...
 		}
+
+		// Release header/footer views.
+		if (hasPropertyAndNotNull(TiC.PROPERTY_HEADER_VIEW)) {
+			final TiViewProxy headerProxy = (TiViewProxy) getProperty(TiC.PROPERTY_HEADER_VIEW);
+			headerProxy.releaseViews();
+		}
+		if (hasPropertyAndNotNull(TiC.PROPERTY_FOOTER_VIEW)) {
+			final TiViewProxy footerProxy = (TiViewProxy) getProperty(TiC.PROPERTY_FOOTER_VIEW);
+			footerProxy.releaseViews();
```
hansemannn, line 372: @drauggres Quick question as you fixed this one and [https://jira.appcelerator.org/browse/TIMOB-28583](TIMOB-28583) now popped up for us: When comparing the changes, you release the views *and set the activity* for list headers / footers, but you don't set the activity here. Is this maybe the missing piece to prevent further crashes?

hansemannn: Okay, thank you! So you would recommedn to add the `setActivity` here as well? And I am wondering if a workaround then may be to wait for the window to be opened!

drauggres: > but you don't set the activity for list section headers / footers

Just checked the code, we call `setActivity` in this case too:
https://github.com/appcelerator/titanium_mobile/blob/2dad43e47e6db455ce6d9f852082ab766bfd81d8/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListSectionProxy.java#L350-L365

In theory it is possible to have a "wrong" `Activity` stored in the parent proxy. We don't really know the right one before it is being added to a `Window`.
I think we should override `ListViewProxy.add` to call `setActivity` for its header/footer and every section (and their headers/footers).
https://github.com/appcelerator/titanium_mobile/blob/2dad43e47e6db455ce6d9f852082ab766bfd81d8/android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java#L581

m1ga: >  And I am wondering if a workaround then may be to wait for the window to be opened!

just tried that and it doesn't work. 

Btw: it just happens with a map, a normal view inside the footerView works fine. And if your footerView is visible directly it works too.

hansemannn: The native map view may cause this issue related to wrong activities. We tried a few things and noticed that we indeed created the listview prior to the `open` event. Moving it into the `open` event seems to be stable right now, although it's not always reproducible. @m1ga Does your test case work with that workaround? 

In any case: Guys, you are the best! Happy to provide some rewards for this work!

m1ga:
```js
const mapView = Ti.UI.createView({
	height: 200
});
mapView.add(require('ti.map').createView());

	title: "test",
	backgroundColor: 'gray'
});

function addListView() {
		footerView: mapView,
		templates: {
			test: {
				childTemplates: [{
					type: 'Ti.UI.View',
					childTemplates: [{
						type: 'Ti.UI.Label',
						bindId: 'label',
						properties: {
							color: 'black',
							bindId: 'label'
						}
					}],
					properties: {
						width: Ti.UI.FILL,
						height: 200,
						cardUseCompatPadding: true,
						backgroundColor: 'white',
					}
				}]
			},
		},
		defaultItemTemplate: 'test'
	});
	var items = [];

		items.push({
			label: {
				text: "item " + i
			},
			template: 'test'
		})
	}

	section.setItems(items);
	win.add(listView);
}

	addListView();
})
win.open();
```

I can reproduce it all the time with this example.

drauggres: @m1ga This is exactly the point. If you move `mapView.add(require('ti.map').createView());` inside `addListView` it will work.

m1ga: nope, tried that too. still crashes
```
Invalid ID 0x00000002.
(main) [9,1947] No view found for id 0x2 (unknown) for fragment Fragment{b347057} (a7bb91cc-cd9d-4aeb-865f-8fcad0acc024 id=0x2)
   androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:513)
   androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:261)
   androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1840)
   androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1758)
   androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1701)
   androidx.fragment.app.FragmentManager$4.run(FragmentManager.java:488)
   android.os.Handler.handleCallback(Handler.java:938)
   android.os.Handler.dispatchMessage(Handler.java:99)
   android.os.Looper.loopOnce(Looper.java:201)
   android.os.Looper.loop(Looper.java:288)
   android.app.ActivityThread.main(ActivityThread.java:7839)
   java.lang.reflect.Method.invoke(Native Method)
   com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
   com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```
I thought I had a working version when the map was hidden. But it only work one time (not sure why) and the next time it didn't

m1ga: I can get it to work when I call `mapView.add(require('ti.map').createView());` when the footer view is visible

hansemannn: That's really hard to do in the quite extended environment we're using it here. It really has to be a ti.map thing then. Not sure where to look then.

hansemannn: Interesting - will test that on Monday as the first thing. Thank you!

drauggres: https://github.com/appcelerator/titanium_mobile/issues/13243

PR #12525, opened 2021-03-08
build: no longer generate parity report or JSCA file for Studio

sgtcoolguy:
**Description:**
We currently generate a parity report (API parity across platforms, which was more useful when we were actively developing Windows Mobile), and a `jsca` file from our docs.

The `jsca` file is a specialized JSON file for content assist used by Studio. Given the OL/EOS of Studio, we should stop generating and packaging it. It bloats unzipped SDK install size by ~34Mb!

I don't know if VSCode/Atom also use this file, or if they rely on typescript definitions (which we do generate during builds but don't actually package into the zip - we manually try to push them to DefinitelyTyped as part of GA release process)

### DIFF
--- a/build/lib/docs.js
+++ b/build/lib/docs.js
@@ -35,24 +35,12 @@ class Documentation {
 		});
 	}
 
-	async generateParityReport() {
-		return this.generateReport('parity', 'parity.html');
-	}
-
-	async generateJSCA() {
-		return this.generateReport('jsca', 'api.jsca');
-	}
-
 	async generateTypeScriptTypeDefinitions() {
 		return this.generateReport('typescript', 'index.d.ts');
 	}
 
 	async generate() {
-		return Promise.all([
-			this.generateParityReport(),
-			this.generateJSCA(),
-			this.generateTypeScriptTypeDefinitions()
-		]);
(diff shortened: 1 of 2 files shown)


ewanharris:
@sgtcoolguy we use the JSCA file in both extensions as a starting point for the autocomplete. Maybe we could drop some of the properties from the JSCA rather than remove the thing altogether? I can draw up a list of the bare minimum set needed for the editors and we could maybe trim back to that?

ewanharris:
@sgtcoolguy, so here's the properties that the completions in the editors uses. I've tested reducing it to just that set and it reduces the overall size to about 9.5MB

- types
	- deprecated
	- description
	- functions
	- events
	- properties
		- name
		- description
		- type
		- permission
		- constants

That will have an impact on Studio most likely (I believe it uses things like userAgent/since), I think in the short term there's maybe a couple things we could do:

* Don't pretty-print the file (saves like 50% of the size)
* Remove the autogenerated getter/setters (another 50% saving)

Those two changes together remove it from 34MB down to 8MB (checking with ls -la), I need to verify that Studio is happy with those changes and then I'll make a PR to titanium-docgen

PR #12534, opened 2021-03-09
chore(deps-dev): update titanium-docgen to 4.10.3

sgtcoolguy:
Relates to #12525 

Update titanium-docgen to version that produces JSCA file without accessors or pretty printing - significantly reducing file size (though it is still large!)


ewanharris (approved):
LGTM!

PR #12537, opened 2021-03-10
chore(android): material theme/widget improvements

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28301

**Summary:**
- Changed Titanium default theme for "statusBarColor" and "navigationBarColor" to use "?attr/colorPrimaryDark".
  * Makes it easier to custom theme an app with just "colorPrimary" and "colorPrimaryDark".
  * Was defaulting to this color before, but in a way that wasn't as easy to override.
- Added null reference guards to proxy classes in case it's used after release.
- Updated `TabGroup` and `TabbedBar` top tab style to scale icons to fit like how bottom style works.
- Improved `Ti.UI.Button` background and touch feedback handling.
  * Setting a "backgroundImage", "backgroundGradient", or "border*" properties used to log a nasty warning from the new `MaterialButton` we're using since it can cause it to be rendered incorrectly. Will now fallback to using `AppCompatButton` if you override its background.
  * Will now apply "touchFeedbackColor" correctly on `MaterialButton` when not overriding its background.

**Test:**
2. Verify that you do **NOT** see the following warning in the log.
`[WARN]  MaterialButton: MaterialButton manages its own background to control elevation, shape, color and states...`
3. Verify that you see an orange button and a blue gradient button like in the screenshot below. _(Note that the bottom blue image button currently fails to render on "master". This PR fixes it.)_
4. Tap on both buttons and verify that you see a yellow ripple effect.

AndroidButtons

```javascript
const colorButton = Ti.UI.createButton({
	title: "Orange Button",
	color: "white",
	backgroundColor: "orange",
	touchFeedbackColor: "yellow",
	touchFeedback: true,
	top: "20%",
});
colorButton.addEventListener("click", () => {
	Ti.API.info("@@@ 'Orange Button' was clicked on.");
});
window.add(colorButton);
const imageButton = Ti.UI.createButton({
	title: "Image Button",
	color: "white",
	backgroundImage: "android.resource://android/drawable/panel_picture_frame_bg_focus_blue",
	touchFeedbackColor: "yellow",
	touchFeedback: true,
	bottom: "20%",
});
imageButton.addEventListener("click", () => {
	Ti.API.info("@@@ Image Button was clicked on.");
});
window.add(imageButton);
window.open();
```


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
```diff
@@ -73,24 +80,51 @@ public TiUIButton(final TiViewProxy proxy)
...
+			|| hasBorder(proxy.getProperties())
+			|| hasGradient(proxy.getProperties());
 
+		// Create and set up the button.
+		// Note: MaterialButton does not support replacing its background drawable. Will log a nasty warning.
+		AppCompatButton btn;
+		if (hasCustomBackground) {
+			btn = new AppCompatButton(proxy.getActivity()) {
+				@Override
+				public boolean onFilterTouchEventForSecurity(MotionEvent event)
```
garymathews, line 96: Potential to move this out and remove duplicate code?

garymathews (approved):
CR: **PASS**

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIButton.java
jquick-axway: I thought about it, but the duplicated code was so small that I didn't think it was worth it. Especially since it would bloat it to the below. 🙄 
```java
public boolean onFilterTouchEventForSecurity(MotionEvent event)
{
	boolean isTouchAllowed = super.onFilterTouchEventForSecurity(event);
	return handleFilterTouchEventForSecurity(isTouchAllowed, event);
}

private boolean handleFilterTouchEventForSecurity(boolean isAllowed, MotionEvent event)
{
	if (!isAllowed) {
		fireSyncEvent(TiC.EVENT_TOUCH_FILTERED, dictFromEvent(event));
	}
	return isAllowed;
}
```

lokeshchdhry:
FR Passed.

PR #12541, opened 2021-03-10
fix(android): Check proxy on TiDrawableReference fromUrl

LalitaGertrudis:
Fix crash when loading images from url

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28385

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
@@ -157,6 +157,10 @@ public class TiDrawableReference
 	 */
 	public static TiDrawableReference fromUrl(KrollProxy proxy, String url)
 	{
+		if (proxy == null) {
+			return new TiDrawableReference(TiApplication.getAppCurrentActivity(), DrawableReferenceType.NULL);
+		}
+
 		if (url == null || url.length() == 0 || url.trim().length() == 0) {
 			return new TiDrawableReference(proxy.getActivity(), DrawableReferenceType.NULL);
 		}
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
@@ -20,6 +20,7 @@ import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.common.Log;
 import org.appcelerator.titanium.TiApplication;
 import org.appcelerator.titanium.TiBlob;
+import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.TiDimension;
 import org.appcelerator.titanium.TiFileProxy;
 import org.appcelerator.titanium.io.TiBaseFile;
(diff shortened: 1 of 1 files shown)


hansemannn:
@LalitaGertrudis Would you mind updating the URL to an actual one? Thx!

hansemannn:
cc @garymathews: @LalitaGertrudis reported that this was a regression in the ListView component from 9.3.0 onwards, indicating that this could be a side effect of the RecyclerView change. So maybe an alternate approach could be to check where in the ListView implementation this is caused and fix it there.

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
```diff
@@ -157,10 +157,21 @@ public static TiDrawableReference fromBlob(Activity activity, TiBlob blob)
...
+		if (proxy != null) {
+			activity = proxy.getActivity();
+		}
+
 		if (url == null || url.length() == 0 || url.trim().length() == 0) {
-			return new TiDrawableReference(proxy.getActivity(), DrawableReferenceType.NULL);
+			return new TiDrawableReference(activity, DrawableReferenceType.NULL);
+		}
+
+		if (proxy == null) {
```
garymathews, line 170: It looks like this doesn't need to use `proxy.resolveUrl()`, it can be simplified to:
```Java
return fromUrl(activity, TiUrl.resolve(TiC.URL_APP_PREFIX, url, null));
```

garymathews (changes requested):

garymathews:
##### TEST CASE
```JS
const items = [];

	items.push({ img: { image: 'https://picsum.photos/200' } });
}

	sections: [ Ti.UI.createListSection({ items }) ],
	templates: {
		template: {
			childTemplates: [{
				type: 'Ti.UI.ImageView',
				bindId: 'img'
			}]
		}
	},
});

win.add(listView);
win.open();
```

android/titanium/src/java/org/appcelerator/titanium/view/TiDrawableReference.java
garymathews: Oops! I should have been more clear, the whole `if (proxy == null) {` conditional scope can be removed so it looks something like this:
```Java
...
if (url == null || url.length() == 0 || url.trim().length() == 0) {
    return new TiDrawableReference(activity, DrawableReferenceType.NULL);
}
return fromUrl(activity, TiUrl.resolve(TiC.URL_APP_PREFIX, url, null));
```

garymathews (approved):
CR: **PASS**

ssjsamir (approved):
FR Passed, Tested using the following test case: 

```
const items = [];

	items.push({ img: { image: 'https://picsum.photos/200' } });
}

	sections: [ Ti.UI.createListSection({ items }) ],
	templates: {
		template: {
			childTemplates: [{
				type: 'Ti.UI.ImageView',
				bindId: 'img'
			}]
		}
	},
});

win.add(listView);
win.open();
```

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.2 Beta
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
iPhone 8 14.2
```

sgtcoolguy:
Manually cherry-picked to `master` and `10_0_X`

PR #12567, opened 2021-03-13
fix(ios): fix a few warnings

hansemannn:
Absolutely work-in-progress. Many warning are related to the "ObjCProxy" / getter-setter changes, so maybe @sgtcoolguy is best to review/finish this if there is time.


hansemannn:
Will do a clean one again.

PR #12580, opened 2021-03-15
refactor(android): use Geocoder for forward and reverse lookup

garymathews:
- Remove dependency on third-party lookup service
- Transition to using Googles Geocoder

##### TEST CASE
```JS
const scrollView = Ti.UI.createScrollView();
	top: 5,
	left: 5,
	color: 'white'
});

Ti.Geolocation.forwardGeocoder('440 N Bernardo Ave, Mountain View, CA 94043')
	.then(e => label.text = `${label.text}forwardGeocoder: ${JSON.stringify(e, null , 1)}\n`);

Ti.Geolocation.reverseGeocoder(37.4222617, -122.0862134)
	.then(e => label.text = `${label.text}reverseGeocoder: ${JSON.stringify(e, null , 1)}\n`);

Ti.Geolocation.forwardGeocoder('440 N Bernardo Ave, Mountain View, CA 94043',
	e => label.text = `${label.text}forwardGeocoderCallback: ${JSON.stringify(e, null , 1)}\n`);

Ti.Geolocation.reverseGeocoder(37.4222617, -122.0862134,
	e => label.text = `${label.text}reverseGeocoderCallback: ${JSON.stringify(e, null , 1)}\n`);

scrollView.add(label);
win.add(scrollView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28395)


android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
```diff
@@ -762,7 +767,36 @@ public boolean getLocationServicesEnabled()
 												   @Kroll.argument(optional = true) final KrollFunction callback)
 	{
 		return KrollPromise.create((promise) -> {
-			tiLocation.forwardGeocode(address, createGeocodeResponseHandler(callback, promise));
+			final KrollDict response = new KrollDict();
+
+			response.put(TiC.EVENT_PROPERTY_SOURCE, this);
+
+			try {
+				final List<Address> addresses = geocoder.getFromLocationName(address, 1);
```
jquick-axway, line 775: The `Geocoder.getFromLocationName()` blocks the current thread. Supposedly if you have network access but no Internet access, then this will block until it times-out. So, we'll need to do this on another thread.

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
```diff
@@ -780,7 +814,39 @@ public boolean getLocationServicesEnabled()
 								@Kroll.argument(optional = true) final KrollFunction callback)
 	{
 		return KrollPromise.create((promise) -> {
-			tiLocation.reverseGeocode(latitude, longitude, createGeocodeResponseHandler(callback, promise));
+			final KrollDict response = new KrollDict();
+
+			response.put(TiC.EVENT_PROPERTY_SOURCE, this);
+
+			try {
+				final List<Address> addresses = geocoder.getFromLocation(latitude, longitude, 10);
```
jquick-axway, line 822: Same as above. We'll need to do this on another thread.

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
```diff
@@ -762,7 +767,36 @@ public boolean getLocationServicesEnabled()
...
+
+					// Could not resolve address.
+					throw new Exception("Could not resolve address to location.");
+				}
+
+				// Success, resolve.
+				response.putCodeAndMessage(0, null);
+				promise.resolve(response);
+
+			} catch (Exception e) {
```
jquick-axway, line 789: **Question:** Will we end up in the `catch` block on a device that doesn't support reverse/forward lookup? Will this happen in an Android emulator without "Google APIs"?

That's fine if the above is true, but that would make this a breaking-change... which we have to eventually do anyways. I just want to make sure everyone is clear when this can happen. We might want to write a comment here to indicate the case(s) that cause this to happen.

jquick-axway (approved):
CR: Pass

sgtcoolguy:
A more apt conventional commit prefix is likely `refactor(android): use Geocoder for forward and reverse lookup` as this is intended as an internal refactoring and not an actual bug fix.

ssjsamir (approved):
FR Passed, tested using the test case above. 

**Test Environment**
```
MacOS Big Sur: 11.1 Beta 1
Xcode: 12.2 Beta
Java Version: 1.8.0_242
Android NDK: 21.3.6528147
Node.js: 12.18.1
""NPM":"5.0.0","CLI":"8.1.1""
Pixel XL 10.0
```

PR #12581, opened 2021-03-15
chore(ios): use CLGeocoder APIs to reverse geocode and forward geocode

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28396


tests/Resources/ti.geolocation.test.js
```diff
@@ -680,7 +680,7 @@ describe('Titanium.Geolocation', () => {
 					should(data.places[0]).have.property('latitude').which.is.a.Number();
 					should(data.places[0]).have.property('longitude').which.is.a.Number();
 					should(data.places[0].country).be.oneOf('USA', 'United States of America', 'United States');
-					should(data.places[0].state).be.eql('California');
+					should(data.places[0].state).be.oneOf('California', 'CA');
```
jquick-axway, line 683: Interesting. This is an example on how this is a breaking-change and where the results may differ between Android and iOS... but we've got to do it anyways.

jquick-axway: We're missing "state" property documentation in `GeocodedAddress` and `ForwardGeocodeResponse`. Would you  mind adding it please? We can indicate that the "state" property might return the full name or abbreviated name.
https://github.com/appcelerator/titanium_mobile/blob/master/apidoc/Titanium/Geolocation/Geolocation.yml

iphone/Classes/GeolocationModule.m
```diff
@@ -502,35 +421,73 @@ - (BOOL)hasCompass
...
+                 completionHandler:^(NSArray<CLPlacemark *> *_Nullable placemarks, NSError *_Nullable error) {
+                   if (error != nil) {
+                     [rcb requestError:error];
+                   } else {
+                     NSMutableDictionary *events = [TiUtils dictionaryWithCode:0 message:nil];
+
+                     NSMutableArray *places = [NSMutableArray array];
+                     for (CLPlacemark *placemark in placemarks) {
+                       NSMutableDictionary *dict = [NSMutableDictionary dictionary];
+                       NSMutableString *address = [NSMutableString string];
```
sgtcoolguy, line 443: I don't think we should be building this string up ourselves. `CLPlacemark` has a `postalAddress` property which is a `CNPostalAddress` - you can then format it via a `CNPostalAddressFormatter`: https://developer.apple.com/documentation/contacts/cnpostaladdressformatter?language=objc

Something like:
``` obj-c
NSString *address = [CNPostalAddressFormatter stringFromPostalAddress:placemark.postalAddress style:CNPostalAddressFormatterStyleMailingAddress];
```

sgtcoolguy (changes requested):

tests/Resources/ti.geolocation.test.js
vijaysingh-axway: Added doc for state property.

iphone/Classes/GeolocationModule.m
vijaysingh-axway: Thanks for pointing. Updated.

sgtcoolguy:
For reference, the conventional commit should be more like `refactor(ios): use CLGeocoder APIs to reverse geocode and forward geocode` since this is intended as an internal refactoring and not a bug fix, new feature/improvement - and definitely isn't `build` or `chore`. `chore` is typically a catch-all for things, but a better prefix should always be preferred first. I think we're just used to seeing it from dependabot/greenkeeper for updating dependencies so everyone just defaults to it. But in fact, even dependabot now uses `build` to tag dependency updates.

sgtcoolguy (approved):

PR #12614, opened 2021-03-17
fix: replace shortened appcelerator.com URLs with long URLs

cb1kenobi:
**JIRA:** https://jira.axway.com/browse/TIMOB-21146

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -1091,13 +1091,13 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
 							logger.error(__('Unable to find any non-expired development provisioning profiles that match the app id "%s".', appId) + '\n');
 						}
 						logger.log(__('You will need to log in to %s with your Apple Developer account, then create, download, and install a profile.',
-							'http://appcelerator.com/ios-dev-certs'.cyan) + '\n');
+							'https://developer.apple.com/account/ios/certificate/certificateList.action?type=development'.cyan) + '\n');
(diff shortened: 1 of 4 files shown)


iphone/cli/commands/_build.js
```diff
@@ -1091,13 +1091,13 @@ iOSBuilder.prototype.configOptionPPuuid = function configOptionPPuuid(order) {
 							logger.error(__('Unable to find any non-expired development provisioning profiles that match the app id "%s".', appId) + '\n');
 						}
 						logger.log(__('You will need to log in to %s with your Apple Developer account, then create, download, and install a profile.',
-							'http://appcelerator.com/ios-dev-certs'.cyan) + '\n');
+							'https://developer.apple.com/account/ios/certificate/certificateList.action?type=development'.cyan) + '\n');
```
ewanharris, line 1094: This way redirects for me, but it looks like the new URL is `https://developer.apple.com/account/resources/certificates/list`, I can't find a way to toggle development/production via the URL.

Should we start using the newer URL?

ewanharris (commented):
Just a comment about the developer.apple URL

sgtcoolguy:
I should note that it's actually pretty likely the new guide content URLs will change. Once we cut over from the wiki to markdown, I would think we'd want to clean up the guide paths/URLs to be less ugly. So it may make sense to use an url shortener/redirect service as an intermediary to protected against that?

iphone/cli/commands/_build.js
cb1kenobi: Fixed!

PR #12621, opened 2021-03-19
style(android): fix java lint warnings

jquick-axway:
**Summary:**
Fixes most (but not all) Java lint warnings reported by Android Studio.
- Removed all `@module.api` javadoc tags.
- Made member variables `final` if assigned upon declaration.
- Made inner classes static if possible.
- Leveraged Java 7 diamond operator. Ex: `ArrayList<String> list = new ArrayList<>();`
- Initialized `toArray()` calls with zero length array. (Slight optimization.)

### DIFF
--- a/android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
+++ b/android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
@@ -541,7 +541,7 @@ public class AndroidModule extends KrollModule
 	public static final int FOREGROUND_SERVICE_TYPE_CAMERA = ServiceInfo.FOREGROUND_SERVICE_TYPE_CAMERA;
 
 	protected RProxy r;
-	private LinkedList<BroadcastReceiverProxy> registeredBroadcastReceiverProxyList = new LinkedList<>();
+	private final LinkedList<BroadcastReceiverProxy> registeredBroadcastReceiverProxyList = new LinkedList<>();
 
 	private static final int REQUEST_CODE = 99;
 
@@ -655,7 +655,7 @@ public class AndroidModule extends KrollModule
 	public boolean hasPermission(Object permissionObject)
 	{
 		if (Build.VERSION.SDK_INT >= 23) {
-			ArrayList<String> permissions = new ArrayList<String>();
+			ArrayList<String> permissions = new ArrayList<>();
 			if (permissionObject instanceof String) {
 				permissions.add((String) permissionObject);
 			} else if (permissionObject instanceof Object[]) {
@@ -705,8 +705,7 @@ public class AndroidModule extends KrollModule
 				if (filteredPermissions.size() > 0) {
 					TiBaseActivity.registerPermissionRequestCallback(REQUEST_CODE, permissionCallback,
 						callbackThisObject, promise);
-					currentActivity.requestPermissions(filteredPermissions.toArray(
-						new String[filteredPermissions.size()]), REQUEST_CODE);
+					currentActivity.requestPermissions(filteredPermissions.toArray(new String[0]), REQUEST_CODE);
 					return;
 				}
 			}
--- a/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
+++ b/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
@@ -102,8 +102,7 @@ import android.os.Message;
  * accuracy, frequency properties or even changing modes are respected and kept but don't actually get applied on the OS until
  * the listener count is greater than 0.
  */
-// TODO deprecate the frequency and preferredProvider property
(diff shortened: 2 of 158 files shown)


android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
```diff
@@ -705,8 +705,7 @@ public boolean hasPermission(Object permissionObject)
 				if (filteredPermissions.size() > 0) {
 					TiBaseActivity.registerPermissionRequestCallback(REQUEST_CODE, permissionCallback,
 						callbackThisObject, promise);
-					currentActivity.requestPermissions(filteredPermissions.toArray(
-						new String[filteredPermissions.size()]), REQUEST_CODE);
+					currentActivity.requestPermissions(filteredPermissions.toArray(new String[0]), REQUEST_CODE);
```
sgtcoolguy, line 708: I was always used to pre-sizing the arrays, but after some Googling it turns out the JDKs have been optimized to handle zero-length arrays as fast or faster now, so was unaware that the ugly boilerplate code is no longer necessary!

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
```diff
@@ -381,21 +377,6 @@ private void propertyChangedFrequency(Object newValue)
 		double frequencyProperty = TiConvert.toDouble(newValue) * 1000;
 	}
 
-	/**
-	 * Handles property change for Ti.Geolocation.preferredProvider
-	 *
-	 * @param newValue					new preferredProvider value
-	 */
-	@SuppressLint("MissingPermission")
-	private void propertyChangedPreferredProvider(Object newValue)
```
sgtcoolguy, line 390: Looks like this was basically a no-op, so I can see why it's being removed - but our docs say deprecated since 2.0 with no removal..

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIMaskedImage.java
```diff
@@ -205,7 +205,7 @@ private void updateImageViewScaleType()
 		ImageView.ScaleType scaleType = ImageView.ScaleType.FIT_XY;
 		if ((layoutParams.optionWidth == null) && (layoutParams.optionHeight == null)) {
 			if (!layoutParams.autoFillsWidth && !layoutParams.autoFillsHeight) {
-				if (layoutParams.sizeOrFillWidthEnabled && layoutParams.sizeOrFillWidthEnabled) {
+				if (layoutParams.sizeOrFillWidthEnabled && layoutParams.sizeOrFillHeightEnabled) {
```
sgtcoolguy, line 208: good catch, looks like a bug was hidden here

android/runtime/common/src/java/org/appcelerator/kroll/KrollLogging.java
```diff
@@ -119,15 +119,8 @@ public void log(String level, String... args)
...
-		String msg;
 		int length = (args == null ? 0 : args.length);
-
-		if (length > 0) {
-			msg = args[0];
-		} else {
-			msg = new String();
-		}
-
+		String msg = (length > 0) ? args[0] : "";
```
sgtcoolguy, line 123: This feels very much like a spot we should be using `StringBuilder`

sgtcoolguy (approved):
LGTM. Made me go cross-eyed looking over 157 files!

Only things I really saw are in my comments:
- We should update the Geolocation docs to note whenever `preferredProvider` was actually removed.
- The logging method that builds up a String should really use `StringBuilder` if it's going to keep appending things.

android/modules/android/src/java/ti/modules/titanium/android/AndroidModule.java
jquick-axway: Yeah, the old approach used to be optimal (only does 1 allocation and uses the array given) but this approach is now the optimal way to do it and simpler to write.

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
jquick-axway: I just noticed the `Ti.Geolocation.frequency` property (deprecated in 2.0.0) is dead code too. I'll clean this up.

jquick-axway: I've updated the docs. Confirmed removed as of Titanium 8.0.0.

android/runtime/common/src/java/org/appcelerator/kroll/KrollLogging.java
jquick-axway: Just optimized it now.
However, our `Ti.API` logging functions no longer call this class' variable arguments methods anymore.

PR #12625, opened 2021-03-19
Fix usage of variables before definition

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28401

This PR is 2 parts:

* Add the [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) rule to the eslint config for variables only
* Fix the errors that arise from that. Whilst not all of them would cause the same issue due to being used in event handlers, it's better to be safe imo

This fixes the issue in the linked ticket (originally from https://github.com/appcelerator/titanium_mobile/discussions/12600)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,7 @@
     "sourceType": "script"
   },
   "rules": {
+    "no-use-before-define": ["error", { "functions": false, "classes": false, "variables": true }],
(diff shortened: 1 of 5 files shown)


sgtcoolguy (approved):

ewanharris:
Not relevant for 9_3_X

PR #12631, opened 2021-03-20
feat(docs): add information about material Buttons

m1ga:
Adding a quick info on how to style buttons when updating to Titanium SDK 10 to get back the old look & feel and lower-case text:

6ZOm0nm
_(image from Stackoverflow, so I didn't include it)_


sgtcoolguy (approved):
LGTM, but @jquick-axway should take a quick look too.

jquick-axway (approved):

jquick-axway:
Yeah, looks good. We can merge it.

m1ga:
I had some Views with Labels as buttons too and suddenly the normal buttons looked different. It was mainly a visual thing since I want the font inside the buttons to have the same spacing as my other labels and textfields

jquick-axway:
@m1ga , I see. Well we did add a new `Ti.UI.BUTTON_STYLE_TEXT` in Titanium 10.0.0 which will show a text-only material button on Android. And it will show a button ripple effect when tapped on.
https://titaniumsdk.com/api/titanium/ui/button.html#style

PR #12665, opened 2021-03-26
feat(ios): specify additional parameters to SFSymbol system image

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28405


apidoc/Titanium/UI/iOS/iOS.yml
```diff
@@ -1710,3 +1713,18 @@ properties:
...
+---
+name: SystemImageParameters
+summary: |
+    Dictionary object of parameters for the 2nd parameter of the <Titanium.UI.iOS.systemButton> method
+    to pass additional configuration options like `weight` and `size`.
+properties:
+  - name: weight
+    summary: |
+        The symbol weight to use for the SF Symbol.
+        Can be one of `ultralight`, `light`, `thin`, `normal`, `semibold`, `bold`, `heavy` or `black`.
```
vijaysingh-axway, line 1726: I think we can make it constant. So you don't have to write utility function to convert it in constants in TiUtils.

hansemannn: Agree, but since it's iOS only right now and people are used to use `fontWeight: 'bold` etc, I made it a bit more "handy" to use. Maybe both like for `textAlign`?

apidoc/Titanium/UI/iOS/iOS.yml
```diff
@@ -27,13 +27,16 @@ methods:
...
-        One can download SF Symbols app from https://developer.apple.com/design/downloads/SF-Symbols.dmg and 
-        use it to lookup the inbuilt SF Symbol name.
+        More details on SF Symbols can be found [here](https://developer.apple.com/design/human-interface-guidelines/sf-symbols/overview).
+        One can download SF Symbols app from [here](https://developer.apple.com/design/downloads/SF-Symbols.dmg) and
+        use it to lookup the built-in SF Symbol name.
     parameters:
       - name: name
         summary: Name of SF Symbol.
         type: String
+      - name: parameters
```
sgtcoolguy, line 37: likely needs a `since` value of `10.0.0` and also possibly an `osversion` value?

sgtcoolguy: oh and `optional: true`

hansemannn: Added `since` and `optional` keys. You mean `osver`? Thats already defined by the method docs (iOS 13+) and the "new" API was also added in iOS 13 already.

hansemannn: > [ERROR] Titanium.UI.iOS: found 1 error(s), 0 warning(s)!
	methods[systemImage].parameters[parameters].since - Invalid key(s) in Titanium.UI.iOS: since

Nearly forgot: The `since` key is not available everywhere!

vijaysingh-axway (approved):
Looks good!

ewanharris (approved):
FR Pass. Able to specify weight and size for SF Symbols

PR #12661, opened 2021-03-25
fix(android): remove getter and setter methods from TabGroup

garymathews:
- Remove `getActiveTab()` and `setActiveTab()` from `Ti.UI.TabGroup`
- Remove `getTabs()` from `Ti.UI.TabGroup`
- Remove overiding of `TabGroup` via common extension

##### TEST CASE
```JS
const win_a = Ti.UI.createWindow({
	backgroundColor: 'blue',
	title: 'Blue'
});
const win_b = Ti.UI.createWindow({
	backgroundColor: 'red',
	title: 'Red'
});
const tab_a = Ti.UI.createTab({
	window: win_a,
	title: 'Blue'
});
const tab_b = Ti.UI.createTab({
	window: win_b,
	title: 'Red'
});
const tabGroup = Ti.UI.createTabGroup({
	tabs: [ tab1, tab2 ],
	activeTab: 1 // Open on second tab.
});

win_a.addEventListener('click', e => {

	// Switch to second tab.
	tabGroup.activeTab = 1;
});

win_b.addEventListener('click', e => {

	// Reset tabs.
	tabGroup.tabs = [ tab_a, tab_b ];
});

tabGroup.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28404)

PR #12677, opened 2021-03-30
docs: mark accessors that were autogenerated but documented as removed in 10

ewanharris:
This PR is related to #12661 and #12654 and removes some accessor functions that were documented properly (i.e existed in the yml docs and were not autogenerated by the docs scripts) but were actually backed by the auto-generation of accessors. Namely:

* Ti.Android.Intent.getData
* Ti.Android.MenuItem.setCheckable
* Ti.Android.MenuItem.setChecked
* Ti.Android.MenuItem.setEnabled
* Ti.Android.MenuItem.setVisible
* Ti.Android.MenuItem.isActionViewExpanded
* Ti.Android.MenuItem.isCheckable
* Ti.Android.MenuItem.isChecked
* Ti.Android.MenuItem.isEnabled
* Ti.Android.MenuItem.isVisible
* Ti.Network.HTTPClient.setTimeout (this is fine to remove as it is not part of the spec)
* Ti.UI.TableView.setHeaderPullView (iOS only)

It does not (yet) deal with the following APIs:

* Ti.UI.Tab.setWindow - This exists on Android but not iOS
* Sound.isLooping - This exists on iOS but not Android
* Sound.isPlaying - This exists on iOS but not Android
* Sound.isPaused - This exists on iOS but not Android

Personally, I think we should just remove the above APIs rather than have a difference between iOS and Android. But, that's like just my opinion so I'm open to suggestions

[JIRA](https://jira.appcelerator.org/browse/TIMOB-28407)

### DIFF
--- a/apidoc/Titanium/Android/MenuItem.yml
+++ b/apidoc/Titanium/Android/MenuItem.yml
@@ -101,7 +101,7 @@ methods:
     deprecated:
       since: "10.0.0"
       removed: "10.0.0"
-      notes: Use the <Titanium.Android.MenuItem.checkable> property instead.
+      notes: Use the <Titanium.Android.MenuItem.visible> property instead.
 
   - name: setCheckable
     summary: Sets the [checkable](Titanium.Android.MenuItem.checkable) state of the menu item.
--- a/apidoc/Titanium/Android/MenuItem.yml
+++ b/apidoc/Titanium/Android/MenuItem.yml
@@ -60,27 +60,48 @@ methods:
         on devices running Android 4.0 (API level 14) and greater.
     returns:
         type: Boolean
+    deprecated:
+      since: "10.0.0"
+      removed: "10.0.0"
+      notes: Use the <Titanium.Android.MenuItem.actionViewExpanded> property instead.
+
 
   - name: isCheckable
     summary: |
         Returns the [checkable](Titanium.Android.MenuItem.checkable) state of the menu item.
     returns:
         type: Boolean
+    deprecated:
+      since: "10.0.0"
+      removed: "10.0.0"
+      notes: Use the <Titanium.Android.MenuItem.checkable> property instead.
 
   - name: isChecked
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/Android/MenuItem.yml
```diff
@@ -60,55 +60,92 @@ methods:
...
+      notes: Use the <Titanium.Android.MenuItem.enabled> property instead.
 
   - name: isVisible
     summary: Returns the [visible](Titanium.Android.MenuItem.visible) state of the menu item.
     returns:
         type: Boolean
+    deprecated:
+      since: "10.0.0"
+      removed: "10.0.0"
+      notes: Use the <Titanium.Android.MenuItem.checkable> property instead.
```
sgtcoolguy, line 104:
```suggestion
      notes: Use the <Titanium.Android.MenuItem.visible> property instead.
```

sgtcoolguy (changes requested):

vijaysingh-axway:
I think, we can deprecate following APIs in favor of corresponding properties.  Probably we can create a ticket to track it.
   Sound.isLooping - This exists on iOS but not Android
   Sound.isPlaying - This exists on iOS but not Android
   Sound.isPaused - This exists on iOS but not Android

sgtcoolguy (approved):

sgtcoolguy:
@ewanharris We chatted on teams about this, but to summarize my thoughts:

Personally I think the get/set methods can be marked deprecated/removed in 10.0.0 since they fell under what we were attempting to get rid of around autogenerated accessors. I think the `is*` boolean query methods are less obvious and we should deprecate them in 10.0.0, but not remove them.

So we should:
-  add back the `@Kroll.method` annotations for the `MenuItem` ones and not mark them removed yet in docs
- mark the `Sound.is*` methods deprecated in docs.
- remove `Ti.UI.Tab.setWindow` and note it as such in the docs

PR #12682, opened 2021-04-01
fix(android): improve selection validation

garymathews:
- Improve selection validation to prevent `OutOfBounds` exception

##### TEST CASE
- Could not reproduce, no test case was provided.

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28410)


jquick-axway (approved):
CR: Pass

PR #12687, opened 2021-04-05
Revert "perf(ios): use new API for rendering view to image"

hansemannn:
This reverts commit c7e65a507a060f49ad4db10888ad9d1c2c7bd745. See the discussion in the attached JIRA ticket + [this SO thread](https://stackoverflow.com/questions/33592307/what-is-expected-performance-of-drawviewhierarchyinrect-vs-renderincontext).

Fixes a couple of regressions:
- Broken navigation transitions (due to blocking UI thread)
- `toImage()` to not work (e.g. for generating annotations or static images) on the first rendering stage

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28394


sgtcoolguy (approved):
I'm fine with this if you can confirm it actually fixes your issues. I was attempting to use a newer API which is purported to be faster, but presumably it had some knock-on side effects I wasn't aware of.

hansemannn:
The move was an absolutely good idea! And I actually saw the commit back then and "silently" approved it. You can definitely reproduce the issue in the latest test case I provided on JIRA, where there should be 100 red squares which are not rendered.

jquick-axway:
Do you think it's an issue with setting `afterScreenUpdates` to `YES`?
This suggests that rendering will be deferred.

hansemannn:
@jquick-axway Unfortunately that does not work:
> Drawing a view (0x7f889c0199d0, TiUIView) that has not been rendered at least once requires afterScreenUpdates:YES.

The problem is that the method is sync by default. But it is already possible to pass a callback, so maybe move the existing one to `toImageSync()` or move the async part (with full promise support) to `toImageAsync()`. That way, the async variant could run on a background thread to be non-blocking, although it may require the UI thread because of the rendering operation. But still, having it async could fix it.

hansemannn:
@jquick-axway Unfortunately that does not work:
> Drawing a view (0x7f889c0199d0, TiUIView) that has not been rendered at least once requires afterScreenUpdates:YES.

The problem is that the method is sync by default. But it is already possible to pass a callback, so maybe move the existing one to `toImageSync()` or move the async part (with full promise support) to `toImageAsync()`. That way, the async variant could run on a background thread to be non-blocking, although it may require the UI thread because of the rendering operation. But still, having it async could fix it.

PR #12697, opened 2021-04-07
build: add back linux when packaging on any os with --all

ewanharris:
Removing linux caused the builds.appc site to break for newer builds as it has an expectation that
there will be 3 platforms for every build released for master/10_0_X branches

### DIFF
--- a/build/lib/builder.js
+++ b/build/lib/builder.js
@@ -17,8 +17,6 @@ const DIST_DIR = path.join(ROOT_DIR, 'dist');
 const TMP_DIR = path.join(DIST_DIR, 'tmp');
 
 // Platforms/OS mappings
-// NOTE: 'linux' could be added, but is not officially supported.
-// Specifying --all will only produce a linux output on a linux host.
 const ALL_OSES = [ 'win32', 'osx', 'linux' ];
 const ALL_PLATFORMS = [ 'ios', 'android' ];
 const OS_TO_PLATFORMS = {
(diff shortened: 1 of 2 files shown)


m1ga:
ahhhh...that's why the tab is empty :smile: :+1:

hansemannn:
I just filed a JIRA for this, funny! :-D Here we go: https://jira.appcelerator.org/browse/TIMOB-28414

build/lib/builder.js
```diff
@@ -19,7 +19,7 @@ const TMP_DIR = path.join(DIST_DIR, 'tmp');
 // Platforms/OS mappings
 // NOTE: 'linux' could be added, but is not officially supported.
 // Specifying --all will only produce a linux output on a linux host.
-const ALL_OSES = Array.from(new Set([ 'win32', 'osx', thisOS() ]));
+const ALL_OSES = [ 'win32', 'osx', 'linux' ];
```
garymathews, line 22: May want to remove the _NOTE_ above?

I'm also wondering if `--all` should explicitly include a platform we don't support? If we've fixed our build page, maybe this should stay as is?

sgtcoolguy: we haven't fixed our build page, or modified our release scripts to expect this (no linux zip) AFAIK. So not having this will break both still. (And may break some CI builds where I think we may build say an Android native module without unit tests so can use a linux node/sdk)

ewanharris: appcelerator/titanium_build/pull/3 fixes the build page, I've just pushed a commit that removes that comment too

PR #12700, opened 2021-04-07
fix(ios): do not pop error dialog, guard when ti.kernel.js cannot be read

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28408

**Optional Description:**

Most likely scenario is failure to decrypt. We need to avoid opening a dialog
because that will prevent the 'Security Violation' error dialog that the
decryption code will pop.

This change is really about guarding and logging errors that can happen during the earliest app startup phase. 

If we fail to load `ti.kernel.js`  that's the best indicator that decryption failed. So log a simple error string about it and *don't* pop our fancy stack trace error dialog - or else the appc-verify "Security Violation" dialog won't open (because a modal dialog is already up).

If the file fails to eval or export a valid bootstrap function, or calling that function fails - log an error string noting it and pop the stack trace error dialog/modal.

In any of the above error cases, *refuse* to load `ti.main.js` (which wraps and runs the `app.js`) - as attempting ti is about 100% guaranteed to fail, and would also pop an error dialog modal (again blocking any possible error dialog the verify encrypt/decrypt code would pop).

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
@@ -348,47 +348,74 @@ CFMutableSetRef krollBridgeRegistry = nil;
   // Load ti.kernel.js (kroll.js equivalent)
   NSURL *bootstrapURL = [TiHost resourceBasedURL:@"ti.kernel.js" baseURL:NULL];
   NSString *source = [AssetsModule readURL:bootstrapURL];
-
-  JSValue *bootstrapFunc = [objcJSContext evaluateScript:source withSourceURL:bootstrapURL];
-  if (objcJSContext.exception != nil) {
+  if (source == nil || source.length == 0) {
+    NSLog(@"[ERROR] Error reading ti.kernel.js, source is nil/empty. Most likely due to failure to decrypt via remote policy and device offline.");
     evaluationError = YES;
-    [TiExceptionHandler.defaultExceptionHandler reportScriptError:objcJSContext.exception inJSContext:objcJSContext];
-  }
-  [bootstrapFunc callWithArguments:@[ global, [[KrollModule alloc] init] ]];
-  if (objcJSContext.exception != nil) {
-    evaluationError = YES;
-    [TiExceptionHandler.defaultExceptionHandler reportScriptError:objcJSContext.exception inJSContext:objcJSContext];
-  }
-
-  JSValue *titanium = global[@"Ti"];
-  TopTiModule *module = [titanium toObject];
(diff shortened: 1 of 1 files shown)


vijaysingh-axway (approved):
CR passed.
I verified on device. It's working fine.

ssekhri:
FR Passed
Verified on:
Mac OS: 10.15.4
SDK: 10.1.0.v20210407101216
Appc CLI: 8.1.1
JDK: 11.0.6
Node: 12.16.1
Studio: 6.0.0.202005141803
Xcode: 12.3
Device: iPhone X(v13.6.1)

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -348,47 +348,74 @@ - (void)didStartNewContext:(KrollContext *)kroll
...
+      NSLog(@"[ERROR] Error eval'ing ti.kernel.js bootstrap script. Please contact Titanium developers and file a bug report.");
+      evaluationError = YES;
+      [TiExceptionHandler.defaultExceptionHandler reportScriptError:objcJSContext.exception inJSContext:objcJSContext];
+    } else if (bootstrapFunc == nil || [bootstrapFunc isUndefined]) {
+      // it didn't export a bootstrap function! Most likely reason is we hosed the file
+      NSLog(@"[ERROR] Error eval'ing ti.kernel.js, bootstrap function is undefined/nil. Please contact Titanium developers and file a bug report.");
+      evaluationError = YES;
+      [TiExceptionHandler.defaultExceptionHandler reportScriptError:objcJSContext.exception inJSContext:objcJSContext];
+    } else {
+      // ti.kernel.js eval'd OK, so now let's run the exported bootstrap function
```
hansemannn, line 368: Just a question: Can the "success" case be called earlier, e.g. with reverted if statements? It feels like, since this is the "default case", it should be ran as quick as possible. Right now, for every new context, there are at least 4 new if-statements that all need to access JSCore selectors before completing. I hope this does not influence performance, especially for larger projects

sgtcoolguy: regardless of the `if` order, it's still going to be branched code under the hood. I don't think this is a significant concern performance-wise. A "context" is only ever created at app startup, or when a background service launches. It's not like we're creating contexts all over the place. The difference here is that we check if the loaded `ti.kernel.js` source is not nil/empty, and that the exported bootstrap function is not nil or undefined. The exception guards already existed.

hansemannn: Thanks for pointing this out!

PR #12706, opened 2021-04-08
fix(ios): use weak instead of assign in properties for Objects

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28413
https://jira.appcelerator.org/browse/TIMOB-28344
In general 'assign' is used for primitive data types and 'weak' is used for objects. If object is deallocated from anywhere, any 'weak' reference to that object is set nil.  In assign, if object is deallocated you will still have a pointer to it's memory (which is garbage now)


sgtcoolguy:
Could this be the cause of https://jira.appcelerator.org/browse/TIMOB-28296 ? There we have pretty consistent crashes with `Ti.UI.Picker` on macOS for dates - where some underlying object is GC'd and we're trying to send it messages.
[mocha_2021-04-09-082446_macos-walle.crash.txt](https://github.com/appcelerator/titanium_mobile/files/6297563/mocha_2021-04-09-082446_macos-walle.crash.txt)

Do we need to make the `TiUIPicker`'s `UIControl *picker` field weak?

vijaysingh-axway:
> Could this be the cause of https://jira.appcelerator.org/browse/TIMOB-28296 ? There we have pretty consistent crashes with `Ti.UI.Picker` on macOS for dates - where some underlying object is GC'd and we're trying to send it messages.
> [mocha_2021-04-09-082446_macos-walle.crash.txt](https://github.com/appcelerator/titanium_mobile/files/6297563/mocha_2021-04-09-082446_macos-walle.crash.txt)
> 
> Do we need to make the `TiUIPicker`'s `UIControl *picker` field weak?

It doesn't look we have to make `UIControl *picker` field weak . I'll look in this crash.

sgtcoolguy (approved):

PR #12712, opened 2021-04-11
feat(android): add password toggle

m1ga:
[jira-archive ticket](https://jira-archive.titaniumsdk.com/TIMOB/TIMOB-27531)

**Optional Description:**
So it won't get lost in the JIRA tickets. Credits go to @drauggres! He made the whole code in the JIRA ticket.

```javascript
    backgroundColor: '#000000',
    layout: 'vertical'
});

const edit1 = Ti.UI.createTextField({
    hintText: 'no pwd mask',
    // common properties

    hintTextColor: '#8f000000',
    backgroundColor: '#ffffff',
    color: '#000000',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    top: 10,
    bottom:10,
    left: 20,
    right: 20
});

const edit2 = Ti.UI.createTextField({
    hintText: 'pwd mask. no toggle',
    passwordMask: true,
    // common properties

    hintTextColor: '#8f000000',
    backgroundColor: '#ffffff',
    color: '#000000',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    top: 10,
    bottom:10,
    left: 20,
    right: 20
});

const edit3 = Ti.UI.createTextField({
    hintText: 'pwd mask. toggle. default toggle colors',
    passwordMask: true,
    passwordVisibilityToggleEnabled: true,

    // common properties
    hintTextColor: '#8f000000',
    backgroundColor: '#ffffff',
    color: '#000000',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    top: 10,
    bottom:10,
    left: 20,
    right: 20
});

const edit4 = Ti.UI.createTextField({
    hintText: 'no pwd mask. toggle. default toggle colors',
    passwordMask: false,
    passwordVisibilityToggleEnabled: true,

    // common properties
    hintTextColor: '#8f000000',
    backgroundColor: '#ffffff',
    color: '#000000',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    top: 10,
    bottom:10,
    left: 20,
    right: 20
});

const edit5 = Ti.UI.createTextField({
    hintText: 'pwd mask. toggle. custom toggle colors',
    passwordMask: true,
    passwordVisibilityToggleEnabled: true,
    passwordVisibilityToggleColor: '#33e33e',
    passwordVisibilityToggleTintColor: '#e33e33',

    // common properties
    hintTextColor: '#8f000000',
    backgroundColor: '#ffffff',
    color: '#000000',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE,
    top: 10,
    bottom:10,
    left: 20,
    right: 20
});


win.add(edit1);
win.add(edit2);
win.add(edit3);
win.add(edit4);
win.add(edit5);

win.open();
```
pwd_toggle


jquick-axway:
The `setPasswordVisibility*()` APIs are deprecated. We're supposed to use the `setEndIcon*()` methods instead.
https://developer.android.com/reference/com/google/android/material/textfield/TextInputLayout

The `setEndIcon*()` methods kind of line-up with iOS' `rightButton*` APIs (which is not a right-to-left friendly name). Unfortunately, iOS does not have a built-in password toggle button. iOS does support a built-in "clear" button and a custom button, both of which Android's `TextInputLayout` supports as well. We could make this cross-platform, with the only issue being RTL (right-to-left) handling is different here. Hmm...
https://titaniumsdk.com/api/titanium/ui/textfield.html#rightbutton
https://developer.apple.com/documentation/uikit/uitextfield/1619596-rightview

m1ga:
Oh, thanks for the info! Didn't look those up. Will try to transform it to use the `setEndIcon` way in the first step

m1ga:
so for now I moved to `setEndIcon` so that this feature could potentially been added to a future version.
The `rightIcon` parity should be handled in a different PR that.

hansemannn:
I feel like this is a very specific, Android-only API that hasn't been requested many times in the past. Therefore I would rather go with a JS implementation of the show / hide behavior that can be cross-platform as well. Would that be okay @m1ga ?

m1ga:
Fine with me. I've added the GIF from the old ticket to the first post to show what it is adding.
I think you can do everything with a imageview + click event on top of the textfield so we don't need properties for it. 

The only thing where this could be useful for is perhaps accessibility? But that is just a guess since, don't have prove for that.

hansemannn:
I think that could be done with the `accessibilityLabel` API if needed. I'll close it for now then! Maybe iOS adds some kind of parity in future versions, then we can always reopen it 👍

PR #12721, opened 2021-04-14
fix(ios): unhandled native exceptions should show an error dialog

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28420

**Summary:**
- Native exceptions invoked by JS should show an error dialog with a trace like Android.
- Currently native exceptions like this are logged and then hard-crashes back to home screen..

---
**Unhandled Native Exception Test:**
2. Verify an exception dialog is shown. _(Used to crash back to the home screen.)_

```javascript
Ti.Filesystem.openStream(0, false);
```

---
**Unhandled JavaScript Exception Test:**
2. Verify an exception dialog is shown.

```javascript
Ti.uhOh();
```


sgtcoolguy (approved):
LGTM

hansemannn:
Same here, tested and working well! Already cherry picked for a new production release :-)

iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -156,8 +156,19 @@ - (id)initWithDictionary:(NSDictionary *)dictionary
     if (_backtrace == nil) {
       _backtrace = [[[dictionary objectForKey:@"stack"] description] copy];
     }
-    _nativeStack = [[dictionary objectForKey:@"nativeStack"] copy];
     _dictionaryValue = [dictionary copy];
+
+    id nativeStackObject = [dictionary objectForKey:@"nativeStack"];
+    if (nativeStackObject) {
+      if ([nativeStackObject isKindOfClass:[NSArray class]]) {
+        nativeStackObject = [nativeStackObject copy];
```
vijaysingh-axway, line 164: NO need to copy here. copy should be used with  _nativeStack = [nativeStackObject copy];

iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -156,8 +156,19 @@ - (id)initWithDictionary:(NSDictionary *)dictionary
...
+      } else if ([nativeStackObject isKindOfClass:[NSString class]]) {
+        nativeStackObject = [[nativeStackObject componentsSeparatedByString:@"\n"] retain];
```
vijaysingh-axway, line 166: Same as above. No need to retain. copy can be used _nativeStack = [nativeStackObject copy];

iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -156,8 +156,19 @@ - (id)initWithDictionary:(NSDictionary *)dictionary
...
+      } else {
+        nativeStackObject = nil;
+      }
+    }
+    _nativeStack = nativeStackObject;
```
vijaysingh-axway, line 171: Probably you can replace it with following -
```suggestion
    if ([nativeStackObject isKindOfClass:[NSString class]]) {
      nativeStackObject = [nativeStackObject componentsSeparatedByString:@"\n"];
    }

    if (![nativeStackObject isKindOfClass:[NSArray class]]) {
      nativeStackObject = nil;
    }
    _nativeStack = [nativeStackObject copy];
```

jquick-axway: What you're recommending does a needless copy if `nativeStackObject` is a string... and a stack trace string is normally quite large. My code retains the array created by the `componentsSeparatedByString` method. I still prefer the way I wrote it.

That said, I probably don't need the `if (nativeStackObject) {}` check since we can do `nil` message calls. So we can probably shorten it to this if we want.
```objc
id nativeStackObject = [dictionary objectForKey:@"nativeStack"];
if ([nativeStackObject isKindOfClass:[NSArray class]]) {
	nativeStackObject = [nativeStackObject copy];
} else if ([nativeStackObject isKindOfClass:[NSString class]]) {
	nativeStackObject = [[nativeStackObject componentsSeparatedByString:@"\n"] retain];
} else {
	nativeStackObject = nil;
}
_nativeStack = nativeStackObject;
```

vijaysingh-axway: In your code you are copying nativeStackObject  in same variable but not releasing. You are retaining nativeStackObject, but not releasing.

_nativeStack is not retaining or copying  any object but it is getting released in dealloc. 
The code which I have suggested, it is also copying (owns memory) the nativeStackObject in _nativeStack.
Other way to do it is  -
`id nativeStackObject = [dictionary objectForKey:@"nativeStack"];
if ([nativeStackObject isKindOfClass:[NSArray class]]) {
	_nativeStack = [nativeStackObject copy];
} else if ([nativeStackObject isKindOfClass:[NSString class]]) {
	_nativeStack = [[nativeStackObject componentsSeparatedByString:@"\n"] retain];
} else {
	_nativeStack = nil;
}
//_nativeStack = nativeStackObject;`

jquick-axway: Updated this PR based on your suggestion.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -156,8 +156,16 @@ - (id)initWithDictionary:(NSDictionary *)dictionary
     if (_backtrace == nil) {
       _backtrace = [[[dictionary objectForKey:@"stack"] description] copy];
     }
-    _nativeStack = [[dictionary objectForKey:@"nativeStack"] copy];
     _dictionaryValue = [dictionary copy];
+
+    id nativeStackObject = [dictionary objectForKey:@"nativeStack"];
+    if ([nativeStackObject isKindOfClass:[NSArray class]]) {
```
garymathews, line 162: We should validate if `nativeStackObject` is `nil` since `isKindOfClass` would fail if it is

garymathews (changes requested):

lokeshchdhry:
FR Passed.

PR #12722, opened 2021-04-14
Clean up Android's HTTPClient proxy

sgtcoolguy:
**Description:**
`Ti.network.HTTPClient` had some methods documented as removed in 3.4.0 but whose implementation as never removed. This yanks them from the apidocs as well as removes the actual implementations.


jquick-axway (approved):
CR: Pass

PR #12724, opened 2021-04-14
fix(ios): expose KrollBridge modules dictionary

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28415

**Description:**
This is purely for native module backwards compatibility. Hans had a drag and drop module that relied in accessing this dictionary to get it's own singleton instance. This patch "fixes" this to still work. That said, I'm not sure if we want to expose this to module developers.

Also note that I cherry-picked Josh's fix from #12721 so that I could se the true underlying errors.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
@@ -34,6 +34,26 @@ CFMutableSetRef krollBridgeRegistry = nil;
 
 @implementation KrollBridge
 
+- (NSDictionary<NSString *, id> *)modules
+{
+  // This is for backwards compatibility with native modules relying on a modules property that
+  // cached module ids -> module instances
+  // Note that this one will only return external native modules and relies on their moduleId
+  // method returning the correct valiue
+  NSDictionary<NSString *, id> *orig = [self.host valueForKey:@"modules"];
+  NSMutableDictionary<NSString *, id> *copy = [orig mutableCopy];
+  for (NSString *key in orig) {
+    id module = [orig objectForKey:key];
+    if ([module respondsToSelector:@selector(moduleId)]) {
+      NSString *moduleId = [module moduleId];
+      if (moduleId) {
+        copy[moduleId] = module;
+      }
+    }
+  }
+  return copy;
+}
+
 + (void)initialize
 {
   if (krollBridgeRegistry == nil) {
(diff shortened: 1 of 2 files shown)


iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m
```diff
@@ -156,8 +156,19 @@ - (id)initWithDictionary:(NSDictionary *)dictionary
     if (_backtrace == nil) {
       _backtrace = [[[dictionary objectForKey:@"stack"] description] copy];
     }
-    _nativeStack = [[dictionary objectForKey:@"nativeStack"] copy];
     _dictionaryValue = [dictionary copy];
+
+    id nativeStackObject = [dictionary objectForKey:@"nativeStack"];
```
vijaysingh-axway, line 161: There are some review comment in PR #12721 for this .

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -34,6 +34,26 @@
...
 @implementation KrollBridge
 
+- (NSDictionary<NSString *, id> *)modules
+{
+  // This is for backwards compatibility with native modules relying on a modules property that
+  // cached module ids -> module instances
+  // Note that this one will only return external native modules and relies on their moduleId
+  // method returning the correct valiue
+  NSDictionary<NSString *, id> *orig = [self.host valueForKey:@"modules"];
+  NSMutableDictionary<NSString *, id> *copy = [orig mutableCopy];
```
jquick-axway, line 44: I think you need to do the following here or else a memory leak will occur.
```objc
[[orig mutableCopy] autorelease];
```

vijaysingh-axway: Instead of autorelease here, generally we should do before while return. Line 54 should be  -
`return [copy autorelease]`

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -34,6 +34,26 @@
 
 @implementation KrollBridge
 
+- (NSDictionary<NSString *, id> *)modules
+{
+  // This is for backwards compatibility with native modules relying on a modules property that
+  // cached module ids -> module instances
+  // Note that this one will only return external native modules and relies on their moduleId
+  // method returning the correct valiue
```
hansemannn, line 42: Small typo in `valiue`. Also, is the `moduleId` the real ID again here, like `ti.facebook` instead of `TiFacebookModule`?

sgtcoolguy: it's whatever the module itself reports from it's `moduleId` method, which should be pre-populated with the correct identifier when the module is created via template.

hansemannn:
Can this be merged?

jquick-axway:
Closing in favor of PR #13161 which implements review feedback and fixes conflicts.

PR #12725, opened 2021-04-15
fix(ios): report script source urls using full file: uris

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28422

**Description:**
When we evaluate JS files we tell the JS engine the source URL for the script. We used to report `file://` with the full path on device (or to the simulator app files). SDK 10.0.0 changed the internals of require and we ended up reporting invalid URIs with the `file://` scheme but whose path was actually relative to the app root. That was intentional internally for `require`, but *not* for the debugger as it causes breakage in Studio. Specifically, Studio uses a regexp to set breakpoints for scripts in many cases and used the `.app/` prefix to avoid the regex matching files deeper in the app source file tree.

So to give some examples, in SDK 9.3.2 we'd report:
- app.js -> file:///Users/cwilliams/Library/Developer/CoreSimulator/Devices/4753950A-0B8D-4EA6-8BD5-5FCB01760D81/data/Containers/Bundle/Application/02857640-4389-400A-8CBE-4F10946EC376/ti-example.app/app.js
- ti.main.js -> file:///Users/cwilliams/Library/Developer/CoreSimulator/Devices/4753950A-0B8D-4EA6-8BD5-5FCB01760D81/data/Containers/Bundle/Application/02857640-4389-400A-8CBE-4F10946EC376/ti-example.app/ti.main.js

Prior to this PR, we'd report:
- app.js -> file:///app.js
- ti.kernel.js -> file:///Users/cwilliams/Library/Developer/CoreSimulator/Devices/4753950A-0B8D-4EA6-8BD5-5FCB01760D81/data/Containers/Bundle/Application/02857640-4389-400A-8CBE-4F10946EC376/ti-example.app/ti.kernel.js
- ti.main.js -> file:///ti.main.js
 
(Note that ti.kernel.js was using the right path because it's the first file we run and doesn't go through the new require code that aligns with Android)


ewanharris (approved):
Changes look good to me, Studio and VS Code can both now hit breakpoints

lokeshchdhry:
FR Passed.
Debugger now hits the breakpoint as expected on IOS.

PR #12728, opened 2021-04-16
fix(android): account for platform path separators

garymathews:
- Account for Windows file path separators when processing assets

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28423)

### DIFF
--- a/cli/lib/gather.js
+++ b/cli/lib/gather.js
@@ -9,10 +9,10 @@ const FILENAME_REGEXP = /^(.*)\.(\w+)$/;
 // iOS specific stuff
 const LAUNCH_IMAGE_REGEXP = /^(Default(-(Landscape|Portrait))?(-[0-9]+h)?(@[2-9]x)?)\.png$/;
 const LAUNCH_LOGO_REGEXP = /^LaunchLogo(@[23]x)?(~(iphone|ipad))?\.(png|jpg)$/;
-const BUNDLE_FILE_REGEXP = /.+\.bundle\/.+/;
+const BUNDLE_FILE_REGEXP = /.+\.bundle[/|\\].+/;
 // Android-specific stuff
-const DRAWABLE_REGEXP = /^images\/(high|medium|low|res-[^/]+)(\/(.*))$/;
-const ANDROID_SPLASH_REGEXP = /^(images\/(high|medium|low|res-[^/]+)\/)?default\.(9\.png|png|jpg)$/;
+const DRAWABLE_REGEXP = /^images[/|\\](high|medium|low|res-[^/]+)([/|\\](.*))$/;
+const ANDROID_SPLASH_REGEXP = /^(images[/|\\](high|medium|low|res-[^/]+)[/|\\])?default\.(9\.png|png|jpg)$/;
 
 /**
  * Merges multiple maps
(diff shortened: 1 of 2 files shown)


jquick-axway:
@sgtcoolguy, would you mind checking this out? This PR changes the file's relative path (used as a key in a dictionary) to use the platform's path separator (back slash on Windows). It no longer normalizes all paths to use forward slashes. Will there be any repercussions to doing this anywhere else?

cli/lib/gather.js
```diff
@@ -197,7 +197,7 @@ class Walker {
 			return;
 		}
 		const info = new FileInfo(name, from, to);
-		const relPath = from.replace((origSrc || src) + path.sep, prefix ? prefix + path.sep : '').replace(/\\/g, '/');
+		const relPath = from.replace((origSrc || src) + path.sep, prefix ? prefix + path.sep : '').replace(/\\/g, path.sep);
```
sgtcoolguy, line 200: Ah, I see the issue. Ok, so I had assumed this would normalize paths to unix style path separators - but the later calls when we walk directories ends up using `path.join()` which uses the OS-path separator. The `path`/`fs` APIs will still handle using either OK, but the regex matching obviously wouldn't work as expected (and there's potential for duplicated paths ending up in the map/sets as separate entries even though they meant to point to the same file but may have mismatching separators.)

sgtcoolguy: So the downstream effects here would likely fall to any of the "tasks" that end up processing the files in these sets/groups of files. So the places to look would be:
- https://github.com/appcelerator/titanium_mobile/tree/master/cli/lib/tasks
- https://github.com/appcelerator/titanium_mobile/blob/master/android/cli/lib/process-drawables-task.js
- https://github.com/appcelerator/titanium_mobile/blob/master/android/cli/lib/process-splashes-task.js

Of those, I think this is the only line that needs to be changed (though technically it's iOS-only so `path.sep` should always be `'/'` on the OS): https://github.com/appcelerator/titanium_mobile/blob/master/cli/lib/tasks/process-js-task.js#L268

The other thing to double-check/confirm is that the underlying code in `apps-tasks` will end up recording hr files using the proper `path.sep` so that the caching/keys it used matched what we used. I think we are the ones who report the directories/filepaths to it, so as long as we're consistent I think we're fine.

sgtcoolguy (changes requested):
One extra line to modify as pointed in my comments (though again technically not necessary as we only support building for iOS on macOS where `path.sep === '/'`

sgtcoolguy (approved):

ssjsamir (approved):
FR Passed, tested using the test steps mentioned in https://jira.appcelerator.org/browse/TIMOB-28423.

**Test environment**
```
Operating System
  Name                        = Microsoft Windows 10 Pro
  Version                     = 10.0.18363
  # CPUs                      = 16
  Memory                      = 63.9GB

Node.js
  Node.js Version             = 12.16.1
  npm Version                 = 6.13.4

Appcelerator CLI
  Installer                   = 6.0.0-1
  Core Package                = 9.0.0-9

Titanium CLI
  CLI Version                 = 5.3.0
  node-appc Version           = 1.1.2
```

PR #12731, opened 2021-04-19
fix(android): support detail property in TableViewRow click event

garymathews:
- Fully support `hasDetail` property

##### TEST CASE
- Clicking row should display `detail: false`
- Clicking detail chevron should display `detail: true`
```JS
	data: [
		Ti.UI.createTableViewRow({
			title: 'hasDetail: false',
			hasDetail: false
		}),
		Ti.UI.createTableViewRow({
			title: 'hasDetail: true',
			hasDetail: true
		})
	]
});

table.addEventListener('click', e => {
	alert(`detail: ${!!e.detail}`);
});

win.add(table);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28199)


jquick-axway:
On iOS, the hasDetail shows an `(i)` info icon.

On Android, there is a built-in `android.R.drawable.ic_menu_info_details` info icon drawable just like this.
However, maybe it would be better to show "3 vertical dots" instead? This is what Google typically uses in their Files, Docs, and Drives apps. I think this is already available via `android.R.drawable.ic_menu_more` but I could be wrong. _(I'm pretty sure Android Studio's vector asset wizard can generate it for us too.)_

Unrelated: I noticed setting "hasChild" to `true` shows a right arrow chevron on iOS, but a down arrow chevron on Android. We may want to change Android to be consistent with iOS and get rid of the down chevron.

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

PR #12744, opened 2021-04-23
refactor: move build code to esm from commonjs

sgtcoolguy:
Related to #12739


android/titanium/libv8-services.mjs
```diff
@@ -129,7 +131,7 @@ async function generateSnapshot(v8SnapshotHeaderFilePath, rollupFileContent) {
 	}
 
 	// Generate 'V8Snapshots.h' from template
-	const template = await util.promisify(ejs.renderFile)('V8Snapshots.h.ejs', { blobs }, {});
+	const template = await promisify(ejs.renderFile)('V8Snapshots.h.ejs', { blobs }, {});
```
ewanharris, line 134: Nit, we might be able to remove the promisify here, it appears `ejs.renderFile` can return a promise since [2.5.8](https://github.com/mde/ejs/blob/main/CHANGELOG.md#v258-2018-03-25)

build/scons.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/opt/homebrew/bin/node
```
ewanharris, line 1: I think this should stay unchanged?
```suggestion
#!/usr/bin/env node
```

android/titanium/prebuild.mjs
```diff
@@ -183,16 +185,14 @@ async function generateSourceCode() {
...
 
-if (require.main === module) {
-	main()
-		.then(() => process.exit(0))
-		.catch((err) => {
-			console.error(err);
-			process.exit(1);
-		});
-}
+main()
```
ewanharris, line 193: These scripts will all be so much nicer when we can use top level await 😅

ewanharris (commented):
Only thing would be the shebang I think. What're the steps to landing this? Move all of the dev build scripts over to esm?

build/lib/builder.mjs
```diff
@@ -1,17 +1,20 @@
...
-
-const git = require('./git');
-const utils = require('./utils');
-const Packager = require('./packager');
-
+import os from 'os';
+import path from 'path';
+import fs from 'fs-extra';
+import { rollup } from 'rollup';
+import babel from '@rollup/plugin-babel';
```
ewanharris, line 5: Oops, shoulda checked CI, this should be as below 
```suggestion
import { babel } from '@rollup/plugin-babel';
```

ewanharris:
Marked as ready for review accidentally in the app.

@hansemannn this is the scons build code which is less problematic imo, I remember reviewing this and what as changed worked ok. There was just no real path determined to getting it merged and it's not a real high priority either

cb1kenobi:
I approve this.

With that said, I don't love Node.js's .mjs extension and once we get the Titanium CLI moved to ES modules, we can probably set the package type to "module" and no longer need the .mjs extension. So, it's a race between the merge conflicts getting fixed in this PR (along with the nitpicks) and the Titanium CLI being massively refactored. Go!

hansemannn:
Okay, then let's fix the merge conflicts and go with it. I think the CLI should rather move to TypeScript instead of modular JS, but I'm not in the position to dictate that. From what I see in our projects, it's all TS and it's superb.

cb1kenobi:
Closing as we've already converted the SDK to ESM.

PR #12752, opened 2021-04-27
fix(ios): fixed large title rendering issue

vijaysingh-axway:
https://jira.appcelerator.org/browse/TIMOB-28428


jquick-axway (approved):
LGTM

Works on iOS 14.5.

However, it crashes on iOS 12.4 when you navigate to the child window, but it looks like the crash is unrelated to this PR's code change.

ewanharris (approved):
FR Pass.

The issue mentioned by Josh does appear to be limited to iOS 12, tested iOS 13 and did not see the issue

jquick-axway:
Regarding the iOS 12 crash, I've written up a ticket here: [TIMOB-28484](https://jira.appcelerator.org/browse/TIMOB-28484)

PR #12755, opened 2021-04-28
fix(ios): xcode project change detection causing unnecessary rebuilds

janvennemann:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28506

**Summary:**
This fixes two issues where our build was unnecessarily doing full Xcode re-builds...

- Always integrate frameworks into into the Xcode project in the same order. Otherwise the Xcode project file is different from the last run. Switched from iterating a map to a sorted array to ensure order. Also fixed an issue with loading previous state data which caused the framework hook to always scan for available frameworks inside the project.
- Trim whitespaces when comparing generated `.entitlements` with existing ones from Xcode extensions before overwriting. Otherwise they may differ just by a trailing newline for example and always cause a full re-build.

**Test:**
1. Build [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) for iOS.
2. Clear the log.
3. Build the project for iOS again without making any changes.
4. Verify that you do **NOT** see the following in the log. _(This is the fix.)_
```
Forcing rebuild: Xcode project has changed since last build
```


hansemannn:
Really cool fix, it works very well! Hope this can be cherry picked into GA

hansemannn:
@jquick-axway May it be possible to include this in the 10.0.1 release? It improves the developer experience for larger projects massively.

iphone/cli/hooks/frameworks.js
```diff
@@ -397,12 +399,10 @@ class InspectFrameworksTask extends IncrementalFileTask {
...
-				this._frameworks.set(frameworkMetadata.name, new FrameworkInfo(
-					frameworkMetadata.name,
-					frameworkMetadata.path,
-					frameworkMetadata.type,
-					new Set(frameworkMetadata.architectures)
-				));
+				this._frameworks.set(frameworkMetadata.name, new FrameworkInfo({
+					...frameworkMetadata,
+					architectures: new Set(frameworkMetadata.architectures)
+				}));
```
jquick-axway, line 405: I'm not sure how to fix this eslint warning.
(If we merge this as-is, then we'll get this warning in all of our PRs.)

It's a false positive and we correctly set the node version via the "package.json" as shown below. For some reason eslint is ignoring it.
https://github.com/appcelerator/titanium_mobile/blob/master/package.json#L163-L165

We can work-around it by adding the following rule to our [.eslintrc](https://github.com/appcelerator/titanium_mobile/blob/master/.eslintrc) file, but it would be a shame to duplicate our min node version in our [package.json](https://github.com/appcelerator/titanium_mobile/blob/master/package.json) here. Especially since the eslint documentation for this rule [here](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features/node-builtins.md) says we shouldn't have to do this. Hmm...
```json
"rules": {
	"node/no-unsupported-features/node-builtins": ["error", {
		"version": ">=12.13.0",
		"ignores": []
	}]
},
```

hansemannn: You can use Object.assign which should do the same as the spread operator

jquick-axway: That's true.

I guess I'm more concerned why eslint is ignoring the node version specified in the "package.json", because according to the docs it should work. Even VS Code gives me the same warning.

jquick-axway (approved):
CR: Pass

iphone/cli/hooks/frameworks.js
ewanharris: I think the issue is down to having multiple `package.json` files in the repo, it looks like the plugin walks backwards from the file with the warning/error to lookup the engines.node property, as that isn't in `iphone/package.json` it's defaulting to >=8.0.0. I've just created #12965 to fix this

jquick-axway: @ewanharris , that makes sense. Why didn't I think of that. :stuck_out_tongue:
For now, I worked-around it by implementing it differently.

ewanharris (approved):
FR Pass. No longer seeing unnecessary rebuilds on kitchensink

PR #12769, opened 2021-05-05
feat(android): implement native selection for ListView and TableView

garymathews:
- Implement ability to select rows/items
- Implement `allowsSelectionDuringEditing` and `allowsMultipleSelectionDuringEditing` properties for both TableView and ListView
- Implement `rowsselected` and `itemsselected` event
- Move both `Ti.UI.iOS.ListViewCellSelectionStyle` and `Ti.UI.iOS.TableViewCellSelectionStyle` to `Ti.UI.SELECTION_STYLE_*`

##### TEST CASE 1
```JS
const btn_edit = Ti.UI.createButton({ title: 'EDITING: false', bottom: 5, left: 5 });
const btn_check = Ti.UI.createButton({ title: 'CHECK: true', bottom: 5, right: 5 });
const btn_scroll = Ti.UI.createButton({ title: 'SCROLL: false', bottom: 50, right: 5 });

function createSection(rows, suffix) {
		headerTitle: `Section ${suffix}`
	});
	const items = [];

	for (let i = 1; i <= rows; i++) {
		items.push({
			properties: {
				title: `Row #${i}`,
				selectionStyle: Ti.UI.SELECTION_STYLE_DEFAULT
			}
		});
	}
	section.items = items;

	return section;
}

	sections: [
		createSection(10, 'A'),
		createSection(10, 'B'),
		createSection(10, 'C')
	],
	editing: false,
	allowsSelectionDuringEditing: true,
	allowsMultipleSelectionDuringEditing: true
});

listView.addEventListener('itemsselected', e => {
	console.log(`itemsselected: ${JSON.stringify(Object.keys(e), null, 1)}`);
});

btn_edit.addEventListener('click', e => {
	listView.editing = !listView.editing;
	btn_edit.title = `EDITING: ${listView.editing ? 'true' : 'false'}`;
});

btn_check.addEventListener('click', e => {
	listView.showSelectionCheck = !listView.showSelectionCheck;
	btn_check.title = `CHECK: ${listView.showSelectionCheck ? 'true' : 'false'}`;
});

let scrollInterval = null;
btn_scroll.addEventListener('click', e => {
	if (scrollInterval) {
		clearInterval(scrollInterval);
		scrollInterval = null;
	} else {
		scrollInterval = setInterval(_ => {
			const sectionIndex =  Math.floor(Math.random() * 3);
			const itemIndex =  Math.floor(Math.random() * 10);

			console.log(`scrollToItem: section -> ${sectionIndex}, item -> ${itemIndex}`);
			listView.scrollToItem(sectionIndex, itemIndex);
		}, 2000);
	}
	btn_scroll.title = `SCROLL: ${scrollInterval ? 'true' : 'false'}`;
});

win.add([ listView, btn_edit, btn_check, btn_scroll ]);
win.open();
```
##### TEST CASE 2
```JS
const btn_edit = Ti.UI.createButton({ title: 'EDITING: false', bottom: 5, left: 5 });
const btn_check = Ti.UI.createButton({ title: 'CHECK: true', bottom: 5, right: 5 });
const btn_scroll = Ti.UI.createButton({ title: 'SCROLL: false', bottom: 50, right: 5 });

function createSection(rows, suffix) {
	const section = Ti.UI.createTableViewSection({
		headerTitle: `Section ${suffix}`
	});

	for (let i = 1; i <= rows; i++) {
		section.add(Ti.UI.createTableViewRow({
			title: `Row #${i}`,
			selectionStyle: Ti.UI.SELECTION_STYLE_DEFAULT
		}));
	}

	return section;
}

	data: [
		createSection(10, 'A'),
		createSection(10, 'B'),
		createSection(10, 'C')
	],
	editing: false,
	allowsSelectionDuringEditing: true,
	allowsMultipleSelectionDuringEditing: true
});

tableView.addEventListener('rowsselected', e => {
	console.log(`rowsselected: ${JSON.stringify(Object.keys(e), null, 1)}`);
});

btn_edit.addEventListener('click', e => {
	tableView.editing = !tableView.editing;
	btn_edit.title = `EDITING: ${tableView.editing ? 'true' : 'false'}`;
});

btn_check.addEventListener('click', e => {
	tableView.showSelectionCheck = !tableView.showSelectionCheck;
	btn_check.title = `CHECK: ${tableView.showSelectionCheck ? 'true' : 'false'}`;
});

let scrollInterval = null;
btn_scroll.addEventListener('click', e => {
	if (scrollInterval) {
		clearInterval(scrollInterval);
		scrollInterval = null;
	} else {
		scrollInterval = setInterval(_ => {
			const sectionIndex =  Math.floor(Math.random() * 3) * 10;
			const itemIndex =  Math.floor(Math.random() * 10);
			const index = sectionIndex + itemIndex;

			console.log(`scrollToIndex: index -> ${index}`);
			tableView.scrollToIndex(index);
		}, 2000);
	}
	btn_scroll.title = `SCROLL: ${scrollInterval ? 'true' : 'false'}`;
});

win.add([ tableView, btn_edit, btn_check, btn_scroll ]);
win.open();
```

##### TEST CASE 3
- See https://github.com/appcelerator/titanium_mobile/pull/12298

[TIMOB-28327](https://jira.appcelerator.org/browse/TIMOB-28327)
[TIMOB-28435](https://jira.appcelerator.org/browse/TIMOB-28435)
[TIMOB-28436](https://jira.appcelerator.org/browse/TIMOB-28436)
[TIMOB-28533](https://jira.appcelerator.org/browse/TIMOB-28533)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -379,11 +381,21 @@ private void processProperty(String name, Object value)
...
+				parent.add(this);
+			}
+
+		} else if (name.equals(TiC.PROPERTY_SHOW_SELECTION_CHECK)) {
 
-			// Update list.
+			// Update and refresh list.
 			update();
 		}
 	}
```
jquick-axway, line 401: I think we "need" to add support for the [selectedItems](https://titaniumsdk.com/api/titanium/ui/listview.html#selecteditems) property so that the app can fetch the currently selected items without an event listener.
```java
@Kroll.getProperty
public KrollDict[] getSelectedItems()
{
	// Return same array as 'itemsselected' event.
}
```

We have an iOS unit-test for the above, but we enable it for Android until we also add support for the [selectItems()](https://titaniumsdk.com/api/titanium/ui/listview.html#selectitem) method... and if do that then we should also support the [deselectItems()](https://titaniumsdk.com/api/titanium/ui/listview.html#deselectitem). _(sorry)_
https://github.com/appcelerator/titanium_mobile/blob/master/tests/Resources/ti.ui.listview.test.js#L239

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewHolder.java
```diff
@@ -202,8 +211,8 @@ public void bind(final ListItemProxy proxy, final boolean selected)
 					}
 
 					// Support selected backgrounds.
-					nativeView.setBackground(generateSelectedDrawable(properties, backgroundDrawable));
-					borderView.setActivated(selected);
+					this.container.setBackground(generateSelectedDrawable(properties, backgroundDrawable));
+					this.container.setActivated(selected);
```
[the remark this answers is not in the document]
garymathews: That's determined [here](https://github.com/appcelerator/titanium_mobile/pull/12769/files#diff-251408f37271ef845002029183bc7f651843fa09b81069fbf9758063509fcabaR207) and passed through the `selected` parameter

iphone/Classes/UIModule.h
```diff
@@ -157,6 +157,10 @@
 @property (nonatomic, readonly) NSNumber *BUTTON_STYLE_OPTION_NEGATIVE;
 @property (nonatomic, readonly) NSNumber *BUTTON_STYLE_OPTION_NEUTRAL;
 
+@property (nonatomic, readonly) NSNumber *SELECTION_STYLE_NONE;
+@property (nonatomic, readonly) NSNumber *SELECTION_STYLE_BLUE;
+@property (nonatomic, readonly) NSNumber *SELECTION_STYLE_GRAY;
+
```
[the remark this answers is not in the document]
garymathews: Oops, good catch. Thought I updated these.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -202,28 +203,66 @@ public Object getSelectionKey()
...
+								}
+							}
+
+							if (selectedItems.size() > 0) {
+								final KrollDict data = new KrollDict();
+
+								data.put(TiC.PROPERTY_SELECTED_ITEMS, selectedItems.toArray(new KrollDict[0]));
+								data.put(TiC.PROPERTY_STARTING_ITEM, selectedItems.get(0));
+								proxy.fireEvent(TiC.EVENT_ITEMS_SELECTED, data);
+							}
```
jquick-axway, line 253: We need to fire an "itemsselected" event when you uncheck all items too, where the selected items count is zero. So, we need to get rid of the `if (selectedItems.size() > 0) {}` part and do the following instead.
```java
final KrollDict data = new KrollDict();
data.put(TiC.PROPERTY_SELECTED_ITEMS, selectedItems.toArray(new KrollDict[0]));
data.put(TiC.PROPERTY_STARTING_ITEM, selectedItems.isEmpty() ? null : selectedItems.get(0));
proxy.fireEvent(TiC.EVENT_ITEMS_SELECTED, data);
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -202,28 +203,66 @@ public Object getSelectionKey()
 				}
 			},
 			StorageStrategy.createLongStorage()
-		)
-			.withSelectionPredicate(SelectionPredicates.<Long>createSelectSingleAnything())
-			.build();
-		this.tracker.addObserver(new SelectionTracker.SelectionObserver()
-		{
```
jquick-axway, line 209: When you set "editing" to `true` but none of the rows are checked, then the end-user has to long-press a row to check a row's checkbox. Once at least 1 row is checked, the user is then able to tap each row to check them. This is a confusing interface. Especially since unchecking all rows means you have to long-press a row to enable checking them again.

There is an easy solution to this. Add the following method to where we create the `ItemDetails` class up above.
```java
return new ItemDetails() {
	// ... your existing methods ...

	@Override
	public boolean inSelectionHotspot(@NonNull MotionEvent e)
	{
		// Returning true allows taps to immediately select this row.
		return true;
	}
};
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -379,11 +402,21 @@ private void processProperty(String name, Object value)
...
```
jquick-axway, line 422: Calling `selectItem()` will throw a `NullPointerException` if property "editing" is set `false`.
_(Does not crash on iOS.)_

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -202,28 +203,66 @@ public Object getSelectionKey()
...
+				this.tracker.addObserver(new SelectionTracker.SelectionObserver()
+				{
+					int selectionCount = 0;
+
+					@Override
+					public void onSelectionChanged()
+					{
+						super.onSelectionChanged();
+
+						if (tracker.hasSelection() && selectionCount != tracker.getSelection().size()) {
```
jquick-axway, line 228: This if-condition's `selectionCount != tracker.getSelection().size()` part prevents single-selection mode from firing events correctly. Only the 1st single selection event fires, but subsequent events do not fire when selecting different rows since the selection count is always 1.

Also, iOS does not fire an "itemsselected" event when in single selection mode. iOS only fires the "itemclick" event.

jquick-axway:
@garymathews, do you think we should add support for ListView's [allowsMultipleSelectionInteraction](https://titaniumsdk.com/api/titanium/ui/listview.html#allowsmultipleselectioninteraction) property? When you set this to `true` on iOS, the ListView will automatically set "editing" to `true` when the user does a particular gesture. On Android, we could automatically enable editing mode when the user long-presses a row and check/select it.

What do you think?

I don't mean to add feature creep, but I'm thinking almost everyone will want to activate editing mode via long-presses since that's the Android standard way of doing it.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -216,34 +217,78 @@ public Object getSelectionKey()
...
+									selectedItems.add(selectedItem);
+								}
+							}
+						}
+
+						final KrollDict data = new KrollDict();
+
+						data.put(TiC.PROPERTY_SELECTED_ITEMS, selectedItems.toArray(new KrollDict[0]));
+						data.put(TiC.PROPERTY_STARTING_ITEM, selectedItems.isEmpty() ? null : selectedItems.get(0));
+						proxy.fireEvent(TiC.EVENT_ITEMS_SELECTED, data);
```
jquick-axway, line 286: On iOS, when in single selection mode, only the "itemclick" event is fired. The "itemsselected" event is only fired when in multi-selection mode.
```java
if (allowsMultipleSelection) {
	final KrollDict data = new KrollDict();
	data.put(TiC.PROPERTY_SELECTED_ITEMS, selectedItems.toArray(new KrollDict[0]));
	data.put(TiC.PROPERTY_STARTING_ITEM, selectedItems.isEmpty() ? null : selectedItems.get(0));
	proxy.fireEvent(TiC.EVENT_ITEMS_SELECTED, data);
} else if (!selectedItems.isEmpty()) {
	selectedItems.get(0).fireSyncEvent(TiC.EVENT_CLICK, null);
}
```
_(I don't know if I'm handling the `EVENT_CLICK` line of code right so you might want to double check.)_

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -216,34 +217,78 @@ public Object getSelectionKey()
...
-						final ListItemProxy proxy = i.next();
+						if (tracker.hasSelection()) {
+							final Iterator<ListItemProxy> i = tracker.getSelection().iterator();
 
-						Log.d(TAG, "SELECTED: " + proxy.getProperties().getString(TiC.PROPERTY_TITLE));
+							while (i.hasNext()) {
+								final ListItemProxy item = i.next();
+								final KrollDict selectedItem = new KrollDict();
+								final ListSectionProxy section =
+									item.getParent() != null ? (ListSectionProxy) item.getParent() : null;
```
jquick-axway, line 270: If you create a section with no rows, then an empty row will appear under the section with a checkbox. If you tap on the checkbox, then the above line of code will throw an exception because `item.getParent()` can't be cast to `(ListSectionProxy)`.

You can test this with your test code by changing `createSection(10, 'A')` to `createSection(0, 'A')`.

We shouldn't show a checkbox in this case of course, but we also want to make sure the empty section header isn't selectable as well. Maybe we return out here? Or perhaps we'll have to override the [SelectionPredicate](https://developer.android.com/reference/kotlin/androidx/recyclerview/selection/SelectionTracker.SelectionPredicate) class' `canSetStateAtPosition()` and `canSetStateForKey()` methods and have them return `false` for empty section headers? I'm not sure what the best solution to this is.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewHolder.java
```diff
@@ -113,6 +116,20 @@ public void bind(final ListItemProxy proxy, final boolean selected)
...
+			&& listViewProperties.optBoolean(TiC.PROPERTY_EDITING, false)
+			&& listViewProperties.optBoolean(TiC.PROPERTY_ALLOWS_SELECTION_DURING_EDITING, false)
+			&& listViewProperties.optBoolean(TiC.PROPERTY_ALLOWS_MULTIPLE_SELECTION_DURING_EDITING, false)) {
+
+			if (selected) {
+				this.leftImage.setImageDrawable(checkcircleDrawable);
+			} else {
+				this.leftImage.setImageDrawable(circleDrawable);
+			}
+			this.leftImage.setVisibility(View.VISIBLE);
```
jquick-axway, line 130: If the section has no items, then this code will show a checkbox in an empty row under the section header.
You can test this with your test code by changing `createSection(10, 'A')` to `createSection(0, 'A')`.

So, maybe we should only set up the `leftImage` if `section.getItemCount() > 0`?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -375,6 +379,69 @@ public void setActivity(Activity activity)
...
+
+		if (contentOffset != null) {
+			this.contentOffset = contentOffset;
+
+			if (listView != null) {
+
+				if (contentOffset.containsKeyAndNotNull(TiC.PROPERTY_INDEX)) {
+
+					// If available, scroll to provided index provided by internal `getContentOffset()` method.
+					listView.getRecyclerView().scrollToPosition(contentOffset.getInt(TiC.PROPERTY_INDEX));
```
jquick-axway, line 422: There is a bug here. This code always assumes the index belongs to the 1st section. _(The `getContentOffset()` does not offset the stored index with the total number of section items above it.)_

In your test code, scroll to the very bottom of the ListView. Note that the top-most visible row should be in `Section B`. Now switch between dark/light mode (or enable/disable editing) and you'll notice that the ListView scrolls to `Section A` equivalent of that row item number.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
```diff
@@ -216,34 +223,92 @@ public Object getSelectionKey()
...
+							while (i.hasNext()) {
+								final ListItemProxy item = i.next();
+
+								if (item.isPlaceholder()) {
+									continue;
+								}
+								if (!allowsMultipleSelection) {
+									item.fireEvent(TiC.EVENT_CLICK, null);
+									return;
+								}
```
jquick-axway, line 285: On iOS, when in single-selection mode, the "selectedItems" property is still returns an array of the 1 selected item. On Android, it's an empty array.

Instead of returning here, I think we should add an else-if block where we fire the "itemsselected" event.
```java
if (allowsMultipleSelection) {
	final KrollDict data = new KrollDict();
	data.put(TiC.PROPERTY_SELECTED_ITEMS, selectedItems.toArray(new KrollDict[0]));
	data.put(TiC.PROPERTY_STARTING_ITEM, selectedItems.isEmpty() ? null : selectedItems.get(0));
	proxy.fireEvent(TiC.EVENT_ITEMS_SELECTED, data);
} else if (!selectedItems.isEmpty()) {
	selectedItems.get(0).fireEvent(TiC.EVENT_CLICK, null);
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiRecyclerViewHolder.java
```diff
@@ -139,25 +182,26 @@ protected Drawable generateRippleDrawable(Drawable drawable, String color)
...
-			stateDrawable.addState(new int[] {}, drawable);
-
-			return stateDrawable;
+			stateDrawable.addState(new int[] { android.R.attr.state_activated }, selectedBackgroundDrawable);
+		} else {
+			stateDrawable.addState(new int[] { android.R.attr.state_activated }, new ColorDrawable(COLOR_SELECTED));
 		}
 
-		return drawable;
+		stateDrawable.addState(new int[] {}, drawable);
```
jquick-axway, line 202: Calling `stateDrawable.addState()` causes an exception to be thrown on Android 6.0 and older.
```
[ERROR] TiExceptionHandler: (main) [198,325] Attempt to invoke virtual method 'android.graphics.drawable.shapes.Shape android.graphics.drawable.shapes.Shape.clone()' on a null object reference
[ERROR] TiExceptionHandler:
[ERROR] TiExceptionHandler:     android.graphics.drawable.ShapeDrawable.mutate(ShapeDrawable.java:501)
[ERROR] TiExceptionHandler:     android.graphics.drawable.LayerDrawable.mutate(LayerDrawable.java:936)
[ERROR] TiExceptionHandler:     android.graphics.drawable.RippleDrawable.mutate(RippleDrawable.java:891)
[ERROR] TiExceptionHandler:     android.graphics.drawable.DrawableContainer.selectDrawable(DrawableContainer.java:443)
[ERROR] TiExceptionHandler:     android.graphics.drawable.StateListDrawable.onStateChange(StateListDrawable.java:111)
[ERROR] TiExceptionHandler:     android.graphics.drawable.StateListDrawable.addState(StateListDrawable.java:94)
[ERROR] TiExceptionHandler:     ti.modules.titanium.ui.widget.listview.TiRecyclerViewHolder.generateSelectedDrawable(TiRecyclerViewHolder.java:202)
```

jquick-axway: It's easy to reproduce. You just have to create a simple ListView and the default drawable the above method creates will cause this exception to be thrown on older Android OS versions.
```javascript
function createListViewSectionItems() {
	const items = [];
	for (let index = 1; index <= 500; index++) {
		items.push({ properties: { title: "Row " + index } });
	}
	return items;
}
window.add(Ti.UI.createListView({
		headerTitle: "ListView",
		items: createListViewSectionItems(),
	})],
}));
window.open();
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -661,7 +770,11 @@ public void selectItem(int sectionIndex, int itemIndex)
 				final ListItemProxy item = section.getListItemAt(itemIndex);
```
jquick-axway, line 770: If I set up a ListView to enable editing mode via a "longpress" event (which is how most Android app do it), you would normally want to check the item you long-pressed and you would do this via our `selectItem()` method (see my "longpress" event listener below). The issue with this is the 1st line in this Java `selectItem()` method calls `scrollToItem()`, which we should do if the item is offscreen... but... we don't want the ListView to suddenly scroll the ListView if the item is already onscreen, such as for our long press use-case.
_(Note that iOS will scroll to the selected item if offscreen, so, that part is okay.)_

Can you use the `getFirstVisibleItem()` and `getLastVisibleItem()` methods please and only call `scrollToItem()` if offscreen?

```javascript
function createSection(rows, suffix) {
	const items = [];
	for (let index = 1; index <= rows; index++) {
		items.push({ properties: { title: `Row #${index}` } });
	}
		headerTitle: `Section ${suffix}`,
		items: items,
	});
	return section;
}

	sections: [
		createSection(10, 'A'),
		createSection(10, 'B'),
		createSection(10, 'C')
	],
	editing: false,
	allowsMultipleSelectionDuringEditing: true,
	allowsSelection: false,
	allowsSelectionDuringEditing: true,
	showSelectionCheck: true,
});
listView.addEventListener('longpress', (e) => { // Event not supported on iOS.
	if (e.section && !listView.editing) {
		listView.editing = true;
		listView.selectItem(e.sectionIndex, e.itemIndex);
	}
});
listView.addEventListener('itemsselected', (e) => {
	if (e.selectedItems.length <= 0) {
		listView.editing = false;
	}

	let text = '';
	for (const nextItem of e.selectedItems) {
		if (text.length > 0) {
			text += ',';
		}
		const section = nextItem.section;
		text += ` [${section.headerTitle}][${section.getItemAt(nextItem.itemIndex).properties.title}]`;
	}
	console.log(`itemsselected: ${text}`);
});
window.add(listView);
window.open();
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewHolder.java
```diff
@@ -203,8 +221,8 @@ public void bind(final ListItemProxy proxy, final boolean selected)
 					}
 
 					// Support selected backgrounds.
-					nativeView.setBackground(generateSelectedDrawable(properties, backgroundDrawable));
-					borderView.setActivated(selected);
+					this.container.setBackground(generateSelectedDrawable(properties, backgroundDrawable));
+					this.container.setActivated(selected);
```
jquick-axway, line 225: I could be wrong, but I **think** you need to move the `setActivated()` method call after/outside the `if (nativeView != null) {} block. The reason may be because of the item recycling changes I made in 10.0.1.

If you enable edit mode, scroll to the bottom of a long list, check the last item, scroll back to the top, disable edit mode, re-enable edit mode, and then scroll back to the bottom... you'll find that the last item is still wrongly checked. (All items should be unchecked when enabling edit mode; it's supposed to reset it which works fine for items on-screen.)

android/titanium/src/java/org/appcelerator/titanium/proxy/TiViewProxy.java
```diff
@@ -640,8 +640,8 @@ public void replaceAt(Object params)
 		}
 		if (children != null && children.size() > position) {
 			TiViewProxy childToRemove = children.get(position);
-			insertAt(params);
 			remove(childToRemove);
+			insertAt(params);
```
garymathews, line 644: Switched this around, as replacing child with existing child (changing position) would have remove the child completely.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -197,17 +197,17 @@ private void add(TiUIView child, int childIndex)
...
-					}
-					if (childIndex == -1) {
-						children.add(child);
-					} else {
-						children.add(childIndex, child);
-					}
-					child.parent = proxy;
 				}
 			}
+			if (children.contains(child)) {
```
garymathews, line 202: Move this into outer conditional block, as it would not add the child if its `nativeView` was `null`

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -238,10 +238,10 @@ public void remove(TiUIView child)
 				View nv = getNativeView();
 				if (nv instanceof ViewGroup) {
 					((ViewGroup) nv).removeView(cv);
-					children.remove(child);
-					child.parent = null;
 				}
 			}
+			children.remove(child);
```
garymathews, line 243: Same here, it would only remove the child if its `nativeView` was not `null`

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiRecyclerViewHolder.java
jquick-axway: @garymathews , I'm still seeing a crash on this line. It happens with TableView on Android 5.0 when I set a row's "backgroundColor" property.
```javascript
const defaultRowColor = OS_IOS ? "white" : "?attr/colorBackground";
const rows = [];
for (let index = 1; index <= 20; index++) {
	rows.push(Ti.UI.createTableViewRow({
		title: "Row " + index,
		backgroundColor: defaultRowColor,
	}));
}
const window = Ti.UI.createWindow({ extendSafeArea: false });
const tableView = Ti.UI.createTableView({ data: rows });
tableView.addEventListener("click", (e) => {
	if (e.row.backgroundColor === "orange") {
		e.row.backgroundColor = defaultRowColor;
	} else {
		e.row.backgroundColor = "orange";
	}
});
window.add(tableView);
window.open();
```

jquick-axway (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -350,6 +379,69 @@ public void setActivity(Activity activity)
...
+				TiDimension.TYPE_WIDTH, COMPLEX_UNIT_DIP).getAsDefault(listView);
+			final int y = (int) new TiDimension(listView.getScrollOffsetY(),
+				TiDimension.TYPE_HEIGHT, COMPLEX_UNIT_DIP).getAsDefault(listView);
+
+			contentOffset.put(TiC.PROPERTY_X, x);
+			contentOffset.put(TiC.PROPERTY_Y, y);
+
+			// NOTE: Since obtaining the scroll offset from RecyclerView is unreliable
+			// when items are added/removed, also grab the current visible item instead.
+			final int currentIndex = listView.getAdapterIndex(listView.getFirstVisibleItem().index);
```
jquick-axway, line 400: This line causes a crash if the ListView is empty, because `getFirstVisibleItem()` will return `null` in this case. Maybe we should remove the "index" entry from the `contentOffset` dictionary in this case?

You can reproduce this the below. Tapping on the "Editing" button will cause the `getContentOffset()` method to be called. Alternatively, switch between dark/light mode will cause this too.
```javascript
const listView = Ti.UI.createListView();
const isEditing = () => `EDITING: ${listView.editing ? 'ON' : 'OFF'}`;
const editButton = Ti.UI.createButton({
	title: isEditing(),
	bottom: 5,
	right: 5,
});
editButton.addEventListener('click', () => {
	listView.editing = !listView.editing;
	editButton.title = isEditing();
});
window.add([listView, editButton]);
window.open();
```

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
```diff
@@ -336,6 +344,69 @@ public String getApiName()
...
+				TiDimension.TYPE_WIDTH, COMPLEX_UNIT_DIP).getAsDefault(tableView);
+			final int y = (int) new TiDimension(tableView.getScrollOffsetY(),
+				TiDimension.TYPE_HEIGHT, COMPLEX_UNIT_DIP).getAsDefault(tableView);
+
+			contentOffset.put(TiC.PROPERTY_X, x);
+			contentOffset.put(TiC.PROPERTY_Y, y);
+
+			// NOTE: Since obtaining the scroll offset from RecyclerView is unreliable
+			// when items are added/removed, also grab the current visible item instead.
+			final int currentIndex = tableView.getAdapterIndex(tableView.getFirstVisibleItem().index);
```
jquick-axway, line 365: This crashes if TableView has no rows. Same issue as ListView's `getContentOffset()`.

jquick-axway (changes requested):

jquick-axway:
If you have a `TableView` where you set the "backgroundColor" on the rows, then those rows will show the wrong backgroundColor (they look like they're using the parent's background color) when moving/deleting the row via "editing" or "moving" modes. This only happens in `TableView`. A `ListView` with "backgroundColor" assigned (or not assigned) works fine.

You can reproduce this issue with the below. Tap on the "Editing" button and the slide-to-delete the 1st row or drag-moving the 2nd row. Notice the background color suddenly changes while touch-sliding those rows.
```javascript
	data: [
		{ title: 'Item #1 (editable: true)', editable: true, backgroundColor: 'gray', color: 'white' },
		{ title: 'Item #2 (movable: true)', movable: true, backgroundColor: 'gray', color: 'white' },
		{ title: 'Item #3 (default)', backgroundColor: 'gray', color: 'white' },
	],
});
const isEditing = () => `Editing: ${tableView.editing ? 'ON' : 'OFF'}`;
const btn_editing = Ti.UI.createButton({
	title: isEditing(),
	bottom: 5,
	right: 5
});
btn_editing.addEventListener('click', () => {
	tableView.editing = !tableView.editing;
	btn_editing.title = isEditing();
});
const isMoving = () => `Moving: ${tableView.moving ? 'ON' : 'OFF'}`;
const btn_moving = Ti.UI.createButton({
	title: isMoving(),
	bottom: 5,
	left: 5
});
btn_moving.addEventListener('click', () => {
	tableView.moving = !tableView.moving;
	btn_moving.title = isMoving();
});
win.add([tableView, btn_editing, btn_moving]);
win.open();
```

jquick-axway (approved):
CR/FR: Pass

lokeshchdhry:
FR Passed.

ewanharris (approved):
FR Pass for TIMOB-28533

PR #12778, opened 2021-05-06
fix(android): always clone when row has parent

garymathews:
- Fix behaviour of duplicate rows added to TableView
  - Always clone row when row has a parent

##### TEST CASE
- Check JIRA

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28438)


jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

PR #12781, opened 2021-05-06
fix(android): improve pre-loading optimization of ListView and TableView

garymathews:
- Improve initial pre-load performance of both ListView and TableView
- Queues pre-loading of rows on UI thread, preventing initial stutter
- Pre-loads until user interaction for smooth initial scrolling
- Increased recycler cache size
- Prevent concurrency exception when attempting to load multiple images
  - **NOTE: This would cause some image views to never load and stay blank**

##### TEST CASE 1
```JS

function createSection(rows, suffix) {
		headerTitle: suffix ? `Section ${suffix}` : null
	});
	const items = [];

	for (let i = 1; i <= rows; i++) {
		items.push({
			properties: {
				title: `Row #${i}`,
				image: 'https://picsum.photos/100'
			}
		});
	}
	section.items = items;

	return section;
}

	sections: [ createSection(256) ]
});

win.add(listView);
win.open();
```
##### TEST CASE 2
- Check JIRA

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28439)


garymathews:
@m1ga I believe you ran into an issue where images would sometimes not load? Let me know if this solves that issue for you

jquick-axway:
@garymathews , would you mind making the following performance improvements as well please?

---
We should replace `getItems().length` to `getItemCount()` here.
https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewHolder.java#L239

---
We can get rid of the `setActivity()` call below since `parent.add()` will set the activity for us.
https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java#L295

---
We should initialize the capacity of our `ReferenceTable` HashMap to `1024`. This will help with our app startup times in general.
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java#L26

And we can remove the `references.remove(key)` calls here...
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java#L80
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java#L93
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/ReferenceTable.java#L108

m1ga:
@garymathews I'll try this PR, sounds great! For me the issue with the "not loading images" was already fixed in the master builds after March 12th but I've linked it in the other issue I had because there was someone else where the master build wasn't working. So perhaps that issues is fixed for him too

jquick-axway (approved):
CR: Pass

lokeshchdhry:
@garymathews , With the fix I see an increase in the time from splash to showing the window of 3 - 4 seconds as compared to 2 seconds without. Is this expected ?

lokeshchdhry:
FR Passed.

PR #12782, opened 2021-05-07
fix(android): production builds crash with Node 14

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28440

**Test:**
1. Set up Node 14 on your machine.
2. Build a Titanium project for Android with deployment type "dist-playstore" (aka: production).
3. Verify the build finishes successfully. _(This is the fix.)_
4. Verify you can install and run the built APK.


ewanharris (approved):
Changes look good to me. I pushed a commit to mirror the change to the iOS build as the same issue was occurring there

PR #12788, opened 2021-05-08
fix(ios): ScrollView setZoomScale()/setContentOffset() methods crash as of 10.0.0

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28443

**Summary:**
- This is a regression as of Titanium 10.0.0.
- Calling `Ti.UI.ScrollView` methods `setZoomScale()` or `setContentOffset()` crash with function not found.
- These methods are needed since they support a 2nd argument.

**Test:**
1. Build and run the code shown in [TIMOB-28443](https://jira.appcelerator.org/browse/TIMOB-28443) on iOS.
2. Double tap on the red square.
3. Verify that square doubles in size with an animation.


garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

PR #12815, opened 2021-05-14
fix(android): guard access to some toast properties

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28448

Prevents error logs like these that could become more severe in future Android versions:
Screenshot 2021-05-14 at 15 19 21

Test case:
```js
    Ti.UI.createNotification({
        title: 'Hello world',
        duration: Ti.UI.NOTIFICATION_DURATION_SHORT
    }).show();
});
window.open();
```


jquick-axway (approved):
CR: Pass

PR #12824, opened 2021-05-20
docs: info about initial value for iOS animation

m1ga:

**Optional Description:**

as recommended in Slack


hansemannn:
Why not just fix this in the core instead, like with a default value? :-)

jquick-axway:
I agree with @hansemannn . This isn't an issue on Android and it should be fixed on iOS.
Closing this PR in favor of PR #12832 .

PR #12829, opened 2021-05-21
fix(ios): hyperloop defineClass() crash

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28458

**Summary:**
- Regression as of Titanium 10.0.0.
- Caused by removal of `KrollBridge.require()` method require by hyperloop [here](https://github.com/appcelerator/hyperloop.next/blob/master/iphone/titanium/HyperloopModule.m#L310).
- Also, the `global.require` function was removed in 10.0.0 which is needed by our core JS code [here](https://github.com/appcelerator/titanium_mobile/blob/master/common/Resources/ti.internal/extensions/binding.js#L36).

**Test:**
1. Build and run [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples) for iOS.
2. Go to every window in the app and verify they were as expected.


common/Resources/ti.internal/extensions/binding.js
```diff
@@ -32,8 +32,8 @@ function isHijackableModuleId(path) {
 	return firstChar !== '.' && firstChar !== '/';
 }
 
-// Hack require to point to this as a core module "binding"
-const originalRequire = global.require;
+// Hack require to point to this as core module "binding". (Note that iOS does not have a global require.)
+const originalRequire = global.require ? global.require : require;
```
sgtcoolguy, line 36: This came up in a liveview change @ewanharris had worked on. I think he changed to using something like Module.main.require as the ultimate fallback, which makes sense to me.

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -563,6 +563,25 @@ - (id)krollObjectForProxy:(id)proxy
...
 }
 
+- (id)require:(KrollContext *)kroll path:(NSString *)path
+{
+  if (!kroll || !path) {
+    return nil;
+  }
+
+  JSContext *jsContext = [JSContext contextWithJSGlobalContextRef:kroll.context];
+  JSValue *jsResult = [jsContext.globalObject invokeMethod:@"require" withArguments:@[ path ]];
```
sgtcoolguy, line 573: This is explicitly invoking `global.require`, which probably is fine for hyperloop's usage (since I think it's always dealing with hyper loop wrappers and "absolute" require paths (relative to app root). I think the "most correct" way would be to grab the `require` function from the current context, but honestly I'm not sure it'd work (ideally we'd want the require function passed in to the current module's context/wrapper to be invoked so relative paths get resolved properly).

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
```diff
@@ -563,6 +563,25 @@ - (id)krollObjectForProxy:(id)proxy
   return result;
 }
 
+- (id)require:(KrollContext *)kroll path:(NSString *)path
```
sgtcoolguy, line 566: Should the method API need to be added back to the header file as well?

jquick-axway: We already do. `KrollBridge` inherits from `KrollDelegate` and this method is defined in its header.
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.h#L22
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.h#L17

common/Resources/ti.internal/extensions/binding.js
jquick-axway: I agree. I should be able to access it via `require.main.require`. I'll make the change.
https://github.com/appcelerator/titanium_mobile/blob/master/common/Resources/ti.internal/kernel/module.js#L520

iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
jquick-axway: Yeah, I'm not sure what else I should do here. It may be bad to assume to use `[JSContext currentContext]` here... and the caller can easily do this anyways. My goal was to resurrect this method for hyperloop which is always given an absolute path anyways.

janvennemann (approved):

PR #12839, opened 2021-05-25
fix(android): minor Ti.UI.ScrollableView refactor

garymathews:
- Minor refactor of `Ti.UI.ScrollableView` to move reference of child views into proxy
- Fix activity assignment of `headerView` and `footerView` in `ListSection`

##### TEST CASE
```JS
const headerView = Ti.UI.createView({ height: 100 });
const scrollableView = Ti.UI.createScrollableView({
	height: 100,
	backgroundColor: 'yellow'
});
const headerLabel = Ti.UI.createLabel({
	text: 'This label should be visible',
	color: 'red'
});

scrollableView.addView(headerLabel);
//scrollableView.views = [ headerLabel ]; // QE: Try this too

headerView.add(scrollableView);

	sections: [ Ti.UI.createListSection({ headerView }) ]
});

win.add(listView);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28459)


android/modules/ui/src/java/ti/modules/titanium/ui/ScrollableViewProxy.java
```diff
@@ -190,40 +121,107 @@ public boolean handleMessage(Message msg)
...
+			scrollableView.getAdapter().notifyDataSetChanged();
+		}
 	}
 
 	@Kroll.method
 	public void removeView(Object viewObject)
 	{
-		TiMessenger.sendBlockingMainMessage(getMainHandler().obtainMessage(MSG_REMOVE_VIEW), viewObject);
+		if (viewObject instanceof Number) {
+			this.views.remove(viewObject);
```
jquick-axway, line 209: I'm pretty sure you need to cast `viewObject` as `(int)` so that it will call the [remove(int)](https://developer.android.com/reference/java/util/ArrayList#remove(int)) method instead of the [remove(Object)](https://developer.android.com/reference/java/util/ArrayList#remove(java.lang.Object)) method.

android/modules/ui/src/java/ti/modules/titanium/ui/ScrollableViewProxy.java
```diff
@@ -190,40 +121,107 @@ public boolean handleMessage(Message msg)
...
+				view.setActivity(getActivity());
+				view.setParent(this);
+				this.views.add(insertIndex, view);
+			}
+
+		} else if (viewObject instanceof Object[]) {
+			Object[] views = (Object[]) viewObject;
+
+			for (int i = 0; i < views.length; i++) {
+				insertViewsAt(insertIndex, views[i]);
```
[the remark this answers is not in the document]
garymathews: That's how our current code did it, but you're right we should have tests for this
[TiUIScrollableView.java#L491](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollableView.java#L491)

android/modules/ui/src/java/ti/modules/titanium/ui/ScrollableViewProxy.java
jquick-axway: The `ScrollableViewAddRemoveTest.js` script I attached to [TIMOB-28459](https://jira.appcelerator.org/browse/TIMOB-28459) proves that calling `ScrollableView.removeView()` by index is now broken. When running this script, tap on the "Remove" button and select "Current" page from the dialog to remove the currently displayed page by index. The script's remove "First" and "Last" buttons remove by reference and work correctly.

jquick-axway:
I've found the following issues. The `ScrollableViewAddRemoveTest.js` script I attached to [TIMOB-28459](https://jira.appcelerator.org/browse/TIMOB-28459) can help test this.
- Setting the `currentPage` property before the window is opened no longer works. It always defaults to the 1st page.
- Passing an index to `removeView()` no longer works.
- Views added to ScrollableView after the window are opened are lost when switching dark/light themes. (Note that the proxy's views are released and a new activity is assigned when this happens.)

android/modules/ui/src/java/ti/modules/titanium/ui/ScrollableViewProxy.java
```diff
@@ -66,164 +65,153 @@ public ScrollableViewProxy()
...
+		} else if (viewObject instanceof TiViewProxy) {
+			final TiViewProxy view = (TiViewProxy) viewObject;
+
+			this.views.remove(view);
+
+		} else {
+
+			// Skip updating adapter.
+			return;
+		}
```
jquick-axway, line 210: There's a bug here where if property "currentPage" references the last page and you remove via this method, the "currentPage" property does not get updated and becomes invalid. This can cause an exception to be thrown if you use this invalid index.

You can reproduce this with my `ScrollableViewAddRemoveTest.js` script attached to [TIMOB-28459](https://jira.appcelerator.org/browse/TIMOB-28459).
1. Scroll to the last page.
2. Tap on the Remove button.
3. Select dialog's "Current" option to delete current page. ("currentPage" is now invalid.)
4. Tap on the Remove button again.
5. Select dialog's "Current" option again.
6. Notice an exception is thrown due to an invalid index.

jquick-axway (changes requested):

jquick-axway (approved):
CR: Pass

lokeshchdhry:
FR Passed.

hansemannn:
I think this caused a regression ([TIMOB-28503](https://jira.appcelerator.org/browse/TIMOB-28503) that's still existing in latest 10_0_X. Would be great to check this, thank you! :-)

PR #12865, opened 2021-06-03
fix(adb): check err.code instead of err.errno on socket error

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28464

Updates node-titanium-sdk to include the fix for the above


jquick-axway (approved):

lokeshchdhry:
FR Passed.

PR #12872, opened 2021-06-08
fix(ios): broken subclass on iOS 15

garymathews:
- For this particular case, `isKindOfClass` comparison always returns `true` for any class, which is clearly wrong
- ~The workaround validates the selector instead~
- Amend `Module` inheritance with `NSObject`

###### NOTES
- Xcode debugger shows `module` inheritance correctly
- Apps built using Xcode 12 and run fine on <iOS 15 are affected by this issue on iOS 15
  - _So all current 10.0.0 Titanium apps fail to launch on iOS 15_
- I have not tested complex apps with this workaround yet

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28483)


hansemannn:
Wait, so all apps in the App Store that do not have this patch will crash, even if built with Xcode 12.x? This will cause a lot of trouble soon :-( Probably requires a lot of backports

garymathews:
That's currently the case, it appears to be an issue in the new Obj-C runtime on iOS 15. It's possible this could be fixed before iOS 15 is released.

janvennemann:
@garymathews The real issue appears to be our definition of the [`Module`](https://github.com/appcelerator/titanium_mobile/blob/90f17bd111b49c86051ad7cc4de33e322beb68d0/iphone/TitaniumKit/TitaniumKit/Sources/API/Module.h#L5) protocol. We use it [here](https://github.com/appcelerator/titanium_mobile/blob/90f17bd111b49c86051ad7cc4de33e322beb68d0/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollModule.m#L58) to tell the compiler that the `module` var conforms to the `Module` protocol. However, our `Module` protocol does not inherit from the `NSObject` protocol. 

In a simple test case the compiler correctly errors our with `No known instance method for selector 'isKindOfClass:'`. I'm not really sure why this will be silently ignored in our case or why it suddenly breaks during runtime now with iOS 15. Long story short, to fix the underlying issue simply inherit from the `NSObject` protocol.

```objc
@protocol Module <NSObject>
```

garymathews:
@janvennemann Updated PR

janvennemann (approved):

ewanharris (approved):
FR Pass. Applications now startup ok on iOS 15

hansemannn:
Btw: The ObjCModule was only introduced in SDK 9 or 10, right? So older apps should not be affected by this

ewanharris:
Yup, this code was introduced in SDK 10 (#11693) so this only impacts SDK 10. I tested an app using SDK 9.3.2.GA/Xcode 12 and it launched successfully on iOS 15

PR #12873, opened 2021-06-09
Bump ioslib, declare xcode 13/ios 15 compat, add ios 15 macro

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28467

Related to #12872 which actually makes it work

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -264,7 +264,7 @@ timestamps {
 
 					// Install dependencies
 					timeout(5) {
-						sh 'npm ci'
+						sh 'npm ci -ddd'
 					}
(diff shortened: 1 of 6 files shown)


garymathews (approved):
LGTM!

hansemannn:
Can this be merged? Then the iOS 15 testing could become a bit less stressful each time :-)

ewanharris:
@hansemannn, this is kinda blocked for now as I'm currently investigating an issue that's cropped up after the latest rebase where the SDK package script gets stuck in an infinite loop and then crashes with an OOM error

PR #12879, opened 2021-06-10
fix(ios): top section can disappear in grouped ListView

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28109

**Summary:**
When a `ListView` with "style" set to `GROUPED` or `INSET_GROUPED` is configured and the top-most section has not title or views, then that top-most section will disappear when tapping on a `SearchBar` or relayout such as an orientation change.

**Test:**
2. Tap on the search bar.
3. Verify that space above the first section does not disappear.
4. Rotate to landscape and verify the space above the section is still there.
5. In the code below, use the `INSET_GROUP` style and rebuild.
6. Repeat steps 2-4.

```javascript
	style: OS_IOS ? Ti.UI.iOS.ListViewStyle.GROUPED : undefined,
//	style: OS_IOS ? Ti.UI.iOS.ListViewStyle.INSET_GROUPED : undefined,
	searchView: Ti.UI.createSearchBar(),
	keepSectionsInSearch: true,
	sections: [
		Ti.UI.createListSection({
			items: [
				{ properties: { title: 'Apples', searchableText: 'Apples' } },
				{ properties: { title: 'Bananas', searchableText: 'Bananas' } },
				{ properties: { title: 'Oranges', searchableText: 'Oranges' } },
			],
		}),
		Ti.UI.createListSection({
			items: [
				{ properties: { title: 'Carrots', searchableText: 'Carrots' } },
				{ properties: { title: 'Onions', searchableText: 'Onions' } },
				{ properties: { title: 'Tomatoes', searchableText: 'Tomatoes' } },
			],
		}),
	],
});
window.add(listView);
window.open();
```


iphone/Classes/TiUIListView.m
```diff
@@ -1749,6 +1749,11 @@ - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSIntege
     if (size < DEFAULT_SECTION_HEADERFOOTER_HEIGHT) {
       size += DEFAULT_SECTION_HEADERFOOTER_HEIGHT;
     }
+  } else if (0 == section) {
+    // Make sure grouped style always shows the 1st header if it contains no text/view.
+    if ((tableView.style == UITableViewStyleGrouped) || (tableView.style == UITableViewStyleInsetGrouped)) {
+      size = 35.0;
```
hansemannn, line 1755: I don't feel well with static sizes, could this maybe be received other way?

jquick-axway: I would if I knew how.

Note that this code will only be used if the 1st section has no text/views and for grouped mode only. So, it's a fallback.

janvennemann: After some testing with a native project it seems like it is the default behavior that the header view is not displayed if it is empty. Although i'm not sure why it renders the header view initially, handling this edge case with a hardcoded value seems to be the only viable solution for now.

Speaking of the hardcoded value, where did you get `35.0` from? I see there already is a `DEFAULT_SECTION_HEADERFOOTER_HEIGHT` constant defined in the code. Why not use that?

jquick-axway: The `DEFAULT_SECTION_HEADERFOOTER_HEIGHT` is `29.0` which does not match Apple's "grouped" section height. I've compared screenshots and looked at the pixels. Apple is using a 35 point height. Others on stackoverflow have noted the same for the past several years. I wish I could grab the height via an API, but I haven't seen anything available. If you can think of a better alternative, then I'm all ears.

janvennemann (approved):

lokeshchdhry:
FR Passed.

PR #12881, opened 2021-06-11
chore(android): add "android:exported" to all activities/services with intent-filters

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28475

**Summary:**
- Adds `android:exported="true/false"` to all Titanium activities/services that have intent-filters.
- Added to Titanium activities without intent-filters in case custom "AndroidManifest.xml" defined by app developer adds an intent-filter to them.
- Needed due to a breaking-change where apps/libraries targeting Android 12 will have a build failure if activities, receivers, or services with an intent-filter don't define an "android:exported" attribute. Must be explicitly defined.

**Test:**
1. Create a "Classic" project with the below "tiapp.xml" setting.
2. Build and run on Android. _(Build failure used to occur when targeting Android 12.)_

```xml
<ti:app>
	<android>
		<manifest>
			<uses-sdk android:targetSdkVersion="31"/>
		</manifest>
	</android>
</ti:app>
```


garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

PR #12882, opened 2021-06-11
refactor(android): remove warning about sigalg compatability for older android versions

ewanharris:
Given that the min SDK is now Android 5 this warning is no longer needed

To test create a keystore using `keytool -genkey -alias testing1234 -keyalg RSA -dname "CN=test,OU=test,O=test,L=test,ST=test,C=TE" -validity 9125 -keypass testing -storepass testing -keystore keystore.jks` and then build using it. The warning should no longer be there

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -266,10 +266,6 @@ AndroidBuilder.prototype.config = function config(logger, config, cli) {
 								if (!alias) {
 									return callback(new Error(__('Invalid "--alias" value "%s"', value)));
 								}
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
LGTM! **APPROVED**

jquick-axway (approved):

jquick-axway:
I confirmed that `SHA256withRSA` is supported on Android 4.3 (aka: API Level 18) and higher.
https://developer.android.com/training/articles/keystore#SupportedSignatures

PR #12885, opened 2021-06-12
test: fix unit test failures

jquick-axway:
**Summary:**
- Fixed `Ti.UI.ScrollView.contentOffset` unit test failure.
  * Will return originally assigned point string if already at given position if hasn't moved yet or already at that position.
- Fixed `Ti.Netwrok.HTTPClient` unit test failures with `https://www.google.com`.
  * It looks like it's trying to redirect which is not the response the unit test expects.


ewanharris (approved):
Looks good to me, thanks for checking these out!

PR #12900, opened 2021-06-16
fix(ios): export original native module object

janvennemann:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28491

**Optional Description:**

Skips creating a wrapper module on iOS to ensure the JavaScriptCore callbacks for a proxy are used to access the native properties.

**Test Case**

Run the attached test module to easily test the fix. The example app included in the module should log the following, indicating the the native property value was successfully changed:

```
[INFO]  Initial value: Foo
[INFO]  New value: Bar
```

[proptest.zip](https://github.com/appcelerator/titanium_mobile/files/6665874/proptest.zip)


sgtcoolguy (approved):

lokeshchdhry:
FR Passed.

PR #12903, opened 2021-06-17
test(macos): disable animate width/height unit test

jquick-axway:
**Note:**
Unit test only fails for Jenkins node. I think the animation "complete" event won't fire if there is no monitor connected to the Mac... or really the animation doesn't happen because it can't be seen.


ewanharris (approved):
LGTM!

PR #12908, opened 2021-06-18
fix(android): amend actionBarStyle for MaterialComponents

garymathews:
- Amend `actionBarStyle` to use `MaterialComponents` instead of `AppCompat`

##### TEST CASE
```JS

window.addEventListener('open', _ => {
	const actionBar = window.activity.actionBar;

	if (actionBar) {
		actionBar.title = 'ActionBar';
		actionBar.homeButtonEnabled = true;
		actionBar.displayHomeAsUp = true;
		actionBar.onHomeIconItemSelected = _ => {
			alert('Home');
		};
	}
});

window.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28493)

### DIFF
--- a/android/titanium/res/values/values.xml
+++ b/android/titanium/res/values/values.xml
@@ -59,7 +59,7 @@
 	<!-- Titanium's light material theme with an action bar. -->
 	<style name="Theme.Titanium.Light" parent="Theme.MaterialComponents.Light.DarkActionBar">
 		<item name="actionBarSize">@dimen/ti_action_bar_height</item>
-		<item name="actionBarStyle">@style/Widget.AppCompat.Light.ActionBar.Solid</item>
+		<item name="actionBarStyle">@style/Widget.MaterialComponents.Light.ActionBar.Solid</item>
 		<item name="colorPrimary">@color/ti_light_primary</item>
 		<item name="colorPrimaryDark">@color/ti_light_primary_dark</item>
 		<item name="colorAccent">@color/ti_light_accent</item>


jquick-axway:
@garymathews, the reason I used "Widget.AppCompat.Light.ActionBar.Solid" for the "actionBarStyle" was to work-around an ActionBar issue where the "subtitle" text would be clipped when shown in landscape. The ActionBar is expected to be shorter in landscape and taller in portrait, but Google's built-in themes mishandles this as can be seen here...
https://github.com/material-components/material-components-android/issues/779

So, I fixed the landscape ActionBar height issue by customizing it myself just like how it used to work.
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/res/values-land/values.xml
https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/res/values/values.xml#L19

But Google's material ActionBar style sets min heights regardless of orientation which will cause a problem withe shorter landscape height. However, the AppCompat style does correctly compensate for landscape which works-around the problem.
https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/appbar/res/values/styles.xml#L98-L104

I didn't write an XML comment about this. So, it's my fault for not clarifying why it was set up this way.

---
**Test Code:**
```javascript
window.addEventListener("open", () => {
	const actionBar = window.activity.actionBar;
	actionBar.title = "My Title";
	actionBar.subtitle = "My Subtitle";
	actionBar.displayHomeAsUp = true;
	actionBar.homeButtonEnabled = true;
	actionBar.onHomeIconItemSelected = () => {
		alert("Hello World!");
	};
});
	var menuItem = e.menu.add({
		title: "Menu Item 1",
		showAsAction: Ti.Android.SHOW_AS_ACTION_IF_ROOM,
	});
	menuItem.checked = true;
	console.log("@@@ menuItem.checked: " + menuItem.checked);
	menuItem.addEventListener("click", function() {
		const dialog = Ti.UI.createAlertDialog({
			title: "Alert",
			message: "Menu item was clicked on.",
			buttonNames: ["OK"],
		});
		dialog.show();
	});
};
window.open();
```

Style "Widget.MaterialComponents.Light.ActionBar.Solid" mishandles text in landscape...
LandscapeBad

Style "Widget.AppCompat.Light.ActionBar.Solid" correctly handles portrait/landscape text sizing...
LandscapeGood

hansemannn:
Since the subtitle issue is not resolved so far, I would leave the API as-is for now. Hopefully, this is resolved in future Material Components updates!

PR #12909, opened 2021-06-19
fix(android): improve ListView scroll performance

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28439

**Summary:**
- Improved `ListView` performance by recycling scrolled-off child proxies and their views.
- Modified `View.add()` to not release native views if they're changing `ViewGroup` parents.

---
**Complex Template Test:**
_(This is the customer's alloy test app converted to a classic app.)_
1. Build and run `ListViewTemplateMultiComplexTest.js` attached to [TIMOB-28439](https://jira.appcelerator.org/browse/TIMOB-28439).
2. Quickly scroll the `ListView` all the way down/up and verify it doesn't stutter.

---
**List of TextAreas/Fields Test:**
_(Verifies ListItem view changes are restored after scrolling it off and back on screen.)_
1. Build and run `ListOfTextAreasTest.js` attached to [TIMOB-28439](https://jira.appcelerator.org/browse/TIMOB-28439).
2. Notice there are 20 text areas in the 1st section and 20 text fields in the 2nd section of the list view.
3. Enter "abc" into the top-most text area in the list.
4. Scroll to the very bottom.
5. Enter "xyz" into the bottom-most text field in the list.
6. Scroll to the top and verify "abc" is still in the top-most text area.
7. Scroll to the bottom and verify "xyz" is still in the bottom-most text field


android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
```diff
@@ -301,6 +260,202 @@ public void call(Object data)
...
+			} else {
+				cachedProperties = new KrollDict();
+				this.childProperties.put(bindsEntry.getKey(), cachedProperties);
+			}
+
+			// Copy all view proxy properties to this list item's dictionary.
+			// Saves view's current state when scrolled offscreen so we can restore it when scrolled back in.
+			for (KrollDict.Entry<String, Object> proxyPropertyEntry : proxyProperties.entrySet()) {
+				cachedProperties.put(proxyPropertyEntry.getKey(), proxyPropertyEntry.getValue());
+			}
```
garymathews, line 394: Looks like this could be simplified a little
```Java
// Fetch properties dictionary to store child view's changes to.
HashMap cachedProperties = null;
Object object = this.childProperties.get(bindsEntry.getKey());
if (object instanceof HashMap) {
    cachedProperties = (HashMap) object;
    cachedProperties.putAll(proxyProperties);
} else {
    cachedProperties = new KrollDict(proxyProperties);
    this.childProperties.put(bindsEntry.getKey(), cachedProperties);
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
```diff
@@ -301,6 +260,202 @@ public void call(Object data)
...
+			if (childPropertiesObj instanceof HashMap) {
+				properties.putAll((HashMap) childPropertiesObj);
+			}
+
+			// Assign properties to child. For best performance, only update property if value is different.
+			for (KrollDict.Entry<String, Object> entry : properties.entrySet()) {
+				if (proxy.shouldFireChange(proxy.getProperty(entry.getKey()), entry.getValue())) {
+					proxy.setPropertyAndFire(entry.getKey(), entry.getValue());
+				}
+			}
```
garymathews, line 447: Since `setPropertyAndFire()` calls `shouldFireChange()`, it's doing the check twice. Although, I doubt this gains anything significant
```Java
for (KrollDict.Entry<String, Object> entry : properties.entrySet()) {
    final String key = entry.getKey();
    final Object newValue = entry.getValue();
    final Object oldValue = proxy.getProperty(key);
    
    if (proxy.shouldFireChange(oldValue, newValue)) {
        proxy.setProperty(key, newValue);
        proxy.firePropertyChanged(key, oldValue, newValue);
    }
}
```
I'm curious if you can get away with not updating the V8 object?
```Java
proxy.getProperties().put(key, newValue);
```

jquick-axway: I think we do need to let it update the V8 object's properties since an event fired for this child proxy will be referenced via the event's "source" property. The [shouldFireChange()](https://github.com/appcelerator/titanium_mobile/blob/e2894aa26b0dbae2cc86436ec891dc756264c99e/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java#L1039-L1049) method is pretty simple so I think it's fine.

**_Side Note**: Only updating properties that have changed (versus update all properties) is a huge performance improvement._

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
jquick-axway: Yeah, I agree with you. And doing a `putAll()` should be faster too.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
```diff
@@ -301,6 +260,195 @@ public void call(Object data)
...
+		// Do not continue if templates don't match. Prevents given proxy from re-using this proxy's views.
+		// Note: Should never happen if ListViewAdapter.getItemViewType() returns different IDs for different templates.
+		String templateId = getTemplateId();
+		if ((templateId == null) || !templateId.equals(proxy.getTemplateId())) {
+			return;
+		}
+
+		// Do not continue if this proxy has no children.
+		// This will force given proxy to regenerate child proxies from template.
+		TiViewProxy[] childProxies = getChildren();
```
garymathews, line 342: Very minor, but saves an allocation and makes use of the `hasChildren()` method you introduced
```Java
if (!hasChildren()) {
    return;
}
final TiViewProxy[] childProxies = getChildren();
if (childProxies[0].peekView() == null) {
    return;
}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListItemProxy.java
```diff
@@ -80,44 +80,32 @@ public ListItemProxy(boolean placeholder)
...
+					setProperty(entry.getKey(), entry.getValue());
 				}
-
-				return new ItemView(this);
 			}
 		}
 
-		return null;
+		// Generate all child proxies from template and create all views.
+		if (this.children.isEmpty()) {
```
garymathews, line 105: Maybe use `hasChildren()` here as it also includes a `null` check

garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.
Wow its is pretty fast & smooth scrolling now :)

PR #12914, opened 2021-06-22
chore(android): update gradle and cmake

garymathews:
**Prerequisite:**
PR #12959 must be merged first. It enables Jenkins to build with JDK 11.

---
**Summary:**
- Update gradle plugin to `7.0.0`
- Update `gradle` to `7.1` 
  - For compatibility with JDK 16
  - Included duplication strategy to `kroll-apt` to handle duplication error
- ~Update `cmake` to `3.18.1`~

[TIMOB-28494](https://jira.appcelerator.org/browse/TIMOB-28494)
~[TIMOB-28495](https://jira.appcelerator.org/browse/TIMOB-28495)~


android/titanium/build.gradle
```diff
@@ -121,6 +125,24 @@ android {
 	}
 }
 
+// Install cmake during project configuration.
+def cmakeDir = new File(sdkDir+'/cmake/'+android.externalNativeBuild.cmake.version)
```
garymathews, line 129: I've reverted updating cmake, however maybe we should keep this check here so we can easily update cmake in the future?

jquick-axway:
Upgrading to gradle 7 does not appear to be enough to support JDK 16. I get the following gradle error when I attempt to build an app with this PR and JDK 16 (but it builds fine with JDK 15). I ran into this same error when attempting to build a simple non-Titanium app created via Android Studio.
```
Execution failed for task ':app:processDebugMainManifest'.
> Unable to make field private final java.lang.String java.io.File.path accessible: module java.base does not "opens java.io" to unnamed module @39b7cf4c
```

I think we have to wait for Google to release Android Google Plugin version 7 for JDK 16 support. When I updated my non-Titanium app project in Android Studio to use AGP "7.0.0-beta04", it built fine with JDK 16.

android/titanium/build.gradle
jquick-axway: Yeah, I think this check is fine. Thanks.

jquick-axway (approved):
CR: Pass

jquick-axway:
@garymathews, can you also update the min JDK in our "package.json" too please?
https://github.com/appcelerator/titanium_mobile/blob/master/android/package.json#L34

lokeshchdhry:
FR Passed.
An app builds fine with this fix.

PR #12923, opened 2021-06-24
feat(android): add headingTime, fix headingFilter

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/[TICKET] **todo**

**Optional Description:**

The current compass heading has a default 250ms waiting interval. Since I didn't want to remove it I'll added a parameter to change it. Changing it to < 100 will result in a smoother update interval.

This PR also fixes headingFilter so it will work on Android now.

```javascript
Ti.Geolocation.headingTime = 100;
Ti.Geolocation.headingFilter = 2;

Ti.Geolocation.addEventListener('heading', function(e) {
  if (e.error) {
  	return;
  }
  console.log(e.heading.magneticHeading);
});
```

full example and JIRA ticket will be added soon


jquick-axway:
@m1ga , just to let you know, when you target the upcoming Android 12, an exception will be thrown if you set the interval less than 200 milliseconds... unless you add the new [HIGH_SAMPLING_RATE_SENSORS](https://developer.android.com/reference/android/Manifest.permission#HIGH_SAMPLING_RATE_SENSORS) permission to the manifest.
https://developer.android.com/about/versions/12/behavior-changes-12#motion-sensor-rate-limiting

m1ga:
@jquick-axway Thanks for the info! I've added it to the apidoc

hansemannn (approved):

PR #12942, opened 2021-07-07
fix(android): ScrollableView "views" property wrongly empty before window open as of 10.0.1

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28504

**Summary:**
- Regression as of Titanium 10.0.1. Caught before release.
- "views" property does not return same array assigned upon creation before window open, but works after window opens.

**Test:**
Covered by this PR's unit tests.


garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

PR #12905, opened 2021-06-17
feat(ios): add LiveView support to Hyperloop

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-28461
- https://jira.appcelerator.org/browse/TIMOB-28508

**Summary:**
- Support running LiveView with Hyperloop.
  * No longer modifies JS files containing Hyperloop requires/imports. _(This was the issue.)_
  * Now injects a `hyperloop.bootstrap.js` providing bindings between native types to JS files.
  * Implemented the same way as Android, which supports LiveView as of Titanium 9.0.0.
- Removed 3 year old native property access migration code.
- Re-added 32-bit support needed by Titanium 9.x.x.
  * This reverts [TIMOB-28334](https://jira.appcelerator.org/browse/TIMOB-28334).
- Updated Hyperloop module version to [7.0.4](https://github.com/appcelerator/hyperloop.next/releases/tag/v7.0.4).
  * Module works with Titanium 9.2.0 and higher.

**Note:**
The following still won't work with LiveView because they require a native rebuild.
- Adding libraries.
- Adding `Hyperloop.defineClass()` calls. _(Because it involves native code generation.)_

---
**Test:**
1. Add this PR's hyperloop module to [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples).
2. Build and run with LiveView: `appc run -p ios --liveview`
3. Go to every window in the example app and verify it works as expected.
4. While LiveView is running, open file: `hyperloop-examples/app/controllers/ios/label.js`
5. Change variable `infoString` from `'We ♥ iOS'` to `'We ♥ LiveView'` and save.
6. Verify app restarts without error.
7. Go to the app's Label window and verify the text has been updated.
8. Do an encrypted build with LiveView: `appc run -p ios -D test --liveview`
9. Go to every window in the example app and verify it works as expected.

PR #12973, opened 2021-07-23
fix(ios): modules to support SDK 9.x.x and iOS 10

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28509

**Summary:**
- Updated bundled modules to include 32-bit libraries and support iOS 10.
- Needed to make modules work with Titanium 9.x.x since their "minsdk" are still set to 9.2.0.

### DIFF
--- a/support/module/packaged/modules.json
+++ b/support/module/packaged/modules.json
@@ -1,32 +1,32 @@
 {
 	"ios": {
 		"urlSession": {
-			"url": "https://github.com/appcelerator-modules/ti.urlsession/releases/download/v4.0.0/com.appcelerator.urlSession-iphone-4.0.0.zip",
-			"integrity": "sha512-nB7R1YIOhH+QJaX0MQN28rByYNR7TfPGm/rdyEGDmiCOyPXWVK5QyKUoGJ6hPbsud+9aVztZQ7mnEmGyVgkFmQ=="
+			"url": "https://github.com/appcelerator-modules/ti.urlsession/releases/download/v4.0.1/com.appcelerator.urlSession-iphone-4.0.1.zip",
+			"integrity": "sha512-F3t4bOjAfqBuxehOYw+Z4YYP3MT9zeIMLkmigQ/vKU8R7WlwNC/hkmsVG9o+xRsy7eykRPN0QcqtBvgDszPHQg=="
 		},
 		"facebook": {
-			"url": "https://github.com/appcelerator-modules/ti.facebook/releases/download/v11.0.0-iphone/facebook-iphone-11.0.0.zip",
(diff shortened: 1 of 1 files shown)


jquick-axway:
@ewanharris , would you mind testing this PR out on your iPhone 5 please? Simplest thing to do is...
1. Install this PR's Titanium SDK. (Which will install its modules.)
2. Set up "tiapp.xml" to build with Titanium 9.2.0 and this PR's "ti.map" version 5.1.1.
3. Build and run with the below for iPhone 5.

```javascript
const map = require("ti.map");
const annotations = [];
annotations.push(map.createAnnotation({
	userData: { type: "waypoint" },
	bubbleParent: false,
	title: "Hello World",
	latitude: 51.507,
	longitude: 0.128,
}));
const mapView = map.createView({ width: Ti.UI.FILL, height: Ti.UI.FILL });
mapView.addAnnotations(annotations);
window.add(mapView);
window.addEventListener("open", () => {
	mapView.showAnnotations(annotations);
});
window.open();
```

For the other modules, maybe simply requiring them in will be enough to test loading their libraries?

garymathews:
Another test for Hyperloop is to include a native module and attempt to utilize it

**\<project\>\Podfile**
```
install! 'cocoapods',
    :integrate_targets => false

platform :ios, '8.0'
use_frameworks!
target 'alloyTest' do
    pod 'MTBBarcodeScanner'
end
```

**app.js**
```JS
const MTBBarcodeScanner = require('MTBBarcodeScanner/MTBBarcodeScanner');
```

Project should build correctly, should not see this error:
```
[TRACE] Undefined symbols for architecture i386:
[TRACE]   "_OBJC_CLASS_$_HyperloopPointer", referenced from:
[TRACE]       objc-class-ref in mtbbarcodescanner.o
[TRACE]   "_OBJC_CLASS_$_HyperloopUtils", referenced from:
[TRACE]       objc-class-ref in mtbbarcodescanner.o
[TRACE] ld: symbol(s) not found for architecture i386
[TRACE] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

jquick-axway:
@ewanharris , @garymathews , regarding the hyperloop test, note that I updated the hyperloop module in PR #12905 (I combined it with my iOS LiveView fix).

ewanharris (approved):
FR pass, able to load the following modules successfully on an iPhone 5 using 9.3.2.GA

* facebook
* ti.coremotion
* ti.webdialog
* ti.identity
* ti.map
* urlSession

Note: I can't test that ti.applesignin loads on the iPhone 5 as it requires iOS 13+

PR #12985, opened 2021-07-26
fix(ios): do not create invalid swift compiler flags from symbol usage

ewanharris:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-28510
- https://jira.appcelerator.org/browse/TIMOB-28496

**Summary:**
* Update node-titanium-sdk to improve detection of Ti API usage
* Add a check into the iOS build where we transform those values to ignore any values that are not just an uppercased string
* Fix build failure on 64-bit Linux


garymathews (approved):
LGTM!

lokeshchdhry:
FR Passed.
No compile errors seen.

PR #12986, opened 2021-07-29
chore: update ti.cloudpush

garymathews:
- Update `ti.cloudpush` to version `8.0.0`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28480)


jquick-axway (changes requested):
FR: Failed

On Android 12 and when targeting API Level 31, tapping a push notification fails to launch app unless its UI is backgrounded. It looks like we're running into the following breaking-change...
https://developer.android.com/about/versions/12/behavior-changes-12
https://developer.android.com/reference/android/app/Notification.Builder#setContentIntent(android.app.PendingIntent)

jquick-axway (commented):
CR: Pass
FR: Pass

Tested on Android 12, Android 11, and Android 5.

jquick-axway (approved):

lokeshchdhry:
FR Passed.
Works fine on Android 12, 10 & 5.

PR #12987, opened 2021-07-29
fix(ios): rare random crashes on 64-bit devices

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-27875

**Summary:**
- Fixes issue where JS property name generation for storing native objects can have name collision on 64-bit systems.
- Code was wrongly casting a 64-bit pointer to a 32-bit integer and trying to generate a unique property name from it.


garymathews (approved):
CR: **PASS**

hansemannn:
@jquick-axway Can you maybe include the changes from @garymathews as well? Then we could cherry-pick these into 10_0_X already and go with it :-)

hansemannn:
We use this in production since a few weeks now and it feels stable. So it'd be great to see it merged!

PR #12988, opened 2021-07-29
fix(ios): properly map user activity for cold starts

### LINKED ISSUE #12944: App crashes on cold launch with Ti.App.getArguments() if opened from App Clip
If an App Clip has been added to an app, and the user opens the App Clip via a QR code etc, the normal process is for either a) the App Clip runs independently if the host app isn't installed or b) the host app is launched if installed.

In the case of b), if the app is running this works and the app clip opens the app and getArguments() can be used to get the resultant data from the app clip.

If however the app is closed, when the user opens from the app clip the app opens and immediately crashes when getArguments is called. The error is:

```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSCFString count]: unrecognized selector sent to instance 0x10981d400'
First throw call stack:
(0x19a802754 0x1af2c97a8 0x19a705c3c 0x19a8052ac 0x19a8075b0 0x103921d2c 0x103921c14 0x19bbc8d4c 0x19a7a1cb0 0x19a7a3284 0x1038e3520 0x1039211d4 0x19bbd5458 0x19a77d22c 0x19a77ce28 0x19a77c278 0x19a77602c 0x19a775360 0x1b1db3734 0x19d1f0584 0x19d1f5df4 0x10252e550 0x19a431cf8)
```

Expected behaviour would be for the app to cold start from an App Clip and getArguments() should return the payload without the app crashing.

hansemannn:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-28513
- https://jira.appcelerator.org/browse/TIMOB-28514

**Note:**
Also fixes https://github.com/appcelerator/titanium_mobile/issues/12944

**Summary:**
The reason is that on a cold start, the user activity contains some native values that cannot be bridged into an `NSDictionary` by default. Just like for other instances (like `CLLocation` or `UIShortcutItem`), we should bridge it to JS-accessible types.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
@@ -350,6 +350,36 @@ TI_INLINE void waitForMemoryPanicCleared(void); //WARNING: This must never be ru
   UILocalNotification *_localNotification = [launchOptions objectForKey:UIApplicationLaunchOptionsLocalNotificationKey];
   NSNumber *launchedLocation = [launchOptions objectForKey:UIApplicationLaunchOptionsLocationKey];
   UIApplicationShortcutItem *shortcut = [launchOptions objectForKey:UIApplicationLaunchOptionsShortcutItemKey];
+  NSDictionary *userActivityDictionary = launchOptions[UIApplicationLaunchOptionsUserActivityDictionaryKey];
+
+  // Map user activity if exists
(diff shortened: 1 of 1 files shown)


jquick-axway (approved):
LGTM

ewanharris (approved):
FR pass. When calling Ti.App.arguments/Ti.App.getArguments after the app has cold launched via a Universal Link there is no longer a crash

PR #12989, opened 2021-07-30
fix(android): views added to ScrollableView can be lost upon window open

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28503

**Summary:**
- Views added to `ScrollableView` after creation, but before its window has been opened, can be all be lost.
- Issue is "views" property is null/empty and the `TiUIScrollableView.processProperties()` method ended up clearing all views that were added via the `ScrollableViewProxy` class' `setViews()` and `addView()` methods.
- This is a regression as of Titanium 10.0.1.


garymathews (approved):
CR: **PASS**

ewanharris (approved):
FR pass. Views added to a ScrollableView after creation but before Window open are now visible (test code in ticket)

PR #12996, opened 2021-08-03
fix(ios): create headers directory if it doesn't exist, write a keep file to ensure it persists

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28511

As the headers directory can contain no files it can be excluded from source control and then lead
to errors during the build process that aren't obvious on how to fix. So add a step to the module
build to ensure that newly built modules always have a file there and a step in the app build
framework detection to ensure modules that are missing it don't cause the error.


jquick-axway (approved):
LGTM

hansemannn:
This causes a build error for me:
```
[TRACE] error: Multiple commands produce '<project/build/iphone/build/Products/Release-iphoneos/include/.keep':
[TRACE] 1) Command: ProcessXCFramework <project>/modules/iphone/facebook/10.0.0/facebook.xcframework <project>/build/iphone/build/Products/Release-iphoneos/include/.keep ios
...
```

ewanharris:
@hansemannn, I've added a change that uses the framework name in the file we write to try and ensure that it stays unique

hansemannn:
Thanks! Could you reproduce our issue? It should already be visible in a project with some more modules

ewanharris:
Yup, it was down to multiple modules having the `.keep` file, so using a unique name avoids that

hansemannn:
Great, will try again this week

ewanharris:
You'll need to manually clear out any of the `.keep` files yourself as I didn't add any handling to go from `.keep`->`.<moduleid>-keep` as this change hasn't landed in public builds yet. I think the error message you posted the snippet from should list all the modules that include that file.

hansemannn:
@ewanharris This does not seem to work for manually (locally) created `Headers` directories. To migrate those to be source-control-safe, could you maybe extend the fix to handle that use case as well?

ewanharris:
@hansemannn, I'd kind of intentionally not handled that case in app builds. If the `Headers` directory exists then xcodebuild should be happy as the folder existing is all that's necessary, just checking for the existence avoids the extra file I/O we'd need to determine if we need to write the keep file.

If there's nothing in that directory, then yes it wont be checked into source control, but when someone checks the project out and runs a build it should be automatically fixed for them and future developers.

The module build has the extra check because it's trying to solve the problem upfront rather than waiting for it to exist in app builds.

hansemannn:
Good point! Then it should be fine now :-)

PR #13000, opened 2021-08-06
fix(android): card view "touchFeedbackColor" property is ignored

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28517

**Summary:**
- Added support for "touchFeedbackColor" to `Ti.UI.Android.CardView`.
  * Before Titanium 10.0.0, card view did not support touch feedback at all.
  * As of Titanium 10.0.0, "touchFeedback" is supported, but ripple color was alway gray and could not be changed via this property.
- Optimized native view creation.
  * It was wrongly being created twice since `processProperties()` method was called twice.
  * Native view is now created once via constructor and `processProperties()` is only called once.

**Test:**
2. Tap on the card view.
3. Verify the touch ripple effect is "yellow".
4. Uncomment the following line:  `// touchFeedback: false,`
5. Build and run on Android.
6. Tap on the card view.
7. Verify a touch ripple effect does **NOT** happen.
8. Verify the following log message does **NOT** appear.
`MaterialCardView: Setting a custom background is not supported.`

```javascript
const cardView = Ti.UI.Android.createCardView({
	elevation: 20,
//	touchFeedback: false,
	touchFeedbackColor: "yellow",
	top: "20dp",
	left: "20dp",
	bottom: "20dp",
	right: "20dp",
});
cardView.add(Ti.UI.createLabel({
	text: 'Tap Me',
	touchEnabled: false,
}));
window.add(cardView);
window.open();
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
@@ -18,20 +18,23 @@ import org.appcelerator.titanium.view.TiUIView;
 import org.appcelerator.titanium.view.TiCompositeLayout.LayoutArrangement;
 
 import android.content.Context;
+import android.content.res.ColorStateList;
+import android.graphics.Color;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
+import androidx.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


garymathews (approved):
CR: **PASS**

ewanharris (approved):
FR Pass. Able to set touchFeedbackColor on CardView

PR #13026, opened 2021-08-16
feat(android): add icon splash screen support

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28473

**Summary:**
- Added support for icon-only splash screens.
  * Will show app icon within a white circle on Android 6.0 and higher.
  * If "android:roundIcon" is assigned, will be used on Android 8.0 and higher.
  * Android 5.x only supports unscaled "android:icon".
- Made "default.png"/"background.png" splash image support **optional**.
  * If **NOT** provided, then icon splash will be used on all Android OS versions.
  * If provided, then default PNG will be used on all OS versions except Android 12+.
- Note that Android 12 forces all apps to use an icon splash screen. You cannot opt-out.
- Updated Titanium SDK test app to "target" Android 12.

---
**KitchenSink Test:**
_(Tests icon-only splash since project doesn't include "default.png" splash.)_
1. Build [kitchensink-v2](https://github.com/appcelerator/kitchensink-v2) on Android 12.
2. Set up device for "light" theme.
3. Launch app and verify app icon is shown on a white background.
4. Set up device for "dark" theme.
5. Launch app and verify app icon is shown within a white circle over a dark background.
6. Repeat steps 2-4 on Android 11. _(In this case Titanium replicates an Android 12 style splash.)_
7. Build and run on Android 5.
8. Verify app icon is displayed on screen. _(Icon might be larger than previous tests since it's not scaled.)_

---
**Hyperloop Test:**
_(Tests legacy default/background PNG and "roundIcon" support.)_
1. Build [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples) on Android 12.
2. Set up device for "light" theme.
3. Launch app and verify app icon is shown on a white background. _(Icon will be bigger than KitchenSink.)_
4. Set up device for "dark" theme.
5. Launch app and verify app icon is shown on a dark background. No white circle.
6. Launch app on Android 11 or older and verify legacy [background.9.png](https://github.com/appcelerator/hyperloop-examples/blob/master/app/platform/android/res/drawable-mdpi/background.9.png) splash image is shown instead of icon.


hansemannn:
Quick question: Does this handle 9-patch images? Because right now, the 9-patch image is shown prior to the circle app icon on Android 12. We do use the `roundIcon` already, so we may need to find a way to only use 9-patch images for Android < 12.

jquick-axway:
@hansemannn, with this PR, all "default.png"/"background.png" splash image files (like your 9-patch) will **NOT** be used on Android 12 and higher. Only your "roundIcon" will be used on Android 12. So, no more double splash on app start up.

On Android 11 and older, your 9-patch splash PNG will still be used just like before. Just note that with this PR makes this PNG "optional". Meaning if you remove all of your "background.9.png" image files, then your "roundIcon" will be used for the splash on Android 8-12... and your app icon will be used on Android 5-7 (because older OS versions do not support adaptive icons). This way an app can have a consistent splash on different OS versions.

hansemannn:
Sounds perfect, thanks!

garymathews (approved):
CR: **PASS**

lokeshchdhry:
FR Passed.

PR #13029, opened 2021-08-17
fix(ios): allow apps with non-arm64 modules to be built on Apple Silicon

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28525

This pull request basically just replaces the way the process architecture is detected. It also improves the error message for non-arm64 modules to make it easier to resolve them one after another.


iphone/cli/commands/_build.js
```diff
@@ -7142,6 +7142,23 @@ function sha1(value) {
...
+	return new Promise((resolve, reject) => {
+		// eslint-disable-next-line security/detect-child-process
+		const exec = require('child_process').exec;
+
+		exec('uname -m', function (error, stdout) {
+			if (error) {
+				reject(error);
+				return;
+			}
+			resolve(stdout);
```
ewanharris, line 7157: I think `stdout` needs to be trimmed here? On my machine `child_process.exec('uname -m', (err, stdout) => console.log(stdout === 'x86_64'))` returns false but `child_process.exec('uname -m', (err, stdout) => console.log(stdout.trim() === 'x86_64'))` returns true and I assume the same is true for arm64 machines

ewanharris (changes requested):

iphone/cli/commands/_build.js
hansemannn: Interesting. Good point!

hansemannn:
@ewanharris Cool?

ewanharris (approved):
LGTM! (although I don't have a shiny Apple Silicon device so can't FR)

hansemannn:
May make sense to combine with #13032, but I can also rebase the other one ones this is merged

ewanharris (approved):
FR pass (confirmed with Hans that this works on an Apple Silicon mac)

jasonkneen:
Is this now in 10.1.0?

ewanharris:
@jasonkneen, yes this shipped in 10.1.0

jasonkneen:
Awesome. Thanks.

—
Jason

PR #13031, opened 2021-08-20
build: specify sha length when calling rev-parse

ewanharris:
In actions this only returns a 7 character string, but the tests expect a 10 character string. It
probably doesn't matter, but best be safe.

This is totally inconsequential (although the tests assert a length of 10), but I figured it's best to maintain the behaviour _just in case_


cb1kenobi (approved):
LGTM!

PR #13032, opened 2021-08-21
fix(ios): be able to build Titanium project from Xcode again

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-26807

Open issues:

- [x] Production builds fail to compile with the following error (only when including some legacy modules):
```
[TRACE] error: Build input file cannot be found: '<project>/build/iphone/Build/Intermediates.noindex/
ArchiveIntermediates/MyApp/InstallationBuildProductsLocation/Applications/MyApp.app/MyApp' (in target 'MyApp'
from project 'MyApp')
```
It seems like there is some kind of static file reference leftover in the `_build.js` that causes this. Can someone help here?

✅ Building the source from Xcode works again (without wild copying and re-referencing files / build settings)
✅ Incremental builds still work
✅ This also fixes the annoying ARM64 error where one had to manually add "arm64" to the "Excluded Architectures" build setting. The reason was that it was added to the `xcodebuild` command instead of the build setting, so it could not be properly reused from Xcode.


iphone/cli/commands/_build.js
```diff
@@ -4856,6 +4874,11 @@ iOSBuilder.prototype.copyTitaniumiOSFiles = function copyTitaniumiOSFiles() {
 		path.join(this.platformPath, 'iphone', 'Titanium.xcodeproj', 'xcshareddata', 'xcschemes', 'Titanium.xcscheme'),
 		path.join(this.buildDir, this.tiapp.name + '.xcodeproj', 'xcshareddata', 'xcschemes', name + '.xcscheme')
 	);
+	copyAndReplaceFile.call(
+		this,
+		path.join(this.platformPath, 'iphone', 'Titanium.xcodeproj', 'xcshareddata', 'WorkspaceSettings.xcsettings'),
+		path.join(this.buildDir, this.tiapp.name + '.xcodeproj', 'project.xcworkspace', 'xcuserdata', `${require('os').userInfo().username}.xcuserdatad`, 'WorkspaceSettings.xcsettings')
+	);
```
hansemannn, line 4881: This is actually where the magic happens

jasonkneen:
Will this be available soon?

hansemannn:
I haven't had the chance to finish the PR, yet and have to fix the remaining production builr issue

jasonkneen:
> I haven't had the chance to finish the PR, yet and have to fix the remaining production builr issue

Awesome -- thanks for all your efforts!

hansemannn:
Hi everyone! This one is ready for review!

iphone/cli/commands/_build.js
```diff
@@ -2265,6 +2270,15 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 						cli.scanHooks(path.join(module.modulePath, 'hooks'));
 					}, this);
 
+					// Exclude arm64 architecture from simulator build in XCode 12+ - TIMOB-28042
+					if (this.legacyModules.size > 0 && parseFloat(this.xcodeEnv.version) >= 12.0) {
+						if (await processArchitecture() === 'arm64') {
```
cb1kenobi, line 2275: Just reference `process.arch`. It will be set to `"arm64"` on Apple Silicon. No need to run `exec()` and no need to make this callback async.

cb1kenobi (changes requested):

hansemannn:
@cb1kenobi Done, thank you! I noticed that the `error: Build input file cannot be found:` error is still happening with _some_ iOS modules. But after repackaging with the latest tooling, it does work. So I guess I'll have to do some more research to offer either an autoatic migration or better error logging to let devs know whats wrong.

cb1kenobi (approved):
CR'd, looks good! Thanks!

caspahouzer:
Can this be merged? Or are there any obstructions?

hansemannn:
There is still one open issue which I couldn't solve so far (see the open To-do in the first post)

hansemannn:
Ready for review! @janvennemann and/or @cb1kenobi, can you take a look? I disabled the project generation for prod builds for now, so it doesn't interfere with sensitive production settings.

iphone/cli/commands/_build.js
```diff
@@ -2260,6 +2265,15 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 						cli.scanHooks(path.join(module.modulePath, 'hooks'));
 					}, this);
 
+					// Exclude arm64 architecture from simulator build in XCode 12+ - TIMOB-28042
+					if (this.legacyModules.size > 0 && appc.version.gte(this.xcodeEnv.version, '12.0.0')) {
+						if (process.arch === 'arm64') {
```
cb1kenobi: **Legacy-module arm64 check isn't gated on target or deploy type** — severity: high

This block runs in `validateModules`, which executes for every target and deploy type. On an Apple Silicon Mac (`process.arch === 'arm64'`), any app using a legacy module now hard-fails validation — including `--target device`, `dist-adhoc` and `dist-appstore` builds, which are unaffected by the simulator-slice problem (those modules do ship arm64 *device* slices). Developers on M-series Macs go from shipping fine to being unable to build at all for device/production.

Simplest fix: gate the block on `this.target === 'simulator'` (matching the comment "Exclude arm64 architecture from simulator build"), e.g. `if (this.target === 'simulator' && this.legacyModules.size > 0 && appc.version.gte(...))`.

—
Reviewed 4fb189ab

iphone/cli/commands/_build.js
```diff
@@ -3234,6 +3248,10 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 		},
 		legacySwift = version.lt(this.xcodeEnv.version, '8.0.0');
 
+	if (this.excludeARM64 && this.deployType !== 'production') {
+		buildSettings['EXCLUDED_ARCHS'] = 'arm64';
```
cb1kenobi: **`EXCLUDED_ARCHS = arm64` is written into project-wide build settings** — severity: high

`buildSettings` here is the project-level settings dict applied to all configurations and destinations, and the only guard is `deployType !== 'production'`. A development or test build to a physical device (arm64-only since iPhone 5s) then has every architecture excluded, so the link step produces no valid slice and the build fails with "no architectures to compile for". The setting is also persisted into the generated `.xcodeproj`, breaking subsequent device builds started from Xcode.

Simplest fix: also require a simulator target — `if (this.excludeARM64 && this.target === 'simulator' && this.deployType !== 'production')` — or scope it to the simulator SDK (`EXCLUDED_ARCHS[sdk=iphonesimulator*]`) so device builds keep arm64.

—
Reviewed 4fb189ab

cb1kenobi (commented):
The new arm64 handling is meant for simulator builds but is applied unconditionally. On Apple Silicon, any app with a legacy module now fails validation even for device and production builds, and EXCLUDED_ARCHS=arm64 is written into project-wide build settings so non-production device builds lose their only valid architecture. Both need to be gated on target === 'simulator' (or scoped with [sdk=iphonesimulator*]).

—
Reviewed 4fb189ab

iphone/cli/commands/_build.js
hansemannn: Fixed in d1e2599 - the check is now gated on `this.target === 'simulator'`, so device and dist builds on Apple Silicon are unaffected.

iphone/cli/commands/_build.js
hansemannn: Fixed in d1e2599 - the setting is now written as `EXCLUDED_ARCHS[sdk=iphonesimulator*]` (same pattern as the existing `CODE_SIGN_IDENTITY[sdk=macosx*]` key), so device builds from the CLI or from the generated Xcode project keep arm64.

hansemannn:
Merged current `main` (resolved the `_build.js` conflict by re-applying the PR's changes on top of the ESM refactor) and addressed the review (#pullrequestreview-5283521835):

- Legacy-module arm64 check in `validate()` is now gated on `target === 'simulator'`.
- `EXCLUDED_ARCHS` is scoped to `[sdk=iphonesimulator*]` in the generated project, so device builds keep arm64 even when started from Xcode.

@cb1kenobi - ready for another look.

cb1kenobi (commented):
The two high-severity issues from the earlier review are properly fixed in this commit. The arm64 legacy-module check is now limited to simulator targets, and the architecture exclusion is scoped to the simulator SDK so device and production builds keep arm64. The Xcode workspace settings file is still only written for non-production builds. No new blocking problems on the changed lines.

—
Reviewed d1e2599c

cb1kenobi (commented):
Both high-severity issues from the earlier review are fixed at this commit. The legacy-module arm64 check now only runs for simulator targets, and the architecture exclusion is scoped to the simulator SDK so device and production builds keep arm64. The Xcode workspace settings file is still written only for non-production builds. No new blocking problems on the changed lines.

—
Reviewed d1e2599c

PR #13038, opened 2021-08-23
Improve reporting of failed test runs

ewanharris:
* Adds a danger check for all of the test reports and fails the build if any are missing (this does mean if we remove or rename one we have to update the check)
* Adds a flag to allow us to distinguish between Android main and Android 5.0 test reports
* Moves the junit danger check to only _warn_ about failed tests

This means that we can make danger the only required check, and remove required status on the Jenkins check. So:

* All builds pass and we have test failures, anyone with merge access can merge
* Build(s) fail (either the app build, or the SDK build), danger will block the merge


garymathews (approved):
LGTM!

PR #13058, opened 2021-09-07
test: remove axway and appcelerator domains in tests

ewanharris:
It seems like axway.com and appcelerator.com domains are taking a while to resolve on the Jenkins nodes, so lets just tear any reference to them out. Works locally, 🤞 for Jenkins


ewanharris:
Of course example.com is blocked, switching out

jquick-axway (approved):
CR: Pass

PR #13061, opened 2021-09-11
fix(ios): fix navigation bar from jumping on iOS 15+

hansemannn:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28488

```js
const btn = Ti.UI.createButton({ title: 'Open window' });

	const navWin = Ti.UI.createNavigationWindow({
		window: Ti.UI.createWindow({
                        title: 'Test',
			backgroundColor: 'white'
		})
	});

	navWin.open({ modal: true });
});

window.add(btn);
window.open();
```

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
@@ -759,6 +759,9 @@
   if (hidesBackButton) {
     [self setHidesBackButton:hidesBackButton];
   }
+  if (isModal && opening) {
+    [self windowDidOpen];
+  }
 
   [self willShow];
 }
@@ -771,9 +774,6 @@
 }
 - (void)viewDidAppear:(BOOL)animated
 {
-  if (isModal && opening) {
-    [self windowDidOpen];
-  }
   if (controller != nil && !self.isManaged) {
     [self gainFocus];
   }


lokeshchdhry:
FR Passed.

iphone/TitaniumKit/TitaniumKit/Sources/API/TiWindowProxy.m
```diff
@@ -759,6 +759,9 @@ - (void)viewWillAppear:(BOOL)animated
   if (hidesBackButton) {
     [self setHidesBackButton:hidesBackButton];
   }
+  if (isModal && opening) {
+    [self windowDidOpen];
+  }
```
jquick-axway, line 764: I've tested this an it works. However, moving the `windowDidOpen()` call here causes the `Ti.UI.Window` object's "open" event to be fired earlier than it used to. I'm not sure if this has negative consequences or not, but an alternative solution would be to revert your code change here and instead change the [TiUINavigationWindowProxy.m](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiUINavigationWindowProxy.m) code to set up the navigation controller via the `viewWillAppear()` method, which is what your code change does. I've tested the below and it works for me.
```objc
// Delete this old code since it doesn't set up nav controller in time.
#if 0
- (void)windowDidOpen
{
  // Set parent for navigation controller
  if (navController) {
    UIViewController *parentController = [self windowHoldingController];
    [parentController addChildViewController:navController];
    [navController didMoveToParentViewController:parentController];
  }
  [super windowDidOpen];
}

- (void)viewWillAppear:(BOOL)animated
{
  // Move the above code here.
  if (navController) {
    UIViewController *parentController = [self windowHoldingController];
    [parentController addChildViewController:navController];
    [navController didMoveToParentViewController:parentController];
  }

  if ([self viewAttached]) {
    [navController viewWillAppear:animated];
  }
  [super viewWillAppear:animated];
}
#endif
```

jquick-axway: Or we can move that block of code to the [TiUINavigationWindowProxy.m](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiUINavigationWindowProxy.m) class' `windowWillOpen()` method. I've tested this and it work as well. The key thing is the nav controller needs to be set up before the window is opened.

hansemannn: Fine with both! Natively, the `open` equivalent is `viewDidAppear`, so that should be kept as-is.

hansemannn:
Superseded by #13067

PR #13069, opened 2021-09-15
fix(android): blob fails to read WebP image info

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28535

**Summary:**
- When loading a WebP image via a `Ti.Blob` then its width/height and mimeType properties return the wrong values and the blob's `image*()` methods wrongly return `null`.
- Caused by Java `URLConnection.guessContentTypeFromStream()` method which wrongly returned "audio/x-wav" instead of "image/webp", which caused the blob to think it wasn't an image.

**Test:**
1. Copy "TiLogoOpaque.webp" from [TIMOB-28535](https://jira.appcelerator.org/browse/TIMOB-28535) to project folder: `./Resources/assets/images`
2. Use "app.js" code shown in ticket [TIMOB-28535](https://jira.appcelerator.org/browse/TIMOB-28535).
3. Build and run on Android.
4. Verify a red Titanium logo with rounded corners appears in app. _(Use to show nothing.)_
5. Verify app's "Image Size" label reads: `1024x1024`
6. Verify app's "Mime-Type" label reads: `image/webp`


garymathews:
@jquick-axway Looks like `.image with Ti.Blob WebP` test is failing on Android 5.0

jquick-axway:
@garymathews , the WebP image loads fine on Android 5.0, but the window's "postlayout" event was happening before the ImageView resized itself to fit its image. I re-worked the event handling to make the unit test more reliable.

garymathews (approved):
CR: **PASS**

PR #13077, opened 2021-09-20
fix(android): auto-scaling mode must stretch ImageView if both width/height set

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28537

**Summary:**
- When `ImageView` properties "width" and "height" are set, it should stretch disproportionally by default when using "scalingMode" `Ti.Media.IMAGE_SCALING_AUTO` which is the default. Instead, it is wrongly letterbox scaling image. This is for backward compatibility with Titanium 10.0.2 and older SDK versions.
- Needs to match what we document [here](https://titaniumsdk.com/api/titanium/media.html#image_scaling_auto).
- Regression as of Titanium 10.1.0.RC.

**Test:**
2. Verify image is stretch width wise as shown in "Actual Result" in [TIMOB-28537](https://jira.appcelerator.org/browse/TIMOB-28537).
3. Comment out the line: `//  width: Ti.UI.FILL,`
4. Rebuild and re-run the app.
5. Verify image is proportionally scaled as shown in "Result" in [TIMOB-28537](https://jira.appcelerator.org/browse/TIMOB-28537). _(In this case, this is the correct behavior because only "width" is defined.)_

```javascript
win.add(Ti.UI.createImageView({
	image: "https://raw.githubusercontent.com/appcelerator/titanium_mobile/master/templates/app/angular-default/template/DefaultIcon.png",
	width: Ti.UI.FILL,
	height: 30,
}));
win.open();
```


garymathews (approved):
CR: **PASS**
FT: **PASS**

lokeshchdhry:
FR Passed.

PR #13095, opened 2021-09-30
feat(android, ios): continuous update for ListView scrolling event

m1ga:
**JIRA:** https://jira.appcelerator.org/browse/AC-6722

When `continuousUpdate` is enabled the `scrolling` event will fire for every visibleItem/Section change. 

**Test:**
* start example app
* start scrolling
* in the top right corner (or log) you'll see the current index/section
* click "change" button to enable/disable `continuousUpdate`

disabled or with 10.1.1GA you only see an update when you change scroll direction.

**Example**
```js
var toggle = true;
var sections = [];
	title: "change",
	bottom: 20
});
	top: 10,
	right: 10,
	width: Ti.UI.SIZE,
	height: Ti.UI.SIZE
})
	height: Ti.UI.FILL,
	width: Ti.UI.FILL,
	continuousUpdate: true
});

	toggle = !toggle;
	listView.continuousUpdate = toggle
	console.log(listView.continuousUpdate);
})

for (var s = 0; s < 5; s++) {
		headerTitle: 'Section ' + s
	});
	var set = [];
		set.push({
			properties: {
				title: 'Item 0 ' + i
			}
		})
	}
	section.setItems(set);
	sections.push(section);
}

listView.sections = sections;

	console.log("Item", e.firstVisibleItemIndex, e.firstVisibleSectionIndex);
})
win.open();
```


jquick-axway:
@m1ga , did you run clang-format on the Java code?

You don't need to do that anymore. We use Java "checkstyles" which will trigger a build error for anything that violates our coding guidelines and explain what the issue is. Also, Android Studio's build log will show a clickable link that will take you directly to the line of code that violated the rule.

m1ga:
Thanks @jquick-axway. I saw those warnings e.g. line too long and then I thought I'll run clang to automatically fix those warnings. Didn't know that it's not needed anymore. Was a quick automatic way of fixing all issues. 
I've fixed the unneeded changes

apidoc/Titanium/UI/ListView.yml
```diff
@@ -902,6 +902,15 @@ properties:
     platforms: [android, iphone, ipad, macos]
     since: 3.2.0
 
+  - name: continuousUpdate
+    summary: Determines if the scrolling event should fire every time there is a new visible item.
+    description: |
+        When setting `continuousUpdate` to `true` the [scrolling](Titanium.UI.ListView.scrolling) event will everytime a new item is visible.
```
jquick-axway, line 908: Would you mind documenting what the use-case for enabling this is in the description please?
I'm curious why you want this feature. And I'm willing to help add it to iOS.

Also, the following sentence is awkward.
`...event will everytime a new item is visible.`
I think the following is better. _(Not trying to be picky.)_
`...event will fire every time a new item is scrolled in.`

m1ga: Sure, will update the docs tomorrow. I need to know where I'm at to change a menu e.g. which element just came into view. Currently I'm using markers but they only fire once so I have to add them again when I've hit one. 
Especially iOS had problems when I've add the markers and it filled the list after adding them. It did fire the marker event for all markers already because it was triggered before the actual data was visible and my 2nd (normally offscreen marker) was visible.
Having the scoll event just telling me the whole time which element is visible (fistVisibleItem/section) I can easily use that for lazy loading, dynamically switching menus (indicator where you are) and other triggers (e.g. changing a headline)

m1ga:
@hansemannn could you have a look at this one.

build/.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
```
hansemannn, line 3: This file should be removed

m1ga: sorry, removed it.

hansemannn:
The code looks fine - all additions to the code are guarded / in separate parts. Same for iOS. Have you tested iOS properly as well? And for custom cells?

hansemannn:
In any case, I would prioritize bug fixes first at this stage and do features / improvements after that, with a prio on features that have full parity.

m1ga:
tested in on iOS too with the example and an app with a ListView with 4 templates, 5 sections. And since its an older PR the whole iOS/Android test-suite was done too.

The new parts are hidden behind a `false` by default property so it won't affect any existing apps. Only if you set `continuousUpdate:true` it will make any difference. But it can wait for other PRs :smile:

hansemannn:
I feel like it's not the most frequently asked-for feature, so let's approve it already, but merge after the fixes have all been merged.

hansemannn (approved):
iOS looks good!

hansemannn:
@m1ga Happy to take this before 12, e.g. 11.1.0!

m1ga:
> @m1ga Happy to take this before 12, e.g. 11.1.0!

ah nice! Wasn't sure :smile: It's covered by a test case too so from a functional state point quick to test/merge

hansemannn (approved):

PR #13097, opened 2021-09-30
chore(android): update v8 to 9.4.146.16

garymathews:
- Update V8 to `9.4.146.16` https://github.com/appcelerator/v8_titanium/tree/9.4-lkgr
  - Maintains backwards compatibility with native modules

##### TEST CASE
- Test with kitchensink-v2
- Various applications that use native modules
- All supported architectures need to be tested, `arm64, armv7, x86_64, x86`

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28539)


jquick-axway (approved):
CR: Pass
FR: Pass

Tested the following in Android 12 emulator and Android 11 Pixel 3a:
- kitchensink-v2
- hyperloop-examples
- Titanium SDK mocha unit tests
- Customer app with several legacy modules

hansemannn:
I'm afraid we have to revert this due to #13395 or find a proper patch. Not sure who is able to do that right now :(

Issue #13053, opened 2021-09-03
Apple M1 Run on iphone : Can't sign application


lionelbouth:
### Describe the bug
Can't run application on iphone (UDID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX ) connect or building Ad-hoc applications on Apple Mac Silicon M1 Big Sur xcode 12.5

Apple developpement certificate is OK 
Provisionning profile including Mac + my iphone is OK

Always having error 

note: Build preparation complete
[TRACE] error: Provisioning profile "MyApp" doesn't include the currently selected device "iMac of me" (identifier ZZZZZZZZZZZZZZZZZZZZZ). (in target 'MyApp' from project 'MyApp')

ZZZZZZZZZZZZZZZZZZZZZ is de UUID of the MAC M1 which is building de app.

But i dont build the app for the Mac M1 ZZZZZZZZZZZZZZZZZZZZZ  but for iphone XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX 

Moreover, ZZZZZZZZZZZZZZZZZZZZZ  is the uuid of the MAC M1 so I can't add it on apple developper devices as ios device

### To Reproduce

appc run -p ios -T device XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX --log-level trace

### Expected behavior
Runing app on iphone with uuid : XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX 

### Environment
Operating System
  Name                        = macOS
  Version                     = 11.5.2
  Architecture                = 64bit
  # CPUs                      = 8
  Memory                      = 16.0GB

Node.js
  Node.js Version             = 12.22.3
  npm Version                 = 6.14.13

Titanium CLI
  CLI Version                 = 5.3.2
  node-appc Version           = 1.1.2

Titanium SDKs
  9.2.1.GA
    Version                   = 9.2.1
    Install Location          = /Users/devmac/Library/Application Support/Titanium/mobilesdk/osx/9.2.1.GA
    Platforms                 = iphone, android
    git Hash                  = 19122961e3
    git Timestamp             = 10/5/2020 22:59
    node-appc Version         = 1.1.1
  10.0.2.GA
    Version                   = 10.0.2
    Install Location          = /Users/devmac/Library/Application Support/Titanium/mobilesdk/osx/10.0.2.GA
    Platforms                 = iphone, android
[...]

PR #13098, opened 2021-10-01
fix(ios): use correct iOS destination platform during build

lionelbouth:
https://github.com/appcelerator/titanium_mobile/discussions/12602

https://github.com/appcelerator/titanium_mobile/issues/13053

JIRA: https://jira.appcelerator.org/browse/TIMOB-28545


m1ga:
awesome, @lionelbouth :+1: 
You could change your title to: `fix(ios): correct iOS destination platform during build` or something like that so it will look nicer.

ewanharris:
Thanks @lionelbouth! It looks like this maybe caused the `macos` target to fail on a build. We'll investigate and push any required fixes to your branch

ewanharris:
@lionelbouth I just pushed a commit to only apply the `-destination generic/platform=iOS` when targeting the `device` target type which fixes the `macos` target issue. Could you try applying that change and see if it works for you on an M1 mac?

lionelbouth:
Tried your commit on 9.2.1.GA,  on my MAC M1 : not working.

ewanharris:
@lionelbouth when you say not working, what are you targeting (device or another target?) and what is the error you are receiving?

lionelbouth:
Sorry my bad.
I tried to build ipa whith Ad hoc provisioning profiles ...
IMO, the correct answer should be **if( this.target ==='device' ||  this.target ==='dist-adhoc' )**

iphone/cli/commands/_build.js
```diff
@@ -7102,6 +7102,8 @@ iOSBuilder.prototype.invokeXcodeBuild = async function invokeXcodeBuild(next) {
 			this.logger.warn(`The app is using native modules (${Array.from(this.legacyModules)}) that do not support arm64 simulators, we will exclude arm64. This may fail if you're on an arm64 Apple Silicon device.`);
 			args.push('EXCLUDED_ARCHS=arm64');
 		}
+	} else if ( this.target === 'device' || this.target === 'dist-adhoc' ){
```
ewanharris, line 7105: Does this also need to include the `dist-appstore` target or is the error not seen when targeting that?

ewanharris: Also, eslint will complain about the spaces in the brackets. If you have this checked out locally you can run `npm run format:js` to fix them and any other issues automatically, you might need to run `npm ci` first.

lionelbouth: No need to include dist-appstore. 
Probably because appstore provisioning profiles don't include devices

m1ga:
btw this also fixed
```

  --- xcodebuild: WARNING: Using the first of multiple matching destinations:
    { platform:iOS, id:dvtdevice-DVTiPhonePlaceholder-iphoneos:placeholder, name:Any iOS Device }
    { platform:macOS, arch:arm64, variant:Designed for [iPad,iPhone], id:00008103-001444810A31001E }
    { platform:iOS Simulator, id:dvtdevice-DVTiOSDeviceSimulatorPlaceholder-iphonesimulator:placeholder, name:Any iOS Simulator Device }
    { platform:iOS Simulator, id:EDAD5055-209E-43BF-A48F-F0A2D8D26F43, OS:15.0, name:iPad (9th generation) }
    { platform:iOS Simulator, id:6D279D6B-91AD-43A0-ADEA-9C19BDE7F867, OS:15.0, name:iPad Air (4th generation) }
    { platform:iOS Simulator, id:CB7171EA-2225-4747-9F52-701A93AAEB11, OS:15.0, name:iPad Pro (9.7-inch) }
    { platform:iOS Simulator, id:5142D23C-2C4B-4A12-B1D7-0F50CC53106F, OS:15.0, name:iPad Pro (11-inch) (3rd generation) }
    ....

```
for me which suddenly appeared. Not sure why, running Xcode 13 for a while now and using an old Macbook Air

sgtcoolguy:
I can confirm this fixes #12602 for me on my M1 Mac mini. Building for device would fail before complaining the host machine was not on the provisioning profile. With this fix, it no longer does.

ewanharris (approved):
CR Pass!

Thanks @lionelbouth!

PR #13099, opened 2021-10-01
fix(ios): dont run simulator detection for macos builds

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28544

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -2013,7 +2013,7 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 			},
 
 			function selectDevice(next) {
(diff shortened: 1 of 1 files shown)


iphone/cli/commands/_build.js
```diff
@@ -2013,7 +2013,7 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 			},
 
 			function selectDevice(next) {
-				if (cli.argv.target === 'dist-appstore' || cli.argv.target === 'dist-adhoc') {
+				if (cli.argv.target.startsWith('dist') || cli.argv.target === 'macos') {
```
jquick-axway, line 2016: **Question:** Can `target` ever be `null` or `undefined`?

jquick-axway (approved):
LGTM

PR #13107, opened 2021-10-07
fix(ios): ListView/TableView multiselection events only fired when tapping checkboxes

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28548

**Summary:**
- `ListView` "itemsselected" event only fired when tapping item checkboxes. Not fired when tapping any other part of the item such as the title.
- `TableView` "rowsselected" event only fired when tapping row checkboxes. Not fired when tapping any other part of the row such as the title.
- Issue is item selection/de-selection still occurs, but not event gets fired. This makes it hard to track user selection unless you use a timer. Use-case would be to display a selection count onscreen.

**ListView Test:**
1. Build and run `ListViewSelectionCountTest.js` attached to [TIMOB-28548](https://jira.appcelerator.org/browse/TIMOB-28548) on iOS.
2. Tap on the "Edit" button.
3. Tap on a list item's text and verify selection count increases. _(This did not work before.)_
4. Tap on the selected list item's text to de-select it and verify count decreases. _(This did not work before.)_
5. Tap on a list item's checkbox and verify selection count increases.
6. Tap on a selected list item's checkbox to de-select it and verify count decreases.

**TableView Test:**
1. Build and run `TableViewSelectionCountTest.js` attached to [TIMOB-28548](https://jira.appcelerator.org/browse/TIMOB-28548) on iOS.
2. Tap on the "Edit" button.
3. Tap on a row's text and verify selection count increases. _(This did not work before.)_
4. Tap on the selected row's text to de-select it and verify count decreases. _(This did not work before.)_
5. Tap on a row's checkbox and verify selection count increases.
6. Tap on a selected row's checkbox to de-select it and verify count decreases.


garymathews (approved):
CR: **PASS**

PR #13123, opened 2021-10-15
fix(ios): drag-and-drop text in TextArea can crash on iOS 15

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28556

**Summary:**
- Drag-and-dropping text into a `TextArea` that exceeds the "maxLength" property will crash the app on iOS 15.
- This looks like a bug on Apple's end where they don't check the final string length before updating the cursor/selection position. Probably because they don't expect us to manipulate the string that's been pasted.

**Test:**
1. Build and run the below on iOS 15.
2. Tap and hold the top TextArea's text.
3. Tap on "Select All" in the context menu.
4. Tap and hold selected text until it becomes draggable.
5. Drag text into bottom TextArea.
6. Verify dragged text is copied into TextArea, but trimmed based on maxLength of 20.

```javascript
	layout: "vertical",
	backgroundColor: "white",
});
window.add(Ti.UI.createTextArea({
	value: "Hello World! This is a test.",
	borderWidth: 1,
	top: 100,
	width: "80%",
	height: 80,
}));
window.add(Ti.UI.createTextArea({
	value: "Paste text here.",
	maxLength: 20,
	borderWidth: 1,
	top: 20,
	width: "80%",
	height: 80,
}));
window.open();
```

### DIFF
--- a/iphone/Classes/TiUITextArea.h
+++ b/iphone/Classes/TiUITextArea.h
@@ -18,6 +18,7 @@
 @property (nonatomic, readwrite, assign) BOOL enableCopy;
 
 - (void)setTouchHandler:(TiUIView *)handler;
+- (NSComparisonResult)comparePosition:(UITextPosition *)position toPosition:(UITextPosition *)other;
 
 @end
 
--- a/iphone/Classes/TiUITextArea.m
+++ b/iphone/Classes/TiUITextArea.m
@@ -38,6 +38,18 @@
   touchHandler = handler;
 }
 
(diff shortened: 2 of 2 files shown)


garymathews (approved):
CR: **PASS**

ewanharris (approved):
FR pass. Just want to note that the error below is still logged, but not crash is seen 

```
*** Assertion failure in -[_UITextKitTextPosition compare:], UITextKitTextViewEditingSupport.m:43
```

Issue #12943, opened 2021-07-07
Empty List View Ti SDK 10 Android


riodiep:
### Describe the bug
I'm using alloy and the list view displays normally when I declare it in XML view and load it with dataCollection and dataTransform. However, when I try to manually create sections and items in controller js and add to the listview it shows an empty list view. Issue only occurs on Ti SDK 10 Android version. 

Please see the attachment for the source code
[refer-patient.zip](https://github.com/appcelerator/titanium_mobile/files/6774150/refer-patient.zip)

### Expected behavior
Expected to show the list of phone contacts but it only shows the search bar. The same code work on older version < Ti SDK 10
I tried to log contactList.sections and its items, everything is there in the object but the view didn't show up.

### Environment
Titanium SDK version: 10.0.0.GA
CLI version: 9.0.1
Alloy: latest

m1ga:
One workaround would be to create a template and use that. Then you wouldn't put everything into
```
properties: {
                    title: title,
                    mobile: cleanMobile(contacts[i].phone.mobile),
                    color: "#000000",
                    backgroundColor: "#FFFFFF",
                    height: 50,
                    searchableText: title
                }
```
but in something like
```
title: {text: title},
mobile: {text: cleanMobile(contacts[i].phone.mobile)]
...
```
```

You have far more control over the layout

riodiep:
Thanks Michael for the advice.

Unfortunately it didn't work. But I figured out the issue was with the SearchBar. When I tried adding the search bar (or searchView) both ways in xml or in js controller file, the list view only showed search bar, but when I removed the search bar, it showed the list sections and items as expected. I also tried to set a fix height for search bar but didn't work.

As a result, I specify to use search bar in IOS platform only for now and wait for a fix in the TI SDK.

m1ga:
I use a normal TextField as a SearchBar on Android. Use the onChange event and set the listview.searchText yourself.

riodiep:
Thanks for this workaround.

drauggres:
@riodiep Please reopen. If there is a bug it should be fixed.

m1ga:
Don't think it is a bug in the SDK. I've used your code (removed all the global functions) and it shows all my contacts:
![Screenshot_20210723-103644](https://user-images.githubusercontent.com/4334997/126757416-6a2149dc-3a55-4dc1-a348-7c23fe165513.png)

```javascript
function showPhoneContacts() {
[...]

PR #13124, opened 2021-10-15
fix(android): re-create list upon visibility change

garymathews:
- `ListView.show()` and `ListView.visibility` may not display list items under certain criterias

##### TEST CASE
```JS
	visible: false, // Set visible equal to true and I work
	searchView: Ti.UI.createSearchBar() // Remove the searchView and I work
});
// Comment out the event listener and I work, change me to 'click' and still broke though
	const data = []
	for (let i = 0; i < 5; i++) {
		const row = {
			properties: {
				title: `Foo ${i}`,
				color: "#000000",
				backgroundColor: "#FFFFFF",
				height: 50
			}
		}
			headerTitle: `Test ${i}`,
			items: [ row ]
		});
		data.push(section);
	}
	listView.sections = data;
        // listView.visible = true;
	listView.show();
});
window.add(listView)
window.open();
```

https://github.com/appcelerator/titanium_mobile/issues/12943
[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28557)


jquick-axway (approved):
CR/FR: Pass

PR #13131, opened 2021-10-18
fix(android): prevent multiple RippleDrawable backgrounds

garymathews:
- Prevent multiple `RippleDrawable` backgrounds from being set

##### TEST CASE
```JS
	searchView: Ti.UI.createSearchBar()
});

	const data = []
	for (let i = 0; i < 5; i++) {
		const row = {
			properties: {
				title: `Foo ${i}`,
				color: '#000000',
				backgroundColor: '#FFFFFF'
			}
		}
			headerTitle: `Test ${i}`,
			items: [ row ]
		});
		data.push(section);
	}
	listView.sections = data;
});

window.add(listView)
window.open();
```
- Entering text into search should not make ripple background darker

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-28560)


jquick-axway (approved):
CR/FR: Pass

PR #13132, opened 2021-10-19
feat: declare node 16 support

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28562

It works, so lets add it to the list!


jquick-axway (approved):
LGTM

PR #13134, opened 2021-10-19
fix(android): ListView not applying i18n "textid" and "titleid" properties on off-screen items as of 10.0.1

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28561

**Summary:**
- As of Titanium 10.0.1, there is a regression in ListView where localization properties such as "textid" and "titleid" are ignored when scrolling-in list items and they show no text.
- Caused by the optimizations we made recycling views/proxies. Properties were being correctly copied to scrolled-in proxies, but we need to apply specialized handling for i18n properties where they're supposed to fetch a localized string and update its associated property (ex: "textid" should update "text" property).

**Test:**
Follow the steps-to-reproduce in [TIMOB-28561](https://jira.appcelerator.org/browse/TIMOB-28561) and verify you see rows with labels/buttons named "Row 01" to "Row 80". There should be no empty rows.


garymathews (approved):
CR: **PASS**
FT: **PASS**

PR #13147, opened 2021-10-27
fix(android): menu and toolbar icons to use ActionBar style colors

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28547

**Summary:**
When setting menu/toolbar icons to a resource ID (ie: a vector drawable), the icon should use the theme/style assigned to the ActionBar or Toolbar. Currently, it uses the style colors assigned to the activity window, which won't be the same as the ActionBar if using a DarkActionBar theme.

**Test:**
1. Download https://github.com/appcelerator/kitchensink-v2/pull/59
2. Build and run on Android.
3. Make sure device is using the Light theme.
3. Go to: Views -> ListView (Multi-Select)
4. Verify all top toolbar icons are white. (They used to be gray.)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiToolbar.java
@@ -17,6 +17,7 @@ import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.TiDimension;
 import org.appcelerator.titanium.proxy.TiViewProxy;
 import org.appcelerator.titanium.util.TiColorHelper;
+import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.view.TiCompositeLayout;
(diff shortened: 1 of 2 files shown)


garymathews (approved):
CR: **PASS**
FT: **PASS**

hansemannn:
I wonder if this change caused the `[ERROR] E/com.example.myapp Invalid ID 0x00000000.` error log. I was able to reproduce it with just these few lines of code:
```xml
<Alloy>
	<Window>
		<Menu platform="android">
			<MenuItem title="Test" />
		</Menu>
	</Window>
</Alloy>
```
cc @m1ga, you may have seen the error log as well.

m1ga:
I'll check that :+1:

m1ga:
@hansemannn it's strange. I can't reproduce it at the moment with 10.2.0. Neither with the demo code nor with my app :thinking: I had this for a long time but now I don't see it anymore. I'll give it another try later today

PR #13125, opened 2021-10-16
feat: add "imageIsMask" property to Ti.UI.Button

jquick-axway:
**JIRA:**
- https://jira.appcelerator.org/browse/TIMOB-28558
- https://jira.appcelerator.org/browse/TIMOB-28559

**Summary:**
- [TIMOB-28558](https://jira.appcelerator.org/browse/TIMOB-28558) Added boolean property "imageIsMask" to `Ti.UI.Button`.
  * When set `true`, the button's "image" is tinted using "tintColor" property or default theme if not assigned.
  * When set `false`, the button's "image" is displayed as-is.
  * iOS used to always tint image by default and Android displayed image as-is, which was a parity issue.
  * This property defaults to `true`, matching iOS' old behavior but is a breaking-change on Android.
- [TIMOB-28559](https://jira.appcelerator.org/browse/TIMOB-28559) Add native resource ID support to `Ti.UI.Button` "image" property.
  * Can now use `Ti.App.Android.R.drawable.*` or `Ti.Android.R.drawable.*` IDs.
  * This let you use native vector drawables, which are typically used as icons.

**Test:**
1. Use the below code as the "app.js".
2. Create folder: `/assets/images`
3. Copy `RedLogo.png` and `controls*.png` images to the above folder.
4. Build and run on Android.
5. Verify 1st button's image and title are white in light theme. (Will be black in dark theme.)
6. Verify 2nd button's image and title are green. (The theme doesn't matter.)
7. Verify 3rd button show a red circle with a titanium logo.
8. Verify 4th button shows a white trash can in light theme. (Will be black in dark theme.)
9. Repeat steps 5-7 on iOS.

```javascript
const window = Ti.UI.createWindow({ layout: "vertical" });
window.add(Ti.UI.createButton({
	title: " imageIsMask true",
	image: "/assets/images/controls" + (OS_IOS ? "" : "@3x") + ".png",
	imageIsMask: true,
	top: 80,
}));
window.add(Ti.UI.createButton({
	title: " imageIsMask true (Green)",
	image: "/assets/images/controls" + (OS_IOS ? "" : "@3x") + ".png",
	imageIsMask: true,
	tintColor: "green",
	top: 20,
}));
window.add(Ti.UI.createButton({
	title: " imageIsMask false",
	image: "/assets/images/RedLogo.png",
	imageIsMask: false,
	top: 20,
}));
if (OS_ANDROID) {
	window.add(Ti.UI.createButton({
		title: "R.drawable",
[...]

PR #13148, opened 2021-10-27
fix(android): restore "res/drawable" support to Ti.UI.Button images

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28558

**Summary:**
- The changes I made in PR #13125 caused a regression where `Ti.UI.Button` image won't load from "res/drawable" DPI scaled images anymore. We need to restore this behavior.
- This issue was caught before release.

**Test:**
1. Download kitchensink-v2 app: https://github.com/appcelerator/kitchensink-v2/pull/59
2. Tap on the "Button" list item.
3. Verify the "Image Button" icon is the same size as shown in the PR's screenshot, except the icon should be "white" when using the Light theme.


garymathews (approved):
CR: **PASS**
FT: **PASS**

PR #13161, opened 2021-11-02
fix(ios): re-add KrollBridge "modules" method some 3rd party modules depend on

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28415

**Summary:**
- This is for backward compatibility with 3rd modules that call this native method directly, such as the [drag-and-drop](https://github.com/hansemannn/titanium-drag-and-drop) module.
- Method was removed in Titanium 10.0.0. Re-added in 10.2.0.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/KrollBridge.m
@@ -34,6 +34,24 @@ CFMutableSetRef krollBridgeRegistry = nil;
 
 @implementation KrollBridge
 
+- (NSDictionary<NSString *, id> *)modules
+{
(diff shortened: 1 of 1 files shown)


jquick-axway:
@hansemannn , does this PR still do what you need? Are there any other modules that depend on this method?

hansemannn:
We migrated our modules, but there may be others that use it right now. To be honest, I think it even can be removed now, because (a) it's an internal API that should not be depended on and (b) Titanium SDK 10.0 was a major release where breaking changes to the ecosystem should be expected.

ewanharris:
I think Chris' original PR stated similar on not exposing this property. Given that it can already be obtained elsewhere and the drag and drop module is seemingly the only report we had of this breakage, maybe we just don't expose it?

PR #13171, opened 2021-11-08
chore: convert issue templates to issue forms

ewanharris:
This allows us greater control over ensuring the forms are filled out as required. The template at
issue_template.md will now be used if a blank issue is opened

You can try these out [over here](https://github.com/ewanharris/playground/issues/new/choose)


.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,60 @@
+name: Bug Report
+description: Create a bug report for the Titanium SDK
+labels: [ bug, needs triage ]
```
ewanharris, line 3: Note: `needs triage` label will need to be added

.github/ISSUE_TEMPLATE/1.bug_report.yml
```diff
@@ -0,0 +1,60 @@
...
+- type: dropdown
+  attributes:
+    label: Platform
+    description: What is the platform that this issue is seen on?
+    multiple: true
+    options:
+      - Android
+      - iOS
+      - CLI
+- type: input
```
ewanharris, line 56: It would be nice to use a dropdown here to select from known versions, but unfortunately there doesn't seem to be a way to pull from the tags (short of having an action that updates the list on every tag)

.github/ISSUE_TEMPLATE/2.docs_issue.yml
```diff
@@ -0,0 +1,34 @@
+name: Docs Issue
```
ewanharris, line 1: I figured it made sense to have a separate form for docs issues

garymathews (approved):
LGTM! Think we need to add the `triage` label before merge?

PR #13178, opened 2021-11-15
fix: always use PHPickerViewController on iOS 14+, cleanup outdated version guards 

hansemannn:
Currently, the `PHPickerViewController` is only used if the target is iOS 14 *and* `allowMultiple` is set to `true`. To benefit from the improved user experience of the picker, it should always be used if the target is iOS 14+, even if `allowMultiple` is set to `false`.

The relevant commit for this change is https://github.com/appcelerator/titanium_mobile/pull/13178/commits/85c9801e7635693fa85752f1392c4a6c41a0a8ef. I also migrated some outdated version guards for iOS 9.x that are not necessary anymore and will make the code more readable.

**JIRA:** https://jira.appcelerator.org/browse/TIMOB-28579

### DIFF
--- a/iphone/Classes/MediaModule.m
+++ b/iphone/Classes/MediaModule.m
@@ -139,7 +139,7 @@ static NSDictionary *TI_filterableItemProperties;
                                                         MPMediaItemPropertyGenrePersistentID, @"genrePersistentID",
                                                         MPMediaItemPropertyComposerPersistentID, @"composerPersistentID",
                                                         MPMediaItemPropertyIsCloudItem, @"isCloudItem",
-                                                        [TiUtils isIOSVersionOrGreater:@"9.2"] ? MPMediaItemPropertyHasProtectedAsset : NO, @"hasProtectedAsset",
+                                                        MPMediaItemPropertyHasProtectedAsset, @"hasProtectedAsset",
                                                         MPMediaItemPropertyPodcastTitle, @"podcastTitle",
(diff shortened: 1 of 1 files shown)


ewanharris:
I believe the original reason for not using PHPickerViewController by default was that it doesn't offer the editing capability that, looking at the Apple docs that doesn't seem to have changed?  However, editing is only enabled if `allowEditing` is actually set to `true` so I'm not sure that reasoning holds up for the default case

How does this look in terms of the success value? If that's different this is (imo) a semver major change

hansemannn:
Very good point. We only check for the returned images instead of the success value itself, but it's a critical change. I wonder if we should deprecate the `editing` for iOS 14+ as Apple removed it as well. The `success` can be filled in the response.

mbender74:
Hi, I made for my personal use some more changes.... maybe you wanna take a look..
[MediaModule.m.zip](https://github.com/appcelerator/titanium_mobile/files/7539143/MediaModule.m.zip)
..

joecorriere:
Will this fix make into 10.2.0GA ? Trying to gauge if we need to build a work around or wait for the updated SDK.

hansemannn:
The differences between the old picker's `success` event and the new one are too large. We cannot get proper backwards compatibility there, mainly because the new one has the `images` property and the old one the single `media` object. Instead, this should be solved with an external module.

PR #13189, opened 2021-11-18
Update Ti.UI.WebView to restore previous default setting

### LINKED ISSUE #13188: Android Webview android:targetSdkVersion=30+ access local files default changed
With Android SDK Version 30+ by default a webview will not have access to local assets like images accessed through file://.

This is therefore a breaking change since previously the default was for this setting to be true.

Android documentation of the change: https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)

Stackoverflow dealing with the problem natively: https://stackoverflow.com/a/63846853/181197

I will submit a PR so that this becomes a property.

markive:
In Android targetSDKVersion=30+ the setAllowFileAcess property now defaults to false.

This is a breaking change for the Appcelerator SDK as it now prevents access to local file assets through the file:// path.

This overrides the new default and restores the expected behaviour within the Appcelerator SDK.

Fixes #13188 

**Optional Description:**

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
@@ -298,6 +298,7 @@ public class TiUIWebView extends TiUIView
 		settings.setJavaScriptCanOpenWindowsAutomatically(true);
 		settings.setLoadsImagesAutomatically(true);
 		settings.setDomStorageEnabled(true); // Required by some sites such as Twitter. This is in our iOS WebView too.
+		settings.setAllowFileAccess(true);
 		File path = TiApplication.getInstance().getFilesDir();
 		if (path != null) {
 			settings.setDatabasePath(path.getAbsolutePath());


ewanharris:
So it seems like Android changed the default in Android 30 for security reasons? I wonder if for that reason this should potentially become an opt-in property rather than on by default, similar to APIs like `ignoreSslError`.

We see things like this get flagged by security scans as a potential issue and if a property is opt-in for those that do need it (as opposed to enabled for everyone) then the justification is easier (i.e "the app doesn't enable this") than if it was enabled by default

garymathews:
It looks like Android wants us to use [WebViewAssetLoader](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader) to access local files. As Ewan mentioned, it's unclear if apps will fail the review process if [setAllowFileAccess](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)) is set to `true` when targeting Android 30+; the docs strongly advise against setting this `true`.

markive:
> So it seems like Android changed the default in Android 30 for security reasons? I wonder if for that reason this should potentially become an opt-in property rather than on by default, similar to APIs like `ignoreSslError`.

Yes I was thinking that it could, but I'm a new contributor and needed to get it fixed for my app asap. I can look into that and resubmit.



> It looks like Android wants us to use [WebViewAssetLoader](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader) to access local files. As Ewan mentioned, it's unclear if apps will fail the review process if [setAllowFileAccess](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)) is set to `true` when targeting Android 30+; the docs strongly advise against setting this `true`.

I can confirm my app has been released and passed any Google Play scan / approval process. I agree with the security scan as my app has to go through these. The fact is a security scan pings all sorts of defaults in the SDK. My response of these are available in the SDK but not used in my code passed the security scan.

m1ga:
a quick side node:
```js
var webView = new WebView($.www);
console.log(webView.getSettings().getAllowFileAccess()); // false
webView.getSettings().setAllowFileAccess(true);
console.log(webView.getSettings().getAllowFileAccess()); // true
```
as a hyperloop workaround. If someone is facing this issue and need a fix right away without building a custom SDK.

@markive if you want to add a property to set it or not have a look at
https://github.com/appcelerator/titanium_mobile/blob/75da8abe6c13ebd3c12b0094dc3d58928a8f8ba6/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java#L316
on how to read a boolean property. The `TiC` properties are defined in TiC.java but you can also use a string there

PR #13208, opened 2021-12-14
fix(android): update log4j version

m1ga:
As discussed in https://github.com/appcelerator/titanium_mobile/discussions/13206 

* Updated the log4j library to the latest 2.16.0
* Build SDK -> works fine
* setting the default SDK to 10.2.0
* rebuilding the SDK -> works fine
* build a module -> works fine


jquick-axway:
@m1ga , thanks for taking the time to address this...
However, my PR #13210 completely removes the "log4j" library. So, I'm going with that. :)

m1ga:
even better 😄👍 less code and libs is always good

PR #13209, opened 2021-12-14
fix(android): drop-down picker to never accept keyboard input

jquick-axway:
**JIRA:**
https://jira.appcelerator.org/browse/TIMOB-28581

**Summary:**
- Regression as of Titanium 10.0.1.
- `Ti.UI.Picker` set up as a drop-down picker will wrongly accept keyboard input when it should be read-only.
- Can happen in the Android emulator when tapping on the picker and typing on the physical keyboard.
- Can happen when there is a TextField before the picker and you tap on the "Next" button on the virtual keyboard.

**Test:**
2. Tap on the text field.
3. Tap on the Next/Enter button in the virtual keyboard.
4. Drop-down picker now has the focus.
5. Type in the virtual keyboard and verify that the picker does **NOT** change.

```javascript
const window = Ti.UI.createWindow({ layout: "vertical" });
window.add(Ti.UI.createTextField({
	top: "30dp",
	width: "80%",
}));
const picker = Ti.UI.createPicker({
	top: "20dp",
	width: "80%",
});
picker.add([
	Ti.UI.createPickerRow({ title: "Item 1" }),
	Ti.UI.createPickerRow({ title: "Item 2" }),
	Ti.UI.createPickerRow({ title: "Item 3" }),
	Ti.UI.createPickerRow({ title: "Item 4" }),
	Ti.UI.createPickerRow({ title: "Item 5" }),
]);
picker.addEventListener("change", (e) => {
	Ti.API.info("@@@ change: " + e.rowIndex);
});
window.add(picker);
window.open();
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
@@ -34,8 +34,13 @@ import ti.modules.titanium.ui.widget.picker.TiUITimePicker;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.graphics.Color;
+import android.text.Editable;
+import android.text.InputType;
+import android.text.method.BaseKeyListener;
 import android.util.Log;
 import android.view.Gravity;
+import android.view.KeyEvent;
+import android.view.View;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


ewanharris (approved):
CR/FR Pass!

PR #13224, opened 2021-12-29
docs: Titanium.UI.iOS.Toolbar was removed in 10.0.0

drauggres:
https://github.com/appcelerator/titanium_mobile/blob/fb732d5acaa08ca8435ca8ec255c4652ab9071d5/apidoc/Titanium/UI/iOS/Toolbar.yml#L43-L46


ewanharris (approved):
Thanks @drauggres!

PR #13238, opened 2022-01-10
fix: pin to colors 1.4.0

ewanharris:
Pins to colors 1.4.0 to avoid the DoS issue.

Note: this does not impact any released version of Titanium


ewanharris:
The problematic versions were removed from npm so closing this

PR #13239, opened 2022-01-13
Add web interface for communication remote webview -> app

arifje:
Add a web interface to communicate from webview to app with events, just like iOS WKWebview is supporting postMessage:

iOS

```
window.webkit.messageHandlers.Ti.postMessage({
	message: { param : value }
},'*');

```
And with this commit, also for Android, from a remote webView:

```
Android.postMessage(JSON.stringify( { param: value }));
```

Listening to events in the app:

```
webView.addEventListener("message", function(e) {
	var response = e.value;
});

```

**Optional Description:**

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -28,6 +28,10 @@ import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.view.TiUIView;
 import ti.modules.titanium.ui.widget.webview.TiUIWebView;
 
+import android.webkit.JavascriptInterface;
+import android.content.Context;
+import org.appcelerator.kroll.KrollProxy;
+
 @Kroll.proxy(creatableInModule = UIModule.class,
 	propertyAccessors = {
 		TiC.PROPERTY_BLACKLISTED_URLS,  // DEPRECATED: Superseded by PROPERTY_BLOCKED_URLS.
@@ -73,6 +77,28 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 		defaultValues.put(TiC.PROPERTY_ZOOM_LEVEL, 1.0);
 	}
 
+	// Web interface
+	public class WebAppInterface
+	{
+		Context mContext;
+		KrollProxy proxy;
+
+		/** Instantiate the interface and set the context */
+		WebAppInterface(Context c, KrollProxy p)
+		{
+			mContext = c;
+			proxy = p;
+		}
+
+		@JavascriptInterface
+		public void postMessage(String value)
+		{
+			KrollDict dict = new KrollDict();
+			dict.put("value", value);
+			proxy.fireEvent("message", dict);
+		}
+	}
+	
 	@Override
 	public TiUIView createView(Activity activity)
 	{
@@ -85,6 +111,10 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 			postCreateMessage = null;
 		}
 
+		webView.getWebView().addJavascriptInterface(
+			new WebAppInterface(TiApplication.getInstance().getApplicationContext(), this), "Android"
+		);
+		
 		return webView;
 	}
 


m1ga:
@arifje does it have to be in this file? The other bindings are here:
https://github.com/appcelerator/titanium_mobile/blob/fe3deddcea07ecb3555673cf2dba8d676fd79611/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java#L105-L107

hansemannn:
I would actually prefer a common namespace for this, e.g. `Ti.Network.sendMessage` that uses `window.webkit.messageHandlers.Ti` on iOS under the hood and Android under the same space. @arifje Do you think thats possible?

arifje:
@hansemannn a common namespace would be perfect, I think it's important communication of events (Web->App and reversed) should be implemented on both platforms. Our app depends on it.

My native Java skills are limited, and I found this method a few years ago. Worked for my, but I agree on your idea :-)

m1ga:
I'm still a bit confused why :) There are 3 of them already: https://github.com/tidev/titanium_mobile/blob/fe3deddcea07ecb3555673cf2dba8d676fd79611/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java#L105-L107
but they are only added when `interfacesAdded` is true (so a local page) but you can inject the files to a remote page with `webview.evalJS` ([slack post](https://tidev.slack.com/archives/D0363HSL2CQ/p1647165463073249)) and use the normal Ti.App.fireEvent stuff.

It just uses `fireEvent` instead of `postMessage`
https://github.com/tidev/titanium_mobile/blob/fe3deddcea07ecb3555673cf2dba8d676fd79611/android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java#L233-L245

So if we add this one it should be in `addJavascriptInterfaces` but we need to check when to call it (e.g. not only for local pages)

arifje:
Good one. We need it for remote web but I am not familiar with that injecting-part. 

Would love to know how that works, but that URL isn't working (https://tidev.slack.com/archives/D0363HSL2CQ/p1647165463073249).

m1ga:
ah sorry. The last code was in a DM :smile: 

It's basically how the cookie tools work: https://stackoverflow.com/questions/44592123/how-to-add-a-js-function-in-a-script-block-in-javascript/44592382#44592382 
you create a new script element and add the [binding file](https://github.com/tidev/titanium_mobile/tree/master/android/modules/ui/assets/Resources/ti.internal/webview) that Ti adds for a local page itself.

```js
/**
 * binding.js
 * to make Ti.App.fireEvent work on remote page
 */
webview.evalJS(
	'javascript=(function addBinding(){' +
	'var s=document.createElement("script");' +
	's.setAttribute("type","text/javascript");' +
	's.innerHTML="' + require('/binding') + '";' 
	+
	'document.getElementsByTagName("body")[0].appendChild(s);' +
	'})()'
);
/**
 * Insert onmessage Ti.App.fireEvent function
 */
webview.evalJS(
	'javascript=(function addTiFireEvent(){' +
	'var s=document.createElement("script");' +
	's.setAttribute("type","text/javascript");' +
	's.innerHTML="' + "window.onmessage=function(_e){ Ti.App.fireEvent('app:fromWebView', { message: _e }); }" + '";' +
	'document.getElementsByTagName("body")[0].appendChild(s);' +
	'})()'
);
```

hansemannn:
@m1ga That would be a great addition to the docs.

m1ga:
@hansemannn https://github.com/tidev/titanium_mobile/pull/13322
Added the bindings.js example. The second part is very specific for this example so I left that out. If the bindings file is there you can just use the normal `Ti.App.fireEvent` stuff like you want.

jquick-axway:
Right. Everyone should use `Ti.App.fireEvent()` to communicate async between the app and WebView. Also note that this is only supported for WebViews showing local pages (not remote pages) for security reasons. I also recommend avoiding using `evalJS()` since that's blocking.

I wrote the following test code a few years ago which tests `Ti.App` and `Ti.API` support within the HTML. (I have more complex examples, but this is probably the simplest.)
```javascript
const htmlText =
		'<!DOCTYPE html>\n' +
		'<html>\n' +
		'	<head>\n' +
		'	</head>\n' +
		'	<body>\n' +
		'		<p>HTML JavaScript Titanium API Test</p>\n' +
		'	</body>\n' +
		'	<script>\n' +
		'		document.addEventListener("DOMContentLoaded", function(e) {\n' +
		'			Ti.API.info("@@@ Logging via Ti.API.info() method.");\n' +
		'			Titanium.API.info("@@@ Logging via Titanium.API.info() method.");\n' +
		'			Ti.App.fireEvent("app:requestLogMessage", {\n' +
		'				message: "@@@ Fired event via Ti.App.fireEvent() method",\n' +
		'			});\n' +
		'			Titanium.App.fireEvent("app:requestLogMessage", {\n' +
		'				message: "@@@ Fired event via Titanium.App.fireEvent() method",\n' +
		'			});\n' +
		'			Ti.App.addEventListener("app:signalHtml", function eventHandler() {\n' +
		'				Ti.API.info("@@@ Received event via Ti.App.addEventListener() method.");\n' +
		'				Ti.App.removeEventListener("app:signalHtml", eventHandler);\n' +
		'			});\n' +
		'			Titanium.App.addEventListener("app:signalHtml", function eventHandler() {\n' +
		'				Titanium.API.info("@@@ Received event via Titanium.App.addEventListener() method.");\n' +
		'				Titanium.App.removeEventListener("app:signalHtml", eventHandler);\n' +
		'			});\n' +
		'			Ti.App.fireEvent("app:webViewReady", {})\n' +
		'		});\n' +
		'	</script>\n' +
		'</html>\n';

Ti.App.addEventListener("app:requestLogMessage", function(e) {
	Ti.API.info(e.message);
});
Ti.App.addEventListener("app:webViewReady", function(e) {
	Ti.App.fireEvent("app:signalHtml", {});
});

const webView = Ti.UI.createWebView({
	html: htmlText,
});
window.add(webView);
window.open();
```

hansemannn:
Closing in favor of the provided alternatives. Thanks folks!

arifje:
Hi guys, thanks for all the input!

I had to update our app again, updated the SDK to 11.0.0.GA and tested the suggested ideas. But, unfortunately I ended up with creating a custom build from the SDK again (with my proposed changes).

The main reason: when I add the bindings with evalJS, this means I can start sending/receiving events when the webview is completely finished loading (correct me if I am wrong), because the bindings can only be added after the webview has completely finished loading. So in practice this means a big delay, depending of the speed of the page of course

In our case it means +/- 3 seconds before the page is loaded completely. But, sometimes remote scripts are delayed (CDN or other network issues), which results in an even bigger delay. De page is already visible then, can be read, scrolled, etc, but some remote images/scripts aren't ready.

When I use my proposed method, I can instantly send events after the webview is created, while the page is still loading some elements (out of the viewport). This is really essential for us. It's simply not an option to wait 3-5 seconds or even longer before we can do some app <> webview communication.

Issue #13278, opened 2022-02-22
Android: Ti.UI.TextField does not handle "Ti.UI.INPUT_BORDERSTYLE_ROUNDED" borderStyle correctly


hansemannn:
### I have searched and made sure there are no existing issues for the issue I am filing

- [X] I have searched the existing issues

### Description

When setting a text-field's `borderStyle` to `Ti.UI.INPUT_BORDERSTYLE_ROUNDED`, there are two major issues that prevent the API's from being used:

1. It cannot handle a fixed size (the padding breaks, since it's seems to try to calculate with a larger height internally)
2. It cannot handle custom background colors (in that case, the border style is not applied at all - possibly the background color is applied to the wrong sub view internally)

### Expected Behavior

- For case 1: The text field should allow fixed heights instead of the ~50px default height
- For case 2: The text field should allow custom background colors without removing the border style

### Actual behavior

- For case 1: The text field values / hint texts are distorted
- For case 2: The text field removes the border style if a background color is applied.

### Reproducible sample

```js
const window = Ti.UI.createWindow();
window.add(Ti.UI.createTextField({
    top: 20,
    height: 40,
    borderStyle: Ti.UI.INPUT_BORDERSTYLE_ROUNDED,
    hintText: 'Enter value …'
}));
window.open();
```

### Steps to reproduce

Run the above code.

### Platform

Android

### SDK version you are using

10.1.1.GA, 10.2.0 (master)

m1ga:
setting `hintType: Ti.UI.HINT_TYPE_ANIMATED` will change the padding size:

https://github.com/appcelerator/titanium_mobile/blob/db1a83a45529993e2d694e55181a881091cfa849/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java#L968-L972

That at least allows you to make the textfield smaller:
![Screenshot_20220222-131308](https://user-images.githubusercontent.com/4334997/155130226-b4f83830-2611-43c9-82af-a48ff4c751ce.png)
![Screenshot_20220222-131319](https://user-images.githubusercontent.com/4334997/155130230-4c0a8810-755e-4377-aca3-f57310b62887.png)

Still has some issues when you want to type into it.

Looking at the "Layout Inspector" it looks like the padding is applied to the outer `TextInputLayout` container (blue line in the middle) and not the `TiUiEditText`. So the Titanium padding is pushing the TextField smaller.
[...]

PR #13279, opened 2022-02-22
fix(android): textfield padding

m1ga:
Partial fix for https://github.com/appcelerator/titanium_mobile/issues/13278

It will apply the padding to the inner TiUITextField and not the outer layout (see screenshots in the issue above):

```js
window.add(Ti.UI.createTextField({
	top: 20,
	height: 40,
	borderStyle: Ti.UI.INPUT_BORDERSTYLE_ROUNDED,
	hintText: 'Enter value …',
	padding: {
		top: 10,
		bottom: 10
	}
}));
window.open();
```
So you can use a smaller padding yourself if the text is cut of by a fixed height. 
I've kept the default padding for the moment which is too high (48 instead of 28) which looks to be because of the animated hintText option. I'll look into that at some point.

**Before PR (10.1.1.GA):**
the actual textfield is pushed together by the padding:
Screenshot_1648396743
Screenshot_1648396750

**With this PR:**
padding is applied to the actual inner textfield:
Screenshot_1648396695


m1ga:
added a test and updated the first post with some images.

Happy to get some feedback!

caspahouzer (approved):
Textfield padding solved great

LGTM

hansemannn:
Unfortunately this caused a very bad regression for our apps, since it's a breaking change inside a minor version update. Therefore, I would revert it for now and merge the PR once the next major version after 11.1.0 is released. Or maybe a glabl opt-in flag to use the new padding behavior. Thoughts?

m1ga:
can you describe the issue you are seeing? It was just "wrong" before how the padding was applied. There wasn't any feedback in https://github.com/tidev/titanium_mobile/issues/13278 or here so I thought it was fine.

Of course we can revert it. Only one person besides you complained about the padding issue before so other people didn't have issues with the current/old way :rofl: The solution for the other person was to use `borderStyle:none` and put it in a view

hansemannn:
Okay, sounds good!

jquick-axway:
I think part of the issue here is that there is more than one padding. Plus the animated hint text complicates things. And if you ever add bottom error text support, which is rendered below the TextField border, then this gets even more "interesting".

There is an outer-padding which is applied outside of the TextField's border. This is what the old code controlled. Correct me if I'm wrong, but doesn't iOS apply the padding outside of its border?

Then there is the inner-padding which is applied within the TextField's border which what this PR is changing it to. (Do you have to worry about the font getting clipped if the padding is too big?)

Maybe the solution is that these should be 2 separate properties?

PR #13281, opened 2022-02-26
feat(android): support flexible border radius in card view

### LINKED ISSUE #13280: Android: flexible borderRadius for CardView

### Description

The current `CardView` uses only a one value `borderRadius` for setting all corners at once. The normal `View` already allows us to set an array to assign a different value for each corner.

https://github.com/appcelerator/titanium_mobile/blob/03ada5dd0d60d900aa7e5236a161c0f298574737/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java#L167-L174

### Solution

It is possible to set different values for a CardView with this code:
```java
float radius = getResources().getDimension(R.dimen.my_corner_radius);
cardView.setShapeAppearanceModel(
  cardView.getShapeAppearanceModel()
      .toBuilder()
      .setTopLeftCorner(CornerFamily.ROUNDED,..)
      .setTopRightCorner(CornerFamily.ROUNDED,..)
      .setBottomRightCorner(CornerFamily.ROUNDED,radius)
      .setBottomLeftCornerSize(0)
      .build());
```
(source: https://stackoverflow.com/a/57837141/5193915)

It would be nice to have this code for TiUICardView too.
The View borderRadius is set in this method: [setRadius(Object)](https://github.com/appcelerator/titanium_mobile/blob/5c3b9d5fa6749b368a7ef1918187635d2c12b8fd/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java#L163)

This can be used as a starting point to add to the current [setRadius(float)](https://github.com/appcelerator/titanium_mobile/blob/03ada5dd0d60d900aa7e5236a161c0f298574737/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java#L173) method.

TODO:
* check if parameter is `instanceof` int or Object[]
* if `int` use the current method
* if `Object[]` use the check from `TiBorderWrapperView` and the `getShapeAppearanceModel()...` part from the demo code above

to test it create a CardView with an array in the borderRadius parameter
```
var cardView = Ti.UI.Android.createCardView({
        padding: 16,
        borderRadius: [20, 0, 20, 0]
        backgroundColor: "yellow"
    });
```
and a normal one
```
var cardView = Ti.UI.Android.createCardView({
        padding: 16,
        borderRadius: 20
        backgroundColor: "yellow"
    });
```
and see if the corners are correct.

To build the SDK look at https://github.com/appcelerator/titanium_mobile#building-locally

### Alternatives

_No response_

### Platforms

Android

hansemannn:
Fixes #13280 

**Example**:
```js
const window = Ti.UI.createWindow({ layout: 'vertical' });

window.add(Ti.UI.Android.createCardView({
	top: 20,
	width: 100,
	height: 100,
	backgroundColor: '#ff0',
	borderRadius: [ 0, 20, 0, 20 ]
}));

window.add(Ti.UI.Android.createCardView({
	top: 20,
	width: 100,
	height: 100,
	backgroundColor: '#ff0',
	borderRadius: '20 0 20 0'
}));

window.add(Ti.UI.Android.createCardView({
	top: 20,
	width: 100,
	height: 100,
	backgroundColor: '#ff0',
	borderRadius: 20
}));

window.add(Ti.UI.Android.createCardView({
	top: 20,
	width: 100,
	height: 100,
	backgroundColor: '#ff0',
	borderRadius: '20'
}));

setTimeout(() => {
	for (const children of window.children) {
		children.borderRadius = 50;
	}
	setTimeout(() => {
		for (const children of window.children) {
			children.borderRadius = '50 50 0 0';
		}
		setTimeout(() => {
			for (const children of window.children) {
				children.borderRadius = [ 0, 0, 50, 50 ];
			}
			setTimeout(() => {
				for (const children of window.children) {
					children.borderRadius = '0';
				}
			}, 1000);
		}, 1000);
	}, 1000);
}, 2000);

window.open();
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUICardView.java
@@ -25,6 +25,7 @@ import android.view.ViewGroup;
 import android.widget.FrameLayout;
 import androidx.annotation.NonNull;
 import com.google.android.material.card.MaterialCardView;
+import com.google.android.material.shape.CornerFamily;
 
 public class TiUICardView extends TiUIView
 {
@@ -165,12 +166,7 @@ public class TiUICardView extends TiUIView
 		}
 
 		if (d.containsKey(TiC.PROPERTY_BORDER_RADIUS)) {
-			float radius = 0;
-			TiDimension radiusDim = TiConvert.toTiDimension(d.get(TiC.PROPERTY_BORDER_RADIUS), TiDimension.TYPE_WIDTH);
-			if (radiusDim != null) {
-				radius = (float) radiusDim.getPixels(cardview);
-			}
-			cardview.setRadius(radius);
+			setRadius(d.get(TiC.PROPERTY_BORDER_RADIUS));
 		}
 
 		if (d.containsKey(TiC.PROPERTY_BORDER_WIDTH)) {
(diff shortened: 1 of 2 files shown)


m1ga:
PR works fine @hansemannn 

Tested 10.1.1.GA and 10.2.0 with this PR on my Pixel 4 (Android 12) :+1:

PR #13283, opened 2022-02-27
fix(android): roundBorder fix for Android 12+

### LINKED ISSUE #13282: Android: Anti-aliasing when setting View borderRadius

### Description

When setting a borderRadius on a View or Button with a backgroundColor I have some dimmed corners. Looks like there is an anti-aliasing issue. 

There is a "hack" in the code:

https://github.com/appcelerator/titanium_mobile/blob/fe3deddcea07ecb3555673cf2dba8d676fd79611/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java#L106-L110

https://github.com/appcelerator/titanium_mobile/blob/fe3deddcea07ecb3555673cf2dba8d676fd79611/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java#L112-L114

but when I remove it it looks better (left side):

Screenshot_1645895949

Not sure if it has some side-effects yet but it is something we need to look for!

### Expected Behavior

Smooth corners, no dark colors

### Actual behavior

Dark corner pixels

### Reproducible sample

```xml
<Button title="connect"/>
```

```
"Button" : {
  borderRadius: 10,
  backgroundColor: "#399",
  height: 40
}
```

### Steps to reproduce

create a button with the code above

### Platform

Android

### SDK version you are using

10.1.1.GA

m1ga:
Fixes https://github.com/appcelerator/titanium_mobile/issues/13282

In my tests it was only with Android12+

```js
const window = Ti.UI.createWindow({});
window.add(Ti.UI.createButton({
 borderRadius: 10,
 backgroundColor: "#399",
 height: 40
}));
window.open();
```
155852967-a4520b17-1953-4f09-a07a-9f38ecbb35f9


m1ga:
The test-suite has a test already but `matchImage` doesn't catch the difference:

10.1.1.GA:
Screenshot_20220321-191216

10.2.0 + this PR:
Screenshot_20220321-191126

My test suite runs with an Android 13 emulator and the old matching images show up as valid with the fix. So it doesn't catch the slight border

hansemannn:
Okay, so the artifacts are only on Android 12+ (due to the old hack), so you made the hack to be used only on Android 11 and earlier? In that case, it's totally fine for me. We also use the patch in production since a few weeks and it's working fine.

m1ga:
yes, it's only for Android 12+

hansemannn (approved):

PR #13316, opened 2022-03-18
feat: remove Ti.Analytics (iOS / Android)

m1ga:
Removing analytics stuff from the SDK.
Since there might be apps that still use the methods I left them but just made them empty.


android/modules/analytics/src/java/ti/modules/titanium/analytics/AnalyticsModule.java
```diff
@@ -6,204 +6,62 @@
  */
 package ti.modules.titanium.analytics;
 
-import java.util.HashMap;
-import java.util.Iterator;
+import com.appcelerator.aps.APSAnalytics;
```
cb1kenobi, line 9: Can't we just nuke this file?

m1ga: Thought about that too but then it has to be a major update since it will break apps that use `setOptedOut` or other analytics mehtods. Yes they shouldn't use them but I think this change will be better for the first "non appc" release.

cb1kenobi: Can we add a debug log message at either build or run time if someone is using `setOptedOut` that the feature is deprecated?

m1ga: you are right! Added a deprection debug log.

m1ga: Local test are fine too, just the Clipboard issue (https://github.com/tidev/titanium_mobile/pull/12368) from time to time

cb1kenobi: Looks good, thanks!

cb1kenobi (approved):
CR'd, not tested, but looks good to me.

hansemannn:
@m1ga Should I help with iOS? And I suppose Axway did not provide the source for the Analytics to self-host it?

m1ga:
@hansemannn go ahead! Just wanted to start with the Android version here so we get rid of the stuff at one point :)

hansemannn:
@m1ga iOS added!

cb1kenobi (approved):

PR #13317, opened 2022-03-19
fix(android): fix build deps

hansemannn:
The local build was failing because of two reasons:
1) If no cached V8 artifact was found, a new one could not be fetched because the https:// cert expired and we don't have any other server for it, yet
2) The build failed on newer versions of Android Studio because of the outdated Kotlin version

### DIFF
--- a/android/titanium/libv8-services.js
+++ b/android/titanium/libv8-services.js
@@ -174,7 +174,7 @@ async function createSnapshot() {
 			// Post rolled-up "ti.main" script to server and obtain a snapshot ID as a response.
 			// We will send an HTTP request for the snapshot code later.
 			console.log('Attempting to request snapshot...');
-			const snapshotUrl = 'https://v8-snapshot.appcelerator.com';
+			const snapshotUrl = 'http://v8-snapshot.appcelerator.com'; // TODO: Migrate to Github Artifacts once ready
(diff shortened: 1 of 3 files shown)


android/titanium/libv8-services.js
```diff
@@ -174,7 +174,7 @@ async function createSnapshot() {
 			// Post rolled-up "ti.main" script to server and obtain a snapshot ID as a response.
 			// We will send an HTTP request for the snapshot code later.
 			console.log('Attempting to request snapshot...');
-			const snapshotUrl = 'https://v8-snapshot.appcelerator.com';
+			const snapshotUrl = 'http://v8-snapshot.appcelerator.com'; // TODO: Migrate to Github Artifacts once ready
```
m1ga, line 177: Tested it on a Mac and I had the same certificate error. This change fixed it 👍

m1ga:
Don't have the Kotlin problem with my Macbook Air (Intel, 2020), macOS 12.2.1, Android Studio 2021.1.1 Patch 2, Java 15.0.1). I can build an SDK with just the URL fix but having a newer Kotlin version doesn't hurt. The SDK doesn't use any Kotlin files so I'm not sure why it would fail :smile:

PR #13319, opened 2022-03-21
Update node dependencies

m1ga:
Updates the package.json/package-lock.json. No outdated packages at the end (`npm outdated`)
Tested a local build and it still runs fine

`npm audit` shows issues in listview, node-appc and node-titanium-sdk. So we'll need to update those repos later.


hansemannn:
Maybe do the minor updates for now (like Dependabot attempted to)? Then we are save for now and can make the major updates in an own epic.

cb1kenobi:
I think we all need to get on the same page.

1. I believe we need to do whatever it takes to unblock the pipelines. We did this with the node-ios-device fix.
2. We need to touch every single node package, update the readme, remove old CI junk, clean up npm access, etc. I've done this for node-ios-device and node-appc. We have about 40 repos to go.
3. We need to drop Node 10 AND Node 12. Node 10 has been EOL for like a year. Node 12 will be EOL in a month.
4. We need to update dependencies for all packages. This will take a lot of time and result in breaking changes. We will need to major rev everything.
5. We need to take a bottom up approach. Fix/update the dependencies lower on the totem pole first. That's why I've started with node-appc and node-ios-device. Then we can tackle node-titanium-sdk, etc. We should not touch titanium_mobile's dependencies until we get things updated unless it blocks the pipeline.

This is going to take a long time. There are a ton of packages/repos. We will need to rip out dead code, update tests, etc.

While I find value in Dependabot, it's super annoying and I don't love it. I'm not saying we get rid of it, but maybe there's a better alternative or maybe we can tweak it to group updates or something?

Pro tip: when updating deps, try using https://www.npmjs.com/package/npm-check-updates. `npm i -g npm-check-updates` You can run `ncu -t minor` and it will only show update minor/patch changes. Run `ncu -t minor -u` to commit the changes. For npm, you probably can just run `npm upgrade`, but for yarn you must run `yarn && yarn upgrade`.

PR #13327, opened 2022-03-26
Started remove appcelerator references

caspahouzer:


support/module/packaged/modules.json
```diff
@@ -55,8 +55,8 @@
 	},
 	"hyperloop": {
 		"hyperloop": {
-			"url": "https://github.com/appcelerator/hyperloop.next/releases/download/v7.0.4/hyperloop-7.0.4.zip",
-			"integrity": "sha512-5Nu6+o5LfE96wp5KHUR8Lj2iyDO3GrOk3WI5jO/RpD3pTxM9Yxuq2oyvfB4xXu/HnDYFsUKNRqIxsVU+9/AAJA=="
-		}
+      "url": "https://github.com/tidev/hyperloop.next/releases/download/v7.0.4/hyperloop-7.0.4.zip",
```
m1ga, line 58: tabs are missing (same for the three lines below

tests/README.md
```diff
@@ -1,6 +1,6 @@
 # Titanium Mocha Suite
 
-The unit testing app for [Axway Titanium](https://github.com/appcelerator/titanium_mobile), built using [TiMocha](https://github.com/tonylukasavage/ti-mocha).
+The unit testing app for [Axway Titanium](https://github.com/tidev/titanium_mobile), built using [TiMocha](https://github.com/tonylukasavage/ti-mocha).
```
m1ga, line 3: "Axway"?

TitaniumKit.podspec
```diff
@@ -10,14 +10,14 @@ Pod::Spec.new do |s|
                    The Axway Titanium iOS Core framework.
                    DESC
 
-  s.homepage     = "https://github.com/appcelerator/titanium_mobile"
+  s.homepage     = "https://github.com/tidev/titanium_mobile"
   s.license      = { :type => "Apache 2", :file => "LICENSE" }
   s.author       = 'Axway Appcelerator'
```
m1ga, line 15: Axway?

support/module/packaged/modules.json
caspahouzer: That’s vscode formatting. Is that critical?

tests/README.md
caspahouzer: Yeah, want to replace that in the next commits

support/module/packaged/modules.json
m1ga: either use tabs or spaces but not both in one file. Here in that file you've mixed both styles. 
And you can always disable "format on save" option if it makes problems. Or set it to use the files default (at least there was an option for Atom to check and use the format the file uses)

apidoc/Titanium/CHANGELOG/1.2.mdoc
```diff
@@ -1,113 +0,0 @@
-Titanium Mobile 1.2.0 - 4/6/2010
```
hansemannn, line 1: IMHO, these files should not be deleted. There are part of a great legacy.

PR #13346, opened 2022-03-29
feat(android): update internal libraries

m1ga:
Changelog of the libraries:

https://developer.android.com/jetpack/androidx/releases/core#core_and_core-ktx_version_17_2
https://developer.android.com/jetpack/androidx/releases/appcompat#1.4.1
https://developer.android.com/jetpack/androidx/releases/fragment#1.4.1
https://developer.android.com/jetpack/androidx/releases/media#version_15_2
https://github.com/material-components/material-components-android/releases/tag/1.5.0

Currently only tested:
* building the SDK
* building apps
* look & feel of the apps didn't change

Hopefully it won't break anything and it will just give us newer libraries/features we can work with in the future

### DIFF
--- a/android/templates/build/ti.constants.gradle
+++ b/android/templates/build/ti.constants.gradle
@@ -6,11 +6,11 @@
  */
 
 project.ext {
-	tiAndroidXAppCompatLibVersion = '1.3.1'
-	tiAndroidXCoreLibVersion = '1.6.0'
-	tiAndroidXFragmentLibVersion = '1.3.6'
-	tiMaterialLibVersion = '1.4.0'
-	tiPlayServicesBaseLibVersion = '17.6.0'
+	tiAndroidXAppCompatLibVersion = '1.4.1'
+	tiAndroidXCoreLibVersion = '1.7.0'
+	tiAndroidXFragmentLibVersion = '1.4.1'
+	tiMaterialLibVersion = '1.5.0'
+	tiPlayServicesBaseLibVersion = '18.0.1'
 	tiManifestPlaceholders = [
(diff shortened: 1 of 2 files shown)


hansemannn:
Funny - I thought the same: https://github.com/tidev/titanium_mobile/compare/master...hansemannn:fix/update-android-core-libs?expand=1

We should definitely upgrade these.

hansemannn:
From what I see, it's save to integrate. The build is passing, the upgrades are mainly minor semver patches and the app still works fine. I just wonder if the play services update should be delayed a bit, since I can remember how cautious @jquick-axway was when updating it due to common bugs that are fixed in later versions.

m1ga:
I was using this [ti.playerservice](https://github.com/tidev/ti.playservices/pull/230) version 18.0.1 in some test apps already. But I can lower that version again to be safe

jquick-axway:
I recommend that you build the "kitchensink-v2" and "hyperloop-examples" apps to see if this causes any breaking-changes. If those apps look fine, then these dependency updates are probably fine too. (Usually it is, but sometimes Google makes breaking-changes.)

jquick-axway:
I recommend that you build the "kitchensink-v2" and "hyperloop-examples" apps to see if this causes any breaking-changes. If those apps look fine, then these dependency updates are probably fine too. (Usually it is, but sometimes Google makes breaking-changes.)

m1ga:
Thanks @jquick-axway :+1: 
Hyperloop-examples and Kitchensink-v2 work and look fine with this PR. Tested with an Android 13 emulator

hansemannn:
Ok, then do all proposed updates! We wil have a beta cycle anyway, so users will be able to test it before going GA. @m1ga Can you change the PR to "Ready for review"?

hansemannn (approved):

PR #13369, opened 2022-04-06
fix: WebView example

caspahouzer:
I've removed a semicolon in the `createWebView` example as it breaks the app on build with it


m1ga (approved):
LGTM

PR #13372, opened 2022-04-08
feat(android): snackbar

m1ga:
Adds the Android [Snackbar](https://material.io/components/snackbars/)

Screenshot_20220408_124929

```js
	title: 'show message'
});
	snack.show();
});
const snack = Ti.UI.Android.createSnackbar({
	message: 'Hello Snackbar!',
	length: Ti.UI.Android.Snackbar.LENGTH_INDEFINITE,
	action: 'click me'
});
snack.addEventListener('click', function(e) {
	console.log(e.action);
});
win.add([btn, snack]);
win.open();
```


hansemannn:
We could try out a new concept here: Providing a polyfill / parity API for iOS using the Javascript bindings! Some existing examples can be found [here](https://github.com/tidev/titanium_mobile/tree/master/common/Resources/ti.internal/extensions/ti/ui). We would just need to follow the [material design guidelines](https://material.io/components/snackbars) for iOS in terms of size, colors and actions. Happy to do that if interested. Then we could take it as a core API without issues. Otherwise, as discussed, we should publish it as a module.

apidoc/Titanium/UI/Android/Snackbar.yml
```diff
@@ -0,0 +1,82 @@
...
+    type: String
+    since: "10.2.0"
+
+examples:
+  - title: Show a Snackbar
+    example: |
+        Display a Snackbar with an action button.
+
+        ``` js
+        var win = Ti.UI.createWindow({});
```
hansemannn, line 63: Use `const`, remove empty contructor

apidoc/Titanium/UI/Android/Snackbar.yml
```diff
@@ -0,0 +1,82 @@
...
+        var btn = Ti.UI.createButton({
+          title: "show message"
```
hansemannn, line 65: Use single quotes (everywhere)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUISnackbar.java
```diff
@@ -0,0 +1,53 @@
...
+import org.appcelerator.titanium.view.TiUIView;
+
+public class TiUISnackbar extends TiUIView
+{
+	private static final String TAG = "TiUISnackbar";
+
+	public TiUISnackbar(TiViewProxy proxy)
+	{
+		super(proxy);
+	}
```
hansemannn, line 26: Is this required if nothing is overwritten?

android/modules/ui/src/java/ti/modules/titanium/ui/android/SnackbarProxy.java
```diff
@@ -0,0 +1,50 @@
...
+	public static final int LENGTH_SHORT = Snackbar.LENGTH_SHORT;
+	@Kroll.constant
+	public static final int LENGTH_INDEFINITE = Snackbar.LENGTH_INDEFINITE;
+
+	private TiUISnackbar snackbar;
+
+	public SnackbarProxy()
+	{
+		super();
+	}
```
hansemannn, line 36: Is this required?

android/modules/ui/src/java/ti/modules/titanium/ui/android/SnackbarProxy.java
```diff
@@ -0,0 +1,50 @@
+/**
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2017 by Appcelerator, Inc. All Rights Reserved.
```
hansemannn, line 3: Use TiDev header if available

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUISnackbar.java
m1ga: yes, it complains about a missing default constructor otherwise

android/modules/ui/src/java/ti/modules/titanium/ui/android/SnackbarProxy.java
m1ga: no, looks like I can remove this one though :smile:

m1ga:
> We could try out a new concept here: Providing a polyfill / parity API for iOS using the Javascript bindings! Some existing examples can be found [here](https://github.com/tidev/titanium_mobile/tree/master/common/Resources/ti.internal/extensions/ti/ui). We would just need to follow the [material design guidelines](https://material.io/components/snackbars) for iOS in terms of size, colors and actions. Happy to do that if interested. Then we could take it as a core API without issues. Otherwise, as discussed, we should publish it as a module.

Interesting! I like that idea. Will look into that

jquick-axway:
@hansemannn, @m1ga, regarding doing a "polyfill", a year ago I was working the ability to allows Android/iOS to set up a JS class to derive from a native `Ti.UI.View` so that you can do exactly this with views. And it was going to allow you to do a `new` on Titanium proxy classes in JavaScript. I started on the iOS, which can be seen in my 2nd commit in the following branch... although it may have conflicts now. Particularly in the `KrollObject.m` which has the most changes.
https://github.com/jquick-axway/titanium_mobile/commits/TIMOB-28332

I definitely got it working on iOS. However, I did run into a bug with Apple's JavaScriptCore where deriving from a native class from the JS side caused the JS prototype not to be applied. V8 doesn't have this problem.
```javascript
class MyView extends Ti.UI.View {
	constructor(properties) {
		const defaultProperties = {
			borderRadius: "20dp",
			borderColor: "red",
			borderWidth: "5dp",
			backgroundColor: "#dddddd",
		};
		properties = Object.assign(defaultProperties, properties);
		super(properties);

		// Work-around Apple JavaScriptCore bug.
		Object.setPrototypeOf(this, MyView.prototype);
	}
}

const window = new Ti.UI.Window({ backgroundColor: "white" });
const myView = new MyView({ width: "50%", height: "50%" });
window.add(myView);
window.open();
```

jquick-axway:
If you want to `extend` from a Titanium proxy type, then my `KrollObject.m` changes are needed to expose it as a JS class and provide a prototype on iOS. As you know, our `Ti.UI.View` base class has a lot of properties/methods and you probably don't want to replicate it all in JS in the common/kernel. That's what my changes provide. But it's up to you guys. If you're nervous about making core changes like this, then I can understand.

caspahouzer (approved):

apidoc/Titanium/UI/Snackbar.yml
```diff
@@ -0,0 +1,82 @@
+---
+name: Titanium.UI.Snackbar
+summary: Snackbars provide brief messages about app processes at the bottom of the screen.
+description: |
+    For design guidelines, see
+    [Material design: Snackbars](https://material.io/components/snackbars)
+since: "10.2.0"
+platforms: [android, iphone, ipad, macos]
```
hansemannn, line 8: Move to `Ti.UI.Android.*` namespace for now and remove the other platforms, until we have a proper parity API

apidoc/Titanium/UI/Snackbar.yml
```diff
@@ -0,0 +1,82 @@
...
+
+methods:
+  - name: show
+    summary: Show the Snackbar
+
+events:
+  - name: click
+    summary: Fired when the action button is clicked
+
```
hansemannn, line 17: Add the `action` parameter that can be returned

hansemannn (changes requested):
Some minor changes, then ready to go!

apidoc/Titanium/UI/Snackbar.yml
m1ga: moved everything to the Android namespace

hansemannn (approved):

PR #13373, opened 2022-04-09
feat(android): apply Android code optimizations

m1ga:
I've ran "Analyze - Inspect Code" from Android Studio for the SDK.

Before:
Screenshot_20220409_160731

and applied some of the changes:
* minor upgrade to gradle 7.0.4
* remove redundant casts
* StringBuilder -> String
* optimize some if's for redundant code
* some array copy loops [got replaced](https://github.com/tidev/titanium_mobile/compare/master...m1ga:codeCheck?expand=1#diff-a819f80eff8e0223eb52ba7d63518b58373fbe0edc4d893fe77bff8d031a06a4R318)
* many `<Type>` -> `<>`
* removed some expecptions (e.g. `IOException, UnknownHostException` -> `IOException` since that will catch the 2nd one too) 
* other minor parts

and changed some `Log.i(..., debug)` to `Log.d(...)` so it's easier to spot and find them.

Code check afterwards:
Screenshot_20220409_162526

Android: 10 -> 6 errors
Java: 955  -> 773 warnings

Some other parts change (e.g. some more errors in C++ and General), might be because of the gradle update.
Not all checks are valid or introduced problems so I didn't just apply everything. But I think we have some room for improvement here :smile: 

*Test*
I've compiled two bigger apps with it and ran the test suite and didn't see any issues there. But it's a big one to check!


hansemannn:
I really don't feel save reviewing this TBH. Maybe someone else with more Android exp can review this?

m1ga:
I think I'll split this, so the "easy" parts are in a own PR

m1ga:
As a first step: this one should be a lot easier to read/follow: https://github.com/tidev/titanium_mobile/pull/13605

We should keep this one open because it still has other changes.
e.g. https://github.com/tidev/titanium_mobile/pull/13373/files#diff-d28d92bfdcb28f363e2151129d2b323644ec5fb11a61d699d88b3dba6fe7ae57R24-R27

or https://github.com/tidev/titanium_mobile/pull/13373/files#diff-942c49a99612fc82eb16610e063d5230f16727620a7271bf06628187b4207554R96-R98

Perhaps that is even faster then the for-loop :shrug:

m1ga:
close this one now since its more difficult to see the changes. I've made two smaller PRs (one is already merged) and since the isses can be found with Android Studio again we don't need this one as a reference

PR #13375, opened 2022-04-10
fix(ios): remove falsy error log

hansemannn:
The error was thrown when checking for `window.navigationWindow`, which is the wrong place because the property can be used to check if it's available or not.


hansemannn:
@m1ga @caspahouzer Okay to proceed?

m1ga:
According to [the docs](https://titaniumsdk.com/api/titanium/ui/window.html#navigationwindow) it is `null` if there is not navigation window, so throwing an error shouldn't be necessary. 
So removing it should be fine

PR #13390, opened 2022-04-12
feat(android): expose base context

m1ga:
Adds a new method we can use in modules to get the (old) base Activity with `TiApplication.getBaseActivity()`. 

**Why**
This base activity is needed e.g. to call the `printmanager`. We "lose" the reference to that base activity when we start the `TiLocalManager.getLocalizedContext()`. 
In native land this happens too (examples: [SO post 1](https://stackoverflow.com/a/58888333/5193915), [SO post 2](https://stackoverflow.com/a/43775680/5193915)) and up until now I only saw that the `printmanager` has a problem: `java.lang.IllegalStateException: Can print only from an activity` when trying to use it. Non of my other modules had this issue before.

So it won't cause any issues with existing modules/apps and we can use printing on Android again (perhaps it will be useful for other modules too).


android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -631,6 +631,7 @@ public void setContentView(View view, LayoutParams params)
 	@Override
 	protected void attachBaseContext(Context newBase)
 	{
+		baseContext = newBase;
```
hansemannn, line 634: Shouldn't this be released / cleaned up again at some point?

m1ga: :thinking: I think only when the app is closed/killed but then it should be gone anyways. During the runtime it should be there all the time since its the root/source activity.

m1ga: But since there is a `onDestroy` I'll do a quick test if it still works when I put it into that and clear it

m1ga: fast builds now :smile: 

Added 
```
if (baseContext != null) {
	baseContext = null;
}
```
to `onDestroy()`. It's going into it when closing the app. Restarting it still works fine with the print module. Feels better like this. Added a second window to my printer test app to and that still works correct too. 

Tested: 
* win1 with all the content, opening win2, closing win2, calling printer in win1
* win2 with all the content, opening win2, calling printer in win2, closing win2

hansemannn:
@drauggres Would you mind taking a peek here? :)

drauggres:
From DM with @m1ga:

> Here is what I found:
> * `newBase` which we receive in `TiBaseActivity.attachBaseContext` has internal field `mOuterContext` (in `ContextImpl` class) which points to this `Activity` instance.
> * printmanager uses `mOuterContext` to detect if the Context is an Activity (otherwise it throws `IllegalStateException` "Can print only from an activity")
> * when we call `TiLocaleManager.getLocalizedContext`, we create new `Context` and lose the orignal, also in this new context `mOuterContext` does not point to our activity
> * `ContextImpl.setOuterContext` is not available for applications (Android uses this method internally to set `mOuterContext`), i.e. we can't set it ourselves
> 
> If I modify `TiLocaleManager.getLocalizedContext` to not create new context, but to update configuration of the existing, it works, but `Resources.updateConfiguration()` is deprecated.
> 
> ```java
>     public static Context getLocalizedContext(Context ctx) {
>         Configuration config = ctx.getResources().getConfiguration();
>         config.setLocale(currentLocale);
>         ctx.getResources().updateConfiguration(config, ctx.getResources().getDisplayMetrics());
>         return ctx;
>     }
> ```

* I think this is a bug in Android: we shouldn't lose `mOuterContext` when we call `createConfigurationContext`
* If we change `getLocalizedContext` as shown above, it will fix exising modules without having to use this new API
* We use deprecated `Resources.updateConfiguration()` anyway
* Commit where this change was introduced: https://github.com/tidev/titanium_mobile/commit/095c3922d9c25356b175390a5937380a0bf3cefc

m1ga:
I don't have any other existing module that is broken at the moment besides `printmanager`. So I'm a bit scared that the `getLocalizedContext` change might bring some other issues (localization not working, modules not working). Otherwise they didn't work since 2019. And adding another place with a deprecated method isn't perhaps the best thing looking into the future 😄 
But I'll test that change tomorrow

drauggres:
If you are going to create a new method, I think it would make more sence to add it directly to `TiBaseActivity` and name it `getBaseContext`: 
* each `Activity` has its own base context (and `Application` has own too)
* return value is not an `Activity` instance, but a `Context`

And also I would make `baseContext` private.

drauggres (approved):

hansemannn (approved):

m1ga:
Renaming the function to `getBaseContext()` brings an issue! I can't resume a killed app:

```
java.lang.IllegalStateException: System services not available to Activities before onCreate()
      at android.app.Activity.getSystemService(Activity.java:6871)
      at com.android.internal.policy.DecorContext.getSystemService(DecorContext.java:94)
      at android.content.Context.getSystemService(Context.java:3936)
      at android.view.ViewRootInsetsControllerHost.getInputMethodManager(ViewRootInsetsControllerHost.java:233)
      at android.view.ImeInsetsSourceConsumer.getImm(ImeInsetsSourceConsumer.java:164)
      at android.view.ImeInsetsSourceConsumer.onWindowFocusLost(ImeInsetsSourceConsumer.java:62)
      at android.view.InsetsController.onWindowFocusLost(InsetsController.java:1374)
      at android.view.ViewRootImpl.dispatchDetachedFromWindow(ViewRootImpl.java:5078)
      at android.view.ViewRootImpl.doDie(ViewRootImpl.java:8282)
      at android.view.ViewRootImpl.die(ViewRootImpl.java:8259)
      at android.view.WindowManagerGlobal.removeViewLocked(WindowManagerGlobal.java:500)
      at android.view.WindowManagerGlobal.removeView(WindowManagerGlobal.java:441)
      at android.view.WindowManagerImpl.removeViewImmediate(WindowManagerImpl.java:196)
      at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:5512)
      at android.app.servertransaction.DestroyActivityItem.execute(DestroyActivityItem.java:47)
      at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
      at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
      at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2253)
      at android.os.Handler.dispatchMessage(Handler.java:106)
      at android.os.Looper.loopOnce(Looper.java:201)
      at android.os.Looper.loop(Looper.java:288)
      at android.app.ActivityThread.main(ActivityThread.java:7870)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```
Everything else still works but killing it and then reopen it will crash with the error above.

I'll rename it to `getAppContext()` so it will work again.

PR #13391, opened 2022-04-12
feat(android): add :local npm run scripts

m1ga:
Adding two more npm scripts to build with `--skip-zip`

`npm run cleanbuild:local` and `npm run build:local` (without the cleaning part for even quicker local builds)

```
cleanbuild            2m11,086s
build without clean   1m7,434s

cleanbuild:local      1m44,826s
build:local           0m46,621s
```

makes it quicker to test local builds


hansemannn (approved):
Cool addition for development!

PR #13400, opened 2022-04-16
test: Android framents reordering

m1ga:
just a test PR to get the action to build a test SDK


m1ga:
It looks like the key issue is to more the `manager = ((FragmentActivity) activity).getSupportFragmentManager();` out of the `onAttachedToWindow`. Still doing some tests and see what is needed and what not. But it's working a lot better at the moment

PR #13416, opened 2022-04-21
feat(ios): textarea contentHeight fix with new return property textareaHeight in "change" event

### LINKED ISSUE #13413: iOS: textarea missing correct contentHeight on content change - event "change" now with correct height (no postlayout needed)

### Description

Textarea on iOS missing a correct contentHeight feature.

### Solution

I fixed this on event "change" for textarea which will always return the correct textarea height when the content of textarea is changing.

```
textarea.addEventListener('chage',function(e) {
        console.log(e.textareaHeight);
        console.log(e.value);
});

```

### Alternatives

_No response_

### Platforms

iOS

mbender74:
Fixes #13413


apidoc/Titanium/UI/TextArea.yml
```diff
@@ -518,6 +518,11 @@ events:
       - name: value
         summary: New value of this text area.
         type: String
+      - name: textareaHeight
```
hansemannn, line 521: Needs to be `textAreaHeight`

iphone/Classes/TiUITextField.m
```diff
@@ -564,7 +564,7 @@ - (BOOL)textFieldShouldEndEditing:(UITextField *)tf
 - (BOOL)textFieldShouldClear:(UITextField *)tf
 {
   // we notify proxy so he can serialize in the model
-  [(TiUITextFieldProxy *)self.proxy noteValueChange:@""];
+  [(TiUITextFieldProxy *)self.proxy noteValueChange:@"":nil];
```
hansemannn, line 567: Why not use the named parameter?

iphone/Classes/TiUITextWidgetProxy.m
```diff
@@ -118,13 +118,22 @@ - (BOOL)focused:(id)unused
...
-- (void)noteValueChange:(NSString *)newValue
+- (void)noteValueChange:(NSString *)newValue:(NSNumber *)contentHeight
 {
   NSString *oldValue = [TiUtils stringValue:[self valueForKey:@"value"]];
   if (![oldValue isEqual:newValue]) {
     [self replaceValue:newValue forKey:@"value" notification:NO];
+
+    NSMutableDictionary *event = [NSMutableDictionary dictionary];
+    if (contentHeight != nil) {
+      [event setValue:contentHeight forKey:@"textareaHeight"];
```
hansemannn, line 129: textAreaHeight as well. Or just `height` since it's in the implicit context

hansemannn (changes requested):
The backwards compatibility looks fine, but please see the changes before merging. Thx!

iphone/Classes/TiUITextField.m
mbender74: What named parameter do you mean?

mbender74 (commented):
changes made
- event property is now "contentHeight" which is more self explaining and conform with similar properties for other views. "height" would not make sense, because, the textArea has a own height, and this event property reflects the inner scrollView height of the text widget.
- API docs updated to that

hansemannn (approved):
Looks solid, let's go! Thanks for this! 🚀

PR #13417, opened 2022-04-22
refactor: change version number

m1ga:
As discussed earlier. `10.2.0` -> `11.0.0`


caspahouzer:
This is the PR we merge on May 1st? ❤️

m1ga:
I think: the quicker it's merged the more time we have to test it and make other changes if needed before May

caspahouzer:
But isn't then SDK 11 automatically released by the workflow?

m1ga:
its called 10.2.0 now and is not released, it will just change that and build the artifacts with the new version number. So its the same as the other weeks but just a different number :smile: 
But with that change we can adjust all other waiting feature PRs since we have to raise their `since:...` number to `11.1.0` or so. 

Also the milestone numbers should be changed (currently it says 10.2.0 is the first TiDev release). But I'll step back a bit and leave it you guys to decide on that and change it if needed :wink:

PR #13429, opened 2022-05-07
fix: package.json broken build:local missing node command on last ins…

DzzD:
Titanium SDK Build fix : 
packager.json=>build:local not working as it is missing a node command for the last part*

Module build command Fix : 
When building module target & device-id command line parameter are now used for the build & deploy of the example application

### DIFF
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -1,0 +1,38 @@
+name: Regen Docs
+on:
+  pull_request:
+    types: [ closed ]
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


m1ga:
ah, you are making both PRs in the same branch! Please create PRs from a separate branch so they only contain one change.

m1ga:
quick info why it wasn't an issue before: on linux/osx those files are executable so you can run them by it's own. But since they are not an exe it might not work on windows. This way it's cleaner and works everywhere :+1: I'll approve it once you've separated the PRs

PR #13430, opened 2022-05-08
Fix module example application build parameters : target & deviceId

DzzD:
**GitHub discussions:** https://github.com/tidev/vscode-titanium/issues/993

When building a module, this apply the command line parameters target & device Id to the build of the module application.

Building a module with 
`ti build --target device --device-id AAAB`
will result in building the module example application with the same parameters
`ti build --target device --device-id AAAB`

Before, the module application was built without those two parameters regardless if they have been used in the module build comand line

NB: also fix package.json build:local command

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "scripts": {
     "android": "node ./build/scons cleanbuild android",
     "build": "node ./build/scons build",
-    "build:local": "node ./build/scons build && node ./build/scons package --skip-zip && ./build/scons install",
+    "build:local": "node ./build/scons build && node ./build/scons package --skip-zip && node ./build/scons install",
     "build:android": "npm run build -- android",
     "build:changelog": "conventional-changelog -n changelog/config.js -i CHANGELOG.md -s -p angular",
     "build:docs": "docgen apidoc -o ./dist",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "android": "node ./build/scons cleanbuild android",
     "build": "node ./build/scons build",
-    "build:local": "node ./build/scons build && node ./build/scons package --skip-zip && ./build/scons install",
+    "build:local": "node ./build/scons build && node ./build/scons package --skip-zip && node ./build/scons install",
```
m1ga, line 22: this one shouldn't be in this PR. Always start from the normal master branch

m1ga:
add some text in the first post what this change does

package.json
DzzD: Initialy I wanted to make two PR, but I have to correct this one if I wanted that compilation to work.

I dont understand how people are supposed to compile on Windows as it is buggy, anyway no idea on how to split the two commits now in two separated PR ? (I mean without too much complexity)

m1ga: I would remove the fork on your machine, clone it again so you have a fresh folder. Then create a branch `git checkout -b fixModuleBuild` and copy the code from here into that branch. Commit and push. Then go the master again (so its still without any changes) and create a second branch `git checkout -b secondFeature` and copy the code from the other PR into that, commit, push. Then you always have a clean master without any changes and two separate branches for your PRs

DzzD: That was my first idea :), I was just looking for something less tedious, but if it is the good workflow for Titanium, I will do that

PR #13431, opened 2022-05-08
fix(build): fix windows build of Titanium SDK - titanium_mobile

DzzD:
Thoses are two small changes required to build titanium_mobile project on windows : 
1 - **node** command is missing in **package.json** for **build:local** (in the last part of the command line)
2 - **packager.js** use an unknow flag (-y) for zip tool on windows

### DIFF
--- a/build/lib/packager.js
+++ b/build/lib/packager.js
@@ -31,7 +31,7 @@ const TITANIUM_PREP_LOCATIONS = [
  */
 async function zip(cwd, filename) {
 	const command = os.platform() === 'win32' ? path.join(ROOT_DIR, 'build/win32/zip') : 'zip';
-	await exec(`${command} -9 -q -r -y "${path.join('..', path.basename(filename))}" *`, { cwd });
+	await exec(`${command} -9 -q -r "${path.join('..', path.basename(filename))}" *`, { cwd });
 
 	const outputFolder = path.resolve(cwd, '..');
 	const outputFile = path.join(outputFolder, path.basename(filename));
--- a/build/lib/packager.js
+++ b/build/lib/packager.js
@@ -31,7 +31,8 @@ const TITANIUM_PREP_LOCATIONS = [
  */
 async function zip(cwd, filename) {
 	const command = os.platform() === 'win32' ? path.join(ROOT_DIR, 'build/win32/zip') : 'zip';
-	await exec(`${command} -9 -q -r "${path.join('..', path.basename(filename))}" *`, { cwd });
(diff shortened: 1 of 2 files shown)


build/lib/packager.js
```diff
@@ -31,7 +31,7 @@ const TITANIUM_PREP_LOCATIONS = [
  */
 async function zip(cwd, filename) {
 	const command = os.platform() === 'win32' ? path.join(ROOT_DIR, 'build/win32/zip') : 'zip';
-	await exec(`${command} -9 -q -r -y "${path.join('..', path.basename(filename))}" *`, { cwd });
+	await exec(`${command} -9 -q -r "${path.join('..', path.basename(filename))}" *`, { cwd });
```
m1ga, line 34: > -y   store symbolic links as the link instead of the referenced file

it looks like this is only unavailable for windows. So we should keep that for the other platforms.

DzzD: But it wont works on windows as it

m1ga: that's why I said: `keep it for the other platforms` :wink: For windows you can remove it. Check the line above how to check for windows

DzzD: I suppose I can use **os.platform() === 'win32'** to check when to use it ?

m1ga: yes, something like `const params = os.platform() === 'win32' ? '-9 -q -r' : '-9 -q -r -y';`

DzzD: yes perfect

DzzD: I copy/paste your code 👍 and udated PR

m1ga (approved):
changes look good. local build still runs fine on Linux. 
Can't test the build on windows but I think you've tested that :smile:

DzzD:
> Can't test the build on windows but I think you've tested that 😄

oups inded 🙄 , that's your code I am confident and it does not work before so...

ewanharris (changes requested):
@DzzD could you please ensure to [sign the CLA](https://github.com/tidev/organization-docs/blob/main/BECOMING_A_COMMITTER.md) so we can merge this. 

Marking with request changes to ensure this doesn't get merged til that's signed (CLA check is misleading)

DzzD:
> @DzzD could you please ensure to [sign the CLA](https://github.com/tidev/organization-docs/blob/main/BECOMING_A_COMMITTER.md) so we can merge this.
> 
> Marking with request changes to ensure this doesn't get merged til that's signed (CLA check is misleading)

I sent an email to Josh today, so I am wainting its answer

caspahouzer (approved):
LGTM

PR #13432, opened 2022-05-08
fix(build): for module build : enable target & device-id command line parameters to be used for the module application build process

DzzD:
**GitHub Issue:** https://github.com/tidev/vscode-titanium/issues/993

When building a module having an application, target & deviceId parameters was not used when building the application.
With that PR, if building a module with : 
`ti build --target device --device-id MYDEVICEID` 
It will deploy the module application to the desired device


android/cli/commands/_buildModule.js
```diff
@@ -889,7 +891,15 @@ AndroidModuleBuilder.prototype.runModule = async function (cli) {
 
 	// Run the temp app.
 	this.logger.debug(__('Running example project...', tmpDir.cyan));
-	const buildArgs = [ process.argv[1], 'build', '-p', 'android', '-d', tmpProjectDir ];
+	let buildArgs = [ process.argv[1], 'build', '-p', 'android', '-d', tmpProjectDir ];
```
ewanharris, line 894: We've discussed exposing these previously and one issue raised was that there would be no verification of chosen target/device ID until the app build has started, do you think that's an issue or maybe just a potential future improvement?

This change should also be done across ios/android at the same time preferably

DzzD: IMO, this is not a big issue as the target and device are not set by user but choosen from a list of avalaible ones detected.

The risk is only that if one device/emultator  shutdown/disconnect during build time before application build.

I think it should not be difficult to make a pre-verification but is that make sens ? the risk will stay the same, i mean if they get shutdown/disconnect after verification during module compilation, that's why I dont think it is much an issue.

DzzD: > This change should also be done across ios/android at the same time preferably

Sorry, I did completely forget this part, I think the Titanium SDK part is made for both android/ios (as it was requiring very small code change) but I forgot in this VSCode PR, having no mac or ios device may make this a bit hard for me to test.

DzzD: https://github.com/tidev/titanium_mobile/blob/6526d2d55c1989fb113572efb03950de87bfef01/iphone/cli/commands/_buildModule.js#L1072-L1088

Does not look very different logic, I will update the PR, but it wont be tested...

m1ga:
Just tested it because I wanted to try to build the module example directly to my device.

I only run `ti build -p android  -T device` and it worked fine! Deployed straight to my device. Very nice @DzzD !
Tried the `--device-id` part too: when I add a wrong device it will prompt me to select the correct one!
```
[ERROR] Invalid Android device "pixel4"

Which device do you want to install your app on?
Devices
   1)  Pixel 4 (12)
Select a device by number or name: 1
```

DzzD:
> ...Very nice @DzzD ! Tried the `--device-id` part too: when I add a wrong device it will prompt me to select the correct one!

So I Hhave achieved my second badges in the Titanium World !
1 - Making a new Titanium module - checked
2 - Making a PR to Titanium - checked

caspahouzer:
@cb1kenobi 
```
Run tidev/tidev-cla-action@v1
[4](https://github.com/tidev/titanium_mobile/runs/6409176078?check_suite_focus=true#step:2:4)
User DzzD for commit 70c341575b5295a689fe3bd37c53e657cee6e7ea not authorized
```
Why isn't he authorized? He's in the [list](https://github.com/tidev/organization-docs/blob/main/AUTHORIZED_CONTRIBUTORS.md). Is the check case sensitive ? DzzD / dzzd

cb1kenobi:
@caspahouzer CLA usernames are case sensitive. I can fix this later today and rerun the CLA check.

caspahouzer:
I also can change his username in the cla list

cb1kenobi:
@caspahouzer CLA action has been fixed!

caspahouzer:
You're awesome. Thanks

PR #13434, opened 2022-05-11
fix(android): revert to V8 8

m1ga:
To maintain compatibility with SDK 11 modules and older SDKs we revert the V8 level back to 8.8.278.17 (Ti 10.1.1.GA)

More infos in https://github.com/tidev/titanium_mobile/issues/13395
For future releases a fix to V8 9 is in the issue

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -19,9 +19,9 @@
 	],
 	"architectures": ["arm64-v8a", "armeabi-v7a", "x86", "x86_64"],
 	"v8": {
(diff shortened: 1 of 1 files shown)


m1ga:
**Test:**
* build an Android module with 11.0.0 (minsdk should still be 9.0.0)
* put the module into an app that has sdk-version 10.1.1.GA
* build the app with 10.1.1.GA

App should still run without an error

hansemannn (approved):

PR #13443, opened 2022-05-12
feat(android): update gradle to 7.4.2

### LINKED ISSUE #13442: Android: upgrade gradle version

### Description

we can raise the gradle version in https://github.com/tidev/titanium_mobile/blob/master/android/gradle/wrapper/gradle-wrapper.properties#L3

Did some quick build tests with 7.3.3 (should bring Java 17 support; [release notes](https://docs.gradle.org/7.3.3/release-notes.html)) and a SDK build worked fine. Same with 7.4.2 (latest) but it showed some deprecation warnings.

Just did some quick build at the moment, nothing else tested. Need to test some apps and module builds. Ticket is to track the progress and not forget about it 😄 

### Solution

-

### Alternatives

-

### Platforms

Android

m1ga:
resolves https://github.com/tidev/titanium_mobile/issues/13442

* update gradle tools to 7.4.2 (from 7.1.0)
* minor update to gradle build tools to 7.0.4 (from 7.0.1)
* minor update to kotlin to 1.6.21 (from 1.6.10)
* will give us Java 17 support (starting with 7.3.3; did all test with 7.3.3 first, then tested 7.4.2 and it worked too)
* added a section to the README how to update gradle

Tests I've did:

**Javac 11**
* `javac --version` -> 11.0.15
* build SDK
* build module with 11.0.0 -> run it in app with 10.1.1
* build app with 11.0.0

**Javac 17** 
* `javac --version` -> 17.0.3
* build SDK
* build module with 11.0.0 -> run it in app with 10.1.1
* bulid app with 11.0.0
* run whole test suite (Android 13 emulator)
* run `./gradlew` in `build/android` folder to check it shows version 7.4.2
* hyperloop examples
* build kotlin module

**Note:**
* javac 17 is not compatible with Ti 10.1.1.GA because of the older gradle version
* raising the gradle-build tools at https://github.com/tidev/titanium_mobile/blob/master/android/build.gradle#L16 to the latest version will show an error compiling the SDK. Not part of this PR but we should look at raising that at some point too


android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
```
jquick-axway, line 3: I recommend that you use the "all" distribution type (not "bin") so that it will include docs.
```sh
./gradlew --gradle-version 7.4.2 --distribution-type all
```

m1ga: Thanks for the info! I don't have that flag `--distribution-type` :thinking:  But I've changed the file to all

android/build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
 		mavenCentral()
 	}
 	dependencies {
-		classpath 'com.android.tools.build:gradle:7.0.1'
+		classpath 'com.android.tools.build:gradle:7.0.4'
```
jquick-axway, line 16: You also need to update the gradle build tool version in the following file to match. Unfortunately, it's the 1 place we have it duplicated in.
https://github.com/tidev/titanium_mobile/blob/master/android/templates/build/root.build.gradle#L10

You should also update Titanium's [android/package.json](https://github.com/tidev/titanium_mobile/blob/master/android/package.json#L23) file's min SDK build tools to `30.0.3` to match what Google documented [here](https://developer.android.com/studio/releases/gradle-plugin#compatibility-7-1-0). Should look like...
```json
{
	"vendorDependencies": {
		"android build tools": ">=30.0.3 <=31.x"
	}
}
```

m1ga: Thanks for all the infos :+1: I've change the template gradle.

Isn't the "gradle plugin" the one with version 7.0.4? and just the gradle tool the one with 7.4.2? So I can keep the [30.0.2 as a minimum](https://developer.android.com/studio/releases/gradle-plugin#compatibility_2). Their naming convention is strange :rofl:

jquick-axway:
Here's an example on how I updated gradle (and thinks related to it) in the past, if you think it helps. I usually test with [hyperloop-examples](https://github.com/appcelerator/hyperloop-examples) to make sure it doesn't break hyperloop.
https://github.com/tidev/titanium_mobile/pull/12821

I'm pretty sure the newest JDK breaks kotlin too. You might want to upgrade the kotlin plugin as well. :)
I hope this helps!

android/build.gradle
jquick-axway: Oh you're right. We only need to update the min if the Android build tool is 7.2.x. Good catch. :stuck_out_tongue:

m1ga:
Hyperloop examples still work fine with this setup and I can compile my kotlin mlkit module + the test app. The `:app:desugarDebugFileDependencies` part takes a bit longer now but when I do a store build the APK is 3mb smaller.

android/gradle/wrapper/gradle-wrapper.properties
jquick-axway: Sorry. I wrote it wrong. Need to add "wrapper" command to it first.
```sh
./gradlew wrapper --gradle-version 7.4.2 --distribution-type all
```

m1ga: ah ok :+1: no changes after that, but good to know. I only used the first part of it from the gradle docs

caspahouzer (approved):
LGTM

PR #13454, opened 2022-05-18
Fix changelog generation

drauggres:
fix `npm run build:changelog`

The `dateformat` package is now a esm module, but we require it from `changelog/config.js` (a commonjs module).
But we can't convert `config.js` to esm, because it is required from `conventional-changelog` which is also a commonjs module.

### DIFF
--- a/changelog/config.js
+++ b/changelog/config.js
@@ -360,10 +360,9 @@ module.exports = {
 				}
 				eosDate.setMonth(eosDate.getMonth() + 6);
 			}
-			const yyyy = eosDate.getUTCFullYear();
-			const mm = (eosDate.getUTCMonth() + 1).toString().padStart(2, '0');
-			const dd = eosDate.getUTCDate().toString().padStart(2, '0');
-			context.eosDate = `${yyyy}-${mm}-${dd}}`;
+
+			// yyyy-mm-dd
+			context.eosDate = eosDate.toISOString().split('T')[0];
 
 			// Gather up the modules shipped with this version and toss into a variable we can put into the changelog
 			const modules = gatherModules();
--- a/changelog/config.js
+++ b/changelog/config.js
@@ -2,7 +2,6 @@
 
 const path = require('path');
 const semver = require('semver');
-const dateFormat = require('dateformat');
 const fs = require('fs-extra');
 // eslint-disable-next-line security/detect-child-process
 const execSync = require('child_process').execSync;
@@ -361,7 +360,10 @@ module.exports = {
 				}
 				eosDate.setMonth(eosDate.getMonth() + 6);
 			}
-			context.eosDate = dateFormat(eosDate, 'yyyy-mm-dd', true);
+			const yyyy = eosDate.getUTCFullYear();
+			const mm = (eosDate.getUTCMonth() + 1).toString().padStart(2, '0');
+			const dd = eosDate.getUTCDate().toString().padStart(2, '0');
+			context.eosDate = `${yyyy}-${mm}-${dd}}`;
 
 			// Gather up the modules shipped with this version and toss into a variable we can put into the changelog
 			const modules = gatherModules();
--- a/package.json
+++ b/package.json
@@ -135,7 +135,6 @@
     "core-js": "^3.18.3",
     "cz-conventional-changelog": "^3.3.0",
     "danger": "^10.6.6",
-    "dateformat": "^5.0.3",
     "eslint": "^8.13.0",
     "eslint-config-axway": "^7.0.0",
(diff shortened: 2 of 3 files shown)


changelog/config.js
```diff
@@ -361,7 +360,10 @@ module.exports = {
 				}
 				eosDate.setMonth(eosDate.getMonth() + 6);
 			}
-			context.eosDate = dateFormat(eosDate, 'yyyy-mm-dd', true);
+			const yyyy = eosDate.getUTCFullYear();
+			const mm = (eosDate.getUTCMonth() + 1).toString().padStart(2, '0');
+			const dd = eosDate.getUTCDate().toString().padStart(2, '0');
+			context.eosDate = `${yyyy}-${mm}-${dd}}`;
```
cb1kenobi, line 366: I see a trailing curly brace that we probably don't want in the date string.

Furthermore, why not just make this whole thing simpler and do:

```js
eosDate.toISOString().split('T')[0]
```

package.json
```diff
@@ -135,7 +135,6 @@
     "core-js": "^3.18.3",
     "cz-conventional-changelog": "^3.3.0",
     "danger": "^10.6.6",
-    "dateformat": "^5.0.3",
```
cb1kenobi, line 138: Are you sure dateformat is no longer needed? Wouldn't it be safer to revert the config.js changes and downgrade dateformat to v4.6.3?

cb1kenobi (changes requested):

package.json
caspahouzer: I locally reverted to this version. After `npm run build:changelog` I got
```
MacBook:titanium_mobile sebastianklaus$ npm run build:changelog

> titanium-mobile@11.0.0 build:changelog
> conventional-changelog -n changelog/config.js -i CHANGELOG.md -s -p angular

fatal: ambiguous argument 'undefined...HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Failed to get file. Error: Command failed: git log --cherry-pick --right-only --no-merges undefined...HEAD --format=%H
fatal: ambiguous argument 'undefined...HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

hansemannn: Same as for @caspahouzer!

caspahouzer: I tried with argument branch and tag and release, but maybe I'm just doing it wrong 🥲

caspahouzer:
@drauggres can you support the right cli call to create the changelog?

changelog/config.js
drauggres: Fixed. `eosDate.toISOString().split('T')[0]` looks much better.

package.json
drauggres: > fatal: ambiguous argument 'undefined...HEAD': unknown revision

You must have a local branch with previous version. In this case `10_1_X`

https://github.com/tidev/titanium_mobile/blob/f80e1dc35b7e1d8795fd096db2b350fddc94e88c/changelog/config.js#L86

> Are you sure dateformat is no longer needed?

It's not used anywhere, but this place. I think, fewer dependencies, fewer problems.

drauggres:
> @drauggres can you support the right cli call to create the changelog?

Right call is `npm run build:changelog`, but it expects that you have a local branch with a previous version, see `guessPreviousBranch` function in `config.js`.

There is a comment at the bottom of it, and I agree that tags is a way to go.

As for support from me: sorry, I quit my job and at the moment I don't have Titanium-projects. So this is probably my last PR.

cb1kenobi:
@drauggres I want to personally thank you for your contributions and wish you the best of luck in your career. Take care man!

package.json
m1ga: @drauggres is right: if you have a 10_1_X branch it will generate the changelog. 

So:
```
git checkout 10_1_1_GA
git switch -c 10_1_X
npm run build:changelog
```
looks fine. Still needs some manual adjustments (e.g. one of my PRs is listed under `m1ga` and of course a nice intro text or some breaking changes [shouldn't be any in this version]).

m1ga:
as mentioned in the other post: if you have a local 10_1_X branch it will generate the changelog file
Screenshot_20220521_171738
looks fine to me

caspahouzer:
@m1ga this looks great.

hansemannn:
Then we should be fine here. @cb1kenobi Can you give your bless?

cb1kenobi (approved):
LGTM **APPROVED**

PR #13463, opened 2022-05-25
fix(android): lower SDK level for borderFix

m1ga:
In my old PR https://github.com/tidev/titanium_mobile/pull/13283 ` fix(android): roundBorder fix for Android 12+` I've added a check so the fix will only be used for Android 12+ since in my tests only those were affected by the issue.

In Slack a user pointed out that a Samsung device with Android 11 is having those issues with 11.0.0.RC:
android-11---11 0 0 RC

This PR will lower the SDK version for the patch to Android 11. 

I've tested it in my simulator:
Screenshot_1653485697
and it didn't make a difference for me (which is great) but the Samsung device looks fine now :+1:

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiBorderWrapperView.java
@@ -105,12 +105,12 @@ public class TiBorderWrapperView extends FrameLayout
 			canvas.drawPath(outerPath, paint);
 
 			// TIMOB-16909: hack to fix anti-aliasing
-			if (Build.VERSION.SDK_INT < 31 && backgroundColor != Color.TRANSPARENT) {
+			if (Build.VERSION.SDK_INT < 30 && backgroundColor != Color.TRANSPARENT) {
 				paint.setColor(backgroundColor);
 				canvas.drawPath(innerPath, paint);
 			}
 			canvas.clipPath(innerPath);
-			if (Build.VERSION.SDK_INT < 31 && backgroundColor != Color.TRANSPARENT) {
+			if (Build.VERSION.SDK_INT < 30 && backgroundColor != Color.TRANSPARENT) {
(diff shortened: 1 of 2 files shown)


hansemannn:
Are there official docs regarding the minimum API level for the border behavior? I would like to prevent a guessing here

jquick-axway:
@hansemannn , @m1ga , Google documents which drawing APIs support hardware acceleration here...
https://developer.android.com/guide/topics/graphics/hardware-accel

@m1ga , if the issue goes away by disabling hardware acceleration via `setLayerType(View.LAYER_TYPE_SOFTWARE, null)`, then we know for sure it's a hardware acceleration issue.

m1ga:
I'm not sure if that is a hardware accel issue since the error started to appear on the same device when the Android version was raised. I think there is dark element (shadow or border) that was added in Material?

But I'll test to disable it and see if that makes any differences.

m1ga:
Disabling HW acceleration looks fine too. A quick test/workaround is to set `backgroundColor: "rgba(51,153,153,254)"` (so 254 instead of 255). That will disable HW acceleration.

But I think we shouldn't put too much work in this PR because I would like to get rid of the `TiBorderWrapper` for (most) round views. When I use the `ViewOutlineProvider` right away I'm getting smooth corners and even a nice shadow below the buttons:

https://user-images.githubusercontent.com/4334997/170727589-5473a05e-86ff-4c03-9f56-7145b520538b.mp4

and since Android 21 is the current min level we don't have issues with the support. Only have some issues with borderColor at the moment since that will be behind the children (which I think is the purpose of TiBorderWrapper, that it is a view in front of the actual view).

**So I would say:** 
* use this PR for 11.0.0 (we keep HW acceleration) 
* for 11.1.0 I'll do the PR to ignore `TiBorderWrapper` and use `ViewOutlineProvider` without any custom painting when we have no borderWidth/borderColor.
* I've added a note about the `backgroundColor` workaround to disable the HW accel if you still have dark artifacts in other Android versions

hansemannn (approved):

PR #13472, opened 2022-06-04
feat(ios/android): be able to drag and drop without explicit editing mode

hansemannn:
This change allows developers to enable drag and drop without the need to go into "editing mode". This brings many improvements to the user experience:
- Users don't need a second step (like a "Edit" button to enable dragging)
- The drag bars are not required
- The drag and drop can be performed over the whole view, not just the drag bars

I will try to add Android parity as well. As it seems, Android per-default doesn't event need the drag bars in order to drag and drop, but we limit it to them for now (which is okay due to existing parity). Using the new `requiresEditingToMove`, Android could allow drag and drop on the whole view without handle bars as well :)

 

Example:

```js
const myTemplate = {
	childTemplates: [
		{
			type: 'Ti.UI.Label',
			bindId: 'title',
			properties: {
				left: 15
			}
		}
	]
};

	backgroundColor: '#fff'
});

	templates: { template: myTemplate },
	requiresEditingToMove: false,
	sections: [
		Ti.UI.createListSection({
			headerTitle: 'Section 1',
			items: [
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 1' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 2' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 3' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 4' } }
			]
		}),
		Ti.UI.createListSection({
			headerTitle: 'Section 1',
			items: [
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 1' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 2' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 3' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 4' } }
			]
		})
	]
});

listView.addEventListener('move', event => {
	console.warn(event);
});

win.add(listView);
win.open();
```


hansemannn:
@caspahouzer Would you mind testing this one? :)

hansemannn:
@m1ga Android parity added as well! The only issue I see is that the item height seems to truncate when dragging the cell:

I checked every place where `PROPERTY_EDITING` is used, so it should just work. Will need to check again there.

apidoc/Titanium/UI/ListView.yml
```diff
@@ -675,6 +678,15 @@ properties:
...
     platforms: [android, iphone, ipad, macos]
 
+  - name: requiresEditingToMove
+    summary: Determines if the list view should allow drag-and-drop even without going into the editing mode
+    description: |
+        For more information see the "Editing Support" section of <Titanium.UI.ListView>.
+    type: Boolean
+    default: true
+    since: {iphone: "11.2.0", ipad: "11.2.0", macos: "11.2.0"}
+    platforms: [iphone, ipad, macos]
```
m1ga, line 688: add Android here too

m1ga (approved):
**Android:**
working fine on Android. I've tested:
* requiresEditingToMove true / false
* editing: true
* editing + requiresEditingToMove true/false
all worked as expected

caspahouzer (approved):
Works superb on iOS.

What's tested
requiresEditingToMove: true and false
editing: true
requiresEditingToMove and editing: true and false

Everything works as expected

LGTM

PR #13473, opened 2022-06-04
feat(android): update support libraries

m1ga:
* updating support libraries
* applying some Studio suggestions

Tested with test suite, kitchensink, hyperloop-examples and my app
Preparations for Android 13 support :wink:

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,9 +13,10 @@ buildscript {
 		mavenCentral()
 	}
 	dependencies {
-		classpath 'com.android.tools.build:gradle:7.0.4'
+		classpath 'com.android.tools.build:gradle:7.2.1'
 		classpath 'com.google.gms:google-services:4.3.10'
 		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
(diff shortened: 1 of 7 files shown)


android/build.gradle
```diff
@@ -16,6 +16,7 @@ buildscript {
 		classpath 'com.android.tools.build:gradle:7.0.4'
 		classpath 'com.google.gms:google-services:4.3.10'
 		classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+		classpath 'org.codehaus.groovy:groovy-json:3.0.11'
```
hansemannn, line 19: This looks like a third-party dependency (or even fork?), can we prevent that?

m1ga: this is because of https://github.com/tidev/titanium_mobile/pull/13473/files#diff-29c0a847a07a95fad8d2db898aca7db913cfa4a759621e5e80276111f59b0ac3R129 and was added by Studio when optimizing that line

hansemannn: Okay, if Android Studio added it, it should be fine!

hansemannn (approved):
Looks and works good!

PR #13474, opened 2022-06-04
feat(android): Android 13 updates

m1ga:
* use new `POST_NOTIFICATIONS` to add parity for `remoteNotificationsEnabled`

This will return `true` is you enable push notifications in the app settings:
Screenshot_20220604_122358

even if it is `false` you will still get a token from FCM! It just won't display a push notification.

* updated the target/compile SDK to 33
* updated gradle tools (only for apps, SDK will still fail so I've kept the version and added `suppressUnsupportedCompileSdk=33`)
* updated build/platform tools
* made two fixes for IOExceptions (Studio says "remove them" but gradle will fail if they are not there)
* removed `setAppCacheEnabled` from WebView settings. Was deprecated and is now removed (https://developer.android.com/sdk/api_diff/33/changes/android.webkit.WebSettings#android.webkit.WebSettings.setAppCacheEnabled_removed(boolean))

Building (debug and dist) still works fine. Tested kitchensink, hyperloop examples and one of my apps.

**Note:**
we could use the current gradle setup but it will show a warning that it wasn't tested with the compile/target level. So it makes sense to raise those versions.


android/modules/network/src/java/ti/modules/titanium/network/NetworkModule.java
```diff
@@ -479,6 +484,24 @@ public void removeAllHTTPCookies()
...
+	@Kroll.getProperty
+	public boolean remoteNotificationsEnabled()
+	{
+		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
+			return true;
+		}
+
+		Context context = TiApplication.getInstance();
+		int result = context.checkSelfPermission(Manifest.permission.POST_NOTIFICATIONS);
+		result &= context.checkSelfPermission(Manifest.permission.POST_NOTIFICATIONS);
```
hansemannn, line 501: Silly question, but why check the same `Manifest.permission.POST_NOTIFICATIONS` value twice?

m1ga: 🤦 you are right. Was testing stuff before, will remove that!

hansemannn: Happens!

android/modules/media/src/java/ti/modules/titanium/media/TiThumbnailRetriever.java
```diff
@@ -48,7 +48,9 @@ public void setUri(Uri uri)
 	public void cancelAnyRequestsAndRelease()
 	{
 		task.cancel(true);
-		mMediaMetadataRetriever.release();
+		try {
+			mMediaMetadataRetriever.release();
+		} catch (IOException ex) {}
```
hansemannn, line 53: Would be great to know when it fails, e.g. with an error log

android/titanium/src/java/org/appcelerator/titanium/util/TiMimeTypeHelper.java
```diff
@@ -222,7 +223,9 @@ public static String getMimeType(Uri uri, String defaultType)
 				}
 			} catch (Exception ex) {
 			} finally {
-				mediaRetriever.release();
+				try {
+					mediaRetriever.release();
+				} catch (IOException ex) {}
```
hansemannn, line 228: Same here

m1ga:
Tested it in my firebase test app
```
console.log("remoteNotificationsEnabled", Ti.Network.remoteNotificationsEnabled);
```

The new push implementation is strange :smile:  Even when I have it off I was able to send a push notification :thinking: 

You can request the permission with:
```js
var permissions = ['android.permission.POST_NOTIFICATIONS'];
Ti.Android.requestPermissions(permissions, function(e) {
  if (e.success) {
    Ti.API.info('SUCCESS');
  } else {
    Ti.API.info('ERROR: ' + e.error);
});
```

and after that `Ti.Network.remoteNotificationsEnabled` is true. I'll add that to `fcm.registerForPushNotifications()` so it will ask for permissions. Thought about using `Ti.Network.registerForPushNotifications()` but we don't get a token, just the permission so we still need the fcm part.

m1ga:
Tested it on an Andorid 8 device. `Ti.Network.remoteNotificationsEnabled` returns true (as expected) :+1:

hansemannn:
This PR is great. But it does much more than mentioned in the title. We should therefore either split it up in Android 13 support and push notification changes or change the title to make it more expressive. Also more testing from the community would be great!

m1ga:
you're right! the `remoteNotificationsEnabled` is only one part since we have the new Android 13 permission now. I've change the title.

android/package.json
```diff
@@ -17,11 +17,11 @@
...
 	"minSDKVersion": "21",
-	"compileSDKVersion": "31",
+	"compileSDKVersion": "33",
 	"vendorDependencies": {
-		"android sdk": ">=23.x <=31.x",
-		"android build tools": ">=30.0.2 <=31.x",
-		"android platform tools": "31.x",
+		"android sdk": ">=23.x <=33.x",
+		"android build tools": ">=30.0.2 <=33.x",
+		"android platform tools": "33.x",
```
hansemannn, line 24: Did we really skip 32?!

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -304,12 +304,6 @@ public TiUIWebView(TiViewProxy proxy)
 			settings.setDatabaseEnabled(true);
 		}
 
-		File cacheDir = TiApplication.getInstance().getCacheDir();
-		if (cacheDir != null) {
-			settings.setAppCacheEnabled(true);
-			settings.setAppCachePath(cacheDir.getAbsolutePath());
-		}
```
hansemannn, line 311: Can you add some note why this needs to be removed?

android/package.json
m1ga: 32 is [Android 12L](https://de.wikipedia.org/wiki/Android_12L)

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
m1ga: because Google removed the methods :smile: 
Screenshot_20220818_145657

hansemannn: I'm just wondering what this means for the usage of web views. Maybe it's just something they do internally now? But if it requires permissions for examples, that should be adopted.

m1ga: Found some more infos in Studio:
> The Application Cache API is deprecated and this method will become a no-op on all Android versions once support is removed in Chromium. Consider using Service Workers instead. See https://web.dev/appcache-removal/ for more information.

and there they say 
> In light of that, Chrome WebView will support AppCache without an origin trial until the final removal takes place, expected in Chrome 90.

hansemannn:
Looks fine from my end! Should we also bump some library versions? For example, material components and some other patch-version bumps?

m1ga:
Sorry, missed the notification! Will have a look at it :+1:

hansemannn:
@m1ga I added the parity, updated the docs and fixed the version check (I suppose - as `Manifest.permission.POST_NOTIFICATIONS` should be checked for Android 13+, not Android 12+). Maybe I missed something and it was on purpose, so it would be great if you could check the commit in detail. After that, we should be good to go!

m1ga:
Found two libs that have a newer version. Will test that together. Rest looks update to date (I've checked those in a different PR before)

m1ga:
First quick tests (firebase push test app, my climbing app) looked good: permission check was shown, app works fine (I have a version of my app in store already with the first part of this PR :wink: )
But I'll do some more tests tomorrow.

BTW: since most people will use the FCM module in combination with this PR it will add `POST_NOTIFICATIONS` permission automatically. But it is still good to mention it in the docs :+1:

hansemannn (approved):

hansemannn:
Ready to merge from my end. If there are no other objections, this is good to go!

m1ga:
Merge it :rocket: 
Hopefully more people will test the next RC to get some more feedback from other apps.

PR #13481, opened 2022-06-10
build: wired up regen build registry

cb1kenobi:
This GitHub action should notify the regen-builds action in the downloads-www repo to fetch and update the static build registry JSON files.

### DIFF
--- a/.github/workflows/regen-builds.yml
+++ b/.github/workflows/regen-builds.yml
@@ -5,6 +5,9 @@ on:
   workflow_run:
     workflows: [Build]
     types: [completed]
+    branches:
+    - master
+    - "[0-9]+_[0-9]+_X"
 
 jobs:
   on-success:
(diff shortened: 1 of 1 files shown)


.github/workflows/regen-builds.yml
```diff
@@ -0,0 +1,20 @@
+name: Regen Builds
+on:
+  release:
+    types: [published, released]
+  workflow_run:
+    workflows: [Build]
+    types: [completed]
```
ewanharris, line 7: Maybe a branch restriction for `master or \d_\d_X `(if a regex is allowed) or a check for the branch name might be good? If someone pushes a branch to the repo I think this would currently run for that branch?

cb1kenobi: Fixed. You can't do regexes, but you can do glob patterns!

ewanharris (approved):

PR #13484, opened 2022-06-12
feat: add option to exclude dir(s) from compiling to assests catalog …

mbender74:
…when app-thinning is used

needs:  https://github.com/tidev/node-titanium-sdk/pull/581

tiapp.xml:
```
 <ios>
        <exclude-dir-from-asset-catalog>tiles</exclude-dir-from-asset-catalog>
```

### DIFF
--- a/cli/lib/gather.js
+++ b/cli/lib/gather.js
@@ -302,7 +302,7 @@ class Categorizer {
 					// managed by their bundles.
 
 					if (this.useAppThinning && !relPath.match(BUNDLE_FILE_REGEXP)) {
-						if (this.excludeAssestsDir != null) {
+						if (this.excludeAssestsDir !== null) {
 							const checkRegEx = new RegExp(this.excludeAssestsDir);
 							if (!relPath.match(checkRegEx)) {
 								results.imageAssets.set(relPath, info);
--- a/cli/lib/gather.js
+++ b/cli/lib/gather.js
@@ -271,6 +271,17 @@ class Categorizer {
 						results.launchImages.set(relPath, info);
 						return;
 					}
+					if (this.excludeAssestsDir !== null) {
+						let testPath = this.excludeAssestsDir;
+						const checkRegEx = new RegExp(`${testPath}`);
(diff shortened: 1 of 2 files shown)


cli/lib/gather.js
```diff
@@ -297,14 +300,23 @@ class Categorizer {
 					// if we are using app thinning, then don't copy the image, instead mark the
 					// image to be injected into the asset catalog. Also, exclude images that are
 					// managed by their bundles.
+
 					if (this.useAppThinning && !relPath.match(BUNDLE_FILE_REGEXP)) {
-						results.imageAssets.set(relPath, info);
-						return;
+						if (this.excludeAssestsDir !== null) {
+							const checkRegEx = new RegExp(this.excludeAssestsDir);
```
hansemannn, line 306: Can you fix the regex warning?

mbender74: sure... on it...

mbender74: no regex warning....
also added missing part for "png"
Usage in tiapp.xml (no "/" at the start!!!):

```
<ios>
        <exclude-dir-from-asset-catalog>maps/tiles|maps/otherimages</exclude-dir-from-asset-catalog>
```

mbender74:
@hansemannn yeah.... now I can PR my ti.map stuff.... ;-)

hansemannn:
@mbender74 This PR causes our app to crash with latest master. I think I found the root cause and will setup a PR now.

cli/lib/gather.js
```diff
@@ -268,6 +271,17 @@ class Categorizer {
...
+						let testPath = this.excludeAssestsDir;
+						const checkRegEx = new RegExp(`${testPath}`);
+						if (!relPath.match(checkRegEx)) {
+							results.imageAssets.set(relPath, info);
+							return;
+						}
+					} else {
+						results.imageAssets.set(relPath, info);
+						return;
+					}
```
hansemannn, line 284: This else-case did not exist before, so it should be removed.

PR #13487, opened 2022-06-19
feat(android): blur effect

m1ga:
Adds a simple blur method to Views

```js
const view = Ti.UI.createView();
const lbl = Ti.UI.createLabel({text: "test"});
view.add(lbl);
win.addEventListener("open", e=> {
	view.blur({ x: 10, y:10});
});
win.add(view);
win.open();
```

Blurs all children and is still interactive:

https://user-images.githubusercontent.com/4334997/174476127-0b50394f-c57a-45d9-ad70-f33eb5b6d3bc.mp4

**edit**: misread the min version :smile: It's only Android 12 or later, so I've added a warning when you want to use it before that.

Parity PR welcome :wink:


m1ga:
`toImage` is currently not working. The blur effect won't be visible. Have the same issue with another PR and round borders. So hopefully I can find a way to make it work for both of them

hansemannn:
I think we should add parity on the existing `BlurView` API - I am happy to move over the `Ti.UI.iOS.BlurView` namespace as well and do the docs. For the API itself, an Android-only property for the blurRadius could be added, so it's mostly reusable while maintaining parity

PR #13492, opened 2022-06-24
ci: restrict the regen actions to the tidev org

ewanharris:
This is so that forks don't get failures for these actions

### DIFF
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   regen:
-    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    if: github.repository_owner == 'tidev' && (github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch')
     runs-on: ubuntu-latest
     name: Trigger Regen
 
--- a/.github/workflows/regen-builds.yml
+++ b/.github/workflows/regen-builds.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   on-success:
+    if: github.repository_owner == 'tidev'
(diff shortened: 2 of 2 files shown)


.github/workflows/regen-docs.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   regen:
-    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    if: github.repository_owner == 'tidev' && (github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch')
```
ewanharris, line 11: I _think_ this should work, not sure how fancy the logic you can do in these `if` blocks is

caspahouzer: We can only merge it to see oft it works, right?

ewanharris: I think I could maybe test on my fork by setting the branch I created this PR from as the default branch. I'll give it a try

ewanharris: It looks like this should be ok, [here's](https://github.com/ewanharris/titanium_mobile/actions/runs/2563755926) a run of `Regen Docs` against this branch which didn't run because of the check, and [here's](https://github.com/ewanharris/titanium_mobile/runs/7059991207) the master one that failed

caspahouzer: Then we should merge this. Are you okay with that?

ewanharris: I'm good with merging! Not sure if @cb1kenobi wants a quick peek as he wrote these actions initially

cb1kenobi (approved):
LGTM! **APPROVED**

PR #13493, opened 2022-06-26
fix(android): fix Ti.UI.ListView scroll state restoration

hansemannn:
One note: At least on the Emulator, the list view seems to lose focus after re-rendering the list view. Not sure if thats an Emulator-only issue.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/TiListView.java
@@ -21,6 +21,7 @@ import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.ShapeDrawable;
 import android.graphics.drawable.shapes.RectShape;
+import android.os.Parcelable;
 import android.view.MotionEvent;
 import android.view.View;
 
@@ -738,6 +739,8 @@ public class TiListView extends TiSwipeRefreshLayout implements OnSearchChangeLi
 			this.proxy.fireEvent(TiC.EVENT_NO_RESULTS, null);
 		}
 
+		Parcelable recyclerViewState = recyclerView.getLayoutManager().onSaveInstanceState();
+
 		// Notify adapter of changes on UI thread.
 		this.adapter.update(this.items, force);
 
@@ -778,6 +781,8 @@ public class TiListView extends TiSwipeRefreshLayout implements OnSearchChangeLi
 						}
 					}
 				}
+
+				recyclerView.getLayoutManager().onRestoreInstanceState(recyclerViewState);
 			}
 		});
 	}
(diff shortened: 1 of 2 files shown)


m1ga:
some kind of test code would be very nice next time to test it. I've quickly created this one:
```js
	templates: {
		test: {
			childTemplates: [{
				type: 'Ti.UI.View',
				childTemplates: [{
					type: 'Ti.UI.Label',
					bindId: 'label',
					properties: {
						color: 'black',
						bindId: 'label',
						bottom: 5
					}
				}],
				properties: {
					width: Ti.UI.FILL,
					height: 100,
					cardUseCompatPadding: true,
					backgroundColor: 'white'
				}
			}]
		}
	},
	defaultItemTemplate: 'test'
});
var items = [];

function addItems(adding) {
		var txt = "Row: " + (i + adding);
		items.push({
			label: {
				text: txt
			},
			template: 'test'
		})
	}
	section.setItems(items);
}

win.add(listView);
	addItems(0);

	setTimeout(function(){
		listView.scrollToItem(0,50);
		setTimeout(function(){
			addItems(100)
		}, 2000);
	}, 1000)
})
win.open();
```

Without the PR it will scroll to the top after 3 sec. With the PR it will stay at the same position and just attach the items.

m1ga (approved):
Looks fine, test code works better with the PR and my existing app still works the same

caspahouzer (approved):
No scrolling detected with this fix

LGTM

hansemannn:
We have to revert it I think, the PR was not finished so far. The potential open issue with the missing focus needs to be tested on device. Sorry for the missing code @m1ga, I was about to add it this evening. Its already happening when generating a long list of items and setting the sections again (causing the internal `update` method to be called).

m1ga:
@hansemannn I've tested it on a device (Pixel 4, Android 12) and emulator (Android 13).

hansemannn:
Okay, so the view did not lose focus / turn gray? Then it's all good.

hansemannn:
Minimalistic test case:
```js

	sections: [
		Ti.UI.createListSection({
			items: generateItems()
		})
	]
});

window.add(listView);
window.open();

setTimeout(() => {
	listView.sections[0].items = generateItems();
}, 3000);

function generateItems() {
	return  Array.from(Array(30).keys()).map(num => ({ properties: { title: `Cell ${num + 1}` } }))
}
```

hansemannn:
I guess we are fine then. Cool!

m1ga:
> Okay, so the view did not lose focus / turn gray? Then it's all good.

not sure what you mean by that :smile: Your example looks the same with 11.0.0.RC and the patched SDK on my device, at start and 3 seconds later :smile: It only fixes the scrolling that my example is showing. Rest looks fine

hansemannn:
In my internal example, we have a text field in the item and after hitting the "Return" key, the list seems to lose focus. But it seems okay.

PR #13501, opened 2022-07-09
fix(ios): fix asset catalog regression

hansemannn:
Fixes a regression caused by https://github.com/tidev/titanium_mobile/pull/13484. The else-case did not exist before. Removing it again fixes the crash.


hansemannn:
cc @mbender74

hansemannn:
I hope it's okay to merge - otherwise I can't build with master right now. Please feel free to report any issues that may apply with this change.

PR #13516, opened 2022-07-22
fix(ios): linter fix

m1ga:
@hansemannn @mbender74 can you test this (introduced in https://github.com/tidev/titanium_mobile/pull/13484)? It should fix the warning `Unchanged files with check annotations` that is below every PR now (e.g. https://github.com/tidev/titanium_mobile/pull/13514/files)

If I run
```js
var testPath = "test";
var checkRegEx = new RegExp(`${testPath}`);
console.log("test".match(checkRegEx));


var testPath = "test";
var checkRegEx = new RegExp(testPath);
console.log("test".match(checkRegEx));
```
in nodejs the result is the same

### DIFF
--- a/cli/lib/gather.js
+++ b/cli/lib/gather.js
@@ -273,7 +273,7 @@ class Categorizer {
 					}
 					if (this.excludeAssestsDir !== null) {
 						let testPath = this.excludeAssestsDir;
-						const checkRegEx = new RegExp(`${testPath}`);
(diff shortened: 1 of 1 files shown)


hansemannn:
It still warns for the non-literal regex for me I think

m1ga:
oh my bad, it only checks it here. Then I'll close it :smile:

PR #13518, opened 2022-07-29
docs(all): add Slider style example

### LINKED ISSUE #13520: Android slide does not have property to set left track tint color

### Description

Slider does not have a property to set left track tint color.

Right now there is only property to set right track tint color is trackTintColor BUT can't set left track tint color...

### Solution

Slider should have property to set left track tint color.

### Alternatives

Right now we need to use custom theme to set it's left track color but we can set right track tint color using trackTintColor so there should be a property exist so that we can set left track tint color.

### Platforms

Android

m1ga:
* Adding some info how to style a Slider with Android themes
* added the missing platforms line to `tintColor`. Was implemented 3 years (8.3.0) ago for Android but doesn't show up in the docs

fixes https://github.com/tidev/titanium_mobile/issues/13520

### DIFF
--- a/apidoc/Titanium/UI/Slider.yml
+++ b/apidoc/Titanium/UI/Slider.yml
@@ -402,6 +402,7 @@ properties:
   - name: tintColor
     summary: The color shown for the portion of the progress bar that is filled.
     type: [String, Titanium.UI.Color]
+    platforms: [iphone, ipad, android, macos]
     since: {iphone: "3.1.3", ipad: "3.1.3", android: "8.3.0"}
     osver: {android: {min: "5.0"}}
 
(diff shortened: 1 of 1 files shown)


caspahouzer (approved):
LGTM

apidoc/Titanium/UI/Slider.yml
```diff
@@ -394,6 +402,7 @@ properties:
   - name: tintColor
     summary: The color shown for the portion of the progress bar that is filled.
     type: [String, Titanium.UI.Color]
+    platforms: [iphone, ipad, android, macos]
```
hansemannn, line 405: This should usually not be necessary, since the default is set to match all platforms

m1ga: the Android icon wasn't visible here before, now it is:

Screenshot_20220731_164158

trackTintColor was visible already because it had the platforms array.

m1ga: perhaps it is a different setting but this was the PR before:

https://github.com/tidev/titanium_mobile/pull/11074/files#diff-53e21b71d097276c9bb920965872bc92c416bcc0bbae08c894d8b25029313437R394-R405

and Android didn't show up for `tintColor`

PR #13526, opened 2022-08-04
fix(android): reduce tableView updates

m1ga:
During older TableView tests I saw that this update function
https://github.com/tidev/titanium_mobile/blob/918388a75a98c3f5458d27f6ed035f11883b303e/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java#L532
is called multiple times. The `shouldUpdate` around it doesn't catch all loops.

Putting a log in the update function showed:
* Without the PR: 43 update logs
* With this PR: 23 update logs
for the example below.

**Change:**
Calling `appendRow` from Ti doesn't change. Only the internal `appendRow` calls won't trigger the `update()` during the loop. It is called at the end of the loop anyways: https://github.com/tidev/titanium_mobile/blob/918388a75a98c3f5458d27f6ed035f11883b303e/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java#L532

Also fixed a potential null pointer exception. 

**Demo**
```js
const tableView = Ti.UI.createTableView({});
var tableData = [];

for (var i = 1; i <= 20; i++) {
	var entryRow = Ti.UI.createTableViewRow({height: 40});
	var entryNameLabel = Ti.UI.createLabel({text: "test"});
	entryRow.add(entryNameLabel);
	tableData.push(entryRow);
}
tableView.data = tableData;

win.add(tableView);
win.open();
```

**Note**
I've only tested some demo tableViews since I don't have an app with an advanced TableView :smile: (ListView FTW!)
So if you have an app with some tables where you add and remove data it would be nice to hear if the updates are quicker now and everything still works.


hansemannn:
Pretty smart! Could this also be an issue with list views or are those not affected?

m1ga:
> Pretty smart! Could this also be an issue with list views or are those not affected?

no, the update in the ListView isn't called in a loop as far as I see.

hansemannn:
Any community users that can test this with a more advanced case? We also only use list views

m1ga:
Finally have a client app with some tableviews :smile: 

Added the log again and could see:
```
[INFO]  I/----    : (main) [10343,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
[INFO]  I/----    : (main) [0,12280] would run update, but skip
```

:+1: tableview didn't look different compared to 12.0.0. Sadly not that much data so couldn't test if it will improve the performance much. But at least 10 less updates!

hansemannn:
Can we get this tested by someone with more complex table views? We also only use it for a maximum of like 12 cells

cb1kenobi (approved):
Looks good! **APPROVED**

cb1kenobi:
Tested, didn't crash, lets go!

PR #13528, opened 2022-08-07
fix: move “movestart” event call to a better place

hansemannn:
This improves the trigger time of the event by ~ 500ms which helps a lot to further improve the UI updates that consume this event.

Test case:
```js
const myTemplate = {
	childTemplates: [
		{
			type: 'Ti.UI.Label',
			bindId: 'title',
			properties: {
				left: 15,
				color: 'black'
			}
		}
	]
};

	backgroundColor: '#fff'
});

	templates: { template: myTemplate },
	requiresEditingToMove: false,
	sections: [
		Ti.UI.createListSection({
			headerTitle: 'Section 1',
			items: [
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 1' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 2' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 3' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 4' } }
			]
		}),
		Ti.UI.createListSection({
			headerTitle: 'Section 1',
			items: [
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 1' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 2' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 3' } },
				{ properties: { canMove: true, height: 43 }, title: { text: 'Title 4' } }
			]
		})
	]
});

listView.addEventListener('movestart', () => {
	console.warn('STARTED MOVING GESTURE');
});

listView.addEventListener('moveend', () => {
	console.warn('STOPPED MOVING GESTURE');
});

listView.addEventListener('move', () => {
	console.warn('MOVED TO A DIFFERENT INDEX');
});

win.add(listView);
win.open();
```

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -1102,12 +1102,9 @@ static TiViewProxy *FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoint
   UIDragItem *dragItem = [[UIDragItem alloc] initWithItemProvider:itemProvider];
   dragItem.localObject = identifier;
 
-  return @[ dragItem ];
-}
-
-- (void)tableView:(UITableView *)tableView dropSessionDidEnter:(id<UIDropSession>)session
-{
   [[self proxy] fireEvent:@"movestart"];
+
+  return @[ dragItem ];
 }
 
 - (void)tableView:(UITableView *)tableView dropSessionDidEnd:(id<UIDropSession>)session


iphone/Classes/TiUIListView.m
```diff
@@ -1102,12 +1102,9 @@ - (BOOL)canMoveRowAtIndexPath:(NSIndexPath *)indexPath
...
   dragItem.localObject = identifier;
 
-  return @[ dragItem ];
-}
-
-- (void)tableView:(UITableView *)tableView dropSessionDidEnter:(id<UIDropSession>)session
-{
   [[self proxy] fireEvent:@"movestart"];
+
+  return @[ dragItem ];
```
janvennemann, line 1107: Just wondering, did you give [`dragSessionWillBegin`](https://developer.apple.com/documentation/uikit/uitableviewdragdelegate/2897382-tableview) a try? I think technically it would be more correct to put it there because then it's clear a drag is about to start. Or is that too late as well?

hansemannn: `dragSessionWillBegin` is only called once the actual drag starts to move from it's initial index position (which is much later than we know about its movement now). Regarding `canMoveRowAtIndexPath`: I don't see any relation to the `itemsForBeginningDragSession` delegate in which it is called right now. Can you elaborate on that?

hansemannn: @janvennemann We ok then? Happy to move it to a better place if availble

janvennemann: @hansemannn My bad, the GitHub diff had me confused because it showed the wrong method name. All good then!

janvennemann (approved):

PR #13540, opened 2022-08-22
feat(android): tabbedbar color properties

m1ga:
Adding `textColor` and `selectedTextColor` to the Android tabbedBar so you can style it in code and don't need to use a Theme.

```js
	layout: "vertical"
});

const bb = Ti.UI.createTabbedBar({
	labels: ['Default', 'Tabbed', 'Bar'],
	backgroundColor: '#336699',
	top: 10,
	width: 300,
});

const bb1 = Ti.UI.createTabbedBar({
	labels: ['Selected', 'is', 'different'],
	backgroundColor: '#336699',
	top: 10,
	width: 300,
	selectedTextColor: "yellow"
});

const bb2 = Ti.UI.createTabbedBar({
	labels: ['Text', 'is', 'different'],
	backgroundColor: '#336699',
	top: 10,
	width: 300,
	textColor: "red",
});

const bb3 = Ti.UI.createTabbedBar({
	labels: ['All', 'are', 'different'],
	backgroundColor: '#336699',
	top: 10,
	width: 300,
	textColor: "red",
	selectedTextColor: "yellow"
});
window.add([bb, bb1, bb2, bb3]);
window.open();
```
_check images below_

**Test**
* run the code
* first bar is the reference color
* 2nd row: only the selected color should be different
* 3rd row: only the normal text color should be different
* 4th row: both colors are different
* check in dark and light mode

**iOS node**
I have to check it on iOS. The docs say:
> See the description of [Titanium.UI.ButtonBar](https://titaniumsdk.com/api/titanium/ui/buttonbar.html) for information on styling tabbed bars and buttons bars.

and there are `textColor` and `selectedTextColor` for iOS already. So if that is correct I can add `ios` in the docs too and we have full partiy already

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUITabbedBar.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUITabbedBar.java
@@ -8,10 +8,12 @@ package ti.modules.titanium.ui.widget;
 
 import android.app.Activity;
 import android.content.Context;
+import android.content.res.Configuration;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.view.MenuItem;
 import androidx.annotation.ColorInt;
+
 import com.google.android.material.bottomnavigation.BottomNavigationView;
 import com.google.android.material.color.MaterialColors;
 import com.google.android.material.tabs.TabLayout;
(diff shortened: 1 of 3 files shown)


hansemannn:
There is a slight iOS parity issue where the default selected state (example 3) is red on iOS, but the default tint color on Android.
example

m1ga:
Ah, thank you for the quick test :+1:  
I can match that behavior but it makes me think: should it change the selected color on iOS if I don't specify a selected color or should it stay at the default (in your image: black)?

m1ga:
Matching iOS colors
Screenshot_20220822-141946

hansemannn (approved):

PR #13549, opened 2022-08-29
fix(ios): optimize ListView continuousUpdate

### LINKED ISSUE #13537: iOS: optimize continuousUpdate

### Description

The new `continuousUpdate` works fine! Already changed some marker events to use `scrolling` now! Much more reliable. 

One think we could optimize is: only call the event on iOS when the item or section changed. Thats how Android is doing it but I'm not able to add that to iOS too 😊  So it should only fire once a new item or section is at the top. 

It should be something like this:

```
int lastVisibleItem = -1;
int lastVisibleSection = -1;

if (lastVisibleItem != NUMINTEGER([indexPath row]) || lastVisibleSection != NUMINTEGER([indexPath section])) {
            [self.proxy fireEvent:@"scrolling" withObject:eventArgs propagate:NO];
            lastVisibleItem = NUMINTEGER([indexPath row]);
            lastVisibleSection = NUMINTEGER([indexPath section]);
}
```

at https://github.com/tidev/titanium_mobile/blob/340abdd08d422d0b7b2dca5f5dda797c20549582/iphone/Classes/TiUIListView.m#L1938

Guess its a quick fix by one of the iOS devs 😉 

Firing the event with the same values won't help so it should only fire when something happened during scrolling. And it would be full Android parity! If you have bigger ListItems it will reduce the event callbacks

### Platforms

iOS

m1ga:
Fixes https://github.com/tidev/titanium_mobile/issues/13537

Optimization for the old PR: https://github.com/tidev/titanium_mobile/pull/13095
It will fire less events, only when the item really changes. Android works like this already. Currently iOS will fire `scrolling` even if the same item is still the top item when you scroll. 

**Workaround** is very simple inside JS but not sending the same event multiple times would be better

```js
var toggle = true;
var sections = [];
	title: "change",
	bottom: 20
});
	top: 10,
	right: 10,
	width: Ti.UI.SIZE,
	height: Ti.UI.SIZE
})
	height: Ti.UI.FILL,
	width: Ti.UI.FILL,
	continuousUpdate: true
});

	toggle = !toggle;
	listView.continuousUpdate = toggle
	console.log(listView.continuousUpdate);
})

for (var s = 0; s < 5; s++) {
		headerTitle: 'Section ' + s
	});
	var set = [];
		set.push({
			properties: {
				title: 'Item 0 ' + i
			}
		})
	}
	section.setItems(set);
	sections.push(section);
}

listView.sections = sections;

	console.log("Item", e.firstVisibleItemIndex, e.firstVisibleSectionIndex);
})
win.open();
```

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -79,6 +79,8 @@ static TiViewProxy *FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoint
   BOOL isSearched;
   UIView *dimmingView;
   BOOL isSearchBarInNavigation;
+  int lastVisibleItem;
+  int lastVisibleSection;
 }
 
 #ifdef TI_USE_AUTOLAYOUT
@@ -1969,6 +1971,14 @@ static TiViewProxy *FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoint
           [eventArgs setValue:NUMINTEGER([indexPath section]) forKey:@"firstVisibleSectionIndex"];
           [eventArgs setValue:section forKey:@"firstVisibleSection"];
           [eventArgs setValue:[section itemAtIndex:[indexPath row]] forKey:@"firstVisibleItem"];
+
+          if (lastVisibleItem == [indexPath row] && lastVisibleSection == [indexPath section]) {
+            // skip event - still the same item at the top
+          } else {
+            [self.proxy fireEvent:@"scrolling" withObject:eventArgs propagate:NO];
+            lastVisibleItem = [indexPath row];
+            lastVisibleSection = [indexPath section];
+          }
         } else {
           section = [[self listViewProxy] sectionForIndex:0];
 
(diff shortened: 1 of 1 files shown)


iphone/Classes/TiUIListView.m
```diff
@@ -1969,6 +1971,14 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
           [eventArgs setValue:NUMINTEGER([indexPath section]) forKey:@"firstVisibleSectionIndex"];
           [eventArgs setValue:section forKey:@"firstVisibleSection"];
           [eventArgs setValue:[section itemAtIndex:[indexPath row]] forKey:@"firstVisibleItem"];
+
+          if (lastVisibleItem == [indexPath row] && lastVisibleSection == [indexPath section]) {
+            // skip event - still the same item at the top
+          } else {
+            [self.proxy fireEvent:@"scrolling" withObject:eventArgs propagate:NO];
```
hansemannn, line 1978: Simplify the if-statement to `lastVisibleItem != [indexPath row] || lastVisibleSection != [indexPath section]`?

PR #13550, opened 2022-08-30
feat(all): remove X-Requested-With header from HTTP requests

m1ga:
Removing "twitter workaround" from the SDK that and removed fixed `X-Requested-With` header. 
If needed user can still add it but having it as a default causes more errors.

Old info for Twitter: https://jira-archive.titaniumsdk.com/TIMOB/TIMOB-729

**Test**
run and check for the returned output
```js
var url = "https://httpbin.org/post";
	onload: function(e) {
		Ti.API.info(this.responseText);
	}
});
client.open("POST", url);
client.send();
```

`X-Requested-With": "XMLHttpRequest",` shouldn't be visible after this PR

Requested in Slack: paySmart3D checks for "ajax requests" and didn't work when the header is there.
Feedback from slack: Removing that line made the API call successful.


apidoc/Titanium/Network/HTTPClient.yml
```diff
@@ -391,6 +391,14 @@ properties:
     platforms: [android, iphone, ipad, macos]
     since: {android: 3.4.0}
 
+  - name: ignoreXRequestedWith
+    summary: Disable sending the `X-Requested-With:XMLHttpRequest` header.
+    type: Boolean
+    permission: read-only
+    default: false
+    platforms: [android]
+    since: {android: 12.0.0}
```
m1ga, line 400:
```suggestion
    since: {android: 12.2.0}
```

cb1kenobi (approved):

PR #13560, opened 2022-09-09
feat(android): ignoreLog config

m1ga:
Adds a quick way to filter out lines in the android log using `ti config`

New key: `cli.ignoreLog = []`
if one of the lines is in the log line it will be ignored.

Example:

```js
"cli": {
  "ignoreLog": [
    "Too many Flogger logs received before configuration",
    "Counters: exceeded sample count in FrameTime",
    "OpenGLRenderer: Davey!"
  ]
}
```
will filter out those nasty map logs.

**TODO:**
we don't have a page with the CLI config parameters. Should be somewhere in the `titanium` repo I guess

### DIFF
--- a/android/cli/hooks/run.js
+++ b/android/cli/hooks/run.js
@@ -18,6 +18,7 @@ exports.cliVersion = '>=3.2';
 
 exports.init = function (logger, config, cli) {
 	let deviceInfo = [];
+	const ignoreLog = config.cli.ignoreLog || [];
 
 	cli.on('build.pre.compile', {
 		priority: 8000,
@@ -264,6 +265,13 @@ exports.init = function (logger, config, cli) {
 						// if it begins with something like "E/SQLiteLog( 1659):" it's not a contination, don't log it.
 						} else if (nonTiLogRegexp.test(line)) {
 							return;
+						}
+
+						// ignore some Android logs in info log level
+						for (let i = 0, len = ignoreLog.length; i < len; ++i) {
+							if (line.includes(ignoreLog[i])) {
+								return;
+							}
 						}
 
 						switch (logLevel) {


hansemannn:
I think there are some ignored pattern already - can we add the new one to them?

m1ga:
there is `nonTiLogRegexp` https://github.com/tidev/titanium_mobile/blob/177c6e1a11ca9737b9433506582c9fc9be0847b0/android/cli/hooks/run.js#L244  to regex some non-ti logs but this one is easier to manage and the user should set them himself (like the filter in Android Studio). I don't want to filter them out completly perhaps someone needs them (I don't think so since native users complain about it too https://stackoverflow.com/questions/70214824/too-many-flogger-logs-received-before-configuration-dropping-old-logs :smile: ).

hansemannn:
Okay, then I suppose it's okay.

m1ga:
Ah I understand! Since there aren't any default settings (so no loop for most users) I see it as a trade-off between usability/cleaner output and performance (if there any at all). It shouldn't affect the app since the loop is on the computer.

The workaround is to use Android Studio LogCat and add the filters in there. So you have Studio running AND it loops through its  filters for every line it logs. Guess that is worse for your machine then the CLI output :smile: 

Not seeing those annoying maps logs in debug mode is soooo much better!

hansemannn:
We should provide a default set of ignored lines in the config file for new installations, what do you think?

m1ga:
I've added the key to the config: https://github.com/tidev/titanium/pull/597/files but I think we should leave it empty. Otherwise the user won't see the log and might miss stuff. Not every app has a map so adding that will create an unnecessary check. 

I think the key is to inform the user that it is now possible to remove those logs. But every dev should see them and opt-out.

Not many people complained about those logs :smile:

PR #13567, opened 2022-09-14
fix(android): use maxMemory as property

### LINKED ISSUE #11068: Titanium using wrong javac & dx max memory
**Describe the bug**
I am running the following command to compile an app for Android:
```
ti build -p android -T device
```

The build process runs a bunch of commands. On the bit it builds the java files, on v7.5.1.GA and older, it run a command like so:
```
javac -J-Xmx1024M -encoding utf8 -bootclasspath ...
```

But on v7.5.1.GA and greater (tested on v7.5.2 and v8.0.0), the build command used more memory like so, ignoring the default values of 1GB as mentioned on the docs:
```
javac -J-Xmx3072M -encoding utf8 -bootclasspath ...
```

This meant when building my app, I get the following unuseful error message (regardless of log-level):

```
[ERROR] Failed to compile Java source files:
[ERROR]
[ERROR]
```

(Side note: It would be good to get more information than just this).

When I tried the javac command manually, I found the error:

```
Error occurred during initialization of VM
Could not reserve enough space for object heap
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

I tried adjusting the the amount of memory used like so:

```
ti config android.javac.maxmemory 1024M
```

but it's just ignored it, and still uses 3GB instead.

I solved this issue by adding the following properties to tiapp.xml

```
<property name="android.javac.maxmemory" type="string">1G</property>
<property name="android.dx.maxmemory" type="string">1G</property>
```

and the javac command changed correctly.

So the issues are:

 1. Titanium v7.5.2 and greater does not use the correct memory for javac and dexter
 2. When adjusting the config, the memory used in the build does not change
 3. Un-useful error message in console output when build fails (this could be a separate issue)

**Environment**
Titanium SDK version: >=v7.5.2.GA 
CLI version: v5.2.1
OS: Windows

**Logs**:
No additional messages.

Also mentioned the issue here:
https://jira.appcelerator.org/browse/AC-6315

m1ga:
resolves https://github.com/tidev/titanium_mobile/issues/11068

Using camel-case for `maxMemory` CLI + tiapp.xml property. 

**Before**
tiapp.xml uses `android.javac.maxmemory` and CLI used `android.javac.maxMemory`

**After**
both use `android.javac.maxMemory`

**Test**
* add `<property name="android.javac.maxmemory" type="string">1G</property>` to your tiapp.xml
* it will stop and showing an error that you should use maxMemory
* use `<property name="android.javac.maxMemory" type="string">1G</property>` and it will build fine

Since many users won't use this at all I've used `process.exit(1);` and stopped the build. This way I don't have to support both values.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -879,11 +879,17 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 	cli.tiapp.properties['ti.deploytype'] = { type: 'string', value: this.deployType };
 
 	// Fetch Java max heap size setting.
+	this.javacMaxMemory = config.get('android.javac.maxMemory', '3072M');
+
+	// TODO remove in the next SDK
 	if (cli.tiapp.properties['android.javac.maxmemory'] && cli.tiapp.properties['android.javac.maxmemory'].value) {
-		logger.error(__('android.javac.maxmemory is deprecated. Please use android.javac.maxMemory') + '\n');
-		process.exit(1);
+		logger.error(__('android.javac.maxmemory is deprecated and will be removed in the next version. Please use android.javac.maxMemory') + '\n');
+		this.javacMaxMemory = cli.tiapp.properties['android.javac.maxmemory'].value;
 	}
-	this.javacMaxMemory = cli.tiapp.properties['android.javac.maxMemory'] && cli.tiapp.properties['android.javac.maxMemory'].value || config.get('android.javac.maxMemory', '3072M');
+
+	if (cli.tiapp.properties['android.javac.maxMemory'] && cli.tiapp.properties['android.javac.maxMemory'].value) {
+		this.javacMaxMemory = cli.tiapp.properties['android.javac.maxMemory'].value;
+	}
 
 	// Transpilation details
 	this.transpile = cli.tiapp['transpile'] !== false; // Transpiling is an opt-out process now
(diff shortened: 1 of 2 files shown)


android/cli/commands/_build.js
```diff
@@ -879,7 +879,11 @@ AndroidBuilder.prototype.validate = function validate(logger, config, cli) {
 	cli.tiapp.properties['ti.deploytype'] = { type: 'string', value: this.deployType };
 
 	// Fetch Java max heap size setting.
-	this.javacMaxMemory = cli.tiapp.properties['android.javac.maxmemory'] && cli.tiapp.properties['android.javac.maxmemory'].value || config.get('android.javac.maxMemory', '3072M');
+	if (cli.tiapp.properties['android.javac.maxmemory'] && cli.tiapp.properties['android.javac.maxmemory'].value) {
+		logger.error(__('android.javac.maxmemory is deprecated. Please use android.javac.maxMemory') + '\n');
+		process.exit(1);
```
ewanharris, line 884: I'm not sure how much work extra it might be, but given that this is deprecation rather than straight removal initially, maybe warning and continuing is better here?

m1ga: Then the user won't see it as it is right in the beginning of the output. Had that first and it passed by so quickly :smile: 
But I'm happy to change it to a warning, accept both values and we remove it in a later SDK :+1:

m1ga: I've kept it as an error so it will better to see. But I accept both values now and print the error (without exit) when you use `maxmemory`.

hansemannn (approved):

PR #13582, opened 2022-09-25
fix(ios): fix touch issue on Apple Silicon simulators

hansemannn:
This PR fixes an issue on Apple Silicon Simulator where touches are not received. I would suggest to also take it for the 11.1.1 release if approved.

### DIFF
--- a/iphone/Classes/TiUIButton.m
+++ b/iphone/Classes/TiUIButton.m
@@ -146,7 +146,10 @@
     touchStarted = NO;
     fireEvent = @"touchend";
     if (button.highlighted) {
-      fireActionEvent = [touch tapCount] == 1 ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
+      // NOTE: The "!touch.tapCount" fixes an issue on Apple Silicon Simulator
+      // where touches are not received.
+      BOOL shouldFireClickEvent = [touch tapCount] == 1 || !touch.tapCount;
+      fireActionEvent = shouldFireClickEvent ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
     }
     break;
   case UITouchPhaseCancelled:
--- a/iphone/Classes/TiUIiOSStepper.m
+++ b/iphone/Classes/TiUIiOSStepper.m
@@ -223,7 +223,11 @@
     touchStarted = NO;
     fireEvent = @"touchend";
     if (stepper.highlighted) {
-      fireActionEvent = [touch tapCount] == 1 ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
+      // NOTE: The "!touch.tapCount" fixes an issue on Apple Silicon Simulator
+      // where touches are not received.
+      BOOL shouldFireClickEvent = [touch tapCount] == 1 || !touch.tapCount;
+
+      fireActionEvent = shouldFireClickEvent ? @"click" : ([touch tapCount] == 2 ? @"dblclick" : nil);
     }
     break;
   case UITouchPhaseCancelled:
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
@@ -1579,10 +1579,14 @@ DEFINE_EXCEPTIONS
       [self handleControlEvents:UIControlEventTouchCancel];
     }
 
+    // NOTE: The "!touch.tapCount" fixes an issue on Apple Silicon Simulator
+    // where touches are not received.
+    BOOL shouldFireClickEvent = [touch tapCount] == 1 || !touch.tapCount;
+
     // Click handling is special; don't propagate if we have a delegate,
     // but DO invoke the touch delegate.
     // clicks should also be handled by any control the view is embedded in.
-    if ([touch tapCount] == 1 && [proxy _hasListeners:@"click"]) {
+    if (shouldFireClickEvent && [proxy _hasListeners:@"click"]) {
       if (touchDelegate == nil) {
         [proxy fireEvent:@"click" withObject:evt propagate:YES];
         return;


jasonkneen:
Is this the issue where some click events aren't received but if you change to touchstart it is? As I had this a LOT

jasonkneen:
YES -- was DRIVING ME CRAZY!

Thanks!

narbs:
@hansemannn thank you for addressing this.

I tested Titanium 12.0.0 master branch as at Oct 21 (that had this fix) using Xcode 14.1 RC (14B47) and this is still happening in our application - view click listeners aren't receiving click events and not firing on the ARM simulator.

Installing the same build on a device works fine, but not in the iOS 16.1 ARM simulator on an M1 Mac. So maybe it is still happening... are you still seeing any issues after applying this fix? The first tap works, then the second fails. After a period of time, the tap works again - it's strange.

The same Titanium 12.0.0 master branch build running on Xcode 13.4.1 (13F100) works fine, so it looks like the basic problem is related to Xcode

hansemannn:
Hi there, thanks for the report! Very odd behavior. Unfortunately I can't reproduce it anymore since a while, so I wonder if you have a small (app.js suitable) test case?

narbs:
@hansemannn I have a very large app but that won't work - I will try to reproduce the issue with a simple test case and send it to you next week

narbs:
@hansemannn Ok I put some debug statements in the code before the patch you wrote and the result was surprising:

```objc
    NSLog(@"[WARN] tapCount 1: %d", [touch tapCount]);
    NSLog(@"[WARN] tapCount 2: %d", touch.tapCount);
    NSLog(@"[WARN] button.highlighted: %@", button.highlighted);
```

> [WARN]  tapCount 1: 2
> [WARN]  tapCount 2: 2
> [WARN]  button.highlighted: 1
> [WARN]  tapCount 1: 3
> [WARN]  tapCount 2: 3
> [WARN]  button.highlighted: 1
> [WARN]  tapCount 1: 4
> [WARN]  tapCount 2: 4

So the tapCount is just increasing and increasing, which is why the current code isn't responding.

This result does not happen on the same build on the x86 simulator, only the arm simulator running the 12 dev build (master branch).

After a period of time that is between 10-30s, something resets and the count goes back to 1 again:

[WARN]  tapCount 1: 1
[WARN]  tapCount 2: 1
[WARN]  button.highlighted: 1

I could of course hack the code to look for 1 tap (click), 2 taps (double-click) or >2 taps (click) events but that seems hacky and wouldn't work well for double-click events. 

Question is, why is that count just increasing?

Hope this helps. I haven't got a sample project yet - this came from the app we write, which is very large.

If you still need a sample project I can try to reproduce it, but at least I've found the symptom.

I tested it on a menu button on a view that causes the view to animate out after being tapped, but it happens on view clicks as well when there is no associated animation.

The rest of the GUI is responsive but the same button/click listener tapped again is returning the increasing tapCount, and it only resets after a short timeout on the ARM simulator.

narbs:
@hansemannn Just completed testing with Xcode 14.1 (14B47b) on Intel x86 Simulator and the tapCount as above always returns 1, so there is a bug in the ARM simulator where the touch.tapCount returned from:

`UITouch *touch = [[event allTouches] anyObject];`

In the ARM simulator is somehow retained and increased until some moment when it resets.

My application doesn't use double-click ever so I can get around this annoying bug by only looking for single clicks (tapCount <> 0) so that's what I've done but unless we can somehow reset the ARM simulator touch handler each time a touch is received, this bug will get in the way of usability.

Because it is possible to detect the ARM simulator, maybe a patch can be made for ARM only - this bug doesn't occur on ARM devices. I will keep working on it.

hansemannn:
Thanks for the insight! This seems related: https://developer.apple.com/forums/thread/668488 Maybe there is more?

dlewis23:
So I have this issue also. I updated to Xcode 14.x two days ago and the problem started. I did a lot installs/reinstalls of Xcode to try and figure it out. Setup a test app, for me the issue happens when I have a module in the app that does not have arm64 simulator support, like Ti.Storekit. If I remove the storekit module (or any module that does not have arm64 simulator support) the problem goes away. 

I setup this simple test app. Build and click and all click events work, add the ti.storekit module to your tiapp.xml build to simulator on an Apple Silicon Mac and clicks don't get registered. touchstart or touchend end do, but click event do not. Remove ti.storekit from the tiapp.xml build and everything works perfect again.   

`var count = 0;

    title: 'Window',
    backgroundColor: '#fff'
});

    text: 'Hello',
});

win.add(label);

    title: 'Click Me!',
    top: 150,
    width: 100,
    height: 50
});

button.addEventListener('click',function(e) {
    count++;
    Titanium.API.info("You clicked the button: " + count);
});

button.addEventListener('longpress',function(e) {
    Titanium.API.info("You longpressed the button");
});

win.add(button);
win.open();
`

hansemannn:
I would vote for reverting this PR until Apple fixes the Simulators, as it's a development-only issue and better there instead of production apps consuming too many events. Is that ok?

dlewis23:
@hansemannn This was all stuff that was changed in 11.1.1.GA right? If I go to 11.1.0.GA it does not correct the issue for me when there is a module that does not have arm64 simulator support and Xcode 14.x

hansemannn:
Yep, but it will be either or, until all modules are updated

dlewis23:
But as far as I can tell neither version, 11.1.1.GA or 11.1.0.GA work with Xcode 14.1 Simulator even when launching it under Rosetta. I tired a few of the work arounds from the apple developer link you posted and none worked for me. 

I would guess that Apple sees no issue and will never fix the simulators.

hansemannn:
We are working with Xcode 14.x since quite a while without these issues

dlewis23:
But now it seems like running in Rosetta doesn't fix the issue in the simulator anymore. Seems like since Xcode 14 went out of beta is when the issue started.

PR #13583, opened 2022-09-29
feat(android): lineCount/visibleText for Ti.UI.Label

m1ga:
Exposing two new parameters in a Ti.UI.Label:
* lineCount: returns the lines of the label
* visibleText: returns the visible text

Can be used in combination with `ellipsize: true` to determine if the text is fully visible or not.

**Example**

```js
var lbl = Ti.UI.createLabel({
	text: "asd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd 111111 fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdf",
	maxLines: 10,
	ellipsize: true
});
win.add(lbl);
setTimeout(function() {
	console.log(lbl.lineCount);
	console.log(lbl.text == lbl.visibleText ? "full text" : "ellipsize");
}, 2000)

win.open();
```

-> returns 4 and `full text`. So the label will use only 4 of the max 10 lines and it shows the full text

```js
var lbl = Ti.UI.createLabel({
	text: "asd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd 111111 fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdfasd fasd fasd fsdf",
	maxLines: 2,
	ellipsize: true
});
win.add(lbl);
setTimeout(function() {
	console.log(lbl.lineCount);
	console.log(lbl.text == lbl.visibleText ? "full text" : "ellipsize");
}, 2000)

win.open();
```

-> returns 2 and `ellipsize`. The `visibleText` only returns the actual text in the label.

**TODO**
* and I'm not sure if there is something like this for iOS, so parity would be nice


hansemannn:
Possible iOS parity? https://stackoverflow.com/questions/28108745/how-to-find-actual-number-of-lines-of-uilabel

m1ga:
@hansemannn  any help with parity here? Would be a great PR to have

apidoc/Titanium/UI/Label.yml
```diff
@@ -175,6 +175,13 @@ properties:
     type: Number
     since: "4.1.0"
 
+  - name: lineCount
+    summary: Returns the amount of lines the content is acually using. Is equal or lower than `maxLines`.
+    platforms: [android]
+    type: Number
+    permission: read-only
+    since: {android: "12.0.0"}
```
m1ga, line 183:
```suggestion
    since: {android: "12.3.0"}
```

hansemannn:
@m1ga Regarding parity:
- `lineCount` is relatively easy -> https://stackoverflow.com/questions/28108745/how-to-find-actual-number-of-lines-of-uilabel
- `visibleText` is more tricky and error prone -> https://stackoverflow.com/questions/11155787/how-to-know-if-nsstring-fits-in-uilabel-or-not-and-index-of-the-last-string-whic/11158418#11158418

m1ga:
If l`lineCount` won't take long to implement then that would be already a great addition! Saw some apps that use the font size calculation + character count to get a rough line count (handling the font as a mono-spaced font). 
And since Android has a property for it to get the exact line count I would prefer to have it inside of Ti so people won't need to calculate it. 
We can skip the `visibleText` part. Didn't see a request for it, just added it because it was a native property too.

hansemannn:
`lineCount` added ✅

m1ga:
iOS version works fine too.
Simulator Screenshot - iPhone 15 Pro Max - 2023-12-31 at 15 35 19

returned `5` for lineCount :+1:

hansemannn (approved):

PR #13585, opened 2022-09-30
fix(android): github build action

m1ga:
Brew is not installed by default anymore (https://github.com/actions/runner-images/issues/6283)
using https://github.com/Homebrew/actions/tree/master/setup-homebrew to add it back again


hansemannn:
The unit tests look fine! Any open issues you have here? Otherwise it's good to go!

m1ga:
nope, can be merged :+1:

hansemannn (approved):

PR #13591, opened 2022-10-07
fix: guard “example” directory and LICENSE during module build

hansemannn:
Fixes an issue where a missing "example" directory or "LICENSE" file would cause a build interruption.


m1ga (commented):
LGTM. Working fine with a test project with and without example folder or license file

PR #13592, opened 2022-10-07
feat(ios): be able to detect an emulated iOS app on Apple Silicon

hansemannn:
Example code to test (compile an app and run it on an Apple Silicon mac, e.g. via Testflight):
```js
    backgroundColor: '#fff'
});

    title: 'Check if emulated'
});

    alert(`Is emulated iOS app on Apple Silicon: ${Ti.Platform.isAppleSiliconTranslated}`);
});

win.add(btn);
win.open();
```

### DIFF
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -273,8 +273,8 @@ properties:
 
   - name: isRunningOnAppleSilicon
     summary: |
-        A Boolean value that indicates whether the current apps is emulated
-        via an Apple Silicon device.
+        A Boolean value that indicates whether the current app is running as a translated
+        binary using Rosetta on an Apple Silicon device.
     description: |
         Use this utility property to guard API's that may not be available on a non-iOS
         device, e.g. motion activity or some authorization components.
--- a/apidoc/Titanium/Platform/Platform.yml
+++ b/apidoc/Titanium/Platform/Platform.yml
@@ -271,6 +271,18 @@ properties:
     permission: read-only
     since: "7.0.0"
 
+  - name: isRunningOnAppleSilicon
+    summary: |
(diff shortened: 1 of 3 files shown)


hansemannn:
@janvennemann Would you mind checking the proxy part? The logic itself is from Apple :)

iphone/Classes/PlatformModule.m
```diff
@@ -256,6 +256,21 @@ - (NSString *)identifierForAdvertising
 }
 #endif
 
+- (bool)isRunningOnAppleSilicon
```
janvennemann, line 259: Maybe name it `isTranslatedBinaryOnAppleSilicon` to indicate that it refers to the Rosetta translation and not just whether the App is running on an Apple Silicon device?

apidoc/Titanium/Platform/Platform.yml
```diff
@@ -271,6 +271,18 @@ properties:
     permission: read-only
     since: "7.0.0"
 
+  - name: isRunningOnAppleSilicon
+    summary: |
+        A Boolean value that indicates whether the current apps is emulated
+        via an Apple Silicon device.
```
janvennemann, line 277:
```suggestion
        A Boolean value that indicates whether the current app is running as a translated
        binary using Rosetta on an Apple Silicon device.
```

janvennemann (changes requested):
Just some small notes regarding naming and the method summary in the docs.

I'm also wondering if it would make sense to have this in an iOS only namespace? The is `Ti.Platform.Android` for Android specific device infos and this likely falls in the same category, just for iOS, right?

janvennemann (approved):
Thanks! Yeah, makes sense to only go through the effort of creating the new namespace when there are actually more iOS only utils.

PR #13595, opened 2022-10-07
chore: clean up default app template

hansemannn:
- Add more spacing to the default app icon (this bugged me for years now 😄)
- Add proper format to the default app.js and use more semantic colors
- Make README more readable (add .md suffix for better editor support, add more useful content)


templates/app/default/template/Resources/app.js
```diff
@@ -1,47 +1,43 @@
...
-        icon: icon,
-        window: win
-    });
-
-    return tab;
+    return Ti.UI.createTab({
+        title,
+        icon,
+        window
+    });;
```
m1ga, line 42: remove one `;`

I would prefer adding the parameter names like `title: title, icon: icon, window: window` since it is a bit easier to understand/quicker to see what it is doing. Both work fine

hansemannn: In some linting rules, this is detected as bad style IIRC, so I added it this way. Not sure, though. @janvennemann what do you think?

hansemannn: @m1ga I restored the changes as you suggested to prevent the PR from being halted.

m1ga (approved):
LGTM!

PR #13028, opened 2021-08-17
feat(android): initial support for Material You themeing

garymathews:
- Initial support for Material You dynamic theming
- Colors obtained from [platform/frameworks/.../colors_device_defaults.xml](https://android.googlesource.com/platform/prebuilts/fullsdk/platforms/android-31/+/refs/heads/master/data/res/values/colors_device_defaults.xml) to match Google's behavior

##### NOTES
- This does not modify the border radius of _Cards_, _Dialogs_, _ProgressBars_. I expect a revision of Material Components
to support the new aesthetic in Android 12

##### TEST CASE
<details>
<summary>SHOW</summary>
<p>

```JS
const win = Ti.UI.createWindow();
const table = Ti.UI.createTableView({ touchFeedback: false });

function section_dayNight() {
    const row = Ti.UI.createTableViewRow();
    const section = Ti.UI.createTableViewSection();
    const view = Ti.UI.createView({
        layout: 'horizontal',
        width: Ti.UI.SIZE,
        height: Ti.UI.SIZE
    });
    const center = Ti.UI.createView({
        height: 60
    });
    const label = Ti.UI.createLabel({
        text: 'Toggle dark theme: '
    });
    const toggle = Ti.UI.createSwitch({
        value: false
    });

    toggle.addEventListener('change', e => {
        if (Ti.UI.overrideUserInterfaceStyle != Ti.UI.USER_INTERFACE_STYLE_DARK) {
            Ti.UI.overrideUserInterfaceStyle = Ti.UI.USER_INTERFACE_STYLE_DARK;
        } else {
            Ti.UI.overrideUserInterfaceStyle = Ti.UI.USER_INTERFACE_STYLE_UNSPECIFIED;
        }
    });

    view.add([label, toggle]);
    center.add(view);
    row.add(center);
    section.add(row);

    return section;
}

function section_buttons() {
    const title = Ti.UI.createLabel({
        text: 'Button',
        left: 10,
        font: {
            fontSize: 26
        }
    });
    const row = Ti.UI.createTableViewRow();
    const section = Ti.UI.createTableViewSection({
        headerView: title
    });
    const view = Ti.UI.createView({
        layout: 'vertical',
        width: Ti.UI.SIZE,
        height: Ti.UI.SIZE,
        bottom: 10
    });

    view.add([
        Ti.UI.createButton({
            top: 10,
            title: 'BUTTON_STYLE_OPTION_POSITIVE',
[...]

PR #13597, opened 2022-10-08
feat(android): material you themes

m1ga:
Creating 3 new themes:
* Theme.Titanium.Material3.DayNight
* Theme.Titanium.Material3.DayNight.NoTitleBar
* Theme.Titanium.Material3.DayNight.Fullscreen
in order to use Material3 Dynamic Color themes. 

Adding `Ti.UI.Android.harmonizedColor()` to find a "harmonizing color"

```js
const win1 = Ti.UI.createWindow({});
const win2 = Ti.UI.createWindow({});

win1.add(Ti.UI.createButton({title: 'I am button.'}));
win2.add(Ti.UI.createButton({title: 'I am button.'}));

var tab1 = Ti.UI.createTab({
    window: win1,
    title: 'Blue',
		icon: "/appicon.png"
}),
tab2 = Ti.UI.createTab({
    window: win2,
    title: 'Red',
		icon: "/appicon.png"
}),
    tabs: [tab1, tab2],
		shiftMode: false,
		style:  Ti.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION,
		theme:"Theme.Titanium.Material3.DayNight.NoTitleBar"
});
tabGroup.open();
```

Screenshot_20221008-195624

```js
const win1 = Ti.UI.createWindow({});
var col1 = "#044";
var col2 = Ti.UI.Android.harmonizedColor(col1);
var col3 = Ti.UI.Android.harmonizedColor(col2);
var v1 = Ti.UI.createView({backgroundColor: col1, width: 100, height: 100, left: 0})
var v2 = Ti.UI.createView({backgroundColor: col2, width: 100, height: 100, left: 100})
var v3 = Ti.UI.createView({backgroundColor: col3, width: 100, height: 100, left: 200})
win1.add([v1,v2,v3]);
win1.open();
```

Screenshot_20221008-201949

**Alternative**
you can also create an own theme with `<style name="AppTheme" parent="Theme.Material3.DynamicColors.DayNight" />`

**Info**
The other PR #13028 maps dynamic colors to existing Ti colors with the OLD Material themes. Since there are more color slots now (see https://material.io/blog/migrating-material-3 `Colors and Themes`) it won't use e.g. `colorPrimaryContainer` with the old PR resulting in some different colors. I think if you want to use dynamic colors you want to use Material 3 themes too. So with this PR it will do that.

**Todo**
Add it to the titanium-docs Android theme list.

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
@@ -6,27 +6,15 @@
  */
 package ti.modules.titanium.media;
 
-import android.Manifest;
-import android.app.Activity;
-import android.content.ClipData;
-import android.content.ContentResolver;
-import android.content.ContentValues;
-import android.content.Context;
-import android.content.Intent;
-import android.content.pm.PackageManager;
-import android.graphics.BitmapFactory;
-import android.hardware.Camera;
-import android.hardware.Camera.CameraInfo;
-import android.media.CamcorderProfile;
-import android.media.MediaMetadataRetriever;
-import android.net.Uri;
-import android.os.Build;
-import android.os.Environment;
-import android.os.Handler;
-import android.os.Message;
-import android.os.Vibrator;
-import android.provider.MediaStore;
-import android.view.Window;
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.InputStream;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.HashMap;
 
 import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollFunction;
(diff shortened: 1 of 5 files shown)


android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
```diff
@@ -867,6 +869,14 @@ private static void ensureExternalPublicMediaDirectoryExists()
 		}
 	}
 
+	@Kroll.method
+	public String harmonizerColor(String value)
```
hansemannn, line 873: I would rather see this under `Ti.UI` or `Ti.UI.Android`, what do you think? Also, I would prefer `harmonizedColor`

m1ga: both valid points :+1: will move it

hansemannn (approved):

hansemannn:
Just to be sure: [This](https://github.com/tidev/titanium_mobile/commit/0c99bfe613dc161b95e853db228e7d811a80647a) were the changes for the last major theme changes (for dark mode support), maybe it includes something else useful. But since this one will not be the default new apps (yet), I think we're good.

m1ga:
Exactly. If we move this to a main theme we have to (perhaps) do some changes. But since its a Material default theme now it might only be moving the main Titanium theme to a different parent.

PR #13599, opened 2022-10-09
chore: do not post app info

hansemannn:
We saw some very rare ANR's regarding this call and since APSAnalytics is ununused anyways, I think it's okay to remove the call entirely.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
@@ -21,9 +21,6 @@ import android.view.accessibility.AccessibilityManager;
 
 import androidx.annotation.NonNull;
 
-import com.appcelerator.aps.APSAnalytics;
-import com.appcelerator.aps.APSAnalyticsMeta;
-
 import org.appcelerator.kroll.KrollApplication;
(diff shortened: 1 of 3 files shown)


android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -442,28 +439,6 @@ public void onTrimMemory(int level)
...
-		if (buildType != null && !buildType.equals("")) {
-			APSAnalyticsMeta.setBuildType(buildType);
-		}
-
-		APSAnalyticsMeta.setAppId(this.appInfo.getId());
-		APSAnalyticsMeta.setAppName(this.appInfo.getName());
-		APSAnalyticsMeta.setAppVersion(this.appInfo.getVersion());
-		APSAnalyticsMeta.setDeployType(deployType);
-		APSAnalyticsMeta.setSdkVersion(getTiBuildVersion());
-		APSAnalytics.getInstance().setMachineId(this);
```
m1ga, line 464: can you check if `Ti.Platform.id` is still working. That is returning `APSAnalytics.getInstance().getMachineId()` and it looks like this is the place it gets set.

hansemannn:
Let's leave it for now. May be to sensitive to quickly change!

PR #13600, opened 2022-10-10
feat(android): support for view animation with xml

Brianggalvez:
Heavily inspired in: https://github.com/tidev/titanium_mobile/pull/13561

put 
`<?xml version="1.0" encoding="utf-8"?>
<set xmlns:android="http://schemas.android.com/apk/res/android">
    <scale android:duration="200" android:fromXScale="0.9" android:toXScale="1.0" android:fromYScale="0.9" android:toYScale="1.0" android:pivotX="50%" android:pivotY="50%" />
</set>`

in 

`platform/android/res/anim/bounce.xml`

now, in any View you can call: 

`view.animate(
    {
      animation: Titanium.App.Android.R.anim.bounce,
      autoreverse: true,
    },
    () => {
      console.log("animation completed");
    }
  );`


m1ga:
Nice one :+1: Can you try to keep the spacing of the old yml files so they are in line with the other yml files. And it will only show your changes at the end ;-)

PR #13605, opened 2022-10-13
fix(android): code optimizations

m1ga:
splitting up https://github.com/tidev/titanium_mobile/pull/13373/ in easier parts

* using `+` instead of StringBuilders (Android Studio recommendation)
* reduce redundant casts 
* moved some `Log.i` to `Log.d` or `Log.w`
* changed tolower + equal to `equalsIgnoreCase`   (Android Studio recommendation)
* removed some duplicated code that was called in IF and ELSE to be after the if/else

Tested Kitchensink, Hyperloop examples and a bigger app with multiple modules.

Test suite: 
`5156 Total Tests: 5150 passed, 6 failed, 0 skipped.`
gps was enabled instead of disabled (2 failed test) and my locale was different (3 failed tests) and the sessionId error (was there before)


android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
```diff
@@ -126,7 +126,7 @@ public String generateRRULEString()
 							weeklyRecurrencesString.append(",");
 						}
 					}
-					finalRRule.append(weeklyRecurrencesString.toString());
+					finalRRule.append(weeklyRecurrencesString);
```
hansemannn, line 129: `weeklyRecurrencesString` is a `StringBuilder`, so I'm curious if this parses to a `String` automatically these days

android/kroll-apt/src/main/java/org/appcelerator/kroll/annotations/generator/KrollJSONGenerator.java
```diff
@@ -509,7 +507,7 @@ protected void visitDynamicProperty(AnnotationMirror annotation, ExecutableEleme
 			Map<Object, Object> dynamicProperty = new HashMap<>(params);
 
 			String methodName = utils.getName(element);
-			String defaultName = new String(methodName);
+			String defaultName = methodName;
```
hansemannn, line 510: Wondering why this was done in the first place …

android/modules/calendar/src/java/ti/modules/titanium/calendar/RecurrenceRuleProxy.java
m1ga: https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#append(java.lang.StringBuffer)
StringBuilder is a valid append parameter

android/kroll-apt/src/main/java/org/appcelerator/kroll/annotations/generator/KrollJSONGenerator.java
m1ga: don't know :) Git blame says it has been there for 11 years :rofl:

hansemannn: Hi @marshall! We missed you during at the Appc alumni meeting last week :) Sorry for dragging you into this, but do you know by any chance if there was a specific reason using the `String` constructor here? I know it's been super long, so I don't expect an answer. Thanks in any case!

marshall: there is definitely no accounting for this spurious line of code.

my best guess.. a naive attempt to clone by Value??

sorry I missed the get together, I would have loved to be there, but I live in TX :))

hansemannn: Thanks! Then I think it's good to go (since it was even recommended by Android Studio). Regarding the get together: Next time!

PR #13614, opened 2022-10-19
docs: correct method name for isTranslatedBinaryOnAppleSilicon

janvennemann:
This fixes the docs for `isTranslatedBinaryOnAppleSilicon` to use the correct method name. We missed that in #13592 after implementing review feedback.


cb1kenobi (approved):
LGTM!

hansemannn (approved):

PR #13624, opened 2022-10-28
fix(android): add null-guard for activity in listview

hansemannn:


m1ga (approved):
Looking good and makes sense.

PR #13629, opened 2022-10-30
feat(android): add FloatingActionButton

m1ga:
Add support for Androids FAB (floating action button)

Screenshot_20221030-193003

```js
const fab = Ti.UI.Android.createFloatingActionButton({
	bottom: 10,
	right: 110,
	image:"/appicon.png",
	iconSize: "mini"
});

const fab2 = Ti.UI.Android.createFloatingActionButton({
	bottom: 10,
	right: 10,
	customSize: 100,
	maxImageSize: 50,
	image:"/appicon.png",
});

const fab3 = Ti.UI.Android.createFloatingActionButton({
	bottom: 10,
	right: 170,
	image:"/appicon.png",
});
fab3.addEventListener("click", function(e){
	console.log("click");
	fab3.backgroundColor = "blue";
})

win.add([fab,fab2, fab3]);
win.open();
```

**Current properties:**
* customSize
* iconSize: mini (preset)
* maxImageSize: the maximum size of the image inside the FAB
* image: the icon

**Note:**
I will just do the normal FAB for now. The extended one (icon + text) can already be done with the normal button but can be extended in the future.


android/modules/ui/src/java/ti/modules/titanium/ui/android/FabProxy.java
```diff
@@ -0,0 +1,29 @@
...
+import android.app.Activity;
+
+import org.appcelerator.kroll.annotations.Kroll;
+import org.appcelerator.titanium.proxy.TiViewProxy;
+import org.appcelerator.titanium.view.TiUIView;
+
+import ti.modules.titanium.ui.widget.TiUIFab;
+
+@Kroll.proxy(creatableInModule = AndroidModule.class)
+public class FabProxy extends TiViewProxy
```
hansemannn, line 18: We should handle it as a `style` property in the existing `Button` proxy or call it `FloatingActionButton`

m1ga: I think I'll rename it to `FloatingActionButton` since its an own element in Android too it shouldn't be mixed with the button. Otherwise we have many non-parity properties

m1ga: done :)

hansemannn:
Is there also a property for the icon tint? See [this discussion](https://stackoverflow.com/questions/31113819/set-fab-icon-color) for reference.

m1ga:
I have to test those. Since it is extending the normal View the feedback parts might already work.
But I'll check that over the weekend

hansemannn:
@m1ga Is this ready for review?

apidoc/Titanium/UI/Android/FloatingActionButton.yml
```diff
@@ -0,0 +1,67 @@
+---
+name: Titanium.UI.Android.FloatingActionButton
+summary: Androids Floating Action Button (FAB)
```
hansemannn, line 3:
```suggestion
summary_ A floating action button (FAB) is a circular button that triggers the primary action in your app's UI.
```
(from the Android docs)

apidoc/Titanium/UI/Android/FloatingActionButton.yml
```diff
@@ -0,0 +1,67 @@
...
+        You can use `customSize` the set a custom width/height and `maxImageSize` to resize the
+        image inside the button.
+    type: String
+    permission: read-only
+    since: "12.0.0"
+
+examples:
+  - title: Show a floating action button (FAB)
+    example: |
+        Show a floating action button (FAB).
```
hansemannn, line 56: Remove duplicate content

hansemannn (changes requested):
Minor docs changes

apidoc/Titanium/UI/Android/FloatingActionButton.yml
m1ga: :+1:

hansemannn (approved):

PR #13640, opened 2022-11-11
feat(modules): update example and readme

m1ga:
* change the readme to a Markdown file
* remove the old marketplace/gittio link
* update the module default example

### DIFF
--- a/templates/module/default/template/example/app.js.ejs
+++ b/templates/module/default/template/example/app.js.ejs
@@ -11,7 +11,7 @@ win.add(label);
 win.open();
 
 // TODO: write your module tests here
-const <%- moduleNameJSSafe %> = require('<%- moduleId %>');
+import <%- moduleNameJSSafe %>  from '<%- moduleId %>';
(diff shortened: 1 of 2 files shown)


templates/module/default/template/example/app.js.ejs
```diff
@@ -5,15 +5,13 @@
...
-});
-var label = Ti.UI.createLabel();
+const win = Ti.UI.createWindow();
+const label = Ti.UI.createLabel();
 win.add(label);
 win.open();
 
 // TODO: write your module tests here
-var <%- moduleNameJSSafe %> = require('<%- moduleId %>');
+const <%- moduleNameJSSafe %> = require('<%- moduleId %>');
```
hansemannn, line 14: Maybe even use modern imports?
```js
import <%- moduleNameJSSafe %>  from '<%- moduleId %>';
```

m1ga: sure, can do that too.

PR #13652, opened 2022-11-23
fix(ios): fix dimBackgroundForSearch in tableview

### LINKED ISSUE #13303: iOS: dimBackgroundForSearch on TableView has no effect

### Description

The dimBackgroundForSearch has ne effect on a TableView. The background is dimmed anyway.

Maybe related to https://github.com/tidev/titanium_mobile/issues/13246

### Expected Behavior

The background dimming should not be visible

### Actual behavior

The background dimming is visible when dimBackgroundForSearch=false

### Reproducible sample

```
const win = Ti.UI.createWindow({title: 'Window'});
var newData = Ti.UI.createTableViewRow({title: 'element 1'})
var newData2 = Ti.UI.createTableViewRow({title: 'row 2'})
var newData3 = Ti.UI.createTableViewRow({title: 'item 3'})
        dimBackgroundForSearch: false,
	search: Ti.UI.createSearchBar(),
	data: [newData, newData2, newData3]
});
win.add(table);
win.open();
``` 

### Steps to reproduce

- Start the example
- search for something

### Platform

iOS

### SDK version you are using

10.1.1.GA

m1ga:
Fixes https://github.com/tidev/titanium_mobile/issues/13303

```js
const win = Ti.UI.createWindow({title: 'Window'});
var newData = Ti.UI.createTableViewRow({title: 'element 1'})
var newData2 = Ti.UI.createTableViewRow({title: 'row 2'})
var newData3 = Ti.UI.createTableViewRow({title: 'item 3'})
        dimBackgroundForSearch: false,
	search: Ti.UI.createSearchBar(),
	data: [newData, newData2, newData3]
});
win.add(table);
win.open();
````

Overlay shouldn't be visible.

**Note:** 
I'm not sure if this is the best fix :wink: I only added some logs and saw that:
https://github.com/tidev/titanium_mobile/blob/master/iphone/Classes/TiUITableView.m#L1865-L1867
is called before the value is set here (so it is always YES):
https://github.com/tidev/titanium_mobile/blob/master/iphone/Classes/TiUITableView.m#L1879-L1883
so it will always create it, even if you say false. If its false I just set it to nil again.

### DIFF
--- a/iphone/Classes/TiUITableView.m
+++ b/iphone/Classes/TiUITableView.m
@@ -1875,6 +1875,7 @@
     [self createDimmingView];
   } else {
     _dimsBackgroundDuringPresentation = [TiUtils boolValue:arg def:YES];
+    dimmingView = nil;
   }
 }
 


hansemannn:
IIRC, this caused another glitch / crash regarding the deallocated view again. But if it works for you and the test cased, I'd be fine to accept it!

m1ga:
The dimming view itself is still behaving weird (like wrong position, [other PR](https://github.com/tidev/titanium_mobile/issues/13246#issuecomment-1221387587)) but at least you can disable it now again! 
I've only tested it with the examples in this other ticket: https://github.com/tidev/titanium_mobile/issues/13246#issuecomment-1221387587

I'll try to get the Slack user to test it. My short examples don't crash but they don't do much :smile:

hansemannn (approved):

PR #13657, opened 2022-12-02
feat(android): parity for WebView.createPDF()

m1ga:
Adds the `createPDF()` method to the Webview. iOS already has it. 

**Attention:**
Since it works a bit differently I have to pass an object to it with more parameters.

**Example**
* run the code
* wait till the page is loaded
* press the "create pdf" button
* log will print pdf content

```js
const www = Ti.UI.createWebView({url: "https://tidev.io"});
const btn = Ti.UI.createButton({title: "create pdf", top: 10});
window.add([www, btn]);
window.open();

	if (OS_ANDROID) {
		www.createPDF({
			pageWidth: 5000,
			pageHeight: 72000,
			success: function(e) {
				console.log(e.data);
			}
		})
	}
})
```


apidoc/Titanium/UI/WebView.yml
```diff
@@ -1121,6 +1152,41 @@ properties:
...
+    since: "12.1.0"
+
+  - name: DIN_A2
+    summary: PDF paper size
+    type: Number
+    platforms: [android]
+    permission: read-only
+    since: "12.1.0"
+
+  - name: DIN_A1
```
hansemannn, line 1183: Can you change to `PDF_PAGE_DIN_A2`, etc.? It makes it more obvious where it's used.

m1ga: done :+1:

hansemannn (approved):

PR #13666, opened 2022-12-07
feat(all): update default tidev modules

m1ga:
Integrate new modules to the SDK:

iOS
* ti.identity 5.0.0
* ti.applesignin 3.1.2
* facebook 13.0.0
* ti.map 7.0.0 

Android:
* facebook 12.0.0
* ti.map 5.5.0
* ti.playservices 18.1.0


m1ga:
@hansemannn I've set ti.playservice to [pre-release](https://github.com/tidev/ti.playservices/releases/tag/v18.1.0) so I could use the download link and generate the json file. If everything is fine we should switch it to a normal release

hansemannn:
We can release it!

PR #13669, opened 2022-12-09
fix(android): code optimizations

m1ga:
follwing up https://github.com/tidev/titanium_mobile/pull/13605

here are some more improvements that are recommended by Android Studio. This time:
* possible memory leak in AsyncTask
* possible memoery leak in "non static inner class" 
> Reports inner classes that can be made static.
> A static inner class does not keep an implicit reference to its enclosing instance. This prevents a common cause of memory leaks and uses less memory per instance of the class.

Tested in my app (UI still looks the same and no errors), test suite (same result as in #13605) and hyperloop example app.


hansemannn:
Excellent change, but I don't feel comfortable reviewing / deciding these major Android changes alone. Do we have someone else here?

PR #13670, opened 2022-12-11
fix(docs): fixing filenames for titanium-docs

m1ga:
The titanium-docs [will fail to build](https://github.com/tidev/titanium-docs/actions/runs/3642007615/jobs/6157561804) at the moment. **Two issues**:

* can't find `progressindicator_android.png` because it's camel-cased.
* snackbar lands in `docs/api/structs/titanium/ui/android/` instead of `docs/api/titanium/ui/android/`

**Solution:**
* I've renamed the file so it will find it
* not sure why the snackbar docs are in the structs folder. I've added the `extends: Titanium.UI.View` part so hopefully that will make it to stay in it's normal folder


cb1kenobi (approved):
LGTM

PR #13678, opened 2022-12-15
fix(android): fix duplicate lifecycle issue

### LINKED ISSUE #13676: Android: Duplicate lifecycle class error (Ti 12.0.0.RC)

### Description

Got this error (had this before in a different app already):
```
Duplicate class androidx.lifecycle.ViewModelLazy found in modules jetified-lifecycle-viewmodel-ktx-2.3.1-runtime (androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.1) and lifecycle-viewmodel-2.4.0-runtime (androidx.lifecycle:lifecycle-viewmodel:2.4.0)
```

It's because of the appcompat library.
https://stackoverflow.com/a/74259107/5193915
> PROBLEM: I got this error when I updated the appcompat dependency to 1.5.1.
SOLUTION: I updated the appcompat dependency to 1.6.0-rc01.

We use 1.5.0 and 1.5.1 "should" also include the fix according to the release notes. But according to [this post](https://stackoverflow.com/questions/69817925/problem-duplicate-class-androidx-lifecycle-viewmodel-found-in-modules#comment130309522_73624805) it will only work with 1.6.x

**Workaround**
1. Adding this to the build.gradle:
```
configurations {
    all {
        exclude group: 'androidx.lifecycle', module: 'lifecycle-viewmodel-ktx'
    }
}
```
(should go into the FAQ page).

2. add `implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.4.0'`
3. use 1.6.0-rc1 (released in October, perhaps the full version will be out soon)
4. go back to 1.4.2 (would prefer the other options)

### Platform

Android

### SDK version you are using

12.0.0.RC

_No response_

m1ga:
Resolves https://github.com/tidev/titanium_mobile/issues/13676

More infos in the issue description.
Also added https://github.com/tidev/titanium-docs/pull/168 in case we keep the 1.5.0 version. Doesn't happen for everyone but the FAQ entry will also solve it for the app.

Edit: checked it on my Mac where the error appeared. Fixed with this PR

### DIFF
--- a/android/templates/build/ti.constants.gradle
+++ b/android/templates/build/ti.constants.gradle
@@ -6,7 +6,7 @@
  */
 
 project.ext {
-	tiAndroidXAppCompatLibVersion = '1.5.0'
+	tiAndroidXAppCompatLibVersion = '1.6.0-rc01'
 	tiAndroidXCoreLibVersion = '1.8.0'
 	tiAndroidXFragmentLibVersion = '1.5.2'
 	tiMaterialLibVersion = '1.6.1'
--- a/android/templates/build/ti.constants.gradle
+++ b/android/templates/build/ti.constants.gradle
@@ -7,6 +7,7 @@
 
 project.ext {
 	tiAndroidXAppCompatLibVersion = '1.6.0-rc01'
+	tiAndroidXCoreLibVersion = '1.9.0'
 	tiAndroidXFragmentLibVersion = '1.5.5'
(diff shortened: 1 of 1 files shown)


m1ga:
@hansemannn can you add a backport PR for 12.0.0 or should I create one against that branch by hand?

hansemannn:
See my Slack message :)

PR #13693, opened 2022-12-29
fix(ios): fix crash on soft-restart

hansemannn:
This crash happens when soft-restarting the app (e.g. with https://github.com/rborn/tiapprestart) after opening an Universal Link / User Activities API. The reason is that we override the internal `NSUserActivity` to a `NSDictionary` during initial launch any try to re-access the native value that is already mapped to a dictionary at this point.

The testing setup is a bit complex, but it's basically this:
1. Download the ti.apprestart module: https://github.com/rborn/tiapprestart
2. Setup Universal Links: https://stackoverflow.com/questions/36542001/appcelerator-how-to-add-universal-links-support-in-ios9-app
3. Open a universally-linked URL
4. Return to the app
5. Trigger an app restart

-> App crashes!


hansemannn:
@m1ga What do you think? I tested locally and it works properly, but I won't merge it before formerly accepted.

PR #13694, opened 2022-12-30
fix(android): gradle warning log

m1ga:
Will show the gradle warnings as warnings and not as errors:
210062035-8f0429e7-d4f1-401c-9c87-19fbcb46cc10

* using `Warning: ` as a string to check. Hopefully other gradle parts won't just have that string in the middle somewhere :smile: 
* default still `info`
* error still `error`


cb1kenobi (approved):
Perfect!

PR #13702, opened 2023-01-04
fix: remember proxies from section data items

### LINKED ISSUE #13433: iOS: ListView with many AttributedStrings crashes

### Description

Using a ListView with many attributedStrings will crash the iOS app.

### Expected Behavior

-

### Actual behavior

Hard crash

### Reproducible sample

```js

function addListView() {
		// footerView: mapView,
		templates: {
			test: {
				childTemplates: [{
					type: 'Ti.UI.View',
					childTemplates: [{
						type: 'Ti.UI.Label',
						bindId: 'label',
						properties: {
							color: 'black',
							bindId: 'label'
						}
					}],
					properties: {
						width: Ti.UI.FILL,
						height: 100,
						cardUseCompatPadding: true,
						backgroundColor: 'white'
					}
				}]
			}
		},
		defaultItemTemplate: 'test'
	});
	var items = [];

	for (let i = 0; i < 700; i++) {
		var txt = "test text test text test text test text test text test text test text test text test text test text";
			text: txt,
			attributes: [{
				type: Titanium.UI.ATTRIBUTE_PARAGRAPH_STYLE,
				range: [0, txt.length],
				value: {
					lineSpacing: 0,
					lineHeightMultiple: 0.8,
					lineBreakMode: Titanium.UI.ATTRIBUTE_LINE_BREAK_BY_WORD_WRAPPIN
				}
			}]
		});

		items.push({
			label: {
				attributedString: attr
			},
			template: 'test'
		})
	}

	section.setItems(items);
	win.add(listView);
}

	addListView();
})
win.open();
```

### Steps to reproduce

* start app
* scroll down

### Platform

iOS

### SDK version you are using

10.1.1

_No response_

janvennemann:
A potential fix for https://github.com/tidev/titanium_mobile/issues/13433 we are currently testing. This remember all proxies that are inside a list data item to ensure they are properly protected from garbage collection between setting the `items` on a list section and the list view rerendering.

Fixes https://github.com/tidev/titanium_mobile/issues/13433

### DIFF
--- a/iphone/Classes/TiUIListSectionProxy.m
+++ b/iphone/Classes/TiUIListSectionProxy.m
@@ -149,30 +149,30 @@
 {
   __weak TiUIListSectionProxy *weakSelf = self;
 
-  [self findProxyInItems:items
-       completionHandler:^(TiProxy *proxy) {
-         TiUIListSectionProxy *strongSelf = weakSelf;
-         if (strongSelf == nil) {
-           return;
-         }
+  [self enumerateProxiesInItems:items
+                     usingBlock:^(TiProxy *proxy) {
+                       TiUIListSectionProxy *strongSelf = weakSelf;
+                       if (strongSelf == nil) {
+                         return;
+                       }
 
-         [strongSelf rememberProxy:proxy];
-       }];
+                       [strongSelf rememberProxy:proxy];
+                     }];
 }
 
 - (void)forgetProxiesInItems:(NSArray<NSDictionary *> *)items
 {
   __weak TiUIListSectionProxy *weakSelf = self;
 
-  [self findProxyInItems:items
-       completionHandler:^(TiProxy *proxy) {
-         TiUIListSectionProxy *strongSelf = weakSelf;
-         if (strongSelf == nil) {
-           return;
-         }
+  [self enumerateProxiesInItems:items
+                     usingBlock:^(TiProxy *proxy) {
+                       TiUIListSectionProxy *strongSelf = weakSelf;
+                       if (strongSelf == nil) {
+                         return;
+                       }
 
-         [strongSelf forgetProxy:proxy];
-       }];
+                       [strongSelf forgetProxy:proxy];
+                     }];
 }
 
 - (void)appendItems:(id)args
@@ -454,7 +454,7 @@
 
 #pragma mark - Utilities
 
-- (void)findProxyInItems:(NSArray<NSDictionary *> *)items completionHandler:(void (^)(TiProxy *proxy))completionHandler
+- (void)enumerateProxiesInItems:(NSArray<NSDictionary *> *)items usingBlock:(void (^)(TiProxy *proxy))block
 {
   for (NSDictionary *item in items) {
     for (NSString *key in item) {
@@ -466,7 +466,7 @@
         for (NSString *bindPropertyName in itemPropertyValue) {
           id propertyValue = itemPropertyValue[bindPropertyName];
           if ([propertyValue isKindOfClass:TiProxy.class]) {
-            completionHandler(propertyValue);
+            block(propertyValue);
           }
         }
       }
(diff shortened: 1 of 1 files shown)


m1ga:
nice, will check it later :+1: Had a quick test run with the old code but couldn't reproduce the crash with the latest xcode. I'll try to use the old app where I've found the crash for the first time, there I could reproduce it better

iphone/Classes/TiUIListSectionProxy.m
```diff
@@ -137,11 +139,52 @@ - (void)setItems:(id)args withObject:(id)properties
...
+    for (NSString *key in item) {
+      if ([key isEqualToString:@"template"] || [key isEqualToString:@"properties"]) {
+        continue;
+      }
+      id itemPropertyValue = item[key];
+      if ([itemPropertyValue isKindOfClass:NSDictionary.class]) {
+        for (NSString *bindPropertyName in itemPropertyValue) {
+          id propertyValue = itemPropertyValue[bindPropertyName];
+          if ([propertyValue isKindOfClass:TiProxy.class]) {
+            [self forgetProxy:propertyValue];
```
hansemannn, line 180: Once validated: You could reuse the loop for both remember/forget to prevent duplicate code

m1ga:
I can reproduce it with the old app (listview is a bit more complex images + attributedString) and once I have some items loaded and scroll up/down it will suddenly crash again. 
Using this PR I can scroll around for some minutes now without any crash :+1: I'll give it some more runs tomorrow but it's looking good. 
Any performances downsides using this approach? I see some nested for loops in there :smile: Not sure how often they'll run or if it depends on the amount of items

hansemannn:
Regarding performance: As it's only done when setting items, it will not impact scroll performance. It *may* impact the speed of setting items, but I'll do some benchmarking there.

janvennemann:
@m1ga As Hans already mentioned the for loops only run when setting new items. I did some time measurements, and the time spend in the for loops can be neglected compared to the total time it takes to set the items. Setting 5000 items on a section takes ~800ms on my iPhone X and only 4ms of that are spend in the for loops. Of course this might increase a little the more attributed string / proxies you use, but even then it shouldn't be noticeable.

BTW, here is the code that we were using to reproduce the crashes. It's a slightly modified version of your initial example, resetting the items every 500ms.

```js

function addListView() {
		templates: {
			test: {
				childTemplates: [{
					type: 'Ti.UI.View',
					childTemplates: [{
						type: 'Ti.UI.Label',
						bindId: 'label',
						properties: {
							color: 'black',
							bindId: 'label'
						}
					}],
					properties: {
						width: Ti.UI.FILL,
						height: 100,
						cardUseCompatPadding: true,
						backgroundColor: 'white'
					}
				}]
			}
		},
		defaultItemTemplate: 'test'
	});

	setInterval(() => {
		console.warn('REDRAW');

		section.items = getItems();
	}, 500);

	win.add(listView);
}

	addListView();
});
win.open();

function getItems() {
	const items = [];

	for (let i = 0; i < 700; i++) {
		var txt = 'test text test text test text test text test text test text test text test text test text test text' + new Date();
			text: txt,
			attributes: [{
				type: Titanium.UI.ATTRIBUTE_PARAGRAPH_STYLE,
				range: [0, txt.length],
				value: {
					lineSpacing: 0,
					lineHeightMultiple: 0.8,
					lineBreakMode: Titanium.UI.ATTRIBUTE_LINE_BREAK_BY_WORD_WRAPPIN
				}
			}, {
				type: Titanium.UI.ATTRIBUTE_FOREGROUND_COLOR,
				range: [0, txt.length],
				value: 'red'
			}]
		});

		items.push({
			label: {
				attributedString: attr
			},
			template: 'test'
		});
	}

	return items;
}
```

hansemannn (approved):
Working fine for my test cases

m1ga (approved):
LGTM

tested the latest version with my app and the new test code. Both crashed without the patch and work fine with the patch :+1:

PR #13715, opened 2023-01-12
feat(android): add rotation to blob

m1ga:
Adding a `rotation` value to Ti.Blob. If it is available in the blob it will return it:

```js
const btn1 = Ti.UI.createButton({title:"1"})
const btn2 = Ti.UI.createButton({title:"2"})
const btn3 = Ti.UI.createButton({title:"3"})
win.add([btn1,btn2,btn3]);

	Ti.Media.openPhotoGallery({
		success: function(e) {
			console.log(e.media.rotation);
			console.log(e.media.width, "x", e.media.height);
			console.log(e.media.uprightWidth, "x", e.media.uprightHeight);
		}
	});
})
	Ti.Media.showCamera({
		success: function(e) {
			console.log(e.media.rotation);
			console.log(e.media.width, "x", e.media.height);
			console.log(e.media.uprightWidth, "x", e.media.uprightHeight);
		}
	});
})
btn3.addEventListener("click", function() {
  var file  = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'test.txt');
  console.log(file.read().rotation);
})
win.open();
```

## Test:
* run the app (manually allow Camera permission!)
* click 1, select an image. Should most of the time return 0
* click 2, make an image in portrait or landscape mode and it should return a rotation value
* click 3, it should return `undefined`


android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -411,6 +415,12 @@ public int getUprightWidth()
 		return this.uprightWidth;
 	}
 
+	@Kroll.getProperty
+	public int getRotation()
+	{
+		return this.rotation;
```
cb1kenobi, line 421: If `rotation` is not avaiable, return `undefined` instead of `-1`.

cb1kenobi (changes requested):

cb1kenobi (approved):
Looks great! **APPROVED**

PR #13723, opened 2023-01-18
fix(android): imageAsResized to squared images

m1ga:
Resizing an image to a squared image doesn't work:

```
	image: 'appicon.png'
});
const blob = img.toBlob();
const b = blob.imageAsResized(60, 60);
console.log(b.width, b.height);
```

Ti 12.0.0.GA: will return 128x128
This PR: will return 60x60

`blob.imageAsResized(60, 61);` works fine in 12.0.0.GA

Added tests to cover resizing a squared image and a non-squared image to different sizes.


m1ga:
About the fix:

```js
const blob2 = Ti.Filesystem.getFile('appicon.png').read();
const b2 = blob2.imageAsResized(50, 50);
console.log(b2.width, b2.height);
```
works fine so I've kept all the "square check" and "sample" stuff as it is better to resize it this.
But using a blob from an ImageView will fail. So I do a "fallback" check at the end to see if the image was resized, if not then use the normal `createScaledBitmap` way to rescale it. 
Then all the current apps where it works will still use the old code and only parts where it failed will rescale it at the end again

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
```diff
@@ -641,7 +641,7 @@ public TiBlob imageAsCropped(Object params)
 	@Kroll.method
 	public TiBlob imageAsResized(Number width, Number height)
 	{
-		boolean valid = (image != null) || (image == null && (this.width > 0 && this.height > 0));
+		boolean valid = (image != null) || (this.width > 0 && this.height > 0);
```
hansemannn, line 644: Can you please go back with git-blame and see why the check was added in the first place? It looks like it was there for another reason

hansemannn (commented):
One thing to validate

android/titanium/src/java/org/appcelerator/titanium/TiBlob.java
m1ga: it was added here: https://github.com/tidev/titanium_mobile/pull/5597/files#diff-a8746380cd1e01787f46def043161a8d66bb98fb0f1e5b8184bd981355f5d7c7R563

in a [TIMOB-1510 Memory Optimizations on TiBlob ](https://jira-archive.titaniumsdk.com/TIMOB-1510)
https://github.com/tidev/titanium_mobile/pull/5597/commits/efe48cbe9b10623561cb9b6c73d48e4e14e1632d

m1ga: btw: looks like I only used that for a different resize way before. It works fine keeping it in there

PR #13726, opened 2023-01-21
feat(android): parity for keyboardframechanged and keyboardVisible

m1ga:
Add Android parity for `keyboardframechanged` event and `Ti.App.keyboardVisible` to Ti.App:

```js
var tf = Ti.UI.createTextField({width: 100});
var tf2 = Ti.UI.createTextField({width: 100, top: 0});
win.add([tf,tf2]);

Ti.App.addEventListener("keyboardframechanged", function(e){
	alert("keyboard visible: " + Ti.App.keyboardVisible);
	if (Ti.App.keyboardVisible) {
		alert("Size: " + e.keyboardFrame.width + " x " + e.keyboardFrame.height);
	}
});
win.addEventListener("open", function(e){
	alert("keyboard visible: " + Ti.App.keyboardVisible);
})
win.open();
```


m1ga:
https://user-images.githubusercontent.com/4334997/213867390-a0f02a21-f573-40d2-8d6d-7fd45ec742d6.mp4

hansemannn:
Could this be changed to finally bring parity for https://titaniumsdk.com/api/titanium/app.html#properties_keyboardvisible and https://titaniumsdk.com/api/titanium/app.html#events_keyboardframechanged ?

m1ga:
I can trigger that event and use the ` keyboardVisible` property :+1: 
Not all fields will be filled in the event but at least it will be there for Android and you can check the keyboard status!

hansemannn:
Yep, great parity already!

m1ga:
It's not in the `Ti.App.` namespace 
* `Ti.App.keyboardVisible`
* and `Ti.App.addEventListener("keyboardframechanged")`

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -763,17 +763,18 @@ public void onChanged(TiActivitySafeAreaMonitor monitor)
...
-					kd.put("keyboardVisible", keyboardVisible);
+				if (isVisible != keyboardVisible) {
 					TiWindowProxy windowProxy = TiBaseActivity.this.window;
 					if (windowProxy != null) {
 						windowProxy.fireSafeAreaChangedEvent();
 					}
-					windowProxy.fireEvent("keyboardChanged", kd);
-					oldKeyboardValue = keyboardVisible;
+					KrollDict kd = new KrollDict();
+					kd.put("keyboardFrame", null);
```
hansemannn, line 774: Could we eventually use this snippet to get some parts of the frame? https://stackoverflow.com/a/72566595/5537752

m1ga: added those values but I'll need to verify it with some different keyboards. height/width is correct on my Samsung A20.

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -760,6 +762,38 @@ public void onChanged(TiActivitySafeAreaMonitor monitor)
...
+						kdX.put("top", keyboardSize.top);
+						kdX.put("bottom", keyboardSize.bottom);
+						kdFrame.put("x", kdX);
+						kdFrame.put("y", kdY);
+						kdFrame.put("height", keyboardSize.bottom);
+						kdFrame.put("width", width - keyboardSize.left - keyboardSize.right);
+						kdAll.put("keyboardFrame", kdFrame);
+						kdAll.put("animationDuration", 0);
+						
+						tiApp.fireAppEvent("keyboardframechanged", kdAll);
```
hansemannn, line 792: Last one: We should guard if there is actually an event listener, so the whole logic is not called for cases where the client does not even listen to the event.

m1ga: sure, I'll add a check.
Always wondered: does it make a difference if there is no listener and you call it?

m1ga: The fix will apply for all `Ti.App` event listeners

m1ga: looks like it didn't like that last change. Some tests failed so I'm testing it without the hasListener

hansemannn: Can't you just add the check here?

m1ga: I'm currently re-arranging to logic since it will crash apps at the moment. The check will be in place after that

m1ga: edit: I have to go into TiApplication and do the check there since it's on `Ti.App` and not the `Window`. Guess I have to add some more checks so it won't crash :disappointed:

m1ga: :facepalm: 
https://github.com/tidev/titanium_mobile/blob/a16d76c767025a67a7375edd9da70a4d99edb597/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java#L781
it does the check already :rofl:

m1ga:
Not sure about some of the errors in the test suite. Don't think they are related to this PR.
e.g. I see
```
Expected to throw exception\\' undefined undefined' to be 'Unable to convert null'"
```
in error.test.js `Native exception surfaced`. It's using `Ti.Geolocation.accuracy = null;` to check for `'Unable to convert null'`.
It says its returning `undefined` but when I just run that line in a test app I see
Screenshot_20230124-193207

so the error is `Unable to convert null` :shrug: 

I've tested my app, your app, kitchensink and hyperloop examples and they look fine :+1:

hansemannn:
I meant the Java callback invocation and construct of the kroll dict, sorry for the confusion

m1ga:
exposed the `hasListener` and added it to the `if` checks:

https://github.com/tidev/titanium_mobile/pull/13726/commits/fec7cc73a90984c2c32dce535da8d4d674c8bfcd#diff-6f50b13a3f7b68de5485b38155383cdbd1ec15887ad7fd84573b826d79154790R772

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
```diff
@@ -760,6 +762,34 @@ public void onChanged(TiActivitySafeAreaMonitor monitor)
...
 				}
 			}
+
+			@Override
+			public void onKeyboardChanged(boolean isVisible, int width, int height, Insets keyboardSize)
+			{
+				if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q
+					&& isVisible != keyboardVisible
+					&& tiApp != null
+					&& tiApp.hasListener("keyboardframechanged")) {
```
hansemannn, line 772: Move the listener-check to the `onApplyWindowInsets` check, so this callback won't even be called if not explicitly listened to.

android/titanium/src/java/org/appcelerator/titanium/view/TiActivitySafeAreaMonitor.java
```diff
@@ -105,6 +108,14 @@ public void onLayoutChange(View view, int left, int top, int right, int bottom,
 			@Override
 			public WindowInsets onApplyWindowInsets(View view, WindowInsets insets)
 			{
+				if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
```
hansemannn, line 111: Move the listener-check to here

android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
m1ga: I keep the check still in here. It should only be called when `tiApp != null && tiApp.hasListener("keyboardframechanged")` anyways but for safety I'll check it still here

android/titanium/src/java/org/appcelerator/titanium/view/TiActivitySafeAreaMonitor.java
m1ga: moved tiapp into the safeareamonitor file and doing the check there now

hansemannn:
Thanks for all the adjustments! Sorry for being so picky..but I think it's worth to double-check the calls before it's getting an issue in production. We should be ready to go here..if someone else can verify the fix, we can merge.

hansemannn (approved):
Code-review approved, waiting for functional testing.

hansemannn:
@m1ga Any external testing done? Maybe the users that requested this can help. This can always be the way to fast-lane changes.

m1ga:
Since this is a Android 11+ fix it wasn't useful for the user (Android 8.1 device) so I don't know the current status. I don't have any use for it so I didn't implement it in a client app at the moment

apidoc/Titanium/App/App.yml
```diff
@@ -394,8 +397,8 @@ events:
       - name: animationDuration
         summary: The duration of the keyboard animation. This parameter is only available on Titanium SDK 4.0.0 and later.
         type: Number
-    platforms: [iphone, ipad, macos]
-    since: {iphone: "3.0.0", ipad: "3.0.0", macos: "9.2.0"}
+    platforms: [android, iphone, ipad, macos]
+    since: {android: "12.1.0", iphone: "3.0.0", ipad: "3.0.0", macos: "9.2.0"}
```
m1ga, line 401:
```suggestion
    since: {android: "12.3.0", iphone: "3.0.0", ipad: "3.0.0", macos: "9.2.0"}
```

apidoc/Titanium/App/App.yml
```diff
@@ -609,9 +612,12 @@ properties:
 
   - name: keyboardVisible
     summary: Indicates whether or not the soft keyboard is visible.
+    description: |
+        On Android it will only work using Android 11 and up. Otherwise it will always be false.
     type: Boolean
     permission: read-only
-    platforms: [iphone, ipad, macos]
+    since: {android: "12.1.0", iphone: "7.5.0", ipad: "7.5.0", macos: "9.2.0"}
```
m1ga, line 619:
```suggestion
    since: {android: "12.3.0", iphone: "7.5.0", ipad: "7.5.0", macos: "9.2.0"}
```

m1ga:
Tested on Android 14 (Pixel 9) and Android 15 (Emulator) again, working fine. Event is fired and keyboardVisible value is correct

hansemannn (changes requested):

m1ga:
fixed all suggestions, thanks for check :+1:

hansemannn (approved):

PR #13727, opened 2023-01-22
feat(android): expose videoPlayer buffer

m1ga:
Expose the Android VideoPlayer `buffer` event

```js
var v = Ti.Media.createVideoPlayer({
	width: 300,
	height: 300,
	url: "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerEscapes.mp4"
});
var lbl = Ti.UI.createLabel({top: 0});

v.addEventListener("buffer", function(e){
	lbl.text = e.percentage
})

win.add([v,lbl]);
win.open();
```
label at the top should fill up to 100

### DIFF
--- a/apidoc/Titanium/Media/VideoPlayer.yml
+++ b/apidoc/Titanium/Media/VideoPlayer.yml
@@ -205,6 +205,16 @@ methods:
     summary: Stops playing the video.
 
 events:
+  - name: buffer
+    summary: |
+        Fired when the video is loading/buffering.
+    properties:
+      - name: percentage
+        summary: Percentage of the video that is buffered/loaded.
+        type: Number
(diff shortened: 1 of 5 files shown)


apidoc/Titanium/Media/VideoPlayer.yml
```diff
@@ -205,6 +205,16 @@ methods:
...
 
 events:
+  - name: buffer
+    summary: |
+        Fired when the video is loading/buffering.
+    properties:
+      - name: percentage
+        summary: Percentage of the video that is buffered/loaded.
+        type: Number
+    since: '12.1.0'
```
m1ga, line 215:
```suggestion
    since: '12.3.0'
```

PR #13729, opened 2023-01-25
fix(ios): fix border refresh issue

### LINKED ISSUE #13728: iOS: border gets bigger when setting borderRadius with multiple values

### Description

Setting the borderRadius as an array for elements like View or Textfield will get bigger when you pause and resume the app.

```js
	width: '94%',
	height: 40,
	borderColor: '#000',
	borderWidth: 1,
	borderRadius: [20, 0, 20, 0]
});
win.add(view);
win.open();
```

https://user-images.githubusercontent.com/4334997/214248377-26631b6a-ad09-4fd1-a838-01466e08d380.mp4

Won't happen with just `borderRadius: 20`

### Platform

iOS

### SDK version you are using

12.0.0.GA

hansemannn:
The line width of the border layer was always adjusted as well, causing rendering artifacts. This PR changes that behavior and also reduces memory load by guarding the setters during a trait collection change.

Fixes https://github.com/tidev/titanium_mobile/issues/13728

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
@@ -335,10 +335,17 @@ DEFINE_EXCEPTIONS
 {
   [super traitCollectionDidChange:previousTraitCollection];
 
+  BOOL isInBackground = UIApplication.sharedApplication.applicationState == UIApplicationStateBackground;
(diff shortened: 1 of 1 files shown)


m1ga:
Border stays the same after pause - resuming the app :+1: 
didn't test anything else :smile: 

not for this PR but I've noticed that `borderWidth: 1` looks thicker when using an array. When I don't set it OR use `borderRadius: 20` it will be a thin border.

PR #13730, opened 2023-01-27
fix(android): change deprecated view.setBackgroundDrawable

m1ga:
switching from `setBackgroundDrawable` to `setBackground` in TiUIView. Parameter is the same

https://developer.android.com/reference/android/view/View#setBackgroundDrawable(android.graphics.drawable.Drawable)

>This method was deprecated in API level 16.
use [setBackground(android.graphics.drawable.Drawable)](https://developer.android.com/reference/android/view/View#setBackground(android.graphics.drawable.Drawable)) instead

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
@@ -1173,7 +1173,7 @@ public abstract class TiUIView implements KrollProxyListener, OnFocusChangeListe
 						background.setBackgroundDrawable(currentDrawable);
 
 					} else {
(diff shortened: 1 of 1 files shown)


hansemannn (approved):

hansemannn:
Any known side effects? Otherwise, good to go.

m1ga:
no, test suite worked fine and my/your app too. I just noticed it while testing some backgroundImage stuff that it was marked as deprecated. 
Android 33 even just calls the deprecated function itself :smile: 
Screenshot_20230127_152813

but I like it better to have less deprecated calls in Ti SDK

PR #13732, opened 2023-01-28
feat(ios): read status bar height programmatically

hansemannn:
Requested via Slack to have a common solution that's more flexible for newer devices.

### Test Case

```js

    backgroundColor: '#fff'
});

const view = Ti.UI.createView({ top: 0, height: 0, backgroundColor: 'red' });

    title: 'Get status bar height'
});

    view.animate({ height: Ti.UI.statusBarHeight });
    console.warn('statusBarHeight = ', Ti.UI.statusBarHeight);
});

win.add([view, btn]);
win.open();
```

### Example

https://user-images.githubusercontent.com/10667698/215262914-e1abb396-cab8-4136-9428-c89056795cc0.mp4

### DIFF
--- a/apidoc/Titanium/UI/UI.yml
+++ b/apidoc/Titanium/UI/UI.yml
@@ -2814,7 +2814,7 @@ properties:
         when allowing rotation, we recommend using the <Titanium.UI.Window.safeAreaPadding> property of the related window.
     type: Number
     permission: read-only
-    platforms: [iphone, ipad, macos]
+    platforms: [iphone, ipad, macos, android]
     since: "12.1.0"
 
 examples:
--- a/apidoc/Titanium/UI/UI.yml
+++ b/apidoc/Titanium/UI/UI.yml
@@ -2807,6 +2807,16 @@ properties:
     osver: {ios: {min: "13.0"}}
     since: "9.1.0"
 
+  - name: statusBarHeight
+    summary: The total height of the status bar including safe area padding.
+    description: |
+        Use this property to determine an absolute spacing to the top in full screen windows. Note: For a more flexible approach, e.g.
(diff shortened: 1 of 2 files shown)


m1ga:
for Android too? `displayCutout`. I thought I did a PR with that already 🤔  Was doing some tests with it while I added the Rive animation to the statusbar. Will look if I still find it

hansemannn:
I'd add Android as well, but feel free to do so as well.

m1ga:
found my code already. 
```java
@Kroll.method
	public void getCutout()
	{
		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
			Window w = TiApplication.getAppCurrentActivity().getWindow();
			if (w == null) return;
			View dv = w.getDecorView();
			if (dv == null) return;
			DisplayCutout dpc = dv.getRootWindowInsets().getDisplayCutout();
			if (dpc == null) return;
			List<Rect> rects = dpc.getBoundingRects();

			int cutouts = rects.size();
			int[] result = new int[cutouts * 4];
			int index = 0;

			for (Rect rect : rects) {
				KrollDict kd = new KrollDict();
				kd.put("left", rect.left);
				kd.put("top", rect.top);
				kd.put("width", rect.width());
				kd.put("height", rect.height());
				fireEvent("cutout", kd);
			}
		}
	}
```

it returned the whole rectangle shape (e.g. a box around the camera cutout).  Will change it to get the height (rect.top + rec.height).

m1ga:
need to clean it up but its running :+1: 
Screenshot_20230128-130704

m1ga:
added it here: https://github.com/tidev/titanium_mobile/pull/13733
I've kept my `cuteoutSize` method so you can get the shape on Android too.

You could add `android` to the apidocs here

PR #13747, opened 2023-02-13
fix(ios): link to AppleWWDRCA

m1ga:
Just saw this post on SO https://stackoverflow.com/q/75433379/5193915
and according to https://www.apple.com/certificateauthority/ the old link (G1) expired 02/07/23

Is it correct to use the next link :shrug:


hansemannn (approved):
Good one!

PR #13749, opened 2023-02-16
feat(android): android optimizations

m1ga:
I've used ChatGPT and put in three methods and asked for optimizations :smile: 

Here are the changes:

<b>ListView:</b>
* Instead of using a LinkedList to store the recyclable items, you can use a LinkedHashSet to store the recyclable items in a more efficient way. A LinkedHashSet provides fast lookups and preserves the order of insertion.
This code uses a LinkedHashSet to store the recyclable items for each template, and it iterates over the recyclable items using a for-each loop to find a recyclable item with child views that can be reused by the current item. When a recyclable item is found, it removes it from the set and transfers its child views to the current item.

<b>removeAllChildren</b>
* One small optimization you could make is to avoid creating a new list if there are no children to remove. This would save a little bit of memory and processing time

<b>remove</b>
* The check for peekView() != null could be moved to the beginning of the function to avoid unnecessary computations if the parent view has not yet been attached to a window.
* The check for child.parent != null && child.parent.get() == this can be replaced with a simpler child.parent = null, since this statement removes the child view from the current parent, if any.

<b>toImage</b>
* The unnecessary blob variable was removed, since it's only used to hold a default result value

Android tests run fine, App runs fine. 

I like the `toImage` and `removeAllChildren` changes! Makes the code a bit easier to read. Don't see any performance improvements. We can ignore the ListView part since it's a bit out of my knowledge if its better or not.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewAdapter.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewAdapter.java
@@ -6,7 +6,7 @@
  */
 package ti.modules.titanium.ui.widget.listview;
 
-import java.util.LinkedList;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.TreeMap;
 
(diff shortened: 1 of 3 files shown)


hansemannn:
Honestly, I don't fully trust GPT-based results so far. Is someone having an issue with the current implementation? If not, I'd rather not use this, yet.

m1ga:
I trust the Android Studio recommendations more 😄 I only liked the part in https://github.com/tidev/titanium_mobile/pull/13749/files#diff-c1ca52b23726a03177273b78aed424203d8f5352bf7ba2022f74e2f8ad536ce3R962 so it won't create a tiblob and returns it at the end. 
But no complaints or improvements. Guess I'll close it and perhaps use the toImage part in a different PR

PR #13757, opened 2023-03-01
feat(android): use example platform folder for module builds

m1ga:
When you build a module with a custom build.gradle you couldn't use the `/example` folder as it didn't copy the build.gradle to the correct place.

This PR will copy a `/platform` folder into the tmp build folder at the correct place so you can use a custom `build.gradle` for your module. Like https://github.com/m1ga/ti.exoplayer/tree/main/example

## Test:

* clone https://github.com/m1ga/ti.exoplayer/
* go to `/android`
* run `ti build -p android` to build the module + the example app
* it should start the example app (before it complaint because it didn't use the build.gradle)

## Test 2:
* create a new module and build it (default app.js)

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -897,9 +897,9 @@ AndroidModuleBuilder.prototype.runModule = async function (cli) {
 	);
 
 	// Copy example/platform to tmp/platform to use a custom build.gradle
-	if (fs.existsSync(this.exampleDir + '/platform')) {
+	if (fs.existsSync(path.join(this.exampleDir, 'platform'))) {
 		appc.fs.copyDirSyncRecursive(
-			this.exampleDir + '/platform',
+			path.join(this.exampleDir, 'platform'),
 			path.join(tmpProjectDir, 'platform'),
 			{
 				preserve: true,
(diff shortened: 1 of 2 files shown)


android/cli/commands/_buildModule.js
```diff
@@ -896,6 +896,18 @@ AndroidModuleBuilder.prototype.runModule = async function (cli) {
 		}
 	);
 
+	// Copy example/platform to tmp/platform to use a custom build.gradle
+	if (fs.existsSync(this.exampleDir + '/platform')) {
```
hansemannn, line 900: Use `path.join` to concatenate paths

android/cli/commands/_buildModule.js
```diff
@@ -896,6 +896,18 @@ AndroidModuleBuilder.prototype.runModule = async function (cli) {
...
+		appc.fs.copyDirSyncRecursive(
+			this.exampleDir + '/platform',
```
hansemannn, line 902: Same here

hansemannn (changes requested):

m1ga:
Updated. Build of the module still works fine :+1:

hansemannn (approved):

PR #13758, opened 2023-03-02
chore: add “Optimize for Mac” option for macOS catalyst builds, add tooltip API

hansemannn:
Enables the "Optimize for Mac" flag like visually in Xcode:

Screenshot 2023-03-02 at 22 11 21

Also adds tooltip support for views and buttons
Screenshot 2023-03-02 at 22 39 55 Screenshot 2023-03-02 at 22 40 52

Example (views):
```js
    backgroundColor: '#fff'
});

const btn = Ti.UI.createView({
    backgroundColor: 'red',
    width: 200,
    height: 100,
    tooltip: 'Click me!'
});

win.add(btn);
win.open();
```

Example (buttons):
```js
    backgroundColor: '#fff'
});

    title: 'Trigger',
    tooltip: 'Click me!'
});

win.add(btn);
win.open();
```

### DIFF
--- a/apidoc/Titanium/UI/Button.yml
+++ b/apidoc/Titanium/UI/Button.yml
@@ -356,6 +356,16 @@ properties:
     description: Only one of `title` or `titleid` should be specified.
     type: String
     since: "1.5"
+  
+  - name: tooltip
+    summary: The default text to display in the control's tooltip.
+    description: |
+         Assigning a value to this property causes the tool tip to be displayed for the view.
+         Setting the property to `null` cancels the display of the tool tip for the view.
+         Note: This property is only used for apps targeting macOS Catalyst.
+    type: String
+    osver: { ios: { min: "15.0" } }
+    since: "12.1.0"
 
   - name: verticalAlign
     summary: |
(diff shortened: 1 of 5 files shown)


hansemannn:
Tested by Cesar, so I assume we can merge this!

macCesar:
In line 3214 of this file:` iphone/cli/commands/_build.js`, you'll also need to check when compiling for the Mac App Store to set the "Optimize for Mac" flag:

```javascript
if (this.target === 'macos' || this.target === 'dist-macappstore') {
  deviceFamily = [...deviceFamily.split(','), '6'].join(',');
}
```

Otherwise, the UI will be at 77% of its original size.

hansemannn:
Interesting. But why would this option only be enabled when deploying for production? I would assume it's a general setting, like recommended by Apple.

macCesar:
But the change I'm suggesting affects both the development and production environments.

Let me explain...

When I tested the production version of my app, I noticed that the entire interface was scaled down, just like before you changed the flag to 'Optimize for Mac'.

After making that change, both versions displayed the interface correctly.

PR #13774, opened 2023-03-16
feat(android): secure Ti.Network.Socket

m1ga:
Adding a `secure` property to `createTCP` in order to create a secure socket [SSLSocket](https://developer.android.com/reference/javax/net/ssl/SSLSocketFactory).

```js
var socket = Ti.Network.Socket.createTCP({
	host: 'www.github.com',
	port: 443,
	secure: true,
	connected: function(e) {
		Ti.API.info('Socket opened!');
		Ti.Stream.pump(e.socket, readCallback, 1024, true);
		Ti.Stream.write(socket, Ti.createBuffer({
			value: "GET / HTTP 1.1\n\rHost: www.github.com\r\nConnection: close\n\r\n\r"
		}), writeCallback)
	},
	error: function(e) {
		Ti.API.info('Error (' + e.errorCode + '): ' + e.error);
	},
});

function writeCallback(e) {
	Ti.API.info('Successfully wrote to socket.');
}

function readCallback(e) {
	try {
		if (e.buffer) {
			var received = e.buffer.toString();
			Ti.API.info('Received: ' + received.length);
		} else {
			Ti.API.error('Error: read callback called with no buffer!');
		}
	} catch (ex) {
		Ti.API.error(ex);
	}
}
socket.connect();
```

Tested by a slack user with a secure connection.

**Test:**
* run the code with Ti 12.1.1 (or `secure:false`) and you should see `Received: 0`
* run the code with this PR and `secure:true` and you should see `Received: 207`

CLI check `printf 'GET / HTTP/1.1\r\nHost: github.com\r\n\r\n' | nc --ssl github.com 443`

Info about the iOS part:
https://ti-qa-archive.github.io/question/63501/does-appcelerator-support-tlsssl-sockets.html#answer-195541
would be nice if someone adds parity


m1ga:
@cb1kenobi Got feedback by the requester and added a test case. We were overthinking it a bit :smile:

cb1kenobi (approved):

PR #13780, opened 2023-03-24
chore: fix packages

### LINKED ISSUE #13778: Android: Karma tests failling - Callsite/Stack issue

### Description

Putting the Slack infos in here:

While testing the Karma part for ti.playservices I've found that node-titanium-sdk is behaving differently in the latest version
When you run (error test from the test suite)
```
try {
	Ti.Geolocation.accuracy = null
} catch (ex) {
	console.log(ex.stack)
}
```
it results in `[{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}]`

before it was
```
"Error: Unable to convert null
\n    at new Controller (/alloy/controllers/index.js:46:24)
\n    at Object.exports.createController (/alloy.js:427:8)
\n    at /app.js:29:7
\n    at Module._runScript (ti:/kroll.js:1346:15)
\n    at Module.load (ti:/kroll.js:829:13)
\n    at Module.loadJavascriptText (ti:/kroll.js:1192:15)
\n    at Module.loadAsFile (ti:/kroll.js:1244:22)
\n    at Module.loadAsFileOrDirectory (ti:/kroll.js:1163:26)
\n    at Module.require (ti:/kroll.js:974:30)
\n    at Module.global.Module.require (<embedded>:5053:32)"
```

I can pin it down to happen between 12.1.0.v20230116130958 and 12.1.0.v20230120162600. That was the [node-titanium-sdk merge](https://github.com/tidev/titanium_mobile/pull/13724).
BUT: when I build a master with the old 5.1.6 I still see the problem :thinking_face:
That confuses me a bit, might be some build issue on my side.
Since the only "big" change in the version has been the node module package raise I suspect that one of the packages is the issue:
[https://github.com/tidev/node-titanium-sdk/commit/70ae6bfc2c75896ccd19c04ebf90098ab667ab7e#diff-7ae45ad102eab3b6d7e789[…]b346470d7bc6507b6481575d519](https://github.com/tidev/node-titanium-sdk/commit/70ae6bfc2c75896ccd19c04ebf90098ab667ab7e#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)
Don't think the ndk error messages or eslint fixes are responsible.

I just can't test it.  So it would be great if someone can test the following:
* revert the package.json changes in node-titanium-sdk
* npm i and use that node-titanium-sdk in 12.1.0 master
* run the test code above (Android) and see if the output is the correct stack again

I can't even build a 12.0.0 (released version works fine) here without the issue :disappointed:

before we make a 12.0.1 we should think about this one. I'm sure that a 12.0.1 build will have the same issue now as it might be a npm package that will install a newer version now.
Run the 4 lines of code with a 12.0.0 build on Android to check if the error output is correct.
It's not a big issue for me as I don't use  ex.stack somewhere and the ex.message is still correct. I only found this running the test suite and karma tests in modules (jasmine looks for an error as a string and its an object).

https://www.diffchecker.com/C7VPEuqB/
diff of the package-lock files.
left: master - right: last working commit fc1e2e108cc66ef38268492e1c065aea80483f2d
ioslib and node-titanium-sdk are the obvious changes but the child packages change a lot too. Babel/core 7.7.5 -> 7.20.12 and so on

after days of debugging I made a tiny step forward! At least I found the place where the error message is created: [https://github.com/tidev/titanium_mobile/blob/ee64cd2d50add46b16c5de9e20607719cd35[…]/Resources/ti.internal/extensions/node/internal/util/inspect.js](https://github.com/tidev/titanium_mobile/blob/ee64cd2d50add46b16c5de9e20607719cd35fa39/common/Resources/ti.internal/extensions/node/internal/util/inspect.js#L754)
and the CallSite stack is mentioned here: https://v8.dev/docs/stack-trace-api#customizing-stack-traces
just have to figure out now if that is something to fix in inspect.js :slightly_smiling_face:

I'm sure the actual "error"/change is somewhere else but since I can't find it (tried many babel versions, babel modules, different node-titanium....) I thought it might be worth seeing if we can use the "new" version and try to make it spit out the error again and not just empty objects :smile:

https://coolaj86.com/articles/callsite-get-your-stack-trace-back/

### Platform

Android

### SDK version you are using

12.1.0

m1ga:
after <i>weeks</i> of testing I found that babel was the issue and some other packages that are newer now when you remove the package-lock file.

Resolves https://github.com/tidev/titanium_mobile/issues/13778

### Attention
<strike><b>Merge this first:</b> https://github.com/tidev/node-titanium-sdk/pull/649 - after that I'll update the version number in this PR for the node-titanium-sdk package</strike> - merged

### Log

Full log in https://tidev.slack.com/archives/C038J3D6M4Y/p1679688501001629?thread_ts=1675721718.383329&cid=C038J3D6M4Y

### Info

Removing e.g. `core-js-compat` or updating to babel `7.12.*` will result in the `callsite {}` issues. Had to add babel/core and types in here so it will force this version. Starting to put the package version info into https://github.com/tidev/titanium_mobile/issues/13781

### Fixed version

<strike>I went ahead and used fixed version in the package.json file! Test suite runs fine besides the **"TableView" issue** ([check slack](https://tidev.slack.com/archives/C038J3D6M4Y/p1679703590799479)).</strike>
* reverted the tableView `Cannot call this method in a scroll callback` fix from https://github.com/tidev/titanium_mobile/pull/13768

Tested two apps (one external), test suite, hyperloop examples, module build

### Follow up

* Someone needs to find out what the issue is for a higher version and fix the callsite part! Babel changelog is here https://github.com/babel/babel/blob/main/CHANGELOG.md#v7120-2020-10-14 
* <strike>we should use fixed versions for the node packages to make it future prove and use only packages we want it to use</strike>

This way we can finally create a 12.1.0 in my opinion. After that we can continue to update node packages, the current versions are the same as 12.0.0 (especially babel).


m1ga:
* reverted the tableView `Cannot call this method in a scroll callback` fix from https://github.com/tidev/titanium_mobile/pull/13768
* using the old tableview test again

cb1kenobi (approved):
Tested on Windows, both building the SDK and the CI build. Works!

PR #13786, opened 2023-04-02
fix(ios): fix production build error in Xcode 14.3

hansemannn:
This was hard to figure out..but it does work with all my sample projects + complex production app. The question is still, why Apple broke this in Xcode 14.3 and also if there might be a better place to create the required folder. Happy to feedback!

[Download SDK that includes this fix](https://github.com/tidev/titanium_mobile/suites/11983689555/artifacts/629010322)

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -3300,7 +3300,7 @@ iOSBuilder.prototype.createXcodeProject = function createXcodeProject(next) {
 			outputPaths: [],
 			runOnlyForDeploymentPostprocessing: 0,
 			shellPath: '/bin/sh',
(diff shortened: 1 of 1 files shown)


hansemannn:
It has now been backported to 12.1.0 (12_1_X) as well

m1ga:
Did this somehow disable the Android builds? The actions get cancelled without any error:

Screenshot_20230404_164925

hansemannn:
Not by me, but I can retrigger it

m1ga:
AH! Nevermind. 
> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04 or ubuntu-22.04 (ubuntu-latest). For more details, see https://github.com/actions/runner-images/issues/6002

and guess when it stopped:
>  Deprecation will begin on 2022/08/08 and the image will be fully unsupported by 2023/04/03

:smile: 

will update it

PR #13788, opened 2023-04-04
chore: fix Android clipboard test

m1ga:
<strike>The clipboard test won't run in every Android version without an issue so I make this iOS only for now.</strike>

Increased the waiting time in the `clipboard` test. It turns out that the next UI test (ImageView) will remove the focus too quickly and on of the clipboard tests will fail. Increasing the waitingTime will fix it.

That way it's <strike>`5166 Total Tests: 5166 passed, 0 failed, 0 skipped.`</strike> `5167 Total Tests: 5167 passed, 0 failed, 0 skipped.
` again. Should we add a `test action` (not for every commit of course) that will run the test suite again?

### DIFF
--- a/tests/Resources/ti.ui.clipboard.test.js
+++ b/tests/Resources/ti.ui.clipboard.test.js
@@ -63,7 +63,7 @@ describe('Titanium.UI.Clipboard', () => {
 				should(Ti.UI.Clipboard.clearData).be.a.Function();
 			});
 
-			it('clears \'text\' data type', () => {
+			it.ios('clears \'text\' data type', () => {
 				setTimeout(function () {
 					Ti.UI.Clipboard.clearData(); // delete all data
 					Ti.UI.Clipboard.text = 'clearData';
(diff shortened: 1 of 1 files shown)


hansemannn:
Do we know *why* they would fail on some Android versions, e.g. only > 11 or only > 13? In that case, we should look into the code

m1ga:
I don't know but it was broken for several months (years) like many other tests so I don't think the tests where used that often :wink: I've tested them but did now that this (and the geo parts) could fail so I ignored them. I rather have it clean so we could add it to the CI again and then add more tests.

It is also only the `should(Ti.UI.Clipboard.hasText()).be.true();` part in the test. All other clipboard tests work fine.

m1ga:
Android 10 worked: https://github.com/tidev/titanium_mobile/pull/13355

m1ga:
It has to do with the current focus.
If I do
```js
Ti.UI.Clipboard.setData('text', "asdf");
console.log(Ti.UI.Clipboard.getData('text'));
```
it will return `null`. But if I add a TextField and use it in the TextField `focus` event with
```js
tf.addEventListener("focus", function(){
	tf.value = Ti.UI.Clipboard.getData('text');
})
```
it does return the correct value and writes it into the TextField. I adjust the test to include a textfield

hansemannn (approved):

PR #13790, opened 2023-04-04
chore: update ubuntu in github action

m1ga:
> We have started the deprecation process for Ubuntu 18.04. While the image is being deprecated, You may experience longer queue times during peak usage hours. Deprecation will begin on 2022/08/08 and the image will be fully unsupported by 2023/04/03

Doing the smaller version update for now to `ubuntu-20.04`  (`ubuntu-22.04` should be the next, but let's see if the 20.04 works fine)

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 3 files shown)


m1ga:
Luckily everything green :smile: :+1: 
Android artifact works fine

cb1kenobi (approved):

PR #13793, opened 2023-04-07
chore: github actions

m1ga:
Update github actions to newer version. 

Some of the current actions use node 12 internally and produce a warning

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -20,13 +20,13 @@ runs:
       shell: bash
 
     - name: Setup Node.js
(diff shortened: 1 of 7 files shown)


m1ga:
Check the warnings before and after:
warnings before: https://github.com/tidev/titanium_mobile/actions/runs/4634996927
warnings after: https://github.com/tidev/titanium_mobile/actions/runs/4637214759

```
JavaScript
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16...
```
those are gone now :+1:

cb1kenobi (approved):

PR #13800, opened 2023-04-11
feat: add sponsors action

hansemannn:
cc @joshualambert @cb1kenobi @ewanharris @m1ga

### DIFF
--- a/.github/workflows/sponsors.yaml
+++ b/.github/workflows/sponsors.yaml
@@ -1,0 +1,15 @@
+name: Sponsors
+on:
+  schedule:
+    - cron: '0 0 * * *' # Every 24 hours
+
+jobs:
+  update-sponsoprs:
+    runs-on: ubuntu-latest
+    name: Update Sponsors
+    steps:
+    - name: Update Sponsors in README.md
+      uses: JamesIves/github-sponsors-readme-action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        file: 'README.md'
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ see the [LICENSE](LICENSE) file for specific details.
 # Table of Contents
 
 - [Table of Contents](#table-of-contents)
+  - [Sponsors](#sponsors)
   - [Features](#features)
   - [Hyperloop](#hyperloop)
     - [Features](#features-1)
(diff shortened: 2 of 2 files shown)


hansemannn:
It seems like the `secrets.GITHUB_TOKEN` is not available for actions, only for workflows. Can someone add that?

.github/actions/package/action.yml
```diff
@@ -103,3 +103,10 @@ runs:
         name: |
           android-build
           ios-build
+
+    ### ONLY FOR TESTING INSIDE THE PR -- See .github/workflows/release.yml for the final destination ##
+    - name: Generate Sponsors
+      uses: JamesIves/github-sponsors-readme-action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
ewanharris, line 111: As this is a composite action not a workflow definition there is no access to secrets. It would need to be declared as an in the `inputs` block on line 4, and then referenced as `inputs.GITHUB_TOKEN`.

But given that you're just trying to test it out maybe just add it to `workflows/build.yml` after the package step that references this file rather than do all that to have it work in this file?

hansemannn: Thank you! Updated!

.github/actions/package/action.yml
```diff
@@ -102,4 +102,4 @@ runs:
       with:
         name: |
           android-build
-          ios-build
+          ios-build
```
cb1kenobi, line 105: What's up with the removal of the new line?

.github/workflows/build.yml
```diff
@@ -86,3 +86,8 @@ jobs:
         node-version: '16.x'
         java-version: '11'
         vtag: ${{ env.vtag }}
+    - name: Generate Sponsors
```
cb1kenobi, line 89: After giving this some thought, the list of sponsors has nothing to do with the build. I suggest creating a new workflow that uses cron to automatically regen the list of sponsors every 12 or 24 hours using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule.

hansemannn: Will do! Just want to see if the general gen works, then we can check the best place.

hansemannn: @cb1kenobi Updated!

cb1kenobi (approved):
Nice, this is dope!

hansemannn:
Let's see if it actually work. I'll merge this once the workflows all completed.

hansemannn:
Mhh, it doesn't look like. I'll need to dig further here.

cb1kenobi:
It's a cron job, it may not run for a bit. I suggest you update the `sponsors.yaml` by adding a `workflow_dispatch:` as such:

```
name: Sponsors
on:
  workflow_dispatch:
  schedule:
    - cron: '0 0 * * *' # Every 24 hours
```

Then we can force the action to run with this handy button:
image

m1ga:
the `<!-- sponsors --><!-- sponsors -->` markers are still in the page (file unchanged) and the `commit` part of the job says `Working tree clean. Nothing to commit.`
Maybe it's because the sponsors aren't on the repo itself but the organization? Can you check https://github.com/JamesIves/github-sponsors-readme-action/issues/450#issuecomment-1221535046

PR #13807, opened 2023-04-18
feat: update minimum node.js version to 16.x

m1ga:
Current min version is 12.x (not used or supported anymore). Raising this to 16.x

_Side note:_ `titanium` has engines: node > 14.15 already

github actions use node 16 already to build everything


cb1kenobi:
Couple things.

1. Node 14 is end of life in 12 days (4/30/2023)
2. This is a breaking change and we need to wait until Ti SDK 13

So, let's update this PR to bump things to Node.js 16.13.0 (the first 16.x LTS release) and then I put a (new) `semver-major` label on it. I'm not sure if `master` is 13.0.0 or what. If so, we can merge this sooner than later.

m1ga:
I saw the EOL for 14 but I've kept it in so it won't break anything for people using Ti SDK 10 (14 was the minimum there already). Node 12 is gone since 11 month and still in the files :smile: 

But if we raise it to 16 then we should go with Ti SDK 13. 
Since support for 19 is in there already (and I'm using it for a while now) I'm for the min node 16 option but wait until we have Ti 13 :+1: 

It's not anything highly requested, just saw the numbers before and wanted to raise it :smile:

cb1kenobi:
Node.js 20 just dropped: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#20.0.0

m1ga:
Screenshot_20230418_200701

no problem :wink:

m1ga:
Updated the version to min 16, removed node 14 and add 18/20 to the list.

m1ga:
Any update here?

Don't think it is a breaking change as with 12.1.0 the minimum nodejs version was 16.x already:
https://titaniumsdk.com/guide/Titanium_SDK/Titanium_SDK_Getting_Started/Installation_and_Configuration/Titanium_Compatibility_Matrix/#supported-version-of-node-js

anything below that didn't work. So we could include that in 12.3.0 and get rid of the node 20 warning.
@hansemannn I saw you've removed your PR with that change.

hansemannn:
Yep, I closed the PR due to the feedback there (+ the fact that this one exists already). I would prefer to keep maintaining this version of the CLI for now (and merge this PR) if Chris is also okay with it

hansemannn (approved):

PR #13808, opened 2023-04-18
fix(ios/build): support passing target and device id to the module build

ewanharris:
The arguments are copied as is over to the app build process at the end of the module build, there is also a change to return early in `initTiappSettings` in the app build. This is because all build options are ran so `--device-id` to a module build triggers the app build checks, which then subsequently fail. There is definitely a better fix for this, but this for now at least allows it to work and unblocks the related VS Code feature

This is related to #13432

### DIFF
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -1091,12 +1091,10 @@ iOSModuleBuilder.prototype.runModule = function runModule(cli, next) {
 			];
 
 			if (this.target) {
-				buildArgs.push('-T');
-				buildArgs.push(this.target);
+				buildArgs.push('-T', this.target);
 			}
 			if (this.deviceId) {
-				buildArgs.push('-C');
-				buildArgs.push(this.deviceId);
+				buildArgs.push('-C', this.deviceId);
 			}
 
 			runTiCommand(buildArgs, cb);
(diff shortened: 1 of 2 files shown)


iphone/cli/commands/_buildModule.js
```diff
@@ -1078,17 +1081,25 @@ iOSModuleBuilder.prototype.runModule = function runModule(cli, next) {
...
+				'-p', 'ios',
+				'-d', tmpProjectDir,
+				'--no-prompt',
+				'--no-colors',
+				'--no-progress-bars'
+			];
+
+			if (this.target) {
+				buildArgs.push('-T');
+				buildArgs.push(this.target);
```
cb1kenobi, line 1095: non-blocking nitpick:

```suggestion
				buildArgs.push('-T', this.target);
```

iphone/cli/commands/_buildModule.js
```diff
@@ -1078,17 +1081,25 @@ iOSModuleBuilder.prototype.runModule = function runModule(cli, next) {
...
+			}
+			if (this.deviceId) {
+				buildArgs.push('-C');
+				buildArgs.push(this.deviceId);
```
cb1kenobi, line 1099: non-blocking nitpick:

```suggestion
				buildArgs.push('-C', this.deviceId);
```

cb1kenobi (approved):
Changes look good. A couple of nitpicks. I can't easily test this, so we can either have someone test it or yolo.

PR #13811, opened 2023-04-19
fix(ios): build module platform/xcframwork symbolic links

mbender74:
**Optional Description:**
fix to keep symbolic links in /platform/*.XCFramework folder for MacCatalyst xcframeworks when packaging the module.
Else the symlinks of MacCatalyst version of xcframework will be resolved and building the app for target 'macos' will fail!


iphone/cli/commands/_buildModule.js
```diff
@@ -915,12 +915,7 @@ iOSModuleBuilder.prototype.packageModule = function packageModule(next) {
 
 		// 3. platform folder
 		if (fs.existsSync(this.platformDir)) {
-			this.dirWalker(this.platformDir, function (file, name) {
-				var stat = fs.statSync(file);
-				if (name !== 'README.md') {
-					dest.append(fs.createReadStream(file), { name: path.join(moduleFolders, 'platform', path.relative(this.platformDir, file)), mode: stat.mode });
-				}
-			}.bind(this));
+			dest.directory(this.platformDir, path.join(moduleFolders, 'platform'));
```
hansemannn, line 918: Does this method method ensure that the README.md is not zipped up as well?

mbender74: no, unfortunatly not....
for now I did not find out how to do the symbolic link fix with dirWalker... but is this really an issue?

hansemannn: It's a behavioral change on how modules are zipped. I think @cb1kenobi will know more here, as we built the initial script.

cb1kenobi: `dest.directory()` supports a 3rd arg which can be a function: https://github.com/archiverjs/node-archiver/blob/master/lib/core.js#L596-L604. We can use that function to filter out the README.

```suggestion
			dest.directory(
				this.platformDir,
				path.join(moduleFolders, 'platform'),
				entryData => entryData.name === 'README.md' ? false : entryData
			);
```

I can't easily test this code out, so I can't say if it will work. Additionally, it would be helpful if there was a test case.

Interestingly, I see `directory()` will tack the `stats` onto the `entryData` which should include whether the file is a symlink, so it does seem like this is a more robust method.

cb1kenobi (changes requested):

iphone/cli/commands/_buildModule.js
```diff
@@ -915,12 +915,11 @@ iOSModuleBuilder.prototype.packageModule = function packageModule(next) {
...
-			this.dirWalker(this.platformDir, function (file, name) {
-				var stat = fs.statSync(file);
-				if (name !== 'README.md') {
-					dest.append(fs.createReadStream(file), { name: path.join(moduleFolders, 'platform', path.relative(this.platformDir, file)), mode: stat.mode });
-				}
-			}.bind(this));
+			dest.directory(
+				this.platformDir,
+				path.join(moduleFolders, 'platform'),
+				(entryData) => entryData.name === 'README.md' ? false : entryData
```
cb1kenobi, line 921: So, did you test it? Does the `README.md` get excluded from the zip file?

mbender74: yes!
when only the 'README.md' is in the /platform folder (no xcframework as content), the module builds and in the zip the platform folder is also not included.
when  'README.md' and xcframework is included in the platform folder, only the xcframework is in the zipped module included.

works very well!

cb1kenobi, line 921: Bah, linting doesn't like the syntax I suggested. I suppose you either need to wrap the return value in parens or use curly braces with a `return`.

```suggestion
				(entryData) => (entryData.name === 'README.md' ? false : entryData)
```

or if needed:

```
				(entryData) => { return entryData.name === 'README.md' ? false : entryData; }
```

cb1kenobi (changes requested):

iphone/cli/commands/_buildModule.js
hansemannn: Yep, that one helped. VSCode shows it locally as well

cb1kenobi (approved):

PR #13812, opened 2023-04-20
fix(ios): enable JSContext to be inspectable

ewanharris:
In iOS 16.4 the default state for whether a JSContext is inspectable became false and the owner of the context must explicitly set it to be inspectable.

This commit enables the JSContext to be inspectable and I have verified this works for both Safari and VS Code debugging. However, for security concerns it's probably best that we disable this in production I assume. How to do that, I have no idea. It probably also needs to be guarded to only be set on iOS 16.4+

To test this just open an app on iOS then go Safari-> Develop -> _simulator/device name_. With this fix, you should see a JSContext listed

Refs
https://webkit.org/blog/13966/webkit-features-in-safari-16-4/ 
https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
@@ -793,7 +793,10 @@ static JSValueRef StringFormatDecimalCallback(JSContextRef jsContext, JSObjectRe
 {
   pthread_mutex_lock(&KrollEntryLock);
   context = JSGlobalContextCreate(NULL);
-  JSGlobalContextSetInspectable(context, true);
+  // Ensure that the JSContext is debuggable
+  if (@available(iOS 16.4, *)) {
+    JSGlobalContextSetInspectable(context, true);
+  }
   JSObjectRef globalRef = JSContextGetGlobalObject(context);
 
   if (appJsKrollContext == nil) {
--- a/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
@@ -793,10 +793,6 @@ static JSValueRef StringFormatDecimalCallback(JSContextRef jsContext, JSObjectRe
 {
   pthread_mutex_lock(&KrollEntryLock);
   context = JSGlobalContextCreate(NULL);
-  // Ensure that the JSContext is debuggable
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
-  JSGlobalContextSetInspectable(context, true);
-#endif
   JSObjectRef globalRef = JSContextGetGlobalObject(context);
 
   if (appJsKrollContext == nil) {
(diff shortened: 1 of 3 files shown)


iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
```diff
@@ -793,6 +793,10 @@ - (void)main
 {
   pthread_mutex_lock(&KrollEntryLock);
   context = JSGlobalContextCreate(NULL);
+  // Ensure that the JSContext is debuggable
+  if (@available(iOS 16.4, *)) {
```
ewanharris, line 797: I thought this would work based on the usage in KrollPromise but I guess not?

iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
```diff
@@ -793,6 +793,10 @@ - (void)main
 {
   pthread_mutex_lock(&KrollEntryLock);
   context = JSGlobalContextCreate(NULL);
+  // Ensure that the JSContext is debuggable
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
+  JSGlobalContextSetInspectable(context, true);
```
hansemannn, line 798: Small thing: In the ObjC world, `YES` is favored over `true` and `TRUE`.

Besides that, great change! It's just a bit complex to disable this per target, as TitaniumKit is precompiled during the SDK build, but I think we have some prod-related checks that we could reuse here. Checking …

hansemannn: Might found a solution:
```objc
  BOOL isProduction = [TiApp.tiAppProperties[@"ti.deploytype"] isEqualToString:@"production"];
```

ewanharris: Just pushed c70ebd011232460f4ad4dd2b1e3fab6b310995d9 with those changes. I don't have production builds setup, but I was able to test that changing to `if (isProduction) {`  causes the JSContext to not be inspectable in dev builds so I think it should be good

hansemannn: @ewanharris That one didn't work, yet, as it's deprecated. So if users don't manually add `<string name="ti.deploytype"">production</string>` in their tiapp.xml, it will not work. Really hard to get these config so early in the app lifecycle

ewanharris: Ah shoot :/ I think we can defer the setting of this potentially to be later in the lifecycle if needed as it can be toggled on/off AFAIK. Is there a place that might be better to do that?

Feel free to push to the PR if you wanna rework it at all

hansemannn: I tried some places already, which all worked in dev, but crashed in prod builds (tested via testflight). I'll revisit it tomorrow 👍

hansemannn: Found something promising! Doing some production tests and then push here.

hansemannn:
@ewanharris Okay, I finally found a solution - it enables it for dev (sim) and test (device), but disables it for prod builds.

hansemannn (approved):

ewanharris:
Awesome, thank you so much for finishing this off @hansemannn!

ewanharris:
Built locally and the changes look good! Created #13817 to backport some recent CI changes as there seems to be issues building Android on 12_1_X due to the Ubuntu 18 removal

PR #13814, opened 2023-04-22
chore: optimize test output

m1ga:
adding two new tests to the `npm`  scripts:
`npm run test:android:onlyFailed` and `npm run test:iphone:onlyFailed`

**What it does**
It will skip outputting all successful tests in the lists at the end. This way it is easier to spot failed tests. 

**Why**
Because we have over 5000 tests and if the 20th tests failed you will only see "failed 1" at the end but it might not be possible to scroll up the whole list to position 20 :smile: 
The rest of the output stays the same.

```
5167 Total Tests: 5167 passed, 0 failed, 0 skipped.
```
and the list of all tests above, just not the `✓ success...` rows

### DIFF
--- a/build/lib/test/test.js
+++ b/build/lib/test/test.js
@@ -42,6 +42,8 @@ const OS_VERSION_PREFIX = 'OS_VERSION: ';
 // Sniff if we're on Travis/Jenkins
 const isCI = !!(process.env.BUILD_NUMBER || process.env.CI || false);
 
+var showFailedOnly = false;
+
 /**
  * Generates a test app, then runs the app for each platform with our
  * test suite. Outputs the results in a JUnit test report,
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -61,12 +61,15 @@
...
     "test:cli": "JUNIT_REPORT_PATH=junit.cli.report.xml nyc mocha **/cli/tests/test-*.js build/**/test-*.js --reporter=mocha-jenkins-reporter",
     "test:ios": "npm run test:iphone",
     "test:iphone": "npm run cleanbuild:ios -- --skip-zip --no-docs --symlink && ti sdk select $npm_package_version && npm run test:integration -- ios -F iphone",
+    "test:iphone:onlyFailed": "npm run cleanbuild:ios -- --skip-zip --no-docs --symlink && ti sdk select $npm_package_version && npm run test:integration:onlyFailed -- ios -F iphone",
     "test:ipad": "npm run cleanbuild:ios -- --skip-zip --no-docs --symlink && ti sdk select $npm_package_version && npm run test:integration -- ios -F ipad",
     "test:mac": "npm run cleanbuild:ios -- --skip-zip --no-docs --symlink && ti sdk select $npm_package_version && npm run test:integration -- ios -T macos",
-    "test:integration": "node ./build/scons test"
+    "test:integration": "node ./build/scons test",
+    "test:integration:onlyFailed": "node ./build/scons test --only-failed true"
```
cb1kenobi, line 72: Why can't `--only-failed` be a flag?

m1ga: :facepalm: you are right! It works as a flag too. I've adjusted the part

build/lib/test/test.js
```diff
@@ -42,6 +42,8 @@ const OS_VERSION_PREFIX = 'OS_VERSION: ';
 // Sniff if we're on Travis/Jenkins
 const isCI = !!(process.env.BUILD_NUMBER || process.env.CI || false);
 
+var showFailedOnly = false;
```
cb1kenobi, line 45:
```suggestion
let showFailedOnly = false;
```

cb1kenobi (approved):

PR #13821, opened 2023-04-27
ci: compile on macos13/xcode 14.3

ewanharris:
GitHub [released these](https://github.blog/changelog/2023-04-24-github-actions-macos-13-is-now-available/) this week so we should look to move to using them

Also fixes a warning from release script due to an outdated github-script action version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,12 +25,12 @@ jobs:
           java-version: '11'
 
   ios:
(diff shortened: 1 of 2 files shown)


cb1kenobi:
@hansemannn, @m1ga are we good to update to Xcode 14.3?

hansemannn (approved):
The build seems stable, so yes, I'm good here!

PR #13826, opened 2023-05-08
fix(android): tableView row parent

m1ga:
Split out of the "optimize tableview updates" PR to test it separately. 

Will keep the click event when reassigning the data:

```js
var menu = [];
var numrows = 0;

  if (e.index == numrows - 1)
    AddRows();
});

function AddRows() {
  if (menu.length == 0)
    row.add(lab);
  menu.push(row);
  numrows++;
  menu.push(row);
  numrows++;
  table.data = menu;
};

AddRows();
self.add(table);
self.open();
```

click on the last row to add more rows and then click on any row above it. The first row should update with the correct clicked number and there is a console output.


hansemannn (approved):

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
```diff
@@ -190,6 +190,7 @@ private void appendRowInternal(Object rows, KrollDict animation, boolean interna
 					row.getProperties().getString(TiC.PROPERTY_FOOTER)));
 
 			// Add row to section.
+			row.setParent(section);
```
prashantsaini1, line 193: @m1ga This is already done by `section.add(row)` in next call. I am not sure what it would fix :).

m1ga: Please run the example code and you can see that it didn't use the event listeners when you reasign data to the table. `section.add(row)` is adding the row to the section, The PR is setting the parent to the row again. It looks like this is needed to get the event listener from the table again

prashantsaini1:
@m1ga This PR broke `TableView`. Since `TableView` is supposed to be used to access and alter child-views directly, but setting the row's parent before adding caused the row to be cloned from initial state, leaving further changes not applied.

Sharing a simple example below where image after timeout does not render on 12.6.1.GA, but renders fine on lower versions.

**IMPORTANT NOTES:**
1. The purpose you were trying to achieve through this PR could already be achieved using `TableView's update../append..` methods which were created for these purposes. 
2. `TableView's data` property must be set only with newly created `TableViewRow`s again. You can see the setData code in `TableViewProxy` which clears out all sections before proceeding.
3. If we only need to update some rows/sections, then use methods mentioned in 1.

<hr>
<details>
<summary>Sample code</summary>

```

function AddRows() {
	const rows = [];
			title: "Row " + i,
			color: 'black',
			height: 100
		});

		row.add(createRowView());
		
		rows.push(row);
	}
	
	table.data = rows;
};

function createRowView() {
		height: 80,
		width: 120
	});
	
	setTimeout(() => {
	}, 1000);

	return imageView;
}

self.add(table);
self.open();
AddRows();

```
</details>

If you want, I can rewrite your code to behave as you were expecting on 12.5.1.GA.

m1ga:
Did you see https://github.com/tidev/titanium-sdk/pull/14169 yet? This looks almost like the code I got on Slack to test it :) The draft PR fixed the issue for the slack user

m1ga:
Your example code works fine with the draft PR. I can see all images

prashantsaini1:
Please check IMPORTANT NOTES. These changes are not required🙂.

m1ga:
But it looks like it was possible to access and update content in previous Ti versions without calling update (note 1).
And without this fix the events did not work, which was also a thing that broke in the past and was working before. I don't know what the correct way is :smile: Just trying to fix the bugs people report

m1ga:
started https://github.com/tidev/titanium-sdk/pull/14170 to fix it

PR #13830, opened 2023-05-10
feat(all): add `forceUpdates` propterty to listView to improve scrolling event

m1ga:
* new parameter `forceUpdates` for ListView `continuousUpdate`
* `top` value will be included for the y-scoll position

https://github.com/tidev/titanium_mobile/assets/4334997/061d6435-d3bf-4b66-9cf7-d009728c508c

Usefull if you want to animate other views during scrolling.

```js
var toggle = false;
var sections = [];
	title: "change",
	bottom: 20
});
	top: 10,
	right: 10,
	width: Ti.UI.SIZE,
	height: Ti.UI.SIZE
})
	height: Ti.UI.FILL,
	width: Ti.UI.FILL,
	continuousUpdate: true,
  forceUpdates: false
});

	toggle = !toggle;
	listView.forceUpdates = toggle
	console.log(listView.forceUpdates);
})

for (var s = 0; s < 5; s++) {
		headerTitle: 'Section ' + s
	});
	var set = [];
		set.push({
			properties: {
				title: 'Item 0 ' + i
			}
		})
	}
	section.setItems(set);
	sections.push(section);
}

listView.sections = sections;

	console.log("Item", e.direction, e.firstVisibleItemIndex, e.firstVisibleSectionIndex, e.top);
})
win.open();
```

**Tests**
* run app
* start scrolling
* log only shows up when the top item changes (current default)
* click "toggle"
* log will fire all the time during scrolling, `top` value is available

I checked 12.1.1.GA and this PR with `continuousUpdate` enabled and not in the profiler and it is the same CPU usage (<15%) when you scroll so it is not a performance issue to fire the event all the time!


m1ga:
@hansemannn can you help me to find the y scroll position in this line (currently -1):

https://github.com/tidev/titanium_mobile/pull/13830/files#diff-1a2a8f1ea092b927374c5e1a188c5d1629165b4b736b79889496d15854969ba9R2019

It should be the y position of the top of the listview (see the video in the first post).
The parameter works already but I can't find the y-position :-(

hansemannn:
The video is unfortunately broken, can you reupload?

m1ga:
Video should still work, tested it on my Mac and Linux machine. But the line already worked :smile: Thanks!
So you can get the scoll y position during scrolling a listview for Android and iOS now. Now people can move out views depending on that position

m1ga:
As discussed in a PR meeting with Chris today:

I removed the parameter again and just made it a default value! There was no benefit of having it disabled when using `continouseUpdates`!

iphone/Classes/TiUIListView.m
```diff
@@ -2013,13 +2012,9 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
...
           [eventArgs setValue:[section itemAtIndex:[indexPath row]] forKey:@"firstVisibleItem"];
+          [eventArgs setValue:NUMINTEGER(topSpacing) forKey:@"top"];
 
-          if (lastVisibleItem != [indexPath row] || lastVisibleSection != [indexPath section]) {
-            // only log if the item changes
-            [self.proxy fireEvent:@"scrolling" withObject:eventArgs propagate:NO];
-            lastVisibleItem = [indexPath row];
-            lastVisibleSection = [indexPath section];
-          }
+          [self.proxy fireEvent:@"scrolling" withObject:eventArgs propagate:NO];
```
hansemannn, line 2017: I am concerned about the performance impact of this change, as it will now always fire and not only if the item changed. The reason the check was added initially was also performance-related, so this could affect lists even if not using the event.

cb1kenobi: Web browsers fire the `onscroll` event constantly, so it was my suggestion to do the same and only always fire the event as scrolling. Is there a way to only fire the event if there are listeners?

hansemannn: The web browser does not have the kroll proxy behind it that blocks the main thread every time :) It's a really critical piece of code here.

m1ga: I've added the `only change when a new item` code so it won't fire the event all the time without new info since there wasn't a y-coordinate in the response before. It was just unnecessary to see it all the time when there was no change.  I didn't do that for performance reasons BUT I didn't do performance analysis on iOS :smile: 

On Android I didn't see any impact in CPU usage with or without this PR. 
But of course it won't be useful if it blocks the main thread for other parts in your app!

I'm happy to add it behind a flag again so you can enable/disable it and it will only fire when you need the value to constantly change:
* default is still normal scroll behavior (firing only when you change direction) - current default in 12.1.1.GA
* you can enable `continuousUpdate` and that will fire every time a new item is at the top (also already in 12.1.1GA)
* add another flag `forceUpdates` that will fire all the time (new way) - with a hint in the docs that it should only be used with caution and when you have a reason for it e.g. moving out elements depending on your scroll depth.

hansemannn: Yep, a flag may actually be good, at least until it's wider adopted. Are there specific community requests or a personal dependency on your end to add it?

m1ga: community request for the `top` value to build something like this http://jsfiddle.net/mariusc23/s6mLJ/31/ in Titanium. And it's only useful if the `top` is reported all the time and not only when the item changes (which is the current workaround)

m1ga: back to the original PR: `forceUpdates` parameter is back :smile:

apidoc/Titanium/UI/ListView.yml
```diff
@@ -951,14 +951,25 @@ properties:
...
 
+  - name: forceUpdates
+    summary: Optimize the `continuousUpdate` scrolling event.
+    description: |
+        If `continuousUpdate` is enabled and you set `forceUpdates` to `true` it will fire the scrolling](Titanium.UI.ListView.scrolling) all the time.
+        Otherwise it will only update when a new item is scrolled in (default).
+    type: Boolean
+    default: false
+    platforms: [android, iphone, ipad, macos]
+    since: 12.2.0
```
m1ga, line 971:
```suggestion
    since: 12.3.0
```

apidoc/Titanium/UI/ListView.yml
```diff
@@ -951,14 +951,25 @@ properties:
     since: 3.2.0
 
   - name: continuousUpdate
-    summary: Determines if the scrolling event should fire every time there is a new visible item.
+    summary: Determines if the scrolling event should fire every time the user is dragging the ListView.
```
m1ga, line 954:
```suggestion
    summary: Determines if the scrolling event should fire every time there is a new visible item.
```

apidoc/Titanium/UI/ListView.yml
```diff
@@ -953,12 +953,23 @@ properties:
   - name: continuousUpdate
     summary: Determines if the scrolling event should fire every time there is a new visible item.
     description: |
-        When setting `continuousUpdate` to `true` the [scrolling](Titanium.UI.ListView.scrolling) event will fire every time a new item is scrolled in.
+        When setting `continuousUpdate` to `true` the [scrolling](Titanium.UI.ListView.scrolling) event will fire every time.
```
m1ga, line 956:
```suggestion
        When setting `continuousUpdate` to `true` the [scrolling]
```

apidoc/Titanium/UI/ListView.yml
```diff
@@ -953,12 +953,23 @@ properties:
...
+        Otherwise it will only change when you change the scroll direction.
```
m1ga, line 957:
```suggestion
        (Titanium.UI.ListView.scrolling) event will fire every time a new item is scrolled in.
```

cb1kenobi (approved):
Looks good! **APPROVED!**

PR #13835, opened 2023-05-12
feat(android): webview injectEventBindings method for external pages

m1ga:
* in the last ["to arrow function" change](https://github.com/tidev/titanium_mobile/pull/13751/commits/a48e92eb4a174b498ed5d467a308de85dbe5b8dd#diff-3b2ea0af3c749abb7b7ad6e90532bf737d59eec880103d4914f2a9c50ef3c2b8L438-L441) on parameter was missed causing it the return an error.

* Adding `injectEventBindings` to the Android webview so you don't have to run a custom evalJS script yourself in the app to inject the event bindings into external pages.
* minified `bindings.js.min` again

I still keep the example in the docs if you want to add it manually in case you'll need to make changes (but with less code!). For iOS it is not needed since you can use `window.webkit.messageHandlers.yourHandlerName.postMessage...` to communicate with the app right away. The manual evalJS code could be used for iOS too!

```js

	title: "add"
});
	url: "https://example.com"
});

win.add([webview, btn]);
win.open();

	webview.injectEventBindings();
});
```

**Before**
Screenshot_20230512_104730

**After clicking the button:**
Screenshot_20230512_111113

Then you can add other scripts that will fire `Ti.App.fireEvents()` to your page.


android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -130,6 +130,15 @@ public Object evalJS(String code, @Kroll.argument(optional = true) KrollFunction
 		return view.getJSValue(code);
 	}
 
+	@Kroll.method
+	public void injectCode()
```
cb1kenobi, line 134: Should this method be called `injectEventBindings()`?

```suggestion
	public void injectEventBindings()
```

android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -130,6 +130,15 @@ public Object evalJS(String code, @Kroll.argument(optional = true) KrollFunction
...
+	{
+		TiUIWebView view = (TiUIWebView) peekView();
+		if (view != null) {
+			view.injectCode();
```
cb1kenobi, line 138:
```suggestion
			view.injectEventBindings();
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -647,6 +647,22 @@ public void setUrl(String url)
...
+			.replace("\n", "").replace("\r", "");
+		code.append("(function addBinding(){");
+		code.append("var s=document.createElement(\"script\");");
+		code.append("s.setAttribute(\"type\",\"text/javascript\");");
+		code.append("s.innerHTML=\"" + outCode + "\";");
+		code.append("document.getElementsByTagName(\"body\")[0].appendChild(s);");
+		code.append("})()");
+		Log.i("---", code.toString());
+		getJSValue(code.toString());
+	}
```
cb1kenobi, line 664: Since the heavy lifting is all done in the `TiWebViewBinding`, can we move this logic there and make this function a passthrough?

```suggestion
	public void injectEventBindings()
	{
		client.getBinding().injectEventBindings();
	}
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java
```diff
@@ -120,7 +120,7 @@ public void destroy()
 		destroyed = true;
 	}
 
-	private static StringBuilder readResourceFile(String fileName)
+	public static StringBuilder readResourceFile(String fileName)
```
cb1kenobi, line 123: No longer needed if the `injectEventBindings()` logic is moved into this file.

```suggestion
	private static StringBuilder readResourceFile(String fileName)
```

cb1kenobi (changes requested):

apidoc/Titanium/UI/WebView.yml
```diff
@@ -315,6 +320,12 @@ methods:
   - name: goForward
     summary: Goes forward one entry in this web view's history list, if possible.
 
+
+  - name: injectCode
```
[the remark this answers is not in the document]
m1ga: I've changed it to `injectEventBindings` already but didn't commit the code :facepalm:

apidoc/Titanium/UI/WebView.yml
```diff
@@ -35,23 +35,28 @@ description: |
...
         'document.getElementsByTagName("body")[0].appendChild(s);' +
       '})()'
     );
     ```
-    The `binding.min.js` is available in the [repository](https://github.com/tidev/titanium_mobile/tree/master/android/modules/ui/assets/Resources/ti.internal/webview).
+    The `binding.min.js` file is available in the [repository](https://github.com/tidev/titanium_mobile/tree/master/android/modules/ui/assets/Resources/ti.internal/webview).
 
-    For iOS check the example in <Titanium.UI.WebView.addScriptMessageHandler>.
+    <b>iOS</b>:
+    For iOS have a look at the example in <Titanium.UI.WebView.addScriptMessageHandler> for custom handler names or if you run the manual evalJS version
```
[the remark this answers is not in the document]
m1ga: Parity would be good. I just didn't know how to integrate the `injectEventBindings()` in ObjectivC :blush:  That's why I wanted to keep the evalJS example since that is working on both platforms. `window.webkit.messageHandlers.yourHandlerName` is not available in the Android webview so we can't directly map that. 

The default Ti way to communicate from webview -> app is still `Ti.App.fireEvent` and that is cross-platform. The docs were just always wrong saying that you can't use it with external URL, you could always inject that code with evalJS (or if it is your external page: just put the file on your server and load it right away) :smile: 

I'll try to see if I can add the iOS part!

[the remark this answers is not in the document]
cb1kenobi: Wow, this PR is looking good! We can dive into the iOS stuff during our meeting. Great job!

apidoc/Titanium/UI/WebView.yml
```diff
@@ -315,6 +321,12 @@ methods:
   - name: goForward
     summary: Goes forward one entry in this web view's history list, if possible.
 
+
+  - name: injectEventBindings
+    summary: Injects the Titanium event bindings from bindings.js code into an external webpage.
+    since: "12.2.0"
```
m1ga, line 327:
```suggestion
    since: "12.3.0"
```

m1ga:
:thinking: I broke it after merging master....strange. Master is still building fine, this PR is failing. Checking at the moment

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java
```diff
@@ -149,6 +149,22 @@ private static StringBuilder readResourceFile(String fileName)
...
+		StringBuilder tiCode = TiWebViewBinding.readResourceFile("binding.min.js");
+		String outCode = tiCode.toString().replaceAll("\"", "\\\\\"")
+			.replace("\n", "").replace("\r", "");
+		code.append("(function addBinding(){");
+		code.append("var s=document.createElement(\"script\");");
+		code.append("s.setAttribute(\"type\",\"text/javascript\");");
+		code.append("s.innerHTML=\"" + outCode + "\";");
+		code.append("document.getElementsByTagName(\"body\")[0].appendChild(s);");
+		code.append("})()");
+		Log.i("---", code.toString());
```
m1ga:
```suggestion
```

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java
```diff
@@ -149,6 +149,21 @@ private static StringBuilder readResourceFile(String fileName)
 		return code;
 	}
 
+	public void injectEventBindings()
+	{
+		StringBuilder code = new StringBuilder();
+		StringBuilder tiCode = TiWebViewBinding.readResourceFile("binding.min.js");
+		String outCode = tiCode.toString().replaceAll("\"", "\\\\\"")
+			.replace("\n", "").replace("\r", "");
+		code.append("(function addBinding(){");
```
cb1kenobi, line 158: **Injection is not idempotent — a second call silently breaks `Ti.App` events** (severity: high)

Each call appends a new `<script>` whose body re-executes `var Ti={_event_listeners:[],...},Titanium=Ti;`. The global `Ti` is replaced with a fresh object whose `_event_listeners` array is empty, but the native side (`TiApp`) still holds the systemId→index mappings created against the previous object.

Failure mode: a page calls `Ti.App.addEventListener('foo', cb)` after the first injection; the user (or app logic) calls `webview.injectEventBindings()` again — e.g. the button in the PR description tapped twice, or on a page that already has bindings. From then on `Ti.executeListener(index, data)` from native hits `getEventListenerByKey('index', …)` on the empty array, returns `null`, and the callback is never invoked. Events stop firing with no error anywhere.

Simplest fix: make the injected wrapper a no-op when bindings are already present:

```java
code.append("(function addBinding(){");
code.append("if(window.Ti&&window.Ti._event_listeners)return;");
```

—
Reviewed 94227d6e

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiWebViewBinding.java
```diff
@@ -149,6 +149,21 @@ private static StringBuilder readResourceFile(String fileName)
...
+		code.append("var s=document.createElement(\"script\");");
+		code.append("s.setAttribute(\"type\",\"text/javascript\");");
+		code.append("s.innerHTML=\"" + outCode + "\";");
+		code.append("document.getElementsByTagName(\"body\")[0].appendChild(s);");
```
cb1kenobi, line 162: **Injection fails silently when `document.body` does not exist yet** (severity: medium)

`document.getElementsByTagName("body")[0]` is `undefined` before the body element is parsed (call made right after `setUrl()`/during load, on a redirect, or on a non-HTML document). `.appendChild` then throws a `TypeError`, and because `getJSValue()` wraps the expression in a `try{...}catch(ti_eval_err){return "";}`, the exception is swallowed — the method returns normally and the caller has no way to know the bindings were never installed.

Simplest fix: append to whichever container is available:

```java
code.append("(document.body||document.documentElement).appendChild(s);");
```

(and consider returning the success flag from the IIFE so callers can react).

—
Reviewed 94227d6e

apidoc/Titanium/UI/WebView.yml
```diff
@@ -313,6 +318,12 @@ methods:
   - name: goForward
     summary: Goes forward one entry in this web view's history list, if possible.
 
+
+  - name: injectEventBindings
+    summary: Injects the Titanium event bindings from bindings.js code into an external webpage.
+    since: "12.3.0"
```
cb1kenobi, line 324: **`since` version is stale** (severity: low)

`since: "12.3.0"` was set when this PR was first opened; 12.3.0 has long since shipped without this method. Merging as-is tells users the API is available in SDK releases where calling `webview.injectEventBindings()` throws "method does not exist", and the generated docs/compatibility tables will be wrong.

Simplest fix: bump `since` to the next unreleased SDK version at merge time.

—
Reviewed 94227d6e

cb1kenobi (commented):
The new injectEventBindings() method re-defines the global Ti object every time it is called, which wipes the JS-side listener registry and silently breaks events that were registered before the second call. It also assumes document.body already exists, so an early call fails silently because getJSValue swallows the exception. Finally the apidoc marks the method as since 12.3.0, a version that has already shipped. Add an 'already injected' guard, fall back to documentElement, and bump the since version.

—
Reviewed 94227d6e

PR #13848, opened 2023-06-18
feat(android): use EncryptedSharedPreferences

### LINKED ISSUE #13846: Enable support for EncryptedSharedPreferences

### Description

We've had white hat hackers test our system for vulnerabilities. One of the things they pointed out was the current use of "SharedPreferences" (Ti.App.properties). 

### Solution

The hackers suggested we should use "EncryptedSharedPreferences" instead.
https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences.

### Alternatives

_No response_

### Platforms

Android

m1ga:
**:warning:  NOTE: we'll have to wait for a stable https://developer.android.com/jetpack/androidx/releases/security 1.1.0 :warning:**

---

fixes https://github.com/tidev/titanium-sdk/issues/13846

Adds `Ti.App.Properties.encrypt` to store EncryptedSharedPreferences.

## Test

```js
Ti.App.Properties.setString("string", "plain value");
console.log("not encrypted - should be 'plain value': " + Ti.App.Properties.getString("string", "default"));

Ti.App.Properties.encrypt = true;
Ti.App.Properties.setString("string", "secure value");
console.log("encrypted - should be 'secure value': " + Ti.App.Properties.getString("string", "default"));

Ti.App.Properties.encrypt = false;
Ti.App.Properties.setString("string", "plain value 2");
console.log("not encrypted - should be 'plain value 2': " + Ti.App.Properties.getString("string", "default"));

Ti.App.Properties.encrypt = true;
console.log("encrypted - should be 'secure value': " + Ti.App.Properties.getString("string", "default"));
```

Stores the values in an encrypted xml:

Screenshot_20230618_142738

## Todo
* [x] Extend test
* [x] Docs


apidoc/Titanium/App/Properties/Properties.yml
```diff
@@ -158,97 +158,107 @@ methods:
...
+
+properties:
+  - name: useEncryption
+    summary: |
+        If enabled it will use Androids [EncryptedSharedPreferences](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences)
+        to store properties.
+    type: String
+    permission: read-only
+    platforms: [android]
+    since: {android: "12.3.0"}
```
m1ga, line 251:
```suggestion
    since: {android: "12.4.0"}
```

apidoc/Titanium/App/Properties/Properties.yml
```diff
@@ -157,97 +157,107 @@ methods:
...
       - name: value
         summary: Property value.
         type: String
-        
+
+properties:
+  - name: useEncryption
+    summary: |
+        If enabled it will use Androids [EncryptedSharedPreferences](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences)
+        to store properties.
```
hansemannn, line 246: `EncryptedSharedPreferences` has been deprecated lately, see the link you've used. It's deprecated in 1.1.0-alpha-7 and we should target 1.0.0, but if the APIs in the main SharedPreferences expose the necessary APIs already, those should be used directly, so we don't need to migrate it again.

Otherwise, this PR should wait for a stable 1.1.0.

hansemannn (changes requested):

android/titanium/src/java/org/appcelerator/titanium/TiProperties.java
```diff
@@ -24,8 +27,10 @@ public class TiProperties
 {
 	private static final String TAG = "TiProperties";
 	private static JSONObject systemProperties;
+	public static boolean useEncryption = false;
 
 	SharedPreferences preferences;
+	SharedPreferences preferencesPlain;
```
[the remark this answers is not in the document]
m1ga: `preferences` will be the normal encrypted or plain preference object throughout the whole file. So that name is correct. 
`preferencesPlain` is the none-encrypted version I only use to assign to `preferences` in the beginning or when you get elements. I could rename it to `noneEncryptedPreferences` but I like `plain` better in this case.

apidoc/Titanium/App/Properties/Properties.yml
m1ga: yeah `MasterKey` was added in 1.1.0-alpha01. I'll check for the deprecation message, updated to 1.1.0-alpha07 and I see it now too. We'll have to wait for the final version 1.1.0.

m1ga:
according to https://www.spght.dev/articles/28-05-2024/jetsec-deprecation the way to do it is just use the normal 
`SharedPreferences` without encryption:
> Additionally, as of Android 10 (Q), all devices encrypt user data on the filesystem by default, adding another level of security.

so you can only access it from a rooted phone and you shouldn't store sensitive data in it. Or encrypt it by hand and store it.

**Conclusion:**
I think we can just close this PR as `EncryptedSharedPreferences` still works in the alpha version but as we don't know how long it will be deprecated we should just not start to use it

hansemannn:
Makes sense! Still thank you for the effort to research this and do the PoC :-)

PR #13871, opened 2023-06-26
Rename dependabot.yml to dependabot.yml.disable

cb1kenobi:
We keep getting Dependabot notifications despite being turned off. I suspect it's because the config file exists. I simply renamed it to hopefully disable Dependabot.


m1ga:
sounds good, hopefully that will stop it 😄

hansemannn (approved):

PR #13874, opened 2023-06-28
feat(android): badgeTextColor property

m1ga:
Currently you can only set the badgeColor (background) but not the text color. `badgeTextColor` can be used to do that now.

added  `badgeBackgroundColor` and added a deprecation warning to `badgeColor` to use the new property

```js
const isAndroid = Ti.Platform.osname === 'android';
let tabGroup;

	style: isAndroid ? Ti.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION : undefined,
	tabs: [
		Ti.UI.createTab({
			window,
			title: "test"
		}),
		Ti.UI.createTab({
			window:Ti.UI.createWindow(),
			title: "test"
		})
	]
});

//tabGroup.tabs[0].badgeColor = "black" // deprecated - shows a warning
tabGroup.tabs[0].badgeBackgroundColor = "black"
tabGroup.tabs[0].badgeTextColor = "white"
tabGroup.tabs[0].badge = 1

tabGroup.open();
```

Screenshot_20230628-122134

Code is a copy of the `badgeColor` part and just using `setBadgeTextColor` instead of `setBackgroundColor`

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/TabProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/TabProxy.java
@@ -25,6 +25,7 @@ import android.app.Activity;
 		TiC.PROPERTY_ACTIVE_TITLE_COLOR,
 		TiC.PROPERTY_BADGE,
 		TiC.PROPERTY_BADGE_COLOR,
+		TiC.PROPERTY_BADGE_TEXT_COLOR,
 		TiC.PROPERTY_ICON,
 		TiC.PROPERTY_TINT_COLOR,
 		TiC.PROPERTY_TITLE,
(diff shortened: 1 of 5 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/TabProxy.java
```diff
@@ -25,6 +25,7 @@
 		TiC.PROPERTY_ACTIVE_TITLE_COLOR,
 		TiC.PROPERTY_BADGE,
 		TiC.PROPERTY_BADGE_COLOR,
```
hansemannn, line 27: Can we deprecate this one and move it to `badgeBackgroundColor` for improved consistency to other similar APIs?

m1ga: sure. update the PR. It will show a warning to use `badgeBackgroundColor` but both values will continue to work

m1ga:
added `badgeBackgroundColor` and made a deprecation warning in `badgeColor` to use the new property now

hansemannn (approved):
Great change, thank you!

PR #13879, opened 2023-07-04
fix(ios): enable webview inspection

m1ga:
Follow up to this PR: https://github.com/tidev/titanium-sdk/pull/13812

This one only enabled inspection for the JSContext and not the the webview itself.
https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/
has two parts for that.

Slack thread: https://tidev.slack.com/archives/C03CVQX2A/p1688465009923559


iphone/Classes/TiUIWebView.m
```diff
@@ -83,7 +83,9 @@ - (WKWebView *)webView
     _willHandleTouches = [TiUtils boolValue:[[self proxy] valueForKey:@"willHandleTouches"] def:YES];
 
     _webView = [[WKWebView alloc] initWithFrame:[self bounds] configuration:config];
-
+#if TARGET_OS_SIMULATOR
+    _webView.inspectable = YES;
```
hansemannn, line 87: The API is only available on iOS 16.4+, so you also need the iOS version check here, otherwise it will crash on older devices

m1ga: ah, good point. I've added the version check from the other PR. Will have to check it on my Mac later

m1ga:
tested with 16.4 simulator works fine but it doesn't use the version code check so it will still crash on 15.5 :thinking:

iphone/Classes/TiUIWebView.m
```diff
@@ -83,7 +83,9 @@ - (WKWebView *)webView
     _willHandleTouches = [TiUtils boolValue:[[self proxy] valueForKey:@"willHandleTouches"] def:YES];
 
     _webView = [[WKWebView alloc] initWithFrame:[self bounds] configuration:config];
-
+#if TARGET_OS_SIMULATOR && __IPHONE_OS_VERSION_MAX_ALLOWED >= 160400
+    _webView.inspectable = YES;
```
hansemannn: This one is still missing the `[TiUtils isiOSVersionOrGreater:@"16.4"]` check, which is used as a runtime check. The other ones ensure that you can build the app with Xcode < 14.3 (pre-compile statement) :-)

hansemannn (changes requested):

hansemannn (approved):

m1ga:
Just saw that the other PR is using `if (@available(iOS 16.4, *)) {` now ([source](https://github.com/tidev/titanium-sdk/blob/4202779b6105bd729ef5df9769aa2a0e553d866d/iphone/Classes/AppModule.m#L384)).

That works now. 15.5 and 16.4 simulators are now working

hansemannn:
If possible, please still add the `__IPHONE_16_4` statement to retain Xcode < 13.4 compatibility

m1ga:
ah I see! I've added the #if.. part again :+1:

PR #13892, opened 2023-08-02
fix(android): adjust picker min/maxDate

m1ga:
Currently the picker (date) shows a different max value when you run this on Android and iOS.
Android will go up to the 19th and iOS to the 20th.

```js

	type: Ti.UI.PICKER_TYPE_DATE,
	maxDate: new Date(2023, 7, 20),
});

win.add(picker);
win.open();
```

The issue is that before it uses `cal.setTime(inDate);` and moved the date:
```
inDate: Thu Aug 03 00:00:00 GMT+02:00 2023
out date: Wed Aug 02 02:00:00 GMT+02:00 2023
```

This fix will keep the correct day, month and year from the `inDate` value.

It should be save since the old workaround was to e.g. use `new Date(2023, 7, 20, 12, 0),`. But since the internal method removes hours anyways it shouldn't matter.

Still would need to have some users tests here so we know that it doesn't break existing apps that use a min/maxDate

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/PickerProxy.java
@@ -645,8 +645,14 @@ public class PickerProxy extends TiViewProxy implements PickerColumnProxy.OnChan
 		if (inDate == null) {
 			return null;
 		}
+
+		Calendar calIn = Calendar.getInstance();
+		calIn.setTime(inDate);
+
 		Calendar cal = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-		cal.setTime(inDate);
+		cal.set(Calendar.MONTH, calIn.get(Calendar.MONTH));
+		cal.set(Calendar.DAY_OF_MONTH, calIn.get(Calendar.DAY_OF_MONTH));
+		cal.set(Calendar.YEAR, calIn.get(Calendar.YEAR));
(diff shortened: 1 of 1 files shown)


hansemannn:
@janvennemann Can you take a look here, as you worked with time zones a lot recently?

janvennemann:
Ahhh, welcome to the wonderful world of JavaScript dates and its many quirks. It's important to note the difference between `new Date(2023, 7, 20)` and `new Date('2023-07-20')` here. The first one will create the date in the local timezone., the latter using UTC.

```js
new Date('2023-07-20')
// Thu Jul 20 2023 02:00:00 GMT+0200 (Central European Summer Time)
new Date(2023, 7, 20)
// Sun Aug 20 2023 00:00:00 GMT+0200 (Central European Summer Time)
```

The datepicker used to assume the date is in UTC. With your changes it will now first parse it using the local timezone. So, depending on what Date constructor someone uses, either one way or the other is correct on the native side. Technically this would be a breaking change since it assumes UTC on Android right now. And at least in our App we have a lot of extra handling that makes sure to pass a UTC based date to the date picker on Android and a local timezone based date on iOS.

m1ga:
interesting! Thanks for the input :+1: Didn't know that and it just works using `'2023-07-20'` :+1: No need for a change here.
Crazy that it doesn't use the 0 base month when you put it a string.

PR #13898, opened 2023-08-16
refactor: modernize Apple Script to open Xcode

hansemannn:
The Apple Script wasn't touched for forever, so I added some improvements:
- Support the new macOS system settings app that has different pane identifiers (this was broken)
- Clean up the dialog and split the large message into title and message to look more native
- Use the new macOS wordings for "System Settings" (new) vs "System Preferences" (old), etc.

Note: Github cannot properly display the diff as it's in binary form. Using the Apple Script app (built-in), the code can be observed.  

Before:

After:


cb1kenobi:
I don't have access to a mac right now. @m1ga can you test this? I think this script is invoked when you package an iOS app for the App Store and it opens the organizer in Xcode after the build.

m1ga:
@cb1kenobi I can test it later :+1:

m1ga:
* I've tested it without the permission so I can see the new dialog:
Screenshot_20230816_201953
* button opens the system settings
* if I enable it and build again it will open Xcode

Tested it on 12.6.7 (Monterey), Xcode 14.2

Is there anything else I need to test? It builds the app and opens XCdoe so it looks fine.

@cb1kenobi I'm running https://github.com/kholia/OSX-KVM on Linux :wink: Think that will run on Windows too

PR #13902, opened 2023-08-26
feat(android): add empty build.gradle

m1ga:
adding an empty build.gradle so people finally know where it needs to be :wink: 

Since it is needed on Android quiet often we should include it. And it will also create the folder for the push file


cb1kenobi (approved):

cb1kenobi:
I'm fine with this. Generally we have avoided adding the `platform` directory to the template as to not confuse, but if you think this will be less confusing, then so be it. Alternatively, we could add a section to the template's `README.md` on how users can create the `platform` folders.

m1ga:
> Alternatively, we could add a section to the template's README.md on how users can create the platform folders.

we have it in some modules readme/install instruction and people still ask for it :smile: 

Since we use gradle to build apps I've added some items to the [FAQ to fix module/build issues](https://titaniumsdk.com/guide/Titanium_SDK/Titanium_SDK_FAQ.html#common-gradle-error-messages-and-fixes) and people keep searching that file and don't know that they can just create it.

I've tested creating a classic and and converting it to Alloy. It will be moved to the correct place, so that will be fine.

cb1kenobi:
Ok, let’s try it. I wonder if we should have one for `ios`? Regardless, I’m cool with this.

PR #13899, opened 2023-08-16
fix(macos): fix layout issue

hansemannn:
Related to https://github.com/tidev/titanium-sdk/pull/13881

PR #13903, opened 2023-09-01
fix(ios): building for MacOS (Catalyst) now works with that fix

mbender74:
Building was not working because of not longer supported removeFromSuperView in MacOSCatalyst Updated Titanium.xcodeproj to correct destination settings for using same bundleId for iOS and MacOS

### DIFF
--- a/iphone/Classes/TiUIScrollableViewProxy.m
+++ b/iphone/Classes/TiUIScrollableViewProxy.m
@@ -6,8 +6,12 @@
  */
 #ifdef USE_TI_UISCROLLABLEVIEW
 
-#import "TiUIScrollableViewProxy.h"
+#if TARGET_OS_MACCATALYST
+#import <AppKit/AppKit.h>
+#endif
+
 #import "TiUIScrollableView.h"
+#import "TiUIScrollableViewProxy.h"
 
 @implementation TiUIScrollableViewProxy
 @synthesize viewProxies;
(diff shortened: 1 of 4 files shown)


iphone/cli/commands/_build.js
```diff
@@ -1899,7 +1909,7 @@ iOSBuilder.prototype.validate = function validate(logger, config, cli) {
 			// make sure they have Apple's WWDR cert installed
 			if (!this.iosInfo.certs.wwdr) {
 				logger.error(__('WWDR Intermediate Certificate not found') + '\n');
-				logger.log(__('Download and install the certificate from %s', 'https://www.apple.com/certificateauthority/AppleWWDRCAG3.cer'.cyan) + '\n');
+				logger.log(__('Download and install the certificate from %s', 'https://www.apple.com/certificateauthority/AppleWWDRCAG2.cer'.cyan) + '\n');
```
m1ga, line 1912: careful: we had people that complaint that 2 is not working and only 3 worked for them: https://github.com/tidev/titanium-sdk/pull/13841 We should keep 3 here

mbender74: ah, sorry, my mistake (when using older code) :-(

iphone/Classes/TiUIScrollableViewProxy.m
mbender74, line 1: I just found that #13899 from @hansemannn is a much smarter and better solution that has to be used!

hansemannn: You can definitely take it over - your other changes are great!

hansemannn:
Once the layout-changes from my other PR are cherry-picked and the iOS build works again, we can merge this and backport it for the 12.2.0.GA :-) No worries, I can do the backport once merged!

mbender74:
@hansemannn I take over #13899 and now all checks passed :-)

hansemannn (approved):

PR #13909, opened 2023-09-10
chore: copyright headers, wiki links

m1ga:
* Copyright headers (I think the info was to change them when we touch the file...but I just replaced all)
* wiki links
* lighthouse links


cb1kenobi (commented):
Can we rename all instances of `TiDev Titanium Mobile` and `Axway Titanium` to `Titanium SDK`? I don't think we need to say `TiDev Titanium SDK` when we say `TiDev` on the next line in the copyright.

m1ga:
Sure! Replace all of them :+1:

cb1kenobi (approved):
This PR is too big to reiew. My browser is just crawling.

PR #13913, opened 2023-09-20
fix(android): fix Android camera with overlay issues

m1ga:
Fixing two bugs:

**1. Unable to focus**

When `[WARN]  TiCameraActivity: (main) [10472,43572] Unable to focus.` is called the app is not informed that there was an issue while taking the photo and the app could get stuck.

To test it you can run the code below and make your camera not focus while pressing the button. You'll see the warning in the code and then `wait` is never set to false again (no success or error is called).

This PR will add a new error code `NO_FOCUS` and trigger the error callback.

---

**2. Crashlytics crash**

> Unable to resume activity {ti.modules.titanium.media.TiCameraActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'org.appcelerator.titanium.view.TiUIView org.appcelerator.titanium.proxy.TiViewProxy.getOrCreateView()' on a null object reference() at ti.modules.titanium.media.TiCameraActivity.onResume(TiCameraActivity.java:249)

Couldn't replicate it but adding `if (localOverlayProxy != null)` and `if (localView != null)` with an error log in the console to eliminate that error. It might not show an overlay but at least it won't crash.

----

**Example**
```js
function generateUI() {
	limit = 10;
	wait = false;
	num = 0;

	function closeCamera() {
		overlay.removeAllChildren();
		overlay = null;

		Ti.Media.hideCamera();
	}

	var scanner = Ti.UI.createView({ width: 200, height: 150,borderColor: "green", borderWidth: 5,	borderRadius: 15});
	var button = Ti.UI.createView({bottom: 40,width: 40,height: 40, backgroundColor: "green"});
	var done = Ti.UI.createLabel({color: "#fff",top: 15,right: 15, font: {fontSize: 20},text: "done"});

	var overlay = Ti.UI.createView();
	overlay.add(scanner);
	overlay.add(button);
	overlay.add(done);

	done.addEventListener('click', function() {
		closeCamera();
	});

	button.addEventListener('click', function() {
		if (wait == false) {
			wait = true;
			scanner.borderColor = "black";

			Ti.Media.takePicture();
		}
	});

	var error = function(error) {
		if (error.code == Ti.Media.NO_FOCUS) {
			wait = false;
			success();
			return;
		}
	};

	var cancel = function() {
		overlay.removeAllChildren();
		overlay = null;
	};

	var success = function(e) {
		console.log("callback")
		num++;
		if (num >= limit) {
			console.log("close camera")
			closeCamera();
		}
		
		scanner.borderColor = "green";
		wait = false;
	};

	Ti.Media.showCamera({
		saveToPhotoGallery: false,
		success: success,
		cancel: cancel,
		error: error,
		overlay: overlay,
		showControls: false,
		mediaTypes: [Ti.Media.MEDIA_TYPE_PHOTO],
		autohide: false
	});
}

	console.log("click")
	generateUI();
})
win.open();
```

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
@@ -93,7 +93,8 @@ public class MediaModule extends KrollModule implements Handler.Callback
 	public static final int NO_CAMERA = 2;
 	@Kroll.constant
 	public static final int NO_VIDEO = 3;
-
+	@Kroll.constant
+	public static final int NO_FOCUS = 4;
 	@Kroll.constant
 	public static final int IMAGE_SCALING_AUTO = -1;
 	@Kroll.constant
--- a/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
@@ -20,6 +20,7 @@ import org.appcelerator.titanium.TiBlob;
 import org.appcelerator.titanium.TiC;
 import org.appcelerator.titanium.io.TiContentFile;
 import org.appcelerator.titanium.proxy.TiViewProxy;
+import org.appcelerator.titanium.view.TiUIView;
 
 import android.Manifest;
 import android.app.Activity;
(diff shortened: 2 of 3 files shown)


hansemannn:
@m1ga Let me know if this is tested, then we can merge it.

m1ga:
Yes, client is using a custom SDK in the app already in a custom branch. `error.code == Ti.Media.NO_FOCUS` in the `error` event to handle the issue :+1:

hansemannn:
Is this code required now? It its a new API and old code without it crashes/breaks, it would be a breaking change and would be solved differently. Otherwise it's fine!

m1ga:
> Is this code required now?

yes, the client is using a build with this PR to have the app not being stuck when it doesn't focus. If you don't use this PR it will just stop/freeze as it will never reach the error or success call. 

* only affects users with camera + overlay and the camera returns `unfocused` while taking the picture
* then it is stuck as it doesn't go to success or error
* currently you have to close the app and restart the process
* existing apps that use the `error` event to catch other errors will use that (it is an error as it is not focused)
* after that PR you can differentiate in your error to show a message for unfocused or other errors now (e.g. in our case where you can take multiple photos at once)


The `null check` won't harm anyone. In the best case it will just work and never be `null`...if it somehow runs into a case where it is `null` it doesn't crash anymore (we couldn't reproduce it but have some Crashlytics errors)

hansemannn:
Okay, so using 12.2.1, users won't need to add additional code to get the same behavior and the new enum is used to improve accuracy in the error handling? We just need to make sure all old implementations work without changing something, otherwise the camera change is a breaking change and would need to be moved to SDK 13.

m1ga:
> Okay, so using 12.2.1, users won't need to add additional code to get the same behavior and the new enum is used to improve accuracy in the error handling

correct. If they use `error` already it will trigger that now instead of freezing.

hansemannn (approved):

PR #13914, opened 2023-09-20
fix(android): fix Android camera with overlay issues (12_2_X)

m1ga:
backport of https://github.com/tidev/titanium-sdk/pull/13913


hansemannn:
I will cherry-pick it directly to keep the upstream head clean.

PR #13923, opened 2023-10-01
[13922] fix(ios): request full access to calendar on ios17 to retain backward compatibility with ios16

narbs:
**fixes:** https://github.com/tidev/titanium-sdk/issues/13922

**Optional Description:**

To write to the calendar in iOS 17, the app now requires a new info.plist entry NSCalendarsFullAccessUsageDescription and needs to implement EKEventStore requestFullAccessToEventsWithCompletion.

Without this, building on iOS 17 using the existing requestAccessToEntityType:completion denies access to the calendar.

So this method needs to be implemented on iOS 17, replacing the iOS 16 approach, to restore existing functionality.


m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/13922

iphone/Classes/CalendarModule.m
```diff
@@ -222,21 +222,45 @@ - (void)requestAuthorization:(JSValue *)callback forEntityType:(EKEntityType)ent
...
-                                         [callback callWithArguments:@[ propertiesDict ]];
-                                       },
-                                       [NSThread isMainThread]);
-                                 }];
+
+        if (@available(iOS 17.0, *)) {
+
+          NSLog(@"[INFO] Calendar requestAuthorization iOS 17...");
+
+          [ourstore requestFullAccessToEventsWithCompletion:^(BOOL granted, NSError *error) {
```
hansemannn, line 230: Please remove the debug logs and empty lines

iphone/Classes/CalendarModule.m
```diff
@@ -222,21 +222,45 @@ - (void)requestAuthorization:(JSValue *)callback forEntityType:(EKEntityType)ent
...
+                                       NSString *errorMsg = granted ? nil : @"The user has denied access to events in Calendar.";
+                                       propertiesDict = [TiUtils dictionaryWithCode:(granted ? 0 : 1) message:errorMsg];
+                                     } else {
+                                       propertiesDict = [TiUtils dictionaryWithCode:[error code] message:[TiUtils messageFromError:error]];
+                                     }
+                                     TiThreadPerformOnMainThread(
+                                         ^{
+                                           [callback callWithArguments:@[ propertiesDict ]];
+                                         },
+                                         [NSThread isMainThread]);
```
hansemannn, line 261: It looks like the logic in the completion handler is reusable, so the block should be moved out of the handler and be called per instance. That way, the if-check if also better to read.

hansemannn (changes requested):
Mostly formatting and clean-code feedback :)

iphone/Classes/CalendarModule.m
narbs: Thank you for the feedback - I have made the changes to remove the logging and moved the logic for each block into a common area - please check

hansemannn (approved):

PR #13924, opened 2023-10-01
docs: updated examples

m1ga:
just a few that I found during browsing, there are still many more:

* removed some `background-color` attributes from windows (we have dark/light mode) as that make issues when you look at a white window in dark mode and the default text color is white
* removed the deprecated 2dmatrix/3dmatrix pages! They always come up in the search and aren't used since 8.0.0
* fixed the bindings.min.js example (it has to be the content of the file)
* removed a `known issue` from the tableview that has been fixed in 3.3.0
* added some windows to a few examples so they are standalone


ewanharris (changes requested):

cb1kenobi (approved):
LGTM

PR #13927, opened 2023-10-02
chore: removing old Appc strings

m1ga:
Found some Appcelerator strings:
* test file headers
* user agent Android
* in some example strings
* in the "do net edit" files

I'm not sure about the iOS user agent: https://github.com/search?q=repo%3Atidev%2Ftitanium-sdk%20kTitaniumUserAgentPrefix&type=code


cb1kenobi:
I think it's fine if we change the user agent this time as I don't think it's a huge impact, the updated user agent is technically correct, and user agents change all the time. So, I say ship it.

cb1kenobi (approved):

PR #13928, opened 2023-10-03
feat(android): more attributedString link features

m1ga:
Currently the attributedString link is not very flexible in terms of color and removing the underline.

This PR adds two new features:
* `linkColor` for the label: allows to set the color of the autoLink or attributedString link. <b>Note:</b> currently we can use another attribute `Ti.UI.ATTRIBUTE_FOREGROUND_COLOR` to color the link. This works for both Android and iOS. But there is not way to color a link with `autoLink` (Android only). So we won't need iOS parity for this property. 

* a new attribute `underline` in the attributedString that can be used in combination with `type: Titanium.UI.ATTRIBUTE_LINK` to remove the underline from the link

Screenshot_20231003-181037

```js
var win = Titanium.UI.createWindow({
	layout: "vertical"
});
win.open();
var text = 'just one link';

	text: text,
	attributes: [{
		type: Titanium.UI.ATTRIBUTE_LINK,
		value: "https://titaniumsdk.com",
		range: [text.indexOf('one'), ('one').length],
		underline: false
	}]
});

// custom link color - attributedString
var label = Titanium.UI.createLabel({
	height: Titanium.UI.SIZE,
	attributedString: attr,
	linkColor: "red"
});

// custom link color - autoLink
var label2 = Titanium.UI.createLabel({
	height: Titanium.UI.SIZE,
	text: " this is a https://titaniumsdk.com test",
	linkColor: "blue",
	autoLink: true
});

// current default link color
var label3 = Titanium.UI.createLabel({
	height: Titanium.UI.SIZE,
	text: " this is a https://titaniumsdk.com test",
	autoLink: true
});

win.add([label, label2, label3]);
```


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -484,6 +484,10 @@ public void processProperties(KrollDict d)
 			}
 		}
 
+		if (d.containsKey("linkColor")) {
```
hansemannn, line 487: There is an iOS-only attribute `ATTRIBUTE_UNDERLINE_COLOR` right now, the parity would be awesome here!

m1ga: https://github.com/tidev/titanium-sdk/blob/58f868b906c10e61fb90e0005938a23dacb5db05/android/modules/ui/src/java/ti/modules/titanium/ui/AttributedStringProxy.java#L264

Should already be there. The label part is for `autoLink` (Android only property) that will convert http links to clickable links

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -120,21 +120,28 @@ properties:
...
         This specifies the range of text to apply the property to, as an array of two numbers: `[from, length]`.
     type: Array<Number>
     optional: false
+  - name: underline
+    summary: Enable or disable the `ATTRIBUTE_LINK` underline.
+    description: |
+        Default value is `true`. If set to false and you use a `ATTRIBUTE_LINK` it won't underline the link.
+    type: Boolean
+    platforms: [android]
+    since: "12.3.0"
```
m1ga, line 134:
```suggestion
    since: "12.7.0"
```

apidoc/Titanium/UI/Label.yml
```diff
@@ -117,6 +117,15 @@ properties:
...
     type: [String, Titanium.UI.Color]
 
+  - name: linkColor
+    summary: Color of the link inside a label, as a color name or hex triplet.
+    description: |
+        If you enable `autoLink` on Android or add an AttributedString link it will use this color.
+        For information about color values, see the "Colors" section of <Titanium.UI>.
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: "12.3.0"
```
m1ga, line 127:
```suggestion
    since: "12.7.0"
```

apidoc/Titanium/UI/Attribute.yml
```diff
@@ -120,21 +120,28 @@ properties:
...
         This specifies the range of text to apply the property to, as an array of two numbers: `[from, length]`.
     type: Array<Number>
     optional: false
+  - name: underline
+    summary: Enable or disable the `ATTRIBUTE_LINK` underline.
+    description: |
+        Default value is `true`. If set to false and you use a `ATTRIBUTE_LINK` it won't underline the link.
+    type: Boolean
+    platforms: [android]
+    since: "12.7.0"
```
m1ga:
```suggestion
    since: "13.1.0"
```

apidoc/Titanium/UI/Label.yml
```diff
@@ -117,6 +117,15 @@ properties:
...
     type: [String, Titanium.UI.Color]
 
+  - name: linkColor
+    summary: Color of the link inside a label, as a color name or hex triplet.
+    description: |
+        If you enable `autoLink` on Android or add an AttributedString link it will use this color.
+        For information about color values, see the "Colors" section of <Titanium.UI>.
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: "12.7.0"
```
m1ga:
```suggestion
    since: "13.1.0"
```

PR #13938, opened 2023-10-24
feat(android): enable Signature Scheme v3

m1ga:
https://source.android.com/docs/security/features/apksigning/v3

> Android 9 supports [APK key rotation](https://developer.android.com/about/versions/pie/android-9.0#apk-key-rotation), which gives apps the ability to change their signing key as part of an APK update. To make rotation practical, APKs must indicate levels of trust between the new and old signing key. To support key rotation, we updated the [APK signature scheme](https://source.android.com/docs/security/features/apksigning/v2) from v2 to v3 to allow the new and old keys to be used. V3 adds information about the supported SDK versions and a proof-of-rotation struct to the APK signing block.

**Test**
* build a store APK
* run `[path to sdk]/build-tools/33.0.1/apksigner verify --verbose output.apk` and check for `Verified using v3 scheme (APK Signature Scheme v3): true`

## Note

I have no knowledge about the Signature Schemes and if we lose support for Android < 9 with this. It says it's the same block format as v2 so it might still work. Just putting this out there if needed and someone has more infos about it.


hansemannn:
Moving to `Draft` so it doesn't get accidentally merged. Losing < 9 compat would be pretty critical, although I think it should be pretty save. Once verified, we can merge this

m1ga:
Think we should be fine
> For maximum compatibility, sign applications with all schemes, first with v1, then v2, and then v3. Android 7.0+ and newer devices install apps signed with v2+ schemes more quickly than those signed only with v1 scheme. Older Android platforms ignore v2+ signatures and thus need apps to contain v1 signatures.

and
> Devices running Android 7.0 and later support APK signature scheme v2 (v2 scheme) and later. (v2 scheme was updated to v3 in Android 9 to include additional information in the signing block, but otherwise works the same.)

source: https://source.android.com/docs/security/features/apksigning

So adding v3 will still fall back to v2 or v1 if needed

m1ga:
just to test it I've uploaded an app with this PR to the store:
Screenshot_20240222_121337

no changes in the supported devices, so we should be good to go here

cb1kenobi:
This looks good to me, but I haven't tested it.

I believe we can set both `enableV3Signing` and `enableV4Signing`: https://stackoverflow.com/questions/59248088/how-can-i-sign-my-app-with-apk-signature-scheme-v3-and-v4. Is that something we should do?

cb1kenobi (approved):
Sounds good. Approving though I didn't test it. Merge when ready.

cb1kenobi (approved):

PR #13939, opened 2023-10-26
feat(ios): add cli.ignoreLog to ios

m1ga:
same as https://github.com/tidev/titanium-sdk/pull/13560 but for iOS.

New key: `cli.ignoreLog = []`
if one of the lines is in the log line it will be ignored.

Example:

```js
"cli": {
  "ignoreLog": [
    "Too many Flogger logs received before configuration",
    "Counters: exceeded sample count in FrameTime",
    "OpenGLRenderer: Davey!"
  ]
}
```

### DIFF
--- a/iphone/cli/hooks/run.js
+++ b/iphone/cli/hooks/run.js
@@ -75,10 +75,8 @@ exports.init = function (logger, config, cli) {
 				}
 
 				// ignore logs from cli ignoreLog
-				for (let i = 0, len = ignoreLog.length; i < len; ++i) {
-					if (line.includes(ignoreLog[i])) {
-						return;
-					}
+				if (ignoreLog.some(ignoreItem => line.includes(ignoreItem))) {
+					return;
 				}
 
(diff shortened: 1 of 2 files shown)


iphone/cli/hooks/run.js
```diff
@@ -72,6 +73,14 @@ exports.init = function (logger, config, cli) {
 					lastLogger = m[2].toLowerCase();
 					line = m[4].trim();
 				}
+
+				// ignore logs from cli ignoreLog
+				for (let i = 0, len = ignoreLog.length; i < len; ++i) {
+					if (line.includes(ignoreLog[i])) {
+						return;
+					}
+				}
```
hansemannn, line 82: A possibly cleaner approach:
```js
if (ignoreLog.some(ignoreItem => line.includes(ignoreItem))) {
    return;
}
```

m1ga: I like that, let's use that for the existing Android one too :+1:

hansemannn (approved):

PR #13940, opened 2023-10-27
chore(android): prepare SDK for Android 34

m1ga:
**Update**

According to the email from today (06/18) :
> As a reminder, from August 31, Google Play requires all new app releases to target Android 14.

it will be required for new apps in August.

The Google pages don't show that yet:
https://developer.android.com/google/play/requirements/target-sdk and https://support.google.com/googleplay/android-developer/answer/11926878?sjid=9819215519281928158-EU

<br/><br/>

**PR Changes**

There is only one change if you use a broadcast receiver in your app and you might see `one of RECEIVER_EXPORTED or  RECEIVER_NOT_EXPORTED should be specified when receiver isn't being registered exclusively for system broadcasts`

* keeping target api 33
* adding EXPORT part to broadcast receivers (https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported)

<br/><br/>

After this PR users can set `<uses-sdk android:targetSdkVersion="34"/>` in their tiapp.xml to target Android 34. It is not set by default inside the SDK.

<br/><br/>

**Test**
```js
var broadcastReceiver = Ti.Android.createBroadcastReceiver({
	onReceived: function(e) {}
});
Ti.Android.registerBroadcastReceiver(broadcastReceiver, ['com.appcelerator.action.ALERT']);
win.open();
```

tiapp.xml:
```
 <android>
        <manifest>
          <uses-sdk android:targetSdkVersion="34"/>
          <uses-permission android:name="com.appcelerator.action.ALERT"/>
        </manifest>
    </android>
```

Results: 
* 12.3.1.GA will show `One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts`
* this PR won't show the error


m1ga:
:warning:  Needs to be tested with different apps/users

Depending on the SDK release timing we can also do the following:

* merge the code changes and keep the targetSDK to 33
* users with new apps can change the targetSDK by hand in tiapp.xml
* users with updates can keep 33 as long as needed
* we have more time to test and get some field reports by users who want to go 34 already

android/gradle.properties
```diff
@@ -14,4 +14,4 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 # remove if com.android.tools.build:gradle can be raised in /android/build.gradle
-android.suppressUnsupportedCompileSdk=33
+android.suppressUnsupportedCompileSdk=33,34
```
cb1kenobi, line 17: Question: where is this property documented? I can't find anything on it. How did you guess it was a comma separated list?

m1ga: I was an error during the build like you see in this post: https://stackoverflow.com/questions/69911807/we-recommend-using-a-newer-android-gradle-plugin-to-use-compilesdkpreview-sv2

Once we merge the gradle update PR we can remove it

hansemannn (approved):

PR #13941, opened 2023-10-31
feat(ios): support multi-scene applications

hansemannn:
This PR adds the base compatibility for multi-scene apps, e.g. for iPadOS, CarPlay and visionOS. I may also add a PoC of defining scene-specific windows on the JS-side as well, but that's a whole different effort. Todos:
- [x] Verify newly generated app projects
- [x] Verify backwards compatibility with older apps -> Old apps will **break** right now. I will need to add a build check to inject the required scene configuration
- [x] Explore multi-scene JS API -> currently not worth it, as it's only used on iPadOS (CarPlay is also a single window). Once visionOS becomes more relevant in the community, I can look into that again
- [x] Add scene configuration via CLI hook to restore backwards compatibility


hansemannn:
@janvennemann ping

janvennemann (approved):
Looks good to me! As per discussion with Hans there are no additional backwards compatibility fixes required and existing app work perfectly fine with this change, so this is good to go.

PR #13948, opened 2023-11-16
build: bump compatibility versions of the iOS SDK and Node.js

hansemannn:
Note: I am not sure if this change will remove the Node.js warning or if the log originates from somewhere else:
```
[WARN]  Support for Node.js v18.16.0 has not been verified for Titanium SDK 12.2.1.GA
[WARN]  If you run into issues, try downgrading to Node.js v16
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "url": "git://github.com/tidev/titanium_mobile.git"
   },
   "vendorDependencies": {
-    "node": "12.x || 14.x || 16.x"
+    "node": "12.x || 14.x || 16.x || 18.x"
   },
   "engines": {
     "node": ">=12.13.0"
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "url": "git://github.com/tidev/titanium_mobile.git"
   },
   "vendorDependencies": {
-    "node": "12.x || 14.x || 16.x || 18.x"
+    "node": "12.x || 14.x || 16.x || 18.x || 20.x"
   },
   "engines": {
     "node": ">=12.13.0"
--- a/iphone/package.json
+++ b/iphone/package.json
@@ -15,8 +15,8 @@
 	"minIosVersion": "13.0",
 	"minWatchosVersion": "6.0",
 	"vendorDependencies": {
-		"xcode": ">=12.0 <=14.x",
-		"ios sdk": ">=13.0 <=16.x"
+		"xcode": ">=12.0 <=15.x",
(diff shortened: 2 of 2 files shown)


m1ga:
It is the correct place. The CLI will check for `vendorDependencies` here https://github.com/tidev/titanium-cli/blob/master/lib/titanium.js#L419

hansemannn:
It seems like Node.js 18 is in maintenance mode already since two weeks, so I am checking if we can go to 20 directly. iOS and Android app builds work fine so far.

package.json
```diff
@@ -166,7 +166,7 @@
     "url": "git://github.com/tidev/titanium_mobile.git"
   },
   "vendorDependencies": {
-    "node": "12.x || 14.x || 16.x"
+    "node": "12.x || 14.x || 16.x || 18.x || 20.x"
```
hansemannn, line 169: @cb1kenobi I wonder if we should remove everything pre-16, but in a SDK 13.0.0

m1ga: https://github.com/tidev/titanium-sdk/pull/13807

m1ga:
My Linux tests with **node 20**:
✔️ building Android SDK
✔️ building Android module
✔️ building own Android app
✔️ building kitchensink app
✔️ building hyperloop-examples app
✔️ building app with liveview (vite version) enabled

works fine too :+1:

package.json
hansemannn: Please, please, please do not rip out anything or break things for now. As nice as the CLI v7 is, it's far away from a stable / tested release and will require some major upgrades to run stable. Fixing the warning in SDK 12.x should be the most convenient way for most users to continue working without updating their Node.js setup, while developers that have the latest LTS tooling can continue working as well, but without warnings.

m1ga: +1 for removing it! 
But I think we should add up to 20 in the package.json for 12.x just to get rid of the warning as there are no issues with using it - keeping the minimum as it is for now. 
This makes it easier for new users to just grab the latest node version from the page and not have a warning right away

cb1kenobi: Sorry, yes, add 18 and 20 to SDK 12 to suppress the warning.

Definitely not suggesting we break anything anytime soon. Once v7 is merged and published as latest (it'll be a while), we can circle back and do a final v6 release that adds a guard around `vendorDependencies` and a message to update to v7. I want to avoid touching the CLI until v6 is in a maintenance branch and v7 is merged into `master`. The merge conflicts are going to be impossible to fix.

hansemannn:
Not accepted.

PR #13956, opened 2023-12-19
docs: add info about iOS foreground notifications

m1ga:
Apparently there is a `showInForground` in `userInfo`: https://github.com/tidev/titanium-sdk/blob/abdc7bf7dd1a0f2c7080f3c6fe4229d4b53dd776/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m#L547 but it was never documented.

Setting `userInfo: {showInForeground:true}` will display a local notification banner if the app is in foreground.


apidoc/Titanium/App/iOS/UserNotificationCenter.yml
```diff
@@ -160,6 +160,19 @@ properties:
...
+name: UserInfoDictionary
+summary: |
+    Dictionary object of parameters.
+platforms: [iphone, ipad, macos]
+
+properties:
+  - name: showInForground
+    summary: Show notification if app is in foreground
+    description: |
+        Only available in iOS >= 10. Boolean if a notification banner is shown if the app is in foreground.
```
hansemannn, line 173: These notes are not required anymore, as Titanium requires iOS 13+

apidoc/Titanium/App/iOS/UserNotificationCenter.yml
```diff
@@ -133,7 +133,7 @@ properties:
 
   - name: userInfo
     summary: Custom data object.
-    type: Dictionary
+    type: Dictionary<UserInfoDictionary>
```
hansemannn, line 136: Is this required? `UserInfoDictionary` should also work

hansemannn (changes requested):
Two quick ones

hansemannn (approved):

cb1kenobi (approved):
LGTM

PR #13957, opened 2023-12-26
fix(android): set correct title after titleAttributes update

m1ga:
Fixes an issue that came with https://github.com/tidev/titanium-sdk/pull/13954 and window titles. Title where not shown in all (sub)windows:

### Test 1:
```js
var win = Titanium.UI.createWindow({
	barColor: "pink",
	backgroundColor: "white",
	title: 'Autos',
	titleAttributes: {
		top: 0,
		color: "blue",
	},
});

	win.titleAttributes = {
		color: "yellow"
	}
})
win.open();
```
open and click

### Test 2:
```js
var win = Titanium.UI.createWindow({
	barColor: "pink",
	backgroundColor: "white",
	title: 'Autos',
	titleAttributes: {
		top: 0,
		color: "blue",
	},
});

var navWin = Titanium.UI.createNavigationWindow({
	window: win,
});

	win.titleAttributes = {
		color: "yellow"
	}
	win.title = "new"

	var w2 = Ti.UI.createWindow({
		barColor: "black",
		backgroundColor: "white",
		title: 'subwindow',
		titleAttributes: {
			top: 0,
			color: "green",
		},
	})
	navWin.openWindow(w2);
	setTimeout(function() {
		w2.title = "subwindow 2";
	}, 2000);
})
navWin.open();
```

open and click, wait, close window after title changed.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
@@ -533,8 +533,8 @@ public class WindowProxy extends TiWindowProxy implements TiActivityWindow
 						&& windowActivity.get().getSupportActionBar() != null) {
 						ActionBar actionBar = windowActivity.get().getSupportActionBar();
 						if (actionBar.getTitle() instanceof SpannableStringBuilder) {
-							SpannableStringBuilder stringBuilder;
-							stringBuilder = new SpannableStringBuilder(TiConvert.toString((Object) (msg.obj), ""));
+							SpannableStringBuilder stringBuilder =
+								new SpannableStringBuilder(TiConvert.toString((Object) (msg.obj), ""));
 							if (barColor != -1) {
 								stringBuilder.setSpan(new ForegroundColorSpan(barColor),
 									0, stringBuilder.length(), Spannable.SPAN_INCLUSIVE_INCLUSIVE);
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
@@ -81,6 +81,7 @@ public class WindowProxy extends TiWindowProxy implements TiActivityWindow
 	protected static final int MSG_LAST_ID = MSG_FIRST_ID + 999;
 
 	private static int id_toolbar;
+	private int barColor = -1;
 
 	private WeakReference<TiBaseActivity> windowActivity;
 
@@ -349,6 +350,7 @@ public class WindowProxy extends TiWindowProxy implements TiActivityWindow
 			ssb = new SpannableStringBuilder(abTitle);
 		}
 
+		barColor = colorInt;
 		ssb.setSpan(new ForegroundColorSpan(colorInt),
 			0, ssb.length(), Spannable.SPAN_INCLUSIVE_INCLUSIVE);
 		actionBar.setTitle(ssb);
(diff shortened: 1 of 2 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -526,6 +528,23 @@ public boolean handleMessage(Message msg)
...
 				if (activity != null) {
 					activity.setTitle(TiConvert.toString((Object) (msg.obj), ""));
+
+					if (windowActivity != null && windowActivity.get() != null
+						&& windowActivity.get().getSupportActionBar() != null) {
+						ActionBar actionBar = windowActivity.get().getSupportActionBar();
+						if (actionBar.getTitle() instanceof SpannableStringBuilder) {
+							SpannableStringBuilder ssb;
+							ssb = (SpannableStringBuilder) actionBar.getTitle();
+							ssb = new SpannableStringBuilder(TiConvert.toString((Object) (msg.obj), ""));
```
hansemannn, line 538: You are re-assigning / overriding it here directly again, is there a reason for that?

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
m1ga: you are right, is not needed. I've remove the first line

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -526,6 +528,22 @@ public boolean handleMessage(Message msg)
 				Activity activity = getWindowActivity();
 				if (activity != null) {
 					activity.setTitle(TiConvert.toString((Object) (msg.obj), ""));
+
+					if (windowActivity != null && windowActivity.get() != null
+						&& windowActivity.get().getSupportActionBar() != null) {
+						ActionBar actionBar = windowActivity.get().getSupportActionBar();
+						if (actionBar.getTitle() instanceof SpannableStringBuilder) {
+							SpannableStringBuilder stringBuilder;
+							stringBuilder = new SpannableStringBuilder(TiConvert.toString((Object) (msg.obj), ""));
```
hansemannn, line 537: Is there a reason why the constructor is split from the assignment?

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -526,6 +528,22 @@ public boolean handleMessage(Message msg)
...
+							if (barColor != -1) {
+								stringBuilder.setSpan(new ForegroundColorSpan(barColor),
+									0, stringBuilder.length(), Spannable.SPAN_INCLUSIVE_INCLUSIVE);
+							}
+							actionBar.setTitle(stringBuilder);
+						} else {
+							actionBar.setTitle(TiConvert.toString((Object) (msg.obj), ""));
```
hansemannn, line 544: Which type is `msg.object` by default?

hansemannn (commented):
2 quick ones left :)

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
m1ga: It should be `Object` already. I used the same lines as the current code:
https://github.com/tidev/titanium-sdk/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java#L528

Android Studio has a hint `Casting '(msg.obj)' to 'Object' is redundant` but I left it there just in case.

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
m1ga: you are right, can be one command (still two lines otherwise it will complain about the length :) )

hansemannn (approved):

PR #13959, opened 2024-01-03
fix(build/ios): simulator and macos builds don't require wwdr

ewanharris:
The current check will always step into the if because it will always not equal one of those, and the WWDR cert is not required for these.

To test, remove your WWDR cert from Keychain


cb1kenobi (approved):
This makes sense. Nice!

PR #13974, opened 2024-01-26
docs: link idleTimerDisabled and keepScreenOn

m1ga:
Link the two properties in the apidocs as it makes it easier to find it for parity


cb1kenobi (approved):
LGTM

PR #13975, opened 2024-01-26
feat(android): new event for empty TextFields

m1ga:
The `change` event on a TextField is only called when there is content in the field. If it is empty and you press "backspace" again you won't get any event. 

I've added an event: `empty` that will fire. 

```js
var tf1 = Ti.UI.createTextField({
	width: 150,
	left: 10
});
var tf2 = Ti.UI.createTextField({
	width: 150,
	right: 10,
	value: "remove me"
});
tf2.addEventListener("empty", function(e) {
	console.log("key", e.keyCode)
	tf1.focus();
})
tf2.addEventListener("change", function(e) {
	console.log("change", tf2.value.length)
})
win.add(tf1);
win.add(tf2);
	tf2.focus();
})
win.open();
```

https://github.com/tidev/titanium-sdk/assets/4334997/f10a081e-8810-4744-a685-d1d4f67d9d9e

TODO:
* [x] Docs
* [ ] rename it?
* [ ] someone needs to check if iOS has something similar


jordanbisato:
Isn't necessary to check if the TextField is empty before trigger the emptyField event?

m1ga:
as far as I can tell from testing it will only fire when the field is empty, so no text is in there and you press a key (See video above). You have to check the keyCode in Ti to see if it is the back key. I can add a check for the length but currently it looks like it is not necessary unless you have an example where it breaks

apidoc/Titanium/UI/TextField.yml
```diff
@@ -610,6 +610,15 @@ events:
         summary: New value of the field.
         type: String
 
+  - name: emptyField
```
hansemannn, line 613: Rename to `empty`?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -135,6 +135,20 @@ public void onLayoutChange(
 		this.tv.setOnEditorActionListener(this);
 		this.tv.setOnFocusChangeListener(this);
 		this.tv.setIncludeFontPadding(true);
+		this.tv.setOnKeyListener(new View.OnKeyListener()
```
hansemannn, line 138: Only add the listener if the event is listening for it

hansemannn (changes requested):

m1ga:
Fixed both comments

hansemannn (approved):

PR #13976, opened 2024-01-29
feat(android): parity for Tab.popToRootWindow()

m1ga:
An iOS TabGroup opens Windows inside of it and you can use `Tab.popToRootWindow()` to close all those tabs. Since Android is opening new windows we don't have that stack and can close them.

The idea is to emulate that behavior by looking at the first activity and check if it is a TabGroup. If so loop through all activity above it and close them.
Works with `style: Titanium.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION` too.

**Test:**

```js
	title: "open"
})
win.add(btn);

	createWindow();
})

function createWindow() {
	let newWin = Ti.UI.createWindow({
		backgroundColor: "#fff"
	});
	newWin.addEventListener("close", function(e){
		console.log("closing window");
	})
	tab1.open(newWin);
	let newButton = Ti.UI.createButton({
		title: "close",
		bottom: 100
	})
	let newButton2 = Ti.UI.createButton({
		title: "new"
	})
	newWin.add([newButton, newButton2]);
	newButton.addEventListener("click", function() {
		tab1.popToRootWindow();
	});
	newButton2.addEventListener("click", function() {
		createWindow()
	});
}

var tab1 = Ti.UI.createTab({
		window: win,
		title: 'a'
	}),
	tab2 = Ti.UI.createTab({
		window: Ti.UI.createWindow(),
		title: 'b'
	}),
		tabs: [tab1, tab2]
	});
tabGroup.open();

```

* start the app.
* open some windows and then press `close`
* you should go back to the TabGroup again

**Todo:**
* Test it :smile: Draft for now as I only tested it with the small demo above

https://github.com/tidev/titanium-sdk/assets/4334997/1cbb4282-be46-4bc4-ab0d-2808d8320623


hansemannn:
@m1ga Can you give me an update on a few of your `Draft` PR's? Many of them are super interesting and I wonder if you just missed to change the state after finishing work on them.

m1ga:
I just don't use this in any app so I don't have many test data to verify it doesn't have side effects. It's only still a draft as I don't have any infos besides this one demo that it is working :smile: It was requested on Slack from time to time so I've added this but no one tested it.

I'll change it to ready for review so we can get some more people to test it

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -1023,4 +1023,27 @@ public AccessibilityManager getAccessibilityManager()
...
 	{
 	}
+
+	public void popToRootWindow()
+	{
+		/**
+		 * emulates the iOS Tab.popToRootWindow() by closing all windows above a TabGroup.
+		 */
+		boolean isTabGroup = (activityStack.get(1).get() instanceof TiActivity)
+			&& ((TiActivity) activityStack.get(1).get())
```
hansemannn, line 1033: Hardcoding the index could potentially cause issues

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -1023,4 +1023,27 @@ public AccessibilityManager getAccessibilityManager()
...
+			.getWindowProxy().getApiName().equals("Ti.UI.TabGroup");
```
hansemannn, line 1034: Consider identifying tab groups differently, e.g via class instance checks

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -1023,4 +1023,27 @@ public AccessibilityManager getAccessibilityManager()
...
+		// no TabGroup - don't do anything
+		if (!isTabGroup) {
+			return;
+		}
+
+		// finish all activities above our TabGroup
+		for (int i = activityStack.size() - 1; i > 1; --i) {
+			if (activityStack.get(i).get() instanceof TiActivity) {
+				TiActivity currentActivity = (TiActivity) activityStack.get(i).get();
+				currentActivity.finish();
```
hansemannn, line 1045: We should try to close it via the Titanium proxy, so close events etc. are tracked correctly

hansemannn (changes requested):
A few comments to make it less error-prone

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
m1ga: You are right. Especially if once we merge https://github.com/tidev/titanium-sdk/pull/13889 it could be 0 instead of 1. I'll try to loop and check for the ID

apidoc/Titanium/UI/Tab.yml
```diff
@@ -130,19 +130,18 @@ methods:
...
         summary: |
-            Options supporting a single `animated` boolean property to determine whether the windows
+            iOS only: Options supporting a single `animated` boolean property to determine whether the windows
             will be animated while being closed (default: false).
         type: AnimatedOptions
         optional: true
         default: "{ animated: false }"
-    since: "6.2.0"
-    platforms: [iphone, ipad, macos]
+    since: {android: "12.4.0", iphone: "6.2.0", ipad: "6.2.0", macos: "9.2.0"}
```
hansemannn, line 143:
```suggestion
    since: {android: "12.7.0", iphone: "6.2.0", ipad: "6.2.0", macos: "9.2.0"}
```

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
```diff
@@ -1020,4 +1021,35 @@ public AccessibilityManager getAccessibilityManager()
...
+
+	public void popToRootWindow()
+	{
+		/**
+		 * emulates the iOS Tab.popToRootWindow() by closing all windows above a TabGroup.
+		 */
+		int tabGroupPosition = -1;
+		boolean isTabGroup = false;
+
+		for (int i = 0; i <= activityStack.size(); ++i) {
```
hansemannn, line 1033: Can you confirm that the "close" events are behaving the same as on iOS?

hansemannn (changes requested):

android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
m1ga: I would need to check iOS but it will fire all `close` events of each window if you call `popToRootWindow` in the example above. The log will show multiple `closing window` . But I'll run it on iOS to see if it is the same in a bit

m1ga: fixed the example so it will run on iOS too and yes, both show `closing window` (close even) for all opened windows.

hansemannn (approved):

PR #13977, opened 2024-01-29
fix: defer loading kroll-core to scene initialization

hansemannn:
It does show the debugger in Safari, let's see!

#### ⚠️  Please do not merge directly - I would clean up the PR for master first and then cherry-pick this change


hansemannn:
Superseded by #13979

PR #13979, opened 2024-01-30
fix(ios): fix debug issues with scenes

hansemannn:
The debugger now works, so I am adding the initial changes + the debug changes here.


ewanharris (approved):
LGTM! Debugging works in VS Code just fine (well just as fine as it ever has) 🚀

PR #13984, opened 2024-02-05
feat(android): add letterSpacing to Label

m1ga:
iOS can do letterspacing with a attributedString already. Android doesn't have this property. As it just a text property I've added it the same way as `lineSpacing` so it is a native Label property.

```js
	layout: "vertical"
});
var textLabel = Ti.UI.createLabel({
	text: "some text",
	letterSpacing: -0.1,
	font: {
		fontSize: 40
	}
});
var textLabel2 = Ti.UI.createLabel({
	text: "some text",
	letterSpacing: 0.1,
	font: {
		fontSize: 40
	}
});

win.add([textLabel, textLabel2]);
win.open();
```
Screenshot_20240205-091745


apidoc/Titanium/UI/Label.yml
```diff
@@ -181,6 +181,12 @@ properties:
     permission: read-only
     since: "12.3.0"
 
+  - name: letterSpacing
+    summary: Letter spacing of the [text](Titanium.UI.Label.text) as a float value.
+    platforms: [android]
+    type: Number
+    since: "12.4.0"
```
m1ga, line 188:
```suggestion
    since: "12.7.0"
```

hansemannn (approved):

PR #13991, opened 2024-02-14
ci: use newer artifact actions

ewanharris:
I was reading about the [deprecation of v1/v2 ](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/)of these and noticed the claim about v4 being stonkingly fast, so figured lets give it a look see. I doubt the time taken to upload things is _that_ slow here


cb1kenobi (approved):
SGTM

cb1kenobi:
Do we know how fast things are currently?

ewanharris:
Short of checking the timestamps not really no, maybe just an eye at the build time between this and the last build (which is a terrible metric).

Realistically, I should probably just bump all actions used in this PR too

ewanharris:
Eurgh looks like the changes for v4 of the third party delete-artifact require passing it a token which is kinda meh

m1ga:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20

I think you can also raise that to 20. SDK works fine with that version.

ewanharris:
@m1ga I think that's coming from the homebrew action which doesn't look to have updated to 20 yet based on their repo, I'll update where we set 16.x to 20.x though, good shout.

hansemannn:
Hi @ewanharris, should we still use this? :) Happy to do so!

ewanharris:
@hansemannn would probably be good to do but I remember facing an issue with this. If you're aiming to do a release this week I'd not worry about this PR

hansemannn:
Thanks a lot - yeah, we're doing a 12.5.0.GA today, but we can merge it into master/12.6.0 once all builds.

ewanharris:
@hansemannn this is ready now, I went ahead and updated all the actions, squashed the commits and also removed the `backport` action as it was using the not recommend `pull_request_target`

hansemannn (approved):

PR #13993, opened 2024-02-16
ci(android): install python before ccache

ewanharris:
Trying to fix failure here https://github.com/tidev/titanium-sdk/actions/runs/7927871631/job/21646099047, looks like ccache is failing to install python so maybe a manual install will work?


hansemannn (approved):

hansemannn:
We can merge once the build completes 👍

PR #13999, opened 2024-02-29
feat(android): more log output

m1ga:
current log length it 4059. This PR will split up long logs and outputs them in chunks.

**Test**
```
Ti.API.info("something long");
console.log("somenthing long");
```

Use https://www.blindtextgenerator.com/lorem-ipsum to generate a long string


android/runtime/v8/src/native/modules/APIModule.cpp
```diff
@@ -115,7 +115,19 @@ void APIModule::logInfo(const FunctionCallbackInfo<Value>& args)
...
+	std::string cppStr(*message);
+
+	int maxChunk = 4050;
+	int chunkCount = cppStr.length() / maxChunk;
+	for (int i = 0; i <= chunkCount; i++) {
+		int max = maxChunk * (i + 1);
+		if (max >= cppStr.length()) {
+			APIModule::logInternal(LOG_LEVEL_INFO, LCAT, cppStr.substr(maxChunk * i).c_str());
+		} else {
+			APIModule::logInternal(LOG_LEVEL_INFO, LCAT, cppStr.substr(maxChunk * i , max).c_str());
```
cb1kenobi, line 128: The second arg, `max` is should be the number of characters to include in the substring. Each iteration, `max` is growing and growing. I think you would be fine omitting the `if` and always passing in `maxChunk` for the second arg.

android/runtime/v8/src/native/modules/APIModule.cpp
```diff
@@ -115,7 +115,19 @@ void APIModule::logInfo(const FunctionCallbackInfo<Value>& args)
 	Isolate* isolate = args.GetIsolate();
 	HandleScope scope(isolate);
 	v8::String::Utf8Value message(isolate, APIModule::combineLogMessages(args));
-	APIModule::logInternal(LOG_LEVEL_INFO, LCAT, *message);
+
+	std::string cppStr(*message);
+
+	int maxChunk = 4050;
+	int chunkCount = cppStr.length() / maxChunk;
```
cb1kenobi, line 122: You are truncating the decimal when the result is cast to an int. You need to do a `ceil(cppStr.length() / maxChunk)`.

You could also completely do away with the `chunkCount` by incrementing `i` by `maxChunk` each iteration.

cb1kenobi (changes requested):

android/runtime/v8/src/native/modules/APIModule.cpp
m1ga: > I think you would be fine omitting the if and always passing in maxChunk for the second arg.

wouldn't that crash if the len is < maxChunk? Have to recheck. If that is not the case then yes, I can always use the 2nd `if` part

m1ga: you are right, not needed. Works fine with short strings:

> Number of characters to include in the substring **(if the string is shorter, as many characters as possible are used**).

android/runtime/v8/src/native/modules/APIModule.cpp
m1ga: done :+1: I've retested it with a short log, a 4050 character log and 4050+ logs. All are visible

android/runtime/v8/src/native/modules/APIModule.cpp
```diff
@@ -115,7 +115,13 @@ void APIModule::logInfo(const FunctionCallbackInfo<Value>& args)
 	Isolate* isolate = args.GetIsolate();
 	HandleScope scope(isolate);
 	v8::String::Utf8Value message(isolate, APIModule::combineLogMessages(args));
-	APIModule::logInternal(LOG_LEVEL_INFO, LCAT, *message);
+
+	std::string cppStr(*message);
+
+	int maxChunk = 4050;
+	for (int i = 0; i < cppStr.length(); i+=maxChunk) {
```
cb1kenobi, line 122: `i` is a signed `int` but `.length()` returns a `size_t` which is an `unsigned int`. There's a loss of data when the value is cast. To prevent the loss of data and speed up the loop by only evaluating the length once, move the length to the for loop initialization variables and use `size_t`.

```suggestion
	for (size_t i = 0, len = cppStr.length(); i < len; i+=maxChunk) {
```

cb1kenobi (changes requested):

android/runtime/v8/src/native/modules/APIModule.cpp
m1ga: runs fine, thanks :+1:

cb1kenobi (approved):
Looks good!

PR #14000, opened 2024-03-01
chore(android): try/catch around unlink snapshots

m1ga:
Same as the [unlink a few lines before](https://github.com/tidev/titanium-sdk/blob/4d793e48591f3d99844d5325811f72f006eb9151/android/titanium/libv8-services.js#L90-L94): adding a try/catch around in case the file is not there.

Happened a few time when I've switched branches and did a `cleanbuild` where it downloaded the v8 again.


m1ga:
@hansemannn can we merge this one please. Really bugs me when testing different branches :smile: The file is not there and it tries to remove it. It just adds a try/catch (as [some lines above it](https://github.com/tidev/titanium-sdk/blob/4d793e48591f3d99844d5325811f72f006eb9151/android/titanium/libv8-services.js#L90-L94)) so for normal builds it won't change anything. Have to do it by hand at the moment. 

Just a Linux SDK thing anyway since it is in `generateSnapshot` which only runs on Linux.

android/titanium/libv8-services.js
```diff
@@ -125,7 +125,11 @@ async function generateSnapshot(v8SnapshotHeaderFilePath, rollupFileContent) {
 		}
 
 		// Delete snapshot blob.
-		await fs.unlink(blobPath);
+		try {
+			await fs.unlink(blobPath);
+		} catch (e) {
+			// Do nothing...
+		}
```
cb1kenobi, line 132: I know there's a dozen ways to do this, but now I'm curious, what's the best way to do this?

As I get older, I don't mind calling things and catching the error like you did. It avoids race conditions. There was a stigma that try/catch was performance bottleneck, but that's a thing of the past and this is a CLI we're talking about, not some super hot loop.

Most of the SDK checks if the file exists before deleting like this:

```
		if (fs.existsSync(blobPath)) {
			await fs.unlink(blobPath);
		}
```

However, `fs` here is `fs-extra` which has a handy `fs.remove()` ([docs](https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove.md)) and it silently continues if the file does not exist:

```suggestion
		await fs.remove(blobPath);
```

m1ga: the last one sounds better. I've just copied the existing line as it was already there :smile: Let me check

cb1kenobi (approved):
Looks great!

cb1kenobi (approved):
Looks great!

android/titanium/libv8-services.js
m1ga: `await fs.remove(blobPath);` works too :+1: 
It's happening when I switch branches and do a `npm run cleanbuild:local`  when I was running a `npm run build:local` before (again: only linux)

PR #14004, opened 2024-03-07
fix(android): Ti.UI.Tab selected event returns no data

m1ga:
Currently the Ti.UI.Tab `selected` event doesn't return the values it should be according to the docs. The issue is that the data is set to `null` instead of `focusEventData`

**Test**
```js
var win1 = Ti.UI.createWindow();
win1.add(Ti.UI.createLabel({text: 'I am a blue window.'}));

win2.add(Ti.UI.createLabel({text: 'I am a red window.'}));

var tab1 = Ti.UI.createTab({
    window: win1,
    title: 'Blue'
}),
tab2 = Ti.UI.createTab({
    window: win2,
    title: 'Red'
}),
    tabs: [tab1, tab2]
});

tab1.addEventListener("selected", function(e){
  console.log(e.index, e.previousIndex)
})
tab2.addEventListener("selected", function(e){
  console.log(e.index, e.previousIndex)
})

tabGroup.open();
```


cb1kenobi (approved):
Tested and works great! Thanks! **APPROVED**

PR #14005, opened 2024-03-08
feat(android): new WebView property 'multipleWindows'

m1ga:
For 14 years the default value for `setSupportMultipleWindows()` is `true` which always tries to create new windows when clicking on a link with target _blank inside a webview. That breaks most webviews so you have to remove target="_blank" or change JS.

The new property `multipleWindows` will allow the user to just set it to `false`

**Test**
```js
var www = Ti.UI.createWebView({
  multipleWindows: false,
  html:"<div onclick=\"window.open('https://player.autopod.xyz/523959', '_blank', 'location=yes,height=810,width=470,scrollbars=no, resizable=no,status=yes');\">click me</div>"
});

win.add(www);
win.open();
```

In 12.3.0.GA or if you set it to `true` (default) it will just show a new black page. If you set it to `false` it will open the link in the same window/webview.

Will also fix the "Watch on YT" button if you have an embedded Youtube player in a webview. In the current SDK it will open a black tab when you click it and you can't go back or prevent it from loading. With this PR it will a) load the mobile YT page in the current tab (so no black screen) and b) you can use the `beforeload` event the prevent it from going to YT.


android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -293,10 +293,11 @@ public TiUIWebView(TiViewProxy proxy)
 		}
 		webView.setVerticalScrollbarOverlay(true);
 
+		boolean multipleWindows = TiConvert.toBoolean(proxy.getProperty(TiC.PROPERTY_MULTIPLE_WINDOWS), true);
 		WebSettings settings = webView.getSettings();
 		settings.setUseWideViewPort(true);
 		settings.setJavaScriptEnabled(true);
-		settings.setSupportMultipleWindows(true);
+		settings.setSupportMultipleWindows(multipleWindows);
```
prashantsaini1, line 300: @m1ga As mentioned in this PR's description, I can't think of any use case where one would need to open links in a new window inside the Ti WebView (it won't work anyways, right). So we should just set it to false forever. What do you think?

m1ga: just wanted to add as few breaking changes as I could, that's why the default is still `true`. We could swap the default to false as more people don't want to have it. But just have the option to still enable it as it was true for so many years now and some people might use it

prashantsaini1 (approved):

PR #14011, opened 2024-03-20
feat(android): use proguard files

m1ga:
Currently you can't use proguard.cfg files during the build process. I've enabled `this.proguard = true;` for the release build. If there is a `proguard.cfg` in the `app/platform/android/` folder it will set it to `proguardConfigFile` which is used in the gradle files later.

If there is no file or if you do a test/debug build it behaves the same as before. 

**Test:**
Download this demo project: [Archiv.zip](https://github.com/user-attachments/files/30224288/Archiv.zip)

**Build it:**
ti build -p android --target dist-playstore --keystore test.keystore --store-password tester123 --key-password tester123 --alias tester --output-dir dist

**Verify minification actually happened:**
1. Generated build.gradle should contain minifyEnabled + both cfg files
grep -A4 minifyEnabled build/android/app/build.gradle

2. R8 mapping file exists and shows renamed classes
head build/android/app/build/outputs/mapping/release/mapping.txt

3. Run it on a device — expect "ALL TESTS PASSED" on screen
adb install -r build/android/app/build/outputs/apk/release/app-release.apk
adb logcat -s TiAPI | grep PROGUARD-TEST

**Negative control (confirms the opt-in gating):** 
mv platform/android/proguard.cfg /tmp/, rebuild, and grep minifyEnabled build/android/app/build.gradle should come back empty.


m1ga:
Looks like http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/dists/focal/InRelease is currently not available...

PR #14008, opened 2024-03-19
feat(android): getter/setter for applicationLocales in Android 12 and up

AbdullahFaqeir:
feat(android): via app compat delegate set application locales.
feat(android): via app compat delegate get application locales.
feat(android): via app compat delegate get default night mode
feat(android): via app compat delegate check if compat vector from resources enabled
feat(android): via app compat delegate set compat vector from recourses enabled or not.

**Description:**

This PR was raised from an issue appeared on android level 33 and later which caused Ti.Locale.setLanguage() not to work properly, which caused random pages not to be translated unless you recall Ti.Locale.setLanguage() on each page.

See [here](https://developer.android.com/guide/topics/resources/app-languages#java)

Nonetheless, I've ported the rest of static methods from AppCompatDelegate.

_edited by Michael:_ for a test code check https://github.com/tidev/titanium-sdk/pull/14008#issuecomment-2009179503

PR #14013, opened 2024-03-23
feat(android): getter/setter for applicationLocales in Android 12 and up

AbdullahFaqeir:
Description:

This PR was raised from an issue appeared on android level 33 and later which caused Ti.Locale.setLanguage() not to work properly, which caused random pages not to be translated unless you recall Ti.Locale.setLanguage() on each page.

See [here](https://developer.android.com/guide/topics/resources/app-languages#java)

Replacing PR #14008 

edited by Michael: for a test code check https://github.com/tidev/titanium-sdk/pull/14008#issuecomment-2009179503


m1ga (changes requested):
Can you revert the changes to the comments in the LocaleModule, please. And the `text.startsWith...` change. Don't think that is needed for the applicationLocale.

Please keep the indent level in the doc file.

AbdullahFaqeir:
> Can you revert the changes to the comments in the LocaleModule, please. And the `text.startsWith...` change. Don't think that is needed for the applicationLocale.
> 
> Please keep the indent level in the doc file.

Should I go back to this?

AbdullahFaqeir:
@m1ga @AbdullahFaqeir can we check on this?

m1ga:
**Documentation:**
Can you change the indent level of the readme file back to the old amount of spaces so it will be inline with the other files and doesn't have that many changes.
And it would be good to say it is reading/setting the "Per-app language" instead of "UI language". Google calls it: https://developer.android.com/guide/topics/resources/app-languages 
Also add "returns an empty array if no per-app language is set yet". So people won't get confused why it is not returning anything (e.g. a default lang).

AbdullahFaqeir:
> **Documentation:** Can you change the indent level of the readme file back to the old amount of spaces so it will be inline with the other files and doesn't have that many changes. And it would be good to say it is reading/setting the "Per-app language" instead of "UI language". Google calls it: https://developer.android.com/guide/topics/resources/app-languages Also add "returns an empty array if no per-app language is set yet". So people won't get confused why it is not returning anything (e.g. a default lang).

@m1ga of course, I'll do it ASAP

AbdullahFaqeir:
@m1ga @hansemannn I believe this one is ready/

hansemannn:
There seem to be many changes which are not related to this PR, e.g. docs indentation and iOS xcode settings. Please split these up into separate PR's, so this one remains slim. Thank you!

m1ga:
And please add the API 33 check again https://developer.android.com/reference/android/app/LocaleManager#setApplicationLocales(android.os.LocaleList) was "Added in API level 33" and Ti still supports lower SDKs

PR #14020, opened 2024-04-15
fix(ios): fix some open issues related to scenes

hansemannn:
This PR should fix two of the three issues described in https://github.com/tidev/titanium-sdk/issues/13997 so far:
 
- 01fca4418ed1fa65d24e592a3fb9065dd5fddac8: fix “handleurl” event when using scenes
- 5d8d617a9124718f7cfb0e362632ba6ec68e5de4: fix log server from not being started on physical device

The `Ti.Network.registerForPushNotifications` issue still needs to be investigates. If someone has a test case, I'm happy to use it!

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
@@ -1198,11 +1198,19 @@ extern void UIColorFlushCache(void);
   [self tryToInvokeSelector:@selector(scene:willConnectToSession:options:)
               withArguments:[NSOrderedSet orderedSetWithObjects:scene, connectionOptions, nil]];
 
-  // If a "application-launch-url" is set, launch it directly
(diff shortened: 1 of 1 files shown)


m1ga:
any update on this one? 
If I remember from Slack you've fixed the registerForPush part too. Would be a good 12.3.1 version if we have those fixes together and could branch it out to get it tested

hansemannn:
I am still waiting for feedback from the reporters if the issues have been fixed. I think the push-related bug was not fixed, yet, after trying around for a day. I would recommend to review and merge this one first and then do the other one separately.

m1ga:
Sounds like a good idea. I'll review it over the weekend as I could reproduce both parts in an app

m1ga (approved):
Device log and handleurl are working fine with this PR. Demo code in https://github.com/tidev/titanium-sdk/issues/13997 :+1:

PR #14036, opened 2024-05-15
feat(android): improve accessibility text

m1ga:
By default all views/items will have say "double click to activate and for a long press double click and hold". Most items don't use a long press/click so it would be nice to deactive that message. This PR will add `accessibilityDisableLongPress` that you can use to disable that message.

**Test:**

* enable TalkBak
* run the code below
* click the label: old behavior - will say double click and long click
* click the button: new behavior - will only say "double click to activate"

```js
var win1 = Titanium.UI.createWindow({layout:"vertical"});
var text1 = Titanium.UI.createLabel({text: 'Hello Wold'});

  title: 'button',
  accessibilityHint: "click",
  accessibilityLabel: "button",
  accessibilityValue: "click me",
  accessibilityDisableLongPress: true
});

win1.add([text1, button]);
win1.open();
```


cb1kenobi (approved):
Looks good! **APPROVED!**

PR #14038, opened 2024-05-16
chore: fixing typos

m1ga:
Just some Android Studio spell checking :smile: 

Screenshot_20240516_133614

most of them are config/build files or variables but inside the normal files I've fixed the issues.
Mostly comments but some log message had some typos too.


cb1kenobi (approved):
Wow. Just wow. **APPROVED**

PR #14039, opened 2024-05-16
fix(android): remove some deprecated classes

m1ga:
Just two minor changes that are currently marked as deprecated:

* config part: https://developer.android.com/reference/android/util/Config#summary
* navigation part: https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/bottomnavigation/LabelVisibilityMode.java#L36-L38


android/runtime/common/src/java/org/appcelerator/kroll/common/TiConfig.java
```diff
@@ -24,9 +22,9 @@ public class TiConfig
 	* &lt;property name="ti.android.debug" type="bool"&gt;true&lt;/property&gt;
 	* </pre>
 	*/
-	public static boolean LOGD = Config.DEBUG;
-	public static boolean LOGV = Config.DEBUG;
-	public static boolean DEBUG = Config.DEBUG;
-	public static boolean RELEASE = !Config.DEBUG;
-	public static boolean PROFILE = Config.PROFILE;
+	public static boolean LOGD = false;
```
cb1kenobi, line 25: The docs say `LOGD` is always `true`. Are we sure we want `false` here? I have no idea what this is even. The Android docs are completely useless.

Just in case, here's the suggestion to expidite the change:

```suggestion
	public static boolean LOGD = true;
```

m1ga: _Config.LOGD_ in the Android Config is true, but here we used `Config.DEBUG` (which is always false) before. And this is the place where it is set:
```
TiConfig.DEBUG = TiConfig.LOGD = appProperties.getBool("ti.android.debug", false);
```
with `false` as default too. So I would keep it as false. 
It is added to every module `private static final boolean DBG = TiConfig.LOGD;` so in case people use that it should be the same value as before (false) unless you change it with the property.

cb1kenobi: Ohhh, that makes sense! Thanks!

cb1kenobi (approved):
Tested and reviewed. LGTM. **APPROVED!**

PR #14041, opened 2024-05-18
feat(ios): add overrideUserInterfaceStyle to Picker

m1ga:
The compact UI picker doesn't allow the user to set a text color. If you use your phone in light mode but have a dark background it will look like this:

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-18 at 15 52 51

This PR will allow you to switch the `overrideUserInterfaceStyle` for the Picker. I took most of the code/docs from the app `overrideUserInterfaceStyle` and modified it to use the picker instead of the whole app.

That way you can at least set the text color to white!

**Test**

```js
var win = Ti.UI.createWindow({backgroundColor:"black"});

  type:Ti.UI.DATE_PICKER_STYLE_COMPACT,
  minDate:new Date(2009,0,1),
  maxDate:new Date(2014,11,31),
  value:new Date(2014,3,12),
  backgroundColor:"transparent"
});
  title:"toggle", bottom: 40
});
var isLight = true;
btn.addEventListener("click", function(e){
  picker.overrideUserInterfaceStyle = isLight ? Titanium.UI.USER_INTERFACE_STYLE_DARK:Titanium.UI.USER_INTERFACE_STYLE_LIGHT;
  isLight = !isLight;
})
win.add([picker,btn]);
win.open();
```

* Run app in light mode
* click the toggle button to switch between the styles.

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-18 at 15 56 22

### DIFF
--- a/iphone/Classes/TiUIPicker.m
+++ b/iphone/Classes/TiUIPicker.m
@@ -225,6 +225,15 @@ USE_PROXY_FOR_VERIFY_AUTORESIZING
   }
 }
 
+- (void)setOverrideUserInterfaceStyle_:(id)args
+{
+  ENSURE_SINGLE_ARG(args, NSNumber);
+  if (picker != nil) {
+    int style = [TiUtils intValue:args def:UIUserInterfaceStyleUnspecified];
+    ((UIDatePicker *)[self picker]).overrideUserInterfaceStyle = style;
+  }
+}
+
 - (void)setDateTimeColor_:(id)value
 {
   // Guard date picker and iOS 14+ date picker style
--- a/apidoc/Titanium/UI/Picker.yml
+++ b/apidoc/Titanium/UI/Picker.yml
@@ -282,6 +282,21 @@ properties:
     since: "5.2.0"
     default: "black"
 
+  - name: overrideUserInterfaceStyle
+    summary: Forces the picker to used assigned theme instead of the system theme.
+    description: |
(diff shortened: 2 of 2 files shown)


iphone/Classes/TiUIPicker.m
```diff
@@ -225,6 +225,15 @@ - (void)setBackgroundColor_:(id)value
   }
 }
 
+- (void)setOverrideUserInterfaceStyle_:(id)args
+{
+  ENSURE_SINGLE_ARG(args, NSNumber);
+  if (picker != nil) {
+    int style = [TiUtils intValue:args def:UIUserInterfaceStyleUnspecified];
+    ((UIDatePicker *)[self picker]).overrideUserInterfaceStyle = style;
```
hansemannn, line 233: I am very curious, why this manual change is necessary, as it should inherit it from `UIControl` internally.

m1ga: because you can't change the text color by hand and if you have a light mode app but a dark background somewhere it will look like the image in the description. Another image/client app is in https://tidev.slack.com/archives/C03CVQX2A/p1716034991145569. Having the option to change the interfaceStyle just for that element will at least show the text again as it will be white.

hansemannn: Okay, thats an interesting edge case in the design spec. Then it's fine!

hansemannn (approved):

PR #14042, opened 2024-05-21
chore: raise android max sdk support

m1ga:
Currently `ti info` will show a warning for Android 14 emulators:
```
  !  Android API Android 14 (android-34) is too new and may or may not work with Titanium SDK 12.3.0.
     The maximum supported Android API level by Titanium SDK 12.3.0 is API level 33.

  !  Android API Android VanillaIceCream (Preview) (android-VanillaIceCream) is too new and may or may 
     not work with Titanium SDK 12.3.0.
     The maximum supported Android API level by Titanium SDK 12.3.0 is API level 33.
```

Titanium works fine as we still have the compile/target SDK set to 33!

This PR will raise the max version in the SDK check so it will get rid of the warning.

 ---

**Side note:** https://github.com/tidev/titanium-sdk/pull/13940 is for updating the targetSDK once we have to (not required or announced by Google at the moment)


cb1kenobi (approved):
Looks good! **APPROVED!**

PR #14046, opened 2024-05-25
fix: create alloy project with spaces with --alloy

m1ga:
If you run `ti create --name "Testing WebP Images" -p android -t app -u --id test -d . --alloy --sdk 12.3.0` to create an alloy project it will run `alloy new /Users/user/Developer/Apps/Testing WebP Images` later on and will break.

This PR will add `'path with space'` so you can uses spaces too.

**Test**
* run `ti create --name "Testing WebP Images" -p android -t app -u --id test -d . --alloy --sdk 12.3.0`


cli/commands/create.js
```diff
@@ -184,7 +184,7 @@ CreateCommand.prototype.run = function run(logger, config, cli, finished) {
 			}
 
 			if (cli.argv.alloy !== undefined) {
-				execSync('alloy new ' + cli.argv['workspace-dir'] + '/' + cli.argv.name);
+				execSync('alloy new \'' + path.join(cli.argv['workspace-dir'], cli.argv.name) + '\'');
```
cb1kenobi, line 187: This doesn't appear to work on Windows. Possibly because if single quotes instead of double quotes?

Replace the `execSync` import with:

```
{ spawnSync } = require('child_process'),
```

```suggestion
				spawnSync('alloy', ['new', path.join(cli.argv['workspace-dir'], cli.argv.name)], { stdio: 'inherit' });
```

cb1kenobi (changes requested):

cb1kenobi (approved):

PR #14047, opened 2024-05-25
fix(android): bundle webp res files

m1ga:
looks like there is a file that collects all image types in the build process and `webp` was not an option in there :) Downside: if you use `assets/android/images/res-xxxhdpi/` with a webp file it won't include that.

Using Android Studio you can see that it only includes res files for JPG images (or PNG):
Screenshot_20240525_143556
the webp file doesn't have dpi folders.

With this PR:
Screenshot_20240525_144917

**Tests**
I've also added a visual test with /res files. Currently only for a xxhdpi emulator. I had to downgrade strip-ansi again as that version was an ESM version and the tests wouldn't open. 6.0.1 is the last non ESM version for now


m1ga:
**Test**:

```js
	image: "/images/image.webp"
});
win.add(img);
win.open();
```

download this ZIP and extract it into /app/assets/
[android.zip](https://github.com/tidev/titanium-sdk/files/15443517/android.zip)
and run the app. It should show the image and e.g. XX or XXX in the image.

package.json
```diff
@@ -157,7 +157,7 @@
     "rollup": "2.76.0",
     "ssri": "10.0.4",
     "stream-splitter": "0.3.2",
-    "strip-ansi": "7.1.0",
+    "strip-ansi": "6.0.1",
```
hansemannn, line 160: Why did we need to downgrade here?

m1ga: it looks like strip-ansi is only used in `npm run test` the rest worked fine and the github action doesn't run the tests and I only tested apps and modules after updating the package.json the last time. 

I can remove the test for now so it will only be the one `webp` line and can add the test with the downgrade later

hansemannn (approved):

PR #14054, opened 2024-06-03
feat(android): track color of the Ti.UI.Switch

m1ga:
Parity for `tintColor` and adding `onTintColor` to color the tracks for the Material Switch.
Also adds `thumbColor` and `onThumbColor`.

```js

var basicSwitch = Ti.UI.createSwitch({
  tintColor: "red",
  onTintColor: "green",
  onThumbColor: "yellow",
  thumbColor: "blue"
});

// basicSwitch.tintColor = "blue"
win.add(basicSwitch);
win.open();

```

Screenshot_20240603-140140
Screenshot_20240603-140145

with the thumbColor:
Screenshot_20240604-142455

**Current workaround**
Using an [Android theme](https://m2.material.io/develop/android/components/switches#theming-switches)


hansemannn:
IIRC, we have similar theming capabilities on iOS already, are these the same properties?

m1ga:
ah, good point. `onTintColor` is the color for "on". I'll rename the `activeTintColor`

m1ga:
Update: 
* renamed `activeTintColor` to `onTintColor` for iOS parity.
* renamed `activeThumbColor` to `onThubColor` to match the other property

apidoc/Titanium/UI/Switch.yml
```diff
@@ -118,17 +118,43 @@ properties:
...
+
+        Android: Active thumb color of the Material Switch.
+    type: [String, Titanium.UI.Color]
+    default: undefined
+    platforms: [android]
+    since: {android: 12.4.0}
+
+  - name: thumbColor
+    summary: The color used to tint the thumb icon of the switch when it is turned off.
+    description: |
```
hansemannn, line 143: If no specific description is provided, you can rather onit the field to prevent duplicate content

m1ga: I removed the summery so it will just use the description as that had more info

PR #14066, opened 2024-06-18
feat(android): optionBar color properties

m1ga:
If you want to change the colors in an OptionBar you'll have to use a theme. This PR adds new properties `selectectedBackgroundColor`, `selectedBorderColor`, `selectedTextColor` and maps the `color` to the inactive text colors.

Screenshot_20240618-120614

```js

const optionBarDefault = Ti.UI.createOptionBar({
	top: 0,
  labels: [ 'Option 1', 'Option 2', 'Option 3' ]
});
const optionBar1 = Ti.UI.createOptionBar({
	top: 50,
	selectedBackgroundColor: "red",
	selectedBorderColor: "#fff",
	selectedTextColor: "blue",
	color: "yellow",
  labels: [ 'Option 1', 'Option 2', 'Option 3' ]
});
const optionBar2 = Ti.UI.createOptionBar({
	top: 100,
	color: "yellow",
  labels: [ 'Option 1', 'Option 2', 'Option 3' ]
});
const optionBar3 = Ti.UI.createOptionBar({
	top: 150,
	selectedTextColor: "blue",
  labels: [ 'Option 1', 'Option 2', 'Option 3' ]
});

win.add([optionBarDefault,optionBar1,optionBar2,optionBar3]);

win.open();
```

### DIFF
--- a/apidoc/Titanium/UI/OptionBar.yml
+++ b/apidoc/Titanium/UI/OptionBar.yml
@@ -60,6 +60,30 @@ properties:
     default: horizontal
     availability: creation
 
+  - name: selectedBackgroundColor
+    summary: Background color of the selected button
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: { android: "12.4.0"}
+
+  - name: selectedTextColor
+    summary: Text color of the selected button
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: { android: "12.4.0"}
+
+  - name: selectedBorderColor
+    summary: Border color of the selected button
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: { android: "12.4.0"}
+
+  - name: color
+    summary: Text color of the unselected button
+    type: [String, Titanium.UI.Color]
+    platforms: [android]
+    since: { android: "12.4.0"}
+
 examples:
   - title: Text-Only Buttons
     example: |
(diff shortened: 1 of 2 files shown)


apidoc/Titanium/UI/OptionBar.yml
```diff
@@ -60,6 +60,30 @@ properties:
     default: horizontal
     availability: creation
 
+  - name: selectedBackgroundColor
```
hansemannn, line 63: I feel like these properties (partially) are available on iOS as well, but under a different naming

hansemannn: EDIT: Yep, there is:
- tintColor (for the selected background item)
- backgroundColor (for the global background of the option bar, could be seen as unselected background color)
- textColor (unselected text color)
- selectedTextColor (...)
- selectedButtonColor (same as tintColor, but an older API)

m1ga: Are all theses available in an OptionBar? I only see `selectedButtonColor` in a ButtonBar. 
I'll move from `selectedBackgroundColor` -> `selectedButtonColor` and `textColor` -> `color`. The rest looks the same.

hansemannn: They are technically exposed, but maybe not documented? If it's not intended to be documented for this component (they both use the UISegmentedControl internally), we can keep this one as-is.

hansemannn (approved):

PR #14068, opened 2024-06-19
feat(android): set targetSDK to Android 34

m1ga:
Merge this after https://github.com/tidev/titanium-sdk/pull/13940 e.g. in combination with iOS 18 changes.

Users with **new apps** in August can target 34 with `<uses-sdk android:targetSdkVersion="34"/>` and the PR above for the code changes.


efeder-amd:
Do we know when this will get merged in and added to a titanium sdk version? Until then, does that mean we can't set targetSdkVersion for our apps to 34 if using one of the current 12.4.x titanium sdk versions?

m1ga:
You can set `<uses-sdk android:targetSdkVersion="34"/>` in any Ti SDK already. Just put it in your tiapp.xml
Check the last blog post: https://tidev.io/blog/sdk_12_4_0_ga

IF you use any BroadcastReceiver in your app I would update to 12.4.0.GA as it includes a fix we had to do for Android 34. Otherwise you can use any other Ti SDK and add the line above

hansemannn (approved):

PR #14074, opened 2024-07-02
feat(android): return and store exif data in blob/file

m1ga:
Two new features:
* get `exif` data in `event.media.exif` from the camera media file
* it will also be store into a file when you run `imageAsCompressed` and write that blob data

```js

	Ti.Media.showCamera({
		success: function(event) {
			console.log(event.media.exif)

			var f = Titanium.Filesystem.getFile(Titanium.Filesystem.applicationDataDirectory + "/image.jpg");
			f.write(event.media);

			var f1 = Titanium.Filesystem.getFile(Titanium.Filesystem.applicationDataDirectory + "/image1.jpg");
			f1.write(event.media.imageAsCompressed(0.9));

			var f1 = Titanium.Filesystem.getFile(Titanium.Filesystem.applicationDataDirectory + "/image2.jpg");
			f1.write(event.media.imageAsCropped({
				width: 10,
				height: 10,
				x: 0,
				y: 0
			}));

			var f1 = Titanium.Filesystem.getFile(Titanium.Filesystem.applicationDataDirectory + "/image3.jpg");
			f1.write(event.media.imageAsResized(20, 20));

		}
	});
})

win.open();
```

**Test**
To copy files quickly you can use `adb exec-out run-as [your bundle id] cat app_appdata/image.jpg > image.jpg`

12.3.1:
* run the above code in Ti 12.3.1 and download the two files from your phone
* image1.jpg won't have any exif data

With this PR:
* image1.jpg, image2.jpg and image3.jpg will have exif data
* you will see a console.log with the exif data from the camera image


apidoc/Titanium/Blob.yml
```diff
@@ -149,6 +149,13 @@ properties:
     since: {android: "12.3.0"}
     permission: read-only
 
+  - name: exif
+    type: Object
+    summary: In the camera `success` event you will get the EXIF information from the media blob if available.
+    platforms: [android]
+    since: {android: "12.7.0"}
```
m1ga:
```suggestion
    since: {android: "13.2.0"}
```

.github/workflows/docs.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         with:
             fetch-depth: 0
       - name: Use Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
```
prashantsaini1, line 20: Is this version supported and compatible with the repo?

.github/workflows/regen-docs.yml
```diff
@@ -21,9 +21,9 @@ jobs:
         fetch-depth: 0
 
     - name: Setup node
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6
       with:
-        node-version: '16'
+        node-version: '24'
```
prashantsaini1, line 26: compatible and supported?

android/titanium/src/java/org/appcelerator/titanium/util/TiExifHelper.java
```diff
@@ -0,0 +1,31 @@
+package org.appcelerator.titanium.util;
+
+import androidx.exifinterface.media.ExifInterface;
+
+public class TiExifHelper
+{
+	public static String[] attributes = new String[]
```
prashantsaini1: As far as I know, there are ~100 of exif tags. In general, you can refer the ExifInterface class and pick every tag [mentioned here](https://developer.android.com/reference/androidx/exifinterface/media/ExifInterface).

android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
```diff
@@ -832,7 +832,7 @@ public void onPictureTaken(byte[] data, Camera camera)
 						previewRect.put(TiC.PROPERTY_HEIGHT, 0);
 					}
 					response.put("previewRect", previewRect);
-
+					response.put("exif", blob.getExif());
```
prashantsaini1, line 848: As discussed on Slack, better to have a new property `preserveExif` which avoids reading it for existing versions as this can be a breaking change for apps not needing the exif data, yet.

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
```diff
@@ -291,7 +297,17 @@ public boolean write(Object[] args)
 				}
 
 				if (args[0] instanceof TiBlob) {
+					ContentResolver contentResolver = TiApplication.getInstance().getContentResolver();
```
prashantsaini1: - Same as the previous comment on having `preserveExif` since the `saveAttributes()` call is expensive.
- Also note that writing exif is only supported in JPEG, PNG and WebP formats [as described here](https://developer.android.com/reference/androidx/exifinterface/media/ExifInterface#saveAttributes()).

.github/workflows/regen-docs.yml
m1ga: yes, I've updated iOS lib https://github.com/tidev/titanium-sdk/pull/14377 so official since 13.1.1 ([see matrix](https://titaniumsdk.com/guide/Titanium_SDK/Titanium_SDK_Getting_Started/Installation_and_Configuration/Titanium_Compatibility_Matrix/#supported-version-of-node-js)) and Android is supported for a while now

android/titanium/src/java/org/appcelerator/titanium/TiFileProxy.java
m1ga: Two fixes made by Claude:

* `saveAttributes` will do the supported file check itself - no fix here
* TiBlob now tracks isExifInherited, set only by inheritExifFrom(). That distinguishes the two ways a blob can have EXIF:
  - Read from its own bytes (loadExifData()) — a camera capture, a file read, an HTTP response. The metadata is physically in the data, so tbf.write() copies it to the destination for free.
  - Inherited in memory (inheritExifFrom()) — the output of imageAsResized/imageAsCropped/imageAsCompressed/etc., re-encoded from a Bitmap with no metadata in its bytes. This is the only case that actually needs saveAttributes().

  copyExifData() now bails on isExifWriteRequired() first. So f.write(event.media) — the most common path, and previously the most expensive — does no EXIF work at all: no temp-file copy, no image rebuild.

PR #14076, opened 2024-07-06
fix(android): add missing “Calendar.Event.remove” method

hansemannn:
Fixes an issue where calendar events could not be deleted programmatically on Android, yet.

### DIFF
--- a/android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
+++ b/android/modules/calendar/src/java/ti/modules/titanium/calendar/EventProxy.java
@@ -559,6 +559,21 @@ public class EventProxy extends KrollProxy
 		contentResolver.insert(extPropsUri, values);
 	}
 
+	@Kroll.method
+	public boolean remove(Object eventId)
+	{
+		ContentResolver contentResolver = TiApplication.getInstance().getContentResolver();
+
+		try {
+			Uri deleteUri = ContentUris.withAppendedId(CalendarContract.Events.CONTENT_URI, TiConvert.toInt(eventId));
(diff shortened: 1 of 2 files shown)


m1ga:
I've created some test code:
```js
	layout: "vertical"
});
var btn1 = Ti.UI.createButton({
	title: "add"
});
var btn2 = Ti.UI.createButton({
	title: "delete"
});
win.add([btn1, btn2]);
win.open();
var event = null;
	var calendar = Ti.Calendar.defaultCalendar;

	// Create the event
	var eventBegins = new Date(2024, 8, 1, 12, 0, 0);
	var eventEnds = new Date(2024, 8, 1, 14, 0, 0);
	var details = {
		title: 'Do some stuff',
		description: "I'm going to do some stuff at this time.",
		begin: eventBegins,
		end: eventEnds
	};

	if (calendar == null) {
		console.log("calendar is empty")
		return;
	}
	event = calendar.createEvent(details);

	console.log("done", event.id);
})

	if (event) {
		console.log(event.remove(event.id));
	}
});
```

It works fine. Looks like the calendar is caching the events for some seconds so in case you test it in split-screen mode it doesn't update the calendar right away when you add/remove an event.

**Small note:**
can you add something like `Android: parameter is the event id you want to remove` to the docs, as it is not clear what the parameter should be because you call it with `event.remove()` already and the `span` is something different on iOS. BTW: it doesn't link to constants

hansemannn:
Ohh, good catch. We don't even need to pass the ID, I'll remove that!

m1ga (approved):
LGTM. Working fine now without a parameter :+1:

PR #14078, opened 2024-07-09
fix(win): permission issues when building a module

### LINKED ISSUE #14077: Windows: building modules will fail with permission error

### Description

When trying to build an Android module on windows it will show 
```
[ERROR] EPERM: operation not permitted, lstat 'C:\ProjectPath\module\android\build\build_android.log'.
```

which is located here: https://github.com/tidev/titanium-sdk/blob/master/cli/commands/build.js#L424

When I comment out that line it will build the module

### Expected Behavior

build the module

### Actual behavior

crash with permission error

### Reproducible sample

```
ti create -t module --id com.test -n test --android-code-base java -p android -d . && cd test/android && ti build -p android -b
```

### Steps to reproduce

run the command above

### Platform

Android

### SDK version you are using

12.3.1

_No response_

m1ga:
Fixes https://github.com/tidev/titanium-sdk/issues/14077

```
ti create -t module --id com.test -n test --android-code-base java -p android -d . && cd test/android && ti build -p android -b
```

will disable logging to file when you build a module on Windows. 
BTW: on Linux I don't even have the build_android.log file but it won't show a permission error.

### DIFF
--- a/cli/commands/build.js
+++ b/cli/commands/build.js
@@ -420,8 +420,13 @@ function patchLogger(logger, cli) {
 
 		fs.ensureDirSync(buildDir, 0o766);
 
-		// create our write stream
-		logger.log.filestream = fs.createWriteStream(path.join(buildDir, 'build_' + platform + '.log'), { flags: 'w', encoding: 'utf8', mode: 0o666 });
+		if (process.platform === 'win32' && cli.argv.type === 'module') {
+			// disable file logging for windows when building a module
+			logger.fileWriteEnabled = false;
+		} else {
+			// create our write stream
+			logger.log.filestream = fs.createWriteStream(path.join(buildDir, 'build_' + platform + '.log'), { flags: 'w', encoding: 'utf8', mode: 0o666 });
+		}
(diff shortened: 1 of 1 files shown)


cb1kenobi:
As mentioned in the [associated issue](https://github.com/tidev/titanium-sdk/issues/14077#issuecomment-2218810795), I don't think this is the best solution.

hansemannn:
@m1ga I've noticed that you've marked this PR as ready-for-review. Looking at @cb1kenobi's comment, there may be additional work needed?

m1ga:
correct, I've explained it in the last comment: https://github.com/tidev/titanium-sdk/issues/14077#issuecomment-2225442113 
This fix will just be an intermediate fix to make module building on windows work again. It's only for Windows users that are trying to build modules. We had one report on Slack so far, so I don't think it is a common platform and people are missing the build log file because it wasn't there with CLI v6 either.

m1ga:
closed for a proper solution to get the log output instead of skipping it.

PR #14086, opened 2024-07-18
feat(android): add maxImages and pathOnly to openPhotoGallery

m1ga:
* adding `maxImages`: max amount of images to select. Use in combination with `allowMultiple`
* adding `pathOnly`: won't include the blob in the returned data
* add the `path` to the returned object

Note: `EXTRA_PICK_IMAGES_MAX` only works with `new Intent(MediaStore.ACTION_PICK_IMAGES)`  so I have to recreate `galleryIntent = new TiIntentWrapper(new Intent(MediaStore.ACTION_PICK_IMAGES));`

**Test:**

```js
var img = Ti.UI.createImageView();
win.add(img);
win.open();

	Ti.Media.openPhotoGallery({
		mediaTypes: [Titanium.Media.MEDIA_TYPE_PHOTO],
		allowMultiple: true,    // test it with and without
		maxImages: 3,
		pathOnly: true,
		success: function(e) {
			if (e.images) {
				for (var i = 0; i < e.images.length; ++i) {
					// no blob data
					console.log(e.images[i])
				}
				img.image = e.images[0].path;
			} else {
				// no blob data
				console.log(e)
				img.image = e.path;
			}
		},
		error: function(e) {
			alert('error opening image: ' + e);
		}
	});
})

```

* click the window and try to select more than 3 images
* change `allowMultiple` to false and test `pathOnly` values


m1ga:
@hansemannn : I'm still testing it a bit but it will just give you:
```
{
 cropRect: { x: 0, width: 1080, y: 0, height: 2400 },
 path: 'file:///sdcard/.transforms/synthetic/picker_get_content/0/com.android.providers.media.photopicker/media/1000043692.png',
 code: 0,
 success: true,
 x: 0,
 width: 1080,
 y: 0,
 mediaType: 'public.image',
 height: 2400
}

```
as a result if you set `pathOnly: true`. Can be used in combination with or without `allowMultiple` and `maxImages`. So theoretically you can replace the external image picker module.

Also added the multi image picker to Kitchensink: https://github.com/tidev/kitchensink-v2/pull/69 to test it.

hansemannn (approved):
Great idea! Was it ready to review or were there issues during testing?

m1ga:
Yes, ready for review! I've tested it in an existing app without the new properties and that works as before. So no issues for existing apps and you can enable `pathOnly: true` to get the new features (no blobs)

PR #14089, opened 2024-07-27
feat(android): statusBarColor for Window

m1ga:
**Android:**
Add `statusBarColor` property to change the statusBar color using code
```js
	statusBarColor:"#0f0"
});
setTimeout(function(){
	win.statusBarColor = "#f00";
},2000)

win.open();
```

**Test:**
run the code: status bar is green and after one second it's red

**Current workaround**
either use a normal Android theme and set it there or Hyperloop
```js
import Activity from 'android.app.Activity';
import Color from 'android.graphics.Color';

win.addEventListener("open", function(){
	const activity = new Activity(win.activity);
	setInterval(function(){
		activity.getWindow().setStatusBarColor(Color.argb(Math.floor(Math.random()*255), Math.floor(Math.random()*255), Math.floor(Math.random()*255), 0));
	}, 500)
})
win.open();
```


cb1kenobi (approved):
Tested and works!

PR #14094, opened 2024-08-08
feat(android): fire `selected` event again when clicking the same Tab again

m1ga:
If you reclick a tab it won't fire the event again. This PR will fire the `selected` event again if you click the tab again

Can be used to reset/reload content in the tab if you click it again.

```js
const tab1 = Ti.UI.createTab({
	window: Ti.UI.createWindow(),
	title: 'Tab 1'
});
const tab2 = Ti.UI.createTab({
	window: Ti.UI.createWindow(),
	title: 'Tab 2'
});

tab1.addEventListener("selected", function(opts) {
	console.log("selected", opts.index)
})
tab2.addEventListener("selected", function(opts) {
	console.log("selected", opts.index)
})

	tabs: [tab1, tab2]
});

tabGroup.open();

```

I didn't check yes how iOS `selected` behaves. Is that a one time event when you reclick the same tab?


cb1kenobi:
This new event feels like bloat. I would think a "click" event with a current tab that you can check would suffice.

m1ga:
This would work too (I would reuse the `selected` event that is there) but then apps that currently use that event will get multiple events now. Even if I add a parameter `reselected: true` in the KrollDict the user has to use that in an existing app.

That was my only concern to move it to a different event. I'm more than happy with having only the `selected`. In my case I wanted to reset a filter when clicking on the tab again and was surprised that the event didn't fire again

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUITabLayoutTabGroup.java
```diff
@@ -411,6 +412,17 @@ public void onTabUnselected(TabLayout.Tab tab)
...
 	public void onTabReselected(TabLayout.Tab tab)
 	{
+		if (tab != null) {
+			int index = tab.getPosition();
+			if ((index >= 0) && (index < this.tabs.size())) {
+				TiViewProxy tabProxy = this.tabs.get(index).getProxy();
+				if (tabProxy != null) {
+					KrollDict data = new KrollDict();
+					data.put("index", index);
+					tabProxy.fireEvent(TiC.EVENT_RESELECTED, data, false);
```
hansemannn, line 422: Guard the listener first to improve performance, e.g. "hasListener"?

cb1kenobi:
Still not sure on this one. Feels unnessary. I don't get why we have a "selected" and "unselected" events either. On the web, we would have "click" and "change". It feels like there's too many events.

hansemannn:
I kind of agree, although the code itself is fine. What was the reason for adding this new API? Was someone unable to achieve this behavior with the current events?

cb1kenobi:
I would be fine with firing `selected` when a tab is selected twice.

I would have to see how the UX is, but I could see how fat fingering a tab and reseting the state may be a bad UX pattern. I would think a "reset" button would be best, but I understand each app is different.

m1ga:
e.g. in Slack, Whatsapp or Instagram it will scroll up to the first item if you click the Tab icon again which is a very handy use case. 
Maybe I could add a property to `selected` so the user knows if it is a first click or if you reselected it. On the other hand you can do that already by tracking the current selected index if you'll need that and compare it in the event

m1ga:
updated the example and code so it will just simply fire the `selected` event again :+1:

PR #14095, opened 2024-08-15
fix: sdk build on windows needs shell: true to run batch files

cb1kenobi:
Not sure if it's a Node 22 thing or what, but I get a `EINVAL` error doing `npm run cleanbuild:local` because it spawns a batch file which requires a shell to be true.

### DIFF
--- a/build/lib/android/index.js
+++ b/build/lib/android/index.js
@@ -173,7 +173,11 @@ class Android {
 
 async function gradlew(args) {
 	await new Promise((resolve, reject) => {
-		const childProcess = spawn(GRADLEW_FILE_PATH, args, { cwd: TITANIUM_ANDROID_PATH, stdio: 'inherit' });
+		const childProcess = spawn(GRADLEW_FILE_PATH, args, {
+			cwd: TITANIUM_ANDROID_PATH,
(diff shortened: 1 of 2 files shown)


hansemannn (approved):
I don't have a Windows PC to verify, but the change does at least not break macOS, so if it helps on Windows, it's fine.

m1ga:
I'll get my VM setup to build the SDK and check it

m1ga:
yes, it's a node 22 thing. With node 20 I didn't see that error. When I updated I saw:

Bildschirmfoto_20240815_121758
(don't have copy & paste running, so screenshot was quicker :smile: )

After this PR it starting building again. It's still at 48% for 5mins now but I blame the VM for that...

Building with 22.5.0 under linux still works fine with the current master but I see many
```
(node:24216) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```
warnings during the build process.

m1ga (approved):
build starts on Windows again after applying this PR (node 22.6.0)

cb1kenobi:
Awesome, thank you guys for the review! Let's go!

PR #14099, opened 2024-08-29
feat(android): webview content will fill up the whole Ti.UI.WebView by default

### LINKED ISSUE #14098: Android: A canvas inside a local webview has the wrong height if it is specified using percentages (i.e. 100%)

### Description

NOTE: Maybe this issue is not directly related to Titanium SDK but to the Canvas - HTML5 framework.

A canvas within a local webview is not sized correctly if its size is specified as a percentage. This applies both to settings via CSS and to the height and width attributes of the canvas itself.

    <body>
     <style>
      body {margin: 0 0;padding: 0 0;}
      canvas {margin: 0 0;padding: 0 0; width: 100%; height:100%; background-color: blue;}
     </style>
     <canvas id="plot2d" width="100%" height="100%"></canvas>
    </body>
    
FIXING WORKAROUND:
 Programmatically set the dimension of the canvas in pixels.

    <script>
     const canvas=document.getElementById('plot2d');
     function rCnv() {
      canvas.width = window.innerWidth;
      canvas.height = window.innerHeight;
      drawStuff();
     }
     rCnv();
     window.addEventListener('resize', rCnv, false);
    </script>

 If you are using the Chart.js plugin instead use:
 
    <script>
     const canvas=document.getElementById('plot2d');
     const chart = new Chart(canvas.getContext('2d'), {} );
     function rCnv(){
      chart.resize(window.innerWidth,window.innerHeight);
     }
     rCnv();
     window.addEventListener('resize', rCnv, false);
   </scrip>
   
 while setting maintainAspectRatio and responsive to false inside the chart options:
   
    options: {
     responisve: false,
     maintainAspectRatio: false
   
   
### Expected Behavior

Canvas sizes correctly

### Actual behavior

Height is reduced or the canvas looks like a square, same height as width!

### Reproducible sample

X

### Steps to reproduce

X

### Platform

Android

### SDK version you are using

12.4.0.GA

_No response_

m1ga:
Fixes #14098 

The Webview content is not filling up the whole space so it can be used with `100%` values. If you use px values it will work. If you inspect the element before this PR you'll see it doesn't have a height:

Bildschirmfoto_20240829_225345
(it still renders the background color but the inner DOM elements can't handle % values).

After this PR it will fill up the whole webview (e.g. Ti.UI.FILL, nothing, or a number value) and the elements inside will respond to percentage values.

Bildschirmfoto_20240829_225547

**Test**
```js
var w = Ti.UI.createWindow({});
var h = Ti.UI.createWebView({
	html: '<meta name="viewport" content="width=device-width, initial-scale=1"><body style="background-color:red">  <canvas style="width: 100%; height: 100%; background-color:blue"></canvas></body>'
});
w.add(h);
w.open();
```

* run this in the current SDK -> red webview
* run this PR -> blue canvas is visible


apidoc/Titanium/UI/WebView.yml
```diff
@@ -1027,6 +1027,13 @@ properties:
     since: "1.8.0"
     default: true
 
+  - name: fill
```
hansemannn, line 1030: I am not a fan of another platform-specific property here. How does iOS behave? If it's the same, we should keep it the same, if it's different, they should be aligned.

hansemannn (changes requested):

apidoc/Titanium/UI/WebView.yml
m1ga: Good point. I didn't check iOS as the issue was only reported on Android. I can check that later. If it will show a blue background I'll remove the flag and make it the default

m1ga: yes, iOS is blue:

Simulator Screenshot - iPhone 16 Plus - 2024-11-09 at 19 50 09

so I've made it a default value for Android :+1:

hansemannn (approved):
Perfect, thanks!

PR #14104, opened 2024-09-05
chore: java/node version update

m1ga:
Testing if I can remove the 
```
 Not a number: 33x:
java.lang.NumberFormatException: Not a number: 33x
	at com.sun.xml.bind.DatatypeConverterImpl._parseInt(DatatypeConverterImpl.java:95)
```
warnings. I don't have them when I build locally. 

Updating various version in the github action to see if anything helps. It still builds the SDK and I don't see them locally (Fedora).

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -16,7 +16,7 @@ runs:
       run: |
         sudo dpkg --add-architecture i386
         sudo apt-get update -y -qq
-        sudo apt-get install -y -qq gperf libatomic1:i386 libc6:i386 libncurses5:i386 libstdc++6:i386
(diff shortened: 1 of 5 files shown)


cb1kenobi:
I wish this was easiser. Thank you for trying!

m1ga:
sorry if you get notifications all the time :blush: It's strange that it is only happening online and not locally. It looks like it is gradle, so the other PR that raises gradle should fix it.

cb1kenobi:
Notifications are fine. :)

Ideally we could/should put the gradle version behind a feature flag, the users can opt-in until the next major where we flip the default. I understand this takes a bit of effort and I have been stalling. Sigh.

PR #14110, opened 2024-09-10
fix(android): fix Material3 BottomNavigation height

m1ga:
## New PR:

I've only add the note in the docs as I've started a rewrite of the BottomNavigation in another PR. 

## Old PR (just as a reference):
SO issue in native land: https://stackoverflow.com/questions/72062902/how-do-i-get-the-bottom-navigation-bar-height-for-material-design-3 and some screenshots.

Currently we use `design_bottom_navigation_height` for the height of the bottom navigation (56dp). It looks like Material 3 is using `m3_bottom_nav_min_height` (80dp). 

We need to check if the app is using a Material3 theme and then use the new dimens value.

**Test**
```js
var win1 = Ti.UI.createWindow({});
var tf = Ti.UI.createTextField({ bottom: 0, width: Ti.UI.FILL, borderWidth: 1, borderColor: "yellow"})
var win2 = Ti.UI.createWindow({});
win1.add(tf);

var tab1 = Ti.UI.createTab({ window: win1, title: 'Blue' }),
	tab2 = Ti.UI.createTab({ window: win2, title: 'Red'}),
	tabGroup = Ti.UI.createTabGroup({ tabs: [tab1, tab2],
		theme: "Theme.Titanium.Material3.DayNight",                              // <- remove for Material 2
		style: Titanium.UI.Android.TABS_STYLE_BOTTOM_NAVIGATION
	});
tabGroup.open();
```

Yellow text field should be fully visible in the tab.

**Questions:**
Can we use `activity.getTheme().toString().indexOf("Material3")` to check for Material3 theme?

**Workaround**
If you use M3 in your app you can add  `/app/platform/android/res/values/dimens.xml`  with
```xml
<resources>
    <dimen name="design_bottom_navigation_height">80dp</dimen>
</resources>
```
to override the current value.


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java
```diff
@@ -79,13 +80,18 @@ public void addTab(TabProxy tabProxy)
...
 	public void addViews(TiBaseActivity activity)
 	{
 		// Manually calculate the proper position of the BottomNavigationView.
-		int resourceID = activity.getResources().getIdentifier("design_bottom_navigation_height", "dimen",
-															   activity.getPackageName());
-		this.mBottomNavigationHeightValue = activity.getResources().getDimensionPixelSize(resourceID);
+		this.mBottomNavigationHeightValue
+			= activity.getResources().getDimensionPixelSize(R.dimen.design_bottom_navigation_height);
+
+		if (activity.getTheme().toString().indexOf("Material3") > -1) {
```
hansemannn, line 91: This looks hacky. Do we even need a manually-set height? I would prefer to let the system decide the value and use constraints to automatically align other views that are dependent on the height.

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java
m1ga: that would be the best of course. Currently the SDK is using

https://github.com/tidev/titanium-sdk/blob/dd18cd1479a8240ab74474c727e87c48c2b584d0/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigationTabGroup.java#L88-L89

for the BottomNav height (https://github.com/tidev/titanium-sdk/pull/10358) which is causing the issue in Material 3 as they've renamed the variable with the proper height.
I left it as a draft to see if it might be a different way once we have the new gradle/material library included. I'll recheck it after that.

m1ga: The new gradle updates didn't help for this one. It still doesn't find the correct height without pointing it to the new R.dimen. 

I think the real issue is that the current implementation relies on the height instead of constraints (like we are doing with the ListView now) because all the parts are created in code. It might even be better to move it to a XML layout but that would be a bigger change.

As I don't like the `if` myself I think I'll close this one and add a note in the documentation how to use the workaround.

m1ga:
This will just add a note to the docs now. As many people still use the old M2 themes and I've started a rewrite of the BottomNavigation in a different PR it should be enough for now. If you use a M3 theme you have to add the XML file with the height yourself

hansemannn:
It can only be seen as a workaround, as we definitely need the "clean" solution from the other PR

hansemannn (approved):
Using this one until the new bottom nav is finished.

PR #14112, opened 2024-09-14
chore(android): forward --sdk in module build process

m1ga:
When testing the gradle8 PR I saw that inside a module when I run `ti build -p android --sdk 12.6.0` it was building the module with 12.6.0 but the example app was still starting with `12.4.0.GA`. 

This PR will forward the `--sdk` parameter and it won't crash with `Unsupported Class File Major Version 61` errors (module is build with java 17 but then run with java 11)

**Testing:**
* build an Android module with an example folder
* `ti build -p android -l trace --sdk 12.6.0`  should show:
```
[DEBUG] Running example project...
[DEBUG] Titanium Command-Line Interface v7.1.1 SDK v12.6.0
```
after building the module ZIP. 

* copy `android/cli/commands/_buildModule.js` to 12.4.0.GA and build the module again:
* `ti build -p android -l trace`  (no custom SDK, it should take the default SDK) should show:
```
[DEBUG] Running example project...
[DEBUG] Titanium Command-Line Interface v7.1.1 SDK v12.4.0
```

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -33,6 +33,8 @@ const androidDetect = require('../lib/detect').detect,
 	__ = appc.i18n(__dirname).__,
 	version = appc.version;
 
+let tiSdkVersion = '12.6.0.GA';
+
 function AndroidModuleBuilder() {
 	Builder.apply(this, arguments);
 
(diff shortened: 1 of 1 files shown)


android/cli/commands/_buildModule.js
```diff
@@ -33,6 +33,8 @@ const androidDetect = require('../lib/detect').detect,
 	__ = appc.i18n(__dirname).__,
 	version = appc.version;
 
+let tiSdkVersion = '12.6.0.GA';
```
hansemannn, line 36: Can't this be taken from the package.json / `tiapp` util? Or should it remain this one, even for newer builds?

m1ga: it will be overwritten in `AndroidModuleBuilder`. I thought it might be good to have a default value

m1ga: but let me check if I can move that right to the initialization

cb1kenobi (approved):
This is a good catch! LGTM

Issue #14051, opened 2024-05-31
iOS: can click speed be improved?


m1ga:
### I have searched and made sure there are no existing issues for the issue I am filing

- [X] I have searched the existing issues

### Description

If you run this simple app on Android and iOS you'll see that the click event on Android fires quicker:
```js
const win = Ti.UI.createWindow();
win.addEventListener("click", function(e) {
	console.log("click")
})


win.open();
```

I briefly looked into the iOS code and found:
https://github.com/tidev/titanium-sdk/blob/2f1212fd9e46a4a8aeae0af7a850aaa7bff34e52/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m#L1612-L1621

If I add another 
```
else if ([touch tapCount] >= 2 && [proxy _hasListeners:@"click"]) { ... }
```
and fire a click event inside that I can click the window a lot quicker and receive click events on iOS. Has this been done on purpose?

Workaround is to use `singleTap` on iOS

### Expected Behavior

Fire click event as fast as the user clicks

### Actual behavior

small cool-down between the clicks is needed on iOS.

### Reproducible sample

```js
const win = Ti.UI.createWindow();
win.addEventListener("click", function(e) {
	console.log("click")
})


win.open();
```

### Steps to reproduce

* click the window and watch the logs 

### Platform

iOS

### SDK version you are using

12.3.1

m1ga:
fixed in https://github.com/tidev/titanium-sdk/pull/14115

PR #14115, opened 2024-09-15
fix(ios): ios click speed

AbdullahFaqeir:
- Enhance click event speed #14051

Used single tab recogniser.


iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
```diff
@@ -1609,12 +1612,7 @@ - (void)processTouchesEnded:(NSSet *)touches withEvent:(UIEvent *)event
     // Click handling is special; don't propagate if we have a delegate,
     // but DO invoke the touch delegate.
     // clicks should also be handled by any control the view is embedded in.
```
hansemannn, line 1614: This comment still sounds reasonable. Have you tested all other click cases as well?

AbdullahFaqeir: but hierarchically the gesturer event will propagate eventually.

hansemannn (approved):

PR #14120, opened 2024-09-20
feat(android): add support for Android's API 33 Per-app languages

AbdullahFaqeir:
This PR adds support for Android's Per-app languages presented in Android API 33/Android 12.

See [Per-app languages preferences](https://developer.android.com/guide/topics/resources/app-languages)

### DIFF
--- a/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
+++ b/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
@@ -43,7 +43,7 @@ public class LocaleModule extends KrollModule
 	public void setApplicationLocales(String locales)
 	{
 		if (Build.VERSION.SDK_INT < 33) {
-			Log.w(TAG, "This property is only supported on android API 33 and above.");
+			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
 			return;
 		}
 		LocaleListCompat appLocale = LocaleListCompat.forLanguageTags(locales);
--- a/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
+++ b/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
@@ -54,7 +54,7 @@ public class LocaleModule extends KrollModule
 	public KrollDict[] getApplicationLocales()
 	{
 		if (Build.VERSION.SDK_INT < 33) {
-			Log.w(TAG, "This property is only supported on android API 33 and above.");
+			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
 			return new KrollDict[0];
 		}
 		LocaleListCompat localeListCompat = AppCompatDelegate.getApplicationLocales();
--- a/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
+++ b/android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
@@ -10,8 +10,10 @@ import java.text.Collator;
 import java.text.DateFormat;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.Locale;
 
+import org.appcelerator.kroll.KrollDict;
 import org.appcelerator.kroll.KrollModule;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.kroll.common.Log;
@@ -21,7 +23,11 @@ import org.appcelerator.titanium.util.TiLocaleManager;
 import org.appcelerator.titanium.util.TiPlatformHelper;
 import org.appcelerator.titanium.util.TiRHelper;
 
+import android.os.Build;
 import android.telephony.PhoneNumberUtils;
+
+import androidx.appcompat.app.AppCompatDelegate;
+import androidx.core.os.LocaleListCompat;
 
 @Kroll.module
 public class LocaleModule extends KrollModule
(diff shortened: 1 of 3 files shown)


AbdullahFaqeir:
@hansemannn @m1ga a fresh clean PR for the Per-app languages.

m1ga:
Nice :+1:  This looks a lot better now.
one test case is in here: https://github.com/tidev/titanium-sdk/pull/14008#issuecomment-2009179503 

Will start testing your PRs tomorrow but from a quick first look it looked fine

AbdullahFaqeir:
@m1ga,

Will test using this, tho I was thinking about the automated unit testing.

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -33,6 +39,62 @@ public LocaleModule()
 		super("Locale");
 	}
 
+	@Kroll.setProperty
+	public void setApplicationLocales(String locales)
+	{
+		if (Build.VERSION.SDK_INT < 33) {
+			Log.w(TAG, "This property is only supported on android API 33 and above.");
```
m1ga, line 46:
```suggestion
			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
```

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -33,6 +39,62 @@ public LocaleModule()
...
+		}
+		LocaleListCompat appLocale = LocaleListCompat.forLanguageTags(locales);
+		AppCompatDelegate.setApplicationLocales(appLocale);
+	}
+
+	@Kroll.getProperty
+	public KrollDict[] getApplicationLocales()
+	{
+		if (Build.VERSION.SDK_INT < 33) {
+			Log.w(TAG, "This property is only supported on android API 33 and above.");
```
m1ga, line 57:
```suggestion
			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
```

apidoc/Titanium/Locale/Locale.yml
```diff
@@ -235,3 +235,14 @@ properties:
...
     type: String
     permission: read-only
+
+  - name: applicationLocales
+    summary: Update or retrieve Android's Per-app languages.
+    platforms: [android]
+    description: |
+        This property holds the values of Android's Per-app languages which is an array of defined Per-app languages, 
+        if no languages defined, an empty array will be returned. You can set the value of it like this `en-US,en-GB,ar-SA`.
+        This property requires android API 33 and above.
```
m1ga, line 245:
```suggestion
        This property requires Android API level 33 and above.
```

m1ga (changes requested):
just some minor string changes you can apply. 

I've used my test code from the comment above and works correctly:
* starting the app: empty array as no per-app language is set
* changing it in the app and then going to "app preferences - per-app language" shows the correct value
* changing the value in "app preferences - per-app language" and opening the app shows the correct value in the app

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
m1ga: that's not fixed

m1ga (approved):

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -33,6 +39,62 @@ public LocaleModule()
...
+		}
+		LocaleListCompat localeListCompat = AppCompatDelegate.getApplicationLocales();
+		int size = localeListCompat.size();
+		KrollDict[] locales = new KrollDict[size];
+		for (int i = 0; i < size; i++) {
+			Locale locale = localeListCompat.get(i);
+			if (locale != null) {
+				KrollDict localeObj = new KrollDict();
+				localeObj.put("country", locale.getCountry());
+				localeObj.put("iso3_country", locale.getISO3Country());
```
hansemannn, line 68: Properties should always be camelCase

apidoc/Titanium/Locale/Locale.yml
```diff
@@ -235,3 +235,14 @@ properties:
...
+
+  - name: applicationLocales
+    summary: Update or retrieve Android's Per-app languages.
+    platforms: [android]
+    description: |
+        This property holds the values of Android's Per-app languages which is an array of defined Per-app languages, 
+        if no languages defined, an empty array will be returned. You can set the value of it like this `en-US,en-GB,ar-SA`.
+        This property requires Android API level 33 and above.
+        See the [Android's Per-app languages](https://developer.android.com/guide/topics/resources/app-languages).
+    type: Array
```
hansemannn, line 247: Add the array typings

android/templates/build/ti.constants.gradle
```diff
@@ -7,7 +7,7 @@
 
 project.ext {
 	tiNdkVersion = '26.2.11394342'
-	tiAndroidXAppCompatLibVersion = '1.4.1'
+	tiAndroidXAppCompatLibVersion = '1.6.1'
```
hansemannn, line 10: Is this required for the change?

hansemannn (changes requested):
Some open issues to look at

android/templates/build/ti.constants.gradle
AbdullahFaqeir: Yes, as both methods, get and set application languages introduced in 1.6^

m1ga: yes, we need to be in the 1.6.0 versioon: `Use the setApplicationLocales() and getApplicationLocales() methods in Appcompat 1.6.0 or higher` ([source](https://developer.android.com/guide/topics/resources/app-languages#androidx-impl))

1.7.x won't work. I've tested that. Needs more libraries to be updated

apidoc/Titanium/Locale/Locale.yml
AbdullahFaqeir: can you please explain

android/templates/build/ti.constants.gradle
AbdullahFaqeir: Roger that.

m1ga: 1.6.1 was fine too as it is a patch release in 1.6.x it should be the preferred version.

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -33,6 +39,62 @@ public LocaleModule()
...
+		if (Build.VERSION.SDK_INT < 33) {
+			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
+			return;
+		}
+		LocaleListCompat appLocale = LocaleListCompat.forLanguageTags(locales);
+		AppCompatDelegate.setApplicationLocales(appLocale);
+	}
+
+	@Kroll.getProperty
+	public KrollDict[] getApplicationLocales()
```
hansemannn, line 54: It is very uncommon to have different types for the getter and setter. And there is `Ti.Locale.setLanguage` already. So I would propose:
- Use the new Android 33+ API in `Ti.Locale.setLanguage(language)`
- Expose the getter as a new Titanium API (if there isn't one already to get language details - if so, update that one)

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
```diff
@@ -33,6 +39,62 @@ public LocaleModule()
...
+	{
+		if (Build.VERSION.SDK_INT < 33) {
+			Log.w(TAG, "This property is only supported on Android API level 33 and above.");
+			return new KrollDict[0];
+		}
+		LocaleListCompat localeListCompat = AppCompatDelegate.getApplicationLocales();
+		int size = localeListCompat.size();
+		KrollDict[] locales = new KrollDict[size];
+		for (int i = 0; i < size; i++) {
```
hansemannn, line 63: The loop looks a bit "oldskool". Things to change:
- Use a more modern loop technique, e.g. [this one](https://stackoverflow.com/a/18410489/5537752) and append it a to an array instead of maintaining an `index`
- Do not add all the logic in the null check. Instead, check for `locale == null` and `continue` early

apidoc/Titanium/Locale/Locale.yml
hansemannn: See the newer comments to split the API. But in general, you just have an array as the return value, but really it's `Array<LocaleDetails>` to document all properties. This should not just refer to the Android docs, as there may be newer properties in the future that are not exposed in Titanium, yet.

hansemannn (changes requested):

android/modules/locale/src/java/ti/modules/titanium/locale/LocaleModule.java
AbdullahFaqeir: Seems like this PR is with no use so I'll close it

PR #14130, opened 2024-10-03
feat(ios): add direction to ListView scrolling

m1ga:
iOS currently only adds `direction` in the scrolling event when you release the finger. This PR adds it to the continuous scrolling ([old PR](https://github.com/tidev/titanium-sdk/pull/13830))

```js
var toggle = false;
var sections = [];
	title: "change",
	bottom: 20
});
	top: 10,
	right: 10,
	width: Ti.UI.SIZE,
	height: Ti.UI.SIZE
})
	height: Ti.UI.FILL,
	width: Ti.UI.FILL,
	continuousUpdate: true,
  forceUpdates: false
});

	toggle = !toggle;
	listView.forceUpdates = toggle
	console.log(listView.forceUpdates);
})

for (var s = 0; s < 5; s++) {
		headerTitle: 'Section ' + s
	});
	var set = [];
		set.push({
			properties: {
				title: 'Item 0 ' + i
			}
		})
	}
	section.setItems(set);
	sections.push(section);
}

listView.sections = sections;

	console.log("Item", e.direction, e.firstVisibleItemIndex, e.firstVisibleSectionIndex, e.top);
})
win.open();
```

https://github.com/user-attachments/assets/f5c36153-0ef1-4e90-a21c-f4a1d121b75b

### DIFF
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -2012,7 +2012,7 @@ static TiViewProxy *FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoint
         TiUIListSectionProxy *section;
         CGFloat topSpacing = scrollView.contentOffset.y + scrollView.adjustedContentInset.top;
 
-        NSString *direction = nil;
+        NSString *direction = [NSNull null];
 
         if (self.lastContentOffset > scrollView.contentOffset.y) {
           direction = @"down";
--- a/iphone/Classes/TiUIListView.m
+++ b/iphone/Classes/TiUIListView.m
@@ -2012,7 +2012,7 @@ static TiViewProxy *FindViewProxyWithBindIdContainingPoint(UIView *view, CGPoint
         TiUIListSectionProxy *section;
         CGFloat topSpacing = scrollView.contentOffset.y + scrollView.adjustedContentInset.top;
 
-        NSString *direction = [NSNull null];
+        NSString *direction = @"unknown";
 
         if (self.lastContentOffset > scrollView.contentOffset.y) {
           direction = @"down";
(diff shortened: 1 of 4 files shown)


hansemannn:
Are there API docs for this already?

m1ga:
yes, since 5.4.0
https://github.com/tidev/titanium-sdk/blob/master/apidoc/Titanium/UI/ListView.yml#L648-L652
With a note `On iOS no event is fired when the finger is not released`. With `continuousUpdate` event you've got it more often but the `direction` was not yet in there. That was still only fired after touching the screen.

iphone/Classes/TiUIListView.m
```diff
@@ -2011,6 +2012,15 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
         TiUIListSectionProxy *section;
         CGFloat topSpacing = scrollView.contentOffset.y + scrollView.adjustedContentInset.top;
 
+        NSString *direction = nil;
```
hansemannn, line 2015:
```suggestion
        NSString *direction = [NSNull null];
```

Otherwise, if the `if` and `else if` are both not matching, it would remain `nil` and crash the app, as `nil` cannot be bridged to a JS type. Therefore, `nil` always should default to `[NSNull null]` which is the equivalent to `null` in JS.

hansemannn (changes requested):

iphone/Classes/TiUIListView.m
m1ga: I see. Should I change it here too: https://github.com/tidev/titanium-sdk/blob/a7ac1d42103859289e61806b24727fc6844e5554/iphone/Classes/TiUIListView.m#L2125 I've copied it from there

iphone/Classes/TiUIListView.m
```diff
@@ -2012,7 +2012,7 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
         TiUIListSectionProxy *section;
         CGFloat topSpacing = scrollView.contentOffset.y + scrollView.adjustedContentInset.top;
 
-        NSString *direction = nil;
+        NSString *direction = [NSNull null];
```
hansemannn, line 2015: Was the typing okay or did Xcode complain?

iphone/Classes/TiUIListView.m
hansemannn: Ideally yes

iphone/Classes/TiUIListView.m
m1ga: I didn't check it with Xcode, I've only copied it with a normal text editor, build the SDK and checked it in an app :smile:

hansemannn:
@m1ga The recent changes should fix that - are you okay with these?

m1ga:
Looking good, code still works fine, thanks :+1:

PR #14131, opened 2024-10-03
feat(android): implement Android Photo Picker

m1ga:
# Update: As mentioned in a comment below: the current picker will use the new Photo Picker already, so no changes are required. Still keeping this code as it adds `maxImages` and we'll might want to switch later

---

Will replace the current openPhotoGallery method and use https://developer.android.com/training/data-storage/shared/photopicker

When I was updating an Android app I was asked in the store to use the Photo Picker instead of using `<uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>`. But you could still opt-in for the image permission, so not a high priority for now.

I have to call `ActivityResultLauncher` in the BaseActivity otherwise I wasn't able to catch the results. And in order to get the results from the BaseActivity back to MediaModule I use a local Broadcast. Not sure if that is the best way to do it but I didn't find another way.

```js
const img = Ti.UI.createImageView({widht: 100,height:100});
const btn1 = Ti.UI.createButton({title: "image - single"});
const btn2 = Ti.UI.createButton({title: "image - multiple"});
const btn3 = Ti.UI.createButton({title: "image - multiple max 3"});
const btn4 = Ti.UI.createButton({title: "video - single"});
const btn5 = Ti.UI.createButton({title: "video - multiple"});
const btn6 = Ti.UI.createButton({title: "video - multiple max 3"});

btn1.addEventListener("click", () => openGal(false, false, 0));
btn2.addEventListener("click", () => openGal(true, false, 0));
btn3.addEventListener("click", () => openGal(true, false, 3));
btn4.addEventListener("click", () => openGal(false, true, 0));
btn5.addEventListener("click", () => openGal(true, true, 0));
btn6.addEventListener("click", () => openGal(true, true, 3));


function openGal(isMultiple, isVideo, maxImages) {
	Ti.Media.openPhotoGallery({
		allowMultiple: isMultiple,
		maxImages: maxImages,
		mediaTypes: isVideo ? Titanium.Media.MEDIA_TYPE_VIDEO : Titanium.Media.MEDIA_TYPE_PHOTO,
		success: function(e) {
			if (!isVideo) {
				if (isMultiple) {
					console.log("Multiple images");
					console.log(e.images.length);
					img.image = e.images[0].media;
				} else {
					console.log("Single image");
					img.image = e.media;
				}
			} else {
				console.log("video")
				if (isMultiple) {
					console.log("Multiple videos");
					e.videos.forEach(function(item) {
						console.log(item.width, item.height)
					})
				} else {
					console.log("Single video");
					console.log(e.width, e.height)
				}
			}
		}
	});
}

win.add([btn1,btn2,btn3,btn4,btn5,btn6,img]);
win.open();
```


m1ga:
I think I have all parts now :smile: Image, Video, Image & Video. Multiple and single. Multiple with `maxImages`

android/modules/media/src/java/ti/modules/titanium/media/MediaModule.java
```diff
@@ -1105,27 +1122,15 @@ public void openPhotoGallery(KrollDict options)
...
+		fCancelCallback = cancelCallback;
+		fErrorCallback = errorCallback;
 
 		Log.d(TAG, "openPhotoGallery called", Log.DEBUG_MODE);
 
 		Activity activity = TiApplication.getInstance().getCurrentActivity();
 		TiActivitySupport activitySupport = (TiActivitySupport) activity;
 
-		TiIntentWrapper galleryIntent = new TiIntentWrapper(new Intent());
-		galleryIntent.getIntent().setAction(Intent.ACTION_GET_CONTENT);
```
prashantsaini1, line 1118: @m1ga Under the hood `ACTION_GET_CONTENT` already invokes new Photo picker when it's available, otherwise it invokes system picker. I tested it out on 12.5.1.GA on Android 15 device and it invokes new Photo picker.

prashantsaini1: Scroll down to `ACTION_GET_CONTENT behaviour change` on [this page](https://medium.com/androiddevelopers/permissionless-is-the-future-of-storage-on-android-3fbceeb3d70a) for more details.

m1ga: ah that is good too know! I thought the new Photo Picker is _only_ used when you use PickVisualMediaRequest as they don't mention the ACTION_GET_CONTENT in their page.

Then I guess we don't need to do anything here :-) I'll put this into draft so we don't need to do anything here but still have to code

prashantsaini1:
@m1ga This [major issue](https://issuetracker.google.com/issues/264952941) is worth keeping an eye even in future as it makes new Photo picker much less useful if it cannot show all albums as of now and unfortunately Google isn't working on this bug.

PR #14137, opened 2024-10-14
feat(ios): added support for iOS 18 dark mode icons

AbdullahFaqeir:
As mentioned this issue #14122, this was missing.


iphone/layout/layout/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,68 +1,117 @@
...
+      "size" : "1024x1024"
+    },
+    {
+      "appearances" : [
+        {
+          "appearance" : "luminosity",
+          "value" : "tinted"
+        }
+      ],
+      "filename" : "DefaultIcon-Tinted.png",
```
hansemannn, line 22: Does this require additional files by default or is it fully backwards compatible? How does the generated Xcode project handle their absence?

dlewis23:
We have to also update the _build.js file to add the dark and tinted icons. Also have tell it to not flatten against a white background. The dark mode icon is not suppose to have a background the system adds a gradient background automatically.

Attached is my _build.js file. I'm sure there is a better way to handle not flattening against a white background. 

[_build.js.zip](https://github.com/user-attachments/files/17450445/_build.js.zip)

iphone/layout/layout/Assets.xcassets/AppIcon.appiconset/Contents.json
dlewis23: We should have no issue because in the _build.js the -dark and -tinted icons have `minXcodeVer: '16.0'` so older versions should skip it and if the icon is not given the os makes its own version.

hansemannn:
@dlewis23 You can do a PR to @AbdullahFaqeir's branch, then it appears here in the PR once that PR is merged :)

dlewis23:
@hansemannn I tried to make a PR to it, but Github desktop keeps giving me an error when I try and publish the PR which I have not been able to figure out yet.

dlewis23:
I can't do a PR to @AbdullahFaqeir branch, it gives an error every time, but somehow went down a loop and ended up making a new PR here on accident with the changes: https://github.com/tidev/titanium-sdk/pull/14138

hansemannn:
Closing in favor of https://github.com/tidev/titanium-sdk/pull/14138 as it seems newer and have been developed together.

PR #14139, opened 2024-10-24
fix(android): optimize cameraX rotation after camera is active

m1ga:
When you use cameraX and rotate the camera AFTER opening it in portrait mode it won't update the target rotation and the image had the wrong orientation.

**Test**

```js
var img = Ti.UI.createImageView();
	Ti.Media.requestCameraPermissions(function(e) {
		if (e.success) {
			var overlay = Ti.UI.createView();
				bottom: 0,
				title: "take"
			})
			overlay.add(btn);
			Ti.Media.showCamera({
				overlay: overlay,
				useCameraX: true,
				success: function(e){
					img.image = e.media;
				}
			});

				Ti.Media.takePicture();
			})
		}
	});
})
win.add(img);
win.open();
```

Run the app and click on the screen to open the camera.
* Hold it in Portrait mode
* click the screen
* rotate it horizontally 
* take a picture
*  rotate it back to portrait and look at the imageview
* it should have the correct orientation

Test 2:
* start in landscape and take a picture


android/modules/media/src/java/ti/modules/titanium/media/TiCameraXActivity.java
```diff
@@ -417,6 +421,30 @@ private void startCamera()
...
+				}
+				int rotation = getWindowManager().getDefaultDisplay().getRotation();
+				if (lastDisplayOrientation != rotation) {
+					imageCapture.setTargetRotation(rotation);
+					lastDisplayOrientation = rotation;
+				}
+
+			}
+		};
+		if (orientationEventListener.canDetectOrientation()) {
```
hansemannn, line 441: Wouldn't it make more sense to do the whole constructore only if orientation is detected? That way, you don't need to `null` it again

hansemannn (changes requested):

android/modules/media/src/java/ti/modules/titanium/media/TiCameraXActivity.java
hansemannn: Ah, of course!

hansemannn (approved):

PR #14141, opened 2024-11-02
fix(android): fix starting activity when launching from home screen

m1ga:
This is a tricky one: 

When I launch my apps from the home screen I most of the time have this during start:
splash screen/icon and then it moves to the left and my app is starting: 

[Bildschirmaufnahme_20241102_124139.webm](https://github.com/user-attachments/assets/7347d657-ea88-4193-930a-b955a79f9323)

It doesn't happen when you launch it from the normal app launcher. And it doesn't happen for all apps and even some only do that from the first home screen and not from the 2nd. Super strange!

Looks like the issue is in the `filterEquals(intent)` part here: https://github.com/tidev/titanium-sdk/blob/master/android/titanium/src/java/org/appcelerator/titanium/TiRootActivity.java#L263

When I dump the intent content the first one is:
```
Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 pkg=com.miga.climberino cmp=com.miga.climberino/.ClimberinoActivity bnds=[1009,1786][1207,2112] } Bundle[null]
```
and the second one is:
```
Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.miga.climberino/.ClimberinoActivity } Bundle[null]
```

It is still the same BUT `pkg` is added in one and not the other depending on where you launch it from. 

The fix is: just adding the package name to both intents so they are always there so the check won't fail when one of the launch intents has that and the other doesn't.


m1ga:
@prashantsaini1 @hansemannn you can see the issue in your live app too: close/kill the app and use the play store to search for your app and launch it with the "open" button. It will show the splash screen, that moves out to the right and the first window will move in from the left. 
With the PR it will fade from splashscreen to the window, like a normal start from the launcher.

Also got feedback by another client that it fixes the launch behavior when starting it from the play store.

prashantsaini1:
Just tested out the PR and find the fix majorly working as we are now explicitly telling intent to choose only the required app package, rather than finding from all and launching it.
The easy way to reproduce this issue is to open the native app info page on long press the app icon and then opening the app from there. It also creates two launcher activities, one being blank. Notifications and intent filters also works fine.

prashantsaini1 (approved):

PR #14147, opened 2024-11-24
chore(android): remove aps-analytics.jar and titanium-verify.jar

m1ga:
Removed the `aps-analytics.jar` library and created the files locally. 
At a future point it should be removed. At the moment this should be the safest way to still have it without using the library.

Looks like `titanium-verify.jar` is also not needed anymore.


prashantsaini1:
We can remove the APS stuff as it's not required anymore with no Appcelerator Dashboard access around for its analytics.

hbugdoll:
Was marked as deprecated over 3 years ago in Ti 11.2.0 (https://github.com/tidev/titanium-sdk/commit/3e549242e8421601dc5f98bf68cb16e723bf7d27). Now we are two major versions further on.

Removing it with release of Ti 13.0.0 could have been the right moment. Should we now wait until next major release?

m1ga:
The files are not use anymore. Just some methods e.g. in `Ti.Platform`:
```java
public String getOstype()
	{
		return APSAnalyticsMeta.getOsType();
	}
```
call methods out of the classes. That's why I've kept the needed parts but just removed the jar files so we can reduce the code to needed methods.

hbugdoll:
> Just some methods e.g. in `Ti.Platform`:
> 
> ```java
> public String getOstype()
> 	{
> 		return APSAnalyticsMeta.getOsType();
> 	}
> ```
> 
> call methods out of the classes.

Exactly!
I was going to ask you about APSAnalyticsMeta and how to deal with it.

m1ga:
we could also move those methods into another file but since it's not making any external connection I just kept the `APS` name and classes

hbugdoll:
>  I just kept the APS name and classes

Ok.

The class attribute [analyticsUrl](https://github.com/tidev/titanium-sdk/blob/9c938933a9e17f4710bb187439d97f964a6dcc70/android/titanium/src/java/org/appcelerator/titanium/util/APSAnalyticsMeta.java#L22) can also be removed.

m1ga:
> > I just kept the APS name and classes
> 
> Ok.
> 
> The class attribute [analyticsUrl](https://github.com/tidev/titanium-sdk/blob/9c938933a9e17f4710bb187439d97f964a6dcc70/android/titanium/src/java/org/appcelerator/titanium/util/APSAnalyticsMeta.java#L22) can also be removed.

thanks :+1: also removed

cb1kenobi (approved):
Tested, no errors, looks good!

PR #14154, opened 2024-12-13
fix(ios): optimize click event

### LINKED ISSUE #14051: iOS: can click speed be improved?

m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/14051

```js
win.addEventListener("click", function(e) {
	console.log("click")
})


win.open();
```

This was my initial idea to improve the click event. 

Currently it only uses "click" and "dblclick" but if you continue to click it won't fire click events. More info in the issue https://github.com/tidev/titanium-sdk/issues/14051

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m
@@ -1617,6 +1617,9 @@ DEFINE_EXCEPTIONS
     } else if ([touch tapCount] == 2 && [proxy _hasListeners:@"dblclick"]) {
       [proxy fireEvent:@"dblclick" withObject:evt propagate:YES];
       return;
+    } else if ([touch tapCount] > 2 && [proxy _hasListeners:@"click"]) {
+      [proxy fireEvent:@"click" withObject:evt propagate:YES];
+      return;
     }
   }
 }


hansemannn:
I honestly don't feel very confident right now merging something so sensitive again right now, for two reasons:
1. The first one broke quite many apps after being unnoticed during CR
2. I didn't see many devs raging against the current way it works, neither with performance issues or non-native behaviors

Therefore, I would either ask for a community feedback to understand what issues they have (we don't have any for example and none of our users ever reported something like slow click behavior) or keep it as-is for now.

m1ga:
sounds good. Think most people used the singleTap on iOS if they wanted to have faster click events (+/- buttons) in the past instead of click.

PR #14158, opened 2024-12-18
fix: make android mocha tests run again

m1ga:
I was wondering why the Android mocha tests weren't running anymore. Turns out: the generated tiapp.xml was invalid. There was a missing `>` and it had two `<application>` (we have a default one in the template now) nodes.

Fixing some tests:
* tls 1.3 url was not working. Using a different one
* redirect url was blocked, using a different one
* one test was throwing a "getText() of null" error for the label so I've added a null check: https://github.com/tidev/titanium-sdk/pull/14174
* ~headerTitle/footerTitle is not working (even with 12.5.0) so disabling that for the moment~  - Fixed in https://github.com/tidev/titanium-sdk/pull/14160
* added another density image for the textfield padding test

Interesting point:
`tableView.deleteRow(tableView.sections[0].rows[2]);` is not working in the emulator but on a device it is working fine :shrug: That's one test that is still failing but couldn't fix it yet.


tests/Resources/ti.ui.listview.test.js
```diff
@@ -1418,7 +1418,8 @@ describe('Titanium.UI.ListView', function () {
 		should(view).matchImage('snapshots/listViewItem_template.png', { maxPixelMismatch: OS_IOS ? 23 : 0 }); // 23 pixels differ on actual iPhone
 	});
 
-	it('ListView header & footer', function () {
+	it.skip('ListView header & footer', function () {
+		// FIXME: Not working in latest SDK
```
m1ga, line 1422:
```suggestion
	itListView header & footer', function () {
```

tests/Resources/ti.ui.listview.test.js
```diff
@@ -1418,7 +1418,7 @@ describe('Titanium.UI.ListView', function () {
 		should(view).matchImage('snapshots/listViewItem_template.png', { maxPixelMismatch: OS_IOS ? 23 : 0 }); // 23 pixels differ on actual iPhone
 	});
 
-	it('ListView header & footer', function () {
+	itListView header & footer', function () {
```
m1ga, line 1421:
```suggestion
	it('ListView header & footer', function () {
```

m1ga:
@prashantsaini1 maybe you can check this one too. Then the testsuite will run again (`npm run tests`). One small null check in the SDK, the rest is just JS fixes.

Just a note: there are still around 10 tests that fail in the emulator, but the outher thousand of tests run fine 😄

prashantsaini1:
This one's fine as I already noticed at other places that `getNativeView` is being used without null checks, whereas it should be guarded.

hansemannn:
@prashantsaini1 Please review

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java
```diff
@@ -809,6 +809,10 @@ public int getLineCount()
 	public String getVisibleText()
 	{
 		MaterialTextView textView = (MaterialTextView) getNativeView();
-		return textView.getLayout().getText().toString();
+		if (textView != null && textView.getLayout() != null) {
```
hansemannn, line 812: This looks unrelated to this change?

prashantsaini1: I believe @m1ga found one test failing because of missing null check here.

prashantsaini1: @m1ga `null` checks are missing in above `getLineCount` method and at some other places as well in this class.

m1ga: yes, one test was failing:

> one test was throwing a "getText() of null" error for the label so I've added a null check

so I've added the fix here. The `getLineCount()` method doesn't have a test but I think you are right it might fail too. We have those tests in [other methods too](https://github.com/tidev/titanium-sdk/blob/4532539fd9cd538f123889d8ccddaf535b42e0f1/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUILabel.java#L715-L718). I'll update it later.

m1ga: I've split it into 2 PRs so the code change is here now: https://github.com/tidev/titanium-sdk/pull/14174

build/lib/test/test.js
```diff
@@ -221,7 +221,7 @@ async function addTiAppProperties() {
 	const tiapp_xml_string = await fs.readFile(tiapp_xml, 'utf8');
 	const content = [];
 	const insertManifest = () => {
-		content.push('\t\t\t<application android:theme="@style/Theme.Titanium.Dark">');
+		content.push('\t\t\t<application android:theme="@style/Theme.Titanium.Dark"  android:icon="@mipmap/ic_launcher">');
```
prashantsaini1, line 224: @m1ga I think the default icon which SDK puts in is `android:icon="@drawable/appicon"`. I do not see any `mipmap` folder in a fresh project.

m1ga: we have mipmaps since 12.4.0: https://github.com/tidev/titanium-sdk/tree/098f8c2785ef3257b61736141d63ffd61b404ef8/templates/app/default/template/platform/android/res

a new project looks like this:
Bildschirmfoto_20250128_103355

prashantsaini1: Yeah, I just double checked, I think I changed the platform folder during some tests. It's fine then!

prashantsaini1 (approved):

PR #14161, opened 2024-12-22
feat(android): --skip-build module parameter to run example/app.js project

m1ga:
## Issue

When building a module you sometimes just make changes in the `example/app.js` and want to run the app using `ti build -p android`. This will always compile the module again, there is no way to skip building the module and just create the temp app project.

## Solution

The `--skip-build` parameter will just zip the existing code and then create the temporary `example/app.js` project to save some time testing modules. It will check if on of the generated folders `/android/build/module/build/outputs/m2repository` is available otherwise it will just run a normal build again.

## Test

Create and build a module with `ti build -p android --skip-build` inside the android folder. The 2nd time it will skip compiling the Java files again and just creates the zip + example app.

## Improvements

If you know what you are doing and use the parameter it can save you ~10sec when you test your modules every build :+1:


cb1kenobi (approved):
LGTM!

PR #14166, opened 2025-01-04
feat(android): expose uiFlags

m1ga:
Expose `uiFlags` on the window.

Use
```js
win.open();
win.uiFlags = 8192;
```
or
```
  uiFlags: 8192
});
win.open();
```

to set `SYSTEM_UI_FLAG_LIGHT_STATUS_BAR` (dark statusbar text color)


android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -70,7 +70,8 @@
 		TiC.PROPERTY_WINDOW_PIXEL_FORMAT,
 		TiC.PROPERTY_FLAG_SECURE,
 		TiC.PROPERTY_BAR_COLOR,
-		TiC.PROPERTY_STATUS_BAR_COLOR
+		TiC.PROPERTY_STATUS_BAR_COLOR,
+		"uiFlags"
```
hansemannn, line 74: New property?

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
```diff
@@ -328,6 +329,10 @@ public void windowCreated(TiBaseActivity activity, Bundle savedInstanceState)
 			win.setStatusBarColor(colorInt);
 		}
 
+		if (hasProperty("uiFlags")) {
```
hansemannn, line 332: Use properties

apidoc/Titanium/UI/Window.yml
```diff
@@ -1537,6 +1537,16 @@ properties:
     since: "3.3.0"
     availability: creation
 
+  - name: uiFlags
+    summary: Additional UI flags to set on the Activity Window.
+    description: |
+        See [WindowManager.LayoutParams](https://developer.android.com/reference/android/view/View#setSystemUiVisibility\(int\)) for
+        list of additional flags that you can assign to this property. You can use e.g. 8192 (SYSTEM_UI_FLAG_LIGHT_STATUS_BAR) to set
+        the statusbar text color to a dark color.
```
hansemannn, line 1545:
```suggestion
        See [WindowManager.LayoutParams](https://developer.android.com/reference/android/view/View#setSystemUiVisibility\(int\)) for a
        list of additional flags that can be assigned to this property. For example, you can use `8192` (SYSTEM_UI_FLAG_LIGHT_STATUS_BAR) to set
        the status bar text color to a dark color.
```

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/WindowProxy.java
m1ga: I wasn't sure if I should add one as I only use it in this place. But I can do it to make it be inline with the other values :+1:

hansemannn (approved):

PR #14168, opened 2025-01-05
fix(build): restore old scons dependencies to fix local Xcode build

hansemannn:
Wow, this took me a bit to debug. So here is what's going on: https://github.com/tidev/titanium-sdk/commit/48afddf5652602ae2f8cc376d9a37156c828d9c1 updated several dependencies (good!), but broke the local iOS build via Xcode that is used to debug the iOS source code.

The reason why it broke is that we have a pre-compile script to generate the ti.kernel and index.json that are need for internal bootstraping of the iOS source. This generator uses `ssri` and `fs-extra` to do some file i/o. The updated packages introduced optional chaining that is not supported by the Xcode runtime, so it broke the pre-compiler. Restoring these two dependencies to their initial values solved it, while we check for a solution that can handle optional chaining in the future.

Screenshot 2025-01-05 at 13 37 03


m1ga (approved):
LGTM. I'll add it to the https://github.com/tidev/titanium-sdk/issues/13781 list so we don't update it for now :+1:

cb1kenobi (approved):
Wow, nice find!

hansemannn:
Merci!

PR #14169, opened 2025-01-08
fix(android): tableview content update fix

m1ga:
## Issue
A user reported an issue on Slack that a label inside a tableview is not updating when setting a new value.
The event fix in https://github.com/tidev/titanium-sdk/pull/13826 seems to be the culprit. While it fixed the event issue it introduced the label update issue.

## Solution
After a few tests it looks like we only have to set/update the parent IF it already had one (when a user reassigns the data) and not if it stays the same data (just updating the label).

I've added a simple null check and update only if there is already a parent. This way both tests (event and label) work fine:

## Test 1:

see https://github.com/tidev/titanium-sdk/pull/13826 for more details

```js
var menu = [];
var numrows = 0;

  if (e.index == numrows - 1)
    AddRows();
});

function AddRows() {
  if (menu.length == 0)
    row.add(lab);
  menu.push(row);
  numrows++;
  menu.push(row);
  numrows++;
  table.data = menu;
};

AddRows();
self.add(table);
self.open();
```

* start and click the last row a few times
* then click any other row and see if the click event still works

## Test 2
```js
var tableData = [];

for (var i=1; i<=10; i++){
    className: 'forumEvent', // used to improve table performance
	height: 50
  });
  var labelUserName = Ti.UI.createLabel({
    text: 'Fred Smith ' + i,
	left: 10
  });
  row.add(labelUserName);
  tableData.push(row);
}

  data: tableData
});

win.add(tableView);
win.open();
tableView.data[0].rows[0].children[0].text = "test"
```

* start and see if the first label says "test"

--

## Test 3
```js

function AddRows() {
	const rows = [];
			title: "Row " + i,
			color: 'black',
			height: 100
		});

		row.add(createRowView());
		
		rows.push(row);
	}
	
	table.data = rows;
};

function createRowView() {
		height: 80,
		width: 120
	});
	
	setTimeout(() => {
	}, 1000);

	return imageView;
}

self.add(table);
self.open();
AddRows();
```
* images should show up

Currently tested by a Slack user: Feedback was: it's working

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
@@ -189,8 +189,10 @@ public class TableViewProxy extends RecyclerViewProxy
 				row.getProperties().optString(TiC.PROPERTY_FOOTER_TITLE,
 					row.getProperties().getString(TiC.PROPERTY_FOOTER)));
 
+			if (row.getParent() != null) {
+				row.setParent(section);
+			}
 			// Add row to section.
-			row.setParent(section);
 			section.add(row);
 		}
 


prashantsaini1:
As mentioned [in important notes #1](https://github.com/tidev/titanium-sdk/pull/13826#issuecomment-2578590919), these issues are not valid ones since this is not how the `TableView's data` property should be used. `update/append/delete` methods are necessary to alter views since they internally triggers model-data and native view updates.

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
```diff
@@ -189,8 +189,10 @@ private void appendRowInternal(Object rows, KrollDict animation, boolean interna
 				row.getProperties().optString(TiC.PROPERTY_FOOTER_TITLE,
 					row.getProperties().getString(TiC.PROPERTY_FOOTER)));
 
+			if (row.getParent() != null) {
```
prashantsaini1, line 192: Please remove this conditional code as it would cause more issues because it's changing the parent-section of a row from one to another without properly removing the row from its previous parent-section first.

prashantsaini1 (changes requested):

prashantsaini1:
@m1ga The issue reported by Slack user can be resolved via existing methods by properly using TableView. This PR can simply revert the previous changes. 😉

android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java
m1ga: But then we have the same code as before and it won't behave the same way it did in 12.5.1 where you were able to set content without running `update`.  Can you create a PR with a proper solution for both cases

PR #14170, opened 2025-01-08
fix(android): fix tableView parent

m1ga:
Reverting https://github.com/tidev/titanium-sdk/pull/13826


m1ga:
More testing tomorrow

prashantsaini1 (approved):
Reverting the change is correct as people who reported issues earlier were actually using TableView differently than it should be. We should stick to use existing methods to update content as it is the correct way to update proxies internally.

PR #14172, opened 2025-01-20
feat(android): layerType property for WebView

m1ga:
## Issue
Capturing a screenshot of a webview with local content (e.g. GraphJS) will capture an empty webview. External URLs work fine.
iOS does capture local content without issues.

## Solution
Setting the WebView into software rendering mode will make the data appear again. Previously it was done by setting a border or transparent background. As these methods don't work and would lead to visual differences I've added a `.layerType = 0|1|2` property. 

## Test

```js
var btn = Ti.UI.createButton({title:"click"});
var img = Ti.UI.createImageView({bottom:0,height:200,borderWidth:1,borderColor:"red",width: 100});
var html =`<div style="height:150px">
	<canvas id="myChart"></canvas>
</div>

<script src="https://cdn.jsdelivr.net/npm/chart.js"></script>

<script>
  const ctx = document.getElementById('myChart');

  new Chart(ctx, {
    type: 'bar',
    data: {
      labels: ['Red', 'Blue', 'Yellow', 'Green', 'Purple', 'Orange'],
      datasets: [{label: '# of Votes',data: [12, 19, 3, 5, 2, 3],borderWidth: 1}]
    },
    options: {
      animation: {duration: 0},
      scales: {y: {beginAtZero: true}}
    }
  });
</script>`;

var webview = Ti.UI.createWebView({top: 0, height: 200, layerType:Ti.UI.WebView.LAYER_TYPE_SOFTWARE, html:html});

console.log(Ti.UI.WebView.LAYER_TYPE_NONE);
console.log(Ti.UI.WebView.LAYER_TYPE_SOFTWARE);
console.log(Ti.UI.WebView.LAYER_TYPE_HARDWARE);

win.add([btn, img,webview]);
win.open();
win.addEventListener("open", function(){
	// webview.layerType = Ti.UI.WebView.LAYER_TYPE_SOFTWARE; // or here
	console.log(webview.layerType)
})
btn.addEventListener("click", function(){
	console.log(webview.layerType)
	Ti.Media.takeScreenshot(function(e){
		img.image = e.media;
	});
})
```

## Workaround
Using Hylerloop to set it:
```js
const WebView = require("android.webkit.WebView");
const View = require("android.view.View");

// this should be in open or your screenshot method
var webView = new WebView($.webview);
webView.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -245,22 +245,22 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 	}
 
 	@Kroll.setProperty
-	public void setSoftwareMode(Boolean value)
+	public void setLayerType(int value)
 	{
 		TiUIWebView currWebView = getWebView();
 		if (currWebView != null) {
-			currWebView.setSoftwareMode(value);
+			currWebView.setLayerType(value);
 		}
 	}
 
 	@Kroll.getProperty
-	public boolean getSoftwareMode()
+	public int getLayerType()
 	{
 		TiUIWebView currWebView = getWebView();
 		if (currWebView != null) {
-			return currWebView.softwareMode;
+			return currWebView.layerType;
 		}
-		return false;
+		return 0;
 	}
 
 	@Kroll.setProperty
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -15,6 +15,7 @@ import android.print.PrintAttributes;
 import android.print.PrintDocumentAdapter;
 import android.print.PrintManager;
 import android.util.DisplayMetrics;
+import android.view.View;
 import android.webkit.ValueCallback;
 import android.webkit.WebView;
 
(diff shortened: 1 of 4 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -244,6 +244,25 @@ public HashMap getRequestHeaders()
 		return new HashMap<String, String>();
 	}
 
+	@Kroll.setProperty
+	public void setSoftwareMode(Boolean value)
```
hansemannn, line 248: Suggestion: Use `softwareModeEnabled` to better indicate the boolean type

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -1070,4 +1076,18 @@ protected void disableHWAcceleration()
 	{
 		Log.d(TAG, "Do not disable HW acceleration for WebView.", Log.DEBUG_MODE);
 	}
+
+	public void setSoftwareMode(Boolean value)
+	{
+		WebView webView = getWebView();
+		if (webView != null) {
+			softwareMode = value;
```
hansemannn, line 1084: Is there a reason to save it as a local variable?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
```diff
@@ -1070,4 +1076,18 @@ protected void disableHWAcceleration()
...
+			if (value) {
+				webView.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
+			} else {
+				webView.setLayerType(View.LAYER_TYPE_HARDWARE, null);
```
hansemannn, line 1088: There are three layer types (LAYER_TYPE_HARDWARE, LAYER_TYPE_SOFTWARE and LAYER_TYPE_NONE), so a boolean may not be the best fit here

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
m1ga: I use that in the getter to return it

android/modules/ui/src/java/ti/modules/titanium/ui/widget/webview/TiUIWebView.java
m1ga: I've change it to be `webview.layerType` and added constants for all three options. Will update the example in the first post in a bit

PR #14173, opened 2025-01-20
feat: TabGroup show/hide support, iOS 18+ elevated tab bar

prashantsaini1:
**Goal:** Show/hide tab-bar programmatically, especially on iOS 18+ [elevated tab bar](https://developer.apple.com/documentation/uikit/elevating-your-ipad-app-with-a-tab-bar-and-sidebar) which obstructs custom nav bar on iPad apps with a TabGroup.

**Properties**
1. `tabBarVisible`: show/hide the tab-bar, now on iOS 18+ also.
   - Default animation is disabled. Animations usually should not run on property updates, like `view.visible` on Android.
   - To enable animation, we can now use new `showTabBar() / hideTabBar()` methods.

2. `enabled`: **Android** **only**: enable/disable the swipe and click events on Tabs.
   - Default is `true`. This has been decoupled from the previous `disableTabNavigation` to allow disabling `TabGroup` interactions without hiding it actually. To hide the `TabGroup`, new properties/methods can be used.
   - Also fixed a bug where Tabs click are active despite the tab-navigation is disabled.

**Methods**
1. `Ti.UI.TabGroup.showTabBar()`: shows the tab bar with animation.
2. `Ti.UI.TabGroup.hideTabBar()`: shows the tab bar with animation.

**Removed methods:** (These methods do not work programmatically)
1. `Ti.UI.Window.showTabBar()`
4. `Ti.UI.Window.hideTabBar()`


hansemannn:
@prashantsaini1 Please finish this PR

prashantsaini1:
@m1ga `disableTabNavigation` now takes care of the conflicted property `enabled`. Can you please check and commit the required updates as per the new experimental branch? Thanks!

prashantsaini1:
@m1ga @hansemannn This PR is ready now.

apidoc/Titanium/UI/Window.yml
```diff
@@ -535,15 +535,6 @@ methods:
...
     platforms: [iphone, ipad, macos]
 
-  - name: hideTabBar
-    summary: Hides the tab bar. Must be called before opening the window.
-    description: |
-        To hide the tab bar when opening a window as a child of a tab, call
-        `hideTabBar` or set `tabBarHidden` to `true` **before** opening the window.
-
-        If the window is not a child of a tab, this method has no effect.
-    platforms: [iphone, ipad, macos]
```
hansemannn, line 545: This can't just be removed. Please find a way to either make it backwards compatible + warning log or keep it and add an error log that it's not available anymore. But then, we cannot release it before Ti SDK 13.0.0

iphone/Classes/TiUITabGroup.m
```diff
@@ -370,20 +370,16 @@ - (void)tabBarController:(UITabBarController *)tabBarController didEndCustomizin
...
+}
 
-  if ([self tabBarIsVisible] == visible) {
-    return;
+- (void)hideTabBar:(BOOL)hidden animated:(BOOL)animated
+{
+  if (@available(iOS 18.0, *)) {
+    [self.tabController setTabBarHidden:hidden animated:animated];
+  } else {
+    self.tabController.tabBar.hidden = hidden;
```
hansemannn, line 381: The version prior to iOS 18 should remain as is, honoring the animation if set

hansemannn (changes requested):
Some iOS feedbacks - maybe @m1ga can check Android?

android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
```diff
@@ -91,27 +91,24 @@ public int getTabIndex(TabProxy tabProxy)
 		return tabs.indexOf(tabProxy);
 	}
 
-	@Kroll.method
-	public void disableTabNavigation(Object params)
+	@Kroll.setProperty
+	public void setEnabled(boolean enabled)
```
m1ga, line 95: should we add a deprecation period here? I saw that you've put 12.7.0 as deprecated and removed in the doc. 
As it was added two years ago to have `animated` as a parameter this change will remove that functionality.

iphone/Classes/TiUITabGroup.m
prashantsaini1: I remember it didn't work as expected on iOS 18 which is why I fixed it this way.

android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
prashantsaini1: @m1ga I think a deprecation period until 13.0.0 is fine in this PR for both platforms as @hansemannn suggested. I will make the further changes.

iphone/Classes/TiUITabGroup.m
prashantsaini1: I checked it again. It moves the TabBar up from the bottom edge and immediately down to the bottom-edge, making the animation look like a glitch and not hiding the TabBar.

Therefore, I created two new methods to show/hide the TabBar with animation, and made this property `tabBarVisible` to perform no animation (can be used at creation time as well which is not possible without this PR).

apidoc/Titanium/UI/Window.yml
prashantsaini1: You're right as I checked it again and found it working fine to temporary hide the TabBar per window basis, except not effective on elevated tab-bar on iPad, which is fine as elevated tab-bar can be tackled via new additions in this PR. I will revert this property for Window.

android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
prashantsaini1: @m1ga I brought back the `disableTabNavigation()`. Internally using new APIs to keep same effects.

iphone/Classes/TiUITabGroup.m
hansemannn: But the fallback is for < iOS 18, so these users should have the exact same behavior (animated using the previous code)

prashantsaini1: I meant that the same animation effects can be achieved by using new APIs on all iOS versions now. Checking it again for < iOS 18 as it wasn't fully functional there.

m1ga:
**Android:**

I've created some test code to test the new functions:

```js
	title: 'Show / hide tab group',
	top: 0
});
const btn2 = Ti.UI.createButton({
	title: 'enable / disable',
	top: 50
});
let isVisible = false;

	// tabGroup.tabBarVisible = !tabGroup.tabBarVisible;
	if (isVisible) {
		tabGroup.hideTabBar()
	} else {
		tabGroup.showTabBar()
	}
	isVisible = !isVisible;
});
btn2.addEventListener('click', () => {
	tabGroup.enabled = !tabGroup.enabled;
});

window.add([btn, btn2]);

let tabGroup = Ti.UI.createTabGroup({
	tabBarVisible: false,
	tabs: [
		Ti.UI.createTab({
			window,
			title: "tab1"
		}),
		Ti.UI.createTab({
			window: Ti.UI.createWindow(),
			title: "tab2"
		})
	]
});

tabGroup.open();
```

There is one minor Android issue: when it is hidden by default and then you click "show" it won't run the animation. If you hide and show it again it will.

Rest is looking good: showTabBar(), hideTabBar()  and enabled work.

prashantsaini1:
@m1ga Thanks for the test cases.
- This is a known issue to me already as the native `view` height is not available until the window is opened and cannot set `translationY` onto it which is why the show() call does not run animation for first time.
- It's doable though by storing the height for the native `view` when it's created and then apply `translationY` again.

I'll give it a try again and see if it works!

iphone/Classes/TiUITabGroup.m
prashantsaini1: @hansemannn Checked it further in 12.6.3.GA. I suggest that we keep the animation behaviour as is with a deprecation warning in 12.7.0 and remove in 13.0.0. There's one more issue I found that this property does not reset the window insets like `Window.tabBarHidden` does, will fix it as well.

prashantsaini1: @hansemannn It's done now.

m1ga:
https://github.com/tidev/titanium-sdk/pull/14190 has the changes for the `experimental` BottomNavigation. I've kept it separate so you can finish this one without any changes by me.

prashantsaini1:
This PR is finally complete from my end. Latest updates:
- The known bug in Android is fixed where `showTabBar()` animation does not work if tab-bar is initially hidden at creation time using `tabBarVisible`.
- Did some code optimisations and refactoring.
- @m1ga I also added all these features in the new experimental `TiUIBottomNavigation` API as well, so no updates are required in this [other PR](https://github.com/tidev/titanium-sdk/pull/14190) now.
   - Storing `mBottomNavigationHeightValue` is not required in this experimental API as the height is captured dynamically now. 
   - I tested all use cases in new API similar to the existing `TabGroup's style` and all works fine. Please feel free to verify.

apidoc/Titanium/UI/TabGroup.yml
```diff
@@ -199,6 +199,10 @@ methods:
         type: disableTabOptions
     platforms: [android]
     since: 3.6.0
+    deprecated:
+        since: "12.7.0"
+        removed: "13.0.0"
```
hansemannn, line 204: We usually only add the removed tag once the release that removed it is actually out, so I'd suggest to remove it and set a reminder to when master will become 13.0.0 (we can plan it as the next release after the 12.7.0 release). Then, you can do a PR with the removal itself and the docs for it.

iphone/Classes/TiUITabGroup.m
```diff
@@ -370,25 +370,52 @@ - (void)tabBarController:(UITabBarController *)tabBarController didEndCustomizin
 
 - (void)setTabBarVisible_:(id)value
 {
-  BOOL visible = [TiUtils boolValue:value];
+  bool isIOSVersionLowerThan18 = [TiUtils isIOSVersionLower:@"18.0"];
```
hansemannn, line 373: The proper ObjC type is `BOOL`, not the c type `bool`. Also rename it to `isiOS17OrLower`

iphone/Classes/TiUITabGroup.m
```diff
@@ -370,25 +370,52 @@ - (void)tabBarController:(UITabBarController *)tabBarController didEndCustomizin
...
+  if (isIOSVersionLowerThan18) {
+    DEPRECATED_REPLACED_REMOVED(@"UI.TabGroup.tabBarVisible's animation on iOS <= 17 is",
```
hansemannn, line 375: Use `[...] animation on iOS < 18 [...]`

iphone/Classes/TiUITabGroup.m
```diff
@@ -370,25 +370,52 @@ - (void)tabBarController:(UITabBarController *)tabBarController didEndCustomizin
...
+        @"12.7.0", @"13.0.0",
+        @"UI.TabGroup.showTabBar/hideTabBar methods.");
+  }
+
+  [self hideTabBar:![TiUtils boolValue:value def:YES] animated:isIOSVersionLowerThan18];
+}
+
+- (void)hideTabBar:(BOOL)hidden animated:(BOOL)animated
+{
+  // iOS 18+
```
hansemannn, line 385: Remove this, the statement below is pretty self-describing

iphone/Classes/TiUITabGroup.m
```diff
@@ -370,25 +370,52 @@ - (void)tabBarController:(UITabBarController *)tabBarController didEndCustomizin
...
+  if (@available(iOS 18.0, *)) {
+    [self.tabController setTabBarHidden:hidden animated:animated];
+    return;
+  }
+
+  // iOS 17 or below without animation.
```
hansemannn, line 391: Same here - it's self-describing

hansemannn (changes requested):
Last iOS changes, thank you!

m1ga (approved):
Android looking good :+1: Features work fine in the normal BottomNavigation and the new one and with the normal TabGroup.

hansemannn (approved):

PR #14176, opened 2025-01-30
fix(tests): update canOpenURL test

m1ga:
Adds the <queries> to the manifest to make the `canOpenURL()` test work again and switching to https.

That brings us to:
`5169 Total Tests: 5165 passed, 4 failed, 0 skipped.`

two failing tests are locale date related ("Date" and "Intl.DateTimeFormat"), one is a notification test ("Notifications enabled by default") and the last one is a `android:resource://` ("Titanium.Filesystem.File.constructed via URIs") issue.


prashantsaini1 (approved):
Looks good to me!

PR #14178, opened 2025-02-06
feat(android): textTransform for Label

m1ga:
Adds CSS like `textTransform` with `uppercase`, `lowercase` and `none` (default):

```js
var toggle = false;
var win = Ti.UI.createWindow({layout: "vertical"});
var lbl1 = Ti.UI.createLabel({ text: "Just a Text wiTH ranDOm leTTerS (random)"});
var lbl2 = Ti.UI.createLabel({ text: "Just a Text wiTH ranDOm leTTerS (uppercase)",
  textTransform: "uppercase"
});
var lbl3 = Ti.UI.createLabel({ text: "Just a Text wiTH ranDOm leTTerS (lowercase)",
  textTransform: "lowercase"
});
var btn = Ti.UI.createButton({ title: "toggle"});

  if (!toggle) {
		lbl1.textTransform = "lowercase";
		lbl2.textTransform = "none";
		lbl3.textTransform = "uppercase";
	} else {
		lbl1.textTransform = "none";
		lbl2.textTransform = "uppercase";
		lbl3.textTransform = "lowercase";
	}
	toggle = !toggle;
})
win.add([lbl1, lbl2, lbl3, btn]);
win.open();
```

Screenshot_20250206-113228

**Note:**
I've used a string here and not constants. That way you can use the normal CSS properties instead of long Ti constants. I can change it but it feels a bit more intuitive to use it like this.


hansemannn:
Is there a reason to do that via the native proxy instead of in JS directly? The proxy round trip will likely not be as performant as handling it via `toUppercase()` and `toLowerCase()` IMHO

m1ga:
yes, it works with Alloy textid strings so you can define e.g. uppercase before and then assign it.

hansemannn:
iOS is done already! Is this as expected?

https://github.com/user-attachments/assets/75ec3cfc-a8a6-4f48-9d53-67389f59a493

m1ga:
Interesting :thinking: I can't find anything when searching for `textTransform` in the code. Is it because it's the same name in CSS?

m1ga:
oh, I didn't see you've added it into my PR :smile: I thought you meant it's in the current SDK already. I'm recovering from a bad flu at the moment, so my brain is not functioning 100% at the moment. But yes, the video looks fine :+1:

hansemannn:
Get well soon! Then I think it's ready to go

PR #14181, opened 2025-02-09
feat(android): target Android SDK level 35

### LINKED ISSUE #14111: Android: Changes for next Android API level 35 (targetSdkVersion 35)

### Description

We'll need to check our windows/statusbar and test it before we change to Android API level 35 (next year or after that). Android API 35 will enforce "edge to edge" design. See https://www.youtube.com/watch?v=YKPqjsYBFvI&t=67s

You can test it already if you use `<uses-sdk  android:targetSdkVersion="35" />` and test it with a TabGroup.

### Solution

Check the SDK and follow the guides from the video or https://developer.android.com/about/versions/15/behavior-changes-15

There is also an opt-out property for the  `edge-to-edge` mode: https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement that we can put into a theme `<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>`

### Workaround:

Add:
```
android {
	defaultConfig {
		compileSdk 35
  }
}
```
to your `[app]/platform/android/build.gradle` and then create `[app]/platform/android/res/values-v35/values.xml` and add:
```xml
<resources>
	<!-- Todo remove windowOptOutEdgeToEdgeEnforcement after implementing Edge-to-edge -->
	<style name="Base.Theme.Titanium.Light" parent="Theme.MaterialComponents.Light.DarkActionBar">
		<item name="android:windowLightStatusBar">false</item>
		<item name="android:windowLightNavigationBar">false</item>
		<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
	</style>

	<style name="Base.Theme.Titanium.Dark" parent="Theme.MaterialComponents">
		<item name="android:windowLightStatusBar">false</item>
		<item name="android:windowLightNavigationBar">false</item>
		<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
	</style>
</resources>
```

### Platforms

Android

m1ga:
First updates for targetSDK 35 (fixes https://github.com/tidev/titanium-sdk/issues/14111).

As mentioned in https://github.com/tidev/titanium-sdk/issues/14111 Android SDK level 35 will enable edge-to-edge by default. This will break a lot of our default themes:

Bildschirmfoto_20250209_202720
_(see the top of the window)_

This PR will enable targetSDK 35 but will opt-out of edge-to-edge. This way we will have more time to implement proper edge-to-edge support and the apps will looks the same as before for everyone. It's not announced when targetSDK 35 will be required but according to https://targetsdk.com/ it will be end of this year (new apps, app updates a year later as always)

With this PR it will look like this again:
Bildschirmfoto_20250209_202653

`suppressUnsupportedCompileSdk` is added to remove the warnings during build as the gradle version needs to be updated (different PR).

**TODO**
Run more tests and see if anything else needs to be implemented for targetSDK 35


cb1kenobi (approved):
Reviewed, but not tested. Looks good.

hansemannn:
@prashantsaini1 Can you please give this a test with our app?

m1ga:
Android 16 was just released so you can test it on Pixel devices now too. At the moment most 3rd party apps opt-out of the edge-to-edge design (even most Google apps) too.

Long term goal of course: add a safe padding around the top and bottom so it will layout the windows correctly in edge-to-edge mode.

m1ga:
a small update to prepare the XML layouts we have. They won't be used because of the opt-out part but when we implement the correct way these windows will work already.

Without the XML fix and edge-to-edge:

| DrawerLayout: | BottomNavigation (experimental:true):  |
|---|---|
| Screenshot_20250612-124753  | Screenshot_20250612-152728  |


With the XML fix and edge-to-edge:

| DrawerLayout: | BottomNavigation (experimental:true):  |
|---|---|
| Screenshot_20250612-124734  | Screenshot_20250612-153326 |

cb1kenobi:
We should get this merged. I have re-reviewed it, but still haven't tested it. :(

When the description says "opt-out of edge-to-edge", how do I opt-in to test this?

m1ga:
you add https://github.com/tidev/titanium-sdk/pull/14181/files#diff-308f820e2c4773f2a8eb23b7c16335e9878e01a11863c33d6b9fd05b44aad626 to your project (/app/platform/android/res/) and set it to `false`

m1ga:
or run 12.7.0 and set `<uses-sdk  android:targetSdkVersion="35" />` in your tiapp.xml (inside `<manifest>`)

prashantsaini1:
@m1ga Can you check if this entire stuff can be done by simply calling the `WindowCompat.setDecorFitsSystemWindows(getWindow(), true);` in `TiBaseActivity` as [described here](https://developer.android.com/develop/ui/views/layout/edge-to-edge-manually#lay-out-in-full-screen)?

This would be helpful to cover all scenarios if we missed anything somewhere.

Edit: Though this PR disables the edge-to-edge feature and saves current apps, however, this may soon be required to enable more immersive UX. I can take a look on it this weekend to find a proper solution to enable/disable it and automatically handle the systems colors as also described in above link.

m1ga:
No, that doesn't work:
Screenshot_20250612-173904

(I'm testing other stuff at the moment from the video in the issue ;-) )

> Though this PR disables the edge-to-edge feature and saves current apps, however, this may soon be required to enable more immersive UX

True, but currently it isn't and target 35 will be required in November. That is the point of this PR: opt-out so people can make updates and their apps look the same. But totally agree: we have to implement it

prashantsaini1:
Can it wait a couple of weeks to merge? I haven't tested it yet with our apps, so if it can wait, I can prioritise to wrap it up asap.

m1ga:
sure, we didn't switch to 35 yet and no user has to do it by hand. If they want they have the workaround in the issue.

I think this one is more important: https://github.com/tidev/titanium-sdk/pull/14226 so we can start updating all the modules to be compatible. The earlier we can start the better. Many people don't like to update modules :smile: They always run very old versions when they ask for support on slack.

m1ga:
emails are rolling out with:

> Starting August 31 2025:
> New apps and app updates must target Android 15 (API level 35) or higher to be submitted to Google Play; except for Wear OS, Android Automotive OS, and Android TV apps, which must target Android 14 (API level 34) or higher.

but you can extend until November

hansemannn:
cc @prashantsaini1 Can you finish the review this sprint? Then we can release a new Ti SDK release mid July ✨

android/titanium/res/values-v35/values.xml
prashantsaini1, line 1: @m1ga This breaks multiple themes and also some alignment issues. We need to redefine only two themes for `values-35` to make rest of the others inherit properly:
```
<resources>
	<!-- Todo remove windowOptOutEdgeToEdgeEnforcement after implementing Edge-to-edge -->
	<style name="Base.Theme.Titanium.Light" parent="Theme.MaterialComponents.Light.DarkActionBar">
		<item name="android:windowLightStatusBar">false</item>
		<item name="android:windowLightNavigationBar">false</item>
		<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
	</style>

	<style name="Base.Theme.Titanium.Dark" parent="Theme.MaterialComponents">
		<item name="android:windowLightStatusBar">false</item>
		<item name="android:windowLightNavigationBar">false</item>
		<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
	</style>
</resources>
```

m1ga: Can you show an example? I've tested it with two apps (already in the store with this PR) and it worked fine and when I run
```js
    "theme": "Theme.Titanium.DayNight.Solid"
});
var lbl = Ti.UI.createLabel({top:0, text:"top"});
win.add(lbl);
win.open();
```
with all of the changed themes it looks correct too. The label is always visible

prashantsaini1 (approved):
Looks good!

hansemannn (approved):

PR #14185, opened 2025-02-17
fix(android): prevent ANR Bad Behaviour warnings from Google

prashantsaini1:
**This PR attempts to fix an old issue with Titanium SDK where the `Thread.setDefaultUncaughtExceptionHandler` in `TiApplication.onCreate` prevents uncaught exceptions to be handled by system, leading to ANRs.**

### Side effect:
- This PR only prevents ANRs happening due to main-thread waiting, but not the underlying code which causes it, so this will likely turn such ANRs into crashes now which is expected in case of uncaught exceptions. 
- Crash reports are always more verbose in identifying the root cause of issue than ANR reports.

In order to improve error output in [this commit](https://github.com/tidev/titanium-sdk/commit/49a0998b640d7fd5d3adb80ad4f08c6319ea9378#diff-78de726bd2715a7fb5f52e6c82dd6a1cc6736ca7d9cc67a2fdad5cb0defe5580L365), the delegate call got missed to handle uncaught exceptions by Android system, which eventually turns into ANR than likely crashing the app. (A crash event seems better than ANR in uncaught exceptions scenarios).

An excerpt from Firebase ANR reports:

> The main thread has finished processing events and is shutting down the app.
As part of this process, it waits for all other threads to finish.  If there is a thread that does not finish, it is possible that the app will show an ANR as the system considers it to still be active.
A common cause for this type of ANR is installing a new uncaught exception handler that does not delegate to the previous handler. Failure to call back into the previous handler will cause an ANR.

[Another link which describes similar problem](https://ashishb.net/programming/android-uncaught-exception-handler-and-anr/), there are more such reports out there.

- The JS side errors will still show same error dialog as they are caught by V8 Runtime and not delegated to Android main thread.
- `ti.crashlytics` module should also work as it relies on `KrollRuntime.dispatchException` method call before delegating the exception to system. (I will double verify it further - `verified`).

### DIFF
--- a/android/runtime/common/src/java/org/appcelerator/kroll/KrollRuntime.java
+++ b/android/runtime/common/src/java/org/appcelerator/kroll/KrollRuntime.java
@@ -538,7 +538,9 @@ public abstract class KrollRuntime implements Handler.Callback
 			}
 
 			// Handle exception with defaultExceptionHandler
-			instance.primaryExceptionHandler.handleException(exceptionMessage);
+			if (instance.primaryExceptionHandler != null) {
+				instance.primaryExceptionHandler.handleException(exceptionMessage);
+			}
 		}
 	}
 
--- a/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java
+++ b/android/titanium/src/java/org/appcelerator/kroll/KrollProxy.java
@@ -1253,7 +1253,7 @@ public class KrollProxy implements Handler.Callback, KrollProxySupport, OnLifecy
 				}
 			}
 		} catch (Throwable t) {
-			Thread.getDefaultUncaughtExceptionHandler().uncaughtException(null, t);
+			TiApplication.handleInternalException(t);
 		}
 
 		return false;
--- a/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiApplication.java
@@ -95,6 +95,7 @@ public abstract class TiApplication extends Application implements KrollApplicat
 	private String defaultUnit;
 	private BroadcastReceiver localeReceiver;
 	private AccessibilityManager accessibilityManager = null;
+	private UncaughtExceptionHandler nativeExceptionHandler = null;
 
 	protected TiDeployData deployData;
 	protected ITiAppInfo appInfo;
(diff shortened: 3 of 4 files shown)


prashantsaini1:
I just finished testing the [Firebase Crashlytics](https://github.com/hansemannn/titanium-crashlytics/tree/main) module and the PR already takes care of it.

Firebase Crashlytics SDK adds its own exception handler which internally propagates to Android Runtime handler on uncaught exceptions but they were not previously delegated, which is what this PR does now.

<hr>

The [last commit](https://github.com/tidev/titanium-sdk/pull/14185/commits/25a2d084ffc98e8d1eff5f333ad4919b452a54da) in this PR also improves Firebase reporting for `uncaught fatal exceptions`, which previously may have skipped due to `null` thread, or probably even reported as non-fatal exceptions as in following code snippet from [ti.crashlytics module here](https://github.com/hansemannn/titanium-crashlytics/blob/ef9309ca51363544fc06d056da1df09453ffefee/android/src/ti/crashlytics/TitaniumCrashlyticsModule.java#L156).

```
public static void onAppCreate(TiApplication app) {
		KrollRuntime.setPrimaryExceptionHandler(new TiExceptionHandler(){
			@Override
			public void handleException(ExceptionMessage error)
			{
				Throwable th = new Throwable();
				th.setStackTrace(generateStackTrace(error.javaStack, error.jsStack, error.sourceName, error.message, error.lineSource, error.line));
				FirebaseCrashlytics.getInstance().recordException(th);
				super.handleException(error);
			}
		});
	}
```

With this PR, Firebase Crashlytics module should now report uncaught exceptions and delegate them to Android Runtime for internal handling, which further close the problematic thread, kill and restart the app.

prashantsaini1:
@m1ga I would appreciate your thoughts on this if any.

m1ga:
Sounds great! I didn't know that these ANR where just a "waiting" issue before we can get the error/crash. Guess that will help a lot of people to find out what the real issue is. 

SDK builds fine and I quickly tested one of my app and it still builds and runs fine. No errors in there so I didn't test that feature yet :smile:

m1ga (approved):
Code changes look good.

Tested Kitchensink, one other normal app and the test suite. All still run fine. 
Crashlytic error still shows up and looks the same.

Didn't have an ANR example that I could test, so I couldn't test the new error log.

prashantsaini1:
This does not effectively fix the actual ANRs, but rather a situation as mentioned in description:
```
The main thread has finished processing events and is shutting down the app.
As part of this process, it waits for all other threads to finish. If there is a thread that does not finish, it is possible that the app will show an ANR as the system considers it to still be active.
A common cause for this type of ANR is installing a new uncaught exception handler that does not delegate to the previous handler. Failure to call back into the previous handler will cause an ANR.
```

This situation happens when some uncaught errors happen on non main threads but Android tries to restart main-thread which makes it waiting for the actual error thread to finish up first. Firebase Crashlytics SDK also does the same, it delegates back to Android Runtime handler, but Titanium SDK does not. So I am quite positive that this would fix such indirect ANRs and creates actual crashes.

hansemannn (approved):

prashantsaini1:
@m1ga @hansemannn I confirm that this PR works as intended. We are now able to see more closer details in tracing earlier ANRs which are now converting into crashes.

PR #14187, opened 2025-03-05
fix(android): android modules with timodule.xml manifest entry breaks since 12.6.x

prashantsaini1:
Since this [Gradle-8 PR](https://github.com/tidev/titanium-sdk/pull/14014), modules with their own manifest in `timodule.xml` are breaking due to:
1. Missing merge process of `timodule.xml manifest` entry in main manifest.
2. Not disabling `android:nonTransitiveRClass` for module build, but disabling for app build.

### DIFF
--- a/android/cli/commands/_buildModule.js
+++ b/android/cli/commands/_buildModule.js
@@ -496,6 +496,7 @@ AndroidModuleBuilder.prototype.generateRootProjectFiles = async function generat
 	// Create a "gradle.properties" file. Will add network proxy settings if needed.
 	const gradleProperties = await gradlew.fetchDefaultGradleProperties();
 	gradleProperties.push({ key: 'android.useAndroidX', value: 'true' });
+	gradleProperties.push({ key: 'android.nonTransitiveRClass', value: 'false' });
 	gradleProperties.push({
 		key: 'org.gradle.jvmargs',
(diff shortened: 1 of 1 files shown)


hansemannn (approved):
It was properly tested with our app, so I can at least provide a FT-pass. If we can get a CR on this as well, it should be fine. I will then prepare a 12.6.3 already, which includes this fix and #14185

m1ga (approved):
* modules aar contains a filled r.txt file and a proper androidmanifest.xml again :+1:
* module build fine, push arrived

But I have to say: 
on my Linux machine I had working modules + apps even without this change. So I can't really test it as I don't have the broken case to verify.

prashantsaini1:
@m1ga For a test-case, try to verify if the `timodule.xml manifest` entry is there in app's build manifest with and without this change.

However, if these entries are added manually in `tiapp.xml`, then it will work fine for obvious reasons which this PR tackles automatically.

I did a diff check and found the entry missing from `firebase.cloudmessaging` module. This is why the not-working module wasn't able to receive PN in our app, but weirdly receives it via Firebase console (I guess Firebase has a different service in OS for this internally and route notification via OS itself).

Screenshot 2025-03-06 at 1 50 44 AM

PR #14189, opened 2025-03-17
fix(android): lifecycle listeners and states

prashantsaini1:
Fixes:
1. Save activity instance state after further states are updated.
2. Add support to remove various listeners from proxies to avoid holding its activity reference and calling them when a proxy is removed.

Further fixes to come:
1. Some internal views like a WebView holds activity reference after being removed and still be able to call its callbacks repeatedly when added/removed multiple times. (check below screenshot).
   - Though I could not find any memory leaks as these are all WeakReference, but it's still a safe check to remove their listeners once the relevant view-proxy is removed.

Screenshot 2025-03-17 at 2 44 40 PM

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -1670,8 +1670,6 @@ public abstract class TiBaseActivity extends AppCompatActivity implements TiActi
 	@Override
 	protected void onSaveInstanceState(Bundle outState)
 	{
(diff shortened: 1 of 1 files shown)


prashantsaini1:
@m1ga Can you please review this?

m1ga:
Any special stuff I should test?

SDK builds fine and my current app runs fine, not seeing any changes. Code changes itself look fine

m1ga (approved):

hansemannn:
@m1ga Can I merge this?

m1ga:
yes :+1:

PR #14191, opened 2025-03-22
feat(android): parity for Ti.UI.View clipMode

m1ga:
Based on https://github.com/mbender74/clipview/tree/main/ti.clipview

Top: this PR with clipping disabled
bottom: default behavior
Screenshot_20250322-115430

Run the code and click the first view to toggle:
```js
var view = Ti.UI.createView({
	top: 0,
	backgroundColor: 'yellow',
	height: 300,
	width: 400,
});
var nonClippingView = Ti.UI.createView({
	backgroundColor: 'red',
	height: 200,
	width: 200,
	clipMode: Ti.UI.CLIP_MODE_DISABLED
});
var insideViewB = Ti.UI.createView({
	backgroundColor: 'blue',
	height: 10,
	width: 300,
});
view.add(nonClippingView)
nonClippingView.add(insideViewB);
win.add(view);

var contentView = Ti.UI.createView({
	top: 330,
	backgroundColor: 'yellow',
	height: 300,
	width: 400,
});
var clippingView = Ti.UI.createView({
	backgroundColor: 'red',
	height: 200,
	width: 200
});
var insideViewA = Ti.UI.createView({
	backgroundColor: 'blue',
	height: 10,
	width: 300
});
contentView.add(clippingView);
clippingView.add(insideViewA);
win.add(contentView);
var toggle = true;
view.addEventListener("click", function() {
	if (toggle) {
		nonClippingView.clipMode = Ti.UI.CLIP_MODE_ENABLED
	} else {
		nonClippingView.clipMode = Ti.UI.CLIP_MODE_DISABLED
	}
	toggle = !toggle;
});
win.open()

```

* iOS works now too with `Ti.UI.CLIP_MODE_DISABLED` and `Ti.UI.iOS.CLIP_MODE_DISABLED` for backwards compatibility


android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1141,6 +1141,16 @@ public void processProperties(KrollDict d)
 			}
 		}
 
+		if (d.containsKey(TiC.PROPERTY_CLIP_MODE) && !nativeViewNull) {
+			if (nativeView != null
+				&& TiConvert.toInt(d, TiC.PROPERTY_CLIP_MODE) == -1
+				&& nativeView instanceof ViewGroup viewGroup
+				&& viewGroup.getParent() == null) {
+				viewGroup.setClipChildren(false);
+				viewGroup.setClipToPadding(false);
```
prashantsaini1, line 1150: - `setClipToPadding` only works if the view has a padding set, which we don't support at this moment.
- This may even work unexpected inside ListView. `setClipToChildren` and `setClipToPadding` serves entirely different purposes.
- In iOS, `clipsToBounds` is equivalent to `setClipToChildren`, but there's no direct equivalent to Android's `setClipToPadding`.

prashantsaini1 (commented):
@m1ga I tested the PR and it does not work as intended. It does not allow to change property once set. Also it does not work if we remove the `clipMode` from the `root yellow view` despite that the `red view` has it `clipMode: disabled` already.

android/titanium/src/java/org/appcelerator/titanium/TiC.java
```diff
@@ -362,6 +362,7 @@ public class TiC
 	public static final String PROPERTY_CLASS_NAMES = "classNames";
 	public static final String PROPERTY_CLEAR_ON_EDIT = "clearOnEdit";
 	public static final String PROPERTY_CLIP_VIEWS = "clipViews";
+	public static final String PROPERTY_CLIP_MODE = "clipMode";
```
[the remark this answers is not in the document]
m1ga: I've added `Android: This property is "creation only" and has to be on the top-level view.` to the readme but I can change that to make it changeable.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
m1ga: Ah, good to know. I only used both because of the example module. I can remove the padding one

m1ga:
> @m1ga I tested the PR and it does not work as intended. It does not allow to change property once set. Also it does not work if we remove the `clipMode` from the `root yellow view` despite that the `red view` has it `clipMode: disabled` already.

Correct. Check the readme notes `Android: This property is "creation only" and has to be on the top-level view.`.

prashantsaini1:
Ok, I didn't check the readme notes, but rather compared this property with iOS where it properly works as expected. I believe we should do more testing around it as Android has a pretty different behaviour to both these properties. And yes, it requires to be on top-level view, which actually we can set ourselves in SDK. 

I suggest that we introduce the `padding` property as well so we can give more flexibilities to use `setClipToPadding()` which allows to create more immersive scrolling/animation experience. Though a low priority for me 😄

m1ga:
> And yes, it requires to be on top-level view, which actually we can set ourselves in SDK.

very good idea! Yeah, I'll do some more testing and adjust the code but thanks for the check and the feedback :+1:

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -974,6 +976,10 @@ public void propertyChanged(String key, Object oldValue, Object newValue, KrollP
 			} else {
 				Log.w(TAG, "Setting the 'viewShadowColor' property requires Android P or later");
 			}
+		} else if (key.equals(TiC.PROPERTY_CLIP_MODE)) {
+			if (nativeView != null) {
+				setClipMode(TiConvert.toInt(newValue));
```
prashantsaini1, line 981: Use `optInt` with default as `CLIP_MODE_DEFAULT`.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1142,13 +1148,7 @@ public void processProperties(KrollDict d)
...
 
 		if (d.containsKey(TiC.PROPERTY_CLIP_MODE) && !nativeViewNull) {
-			if (nativeView != null
-				&& TiConvert.toInt(d, TiC.PROPERTY_CLIP_MODE) == -1
-				&& nativeView instanceof ViewGroup viewGroup
-				&& viewGroup.getParent() == null) {
-				viewGroup.setClipChildren(false);
-				viewGroup.setClipToPadding(false);
-			}
+			setClipMode(TiConvert.toInt(d, TiC.PROPERTY_CLIP_MODE));
```
prashantsaini1, line 1151: Same here, a default optional to reset the value.

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1207,6 +1207,21 @@ private void applyCustomBackground(boolean reuseCurrentDrawable)
 		}
 	}
 
+	private void setClipMode(int clipMode)
+	{
+
+		if (nativeView instanceof ViewGroup viewGroup) {
+			if (clipMode == UIModule.CLIP_MODE_DISABLED) {
```
prashantsaini1, line 1214: Just the `setClipChildren(clipMode != UIModule.CLIP_MODE_DISABLED)` is fine. Further we can internally set this on top-most view as discussed earlier.

m1ga: Yes, thanks for that :+1: 

I've put it into draft as I know that parts are still missing especially the parent part. I've added your other suggestions. Thanks for the review :+1:

android/titanium/src/java/org/appcelerator/titanium/view/TiUIView.java
```diff
@@ -1197,6 +1207,14 @@ private void applyCustomBackground(boolean reuseCurrentDrawable)
 		}
 	}
 
+	private void setClipMode(int clipMode)
```
prashantsaini1, line 1210: I think this should work to get the root view (I haven't tested it though).
```
private void setClipMode(int clipMode)
{
	if (nativeView == null) {
		return;
	}

	View rootView = nativeView;
	while (rootView.getParent() != null && rootView instanceof ViewGroup) {
		rootView = (View) rootView.getParent();
	}

	if (rootView instanceof ViewGroup viewGroup) {
		viewGroup.setClipChildren(clipMode != CLIP_MODE_DISABLED);
	}	
}
```

m1ga: ` Error: android.view.ViewRootImpl cannot be cast to android.view.View` think it's one level to high. But I'll do some more tests at another point

prashantsaini1: I think we can remove the `(View)` casting, and declare `rootView` as `ViewParent`, even or simply keep it as an Object as it's already casted in next `if` condition.

m1ga:
The loop is running now. The only issue: we should stop it so it won't run to the root item:
Bildschirmfoto_20250522_185109
Here the marked `TiCompositeLayout` should be the last one, the next parent is `Ti.UI.Window`. Currently I've added a check for the API name but of course that should be more flexible.

m1ga:
Claude helped here and now it's behaving the same way as iOS :+1:

PR #14194, opened 2025-03-24
fix(android): add networkMode ethernet to networkType

m1ga:
`NetworkModule.NETWORK_LAN` is currently not used in the `networkType` property but mentioned in the docs.

Found and tested by a Slack user.


dlewis23:
Tested on my device and it works great. Thanks a ton @m1ga

hansemannn (approved):

PR #14201, opened 2025-04-08
fix(android): add null checks on ListView/TableView item/row proxy

prashantsaini1:
Adds the required null checks where a ListItem proxy can be null when asynchronous updates happen to the ListView.


m1ga (approved):
LGTM

PR #14203, opened 2025-04-09
revert(android): revert method signatures for compatibility

prashantsaini1:
- This PR reverts the method  `removeOnLifecycleEventListener` return type which was earlier `void` but changed to `boolean` in [another PR here](https://github.com/tidev/titanium-sdk/pull/14189).
- It also reverts all other similar methods return type for consistency (though all other methods were newly added).

It allows modules, compiled with older SDKs (<= 12.6.4), to run on newer SDK versions with the [another PR](https://github.com/tidev/titanium-sdk/pull/14189) merged.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -1317,39 +1317,39 @@ public abstract class TiBaseActivity extends AppCompatActivity implements TiActi
 		onPrepareOptionsMenuListeners.add(new WeakReference<>(listener));
 	}
 
-	public boolean removeOnLifecycleEventListener(OnLifecycleEvent listener)
+	public void removeOnLifecycleEventListener(OnLifecycleEvent listener)
 	{
-		return lifecycleListeners.remove(listener);
+		lifecycleListeners.remove(listener);
 	}
 
-	public boolean removeOnInstanceStateEventListener(OnInstanceStateEvent listener)
+	public void removeOnInstanceStateEventListener(OnInstanceStateEvent listener)
 	{
-		return instanceStateListeners.remove(listener);
+		instanceStateListeners.remove(listener);
 	}
 
-	public boolean removeOnWindowFocusChangedEventListener(OnWindowFocusChangedEvent listener)
+	public void removeOnWindowFocusChangedEventListener(OnWindowFocusChangedEvent listener)
 	{
-		return windowFocusChangedListeners.remove(listener);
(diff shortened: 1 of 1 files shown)


m1ga (approved):
LGTM.

Which module was causing issues here?

prashantsaini1:
The latest Android code of [ti.bottomsheet](https://github.com/hansemannn/titanium-bottom-sheet/commit/9195fb7072ec08dc7a316db38059120a3d74f616) module which uses `removeOnLifecycleEventListener`.

- It will crash the app if we compile the module with `<= 12.6.4.GA` but build the app with the latest `ti-sdk master (12.7.0)` as the signatures will be different in this case.
- However, using the same SDK version for both the module and the apps will be fine.

m1ga:
Ah ok. I was scanning through some Ti modules and didn't find it in there. Still strange because you are not using the return value so in my opinion it should call the same method. But glad this fixes it

prashantsaini1:
Whether or not we use the return value, at compilation of the module the Java Class will be used based on the SDK version which will differ for different SDK versions without this PR.

m1ga:
I see, just tested it and can confirm:
```
No virtual method removeOnLifecycleEventListener(Lorg/appcelerator/titanium/TiLifecycle$OnLifecycleEvent;)V in class Lorg/appcelerator/titanium/TiBaseActivity; or its super classes (declaration of 'org.appcelerator.titanium.TiBaseActivity'
```

with this changes it works again. 

Side note: you might want to update the ` Calling .toColor() without Context parameter is deprecated` in your module.

prashantsaini1:
Yes, that's the crash error. Thanks for the hint on `toColor`, just pushed it too.

However, I think that we should use current-activity context if available, otherwise the TiApplication context in the SDK itself to make it backward compatible and remove that deprecated warning. Reason being that majority of the use cases will rely on the current activity context. What do you think?

e.g. 
```
Configuration config;

if (context != null) {
    config = context.getResources().getConfiguration();
} else if (TiApplication.getAppCurrentActivity() != null) {
    config = TiApplication.getAppCurrentActivity().getResources().getConfiguration();
} else {
    config = TiApplication.getInstance().getResources().getConfiguration();
}
```

m1ga:
The deprecation note is a preparation of this PR https://github.com/tidev/titanium-sdk/pull/13273 
But I don't have any more details about that PR. I just know it will break all modules that don't use the new toColor way. But I think most first party modules are update by now.

prashantsaini1:
@m1ga Thanks. @hansemannn We can merge this PR now.

PR #14204, opened 2025-04-10
feat: expose document title to Ti.UI.iOS.DocumentViewer

hansemannn:
This allows the document viewer to show a human readable title instead of the URL.

### DIFF
--- a/apidoc/Titanium/UI/iOS/DocumentViewer.yml
+++ b/apidoc/Titanium/UI/iOS/DocumentViewer.yml
@@ -1,120 +1,128 @@
 ---
 name: Titanium.UI.iOS.DocumentViewer
 summary: |
-    A DocumentViewer provides in-app support for managing user interactions with files on the
-    local system.
+  A DocumentViewer provides in-app support for managing user interactions with files on the
+  local system.
 description: |
-    The DocumentViewer is created by the <Titanium.UI.iOS.createDocumentViewer> method.
+  The DocumentViewer is created by the <Titanium.UI.iOS.createDocumentViewer> method.
 
-    Use this class to present an user interface for previewing files,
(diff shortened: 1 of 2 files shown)


hansemannn:
@prashantsaini1 can you review this please? I've tested it in our apps of course.

iphone/Classes/TiUIiOSDocumentViewerProxy.m
```diff
@@ -25,6 +25,7 @@ - (UIDocumentInteractionController *)controller
   if (controller == nil) {
     NSURL *url = [self _toURL:[self valueForUndefinedKey:@"url"] proxy:self];
     controller = [[UIDocumentInteractionController interactionControllerWithURL:url] retain];
+    controller.name = [self valueForUndefinedKey:@"title"];
```
prashantsaini1, line 28: `controller.name` will be `nil` if the `title` property is not passed, is it fine as well?

apidoc/Titanium/UI/iOS/DocumentViewer.yml
prashantsaini1, line 1: Looks like a prettier pitched in here.

iphone/Classes/TiUIiOSDocumentViewerProxy.m
hansemannn: Yes, that is the expected behavior, as it fallbacks to the default.

prashantsaini1 (approved):

iphone/Classes/TiUIiOSDocumentViewerProxy.m
prashantsaini1: I can't see the option to mark this conversation resolved myself. Seems I'm missing some permissions.

PR #14208, opened 2025-04-16
feat(android): support html in text areas

m1ga:
Parity for: https://github.com/tidev/titanium-sdk/pull/14207
```js
const txt = Ti.UI.createTextArea({html:"<b>test</b> <i>test</i> test", width: 200,height:100});
setTimeout(function(){
	txt.html = "new <b>text</b> in field"
},2000);
win.add(txt);
win.open();
```


hansemannn (approved):
🚀

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -288,6 +289,13 @@ public void processProperties(KrollDict d)
 			}
 		}
 
+		if (d.containsKey(TiC.PROPERTY_HTML)) {
+			Spanned text = Html.fromHtml(TiConvert.toString(d, TiC.PROPERTY_HTML));
+			if (text != null) {
+				tv.setText(text, TextView.BufferType.SPANNABLE);
```
prashantsaini1, line 295: @m1ga I believe this should be `TextView.BufferType.EDITABLE` to allow editing as well. `SPANNABLE` is only for formatting with styled text.

hansemannn: It should actually not be editable, same as for iOS. If we use HTML + editing, that would be a RTE like https://github.com/hansemannn/titanium-rich-text-editor

m1ga: it's still an editable text field and you can change the text but it doesn't allow you to change the format of course. 

I don't know what the iOS version does :) This PR will just allow you to set a HTML text in a normal TextArea:
Screenshot_20250416-141650
same way the label does it, but here just as an editable field

prashantsaini1: The docs didn't mention that this property makes the TextArea uneditable. We should actually check for the existing `editable` property to decide the `BufferType` value for flexibility.

m1ga: > The docs didn't mention that this property makes the TextArea uneditable. We should actually check for the existing `editable` property to decide the `BufferType` value for flexibility.

https://developer.android.com/reference/android/widget/TextView.BufferType#EDITABLE
is not really helpful :smile: No explanation.

According to https://stackoverflow.com/questions/65642029/the-effect-of-buffertype-normal-and-buffertype-spannable-for-textview or https://stackoverflow.com/a/41002135/5193915 it doesn't really matter in this case and I think SPANNBLE might even be better as I'm adding a spanned element in there and not a plain text.

PR #14213, opened 2025-05-01
feat(android): clearUserCache method

m1ga:
Add new method `Ti.App.Android.clearUserCache()` to clear the user data and app cache. It's basically the same as going into the app info menu - cache - clear data. It will close the app.

```js
var win = Titanium.UI.createWindow({});
win.open();
var btn = Ti.UI.createButton({title:"clear"})
win.add(btn);
btn.addEventListener("click", function(){
	console.log(Ti.App.Android.clearUserCache());
})
```


Vinnn021414:
E

prashantsaini1 (commented):
This will perform the **Clear Data** action rather than the **Clear Cache**. Curious to know when someone would want to do this with their app?

m1ga:
Correct. As mentioned in the docs `Clears app data and cache` so the whole app will be reset. It might be useful if you have user data that you want to clear completely and want to start with a fresh app. I've used it a lot on apps during debugging. Having it in the app would speed up that process

apidoc/Titanium/App/Android/Android.yml
```diff
@@ -64,6 +64,11 @@ properties:
     permission: read-only
     since: 3.3.0
 
+methods:
+  - name: clearUserCache
+    summary: Clears app data and cache. This will close the app.
+    since: 12.8.0
```
m1ga:
```suggestion
    since: 13.1.0
```

prashantsaini1 (approved):

PR #14221, opened 2025-05-05
feat: option to remove X-Titanium-Id in request header

m1ga:
Currently the httpclient will always set a `X-Titanium-Id`. You can set it to be a different string or empty but you can't remove it.

This PR will add a propety `titaniumId: false` so you can disable it. If it is not set it will still be `true` so it doesn't change anything for users that rely on it. 

```js

var url = "httpbin.org/get";
    onload : function(e) {
        Ti.API.info("Received text: " + this.responseText);
    },
    onerror : function(e) {},
    timeout : 5000
});

client.open("GET", url);
// client.setRequestHeader("X-Titanium-Id", null);  // doesn't work
// client.setRequestHeader("X-Titanium-Id", '');  // still there but empty
client.send();
```

output without the PR:
```
[INFO]  Received text: {
[INFO]   "args": {},
[INFO]   "headers": {
[INFO]     "Accept-Encoding": "identity",
[INFO]     "Host": "httpbin.org",
[INFO]     "User-Agent": "Titanium SDK/12.7.0 (Pixel 9 Pro; Android API Level: 35; de-DE;)",
[INFO]     "X-Amzn-Trace-Id": "Root=1----",
[INFO]     "X-Titanium-Id": "xxxx-xxxxx-xxxxx-xxxx"                <----- this will be gone
[INFO]   },
[INFO]   "origin": "...",
[INFO]   "url": "http://httpbin.org/get"
[INFO]  }
```


hansemannn:
Honestly, we should remove it altogether. This does not bring any benefit.

m1ga:
True, and I think if people rely on it to identify e.g. a webview they could use other headers or set it again if needed. I'll remove it

cb1kenobi:
Could always make it an option to keep it, but I’m all for removing it.

cb1kenobi (approved):

PR #14235, opened 2025-06-10
fix(android): try to fix the tab-group crash on some devices

prashantsaini1:
In follow up with this [previous PR](https://github.com/tidev/titanium-sdk/pull/14210), we are still seeing same crashes.

After doing more research on this issue, the only clues I could find is that it's OEM specific issue (also highly indicated by our reports as well), and that some transitions during the early window startup from the splash-screen is causing this issue.

So I added some delay in opening the window, and also removed the initial transitions from splash-screen. I really hope it should fix the issue!

_Note: We are still throwing that exception so we can see it in Firebase._


m1ga (approved):
Should be fine as most of the devices shouldn't reach that catch part in the best case.

I've tested it with 3 different apps and all behave normally and open fine.

PR #14236, opened 2025-06-10
fix(ios): prevent crash on iOS 26 related to Ti.UI.ScrollableView

hansemannn:
This pull request includes a small improvement to the `frameSizeChanged:bounds:` method in `TiUIScrollableView.m`. The change ensures that subviews implementing the `checkBounds` method are properly checked for bounds updates before proceeding, fixing a crash happening on iOS 26+.


hansemannn:
@janvennemann Reminder - there is one open issue that sometimes causes the scrollable view to render an empty view. That should be planned in one of the next sprints please. EDIT:

It seems like setting "top" / "bottom" causes the issue in the first place. Repro case:
```js
const window = Ti.UI.createWindow({ title: 'Window 1' });

    tabs: [
        Ti.UI.createTab({ window, title: 'Tab 1' })
    ]
})

const scrollableView = Ti.UI.createScrollableView({
    bottom: 0,
    top: 0,
    views:
    [
        Ti.UI.createView({ backgroundColor: 'blue'}),
        Ti.UI.createView({ backgroundColor: 'green'}),
        Ti.UI.createView({ backgroundColor: 'yellow'})
    ]
});

window.add(scrollableView);
tabGroup.open();
```

Workaround: Remove the properties, as the scrollable view is set to width/height = `Ti.UI.FILL` by default.

janvennemann:
@hansemannn I've made a few more changes to the scrollable view. Since iOS 26 it seems there are some internal subviews added to the the `subviews` array of the `UIScrollView`.

Screenshot 2025-08-28 at 17 50 00

Not sure if this intentional or a bug, but with this new change we make sure to only consider our own subviews for layout computation.

janvennemann (approved):

designbymind:
Just wanted to follow up here: this fixed a few issues (strange behavior) that I was experiencing with `ScrollableView` on iOS 26 b7.
- When `ScrollableView` `height` was set `Ti.UI.FILL`, I noticed pages "disappearing" when they weren't on < iOS 18.
- When using `.currentPage = x` to set view index, the pages would often be blank and the entire `.views` array would fall out of 'sync'.
- When manually _and_ programmatically scrolling between views, often times they would "disappear", especially if there were more than 2 pages in the `.views` array.

All that said, I just patched in this PR and all of the above issues seem to be fixed on both sim and device. — Once again, **thank you**!

janvennemann:
Great to hear it's working again for you! We were seeing similar issues and they should all have been addressed with [bb95aca](https://github.com/tidev/titanium-sdk/pull/14236/commits/bb95aca1687a04e4efb6d825d511b7d1b66dfd2e).

PR #14238, opened 2025-06-10
feat(ios): add support for iOS 26+ glass effects

hansemannn:
This pull request introduces a new method for setting glass effects in `TiUIiOSBlurView.m`, specifically targeting iOS 26 and above. The method allows customization of interactive behavior and tint color for the glass effect.

### New functionality for iOS 26+:

* Added the `- (void)setGlassEffect_:(id)value` method to support setting glass effects using the `UIGlassEffect` class. The method checks for iOS 26 compatibility, ensures type safety for input, and configures properties like `interactive` and `tintColor`.

simulator_screenshot_A067E11F-0787-4A83-AEA1-A5585DB2769E


cb1kenobi:
How do I test this?

hansemannn:
Still wip - example code, docs and some screen recording will be delivered next week 😊

designbymind:
Also willing to test this (and looking forward to it)! — Thank you for the great work, @hansemannn 🙂

hansemannn:
@cb1kenobi @designbymind The code is now fully completed, tested, documented and ready for review/merge!

apidoc/Titanium/UI/iOS/iOS.yml
```diff
@@ -334,6 +334,28 @@ properties:
...
 
+  - name: GLASS_EFFECT_STYLE_REGULAR
+    summary: |
+        Use with [BlurView.glassEffect](Titanium.UI.iOS.BlurView.glassEffect) to specify a regular glass effect style.
+    description: |
+        Creates a standard glass effect with regular translucency and blur intensity. This is the default
+        style for glass effects and provides a balanced appearance suitable for most use cases.
+    type: Number
+    permission: read-only
+    since: "14.0.0"
```
janvennemann: Shouldn't this be `13.0.0` as well?

iphone/Classes/TiUIiOSBlurView.m
```diff
@@ -49,6 +49,28 @@ - (void)setEffect_:(id)value
...
+  if (@available(iOS 26.0, *)) {
+    UIGlassEffect *glassEffect = [UIGlassEffect effectWithStyle:style];
+    glassEffect.interactive = isInteractive;
+    glassEffect.tintColor = tintColor.color;
+
+    [[self blurView] setEffect:glassEffect];
+    [[self proxy] replaceValue:value forKey:@"glassEffect" notification:NO];
+  } else if (@available(iOS 26.0, *)) {
+    NSLog(@"[ERROR] Setting glass effects require iOS 26+");
+  }
```
janvennemann: Move this to the top and return early from the function to skip all glass effect related code entirely. I would also remove the `NSLog` as this is not really an error but expected behavior.

janvennemann (changes requested):

janvennemann (approved):
Looks good to me for the basic functionality. I was curious about the corner radius issue and apparently you have to use the new [UICornerConfiguration](https://developer.apple.com/documentation/uikit/uicornerconfiguration-c.class?language=objc) to properly configure the corner radius for glass effect views. That requires overriding our border radius handling which we should probably look at in a separate PR.

I'll let it up to you to merge this PR it in case you want to wait for testing feedback from the others.

hansemannn:
@janvennemann Funny that you mention it - I've just added that in the mean time. The click events also work now (never worked on the blur view).

janvennemann:
> @janvennemann Funny that you mention it - I've just added that in the mean time. The click events also work now (never worked on the blur view).

Ha,. nice! But i think you might need to revisit that :P The `borderRadius` property also supports string and arrays. Using an array like `borderRadius: [20, 20, 0, 0]` on a blur view will now crash as the override treats everything as `NSNumber`.

hansemannn:
Yeah, I've considered that, but thought that may extend the scope of the PR too much. Adding an own PR for it tomorrow

PR #14241, opened 2025-06-29
fix(android): apply required null checks on TiUIEditText

prashantsaini1:
Fixes some cases where the `TextField/TextArea` can be accessed after their native view is released, and might trigger events like `change/return` etc.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
@@ -143,7 +143,8 @@ public class TiUIText extends TiUIView implements TextWatcher, OnEditorActionLis
 				@Override
 				public boolean onKey(View v, int keyCode, KeyEvent event)
 				{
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -902,6 +916,8 @@ private void handleFullscreen(KrollDict properties)
 
 	public void setSelection(int start, int end)
 	{
+		if (tv == null) return;
```
m1ga, line 919: shouldn't you use `getText()` here too?  
Rest looks fine

prashantsaini1: Essentially same effect, since we are using other methods on `tv` next, so it's fine here.

m1ga (approved):
Looking good :+1: 

tested 3 apps, all textfields/textares still work fine

PR #14245, opened 2025-07-23
fix(android): fix transparent TextField backgroundColor

m1ga:
The previous fix for the TextField background color https://github.com/tidev/titanium-sdk/pull/14222 fixed setting a backgroundColor in a textfield. Looks like "transparent" still shows a background color.

12.8.0:
Screenshot_20250817-183957

```js
	backgroundColor: "#999"
});
var tf = Ti.UI.createTextField({
	width: 200,
	height: 40,
	top: 10,
	value: "default"
})
var tf2 = Ti.UI.createTextField({
	width: 200,
	height: 40,
	top: 60,
	backgroundColor: "transparent",
	value: "transparent"
})
var tf3 = Ti.UI.createTextField({
	width: 200,
	height: 40,
	top: 110,
	backgroundColor: null,
	value: "null"
})


win.add([tf, tf2, tf3]);
win.open();
```

With this PR it will create a transparent TextField again.

Screenshot_20250817-185127

Note: `borderStyle: Titanium.UI.INPUT_BORDERSTYLE_NONE` would be the better choice but people use `backgroundColor: "transparent"` so we should still support that :-)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIText.java
```diff
@@ -203,8 +203,12 @@ public void processProperties(KrollDict d)
 		if (d.containsKey(TiC.PROPERTY_BACKGROUND_COLOR)) {
 			// Why transparent?
 			tv.setBackgroundColor(Color.TRANSPARENT);
-			textInputLayout.setBoxBackgroundColor(TiConvert.toColor(d.get(TiC.PROPERTY_BACKGROUND_COLOR),
-				TiApplication.getAppCurrentActivity()));
+			if (d.get(TiC.PROPERTY_BACKGROUND_COLOR).equals("transparent")) {
```
prashantsaini1: - This would probably crash if the `backgroundColor` is passed `null`. A safe condition would be this `"transparent".equals(d.get(TiC.PROPERTY_BACKGROUND_COLOR)))` using the `equals()` on the literal.
- Also, passing the `null` should logically be equal to passing a `transparent` value, right?

m1ga: thanks for the hint. let me check it with `null`. I think `null` should be the default (unstyled) color and not transparent.

m1ga: I've checked it with 12.6.0 and `null` was transparent there too, so you are right. Updated the code and screenshots.

prashantsaini1 (approved):

m1ga:
@hansemannn can we merge this to 13.0.0 too?

PR #14246, opened 2025-07-25
fix(android): add deselectItem to Androids ListView

m1ga:
On iOS you use `$.lst_open.deselectItem(0, e.itemIndex)` to deselect a clicked item. This method is not needed on Android but also not available in the SDK so you have to add an `if OS_IOS` all the time.

This PR will just add an empty method to the Android ListView so you can remove the `if OS_IOS` statement.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
@@ -314,6 +314,12 @@ public class ListViewProxy extends RecyclerViewProxy
 		}
 	}
 
+	// not needed on Android but parity for iOS method call
+	@Kroll.method
+	public void deselectItem(int section, int item)
+	{
+	}
+
 	@Override
 	public String getApiName()
 	{


prashantsaini1:
It would be better to trigger the automatic deselection of a `ListItem` for iOS in the SDK itself as majority of the Ti apps use `deselectItem` on iOS anyways. It would also make the UX same on Android/iOS by showing a touch-ripple effect on item click and automatically fading it away.

hansemannn:
For parity with the current iOS implementation, this is the correct way. Changing the default behavior could be inside future versions then. As it was not requested so far, it seems like low prio though.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/listview/ListViewProxy.java
```diff
@@ -314,6 +314,12 @@ public void deleteSectionAt(int index, @Kroll.argument(optional = true) KrollDic
 		}
 	}
 
+	// not needed on Android but parity for iOS method call
+	@Kroll.method
+	public void deselectItem(int section, int item)
```
hansemannn, line 319: I would expect an empty-function warning or something. Isn't that the case?

prashantsaini1: A debug warning is better here.

hansemannn: Maybe, if Android does not have the concept, it really should just not be exposed. I thought about it a bit more and think that it's much more clean instead of an empty function.

m1ga: I thought it might be a good empty function as you have to use it on iOS so often and you always have to make the `if OS_IOS` check before it because it will throw a "method not found" error on Android.

but as you've said: no requests by other people yet, just thought it might be a good convenient method

hansemannn: You can just write `$.listView.deselectItem?.(event.sectionIndex, event.itemIndex)` for that :)

m1ga:
`$.listView.deselectItem?` is the better solution here :smile: Forgot to close this.

PR #14259, opened 2025-08-30
feat(ios): support for iOS 26+ bottomAccessoryView API

hansemannn:
This PR adds support for the iOS 26+ "bottomAccessoryView" API on the Ti.UI.TabGroup. It creates a floating bottom view that can be shown / hidden by setting the property. It also adds support for the "minimizeBehavior" property to control the automatic showing / hiding of the view based on certain gestures.

```js
	title: "Window 1",
	backgroundColor: "#fff",
});

const items = Array.from({ length: 100 }, (_, i) => ({ properties: { title: `Item ${i + 1}` } }));
const listView = Ti.UI.createListView({ sections: [Ti.UI.createListSection({ items })] });
window.add(listView);

const bottomAccessoryView = Ti.UI.createView();
bottomAccessoryView.add(Ti.UI.createLabel({ text: 'Hello world', width: Ti.UI.FILL, height: Ti.UI.FILL, textAlign: 'center' }));


	tabs: [Ti.UI.createTab({ window, title: "Tab 1" })],
	bottomAccessoryView,
	minimizeBehavior: Ti.UI.iOS.TAB_GROUP_MINIMIZE_BEHAVIOR_ON_SCROLL_DOWN
});

tabGroup.open();
```

https://github.com/user-attachments/assets/08366f82-eca3-40ed-8718-6a5252783f5e

### DIFF
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -413,7 +413,7 @@ DEFINE_EXCEPTIONS
       }];
 }
 
-#ifndef IS_SDK_IOS_26
+#if IS_SDK_IOS_26
 - (void)setBottomAccessoryView_:(id)bottomAccessoryViewProxy
 {
   if (bottomAccessoryViewProxy == [NSNull null]) {
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -413,6 +413,7 @@ DEFINE_EXCEPTIONS
       }];
 }
 
+#ifndef IS_SDK_IOS_26
 - (void)setBottomAccessoryView_:(id)bottomAccessoryViewProxy
 {
   if (bottomAccessoryViewProxy == [NSNull null]) {
@@ -438,6 +439,7 @@ DEFINE_EXCEPTIONS
 {
   [[self tabController] setTabBarMinimizeBehavior:minimizeBehavior.integerValue];
 }
+#endif
 
 - (void)setTabsBackgroundColor_:(id)value
 {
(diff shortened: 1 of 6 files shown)


iphone/Classes/TiUITabGroup.m
```diff
@@ -413,6 +413,34 @@ - (void)hideTabBar:(BOOL)hidden animated:(BOOL)animated
       }];
 }
 
+#if IS_SDK_IOS_26
+- (void)setBottomAccessoryView_:(id)bottomAccessoryViewProxy
+{
+  if (bottomAccessoryViewProxy == [NSNull null]) {
+    [[self tabController] setBottomAccessory:nil];
+    return;
+  }
```
janvennemann, line 427: If this had a bottom accessory view before you also need to forget the proxy and release the view here.

janvennemann (changes requested):

janvennemann (approved):

designbymind:
**This is awesome!** — I've had a chance to check it out over the weekend and all is working great on my end (both sim and device). Below are some nuances that I found when working with `TabGroup` `bottomAccessoryView`:

- To remove, setting `bottomAccessoryView: null` can be used.
- Setting `tabBarHidden: true` on a `Window` does not remove/hide the `bottomAccessoryView`, manual tracking is a must as of now.
- `bottomAccessoryView` can be set on a per-`Tab` bases (for a more context-aware experience). Simply setting the `bottomAccessoryView` a second time will override the current `bottomAccessoryView`.
- For those curious, I found the `bottomAccessoryView` to be `48pt` in height (iPhone SE-16 Pro Max). **Note:** Apple seems to be moving towards dynamic sizing all around, so this number may not be accurate in the future and probably shouldn't be heavily relied on.

Looking forward to using this, thank you so much for the great work!

PR #14266, opened 2025-09-03
Revert "fix(ios): properly handle containment of tab group to fix crash on start up"

hansemannn:
Reverts tidev/titanium-sdk#14261

### DIFF
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -692,13 +692,12 @@ DEFINE_EXCEPTIONS
 {
   TiThreadPerformOnMainThread(
       ^{
-        UIViewController *parentController = TiApp.controller.topPresentedController;
-        // Establish proper containment: add child, add its view, then notify didMove
-        [parentController addChildViewController:self.tabController];
+        [self.tabController willMoveToParentViewController:TiApp.controller.topPresentedController];
+
         self.tabController.view.frame = self.bounds;
         [self addSubview:self.tabController.view];
         isTabBarHidden = NO;
-        [self.tabController didMoveToParentViewController:parentController];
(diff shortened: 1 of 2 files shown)


designbymind:
**Just confirming:** the revert returned things to 'normal', crash is no longer on my end. Thank you!

hansemannn:
@designbymind Would you be able to attach a simple app.js reproducing your crash? It would help a lot to fix the issue, a sthe general fix here was valid and our app didn't crash, for reference. Thanks for taking the time to contribute this!

designbymind:
@hansemannn Of course, happy to help here, anytime! — I will get that `app.js` reproduction out to you soon.

designbymind:
@hansemannn I'm pretty sure that it's on my end. — Built that simple two-tab test app (w/ Alloy too) using TiSDK 13.1.0 and iOS 26. Everything is working great, this PR included. — I'll be looking into my issue more (as the app is much more complex) and should I discover something new, I will share. — As of now, your PR looks valid. Thank you

hansemannn:
I think we should still look into it, as it would be a breaking change. Are you doing something special on start up, e.g. opening a window directly again?

designbymind:
Yes, that was exactly it! — What I had was a `TabGroup` contained within the `centerWindow` of [this module](https://github.com/mbender74/NappDrawer). — For me, the drawer was unnecessary, so I simply removed it, but that what was causing the crash on my end.

hansemannn:
Yeah, the module has it's own containment / child controller issues, which became critical with the recent fixes. So we can potentially re-open this - what do you think @janvennemann

janvennemann:
The tab group controller relationship was also slightly changed in #14267 so there shouldn't be a need to re-open this. As far as i have seen during all the testing with safe area insets recently, the controllers are correctly connected with each other.

PR #14267, opened 2025-09-05
fix(ios): improve safe area detection and trigger relayout

janvennemann:
## Summary

- Improves safe area inset detection on iOS by using view controller's lifecycle method viewSafeAreaInsetsDidChange
  instead of manual triggers
- Fixes tab group containment hierarchy to properly add child view controllers
- Simplifies safe area calculation logic for better accuracy

## Changes

- Added viewSafeAreaInsetsDidChange callback to properly detect safe area changes in TiViewController
- Fixed tab group view controller containment by calling didMoveToParentViewController after adding child
- Removed redundant processForSafeArea calls that were triggered manually
- Simplified safe area inset calculation to use controller's own view hierarchy
- Trigger proper relayout when navigation bar visibility changes

## Technical Details

With iOS 26 and the new glass effect design safe areas become more important for correctly laying out views. Titanium currently differs from the native iOS behavior in two properties:

- `extendsEdges`: This is currently set to `UIRectEdgeNone` by default, which means that a window does not extend under bars from navigation controllers or tab bar controllers. The matching native iOS property is `edgesForExtendedLayout` which usually defaults to `UIRectEdgeAll`.
- `autoAdjustScrollViewInsets`: This maps to [automaticallyAdjustsScrollViewInsets](https://developer.apple.com/documentation/uikit/uiviewcontroller/automaticallyadjustsscrollviewinsets?language=objc) on iOS and is actually deprecated since iOS 11. Titanium will set this to `false` by default, which differs from the native default value, which is `true`.

To achieve a native like look and feel for views that contain list views or scroll views the following style has to be set:

```
'Window[platform=ios]': {
  extendEdges: [Ti.UI.EXTEND_EDGE_ALL],
  autoAdjustScrollViewInsets: true
}
```

Using this window configuration revealed that we have issues that would not compute the correct safe area for windows inside navigation windows or tab groups. This PR tries to address this so Titanium apps that use this property combination can layout content at the top or bottom of a window using proper safe area insets that respect navigations bars and tab bars.

### DIFF
--- a/iphone/Classes/TiUINavigationWindowProxy.m
+++ b/iphone/Classes/TiUINavigationWindowProxy.m
@@ -265,7 +265,6 @@
     }
   }
   TiWindowProxy *theWindow = (TiWindowProxy *)[(TiViewController *)viewController proxy];
-  [theWindow processForSafeArea];
   if ((theWindow != rootWindow) && [theWindow opening]) {
     [theWindow windowWillOpen];
     [theWindow windowDidOpen];
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -666,8 +666,7 @@ DEFINE_EXCEPTIONS
       focusedTabProxy = [theActiveTab retain];
     }
 
-    [self tabController].viewControllers = nil;
-    [self tabController].viewControllers = controllers;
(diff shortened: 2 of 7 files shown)


hansemannn (approved):
Excellent work!

PR #14268, opened 2025-09-09
fix(ios): include top safe area inset in navigation window

janvennemann:
This is a follow up fix for #14267 to always include the top safe area inset inside a navigation window context. We already fixed the inset in tab bar context (which automatically includes a navigation window) in the previous PR.


hansemannn (approved):

hansemannn:
The base branch was incorrect, but I've cherry-picked it into `master` as well to have them in sync again.

janvennemann:
Ah sorry, the last time i checked master was behind 13_0_X but maybe i just didn't update my local branch.

PR #14271, opened 2025-09-10
Drop Node 16 and 18, remove engines.node

cb1kenobi:
Back in the day, we talked about publishing `android` and `iphone` (as well as `mobileweb`, `windows`, and `blackberry`) as individual npm packages. This never happened. The `engines` property in the top-level and platform `package.json` files have never been used.

In the top-level SDK's `package.json`, the `vendorDependencies` only has one property: `node`. It is referenced by the CLI, but it's not actually enforced. If the Node.js version is incompatible, it throws and the catch does nothing with the error. In other words, it's unused, but the CLI should probably and if it does, we might as well bump the Node version.

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -26,9 +26,6 @@
 		"android ndk": ">=r21 <=r22b",
 		"java": ">=11.x"
 	},
-	"engines": {
-		"node": ">=16.0.0"
-	},
 	"repository": {
 		"type": "git",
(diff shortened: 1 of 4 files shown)


m1ga (approved):
LGTM. Since the CLI is already node 20+ we don't need to support lower versions in the SDK

PR #14274, opened 2025-09-11
chore: remove interactive dismiss mode

hansemannn:
The dismiss mode is handled by the operating system in newer iOS versions, so a manual interception causes more problems than it solves and should therefore be discontinued. Spotted by @designbymind

### DIFF
--- a/apidoc/Titanium/UI/NavigationWindow.yml
+++ b/apidoc/Titanium/UI/NavigationWindow.yml
@@ -40,6 +40,10 @@ properties:
     platforms: [iphone, ipad, macos]
     since: "12.5.0"
     availability: creation
+    deprecated:
+      since: "13.0.0"
+      removed: "13.0.0"
+      notes: This is handled by the operation system in newer versions of iOS.
 
 methods:
   - name: closeWindow
--- a/apidoc/Titanium/UI/TabGroup.yml
+++ b/apidoc/Titanium/UI/TabGroup.yml
@@ -683,6 +683,10 @@ properties:
     platforms: [iphone, ipad, macos]
     since: "12.5.0"
     availability: creation
+    deprecated:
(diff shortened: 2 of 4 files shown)


designbymind:
**Hey Everyone!** — For TiDevs interested in additional infos on this, below are some simple tips to get this working in your apps 📲

**What happened?** — Apple finally agreed that being able to swipe from _**anywhere**_ on the screen (instead of just the left-most edge) to return to the previous screen in your `NavigationWindow` and `TabGroup` controllers is a better UX pattern and made this the default functionality in iOS 26+. This will enable your apps to work like some of the most popular apps out there, such as: _Facebook_, _Instagram_, _TikTok_, _Slack_, _Discord_, and many others. For anyone running iOS 26, you can test this functionality in many of Apple’s default apps, such as: _Settings_, _Notes_, _Music_, _Files_, _Podcasts_, and the _App Store_. Just swipe from the middle of the screen (or farther to the right) and you will be able to see how much easier it is to navigate your apps with just one hand 👆 

There are some caveats to making this possible with Ti.
**First:** if you have any custom gestures attached to your `Views` or are using built-in gestures (like swipe-to-delete actions on `ListView`/`TableView`), those gestures will take priority on their respective areas.
**Second:** if you’re using a custom NavBar or a custom ‘back button’ in the native NavBar, this functionality will not work. You must use the native NavBar with the `hidesBackButton` property (of your `Window`) set to `false`.
**Tip💡:** use the following `Window` property to hide the previous `Window`'s `title` property from being displayed in the back button (and only display the back button icon):
```
backButtonTitle: ''
```

**Questions?** — Drop a message on TiDev’s Slack 📝

**Swipe-to-close area iOS <18**
image

**Swipe-to-close area iOS 26**
image

PR #14277, opened 2025-09-13
feat(android): add parity for BlurView API, move iOS to Ti.UI namespace

hansemannn:
As discussed in https://github.com/m1ga/ti.blurview/issues/2

This pull request introduces a new cross-platform-style `BlurView` for Android, providing a background blur effect similar to iOS' `UIVisualEffectView`. It adds a new `Ti.UI.BlurView` component, exposes cross-platform blur effect constants, and updates build dependencies to support the feature. Documentation is also added for the new component and its properties.

It uses `Dimezis/BlurView` for maximized backwards compatiblity - and iOS has also moved to the same API!

```js
const win = Ti.UI.createWindow({ backgroundColor: '#fff' });
const target = Ti.UI.createView({ width: Ti.UI.FILL, height: Ti.UI.FILL });

target.add(Ti.UI.createImageView({
	image: '/DefaultIcon.png'
}));

const blur = Ti.UI.createBlurView({
	width: 200,
	height: 200,
	effect: Ti.UI.BLUR_EFFECT_STYLE_LIGHT
});

win.add(target);
win.add(blur);
win.open();
```

| Static | Animating |
|---|---|
| Screenshot_1757782898 | https://github.com/user-attachments/assets/1d6df851-4706-4d49-a970-2e8b19b7b64d |


android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIBlurView.java
```diff
@@ -0,0 +1,228 @@
...
+			.setOverlayColor(this.overlayColor);
+	}
+
+	private void applyEffectPresetIfAny()
+	{
+		// Map iOS-like styles to Android parameters.
+		// These constants are defined on Ti.UI.*
+		// Default values already set on fields.
+		switch (this.effectStyle) {
+			case 0: // EXTRA_LIGHT
```
m1ga: maybe use the constant in here instead of its values

android/build.gradle
```diff
@@ -24,6 +25,14 @@ allprojects {
 	repositories {
 		google()
 		mavenCentral()
+		maven { url 'https://jitpack.io' }
+	}
+
+	// Ensure BlurView v2 is used if other modules request a different version
+	configurations.all { cfg ->
+		resolutionStrategy {
+			force 'com.github.Dimezis:BlurView:version-2.0.5'
```
m1ga: this is not the latest version https://github.com/Dimezis/BlurView/releases
especially the never 3.0.0+ versions have some API 31 updates which sound nice ( "fast blurring pipeline for API 31+.").

The thing with these external jitpack modules is that it can cause build issues if they are not available anymore. 
I've not looked at the module license but I personally would rather use the source folder in the SDK instead so it's build it (if license match).

m1ga: https://github.com/Dimezis/BlurView/blob/master/LICENSE.md
it would be Apache too! So we could add https://github.com/Dimezis/BlurView/tree/master/library/src/main/java/eightbitlab/com/blurview to the SDK and use that in your files

hansemannn: Yep, I've moved from v3 to v2 on purpose, as v3 requires a fixed "target" on creation, making it much more restrictive regarding background views - something that iOS and the v2 version do not require. If you can find a way to solve this, go ahead! 👍🏼

m1ga: I see, I'll check it tomorrow.

V3 sounds nice, beside the last sentece :smile: 
 
> The new 3.0 version comes with a new API that's based on RenderNodes and RenderEffect (for API 31+ only, on older versions it falls back to the old code path).
The View snapshotting is now fully hardware-accelerated, it supports all of the previously unsupported targets.
The View snapshotting has basically zero overhead. The BlurTarget records the snapshot on a RenderNode, and then draws the same RenderNode on the system canvas.
The RenderNode snapshot is automatically updated whenever the View hierarchy changes, there's 0 additional invalidate() or draw() calls.
All this comes at the cost of a more complex API.

hansemannn:
@m1ga iOS is now also moved over - we have the unique chance to do deprecate the old one in 13.0.0, so I will restore the Ti.UI.iOS.BlurView deprecate it in 13.0.0 and then we can remove it in 14.0.0

android/build.gradle
m1ga: Ok, I see the issue with it now. When I create a BlurTarget and add the whole scene it will complain that the blurview is already added because it has to be outside of the target but our whole window has to be inside.
Think we'll keep it at the lower version for now.

That said: 
removing the gradle stuff and adding the whole eightbitlab folder (from the 2.0.6 source zip, library/src/main/java)  into `android/titanium/src/java/` works fine too. There are a few quick lint changes:
* right click the folder in Android Studio and run `Reformate Code` to fix most of it:
Bildschirmfoto_20250914_143332-1

* run `npm run lint:android` to see the rest of them (some lines are to long: add a break, some { should be in the same line and one string output can be in one line)
* edit `attrs.xml` (android/modules/ui/res/values/) and add 
```xml
	<declare-styleable name="BlurView">
		<attr name="blurOverlayColor" format="color"/>
	</declare-styleable>
```
* go to BlurView.java, line 54 and right click `R` and import the source

apidoc/Titanium/UI/BlurView.yml
```diff
@@ -0,0 +1,176 @@
...
+extends: Titanium.UI.View
+platforms: [android, iphone, ipad, macos]
+since: { android: "13.0.0" }
+properties:
+
+  - name: effect
+    summary: The blur effect to apply to the effect view.
+    description: |
+        On iOS, you can also use the Titanium.UI.iOS.BLUR_EFFECT_STYLE_* effects
+        that are platform-specific.
```
m1ga, line 17:
```suggestion
        On iOS, you can also use the Titanium.UI.iOS.BLUR_EFFECT_STYLE_* effects
        that are platform-specific.
        On Android it will add a default blurRadius of 16px. If you want to change the value you have to use `blurRadius` and `overlayColor` instead of the `effect` preset.
```

That took me a bit to understand that I can't change the blurRadius in an `effect` preset. So I guess it should be in the docs.

android/build.gradle
hansemannn: Why would we want to manually copy over the source files? Maintaining it through Gradle seems like the perfect solution, especially for future updates in case we find a solution to move to 3.x at some point.

m1ga: so we don't have an external resource in every app that can potentially break building even if you don't use blurview  in your app.

E.g. if I change the version number in the app gradle to a invalid version and try to build a test app it will stop with:
```
[ERROR] [GRADLE] 
[ERROR] [GRADLE] FAILURE: Build failed with an exception.
[ERROR] [GRADLE] 
[ERROR] [GRADLE] Configuration cache state could not be cached: field `__librarySourceSets__` of task `:app:mapDebugSourceSetPaths` of type `com.android.build.gradle.tasks.MapSourceSetPathsTask`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
[ERROR] [GRADLE] > Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
[ERROR] [GRADLE]    > Could not find com.github.Dimezis:BlurView:version-1.0.5.
[ERROR] [GRADLE]      Searched in the following locations:
[ERROR] [GRADLE]        - https://dl.google.com/dl/android/maven2/com/github/Dimezis/BlurView/version-1.0.5/BlurView-version-1.0.5.pom
[ERROR] [GRADLE]        - https://repo.maven.apache.org/maven2/com/github/Dimezis/BlurView/version-1.0.5/BlurView-version-1.0.5.pom
[ERROR] [GRADLE]        - file:/home/miga/.titanium/mobilesdk/linux/13.1.0/android/m2repository/com/github/Dimezis/BlurView/version-1.0.5/BlurView-version-1.0.5.pom
[ERROR] [GRADLE]      Required by:
[ERROR] [GRADLE]          project :app > org.appcelerator:titanium:13.1.0
[ERROR] [GRADLE] 
[ERROR] [GRADLE] * Try:
[ERROR] [GRADLE] > Run with --stacktrace option to get the stack trace.
[ERROR] [GRADLE] > Run with --info or --debug option to get more log output.
[ERROR] [GRADLE] > Run with --scan to get full insights.
[ERROR] [GRADLE] > Get more help at https://help.gradle.org.
[ERROR] [GRADLE] 
[ERROR] [GRADLE] BUILD FAILED in 7s
```

and jitpack had issues with availability each year, last time in June (see https://github.com/jitpack/jitpack.io/issues?q=is%3Aissue%20state%3Aclosed%20down%20sort%3Acomments-desc )
and some modules were removed or moved to maven. All that would stop people from building apps.

If this is in a module we can quicker switch to a different repo or module or people can just remove it to build the app again. 

Having the source in the SDK would eliminate all that.

hansemannn: Thanks, that makes sense. Then it may even make sense to fully move it to a module (cross platform), easing maintenance for new glass effect APIs in the future as well. WDYT? Copying source files is in most cases not encouraged, but if you can commit the changes with formatting etc (you have write access), it should be fine

m1ga: I've pushed my changes but you are right. Maybe this should be a module.

I'll revise https://github.com/tidev/titanium-sdk/pull/13487 next week again to see if we can have a version with this for a quick blur effect inside the SDK with a hint to a module for more features and performance.

android/modules/ui/res/values/attrs.xml
```diff
@@ -10,4 +10,7 @@
 	<attr name="ti_footerTitleColor" format="color" />
 	<attr name="ti_footerTitleBackground" format="reference" />
 	<attr name="ti_footerTitleBackgroundColor" format="color" />
+	<declare-styleable name="BlurView">
```
hansemannn, line 13: prefix with `Ti` to prevent clashes with other libs

android/modules/ui/res/values/attrs.xml
```diff
@@ -10,4 +10,7 @@
...
+		<attr name="blurOverlayColor" format="color"/>
```
hansemannn, line 14:
```suggestion
		<attr name="ti_blurOverlayColor" format="color" />
```

android/titanium/src/java/eightbitlab/com/blurview/BlurView.java
```diff
@@ -0,0 +1,171 @@
+package eightbitlab.com.blurview;
```
hansemannn, line 1: Can you add a readme file to the package to link to the repo and explain why it's bundled like this?

android/titanium/src/java/eightbitlab/com/blurview/BlurView.java
```diff
@@ -0,0 +1,171 @@
...
+	public BlurView(Context context, AttributeSet attrs, int defStyleAttr)
+	{
+		super(context, attrs, defStyleAttr);
+		init(attrs, defStyleAttr);
+	}
+
+	private void init(AttributeSet attrs, int defStyleAttr)
+	{
+		TypedArray a = getContext().obtainStyledAttributes(attrs, R.styleable.BlurView, defStyleAttr, 0);
+		overlayColor = a.getColor(R.styleable.BlurView_blurOverlayColor, TRANSPARENT);
```
hansemannn, line 55: Should be like this then?
```suggestion
		overlayColor = a.getColor(R.styleable.TiBlurView_ti_blurOverlayColor, TRANSPARENT);
```

hansemannn: Do we even need that in the XML? Devs should rather use the top-level APIs, making this more clean / genuine as well

hansemannn:
Let's move it to a module - but not in 13.0.0 anymore, the release is big enough already. We can keep this open and I'll prepare a blur view module under tidev that has the sources of this PR bundled in one repo. Alright?

m1ga:
> Let's move it to a module - but not in 13.0.0 anymore, the release is big enough already. We can keep this open and I'll prepare a blur view module under tidev that has the sources of this PR bundled in one repo. Alright?

sounds good. There we could also use the gradle module to reduce the code size if you want or this internal code setup

PR #14280, opened 2025-09-17
14279 Patch ActivityKit so it works with macCatalyst

narbs:
**GitHub discussions:** https://github.com/tidev/titanium-sdk/issues/14279

**Optional Description:**

The macCatalyst build fails for 13.x because ActivityKit is not supported. TiActivityAttributes and TiAppiOSActivityAttributesProxy+startActivity need to exclude macCatalyst specifically - #if canImport(ActivityKit) is not enough.

I have come up with a patch for this issue that fixes the problem by adding #if !targetEnvironment(macCatalyst) to these files.

Apps now build with macCatalyst with these fixes, and continues to build under iOS, using Xcode 16.4 and 26.0

See issue above

### DIFF
--- a/iphone/Classes/TiActivityAttributes.swift
+++ b/iphone/Classes/TiActivityAttributes.swift
@@ -6,6 +6,7 @@
  */
 
 #if canImport(ActivityKit)
+#if !targetEnvironment(macCatalyst)
 import Foundation
 import ActivityKit
 
@@ -15,4 +16,5 @@ public struct TiActivityAttributes: ActivityAttributes {
     var value: [String: String]
   }
 }
+#endif
 #endif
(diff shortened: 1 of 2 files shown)


narbs:
I have signed up to be a contributor, and my name is in the list: https://github.com/tidev/organization-docs/blob/main/AUTHORIZED_CONTRIBUTORS.md

cb1kenobi:
@narbs apologies, you did sign a CLA, but I somehow missed it when I migrated away from Docusign. You're good to go!

hansemannn (approved):
Execellent catch, thank you @narbs! I would have thought that the `canImport` would catch target incompatbilities already, but it seems like it doesn't.

PR #14282, opened 2025-09-20
[14281]: fix (ios): ensure eventStoreChanged notification is not over-registered

narbs:
**GitHub discussions:** https://github.com/tidev/titanium-sdk/discussions/14281

**Optional Description:**
Ensure eventStoreChanged notification is not over-registered. See issue for details


hansemannn (approved):
Thanks - makes total sense to register it during init only!

PR #14293, opened 2025-09-27
fix(ios): simulator background modes warnings

### LINKED ISSUE #14288: iOS: Background modes warnings in simulator

### Description

The iOS simulator shows the following warnings when an app runs **without**  `<key>UIBackgroundModes</key>` in `tiapp.xml`:

```
[DEBUG] You've implemented -[<UIApplicationDelegate> application:performFetchWithCompletionHandler:], but you still need to add "fetch" to the list of your supported UIBackgroundModes in your Info.plist.
[DEBUG] You've implemented -[<UIApplicationDelegate> application:didReceiveRemoteNotification:fetchCompletionHandler:], but you still need to add "remote-notification" to the list of your supported UIBackgroundModes in your Info.plist.
```

There are two symbol macros `USE_TI_FETCH` and `USE_TI_SILENTPUSH` and their absence should actually
prevent the implementation of `application:performFetchWithCompletionHandler:` resp. `application:didReceiveRemoteNotification:fetchCompletionHandler:`.

### Expected Behavior

no warnings

### Actual behavior

2 warnings

### Reproducible sample

- use an app without `<key>UIBackgroundModes</key>` in `tiapp.xml`

### Steps to reproduce

1. Set log level to `debug`.
2. Run an app that doesn't use any background mode in the iOS simulator.

### Platform

iOS

### SDK version you are using

13.0.0

_No response_

hbugdoll:
Fixes https://github.com/tidev/titanium-sdk/issues/14288.

**Description:**

- removed symbol macros `USE_TI_FETCH` and `USE_TI_SILENTPUSH` from `iphone/iphone/project.xcconfig`
  - so the program logic in `_build.js` (function `processTiSymbols()`) works as intended, see [comment](https://github.com/tidev/titanium-sdk/issues/14288#issuecomment-3333236552)
  - so the implementation of background mode handlers can be controlled via `tiapp.xml` (`<key>UIBackgroundModes</key>`) in the iOS simulator too

**Note:**

It seems that GitHub's diff view has a problem with long lines.
The changes inside the long line aren't highlighted:

Bildschirmfoto 2025-09-27 um 00 10 11

compared to VSC:
Bildschirmfoto 2025-09-27 um 00 16 05


hansemannn (changes requested):
By removing them, you disable the use of background APIs on the Simulator, as the pre-compile macros are used at build time to enable the features. A proper fix could be to remove the background modes by default, so developers who are using teh Simulator can enable them if needed. Note that the warning is there since ages and does not affect the apps.

hbugdoll:
Please see my [comment](https://github.com/tidev/titanium-sdk/issues/14288#issuecomment-3333236552) linked above.

`USE_TI_FETCH` and `USE_TI_SILENTPUSH` are/should be added (just like `TI_USE_AUTOLAYOUT`) to the list of symbol macros depending on the settings in `tiapp.xml`.
This is the intended behavior of: 
https://github.com/tidev/titanium-sdk/blob/03339a16dbd8153e5f50617b825945145c5c1214/iphone/cli/commands/_build.js#L6718-L6780



However, the warning is currently misleading. It's false positive in case of absence of `<key>UIBackgroundModes</key>`.

hansemannn:
Did you test this on Simulator builds running from Xcode? As they are now compiled out and Xcode does not use the CLI, testing background modes in development will not be possible anymore - that's what I meant. I just don't see a benefit in removing it, as it causes no harm and is not relevant for device / production builds (where it would indeed matter).

hbugdoll:
Thanks for yours response.

> Did you test this on Simulator builds running from Xcode? As they are now compiled out and Xcode does not use the CLI, testing background modes in development will not be possible anymore - that's what I meant.

No, default usage in VS Code with Ti Extention building app for simulator.

> I just don't see a benefit in removing it, as it causes no harm and is not relevant for device / production builds (where it would indeed matter).

My bad feelings about this is that the warning included an instruction in the sense of “you have to” (_“... but you still need to add ...”_).

For better illustration, I would like to give the following example scenario:
A developer tests his app, which does not require background modes, in the iOS Simulator. He sees the warning and is confused.
Due to concerns that the app might be rejected when uploaded to the App Store, he then unnecessarily activates the background modes via `tiapp.xml`.
I myself was unsure about this! Not everyone will then perform a deeper code analysis and understand what is really going on (resp. has no knowledge of the symbol macros).

hansemannn:
Yep, for production, I agree (and you definitely shouldn't include it if not used, as your app would be rejected otherwise). For development, the only reason would be the possible confusion, which I can also understand.

Let's merge this and hope that core-developers remember to re-enable it when working on the APIs. As there weren't too many changes to these APIs over the last years, it's fine.

hansemannn (approved):

hbugdoll:
Thanks a lot for merging 👍

PR #14296, opened 2025-09-29
fix: properly set Titanium SDK version for runtime usage

### LINKED ISSUE #13254: iOS: Ti.userAgent is incorrect

### Description

The `Ti.userAgent` is set incorrectly on iOS.

### Expected Behavior

The `Ti.userAgent` property should be `Appcelerator Titanium/10.2.0 (...)`.

### Actual behavior

 It returns `Appcelerator Titanium/0.0.0 (...)` on iOS. The same works fine on Android.

### Reproducible sample

console.warn(Ti.userAgent);

### Steps to reproduce

Run the above on iOS.

### Platform

iOS

### SDK version you are using

10.2.0

hansemannn:
Fixes https://github.com/tidev/titanium-sdk/issues/13254

## Test Case

```js

    text: ${Ti.userAgent} / ${Ti.version}`
});

window.add(label);
window.open();
```

### Console
Screenshot 2025-09-29 at 15 40 54

### API
Simulator Screenshot - iPhone 17 Pro - 2025-09-29 at 15 40 41


cb1kenobi (approved):
So you're moving `TI_VERSION` from the config file to the project file. I can see this working, so approved, but I don't have a machine to test (yet).

hansemannn:
I am moving it from the main app target to the TitaniumKit target. As TitaniumKit is compiled during the SDK build (prior to a release, by the CI), it needs to set the pre-compile macro before that. Right now, it was set in the main app target during app build, causing it to be ignored (and displying "0.0.0").

PR #14298, opened 2025-10-01
fix(ios): reduced TitaniumKit build warnings & fixed TiBlob crash

hbugdoll:
Related to https://github.com/tidev/titanium-sdk/issues/14290. Edit: Also fixes https://github.com/tidev/titanium-sdk/issues/14306.
 
**Description**

Removed over 40 warnings during TitaniumKit build process:
- **Class method not found warnings**
  - `+resourceBasedURL:baseURL:` in `iphone/TitaniumKit/TitaniumKit/Sources/API/ScriptModule.m`
    - added missing header file
- **Instance method not found warnings**
  - `-JSValueInContext:` in `iphone/TitaniumKit/TitaniumKit/Sources/API/KrollModule.m`
    - added explicit cast
  - `-applyProperties:` in `iphone/TitaniumKit/TitaniumKit/Sources/API/TiProxy.h`
    - added to header file
- **Incompatible pointer types warnings**
  - `JSValueRef *` vs. `JSValueRef _Null_unspecified` in `iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m`
    - use of `JSValueRef`, which is already a pointer (Edit: **crash relevant**)
    - further improvements here:
       - added explicit initialization of `JSValueRef`
       - use of `[self currentContext]` instead of `JSContext.currentContext` in `toArrayBuffer` to make it thread-safe
       - added exception handler in `toArrayBuffer`
       - added also `free(arrayBytes)` in `arrayBuffer` when the exception is raised
  - `CAShapeLayer *` vs. `CALayer *` in `iphone/TitaniumKit/TitaniumKit/Sources/API/TiUIView.m`
    - added explicit cast
- **Receiver is a forward class and corresponding interface may not exist warnings**
  - `TiHost` in `iphone/TitaniumKit/TitaniumKit/Sources/API/ScriptModule.m`
    - added missing header file
- **Receiver type warnings**
  - `NSUncaughtExceptionHandler *` in `iphone/TitaniumKit/TitaniumKit/Sources/API/TiExceptionHandler.m`
    - added explicit cast
- **Enumeration value not handled in switch warnings**
  - `TiBlobTypeSystemImage` in `iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m`
    - added missing `default` case


iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
```diff
@@ -481,8 +484,7 @@ - (JSValue *)toArrayBuffer
 
   // Now make an ArrayBuffer with the copied bytes
   JSContext *context = JSContext.currentContext;
-  JSValueRef *exception;
-  JSObjectRef arrayBuffer = JSObjectMakeArrayBufferWithBytesNoCopy(context.JSGlobalContextRef, arrayBytes, len, jsArrayBufferFreeDeallocator, nil, exception);
+  JSObjectRef arrayBuffer = JSObjectMakeArrayBufferWithBytesNoCopy(context.JSGlobalContextRef, arrayBytes, len, jsArrayBufferFreeDeallocator, nil, NULL);
```
hansemannn: Why passing NULL here instead of initializing the exception to NULL and passing the reference like you did below?

hansemannn (changes requested):

iphone/TitaniumKit/TitaniumKit/Sources/API/TiBlob.m
hbugdoll: Because `exception` is not evaluated resp. used further here.

hansemannn: I think it should though - any exception handling can be useful

hbugdoll: But how?
Handling it like in `arrayBuffer` via `KrollPromise` then it would make `toArrayBuffer` and `arrayBuffer` redundant.

Handling it with the JavaScriptCore API (I am not familiar with it), ChatGPT suggests the following:

```
    if (exception) {
        JSStringRef exceptionStr = JSValueToStringCopy(context.JSGlobalContextRef, exception, NULL);
        NSString *message = (__bridge_transfer NSString *)JSStringCopyCFString(kCFAllocatorDefault, exceptionStr);
        JSStringRelease(exceptionStr);

        // e.g. handle message
        
        free(arrayBytes);
        return nil;
    }
```

hansemannn: Mhh, we are there even both then, if the only difference is the exception?

hbugdoll: Oh, there is still a difference &ndash; a different context may be used (`JSContext.currentContext` vs. `[self currentContext]`).

hbugdoll: I've done the following changes:
- use of `[self currentContext]` instead of `JSContext.currentContext` in `toArrayBuffer` to make it thread-safe
- added exception handler in `toArrayBuffer`
- added also `free(arrayBytes)` in `arrayBuffer` when the exception is raised

hbugdoll: What do you think about the latest commit https://github.com/tidev/titanium-sdk/pull/14298/changes/c0dd252ebcd308a89f99d4b2c06ceaf9b55b69e3?

It would be cool if the iOS test would run until the end again...

PR #14300, opened 2025-10-02
chore: remove URL question in 'ti create'

### LINKED ISSUE #845: Optimize app creation steps
Do we still need the `Your company/personal URL` step in `ti create`?
https://github.com/tidev/titanium-cli/blob/16a96975b4909e4ae553163b8ea9332ea6937c7f/test/mock-sdk/cli/lib/creator.js#L276-L302
you can always set it in your tiapp.xml and I don't think it is actually used in the app/play store.

If it is not relevant I would say we can remove that step in the `ti create` flow.

m1ga:
fixes https://github.com/tidev/titanium-cli/issues/845

When you run `ti create` the current questions are:
* sdk version
* type (app, module)
* target platform
* project name
* app id
* **company URL**
* directory

The *company URL* is not used inside the app project and can be set in the tiapp.xml if needed. But I don't think it should be part of the `ti create` menu as it is not that important. The CLI parameter `-u` is still there so it shouldn't break any editor package

### DIFF
--- a/cli/lib/creators/app.js
+++ b/cli/lib/creators/app.js
@@ -79,7 +79,6 @@ AppCreator.prototype.init = function init() {
 			name:          this.configOptionName(140),
 			platforms:     this.configOptionPlatforms(120),
 			template:      this.configOptionTemplate(110),
-			url:           this.configOptionUrl(160),
 			'workspace-dir': this.configOptionWorkspaceDir(170)
 		}
 	};


cb1kenobi:
Can we also remove the `configOptionUrl()` method? https://github.com/tidev/titanium-sdk/blob/0c9cf2bb7cf50eae47b31614ffe1052f0e295d98/cli/lib/creator.js#L422-L455

m1ga:
that would remove the `ti create -u` parameter. Not sure if we want to do that as this might be a breaking change

cb1kenobi:
> that would remove the `ti create -u` parameter. Not sure if we want to do that as this might be a breaking change

Ah, good point. Sigh.

cb1kenobi (approved):

PR #14302, opened 2025-10-02
build(ios): dropped iTunes artwork generation

hbugdoll:
**Introduction:**

- Previously, the files `iTunesArtwork` (512x512px) and `iTunesArtwork@2x` (1024x1024px) &ndash; both PNG files without file extension &ndash; were required for Ad Hoc installation using iTunes to display the app icon.
- Ad Hoc installation for apps was removed in iTunes 12.7 shipped with macOS 10.13 in **2017**.
- This was the time of iOS 11, now `minIosVersion` is `15.0`.
- And all important MDM systems should be now updated to not require iTunes artwork files anymore.

**Description:**

- removed outdated  iTunes artwork generation during build process


cb1kenobi (approved):
This is awesome! Thanks @@hbugdoll 

@hansemannn you good with this?

PR #14303, opened 2025-10-03
chore(android): update minSdk to API Level 24

m1ga:
Setting minSDK to API Level 24 (was 21). API Levels overview: https://apilevels.com/

According to https://support.google.com/googleplay/answer/9037938?hl=en support for Play Services is already >= 23
> Google Play services are only supported on devices with Android 6.0 (API level 23) or higher. [Learn how to check and update your Android version](https://support.google.com/android/answer/7680439).

And the latest Firebase FCM Module is also now 23+: https://firebase.google.com/support/release-notes/android#messaging_v25-0-0

We should follow this at some point and make the minSDK 23.

**Note:** the code changes are just removing `if (Build.VERSION.SDK_INT >= 23) {}` checks at places where we had those and they are not needed. In the Git preview it looks like a lot more changed.

People that currently have a fixed version like 22 will see:
```
[ERROR] The minimum supported SDK API version must be 24 or newer, but is currently set to 22

Update the android:minSdkVersion in the tiapp.xml or custom AndroidManifest to at least 24:

<ti:app xmlns:ti="http://ti.tidev.io">
    <android>
        <manifest>
            <uses-sdk android:minSdkVersion="24" />
        </manifest>
    </android>
</ti:app>
```


prashantsaini1:
@m1ga There could be some apps with `minSdkVersion: 21` and devices running on Android API 21 out there. This would stop running their apps on those devices. Is there any concrete reason we are removing 21 support?

m1ga:
don't worry. doesn't have to be merged right away or even this year. But as you can see: Google is updating the Firebase modules to be minSdk 23. And it looks like androidx and latest Material are also already >=23 (https://issuetracker.google.com/issues/380448311)
So I'm sure other parts will follow soon and then it's required by the store (again: not yet). Then we have this ready and can use it.

prashantsaini1:
@m1ga Keep it as Draft until it's mandatory and to make any further changes if required just before merge?

m1ga:
good point, done :+1:

hbugdoll:
Nice overview of the Android versions fragmentation: https://nfctools.net/android/versions.php

hbugdoll:
Just for comparison:
- _React Native_ bumped `minSdkVersion` from `21` to `23` in 2023 (https://github.com/facebook/react-native/commit/70fcb1ebf664203ca9934fa12376cfa8a32d373b), and to `24` in 2024 (https://github.com/facebook/react-native/pull/45906/commits/8f57212a7ed5df6c29a83c9cc6877365d2b5e4a2)
- _Flutter_ bumped `minSdkVersion` from `21` to `24` in July 2025 (https://github.com/flutter/flutter/commit/1dfb59d2811ce784bfa8a11231509bd835fdb4a7)

m1ga:
Thanks for the research @hbugdoll :+1: 

once we have more and more modules that require it to be minsdk 23 we should think about merging this PR. Pretty sure the playstore will required it soon too.

hansemannn:
We can do this now - happy to approve!

m1ga:
@hansemannn go ahead. I think it should be fine to raise the minsdk to 23. Even that is Android 6 from 2015 :smile:

hansemannn:
I'm actually also okay with 24 - thats what we also target. And we'd be aligned with other frameworks.

m1ga:
let me check, we can remove more checks if we go to 23

hansemannn (approved):

PR #14304, opened 2025-10-07
chore: update iOS workflow

m1ga:
https://github.com/actions/runner-images/releases/tag/macos-15-arm64%2F20250928.2397

`/Applications/Xcode_26.0.0.app` was removed and replace with `/Applications/Xcode_26.0.app` for 26.0.1


cb1kenobi (approved):
Ah! I was wondering where the `26.0.0` was coming from haha

PR #14308, opened 2025-10-11
chore: rename APS classes and remove unused analytics class

m1ga:
Some analytics pages (https://reports.exodus-privacy.eu.org/en/trackers/255/) still list Titanium apps as potentially tracking apps because they scan for `APS` and `Analytics` in the code and find these names in Ti apps even if those don't do anything.

This PR will remove the empty AnalyticsModule file and rename APS parts:
* AnalyticsModule is just returning deprecation strings, so I remove that
* Renaming the `APS*` classes to `TiSession`/`TiSessionMeta` as it will just return the session ID and some meta info like platform, OS, architecture and so on.


cb1kenobi (approved):

cb1kenobi:
LGTM!

hbugdoll:
@m1ga Can you please check whether the following three lines are relevant in this context?

https://github.com/tidev/titanium-sdk/blob/40a61fa76765bdb2ae46d04207265cdee050a7cd/STARTUP.md?plain=1#L13

https://github.com/tidev/titanium-sdk/blob/40a61fa76765bdb2ae46d04207265cdee050a7cd/android/titanium/build.gradle#L98

https://github.com/tidev/titanium-sdk/blob/40a61fa76765bdb2ae46d04207265cdee050a7cd/android/cli/commands/_build.js#L2687

m1ga:
@hbugdoll the folder is still there but no files in it so the 2nd line doesn't make an issue. Even deleting the folder won't show an error and the SDK builds fine. So I guess it will just skip that line. Will do some tests if it can be removed to have it cleaner.

The last one is a check for API and permissions if I see it correctly. They just analyze the used methods.

Guess the line from the startup.md can also be removed.

I'll create a PR for the two parts later so we have a cleaner version. Thanks for the notes :+1:

hbugdoll:
Thanks @m1ga for the details!

cb1kenobi:
Initially, `tiSymbols` was added so Appc could determine which APIs are being used by an app and try to detect if we could change or deprecate APIs. For example, if we had an iOS specific constant, then added Android parity, could we move it to the common namespace without breaking millions of apps. It was also more useful back when we also supported Mobile Web, Blackberry, and Windows Phone.

It looks like `tiSymbols` is used today to detect possible misconfigured app permissions, which is a nice feature. 

As far as collecting the `tiSymbols`, I don't think we ever did anything with the data. The captured data wasn't good data. We could detect `Ti.UI.createWindow()`, but we couldn't detect `window.open()`. Nowadays, that would be doable.

hbugdoll:
Thanks @cb1kenobi for all the background information.

PR #14311, opened 2025-10-12
feat(android): parity for `screenshotcaptured`  event

m1ga:
Android parity for `screenshotcaptured` using Android 14 `registerScreenCaptureCallback`:  https://developer.android.com/about/versions/14/features/screenshot-detection#java 

**Changes:**
* add Android parity for `screenshotcaptured` event
* move `Ti.App.iOS.screenshotcaptured` to `Ti.App.screenshotcaptured`
* add deprecation message for `Ti.App.iOS.screenshotcaptured`

**Test:**
```js
	id: "win1",
	backgroundColor: "#fff"
});
	top: 140,
	title: "open"
})
		backgroundColor: "#fff",
		id: "win2"
	})
		color: "#000",
		text: "app even is still fired. Close window and check the log"
	});
	win2.open();
	win2.add(lbl2);
})
var lbl = Ti.UI.createLabel({
	color: "#000",
	text: ""
});
win.add([btn, lbl]);
win.open();

// iOS parity
Ti.App.addEventListener("screenshotcaptured", function(e) {
	console.log("Ti.App event: " + JSON.stringify(e));
	lbl.text += "Ti.App event: " + e.source.id + "\n";
})
if (OS_IOS) {
	Ti.App.iOS.addEventListener("screenshotcaptured", function(e) {
		console.log("Old Ti.App.iOS event: " + JSON.stringify(e));
		lbl.text += "Old Ti.App.iOS event\n";
	})
}

```
* and add `<uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE"/>` to your tiapp.xml
* Run the app and take a screenshot. 
* log and label will show that the event is fired
* click the open button to open a new window
* take a screenshot
* the app event is fired

### DIFF
--- a/apidoc/Titanium/App/App.yml
+++ b/apidoc/Titanium/App/App.yml
@@ -263,11 +263,11 @@ events:
 
   - name: screenshotcaptured
     summary: Fired after the user takes a screenshot, e.g. by pressing both the home and lock screen buttons.
-    descition: |
+    description: |
         Android is only available on devices with Android 14+ and you'll have to add the following permission
         to your tiapp.xml: `<uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE"/>`
     platforms: [android, iphone, ipad, macos]
-    since: {android: "13.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
+    since: 13.1.0
 
   - name: uncaughtException
     summary: Fired when an uncaught JavaScript exception occurs.
--- a/apidoc/Titanium/App/App.yml
+++ b/apidoc/Titanium/App/App.yml
@@ -268,6 +268,7 @@ events:
         to your tiapp.xml: `<uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE"/>`
     platforms: [android, iphone, ipad, macos]
     since: 13.1.0
+    osver: {android: {min: "14.0"}}
 
   - name: uncaughtException
     summary: Fired when an uncaught JavaScript exception occurs.
--- a/apidoc/Titanium/App/App.yml
+++ b/apidoc/Titanium/App/App.yml
@@ -261,6 +261,14 @@ events:
     platforms: [iphone, android]
     since: {android: "3.3.0"}
 
+  - name: screenshotcaptured
+    summary: Fired after the user takes a screenshot, e.g. by pressing both the home and lock screen buttons.
+    descition: |
+        Android is only available on devices with Android 14+ and you'll have to add the following permission
+        to your tiapp.xml: `<uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE"/>`
+    platforms: [android, iphone, ipad, macos]
+    since: {android: "13.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
+
   - name: uncaughtException
     summary: Fired when an uncaught JavaScript exception occurs.
     description: |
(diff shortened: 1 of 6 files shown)


prashantsaini1:
@m1ga 
1. We are firing `Ti.App` level events generated from per-activity level, it would be safe to test that the screen-capture callback is always invoked by the Android OS only when that activity is visible.
2. After the above point, we should also let users know that which activity/ti-window this event is being fired for.
    - e.g. A window where some private content is shown and user takes screenshot, the app may only want to handle that specific use-case, without worrying about rest of the screens.
    - It can be achieved by firing `ti-window` level event also, and keeping the `Ti.App` event as is. So this covers both use cases - if app just needs to know app-wide, it'll be through Ti.App, and if required for only specific window, then using the window's `screenshot` event. Just need to add this info in docs?

m1ga:
Thanks for the comments!

I have to check how the iOS event is fired. I've added the window event and changed the example.
Currently it behaves like this:
* Ti.App event: will fire in all windows
* win event: will only fire if the window is open, if you open the 2nd window it won't fire in the 1st window

Update: the Ti.App event behaves the same now on both platforms. Android will also have the per Window event.

prashantsaini1 (approved):
LGTM!

iphone/Classes/TiAppiOSProxy.m
```diff
@@ -217,6 +217,7 @@ - (void)didChangeTraitCollection:(NSNotification *)info
 
 - (void)didTakeScreenshot:(NSNotification *)info
 {
+  DEPRECATED_REPLACED(@"App.iOS.screenshotcaptured", @"9.1.0", @"App.screenshotcaptured");
```
m1ga:
```suggestion
  DEPRECATED_REPLACED(@"App.iOS.screenshotcaptured", @"13.1.0", @"App.screenshotcaptured");
```

apidoc/Titanium/UI/Window.yml
```diff
@@ -741,6 +741,14 @@ events:
         The listener for this event must be defined before this window
         is opened.
 
+  - name: screenshotcaptured
```
hansemannn: Why is this needed on a window instance and Android only? Isn't it on Ti.App now?

apidoc/Titanium/App/App.yml
```diff
@@ -261,6 +261,14 @@ events:
     platforms: [iphone, android]
     since: {android: "3.3.0"}
 
+  - name: screenshotcaptured
+    summary: Fired after the user takes a screenshot, e.g. by pressing both the home and lock screen buttons.
+    description: |
+        Android is only available on devices with Android 14+ and you'll have to add the following permission
+        to your tiapp.xml: `<uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE"/>`
+    platforms: [android, iphone, ipad, macos]
```
hansemannn, line 269: add `minver` for Android

apidoc/Titanium/UI/Window.yml
m1ga: it came for free as it is registered on the baseactivity and that way you use it on one window if needed and don't need to make the checks yourself to see if it is active and in front ([see this comment](https://github.com/tidev/titanium-sdk/pull/14311#issuecomment-3396006674)).

hansemannn: Why not just pass the window proxy as the "source" then? That's how they are usually identified

m1ga: you are right, sounds better. The Android event will now have current window as a `source`. iOS still needs to update the source.

apidoc/Titanium/App/App.yml
```diff
@@ -261,6 +261,15 @@ events:
...
```
hansemannn, line 269: Document the "source" on Android for windows

hansemannn (approved):

PR #14318, opened 2025-10-16
feat: added trace log option for tests

hbugdoll:
**Introduction:**

Due to the current issues with errors and warnings at the `trace` log level, it is difficult to investigate problems when building the unit tests/mocha test app.
In this context, the log level is currently fixed at `info` in https://github.com/tidev/titanium-sdk/blob/214b762441e5bfa235b0f31fde94f857286ea61a/build/lib/test/test.js#L405

**Description:**

- added log-level option for scons tests (default: `info`)
- BTW added some missing doc comments
- added scripts:
  - `test:android:trace`
  - `test:iphone:trace` (plus "alias" `test:ios:trace`)
  - `test:ipad:trace`
  - `test:mac:trace`

**Bonus:**

- removed script `test:integration:onlyFailed` which couldn't run alone
  - used by`test:android:onlyFailed` and `test:iphone:onlyFailed`
- added scripts for completeness:
  - `test:ios:onlyFailed`
  - `test:ipad:onlyFailed`
  - `test:mac:onlyFailed`

### DIFF
--- a/build/lib/test/test.js
+++ b/build/lib/test/test.js
@@ -62,9 +62,10 @@ let showFailedOnly = false;
  * @param {string} [snapshotDir='../../../tests/Resources'] directory to place generated snapshot images
  * @param {string} [failedOnly] Show only failed tests
  * @param {string} [sdkVersion] The SDK version to use
+ * @param {string} [logLevel] The log level
  * @returns {Promise<object>}
  */
-export async function test(platforms, target, deviceId, deployType, deviceFamily, junitPrefix, snapshotDir = path.join(__dirname, '../../../tests/Resources'), failedOnly, sdkVersion) {
+export async function test(platforms, target, deviceId, deployType, deviceFamily, junitPrefix, snapshotDir = path.join(__dirname, '../../../tests/Resources'), failedOnly, sdkVersion, logLevel) {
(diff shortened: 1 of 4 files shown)


build/lib/test/test.js
```diff
@@ -402,7 +404,7 @@ async function runBuild(platform, target, deviceId, deployType, deviceFamily, sn
 		'--platform', platform,
 		'--target', target,
 		'--sdk', sdkVersion,
-		'--log-level', 'info'
+		'--log-level', logLevel
```
m1ga, line 407: not tested it yet but is the default `info` if you don't set it somewhere? Or is it undefined/empty and it needs to be set as a default value in the parameter list to `, logLevel = 'info')` ?

hbugdoll: Thanks @m1ga for reviewing.

The default value is set here:
https://github.com/tidev/titanium-sdk/blob/23b207b8d7a5832019d4cf8f79eb13afa5c79667/build/scons-test.js#L13

See also https://github.com/tj/commander.js?tab=readme-ov-file#default-option-value.

m1ga (approved):
tested the android parts and they look fine.

iOS looks good to but I didn't run all the tests

PR #14325, opened 2025-10-23
fix(android): fix screenCaptured crash on Android 13 and lower

m1ga:
Startup crash introduced by https://github.com/tidev/titanium-sdk/pull/14311 and devices with Android 13 or lower. `new Activity.ScreenCaptureCallback()` is not available on these devices but it still tries to create the variable.

**Fix:** 
initialize the variable in onCreate and add a Android SDK version check around it 

**Test:**
Start any app on an Android 13 device. Master will crash right away.
```
10-23 16:37:55.291 28706 28706 E AndroidRuntime: java.lang.NoClassDefFoundError: Failed resolution of: Landroid/app/Activity$ScreenCaptureCallback;
10-23 16:37:55.291 28706 28706 E AndroidRuntime:        at java.lang.Class.newInstance(Native Method)
```


cb1kenobi (approved):
LGTM!

PR #14327, opened 2025-10-26
feat(ios): support swift package manager (SPM) module dependencies

hansemannn:
This pull request introduces first-class support for Swift Package Manager (SPM) dependencies in Titanium iOS modules and applications. The main changes add a new `spm.json` file for module authors to declare Swift package dependencies, embed this metadata in module zips, and update the build system to inject required packages into the app's Xcode project. Additionally, documentation and templates are updated to guide developers on using this feature.

Note: This PR currently targets 13_0_X, as master seem to be partially broken for me. I'll add more comments about that soon (cc @cb1kenobi), but it seems like the app.js / `Resources/*` file handling is not working as expected. After it's addressed, I can migrate the changes to the new master structure. As this module is needed in production asap, targeting 13_0_X will make sure it integrates properly.

Also, I think that the CLI / dependency changes should not have landed in master, as they are major breaking changes that should go into a `next branch`. Alternatively, we could use the commit before these changes as 13_1_X and bump master to 14.0.0. Let me know your thoughts.


cb1kenobi:
The dependencies we updated were not a breaking change. We made sure of that. We can bump majors of dependencies as long as we don't break the SDK. If the SDK is broken, it's likely a bug, not a breaking change. If you found a bug, file a ticket, assign it to me, and I'll fix it asap.

I don't have a mac to develop on, but I'm working on it. I can't test iOS device or dist builds, but I did test iOS sim builds of the Kitchen Sink and two tabbed app.

After a massive effort from @m1ga and myself, we reduced the npm dependency vulnerabilities from 57 to 37. We still have a ways to go. Packages like `request` will be replaced with `undici`, `rollup` with `rolldown`, and so on.

We have way too many packages to maintain and too small of a team. I want to consolidate packages like `node-titanium-sdk`, `node-appc`, `liveview`, `alloy`, `androidlib`, and possibly hyperloop into `titaniumlib`. This should not be a breaking change.

cb1kenobi:
I don't think this should go in `13_0_X`, but I understand the urgency. These changes feels like a minor rev, so a 13.1.0 makes sense.

You can branch `13_1_X` from `13_0_X` and re-target this PR to that branch. 13.1.0 will miss out on fixes. `master` can be bumped to 13.2.0.

Note that there is no easy way to convert this PR to ESM. You will need to rewrite your imports and exports. It will be fun.

hansemannn:
It will not, it's just so we can build based on it and another developer who needs it can use it as well.

yozef:
This works for a titanium module wrapped for a library that used SPM packages as dependancies, and it worked nicely :)

hansemannn:
I've refactored the PR to now target `main`. @yozef You may need to run an additional test before moving to 13.1.0 once we release it, so we can be sure that the SPM changes were correctly ported.

cb1kenobi (approved):
This is awesome!

PR #14328, opened 2025-10-27
feat: add script message handler support to Android WebView

prashantsaini1:
This PR implements the `addScriptMessageHandler` and the `removeScriptMessageHandler` methods for Android platform, bringing parity with the iOS.

**Titanium App:**
```
const HANDLER_NAME = 'yourHandlerName';
const webView = Ti.UI.createWebView({ url: 'some-remote-url' });

webView.addScriptMessageHandler(HANDLER_NAME, true);
webView.addEventListener('message', (e) => {
	if (e.name === HANDLER_NAME) {
		alert(e.body.message);
	}
});

window.add(webView);
window.open();

```

**Web page:**
```
<script type="text/javascript">
  const body = { message: 'Hello Titanium!'};

  // For both Android/iOS.
  if (window.tisdk) {
    window.tisdk.emit('yourHandlerName', JSON.stringify(body));
    return;
  }
  
  // iOS only: to keep the backward compatibility with the old syntax.
  if (window.webkit?.messageHandlers?.yourHandlerName) {
    window.webkit.messageHandlers.yourHandlerName.postMessage(body);
  }
</script>
```

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -16,6 +16,7 @@ import android.print.PrintDocumentAdapter;
 import android.print.PrintManager;
 import android.util.DisplayMetrics;
 import android.view.View;
+import android.webkit.JavascriptInterface;
 import android.webkit.ValueCallback;
 import android.webkit.WebView;
 
@@ -34,12 +35,17 @@ import org.appcelerator.titanium.io.TiBaseFile;
 import org.appcelerator.titanium.io.TiFileFactory;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.view.TiUIView;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 import java.io.File;
+import java.lang.ref.WeakReference;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import ti.modules.titanium.ui.widget.webview.TiUIWebView;
 
@@ -76,6 +82,7 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 	private static String fpassword;
 	PrintManager printManager;
 	private Message postCreateMessage;
+	private JSInterface jsInterface;
 
 	@Kroll.constant
 	public static final int PDF_PAGE_DIN_A4 = 0;
@@ -155,6 +162,24 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 		}
 		// TODO deprecate the sync variant?
 		return view.getJSValue(code);
+	}
+
+	@Kroll.method
+	public void addScriptMessageHandler(String name)
+	{
+		if (jsInterface == null) {
+			jsInterface = new JSInterface(this);
+		}
+
+		jsInterface.addScriptMessageHandler(name);
+	}
+
+	@Kroll.method
+	public void removeScriptMessageHandler(String name)
+	{
+		if (jsInterface != null) {
+			jsInterface.removeScriptMessageHandler(name);
+		}
 	}
 
 	@Kroll.getProperty
@@ -634,6 +659,11 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 	@Override
 	public void onDestroy(Activity activity)
 	{
+		if (jsInterface != null) {
+			jsInterface.destroy();
+			jsInterface = null;
+		}
+
 		TiUIWebView webView = (TiUIWebView) peekView();
 		if (webView == null) {
 			return;
@@ -691,6 +721,91 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 					callback.callAsync(krollObject, new Object[] { value });
 				}
 			});
+		}
+	}
+
+	private class JSInterface
+	{
+		private Set<String> scriptMessageHandlers = new HashSet<>();
+		private final HashMap<String, Integer> appListeners = new HashMap<>();
+		private WeakReference<WebViewProxy> proxy;
+
+		public JSInterface(WebViewProxy proxy)
+		{
+			this.proxy = new WeakReference<>(proxy);
+		}
+
+		public void addScriptMessageHandler(String name)
+		{
+			if (view == null) {
+				return;
+			}
+
+			scriptMessageHandlers.add(name);
+			((TiUIWebView) view).getWebView().addJavascriptInterface(this, name);
+		}
+
+		public void removeScriptMessageHandler(String name)
+		{
+			if (view == null) {
+				return;
+			}
+
+			scriptMessageHandlers.remove(name);
+			((TiUIWebView) view).getWebView().removeJavascriptInterface(name);
+		}
+
+		@JavascriptInterface
+		public void postMessage(String json)
+		{
+			if (this.proxy == null || this.proxy.get() == null) {
+				return;
+			}
+
+			KrollDict dict = null;
+			try {
+				if (json != null && !json.equals("undefined")) {
+					dict = new KrollDict(new JSONObject(json));
+				}
+			} catch (JSONException e) {
+				Log.e(TAG, "Error parsing scriptMessageHandler's JSON message", e);
+			}
+
+			if (dict == null) {
+				return;
+			}
+
+			// This is just to keep parity with the iOS structure, no other use.
+			String name = dict.getString("name");
+			if (name == null) {
+				Log.e(TAG, "scriptMessageHandler 'name' missing");
+				return;
+			}
+
+			if (!scriptMessageHandlers.contains(name)) {
+				Log.e(TAG, "scriptMessageHandler missing for name: " + name);
+				return;
+			}
+
+			this.proxy.get().fireEvent("message", dict);
+		}
+
+		// Clear local proxy, all script message handlers and their interfaces.
+		public void destroy()
+		{
+			if (this.proxy != null) {
+				this.proxy.clear();
+				this.proxy = null;
+			}
+
+			if (view != null) {
+				WebView webView = ((TiUIWebView) view).getWebView();
+				for (String name : scriptMessageHandlers) {
+					webView.removeJavascriptInterface(name);
+				}
+			}
+
+			scriptMessageHandlers.clear();
 		}
 	}
 }
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -170,7 +170,7 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 	}
 
 	@Kroll.method
-	public void addScriptMessageHandler(String name, @Kroll.argument(optional = true) boolean unusedParam)
+	public void addScriptMessageHandler(String name)
 	{
 		if (jsInterface == null) {
 			jsInterface = new JSInterface(this);
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -134,6 +134,11 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 			postCreateMessage = null;
 		}
 
+		// Add script message handlers if attempted before the web-view could be created.
+		if (jsInterface != null) {
+			jsInterface.addPendingScriptMessageHandlers(webView);
+		}
+
 		return webView;
 	}
 
@@ -726,6 +731,8 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 
 	private class JSInterface
 	{
+		// Store references to add later the web-view is created.
+		private Set<String> pendingScriptMessageHandlers = new HashSet<>();
 		private Set<String> scriptMessageHandlers = new HashSet<>();
 		private final HashMap<String, Integer> appListeners = new HashMap<>();
 		private WeakReference<WebViewProxy> proxy;
@@ -735,9 +742,22 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 			this.proxy = new WeakReference<>(proxy);
 		}
 
+		// To be called only when the web-view is created and before setting any URL.
+		public void addPendingScriptMessageHandlers(TiUIWebView tiUIWebView)
+		{
+			WebView webView = tiUIWebView.getWebView();
+			for (String name : pendingScriptMessageHandlers) {
+				scriptMessageHandlers.add(name);
+				webView.addJavascriptInterface(this, name);
+			}
+
+			pendingScriptMessageHandlers.clear();
+		}
+
 		public void addScriptMessageHandler(String name)
 		{
 			if (view == null) {
+				pendingScriptMessageHandlers.add(name);
 				return;
 			}
 
@@ -793,6 +813,8 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 		// Clear local proxy, all script message handlers and their interfaces.
 		public void destroy()
 		{
+			pendingScriptMessageHandlers.clear();
+
 			if (this.proxy != null) {
 				this.proxy.clear();
 				this.proxy = null;
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -136,7 +136,7 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 
 		// Add script message handlers if attempted before the web-view could be created.
 		if (jsInterface != null) {
-			jsInterface.addPendingScriptMessageHandlers(webView);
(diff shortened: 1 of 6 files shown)


m1ga:
The example code will show `No app side handler available.` when I run it on a Pixel 9

prashantsaini1:
@m1ga It might have something to do with the synchronisation of the web-page script loading and adding the app side message handler. Can you try to call the web-page side method `postMessage` upon a button click or anything which does not run the script instantly?

I tested both Android and iOS changes upon a button click rather than instantly running the script, but I added the sample code here for simplicity.

m1ga:
I've used the Chrome Devtools to inspect the webview and even using the console I don't have `window.MyScriptMessageHandler`

prashantsaini1:
@m1ga I checked the native docs, and found that the script must be injected either before the URL is loaded or the URL must be reload after. In my tests, the web-page does some redirections. Can you please set the webView URL after the `window.open()` call in the Ti App side and let me know? Thanks!

m1ga:
Still no luck:

```js
const HANDLER_NAME = 'MyScriptMessageHandler';

const window = Ti.UI.createWindow({ title: 'Android Script Message' });
const webView = Ti.UI.createWebView({ });

window.add(webView);

webView.addScriptMessageHandler(HANDLER_NAME);
webView.addEventListener('message', (e) => {
	if (e.name === HANDLER_NAME) {
		alert(e.body?.message ?? 'No message');
	}
});

window.addEventListener("open", function(){
	webView.url = '/page.html'
})

window.open();
```
and 
```html
<h1>test</h1>
<script>
  const androidHandler = window.MyScriptMessageHandler?.postMessage;
  const iOSHandler = window.webkit?.messageHandlers?.MyScriptMessageHandler?.postMessage;


  document.querySelector("h1").addEventListener("click", function() {
    if (androidHandler) {
      androidHandler(
        JSON.stringify({
          name: 'MyScriptMessageHandler',
          body: {
            message: 'Titanium WebView Rocks!'
          },
        }),
      )
    } else if (iOSHandler) {
      iOSHandler({
        message: 'Titanium WebView Rocks!'
      });
    } else {
      // No app side handlers available.
      alert('No app side handler available.')
    }
  })
</script>
```
and I click on the h1 and get the "no app side..." alert. I've fetch the PR again and rebuild the SDK just to make sure it's running the correct version. Tested liveview and no liveview. But I don't see the correct alert

m1ga:
This example code works now:

```js
const HANDLER_NAME = 'MyScriptMessageHandler';

	title: 'Android Script Message'
});
const webView = Ti.UI.createWebView({});
window.add(webView);

window.addEventListener("open", function() {
	webView.addScriptMessageHandler(HANDLER_NAME);
	webView.addEventListener('message', (e) => {
		if (e.name === HANDLER_NAME) {
			alert(e.body?.message ?? 'No message');
		}
	});
	webView.url = '/page.html'
})

window.open();
```
`addScriptMessageHandler` has to be called after `open` otherwise `view` is null in the check and it won't add the listeners.

HTML part:
```html
<!DOCTYPE html>
<html lang="en" dir="ltr">

<head>
  <meta charset="utf-8">
  <title></title>
</head>

<body>
  <h1>test</h1>
  <script>
    document.querySelector("h1").addEventListener("click", function() {

      const androidHandler = window.MyScriptMessageHandler;
      const iOSHandler = window.webkit?.messageHandlers?.MyScriptMessageHandler;

      document.querySelector("h1").addEventListener("click", function() {
        if (androidHandler) {
          androidHandler.postMessage(
            JSON.stringify({
              name: 'MyScriptMessageHandler',
              body: {
                message: 'Titanium WebView Rocks!'
              },
            }),
          )
        } else if (iOSHandler) {
          iOSHandler.postMessage({
            message: 'Titanium WebView Rocks!'
          });
        } else {
          // No app side handlers available.
          alert('No app side handler available.')
        }
      })
    })
  </script>

</body>

</html>
```
(works locally too)

prashantsaini1:
@m1ga Yep! First issue was referencing the `postMessage` in Web Page code directly, brings in a different closure. Fixing the `addScriptMessageHandler` order sequence issue as I noticed it too. Just updated the example as well.

Edit: The problem is calling the `addScriptMessageHandler` without the window being properly opened or when the proxy's `view` is `null`.

I've updated the code in description for now, and will check further if we can handle it inside the SDK itself.

prashantsaini1:
@m1ga Just fixed the code (refer description), tested with http://migaweb.de/page.html?123 and our local web page as well.

prashantsaini1:
@janvennemann PR's ready now to proceed further as discussed.

m1ga:
@prashantsaini1 if you click on the text multiple time it will fire the event multiple times

prashantsaini1:
@m1ga It's because your web page code which you posted is adding the event listener again inside the listener itself. ;)

m1ga:
oh :smile: didn't notice the double click event! Then it's working fine now :+1:

m1ga (approved):

janvennemann:
First off, thanks a lot @prashantsaini1 for investigating this matter and adding the basic feature parity. This is a great basic start. I'd love to add a unified API for this that does the heavy lifting behind the scenes and does not require so much platform specific extra logic. Right now it is very cumbersome to set up all the required handlers.

I'm thinking of a simple event based API that can be used like this inside the web view:

```ts
window.tiWebViewBridge.emit(eventName: string, data: any)
```

And then on the Titanium side handle it like any other event:

```ts
webView.addEventListener('my-custom-event', ({ name, data }) => {
  console.log({ name, data })
})
```

For this we would use automatically injected JS code that prepares a single script message handler which is used to send the event that we unwrap and forward on the native side into the Titanium event bus. I totally understand this may be out of scope for this PR but i just wanted to leave my two cents how we could improve this with a proper cross-platform API.

m1ga:
for that you can look at https://github.com/tidev/titanium-sdk/pull/13835 which will inject the current binding.js file in an external page.

janvennemann:
> for that you can look at #13835 which will inject the current binding.js file in an external page.

Ah nice, but that is also Android only and is again a totally different API than what iOS uses. That's why i mentioned a unified API that works the same on both platforms.

Just to be sure i'm not missing something what's the benefit of this PR compared https://github.com/tidev/titanium-sdk/pull/13835. Doesn't it practically do the same, e.g. sending events from the web view to the Titanium app.

m1ga:
@janvennemann I just wanted to show you how you could inject the bindings. My old PR will just use the normal `Ti` namespace (https://titaniumsdk.com/guide/Titanium_SDK/Titanium_SDK_How-tos/Integrating_Web_Content/Communication_Between_WebViews_and_Titanium.html) that has been there for ages and is added in local pages automatically. My PR will just add it for external pages (and yes, only for Android as I don't know how to do it for iOS :) )

You could use that and add a different namespace too but the Ti should stay as many people use that.

prashantsaini1:
@m1ga @janvennemann Well, I found some interesting stuff on iOS.

If we enable the `_Ti_` message handler on the app side, we can actually invoke the `Ti.App.fireEvent(...)` on the web-page, just have to wait for the script to be initially loaded.

I tested it on our web-page and already works fine without the `window.webkit/window.Android` stuff.

**App side:**
```
Ti.App.addEventListener('myCustomEvent', (e) => {
	Ti.API.error('Event from webview = ' + JSON.stringify(e));
});

// iOS already injects the `Ti.App` namespace script under the `_Ti_` handler.
webView.addScriptMessageHandler('_Ti_');
```

**Web page side:**
```
 Ti.App.fireEvent('myCustomEvent', 'pass any data here, even a json object as usual');
```

**Next Steps:**
- [ ] **iOS**: Auto invoke the `webView.addScriptMessageHandler('_Ti_');` in iOS SDK when the relevant script is injected.
- [ ] **Android**: review the PR from @m1ga  if that works already, or check how it can be done on Android.

janvennemann:
Ha, thanks @prashantsaini1, you beat me to it. I found the same after digging through our web view implementation thanks to @m1ga's hints.

The iOS implementation is also limited to local URLs, by manually invoking `webView.addScriptMessageHandler('_Ti_');` you basically manually doing [this step](https://github.com/tidev/titanium-sdk/blob/b2b4ac8c94d89ce8cceda52dad6461f475f0f5d3/iphone/Classes/TiUIWebView.m#L196)

I think we should just introduce a new property like `autoInjectEventBridge` which controls this behavior. On iOS it set's the script message handler, on Android it loads the bindings file. It can be `false` by default to match existing behavior and possibly for security concerns.

prashantsaini1:
@janvennemann Sounds good. @m1ga I'd take your PR further to enable it for Android as well. :)

m1ga:
sounds good :+1: Not sure if that might be relevant for your page too (in case links that open with target:_blank): https://github.com/tidev/titanium-sdk/pull/14005 but it can be useful when you work with external pages and you don't want to remove `target:_blank` by hand all the time

prashantsaini1:
@m1ga @janvennemann Finally achieved the following syntax for both platforms. To keep backward compatibility without breaking existing iOS apps, added 2nd argument to enable the unified API.

```
const body = { message: 'Hello Titanium!'};
window.tisdk.emit('myCustomHandler', JSON.stringify(body));
```
Screenshot 2025-10-31 at 12 22 57 PM

apidoc/Titanium/UI/WebView.yml
```diff
@@ -419,39 +419,53 @@ methods:
...
+        Important: Call this method before setting the URL, or reload the URL if added later.
+        JS code usages on your web page script:
         ```
         <script type="text/javascript">
-            window.webkit.messageHandlers.yourHandlerName.postMessage({
-                message: 'This is an example string.'
-            });
+            // For iOS with second argument as `false`:
+            const body = { message: 'Hello Titanium!'};
+            window.webkit.messageHandlers.yourHandlerName.postMessage(body);
```
m1ga:
```suggestion
            window.webkit?.messageHandlers.yourHandlerName.postMessage(body);
```

apidoc/Titanium/UI/WebView.yml
```diff
@@ -419,39 +419,53 @@ methods:
...
+            
+            // For Android, or for iOS with second argument as true:
+            const body = { message: 'Hello Titanium!'};
```
m1ga, line 438:
```suggestion
```

android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
```diff
@@ -127,6 +134,11 @@ public TiUIView createView(Activity activity)
 			postCreateMessage = null;
 		}
 
+		// Add script message handlers if attempted before the web-view could be created.
+		if (jsInterface != null) {
+			jsInterface.initializeJSInterface(webView);
+		}
```
m1ga, line 140: maybe add an error log in an `else` case? That way you'll get noticed if you call it too early (e.g. I still have to do it in the `open` event.

m1ga (changes requested):
added some notes. The changes in the Web code will make it work on both platforms if you copy and paste it. Otherwise it will complain on Android because webkit is not there and `const body` is already there.

android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
prashantsaini1: That's the purpose of this code to not worry about the timing of the call on the app side. You can now call it anywhere just like any properties are taken care by the SDK.

m1ga: you are right, sorry about that. Working fine now.

m1ga (approved):
:+1:

iphone/Classes/TiUIWebViewProxy.m
```diff
@@ -384,12 +384,25 @@ - (void)removeAllUserScripts:(id)unused
...
+  BOOL enableParity = NO;
+  if ([args count] > 1) {
+    enableParity = [TiUtils boolValue:[args objectAtIndex:1] def:NO];
+  }
 
   WKUserContentController *controller = [[[self wkWebView] configuration] userContentController];
-  [controller addScriptMessageHandler:[self webView] name:value];
+
+  // Enable web pages to use unified API for both Android and iOS as:
+  // window.tiBridge.emit('handlerName', JSON.strinigify(body));
```
hansemannn:
```suggestion
  // window.tisdk.emit('handlerName', JSON.strinigify(body));
```

iphone/Classes/TiUIWebView.m
```diff
@@ -534,6 +534,21 @@ - (WKUserScript *)userScriptDisableContextMenu
...
+- (WKUserScript *)userScriptForMessageHandlerParity:(NSString *)handlerName
+{
+  NSString *script = @"(function(name){ \
+    window.tisdk={ \
+      emit:function(name, payload){ \
+        window.webkit.messageHandlers[name].postMessage(JSON.parse(payload)); \
+      } \
+     }; \
+    })('%@'); \
+    ";
```
janvennemann: Why is this wrapped in an IIFE? The outer scoped `name` will be masked by the inner `name` parameter of the `emit` function so it is never used.

iphone/Classes/TiUIWebViewProxy.m
```diff
@@ -384,12 +384,25 @@ - (void)removeAllUserScripts:(id)unused
...
+- (void)addScriptMessageHandler:(id)args
 {
-  ENSURE_SINGLE_ARG(value, NSString);
+  NSString *handlerName = nil;
+  ENSURE_ARG_AT_INDEX(handlerName, args, 0, NSString);
+
+  BOOL enableParity = NO;
+  if ([args count] > 1) {
+    enableParity = [TiUtils boolValue:[args objectAtIndex:1] def:NO];
+  }
```
janvennemann: Just wondering why do we set this to `false` by default? If I understand correctly this only adds a utility script to expose the new `window.tisdk` helper. It does not include any breaking changes to calling it the old fashioned way, so why not have this enabled by default? That way you can use the new syntax on both platforms without having to pass different parameters.

Also, i think it would be nice to just set this once as a web view option instead of having to set it in each `addScriptMessageHandler` call (especially if we enable it by default anyway). Something like `autoInjectJSBridge` with the following description: "Enables the `window.tisdk` helper inside the WebView for sending events between the WebView and Titanium." On iOS this toggles the additional user script, and on Android it can toggle adding the entire JS interface that provides `window.tisdk`.

janvennemann (changes requested):

iphone/Classes/TiUIWebView.m
prashantsaini1: Ah, right. I'll modify it.

iphone/Classes/TiUIWebViewProxy.m
prashantsaini1: 1. Setting the 2nd argument as `false` was based on my assumption that it may break some web apps. But I think I was wrong, as the `window.webkit.messageHandlers.*` will always remain active web pages on iOS. I think we can even remove it as it's being supported first time on Android and already provides both ways to access handlers on iOS. It'd be much simpler without this argument then.
2. Using the `addScriptMessageHandler` method allows mobile side code to take different actions based on different handlers by different web-pages. I'd keep it this way itself. :)

iphone/Classes/TiUIWebViewProxy.m
```diff
@@ -389,6 +389,11 @@ - (void)addScriptMessageHandler:(id)value
   ENSURE_SINGLE_ARG(value, NSString);
 
   WKUserContentController *controller = [[[self wkWebView] configuration] userContentController];
+
+  // Enable web pages to use unified API for both Android and iOS as:
+  // window.tisdk.emit('handlerName', JSON.stringify(body));
+  [controller addUserScript:[[self webView] userScriptForMessageHandlerParity]];
```
janvennemann: This will repeatedly add the script every time `addScriptMessageHandler` is called. We should either do this only once when the web view is initialised or guard appropriately if it has to stay here in the `addScriptMessageHandler:`.

prashantsaini1: The script will just do the same assignment `window.tisdk.emit(...)` every time, so this should not affect anything practically.

janvennemann: It does, but it is still unnecessary. This only needs to be done once. This works for now since there is no state inside the window.tisdk object. But if we decide to add the option to send events from Titanium to the WebView later this will likely break. Better do it right immediately than to rely on the behavior of the script we inject.

prashantsaini1: Added. Please check.

iphone/Classes/TiUIWebViewProxy.m
```diff
@@ -14,7 +14,9 @@
 #import "TiNetworkCookieProxy.h"
 #import "TiUIWebView.h"
 
-@implementation TiUIWebViewProxy
+@implementation TiUIWebViewProxy {
+  BOOL _hasAddedTisdkScript;
+}
```
janvennemann: Ok, almost there! Use a private property in a class extension in the .m file, not a raw ivar. This gives you clear encapsulation, dot-syntax, and proper semantics under ARC/KVC. That's basically the modern Objective-C way to do this.

```objc
@interface TiUIWebViewProxy ()
@property (nonatomic, assign) BOOL hasAddedTisdkScript;
@end

@implementation TiUIWebViewProxy
```

janvennemann (changes requested):

iphone/Classes/TiUIWebViewProxy.m
janvennemann: @prashantsaini1 Double check again what i wrote here. The class extension has to go into the .m file. not the header. In the header it's a normal public property. By declaring it in the .m file as a class extension you make sure it's private and only accesible from within the class.

janvennemann (approved):

PR #14330, opened 2025-10-28
fix: show error if build command's config fails

cb1kenobi:
I don't know what I was thinking over 10 years ago when I thought it was a good idea to not bubble up errors from a command's `config()` function. We recently saw a bug where something was blowing up in `config()`, but the CLI continued on and eventually the command failed with an error that was unrelated to the actual problem. This PR surfaces the error and causes the CLI to exit with code 1.


m1ga (approved):
Nice :+1:

PR #14332, opened 2025-10-29
chore(ios): removed old iTunes references in apidoc

hbugdoll:
**Description:**

- removed old references to _iTunes_ in the API documentation (e.g. iTunes Store &rarr; App Store, iTunes &rarr; iCloud)
- only at `CoverFlowView` is it still listed for historical reasons 😄


cb1kenobi (approved):
Awesome!

PR #14333, opened 2025-11-02
feat(ios): add parity for “backgroundSelectedColor” API

hansemannn:
This pull request adds support for selected background color states in `Ti.UI.View` instances, aligning its behavior with Android and expanding platform coverage. Advantages compared to existing approaches:

- Increased performance: No additional tap gesture recognizer / event listeners necessary, no need to cross the JS bridge
- Parity: 1:1 compatibility with existing Android API

Test case:
```js
    backgroundColor: '#fff'
});

    width: 200,
    height: 50,
    backgroundColor: 'green',
    backgroundSelectedColor: 'red'
});

win.add(view);
win.open();
```

https://github.com/user-attachments/assets/5a2fcb1d-d006-4ff1-b720-4edf4632ef89

### DIFF
--- a/apidoc/Titanium/UI/View.yml
+++ b/apidoc/Titanium/UI/View.yml
@@ -1365,7 +1365,8 @@ properties:
 
         Defaults to background color of this view.
     type: [String, Titanium.UI.Color]
-    platforms: [android]
+    since: {android: "0.9", iphone: "13.1.0", ipad: "13.1.0", "macos": "13.1.0"}
+    platforms: [android, iphone, ipad, macos]
 
   - name: backgroundSelectedImage
     summary: Selected background image url for the view, specified as a local file path or URL.
@@ -1483,10 +1484,14 @@ properties:
     since: "9.3.0"
 
   - name: focusable
-    summary: Whether view should be focusable while navigating with the trackball.
+    summary: Enables the view to receive focus and selected background feedback when interacted with.
+    description: |
+        On Android this controls whether the view participates in focus navigation such as with a D-Pad or trackball.
+        On all platform, this property is required to use the `backgroundSelectedColor` property.
     type: Boolean
     default: false
-    platforms: [android]
+    since: {android: "0.9", iphone: "13.1.0", ipad: "13.1.0", "macos": "13.1.0"}
+    platforms: [android, iphone, ipad, macos]
 
   - name: height
     summary: View height, in platform-specific units.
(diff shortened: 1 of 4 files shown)


apidoc/Titanium/UI/View.yml
```diff
@@ -1861,19 +1861,19 @@ properties:
...
-        For Titanium versions older than 9.1.0, touch feedback only works if you set the
-        <Titanium.UI.View.backgroundColor> property to a non-transparent color.
+        On Android this maps to the native ripple effect. On iOS the view background briefly switches to the feedback
+        color for the duration of the touch.
     type: Boolean
     default: false
-    platforms: [android]
-    since: "6.1.0"
+    platforms: [android, iphone, ipad]
+    since: {android: "6.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
```
m1ga:
```suggestion
    since: {android: "6.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
```

apidoc/Titanium/UI/View.yml
```diff
@@ -1861,19 +1861,19 @@ properties:
...
 
   - name: touchFeedbackColor
     summary: Optional touch feedback ripple color. This has no effect unless `touchFeedback` is true.
-    description: Defaults to provided theme color.
+    description: Defaults to the native highlight color on each platform.
     type: String
-    platforms: [android]
-    since: "6.1.0"
+    platforms: [android, iphone, ipad]
+    since: {android: "6.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
```
m1ga:
```suggestion
    since: {android: "6.1.0", iphone: "13.1.0", ipad: "13.1.0", macos: "13.1.0"}
```

m1ga (changes requested):
in case it works for macos you should either have `macos` in the platforms or removing it to make the linter happy

m1ga:
Builds fine locally with 13.1.0 and works as in your video :+1:

prashantsaini1 (commented):
@hansemannn I believe this brings parity with the [backgroundSelectedColor](https://titaniumsdk.com/api/titanium/ui/view.html#properties_backgroundselectedcolor) rather than the `touchFeedback`.

hansemannn:
Good question, what's the difference? Just the ripple?

prashantsaini1:
Yep, touchFeedback is ripple animation, backgroundSelectedColor is flat background switching(100% like in this PR).

apidoc/Titanium/UI/View.yml
```diff
@@ -1360,12 +1360,10 @@ properties:
     summary: Selected background color of the view, as a color name or hex triplet.
     description: |
         For information about color values, see the "Colors" section of <Titanium.UI>.
-
-        `focusable` must be true for normal views.
```
hansemannn, line 1364: This was removed as discussed with Prashant, as Android also does not require it.

hansemannn:
@m1ga @prashantsaini1 Ready when you are

prashantsaini1 (approved):

m1ga (approved):
tested on Android and iOS and it looks the same :+1:

designbymind:
Working great on my end! — TiSDK `13.1.0.v20251103155153` w/ iOS 26.0.1 (simulator & device).

On a similar note and I know this isn't the ideal place to file, but since it's very closely related (and important), I might as well bring it up. — I've noticed that `backgroundSelectedColor` does not use the correct semantic theme color _after the theme has been switched_. **Example:** after switching from light mode to dark mode, the `backgroundSelectedColor` property will be the light mode value when in dark mode (and vice-versa). This is only true for the `Button` element as I've just tested for this new `View` implementation and the theme (`backgroundSelectedColor` value) is respected for each `userInterfaceStyle` (mode). — If any additional information is needed, please ask me, I'd be happy to help test and get this fixed as it's been bugging me for quite some time 🙂

hansemannn:
@designbymind Thank you! Do you have an example code for that? Then we can still address that prior to 13.1.0 as a follow-up effort

designbymind:
@hansemannn I'll put something together for testing and send over ASAP

PR #14334, opened 2025-11-02
fix(ios): fix local Xcode build by selecting compatible Node.js version

hansemannn:
This pull request introduces a new shell script for reliably running Node.js build scripts with enforced version requirements, updates the Xcode build phase to use this new script, and makes minor improvements to the build script argument handling. 

Previously, it has used Node.js 12.x (for whatever reason), so using this local utility script, it will make sure to a) find a proper version and b) display are clear error message if not.

### DIFF
--- a/build/run-node.sh
+++ b/build/run-node.sh
@@ -1,0 +1,105 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ $# -lt 1 ]]; then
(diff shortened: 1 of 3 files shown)


cb1kenobi (approved):
Tested, works! Bizarre it was finding Node 12, but this looks way more robust. Thanks!

PR #14338, opened 2025-11-10
feat(android): enable accessibility support for Ti.UI.Window

prashantsaini1:
This PR comes in as a workaround to this ongoing [Google Android Maps crash outage](https://issuetracker.google.com/issues/457601635) since last week where the apps are crashing while using `liteMode` enabled maps.

Some apps have found the workaround by disabling the accessibility on the map container itself, but it does not work for all since the MapView may create the SurfaceView internally for optimised rendering.

Debugging the stack-trace leads to that some views are still being queried for accessibility before the crash, so we want to try disable the accessibility for the entire activity where the MapView draw its content and see if it works. Unfortunately, there're no other known workarounds as of now, so any other ideas will be appreciated.

Changes in this PR:
- Enable the `accessibilityHidden` property on a `Ti.UI.Window` itself (the current one does not work) which is usually supposed to apply the accessibility value on its `Activity` level.
- This property is required to be applied before inflating any view so it properly takes effect on the entire Activity, which is why I have implemented it using Application level activity callbacks.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiApplicationLifecycle.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiApplicationLifecycle.java
@@ -9,6 +9,7 @@ package org.appcelerator.titanium;
 import android.app.Activity;
 import android.app.Application;
 import android.os.Bundle;
+import android.view.View;
 
 import org.appcelerator.kroll.KrollModule;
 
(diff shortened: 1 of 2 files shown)


hansemannn (approved):

hansemannn:
@prashantsaini1 Please rebase to `main`

prashantsaini1:
@hansemannn Rebasing this branch has multiple conflicts. Should I raise the PR again from a new branch based from `main`?

prashantsaini1:
@hansemannn Rebasing had multiple conflicts to resolve, raised new [PR here](https://github.com/tidev/titanium-sdk/pull/14339) based from `main`. Closing this one.

PR #14343, opened 2025-11-17
chore: allowed multiple arguments for further scons commands

hbugdoll:
**Introduction:**

https://github.com/tidev/titanium-sdk/pull/14312 allowed multiple arguments for the scons command `build`, `cleanbuild` and `test`.
Running e.g. `npm run clean:ios` or directly `node ./build/scons clean ios` still produces:

`error: too many arguments. Expected 0 arguments but got 1.`

**Description:**
- allowed multiple arguments also for`clean`, `install` and `package` command


cb1kenobi (approved):
Great catch, thank you!

PR #14347, opened 2025-11-18
fix(android): fix BottomNavigation selection when using activeTab

m1ga:
Currently when using the `experimental` Android BottomNavigation the selection won't change when you use `activeTab` to select the tab using code. The content is changing but the old button is still highlighted.

https://github.com/user-attachments/assets/fb2cad7b-429a-42b2-956c-565bfa26dbc1

**Test:**
```js
let win1 = Ti.UI.createWindow({backgroundColor:"red"});
let win2 = Ti.UI.createWindow({backgroundColor:"green"});
let tab1 = Ti.UI.createTab({ window: win1, title: 'tab1', tintColor:"red", activeTintColor:"green", icon: '/images/appicon.png'});
let tab2 = Ti.UI.createTab({window: win2, title: 'tab2', tintColor:"red",  activeTintColor:"green", icon: '/images/appicon.png'});

let bottomNav = Ti.UI.createTabGroup({
	tabs: [tab1, tab2],
	theme: "Theme.Titanium.Material3.DayNight",
  experimental: true,
  activeTab: 1
});

bottomNav.open();

setTimeout(function(){
	bottomNav.activeTab = 0
}, 2000)
```

* run the app
* wait 2 seconds
* current issue (13.0.1): tab is changing but selection is still on the first menu item
* with this fix: tab is changing and selection is changing

[fixed](https://github.com/user-attachments/assets/50e7616e-d2ae-426b-b92e-698c680df26a)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
@@ -531,7 +531,9 @@ public class TiUIBottomNavigation extends TiUIAbstractTabGroup implements Bottom
 		currentlySelectedIndex = tabIndex;
 
 		if (bottomNavigation != null && bottomNavigation.getMenu() != null) {
-			bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
+			if (tabIndex < bottomNavigation.getMenu().size()) {
+				bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
+			}
 		}
 
 		TabProxy tp = ((TabProxy) tabsArray.get(tabIndex));
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
@@ -530,7 +530,9 @@ public class TiUIBottomNavigation extends TiUIAbstractTabGroup implements Bottom
 		}
 		currentlySelectedIndex = tabIndex;
 
-		bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
+		if (bottomNavigation != null && bottomNavigation.getMenu() != null) {
+			bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
+		}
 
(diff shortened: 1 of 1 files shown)


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
```diff
@@ -530,6 +530,8 @@ public void selectTab(int tabIndex)
 		}
 		currentlySelectedIndex = tabIndex;
 
+		bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
```
prashantsaini1: Is it fine here without any null checks?

m1ga: I've added a checkout around it. The TabGroupProxy already checks for the tabIndex to be in range (it will show `Invalid tab index` otherwise) and there is another `tabsArray` check a few lines above it so we know there are tabs still available.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/tabgroup/TiUIBottomNavigation.java
```diff
@@ -530,6 +530,10 @@ public void selectTab(int tabIndex)
 		}
 		currentlySelectedIndex = tabIndex;
 
+		if (bottomNavigation != null && bottomNavigation.getMenu() != null) {
+			bottomNavigation.getMenu().getItem(tabIndex).setChecked(true);
```
prashantsaini1: @m1ga Just two more things:

- Apply index check on `getItem(…)` since we are directly accessing tabs using the native view.
- Verify whether the `activeTab` works as intended if set before the TabGroup native view is created. e.g. set `activeTab to 1` in `tss`, or before window opens.

m1ga: done :+1: 

I've changed the example so it will select tab 1 using `activeTab` in the creator and then switch to `activeTab = 0` in code. Both selections show up fine.

Added an index check but can't really test it as the `TabGroupProxy` will already stop it from executing in:
https://github.com/tidev/titanium-sdk/blob/main/android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java#L302-L303
but in case that will be passed the new check will prevent it from trying to access it at the end

prashantsaini1 (approved):
Looks good!

PR #14352, opened 2025-11-28
fix(ios): fixed Swift compilation conditions and flags

hbugdoll:
**Introduction:**

These warnings have caused confusion for several times, particularly as false positive errors ([[1]](https://github.com/tidev/titanium-sdk/issues/13792), [[2]](https://github.com/tidev/titanium-sdk/issues/13792), [[3]](https://github.com/tidev/titanium-sdk/discussions/13701), [[4]](https://github.com/tidev/titanium-sdk/discussions/13847), [[5]](https://tidev.slack.com/archives/C03CVQX2A/p1759151320686219)):

```
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DEPLOYTYPE=development') ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than '__LOG__ID__=******')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DEBUG=1')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'LAUNCHSCREEN_STORYBOARD=1')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DEFAULT_BGCOLOR_RED=1')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DEFAULT_BGCOLOR_GREEN=1')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DEFAULT_BGCOLOR_BLUE=1')  ...
warning: Conditional compilation flags do not have values in Swift; they are either present or absent (rather than 'DISABLE_TI_LOG_SERVER=1')  ...
```

The values of the Swift compiler variable `SWIFT_ACTIVE_COMPILATION_CONDITIONS` are copied from `GCC_PREPROCESSOR_DEFINITIONS` in `iphone/iphone/Titanium.xcodeproj/project.pbxproj`.
But `SWIFT_ACTIVE_COMPILATION_CONDITIONS` doesn't allow key-value pairs,
only a list of conditional compilation flags like in `TI_SYMBOL_MACROS` &ndash; that is why the above warnings are raised.

> [!NOTE]
> My initial naive solution did not work because the Swift compiler does not support key-value compiler flags at all!
>
> But do we even need key-value compiler flags?
> Of course, it's a simple "traditional" way &ndash; easy to maintain.
> But I don't think so in this case. So far, they have only been used in `main.m`.
> And in my opinion, compiler optimizations will make the minimal performance advantage for these few values obsolete.
>
> 1. The `DEPLOYTYPE` flag was no longer used at all; instead, `TI_APPLICATION_DEPLOYTYPE` is available.
> 2. The `__LOG__ID__` flag is ~~not needed because it has exactly the same value as `TI_APPLICATION_GUID`~~ replaced by a conditional flag `LOGTOFILE` and a new constant `TI_LOG_ID` which is filled via templated value `__LOG_ID__`; ~~instead, introduction of a conditional flag `LOGTOFILE` is sufficient.~~
> 3. Only the following flags remain: `TI_LOG_SERVER_PORT` and `DEFAULT_BGCOLOR_RED`/`..._GREEN`/`..._BLUE`. These four can easily be replaced as templated values in `writeMain()` too, analogous to App-ID/GUID/etc.

**Description:**

- ~~splitted~~ transferred non-key-value flags of `GCC_PREPROCESSOR_DEFINITIONS` for use with Swift ~~into~~ to
  - `SWIFT_ACTIVE_COMPILATION_CONDITIONS` (conditional compilation flags)
  - ~~and`OTHER_SWIFT_FLAGS` (key-value pairs)~~
- therefore, added new array `swiftConds` to clearly distinguish from `gccDefs` (this also prevents similar programming mistakes in the future) and a new config variable `SWIFT_CONDITIONS` in `project.xcconfig`
- replaced all current key-value compiler flags (see **Note** above)
- additionally made some "chore":
  - changed `TI_APPLICATION_DEPLOYTYPE` into `TI_APPLICATION_DEPLOY_TYPE` to consistently take camel case into account (comp. with `deployType`)
  - changed templated value `__APP_DEPLOY_TYPE__` (=`buildType`) into `__BUILD_TYPE__` to avoid confusion with `__DEPLOY_TYPE__` (=`deployType`)


hbugdoll:
Changes from initial draft:
- ~~replaced undocumented `SWIFT_COMPILER_FLAGS` with `OTHER_SWIFT_FLAGS`[^1]~~
- use of JS arrow functions

[^1]: https://developer.apple.com/documentation/xcode/build-settings-reference#Other-Swift-Flags

hbugdoll:
New changes:
- rework: replaced key-value compiler flags overall

iphone/Classes/AppModule.m
```diff
@@ -18,7 +18,7 @@
 #import <UIKit/UILocalNotification.h>
 #import <unistd.h>
 
-extern NSString *const TI_APPLICATION_DEPLOYTYPE;
```
hansemannn, line 21: Did you check if these are still properly set? As they are generated by the build

support/iphone/main.m
```diff
@@ -50,23 +47,20 @@ int main(int argc, char *argv[])
...
 #if defined(DEBUG) || defined(DEVELOPER)
   [[TiSharedConfig defaultConfig] setDebugEnabled:YES];
 #endif
 
-#ifdef __LOG__ID__
+#ifdef LOGTOFILE
   NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
   NSString *documentsDirectory = [paths objectAtIndex:0];
-  NSString *logPath = [documentsDirectory stringByAppendingPathComponent:[NSString stringWithFormat:@"%s.log", STRING(__LOG__ID__)]];
+  NSString *logPath = [documentsDirectory stringByAppendingPathComponent:[NSString stringWithFormat:@"%@.log", TI_APPLICATION_GUID]];
```
hansemannn: Was there a case where this caused issues?

hansemannn (commented):
Minor questions!

support/iphone/main.m
hbugdoll: No, but as I wrote (see *Note* above), the intention here is to replace all key-value compiler flags.
The only use of `__LOG__ID__` is in `support/iPhone/main.m` and
> The `__LOG__ID__` flag is not needed because it has exactly the same value as `TI_APPLICATION_GUID`

iphone/Classes/AppModule.m
hbugdoll: It is a SDK-wide renaming of this constant.
But I will check it again!

support/iphone/main.m
hansemannn: The intention of the flag (initially used for the logger from @cb1kenobi) IIRC was to be able to attach different loggers if wanted

iphone/Classes/AppModule.m
hbugdoll: Tested successfully `Ti.App.deployType` in `development` (`-T simulator`) and `production` (`-T dist-appstore`).

hansemannn (approved):

support/iphone/main.m
hbugdoll: > The intention of the flag (initially used for the logger from @cb1kenobi) IIRC was to be able to attach different loggers if wanted

Thanks for the clarification.

So I added a separate constant `TI_LOG_ID`. The build process filled then the placeholder `__LOG_ID__` with `tiapp.guid` like before &ndash; here one can use another ID instead of the app's GUID.

hbugdoll:
@cb1kenobi For Swift compatibility the key-value compiler flag `__LOG__ID__` was replaced by templated value `__LOG_ID__`. To use another logger you can now deal with it at: https://github.com/tidev/titanium-sdk/blob/23c26d313daff361763395b397e427190f649683/iphone/cli/commands/_build.js#L4837

support/iphone/main.m
cb1kenobi: If I had to guess, the only reason I used `__LOG__ID__` was because I didn't know `TI_APPLICATION_GUID` existed. :)

PR #14362, opened 2026-01-05
chore(android): update kotlin to 2.3.0

m1ga:
Update Ti Kotlin version to 2.3.0

Tested with my MLKit module (kotlin based) and it will fix a titanium-stripe (https://github.com/hansemannn/titanium-stripe) build error with the latest version. That is using a higher kotlin version in their strip-android version ("The actual metadata version is 2.1.0, but the compiler version 1.9.0 can read versions up to 2.0.0.").

### DIFF
--- a/android/templates/build/root.build.gradle
+++ b/android/templates/build/root.build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlin_version = '1.9.23'
+	ext.kotlin_version = '2.0.0'
 
 	repositories {
 		google()
--- a/android/templates/build/root.build.gradle
+++ b/android/templates/build/root.build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlin_version = '2.0.0'
+	ext.kotlin_version = '2.3.0'
 
 	repositories {
 		google()
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -6,7 +6,7 @@
  */
 
 buildscript {
-	ext.kotlin_version = '1.9.23'
+	ext.kotlin_version = '2.0.0'
 
 	repositories {
 		google()
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -6,7 +6,7 @@
  */
 
 buildscript {
-	ext.kotlin_version = '2.0.0'
+	ext.kotlin_version = '2.3.0'
 
 	repositories {
 		google()


hansemannn:
Any breaking changes related to this change? I'm not sure if we can take this pre-14.0.0

m1ga:
I've tested an existing app (created with 13.0.1.GA and with some existing modules), kitchensink and building a none-kotlin module with it. Didn't see any issue or difference.
SDK itself is not using Kotlin so it should only affect custom modules and there I've tested two that build fine.

sorry: pressed the wrong button :smile:

hansemannn:
Yeah, we have a loooot of Kotlin-based modules, so I'm a bit curious if that will cause issues on runtime that we may not see now. @prashantsaini1 WDYT?

m1ga:
My ti.mlkit module still works fine (building and running).
ti.giphy builds and runs fine with this SDK too. ~I don't have an API key so it won't show me giphs but it opens the selector.~ Update: found my API key, works fine.

The starting point for this change was that the ti.stripe module is not 16kb ready and the stripe-android version that supports it can't be compiled with the current Ti SDK. It's currently being tested on Slack.

prashantsaini1:
Acknowledged. I'll do the required research and update soon.

android/templates/build/root.build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-	ext.kotlin_version = '1.9.23'
+	ext.kotlin_version = '2.0.0'
```
hansemannn: Any reason why not going to 2.3.0 directly?

m1ga: just lower risk of breaking changes and we've only found the stripe module so far that bumped the kotlin version https://github.com/stripe/stripe-android/blob/master/CHANGELOG.md#2160---2025-03-03 (2.1.10).

The current kotlin 1.9.x supports building up to kotlin 2.0: "...the compiler version 1.9.0 can read versions up to 2.0.0" and the new 2.0.0 supports building 2.1.x. (that's why stripe compiles).

hansemannn: Okay, then we'll do it iteratively. I've just tested our app with 2.3.0 and even that works. Prashant will leave his analysis here as well.

prashantsaini1:
Here's my current analysis. The Gradle dependency graph shows that the:
- Kotlin `org.jetbrains.kotlin:kotlin-stdlib` is already being resolved to `2.0.20` at runtime. `kotlin-stdlib-common` is also updated to `2.0.20`.
- We have multiple libraries explicitly depending on older Kotlin versions, but Gradle is resolving them upwards like from **ti.googlesign** `1.8.x to 2.0.20`,  **ti.polyfill** `1.3.x to 2.0.20` etc.

It is already safe to directly go to `2.3.0`, but I'd like to compile few modules as they may break at compile time only, not the app itself until that feature is actually used.

Edit: The Gradle and AGP are already in supported sync with the Kotlin 2.3.0.

m1ga:
I've also tested it with 2.3.0 in case you want me to use that version. Building modules + using them (mlkit, giphy) and building an app with old kotlin modules still works too.

Or we can use 2.0.0 for this PR and once it is merged create a new one for 2.3.0 for a later release so we don't have such a big version jump

hansemannn:
As 2.0.0 -> 2.3.0 is also a non-major version bump, I'd actually include it here. We can do a 13.1.0 tomorrow then?

hansemannn (approved):

prashantsaini1:
I've one minor feature update for 13.1.0 if this can wait until this weekend? The Android Light/Dark theme support to change at run-time [as here](https://github.com/tidev/titanium-sdk/pull/14113#issuecomment-2572899780). Just saw it, my last update was exactly Jan 6, 2025 … 1 year passed already 😄

PR #14369, opened 2026-01-10
fix(ios): added dSYM for Swift framework-based modules

### LINKED ISSUE #17: iOS: missing dSYM for TiReviewdialog.framework
- Since the v4 refactor the module is an xcframework.
- If you select *"Upload your app's symbols to receive symbolicated reports from Apple"* during distribution process in Xcode Organizer you currently get a warning `The archive did not include a dSYM for the TiReviewdialog.framework` and therefore unsymbolicated crash reports.
  - similar to https://github.com/tidev/titanium-sdk/issues/14157
Image

@hansemannn Do you use a build script like https://github.com/tidev/APSHTTPClient/blob/master/build.sh or build it directly in Xcode?

hbugdoll:
Fixes https://github.com/hansemannn/titanium-review-dialog/issues/17.

**Introduction:**

- If you select _"Upload your app's symbols to receive symbolicated reports from Apple"_ during distribution process in Xcode Organizer you currently get a warning `The archive did not include a dSYM for the ******.framework` and therefore unsymbolicated crash reports, if you use a Titanium module which is based on a Swift xcframework.
  - e.g. https://github.com/hansemannn/titanium-review-dialog since iOS-v4
- You can test it with _Validate App_ in Xcode Organizer.

**Description:**

- added dSYM debug symbols via `-debug-symbols` in the command `xcodebuild -create-xcframework`
  - added helper function `findDSym()` for this purpose
- BTW improved some logs regarding creating the xcframework

### DIFF
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -757,16 +757,16 @@ FRAMEWORK_SEARCH_PATHS = $(inherited) "$(TITANIUM_SDK)/iphone/Frameworks/**"`);
 
 		const moduleId = this.isFramework ? this.moduleIdAsIdentifier : this.moduleId;
 		const findLib = function (dest) {
-			let libPath = this.isFramework ? '.xcarchive/Products/Library/Frameworks/' + moduleId + '.framework' : '.xcarchive/Products/usr/local/lib/lib' + this.moduleId + '.a';
+			let libPath = this.isFramework ? `.xcarchive/Products/Library/Frameworks/${moduleId}.framework` : `.xcarchive/Products/usr/local/lib/lib${moduleId}.a`;
 			const lib = path.join(this.projectDir, 'build', dest + libPath);
 			this.logger.info(`Looking for ${lib}`);
 
 			if (!fs.existsSync(lib)) {
 				// unfortunately the initial module project template incorrectly
 				// used the camel-cased module id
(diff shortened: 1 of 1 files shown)


hbugdoll:
Oh, there is one **requirement**:

You need the build setting `DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym"` in the *Release* configuration of the module's `project.pbxproj`.

**Edit:**
The module template `iphone/templates/module/objc/template/ios/{{ModuleName}}.xcodeproj/project.pbxproj.ejs` already meets the requirement.

Successfully tested with:
```
ti create -p ios --type module --ios-code-base swift
ti build -p ios --build-only
```

hansemannn (approved):
Good work! I've tinied up some parts of the code, but it should still work properly. Can you confirm once please? We can merge afterwards and include it in next week's 13.1.0 release.

hbugdoll:
> Good work! I've tinied up some parts of the code, but it should still work properly. Can you confirm once please?

Confirmed ✅ 

Since `findDSym()` was an adaptation of `findLib()`, I also made your adjustments in `findLib()` to make the code consistent again.
And applied the style `args.push(<option>, <argument>)` to the other options of the `xcodebuild -create-xcframework` command.

> We can merge afterwards and include it in next week's 13.1.0 release.

That would be great.
In my opinion, it is important that the three dSYM fixes are released together.

PR #14370, opened 2026-01-10
fix(ios): Resolve Mac Catalyst build failures and App Store distribution issues

macCesar:
## Summary

This PR fixes critical build failures that prevented Mac Catalyst apps from compiling and being distributed through the App Store. These issues have affected developers since at least SDK 12.x, requiring workarounds that disrupted normal development workflows.

### 1. Framework Search Path Issue
**Symptom:** Intermittent build failures with framework linking errors  
**Root cause:** Relative framework search path `"Frameworks"` failed to resolve correctly in certain build scenarios  
**Solution:** Use absolute path `"$(PROJECT_DIR)/Frameworks"` for reliable framework location

### 2. Missing TitaniumKit.framework Symlinks
**Symptom:** Compilation errors and App Store rejections  
**Build-time error:**
```
<module-includes>:1:9: error: could not build module 'TitaniumKit'
@import TitaniumKit;
        ^
<unknown>:0: error: Unable to find module dependency: 'TitaniumKit'
```

**App Store rejection:**
```
ERROR ITMS-90206: "Invalid Bundle. The bundle at 'YourApp.app/Contents/Frameworks/
TitaniumKit.framework' contains disallowed file 'Versions'."

Asset validation failed: The bundle 'YourApp.app/Contents/Frameworks/
TitaniumKit.framework' does not exist.
```

**Root cause:** Mac Catalyst requires macOS framework structure with proper symlink hierarchy (`Versions/Current -> A` and root-level symlinks), but builds were producing iOS framework structure  
**Solution:** Automatically create required symlinks at two critical points:
- Before xcodebuild (fixes compilation errors)
- After xcodebuild in final app bundle (fixes App Store validation)

### 3. Incremental Build Issues
**Symptom:** `EEXIST: file already exists` errors when rebuilding without `ti clean`  
**Root cause:** Previous SDK versions created directories instead of symlinks, or symlinks became broken during incremental builds  
**Solution:** Detect and remove directories, files, and broken symlinks before creating new ones

### Modified Files
- `iphone/cli/commands/_build.js`

### Specific Changes

1. **Line 3524:** Updated `FRAMEWORK_SEARCH_PATHS` from `"Frameworks"` to `"$(PROJECT_DIR)/Frameworks"`

2. **Lines 7116-7242:** Added `createTitaniumKitSymlinks()` function
   - Guards execution to Mac Catalyst targets only (`macos`, `dist-macappstore`)
   - Creates proper macOS framework structure with symlinks
   - Handles both build directory and final app bundle locations
   - **Intelligently handles incremental builds:**
     - Detects directories created by previous SDK versions
     - Detects broken symlinks (symlinks pointing to non-existent targets)
     - Verifies existing symlinks point to correct targets
     - Removes incorrect/broken items before creating new ones
   - Uses `lstatSync()` instead of `existsSync()` to detect broken symlinks
   - Provides informative logging for debugging

3. **Lines 7101, 7104, 7433:** Integrated symlink creation into build pipeline
   - Called before xcodebuild to fix compilation
   - Called after xcodebuild to fix App Store packaging

### What This Fixes
- ✅ Mac Catalyst apps now compile successfully
- ✅ Incremental builds work without `ti clean` required
- ✅ Broken symlinks from previous builds are automatically fixed
- ✅ Directories created by SDK 13.0.1.GA are automatically converted to symlinks
- ✅ App Store validation passes for Mac Catalyst apps
- ✅ Enables distribution through Mac App Store
- ✅ Applies to all Titanium apps targeting Mac Catalyst

### What This Doesn't Affect
- Standard iOS device/simulator builds continue to work identically
- Android builds are completely unaffected
- Changes only execute for Mac Catalyst targets due to explicit guards

## Testing

Verified across multiple scenarios with different Titanium projects:
- ✅ Fresh project builds successfully
- ✅ Incremental builds work without `ti clean`
- ✅ Development builds (Debug-maccatalyst) compile and run
- ✅ Distribution builds (Release-maccatalyst) create valid xcarchive
- ✅ App Store validation accepts framework structure
- ✅ Broken symlinks from previous builds are automatically repaired
- ✅ Standard iOS builds unaffected
- ✅ Existing iOS projects continue working normally

## Additional Context

The issues fixed by this PR affected any Titanium application attempting to:
- Compile for Mac Catalyst (Target: `macos`)
- Create distribution builds for Mac App Store (Target: `dist-macappstore`)

These were fundamental build system issues, not app-specific problems.

## ⚠️ Module Compatibility

Mac Catalyst builds require native modules to include an `ios-arm64_x86_64-maccatalyst` slice in their xcframework.

**What this means:**
- If a module doesn't support Mac Catalyst, builds targeting Mac will fail
- iOS device and simulator builds are completely unaffected
- This is expected behavior - not all modules can or should support Mac

**Example error when a module lacks Mac Catalyst support:**
```
error: While building for Mac Catalyst, no library for this platform was found in
'module.xcframework'
```

**For module developers:**
To enable Mac Catalyst support in your module:
1. Add `mac: true` to your module's `manifest` file
2. Rebuild the module - the SDK will automatically build all three slices

**Note:** Some modules may have legitimate reasons to not support Mac Catalyst (iOS-only APIs, hardware dependencies, etc.).

## Backward Compatibility

This change is fully backward compatible:
- No breaking changes to existing APIs
- No changes to tiapp.xml configuration
- No impact on existing iOS/Android workflows
- Mac Catalyst builds that previously required workarounds now work natively
- Automatically migrates projects from SDK 13.0.1.GA directory structure to proper symlinks

---

**Note:** This fix enables Mac Catalyst as a viable deployment target for all Titanium developers, opening the Mac App Store distribution channel that was previously blocked by these issues.

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -3521,7 +3521,7 @@ class iOSBuilder extends Builder {
 				DEAD_CODE_STRIPPING: 'YES',
 				SDKROOT: 'iphoneos',
 				CODE_SIGN_ENTITLEMENTS: '"' + appName + '.entitlements"',
-				FRAMEWORK_SEARCH_PATHS: [ '"$(inherited)"', '"Frameworks"' ]
+				FRAMEWORK_SEARCH_PATHS: [ '"$(inherited)"', '"$(PROJECT_DIR)/Frameworks"' ]
 			},
 			legacySwift = version.lt(this.xcodeEnv.version, '8.0.0');
 
@@ -7098,10 +7098,10 @@ class iOSBuilder extends Builder {
 						next();
 					});
 				}.bind(this), function () {
-					done();
-				});
+					this.createTitaniumKitSymlinks(); done();
+				}.bind(this));
 			} else {
-				done();
+				this.createTitaniumKitSymlinks(); done();
 			}
 		});
 
@@ -7109,6 +7109,101 @@ class iOSBuilder extends Builder {
 			this.logger.debug('Removing empty directories');
 			appc.subprocess.run('find', [ '.', '-type', 'd', '-empty', '-delete' ], { cwd: this.xcodeAppDir }, next);
 		}.bind(this));
+	}
+
+	createTitaniumKitSymlinks() {
+		// Fix for Mac Catalyst: Create required symlinks in TitaniumKit.framework
+		// Mac Catalyst requires macOS framework structure with symlinks
+		// Bug #3: Build-time symlinks in build/Frameworks directory
+		// Bug #4: Distribution symlinks in final app bundle
+
+		if (this.target !== 'macos' && this.target !== 'dist-macappstore') {
+			// Only needed for Mac Catalyst builds
+			return;
+		}
+
+		const frameworkLocations = [];
+
+		// Location 1: Build directory (Bug #3)
+		const buildFrameworkPath = path.join(
+			this.buildDir,
+			'Frameworks',
+			'TitaniumKit.xcframework',
+			'ios-arm64_x86_64-maccatalyst',
+			'TitaniumKit.framework'
+		);
+		if (fs.existsSync(buildFrameworkPath)) {
+			frameworkLocations.push({
+				path: buildFrameworkPath,
+				description: 'build directory'
+			});
+		}
+
+		// Location 2: Final app bundle (Bug #4)
+		// For macos target, the app is in: build/iphone/build/Products/Debug-maccatalyst/AppName.app
+		// For dist-macappstore: build/iphone/build/Products/Release-maccatalyst/AppName.app
+		const productConfig = this.target === 'dist-macappstore' ? 'Release-maccatalyst' : 'Debug-maccatalyst';
+		const appBundleFrameworkPath = path.join(
+			this.buildDir,
+			'build',
+			'Products',
+			productConfig,
+			this.tiapp.name + '.app',
+			'Contents',
+			'Frameworks',
+			'TitaniumKit.framework'
+		);
+		if (fs.existsSync(appBundleFrameworkPath)) {
+			frameworkLocations.push({
+				path: appBundleFrameworkPath,
+				description: 'app bundle'
+			});
+		}
+
+		// Create symlinks in all found locations
+		frameworkLocations.forEach(location => {
+			const frameworkPath = location.path;
+			const versionsPath = path.join(frameworkPath, 'Versions');
+
+			if (!fs.existsSync(versionsPath)) {
+				this.logger.warn(`Versions directory not found in ${location.description}: ${versionsPath}`);
+				return;
+			}
+
+			try {
+				// Create Versions/Current -> A symlink
+				const currentLink = path.join(versionsPath, 'Current');
+				if (!fs.existsSync(currentLink)) {
+					fs.symlinkSync('A', currentLink);
+					this.logger.debug(`Created symlink: Versions/Current -> A in ${location.description}`);
+				}
+
+				// Create root-level symlinks
+				const symlinks = [
+					{ link: 'TitaniumKit', target: 'Versions/Current/TitaniumKit' },
+					{ link: 'Resources', target: 'Versions/Current/Resources' },
+					{ link: 'Headers', target: 'Versions/Current/Headers' },
+					{ link: 'Modules', target: 'Versions/Current/Modules' }
+				];
+
+				let createdCount = 0;
+				symlinks.forEach(item => {
+					const linkPath = path.join(frameworkPath, item.link);
+					if (!fs.existsSync(linkPath)) {
+						fs.symlinkSync(item.target, linkPath);
+						createdCount++;
+					}
+				});
+
+				if (createdCount > 0) {
+					this.logger.info(`✓ Created ${createdCount} Mac Catalyst framework symlinks in ${location.description}`);
+				} else {
+					this.logger.debug(`Mac Catalyst symlinks already exist in ${location.description}`);
+				}
+			} catch (err) {
+				this.logger.warn(`Failed to create symlinks in ${location.description}: ${err.message}`);
+			}
+		});
 	}
 
 	optimizeFiles(next) {
@@ -7334,6 +7429,9 @@ class iOSBuilder extends Builder {
 				}
 
 				// end of the line
+			// Fix Bug #4: Create symlinks in final app bundle after xcodebuild
+			this.createTitaniumKitSymlinks();
+
 				done(code);
 			}.bind(this));
 		});
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -7164,16 +7164,24 @@ class iOSBuilder extends Builder {
 
 		// Helper function to ensure a symlink exists with the correct target
 		const ensureSymlink = (linkPath, target, logger) => {
-			if (fs.existsSync(linkPath)) {
-				// Something exists at this path - check what it is
-				const stats = fs.lstatSync(linkPath); // lstat doesn't follow symlinks
+			// Use lstatSync to check if something exists at this path without following symlinks
+			// This catches broken symlinks that existsSync would miss
+			try {
+				const stats = fs.lstatSync(linkPath);
 
 				if (stats.isSymbolicLink()) {
 					// It's a symlink - check if it points to the correct target
 					const currentTarget = fs.readlinkSync(linkPath);
 					if (currentTarget === target) {
-						// Symlink is correct, nothing to do
-						return false;
+						// Symlink points to correct target - verify it's not broken
+						if (fs.existsSync(linkPath)) {
+							// Symlink is valid and points to correct target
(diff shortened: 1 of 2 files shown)


iphone/cli/commands/_build.js
```diff
@@ -7111,6 +7113,101 @@ class iOSBuilder extends Builder {
...
+			if (!fs.existsSync(versionsPath)) {
+				this.logger.warn(`Versions directory not found in ${location.description}: ${versionsPath}`);
+				return;
+			}
+
+			try {
+				// Create Versions/Current -> A symlink
+				const currentLink = path.join(versionsPath, 'Current');
+				if (!fs.existsSync(currentLink)) {
+					fs.symlinkSync('A', currentLink);
```
cb1kenobi: I think I know what this is trying to do. I'm concerned that the first arg is a relative path, it will resolve against the current directory, not `versionsPath`, right?

macCesar: @cb1kenobi Thanks for the review!

You're right to question this - the first arg `'A'` is indeed a relative path, but that's intentional and correct.

With symlinks, relative paths are resolved **from the symlink's location**, not from the current working directory. So when we do:

```javascript
const currentLink = path.join(versionsPath, 'Current');
// currentLink = .../TitaniumKit.framework/Versions/Current
fs.symlinkSync('A', currentLink);
```

The symlink `Current` is created in `.../Versions/` and points relatively to `A`, which resolves to `.../Versions/A`. This is exactly the standard macOS framework structure.

We can verify this works correctly:
```bash
$ ls -la TitaniumKit.framework/Versions/
lrwxr-xr-x  1 cesar  staff    1 Jan 10 18:31 Current -> A
drwxr-xr-x  7 cesar  staff  224 Jan 10 18:31 A/
```

Using relative symlinks (instead of absolute paths) is the standard practice for macOS frameworks - it makes them relocatable and portable across different installation paths.

This has been tested successfully across multiple build scenarios:
- Fresh builds
- Incremental builds without \`ti clean\`
- Distribution builds for App Store
- Projects migrating from SDK 13.0.1.GA

The symlinks are created correctly in both locations (build directory and final app bundle) and pass App Store validation.

Let me know if you'd like me to add a comment in the code explaining why we use relative paths!

iphone/cli/commands/_build.js
```diff
@@ -7111,6 +7113,147 @@ class iOSBuilder extends Builder {
 		}.bind(this));
 	}
 
+	createTitaniumKitSymlinks() {
+		// Fix for Mac Catalyst: Create required symlinks in TitaniumKit.framework
+		// Mac Catalyst requires macOS framework structure with symlinks
+		// Bug #3: Build-time symlinks in build/Frameworks directory
+		// Bug #4: Distribution symlinks in final app bundle
```
hansemannn: Remove these comments, it's looks odd as part of an overall build script. If necessary, add it to the comments

iphone/cli/commands/_build.js
```diff
@@ -7111,6 +7113,147 @@ class iOSBuilder extends Builder {
...
+				// lstatSync throws if path doesn't exist - this is fine, we'll create it
+				if (err.code !== 'ENOENT') {
+					throw err;
+				}
+			}
+
+			// Create the symlink
+			fs.symlinkSync(target, linkPath);
+			return true;
+		};
```
hansemannn: Why do we need symlinks in the first place I don't see why this would be necessary on macOS deployments specifically. Also, your check includes many redundant checks, is this possibly AI-generated / AI-assisted?

hansemannn (changes requested):
Some of these changes look quite verbose, so I'm wondering if only parts of these changes are actually relevant for this precise fix.

iphone/cli/commands/_build.js
hbugdoll: > Also, your check includes many redundant checks, is this possibly AI-generated / AI-assisted?

&rarr; _Co-Authored-By: Claude Sonnet 4.5_

iphone/cli/commands/_build.js
cb1kenobi: Ah, ok, that makes sense, thanks!

iphone/cli/commands/_build.js
macCesar: **This PR is 100% AI-generated/AI-assisted**.

**I’m not a native iOS developer**; Claude Code initially created a post-compilation script in my app so it could compile.

I was so excited when I saw my app successfully compile with **SDK 13.0.1.GA** that I asked Claude Code to apply the same adjustments directly to the SDK.

In the end, the app compiled with the updated SDK, without the need for the post-compilation script. I also created a new app, and it compiled for macOS without any issues.

> Why do we need symlinks in the first place? I don't see why this would be necessary on macOS deployments specifically.

When I submitted my app for review (almost 3 years ago), Apple immediately rejected it because `TitaniumKit` was NOT included.

Claude Code created the symlinks to ensure `TitaniumKit` is included when compiling for distribution. It even performed all the distribution build tests and ensured that `TitaniumKit` was included in the compiled file.

If you want, I can try to optimize the redundant checks!

hansemannn: Please don't get me wrong: I **LOVE** this, for the simple reason that it enables you to fix very specific issues that weren't possible that way before. And I don't think this PR should be closed or discarded. What I'd want to see is the following:
- Your PR currently addresses the issues using two different approaches: adjusting framework search path and creating symlinks
- Check exactly if both changes are needed or certain changes are rather unrelated. For example `'"Frameworks"' ` -> `'"$(PROJECT_DIR)/Frameworks"' ` should usually not be necessary, as it has the relative path to the project directory by default I think. But maybe it's better!
- Remove the very specific "`Bug #4`" etc logs, as developers checking the code later may be wondering which bug you mean, as it's in a big file that contains many other changes as well
- Most importantly: Check why symlinks are necessary and make the "ensureSymlink" function a bit more lightweight.  For example, `isSymbolicLink` should already return `false` if the path does not exist, so the inner `fs.existsSync(linkPath)` call should not be necessary

After that, I can review (and test) it again and we should be good to go soon :) Thank you for this!

macCesar: Thanks a lot for the feedback and for taking the time to review the PR.. Really appreciate it.

I’ll be out for a few days this week, but once I’m back I’ll go through your suggestions, test them, and clean things up accordingly.

Thanks again for the review and suggestions!

macCesar:
@hansemannn Thanks for the detailed feedback! I've addressed your concerns in the latest commit:

**Changes made:**

1. **Removed "Bug #X" comments** - Replaced with descriptive explanations that make sense in context

2. **Simplified `ensureSymlink` function** - Reduced from 44 to 23 lines (~50% less code):
   - Removed redundant `fs.existsSync` check as you suggested
   - Consolidated the symlink validation logic
   - Removed verbose debug logging
   - Removed unused `logger` parameter

3. **Kept `FRAMEWORK_SEARCH_PATHS`** - The `$(PROJECT_DIR)/Frameworks` path is necessary for Mac Catalyst. Without it, the linker fails with "Unable to find module dependency: TitaniumKit"

**Testing performed:**
- ✅ Clean build (`ti build -p ios -T macos`)
- ✅ Incremental build (detects existing symlinks correctly)
- ✅ Distribution build (`ti build -p ios -T dist-macappstore`)
- ✅ Framework structure verified with correct symlinks

Ready for your review when you have a chance!

hansemannn (approved):
Excellent work, we can approve it now! :)

hansemannn:
@macCesar After merging all changes for 13.1.1 (incl. yours) to 13_1_X, the build fails due to a Mac Catalyst error:

https://github.com/tidev/titanium-sdk/actions/runs/21472015632/job/61846652200

Can you please review the issue? Note: The build was triggered by the tab-bar fixes, but only because that was the last cherry-picked commit before pushing it to the branch where the CI starts.

hbugdoll:
@hansemannn I don't think it was a Mac Catalyst error.
After macOS and Simulator xcarchives were created, the build process aborted at the beginning of the iOS step with:
`error:iOS 26.0 is not installed.`

Therefore `error: Unable to find a destination matching the provided destination specifier: { ... platform:iOS }` and only `platform:macOS` destitations were available.

Now with your update to Xcode 26.2, it seems to be working 👍
https://github.com/tidev/titanium-sdk/actions/runs/21473518471/job/61851657778

**Edit:** Ah, I saw your discussion on Slack too late (I only use it in the browser, not the app).

macCesar:
@hansemannn @hbugdoll Thanks for clarifying!

So the failure was NOT caused by my PR, is that correct?

Sorry, I'm not very familiar with this type of CI/workflows! Still learning 🙂

Glad to hear everything is working now with Xcode 26.2!

hbugdoll:
> So the failure was NOT caused by my PR, is that correct?

Yep.

hbugdoll:
For the sake of completeness: GitHub's macOS 15 runner was changed...
- https://github.com/actions/runner-images/issues/13570
- https://github.com/actions/runner-images/pull/13575

PR #14371, opened 2026-01-12
chore(ios): completed Xcode task names for pretty build log

hbugdoll:
**Introduction:**

- build system's `invokeXcodeBuild()` add blank lines between Xcode tasks (as in the original `xcodebuild` log) and colored the first line of a task to make things easier to read
  - done for 23 Xcode task names so far
  - e.g. `GenerateAssetSymbols` or `WriteAuxiliaryFile` were missing

**Description:**

- completed Xcode task name list in `invokeXcodeBuild()`
  - now 63 Xcode task names 
- only relevant for log level `trace`

| Before  | After |
| ------------- | ------------- |
|  |  |


iphone/cli/commands/_build.js
```diff
@@ -7210,33 +7210,74 @@ class iOSBuilder extends Builder {
...
+					'AppIntentsSSUTraining ',
+					'ClangStatCache ',
+					'CodeSign ',
+					'CompileAssetCatalog ',
+					'CompileAssetCatalogVariant ',
+					'CompileC ',
+					'CompileStoryboard ',
+					'ComputePackagePrebuildTargetDependencyGraph',
+					'ComputeTargetDependencyGraph',
+					'Copy ',
```
hansemannn: The diff is larger than it needs to, as you've added empty spaces after each keyword consistently. If the empty space is needed, it can still be computed via:
```
.map(keyword => `${keyword} `)
```
afterwards.

hansemannn (changes requested):

iphone/cli/commands/_build.js
hbugdoll: Previously, a empty space was always added:
https://github.com/tidev/titanium-sdk/blob/75ebb2806bfee677fd4dd626f0ba1610f5f61ebc/iphone/cli/commands/_build.js#L7239

But some of the new task names stand alone in a line with no space following, e.g. `CreateBuildDescription` or `SendProjectDescription`.
I think there were false positives when I omitted the space in all the other keywords.

cb1kenobi:
I get what you're trying to do, but this is silly. If it were me, I would completely remove all of the colors.

But I like color too. I get it. I have to believe there's a much simpler solution that works for 90% of the scenarios using a regex that does some whitespace lookbehinds and lookaheads. It doesn't seem that complex.

iphone/cli/commands/_build.js
hbugdoll: Tested again, removed the spaces after the keyword where they were present and add an optional space `' ?'` at the end of the regex.

hbugdoll:
> I get what you're trying to do, but this is silly. If it were me, I would completely remove all of the colors.

The colors are just additional highlights (very useful in my opinion in the gray text desert at log level `trace`).
The main point is to restore the line formatting of the `xcodebuild` log a bit.
 
> But I like color too. I get it. I have to believe there's a much simpler solution that works for 90% of the scenarios using a regex that does some whitespace lookbehinds and lookaheads. It doesn't seem that complex.

I improved the regex, so that now all that remains is the adding of keywords and the optional space `' ?'`.

hansemannn (approved):

PR #14372, opened 2026-01-12
fix: improve changelog generation

hansemannn:
- Move to ESM
- Fix `Error in conventional-changelog-writer: Cannot modify immutable object.` error
- Remove employee-check -> everyone is community <3 
- Only render `## BREAKING CHANGES` block if there are breaking changes 

The 13.1.0 is ready after this one!


cb1kenobi (approved):
Love this!

PR #14379, opened 2026-01-21
fix(android): apply null checks on `localOverlayProxy`

prashantsaini1:
Based on some internal crash reports, the `localOverlayProxy` is null and prevents the activity to be paused properly.

- Also moved the surrounding code into try-catch to cover any hidden issues.

> Fatal Exception: java.lang.RuntimeException: Unable to pause activity {ti.modules.titanium.media.TiCameraActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'org.appcelerator.titanium.view.TiUIView org.appcelerator.titanium.proxy.TiViewProxy.getOrCreateView()' on a null object reference
       at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:4755)
       at android.app.ActivityThread.performPauseActivity(ActivityThread.java:4706)
       at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:4657)

### DIFF
--- a/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
+++ b/android/modules/media/src/java/ti/modules/titanium/media/TiCameraActivity.java
@@ -291,16 +291,27 @@ public class TiCameraActivity extends TiBaseActivity implements SurfaceHolder.Ca
 	{
 		super.onPause();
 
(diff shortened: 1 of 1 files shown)


m1ga (approved):
LGTM

PR #14381, opened 2026-01-23
chore(android): upgrade gradle plugin

m1ga:
Since we jumped to Kotlin 2.3 in the last PR the gradle build plugin is using the wrong version:
https://developer.android.com/studio/build/kotlin-d8-r8-versions (2.3 needs 8.13.2)

This will result in a build error when you run a 13.1.0 module in a an app. 

This PR fixes that.

Note: gradle plugin 9.0.0 was released a few days ago but the page still says 8.13.2 is required so I've used that version.


prashantsaini1:
I believe it's also time to use **4.4.4** for `com.google.gms:google-services:4.4.1`.

m1ga:
Sounds good. Change log for that is only:
```
Fixed issue where the plugin was eagerly realizing configurations. (GitHub [#307](https://github.com/google/play-services-plugins/pull/307))
Integration with the Crashlytics Gradle plugin is now aware of variants. (GitHub #293) ## Update - May 13, 2024
Internal improvements. ## Update - September 18, 2025
```
https://firebase.google.com/support/release-notes/android#google-services_plugin_v4-4-4

prashantsaini1:
Yeah, that actually puts us ahead to even update GP to 8.14.x since we already use Gradle 8.14.x as well.

prashantsaini1 (approved):

PR #14387, opened 2026-01-28
feat(cli): add Alloy app to create menu

m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/14386

Add an option to create an Alloy app into the create menu:

Bildschirmfoto_20260128_155300

**Notes**
I've added a new type `app_alloy` to have it in the menu. As soon as it's selected I'll store `isAlloy` to run the alloy command at the end and set the type back to `app` so it will use the same flow as before. In order to have less duplicated code I've created a base_app file and include that in `app_alloy` and `app`.

**Test:**
* `ti create`
* select 1 or 2
* check the project at the end


cli/commands/create.js
cb1kenobi: It's awkward to have the prompt default to `'app'` when you don't know you're supposed to type `'app_alloy'`. I propose we change the prompt to numbers instead. That means change the `default` from `'app'` to `'1'` (as a string).

Then you will need to change the `default` below on line 109 also to `'1'` (as a string).

cli/lib/creators/app_alloy.js
```diff
@@ -0,0 +1,39 @@
...
+	 * @param {Object} config - The CLI config
+	 * @param {Object} cli - The CLI instance
+	 */
+	constructor(logger, config, cli) { // eslint-disable-line no-unused-vars
+		super(logger, config, cli, {
+			title: 'Titanium App (Alloy)',
+			titleOrder: 2,
+			type: 'app_alloy'
+		});
+	}
```
cb1kenobi, line 37: If alloy is not installed, it blows up with:
```
'alloy' is not recognized as an internal or external command,
operable program or batch file.
[LiveView] Error: Command failed: alloy new "C:\Users\chris\projects\ti\myapp2"
    at genericNodeError (node:internal/errors:985:15)
    at wrappedFn (node:internal/errors:539:14)
    at checkExecSyncError (node:child_process:925:11)
    at execSync (node:child_process:997:15)
    at file:///C:/ProgramData/Titanium/mobilesdk/win32/13.2.0/cli/commands/create.js:189:6
    at file:///C:/Users/chris/projects/ti/titanium-cli/src/cli.js:400:26
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 73520,
  stdout: null,
  stderr: null
}
Warning: Detected unsettled top-level await at file:///C:/Users/chris/projects/ti/titanium-cli/src/main.js:8
        await cli.go();
        ^
```

This easliy remedied by adding:

```suggestion
	}
	
	init() {
		// check if alloy is installed
		const alloyInstalled = execSync('alloy --version', { stdio: 'ignore' });
		if (!alloyInstalled) {
			throw new Error('Alloy is not installed');
		}
		return super.init();
	}
```

Don't forget the `import { execSync } from 'node:child_process';` at the top.

cb1kenobi (changes requested):

cli/commands/create.js
m1ga: Good point! Same goes for `Apple Watch™ App` where you have to type applewatch. What about doing it like this:

Bildschirmfoto_20260218_104532

Changing from `app_alloy` to just `alloy` and then show the names of each one in the line. Most people will use the numbers in that menu but this way we'll show the proper names

cli/lib/creators/app_alloy.js
m1ga: good point :+1: will add that

m1ga: one issue (maybe Linux?):
```
execSync('alloy --version', { stdio: 'ignore' })
```
is returning NULL for me and doesn't show alloy even if it is installed. What works for me is:
```js
	init() {
		// check if alloy is installed
		try {
			execSync('alloy --version');
		} catch (err) {
			throw new Error('Alloy is not installed');
		}
		return super.init();
	}
```

would that be fine too?

cb1kenobi: Yup, try/catch works, good catch!

cli/commands/create.js
cb1kenobi: I don't love showing the name. It's confusing. I prefer to stick to numbers in the prompt as it makes the most sense. We show the `type` values in `ti create --help`, though I don't know why `alloy` isn't showing up on my local hacked version. I wonder if we need to do the alloy installed check in the constructor? I believe the creator constructor is currently outside the try/catch, so that would need to be adjusted.

+1 on changing `app_alloy` to `alloy`.

m1ga: I've switched to numbers:
Bildschirmfoto_20260218_185659

And the `--alloy` part is shown here:
Bildschirmfoto_20260218_185614

And yes, if you don't have alloy installed and run that command it will fail BUT the project is created as it will just go into that folder and run `alloy new`. I'll add a try/catch in create.js to compensate for that

m1ga: Done:
it will show
```
logger.error('Alloy is not installed. Please install it and run "alloy new" inside the project folder.');
```
if you don't have Alloy but still run `ti create --alloy`

cli/commands/create.js
```diff
@@ -179,9 +185,12 @@ export class CreateCommand {
 				} else {
 					logger.info(`Project created successfully in ${appc.time.prettyDiff(cli.startTime, Date.now())}\n`);
 				}
-
-				if (cli.argv.alloy !== undefined) {
-					execSync(`alloy new "${path.join(cli.argv['workspace-dir'], cli.argv.name)}"`, { stdio: 'inherit' });
+				if (cli.argv.alloy !== undefined || useAlloy) {
+					try {
+						execSync(`alloy new "${path.join(cli.argv['workspace-dir'], cli.argv.name)}"`, { stdio: 'ignore' });
```
cb1kenobi: Can we output the alloy compiler output like this:

```suggestion
						const output = execSync(`alloy new "${path.join(cli.argv['workspace-dir'], cli.argv.name)}"`, { encoding: 'utf8' });
						(output?.trim() || '').split('\n').forEach(line => logger.info(line));
```

cb1kenobi: Also, this code block should go in the "Project created successfully" else block. No sense running alloy if project creation failed. Run alloy before printing the project was created successfully. That should be the last log message.

cli/commands/create.js
```diff
@@ -179,9 +185,12 @@ export class CreateCommand {
...
+					} catch (_alloyError) {
+						logger.error('Alloy is not installed. Please install it and run "alloy new" inside the project folder.');
```
cb1kenobi: Technically this shouldn't happen. Regardless, maybe we should say run `npm i -g alloy` to install alloy?

```suggestion
						logger.error('Alloy is not installed. Run "npm i -g alloy" to install it, then run "alloy new" inside the project folder.');
```

cb1kenobi (changes requested):

cli/commands/create.js
m1ga: you are right, good catch. I've moved it there.

about the `ignore` part: When I run the command now it will show:
```
/bin/sh: Line 1: alloy: Command not found
[ERROR] Alloy is not installed. Run "npm i -g alloy" to install it, then run "alloy new" inside the project folder.
```
That's why I've used `ignore` to hide that not found message. But I think it's fine so you'll see that this is failing and  then the error message to help you

cli/commands/create.js
```diff
@@ -179,9 +185,13 @@ export class CreateCommand {
 				} else {
 					logger.info(`Project created successfully in ${appc.time.prettyDiff(cli.startTime, Date.now())}\n`);
 				}
-
-				if (cli.argv.alloy !== undefined) {
-					execSync(`alloy new "${path.join(cli.argv['workspace-dir'], cli.argv.name)}"`, { stdio: 'inherit' });
+				if (cli.argv.alloy !== undefined || useAlloy) {
```
cb1kenobi: Be sure to move this entire if block before line 186. :)

m1ga: oh, forgot to push that :-( Will do that tomorrow morning

cli/commands/create.js
```diff
@@ -178,10 +184,14 @@ export class CreateCommand {
 					logger.error(`Failed to create project after ${appc.time.prettyDiff(cli.startTime, Date.now())}\n`);
 				} else {
 					logger.info(`Project created successfully in ${appc.time.prettyDiff(cli.startTime, Date.now())}\n`);
-				}
-
-				if (cli.argv.alloy !== undefined) {
-					execSync(`alloy new "${path.join(cli.argv['workspace-dir'], cli.argv.name)}"`, { stdio: 'inherit' });
+					if (cli.argv.alloy !== undefined || useAlloy) {
```
cb1kenobi: I think this should be moved before the "Project created successfully".

m1ga: Bildschirmfoto_20260219_161500

vs

Bildschirmfoto_20260219_161615

Not sure if we should end with the error message or not. I see the argument that in the first one it will be more visible that the project is created.

cb1kenobi: I think what I am trying to suggest is we have multiple sad paths and one happy path.

- If the create fails
  - then show error
  - exit w/ code 1
- If useAlloy
  - run alloy new
  - if error
    - show error
    - exit w/ code 1
- show success message

m1ga: trying to move it above the message "Project created successfully" but I'm having troubles with github at the moment :disappointed: 

update: looks like it worked, guess it was some caching.

cb1kenobi: I was thinking we should `process.exit(1)`, but now I'm wondering we should throw and let the CLI catch it and exit w/ code 1.

m1ga: I've pushed an update that will show the `Project created successfully` message ONLY if it's really created.

So: `ti create --alloy` without Alloy installed will now just show the `ERROR` message as you wanted to have an Alloy project and it's not "successful". 
All other cases (alloy installed or creating a classic project) will show `Project created successfully` at the end.

cb1kenobi (approved):
This is perfect! Thanks for the back and forth!

PR #14391, opened 2026-02-07
fix(ios): support module builds with mac=true on Catalyst

macCesar:
## Summary

This PR fixes iOS module builds when `manifest` includes `mac: true` for Mac Catalyst support.

It addresses 3 required areas:

1. Ensure `TitaniumKit.xcframework` Mac Catalyst slice keeps expected framework symlinks.
2. Ensure new module templates include Catalyst-compatible Xcode build settings.
3. Ensure legacy modules are auto-patched at build time when `mac: true` is set.

### 1) TitaniumKit packaging
- `support/iphone/build_titaniumkit.sh`
- After `xcodebuild -create-xcframework`, restore symlinks in:
  - `ios-arm64_x86_64-maccatalyst/<Framework>.framework`

### 2) Module templates
- `iphone/templates/module/objc/template/ios/{{ModuleName}}.xcodeproj/project.pbxproj.ejs`
- `iphone/templates/module/swift/template/ios/{{ModuleName}}.xcodeproj/project.pbxproj.ejs`

Added in iOS build configurations:
- `SUPPORTED_PLATFORMS = "iphoneos iphonesimulator"`
- `SUPPORTS_MACCATALYST = YES`

### 3) Build-time auto-fix for existing modules
- `iphone/cli/commands/_buildModule.js`
- Added `ensureMacCatalystBuildSettings()`:
  - Runs only when `manifest.mac === 'true'`
  - Patches `project.pbxproj` with:
    - `SUPPORTED_PLATFORMS = "iphoneos iphonesimulator"`
    - `SUPPORTS_MACCATALYST = YES`

Also improved module build diagnostics and SDK consistency:
- Logs `ti` command exit code and full command.
- Propagates `--sdk <version>` to staged temp app create/build when provided.

## Validation

Validated locally by building SDK and module with this branch:
- `npm run cleanbuild -- ios`
- `ti build -p ios --build-only --sdk 13.2.0` on a module with `mac: true`

Result:
- Module build succeeds.
- Generated xcframework includes iOS device, iOS simulator, and Mac Catalyst slices.

### DIFF
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -130,15 +130,14 @@ export class iOSModuleBuilder extends Builder {
 
 		const pbxFilePath = path.join(this.projectDir, `${this.moduleName}.xcodeproj`, 'project.pbxproj');
 		if (!fs.existsSync(pbxFilePath)) {
-			this.logger.warn(`Unable to auto-fix Mac Catalyst build settings. File not found: ${pbxFilePath}`);
+			this.logger.warn(`Unable to verify Mac Catalyst build settings. File not found: ${pbxFilePath}`);
 			return;
 		}
 
 		const xcodeProject = xcode.project(pbxFilePath).parseSync();
 		const configurations = xcodeProject.hash.project.objects.XCBuildConfiguration;
 		const desiredSupportedPlatforms = '"iphoneos iphonesimulator"';
-		const touchedConfigurations = [];
-		let shouldWrite = false;
+		const misconfigured = [];
 
 		for (const key of Object.keys(configurations)) {
 			const configuration = configurations[key];
@@ -151,27 +150,20 @@ export class iOSModuleBuilder extends Builder {
 				continue;
 			}
 
-			let didChangeConfiguration = false;
-
-			if (buildSettings.SUPPORTED_PLATFORMS !== desiredSupportedPlatforms) {
-				buildSettings.SUPPORTED_PLATFORMS = desiredSupportedPlatforms;
-				didChangeConfiguration = true;
-			}
-
-			if (buildSettings.SUPPORTS_MACCATALYST !== 'YES') {
-				buildSettings.SUPPORTS_MACCATALYST = 'YES';
-				didChangeConfiguration = true;
-			}
-
-			if (didChangeConfiguration) {
-				shouldWrite = true;
-				touchedConfigurations.push(configuration.name);
+			if (buildSettings.SUPPORTED_PLATFORMS !== desiredSupportedPlatforms
+				|| buildSettings.SUPPORTS_MACCATALYST !== 'YES') {
+				misconfigured.push(configuration.name);
 			}
 		}
 
-		if (shouldWrite) {
-			fs.writeFileSync(pbxFilePath, xcodeProject.writeSync());
-			this.logger.info(`Auto-fixed Mac Catalyst build settings in ${path.basename(pbxFilePath)} (${Array.from(new Set(touchedConfigurations)).join(', ')}) because manifest has "mac: true".`);
+		if (misconfigured.length > 0) {
+			const configs = Array.from(new Set(misconfigured)).join(', ');
+			this.logger.error(`Mac Catalyst build settings are not configured correctly in ${path.basename(pbxFilePath)} (${configs}).`);
+			this.logger.error('Open your Xcode project and for each affected configuration set:');
+			this.logger.error('  • SUPPORTS_MACCATALYST = YES');
+			this.logger.error('  • SUPPORTED_PLATFORMS = iphoneos iphonesimulator');
+			this.logger.error('Then run the build again.\n');
+			process.exit(1);
 		}
 	}
 
--- a/iphone/cli/commands/_buildModule.js
+++ b/iphone/cli/commands/_buildModule.js
@@ -122,6 +122,59 @@ export class iOSModuleBuilder extends Builder {
 		return isMacOSEnabled;
 	}
 
+	ensureMacCatalystBuildSettings() {
+		// Keep existing behavior untouched unless module author explicitly enables mac builds.
+		if (this.manifest.mac !== 'true') {
+			return;
+		}
+
+		const pbxFilePath = path.join(this.projectDir, `${this.moduleName}.xcodeproj`, 'project.pbxproj');
+		if (!fs.existsSync(pbxFilePath)) {
+			this.logger.warn(`Unable to auto-fix Mac Catalyst build settings. File not found: ${pbxFilePath}`);
+			return;
+		}
+
+		const xcodeProject = xcode.project(pbxFilePath).parseSync();
+		const configurations = xcodeProject.hash.project.objects.XCBuildConfiguration;
+		const desiredSupportedPlatforms = '"iphoneos iphonesimulator"';
+		const touchedConfigurations = [];
+		let shouldWrite = false;
+
+		for (const key of Object.keys(configurations)) {
+			const configuration = configurations[key];
+			if (typeof configuration !== 'object' || !configuration.buildSettings) {
+				continue;
+			}
+
+			const buildSettings = configuration.buildSettings;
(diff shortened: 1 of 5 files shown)


cb1kenobi:
This is awesome! The code looks good, but I would want it to be tested. I'm not quite able to test it yet.

m1ga:
when I create a new module with the PR SDK, keep everything as it is and then run the build commandwill fail with
```
[TRACE] [xcode-macos] /Users/miga/Library/Application Support/Titanium/mobilesdk/osx/13.2.0.v20260207034456/iphone/include/TiModule.h:1:9: error: 'TitaniumKit/TiModule.h' file not found (in target 'test_mac' from project 'test_mac')
[TRACE] [xcode-macos]     /Users/miga/test_mac/ios/Classes/TestMacModule.m:8:9: note: in file included from /Users/miga/test_mac/ios/Classes/TestMacModule.m:8:
[TRACE] [xcode-macos]     /Users/miga/test_mac/ios/Classes/TestMacModule.h:8:9: note: in file included from /Users/miga/test_mac/ios/Classes/TestMacModule.h:8:
[TRACE] [xcode-macos]     note: Did not find header 'TiModule.h' in framework 'TitaniumKit' (loaded from '/Users/miga/Library/Developer/Xcode/DerivedData/test_mac-drmjdjycerxklleagbrwkxzhjxks/Build/Intermediates.noindex/ArchiveIntermediates/test_mac/BuildProductsPath/Release-maccatalyst')
```
setting `mac: false` in the manifest will build the module

macCesar:
Hey Michael! Thanks for testing this.

I think the symlinks in the Mac Catalyst slice of `TitaniumKit.xcframework` might be getting lost during SDK installation. When I build from source locally, the symlinks stay intact and everything works. But depending on how the SDK gets installed (ZIP extraction, etc.), they could break.

Could you check two things?

1. How did you install the PR SDK? (`ti sdk install` from a CI URL, manual ZIP download, etc.)

2. Can you verify the symlinks in the maccatalyst slice are still there? Run this:
   ```
   ls -la "~/Library/Application Support/Titanium/mobilesdk/osx/13.2.0.v20260207034456/iphone/Frameworks/TitaniumKit.xcframework/ios-arm64_x86_64-maccatalyst/TitaniumKit.framework/"
   ```
   You should see symlinks like `Headers -> Versions/Current/Headers`. If those show up as regular directories instead of symlinks (no `@` symbol), that's the problem.

The maccatalyst slice uses a versioned macOS framework structure (`Versions/A/`) that depends on symlinks. I added a fix in `build_titaniumkit.sh` to restore them after `xcodebuild -create-xcframework`, but they might not survive the packaging and extraction pipeline.

I'm also looking into changing the build destination to produce a flat framework structure that wouldn't need symlinks at all, which would fix the root cause instead of patching it.

m1ga:
the folder only contains der "Versions" folder, no symlinks. I've took the SDK from the workflow action and ran `ti sdk install filename.zip` 

will test a local build over the weekend

macCesar:
Hey Michael, thanks for that output, it confirmed the issue. The `Versions` folder without symlinks is exactly what I suspected.

The problem is that `ti sdk install` (ZIP extraction via node-appc) doesn't preserve the symlinks in the Mac Catalyst slice of TitaniumKit.xcframework. Built from source they're intact, but after ZIP extraction they're gone.

I pushed a fix in `475a34ec68` that takes the same approach used for app builds in `_build.js`: restore the symlinks at module build time, right before the Mac Catalyst xcodebuild step. Works regardless of how the SDK was installed.

Also removed the previous `build_titaniumkit.sh` patch since the fix now happens at the right time.

Can you pull the latest and test again when you get a chance?

m1ga:
nice!

```
Command
  /usr/local/bin/node /usr/local/bin/ti build -p ios -b --sdk 13.2.0.v20260212205238

[INFO]  Project directory: /Users/miga/testmod/ios
[INFO]  Module ID: test.mod
[INFO]  Module Type: Static Library (Objective-C)
[WARN]  Please update the LICENSE file with your license text before distributing.
[INFO]  Created 5 Mac Catalyst framework symlinks for TitaniumKit
[INFO]  Creating universal framework/library
[INFO]  Looking for /Users/miga/testmod/ios/build/iphoneos.xcarchive/Products/usr/local/lib/libtest.mod.a
[INFO]  Looking for /Users/miga/testmod/ios/build/iphonesimulator.xcarchive/Products/usr/local/lib/libtest.mod.a
[INFO]  Looking for /Users/miga/testmod/ios/build/macosx.xcarchive/Products/usr/local/lib/libtest.mod.a
[INFO]  Verifying universal framework/library
[INFO]  Simulator has architectures: arm64,x86_64
[INFO]  Mac has architectures: arm64,x86_64
[INFO]  Device has architectures: arm64
[INFO]  Binary framework/library file: /Users/miga/testmod/ios/build/test.mod.xcframework
[INFO]  Creating module zip
[INFO]  Writing module zip: /Users/miga/testmod/ios/dist/test.mod-iphone-1.0.0.zip
[INFO]  Project built successfully in 10s 895ms
```

That worked!

Bildschirmfoto_20260213_095743

That's so great that you can finally do a "ti create" (module) and run a build in the ios folder and it works. At the moment you always had to go in and set `mac:false` in order to build, which was a strange first impression when you start an ios module.

macCesar:
Thanks for testing, Michael! Glad it worked on your end too.

And yeah, good call on the `mac:false` workaround — that was a bad first impression for anyone creating a new iOS module. Now it should just work out of the box.

Hopefully this can get merged soon!

iphone/cli/commands/_buildModule.js
```diff
@@ -122,6 +122,111 @@ export class iOSModuleBuilder extends Builder {
...
+
+			const buildSettings = configuration.buildSettings;
+			if (buildSettings.SDKROOT !== 'iphoneos') {
+				continue;
+			}
+
+			let didChangeConfiguration = false;
+
+			if (buildSettings.SUPPORTED_PLATFORMS !== desiredSupportedPlatforms) {
+				buildSettings.SUPPORTED_PLATFORMS = desiredSupportedPlatforms;
```
hansemannn: It should rather error out in this case and ask to update the build settings, so the physical xcode project state is not out of sync with the build state.

iphone/cli/commands/_buildModule.js
```diff
@@ -122,6 +122,111 @@ export class iOSModuleBuilder extends Builder {
...
+				didChangeConfiguration = true;
+			}
+
+			if (buildSettings.SUPPORTS_MACCATALYST !== 'YES') {
+				buildSettings.SUPPORTS_MACCATALYST = 'YES';
```
hansemannn: Same here

iphone/cli/commands/_buildModule.js
```diff
@@ -122,6 +122,111 @@ export class iOSModuleBuilder extends Builder {
...
+
+			if (didChangeConfiguration) {
+				shouldWrite = true;
+				touchedConfigurations.push(configuration.name);
+			}
+		}
+
+		if (shouldWrite) {
+			fs.writeFileSync(pbxFilePath, xcodeProject.writeSync());
+			this.logger.info(`Auto-fixed Mac Catalyst build settings in ${path.basename(pbxFilePath)} (${Array.from(new Set(touchedConfigurations)).join(', ')}) because manifest has "mac: true".`);
```
hansemannn: A build command should never alter source files. I get the "ease of use" idea, but any irregular behavior should error out instead

hansemannn (changes requested):

macCesar:
Hey Hans, fair point. Fixed in badf897dc2 — the function no longer touches the `.pbxproj`. It now checks for misconfigured build settings and exits early with an error that tells the developer exactly what to update in Xcode (`SUPPORTS_MACCATALYST = YES`, `SUPPORTED_PLATFORMS = iphoneos iphonesimulator`).

hansemannn (approved):
Very well, thank you for another excellent contribution 🤘

PR #14394, opened 2026-02-23
Switch to oxlint

cb1kenobi:
Replace eslint with oxlint. Oxlint is written in Rust and way faster than eslint.

The motivation for this stems from the eslint version being out-of-date and didn't support modern JS patterns such as initializing class properties.

The eslint ignore patterns have been added to the oxlint config.

Nothing was kept from the eslint config. From what I can tell, the overrides and the old axway config plugin were used to declare globals and oxlint doesn't seem to care.

I removed the eslint call from the danger file. I'm not entirely sure what danger is good for.


hansemannn (approved):
Very nice! We should also archive the away-eslint plugin repos, but thats another topic.

PR #14396, opened 2026-02-23
feat(android): move ScrollableView to ViewPager2

### LINKED ISSUE #13690: Android: migrate ViewPager to ViewPager2

### Description

Currently the SDK still uses the normal ViewPager. Migrating to ViewPager2 would allow to use vertical orientation, RTL support and it's based on RecyclerView tech.

https://developer.android.com/develop/ui/views/animations/vp2-migration

### Solution

-

### Alternatives

-

### Platforms

Android

m1ga:
fixes: https://github.com/tidev/titanium-sdk/issues/13690

the TabGroup stlil uses the ViewPager1 so I've kept it in the build.gradle file for now as this will only move the ScrollableView to ViewPager2. ViewPager2 supports RTL layouts or vertical orientation.

**New feature:**
* `scrollType` with `vertical` (default is `horizontal`). Same as the `ScrollView`

**Tested:**
* internal Ti.UI.ScrollableView tests
* kitchensink
* padding example from the docs
* `scrollingEnabled` true and false

**Test:**
use the KitchenSink App: Views - ScrollableView and with https://github.com/tidev/kitchensink-v2/pull/75 for `scrollType: "vertical"`

**Note:** 
:robot: AI generated! Have to do some more testing, but my first test in a real app looked fine already :+1: 

[Bildschirmaufnahme_20260223_220124.webm](https://github.com/user-attachments/assets/2024ee97-cb80-4762-9f6b-1012344c0b49)


apidoc/Titanium/UI/ScrollableView.yml
```diff
@@ -329,6 +329,17 @@ properties:
...
+  - name: scrollType
+    summary: Direction of the ScrollableView (`horizontal` or `vertical`).
+    description: |
+        The default ScrollableView will add multiple views horizontally. Setting it to `vertical`
+        will create a vertical list of views. Make sure to set `showPagingControl` to false to hide
+        the left/right arrows.
+    type: String
+    default: 'horizontal'
+    platforms: [android]
+    since: "12.2.0"
```
m1ga:
```suggestion
    since: "13.2.0"
```

.github/workflows/docs.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         with:
             fetch-depth: 0
       - name: Use Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
-          node-version: '20.x'
+          node-version: '24'
           cache: 'npm'
```
hansemannn: This does not relate to this PR

.github/workflows/regen-docs.yml
```diff
@@ -21,9 +21,9 @@ jobs:
         fetch-depth: 0
 
     - name: Setup node
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6
       with:
-        node-version: '16'
+        node-version: '24'
```
hansemannn: Same

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollableView.java
```diff
@@ -587,134 +557,78 @@ public ViewPagerAdapter(Activity activity, ArrayList<TiViewProxy> viewProxies)
...
-			// Remove the item's view from the ViewPager.
-			// Note: The Titanium view's is wrapped by a "TiCompositeLayout" parent view.
-			if (object instanceof View) {
-				ViewParent parentView = ((View) object).getParent();
-				if (parentView instanceof ViewGroup) {
-					if ((parentView instanceof ViewPager) == false) {
-						((ViewPager) container).removeView((View) parentView);
-					}
-					((ViewGroup) parentView).removeView((View) object);
-				}
```
hansemannn, line 606: Why is this not necessary anymore? Does it more layout magic internally?

android/titanium/build.gradle
```diff
@@ -299,6 +299,7 @@ dependencies {
 	implementation 'androidx.vectordrawable:vectordrawable:1.2.0'
 	implementation 'androidx.vectordrawable:vectordrawable-animated:1.2.0'
 	implementation 'androidx.viewpager:viewpager:1.1.0'
```
hansemannn, line 301: Still needed?

.github/workflows/docs.yml
m1ga: This was the fix for the docs runner so the PR will be green. But it's merged already in a different PR.

android/titanium/build.gradle
m1ga: See the first line of this PR 😄  "the TabGroup stlil uses the ViewPager1 so I've kept it in the build.gradle file for now"

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollableView.java
m1ga: There is not `removeView()` method anymore ([docs](https://developer.android.com/reference/kotlin/androidx/viewpager2/widget/ViewPager2)). The Adapter parts is handling that now
Bildschirmfoto_20260315_123541

android/titanium/build.gradle
hansemannn: Oh, sorry! Read too fast here. You're right!

hansemannn:
PR looks and works good - lete know when it's ready for review

m1ga:
Thanks. Will run some last tests and then we should be good to go :+1:

m1ga:
Tested it with two different client apps where I didn't write the initial code and both behave the same way with the ViewPager2. And since all other tests worked fine too we should be good to go here :+1:

hansemannn (approved):

PR #14399, opened 2026-02-28
feat(android): prepare SDK for Android target SDK 36

### LINKED ISSUE #14353: Android: Target SDK 36

### Description

Just a ticket so we can prepare the Android part for the next target API level update to 36.

In 36 our edge-to-edge workaround is getting remove:
https://developer.android.com/about/versions/16/behavior-changes-16#edge-to-edge
so we have to make sure all UI elements work in edge-to-edge mode.

### Solution

Make all UI elements edge-to-edge ready

### Alternatives

_No response_

### Platforms

Android

m1ga:
(hopefully) fixes https://github.com/tidev/titanium-sdk/issues/14353

Targeting Android SDK 36 will remove the current opt-out XML for the edge-to-edge stuff and this will break our layouts:

Screenshot_20260227-220449<br>
_(Ti 13.1.1.GA with targetSDK 36)_

**Whats fixed:**

This PR will remove the opt-out part from the XML file, adds a new `EdgeToEdgeHelper` that will set the correct padding and fixes the (current/old) BottomNavigation height calculation. Otherwise the views are slightly behind the bottom navigation again. The `experimental:true` (new) BottomNavigation using the XML is working fine.

Screenshot_20260227-220530<br>
_(this PR)_

**How to test it**
* add `<uses-sdk android:targetSdkVersion="36" />` in the `<manifest>` node in your tiapp.xml
* build 13.1.1 and this PR and check the top and bottom of your windows

**Some images:**

Screenshot_20260228-114547
Screenshot_20260228-114556
Screenshot_20260228-114623
Screenshot_20260228-114639
Screenshot_20260228-114705

**Notes:**

This will make sure a bottomnavigation will be behind the navigation bar BUT the content will stay above it. So it technically is not using edge-to-edge in order to make existing apps look the same without any changes.  It is not using safeArea or allows content behind the bottom navigation. This should be implemented at some point. But this PR will just make sure we have something that is Android SDK 36 ready once there is a requirement by Google!

Removing the current bottom padding for normal windows will make them look like this: 
Screenshot_20260227-221432
so you have to make sure to set a proper bottom position (like iOS)


cb1kenobi:
Wow, excellent work!

hbugdoll:
Very good. I was just about to ask how things were going on with target API level 36...

hansemannn:
Is the app icon <> title spacing also that way natively? It clips to the app icon

m1ga:
Don't think so, but this is how it looks with the current SDK too if I remember correctly.

Ti also still use Actionbars in some places. We would need to move all those to Toolbars to have a better look in general.

m1ga:
I've added some more tasks to look at in the first post of the issue:
https://github.com/tidev/titanium-sdk/issues/14353#issue-3682411066
we don't need to fix all right away as some are only deprecated but currently still work. This PR will already work with Android 36 but I'll see if I can add some more changes to it.

m1ga:
big update:


**Fix: Predictive Back Gesture Support**
- Added android:enableOnBackInvokedCallback="true" to all 3 manifest files
- TiBaseActivity.java: Added OnBackInvokedCallback registration in onCreate(), unregistration in onDestroy(), extracted shared handleBackNavigation() method
- TiCameraActivity.java / TiCameraXActivity.java: Added handleBackNavigation() override so camera back handling works with predictive back

**Fix: Gradle Config**
- gradle.properties: Updated suppressUnsupportedCompileSdk from 35 to 36

**Fix: Migrate setSystemUiVisibility() to WindowInsetsController**
- TiBaseActivity.java: Replaced setFullscreen() and windowCreated() to use WindowInsetsControllerCompat
- WindowProxy.java: Replaced 3 occurrences (window creation + 2 property change handlers)
- TabGroupProxy.java: Replaced light status bar handling
- TiToolbar.java: Replaced status bar overlap handling
- TiUIBottomNavigationTabGroup.java / TiUIBottomNavigation.java: Replaced light navigation bar handling

**Fix: Orientation Restrictions on Large Screens**
- TiWindowProxy.java, DecorViewProxy.java, ActivityProxy.java: Added large screen detection (>600dp) with warnings when fixed orientation is requested on Android 16+

**Fix: Remove Reflection**
- TiToolbarStyleHandler.java: Removed reflection on private mMaxButtonHeight field, kept setMinimumHeight() + requestLayout()


And since statusBarColor is not available anymore in 36+ I'm using the backgroundColor of the e.g. TabGroup to fill that color. That way you can now do:
Screenshot_20260528-180512
Screenshot_20260528-180538

before it was just black/white.

apidoc/Titanium/UI/Window.yml
```diff
@@ -1373,6 +1373,11 @@ properties:
     type: [Number]
     platforms: [android]
     since: {android: "12.5.0"}
+    description: |
+        Starting with Android 16 (target SDK 36), the status bar is always transparent as part of
+        edge-to-edge enforcement, and `statusBarColor` is ignored by the system. On Android 16+,
+        the status bar area will show the window's `backgroundColor` instead. Use `backgroundColor`
+        to control the color behind the status bar on Android 16+.
```
hansemannn, line 1386: Should we deprecate it already, as it's still usable, but not encouraged to be set anymore? It probably depends on the app's target SDK level to decide, but looking at how Android itself deprecates, it may make sense.

m1ga: It still works on an Android 11 phone so it's just for Anrdoid 16+ where the edge-to-edge design is used that it won't work.  We could add a warning log that if you use `statusBarColor` and your phone is Android 16+ it will show that you have to set a backgroundColor. On the other hand: you would see that on your device too if you test it (it's black/transparent or has your backgroundColor already).

m1ga:
successfully tested it on three apps today (two are mine, one is a classic client app with 300 files). All run fine and the only change I had to do was to adjust the "statusbar" color. Because there is no area there now (edge-to-edge) you have to set the backgroundColor of the window now to compensate that color. Everything else looks the same as with 13.3.1.GA :+1: 

Anyone else tested it with real apps already?

hansemannn:
@prashantsaini1 FYI

prashantsaini1:
@m1ga I can give it a full smoke test this weekend on our apps.

prashantsaini1:
@m1ga Our apps also work quite good with this PR. Tested these elements/scenarios:
- full screen window
- navigation-window
- bottom-sheet
- sliding panel
- scroll-view
- table-view
- list-view
- text-area with keyboard
- menu items
- **action bar** (it now shows elevation/shadow at the top edge, tried to find some solutions, but rather investing efforts here, I would try to move this to MaterialToolbar - low priority and does not affect apps visual look much).
- Notifications (to ensure that layout changes do not affect anything here as our apps do lots of redirects)
- Deep Links
- Back behaviour on different types of systems like window/nav-window/bottom-sheet, etc.

_There are couple of handling we would need to do in our apps, but definitely not via the SDK._ 

**Pass score: 8/10.** 🚀  (keeping two ratings to cover any untouched use case so far)

m1ga:
@prashantsaini1 thanks for the deep testing! That sounds great already. I'll definitely try to find some other apps to test it with. I had the "shadow issue" in one app too but there they were using a layout with an actionbar and where hiding it in code right away. I've simply moved it to a theme without actionbar and it was fixed :smile: But I'll test the other scenario again as it looks like to be a bug

hansemannn:
Shall we go with this one already then?

m1ga:
found the issue:
Screenshot_20260726-190750

it was only hiding the Actionbar but not the ActionbarContainer so that was still visible. Fixes in ActionbarProxy.java (with that I can use the "wrong theme" and the old code and it looks fine).

Since hiding it after the window was open shows a slight jump I've checked the `navBarHidden` property and that can be used now as a creation property :+1: 

TL;DR: `actionbar.hide()` works again; property `navBarHidden` works again

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -29,15 +30,22 @@ public class ActionBarProxy extends KrollProxy
 	private static final String TAG = "ActionBarProxy";
 	private static final String ACTION_BAR_NOT_AVAILABLE_MESSAGE = "ActionBar is not enabled";
 
+	private final AppCompatActivity activity;
```
prashantsaini1: @m1ga Can you please make it a `WeakReference` to avoid activity leak? [example reference](https://github.com/tidev/titanium-sdk/blob/8cae2200bc1b65530c3348d430112e34c074f6a3/android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java#L79)

prashantsaini1:
As a general discussion, usually apps often need to adjust safe area padding for individual elements like some primary action button at the bottom, or a fab button, or maybe an input field.
 
- As of now, every such element has to be handled using the `postlayout` event on every window. This makes a lot of repeated code just to adjust the positioning.
- What if we could have one property to add on any type of element for its inset positioning and the SDK would handle it internally. e.g., `extendSafeArea` could be reused for other such elements too and SDK can position these elements on every layout change.

This would also enable us to make ScrollView, ListView and TableView have auto spacing at the bottom after the last child element.

This sounds more future proof to me to let SDK handle things internally and apps can focus on core UI/functionality.

Happy to hear your thoughts!

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
m1ga: good catch :+1: It's adjusted: https://github.com/tidev/titanium-sdk/pull/14399/commits/747f6029ac0cd4be9b18e2c2f4c02892606b3845

> Both fields are now weak references in ActionBarProxy.java:
> - activityRef and toolbarRef replace the strong final fields, with a comment explaining why (a JS-retained proxy must not pin a destroyed activity or its view hierarchy).
> - Each method now snapshots the toolbar once into a local via getToolbar(), so the reference can't clear between the null check and use within a single call.
> - getSupportActionBar() dereferences activityRef the same way.

m1ga:
That sounds great if we could reuse that for an automatic real edge-to-edge design at the end! I've honestly only tested it with some custom padding which would require lots of changes to your app. If we can can solve it using the safeArea that would be perfect.

prashantsaini1 (approved):

hansemannn (approved):

PR #14403, opened 2026-03-13
fix: upgrade core-js to remove baseline-browser-mapping warning

m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/14388

as far as I can see the `[baseline-browser-mapping] The data in this module is over two months old.  To ensure accurate Baseline data, please update: `npm i baseline-browser-mapping@latest -D`` warning is gone. 

**Test**
build an app with 13.1.1.GA and you'll see the warning in the first line of the build output.
With this PR I don't see it anymore


hbugdoll:
Tested with Ti SDK 13.1.1 - building app and module works fine.

cb1kenobi (approved):

PR #14405, opened 2026-03-14
feat(android): use ConcurrentHashMap in TiRhelper and TiUIHelper

m1ga:
Some AI recommendations:

**High Priority**: Replace Collections.synchronizedMap() in TiRHelper.java - this is called frequently during resource lookups
**Medium Priority**: Replace Collections.synchronizedMap() in TiUIHelper.java - used for font and image caching

These changes improve concurrent read/write performance by using ConcurrentHashMap which provides:
- Non-blocking reads
- Fine-grained locking (lock striping) instead of global synchronization
- Better scalability under contention

There was one issue with fonts caching that is fixed with:

 ConcurrentHashMap doesn't allow null values, but the original code cached negative lookups with null. Changes made:
1. Added null check for fontFamily 
2. Removed the synchronized block (ConcurrentHashMap handles concurrency)
3. Removed caching of negative lookups (the put(null) at line 452)

Tested two apps, kitchensink and hyperloop examples. All worked fine, they are pretty fast so I didn't see any big improvement :-)

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
@@ -14,9 +14,9 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/util/TiUIHelper.java
```diff
@@ -438,18 +439,14 @@ private static Typeface loadTypeface(Context context, String fontFamily)
 				if (f.equalsIgnoreCase(fontFamily)
 					|| f.toLowerCase().startsWith(fontFamily.toLowerCase() + ".")) {
 					Typeface tf = Typeface.createFromAsset(mgr, customFontPath + "/" + f);
-					synchronized (mCustomTypeFaces)
-					{
-						mCustomTypeFaces.put(fontFamily, tf);
-					}
```
hansemannn, line 444: Are you sure the `synchronized` can be removed? Dealing with this in Titanium is different to normal apps (regarding main thread access).

m1ga: as far as I read in posts like this: https://stackoverflow.com/a/25570488 or https://stackoverflow.com/a/2929410
"ConcurrentHashMap is thread safe without synchronizing the whole map. Reads can happen very fast while write is done with a lock."

I didn't have any issues runnig multiple bigger apps with it. The only issue I had was with the `mCustomTypeFaces.put(fontFamily, null);` part that the old implementation added when the font was not there and `ConcurrentHashMap` can't handle the null. that's why there are these changes in the Font part so it will return null early in the process and not add that part. That whole part is used by AttributedStrings, TiIconDrawable (TabGroup/BottomNavigation Icons) or a custom fontFamily

hansemannn (approved):
Thanks for the feedback approved!

PR #14409, opened 2026-03-15
fix(android): fix blob LruCache crash on recycled bitmaps

hansemannn:
## Summary

- Fixes Android crash: `TiBlobLruCache.sizeOf() is reporting inconsistent results!` when repeatedly resizing images via `imageAsResized()` under memory pressure
- Tracks original bitmap sizes in a `ConcurrentHashMap` so `sizeOf()` returns consistent values even after a bitmap is recycled externally, satisfying the `LruCache` contract
- Overrides `entryRemoved()` to clean up tracked sizes when entries are evicted

## Root Cause

PR #14156 made `sizeOf()` return `0` for recycled bitmaps to suppress warnings. This broke `LruCache` internal accounting — when `evictAll()` (called from `onTrimMemory`) subtracted `0` instead of the original size, the cache ended up with an empty map but non-zero size counter, triggering an `IllegalStateException` crash.

## Test plan

- [ ] Run the reproduction case from #14232: open photo gallery, select and resize 4+ different images in sequence on a Pixel device
- [ ] Verify no crash and `imageAsResized()` returns valid blobs each time
- [ ] Verify no "Called getRowBytes() on a recycle()'d bitmap!" warnings in logcat
- [ ] Test `onTrimMemory` / `onLowMemory` paths by simulating memory pressure while images are cached

Fixes #14232

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/util/TiBlobLruCache.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/util/TiBlobLruCache.java
@@ -9,6 +9,8 @@ package org.appcelerator.titanium.util;
 import android.graphics.Bitmap;
 import androidx.collection.LruCache;
 
+import java.util.concurrent.ConcurrentHashMap;
+
 public class TiBlobLruCache extends LruCache<String, Bitmap>
 {
 	// Get max available VM memory, exceeding this amount will throw an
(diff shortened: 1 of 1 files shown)


m1ga (approved):
working fine. Was able to open 10 images in a row with the example code and no crash or bitmap warning 👍  Before it was having issues with ~4 images

PR #14410, opened 2026-03-18
chore(android): update gradle

m1ga:
Changelog: https://docs.gradle.org/9.4.0/release-notes.html
"Supports Java 26"

**Tested** (Linux only)
- SDK build
- module build
- app build (kitchensink, hyperloop-examples, custom app)
- 13.2.0 module in 13.1.1 app

all works fine with OpenJDK 25 and ~OpenJDK 26~

**Changes:**
* version number update
* android/titanium/libv8-services.js (-b was removed)
* android/build.gradle to fix the untar part

gradlew change is automatically done when importing the new gradle. We should keep the naming in the file as it is. 

**NOTE:**
* https://github.com/tidev/hyperloop.next/pull/420 is needed because of gradle 8 -> gradle 9 switch


m1ga:
This PR will opt-out the new gradle 9 "build-in kotlin" stuff. You'll see some warnings that we have to remove it before switching to gradle 10. But apps and modules build fine :+1: Hyperloop needs a patch in order make those build.gradle files work with gradle 9.

m1ga:
Now tested with Java 26 :+1: Fedora 44 removed everything below Java 25 so I have to use this PR now to build apps again :smile:

hansemannn (approved):
Happy to merge this! Let's hope it doesn't have any affects to older modules

m1ga:
> Happy to merge this! Let's hope it doesn't have any affects to older modules

existing modules from a 13.2.0 app still work fine when building with this SDK (Java 21, 25 and ~26~).
Your app builds fine with this PR and all modules. gradle will show a few _warnings_ in some modules like:
```
Activity#ti.deeply.DeepLinkHandlerActivity@android:exported was tagged at AndroidManifest.xml:91 to replace other declarations but no other declaration present
```
but the app still builds and run.

m1ga:
**note:** @hansemannn  we only have Java 25 support, not Java 26! There is an issue with the gradle plugin  https://issuetracker.google.com/issues/486844145 and due to gradle caching I was still using java 25 when I've switched to Java 26 and compiled the SDK and apps (looks like I always have to restart after switching java versions).

But Java 25 works fine (apps + sdk build). So when we update the compatibility matrix: 13.3.0 support is up to 25 now

PR #14411, opened 2026-03-19
feat(android): improve pinch&zoom in ImageView

m1ga:
Will improve pinch & zoom (especially zoom out) in an Android ImageView

```js
	image:"https://picsum.photos/1000/2000",
	enableZoomControls:true,
	width: Ti.UI.FILL,
	height: Ti.UI.FILL,
})
win.add(img);
win.open();
```

**Test:**
* run the code above and zoom in & out
* Ti 13.1.1: you sometimes are stuck in "moving mode" when you try to zoom out
* This PR: should be able to zoom out 

**Note:**
* there is https://github.com/tidev/titanium-sdk/pull/14227 to improve the behavior in ScrollableViews

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiImageView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiImageView.java
@@ -454,6 +454,9 @@ public class TiImageView extends ViewGroup
 		/** Last scale center point used during a pinch-zoom along the y-axis. */
 		private float lastFocusY;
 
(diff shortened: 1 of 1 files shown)


prashantsaini1:
@m1ga Thanks for the update. Can we take both PRs into one since they are mainly for same feature to make it properly working?

m1ga:
of course! I thought it might be easier to have this one separated but I'll see if I can combine it into yours

m1ga:
now included in https://github.com/tidev/titanium-sdk/pull/14227

PR #14413, opened 2026-03-28
feat(android): update cpp files for unpachted v8 versions

m1ga:
just a draft PR for an unpatched v8 version: https://github.com/tidev/v8_titanium/pull/57

Changes also include:

1. V8Runtime.cpp:80-110 - Fixed O(n²) string concatenation in krollLog:
- Uses std::string for accumulation (O(n))
- Only converts to V8 strings at the end for logging

2. V8Util.cpp - Added JSON.stringify caching:
- Added static jsonStringifyFunction Persistent to cache the function
- First call looks up JSON.stringify and caches it
- Subsequent calls use the cached function directly
- Updated dispose() to clean up the cached function

### DIFF
--- a/android/runtime/v8/src/native/V8Runtime.cpp
+++ b/android/runtime/v8/src/native/V8Runtime.cpp
@@ -6,6 +6,7 @@
  */
 #include <stdio.h>
 #include <cstring>
+#include <string>
 
 #include <v8-platform.h>
 #include <libplatform/libplatform.h>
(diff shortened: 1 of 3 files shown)


m1ga:
Closing this one. Unpatched is fine but it cause all modules to fail and 9.8 ignores our Chrome polyfills for e.g. `Array.at()`. There is a check that will only load `core-js` when the Chrome version is not using them natively. For Chrome 88 (v8 8.8) it's true but starting with 9.2 some stuff should be available and therefore core-js is not loaded. BUT our snapshot builds will strip those out again. So we have to: add custom JS polyfills again, say that core-js is always loaded or make sure the snapshop v8 features are available in v9.8. 
Had it running for hours again without success so I'll close it :) v8 9.8 is more work...

PR #14416, opened 2026-04-01
fix(android): fix endless loop in setLanguage pre API 33

m1ga:
the API 33+ route has a check for setting the same language again:
https://github.com/tidev/titanium-sdk/blob/main/android/titanium/src/java/org/appcelerator/titanium/util/TiLocaleManager.java#L85-L88

pre API 33 doesn't have this and running:
```js
Ti.Locale.setLanguage('en');
```
on an Android 12 phone will end up in an endless restart loop.
With this PR it will max restart once if the locale is different


hansemannn (approved):
Good catch!

PR #14423, opened 2026-04-11
feat(android): add deprecation note for old events in ScrollableView

m1ga:
the `TODO` was added 14 years ago :smile: but there was no warning. 

There are a few of these warnings (search for ` is deprecated. Use`). I think in the next bigger SDK update we should search for these and remove them.


android/modules/ui/src/java/ti/modules/titanium/ui/ScrollableViewProxy.java
```diff
@@ -311,6 +312,7 @@ public void fireDragEnd(int currentPage, TiViewProxy currentView)
 		}
 		// TODO: Deprecate old event
```
hansemannn, line 313: Remove the todo in this case. Below as well

PR #14424, opened 2026-04-11
fix(android): apply dark/light mode switch to experimenta BottomNavigation

m1ga:
**Issue**

When using the new BottomNavigation (`experimental:true`) the content is not updating when changing dark/light mode.

https://github.com/user-attachments/assets/65e42718-887f-4141-9d5e-cc2f56cf690f

**Fix**

* When the app resumes it will check the previous and current UI state and if it changed it will rebuild the content to fix the color scheme.
* small cleanup of variables

**Test**

```js
let tab1 = Ti.UI.createTab({ window: Ti.UI.createWindow({backgroundColor: "backgroundColor",theme:"Theme.Titanium.Material3.DayNight.NoTitleBar"}), title: 'tab1'});
let tab2 = Ti.UI.createTab({window: Ti.UI.createWindow({backgroundColor: "backgroundColor",theme:"Theme.Titanium.Material3.DayNight.NoTitleBar"}), title: 'tab2'});

let bottomNav = Ti.UI.createTabGroup({
  tabs: [tab1, tab2],
	theme: "Theme.Titanium.Material3.DayNight.NoTitleBar",
  experimental: true,
});

bottomNav.open();
```

Note: make sure that `backgroundColor` is present in `semantic.colors.json`

https://github.com/user-attachments/assets/44f3a5e3-b50e-4df9-ac8f-985958d5fdfd


hansemannn (changes requested):
Two issues worth addressing before merge:

### 1. `_lastNightMode` leaks into JavaScript

```java
proxy.setProperty("_lastNightMode", currentNightMode);
```

`KrollProxy.setProperty` makes this visible to JS — `tabGroup._lastNightMode` will be readable/writable from user code and appear in `JSON.stringify`. The leading underscore is convention only; nothing enforces privacy. It also means a per-instance integer pays a HashMap lookup + boxing/unboxing on every resume.

**Suggestion:** store it as a plain field on `TiUIBottomNavigation`. The lifecycle listener is an inner class and has direct access:

```java
private int lastNightMode = -1;
```

…and replace the `proxy.getProperty` / `proxy.setProperty` pair with direct field access.

### 2. `new TiUITab((TabProxy) tab)` in a loop overwrites the proxy's view

`TiUITab`'s constructor calls `proxy.setView(this)` (`TiUITab.java:25`), so each instantiation **replaces** whatever `TiUIView` was previously bound to that `TabProxy`. The priming loop allocates a fresh `TiUITab` for every tab on every qualifying resume, then does it again for the selected tab — discarding the previously bound view each time and risking lost state on it.

You don't need a `TiUITab` here at all — you only want the `TiWindowProxy`. That's already reachable directly:

```java
Object windowProp = ((TabProxy) tab).getProperty(TiC.PROPERTY_WINDOW);
if (windowProp instanceof TiWindowProxy windowProxy) {
    ((TabProxy) tab).setActivity(activity);
    windowProxy.setActivity(activity);
    windowProxy.getOrCreateView();
}
```

Same for the selected-tab rebuild block. This avoids the spurious `setView` overwrites and the allocations.

Nice fix overall — the `onResume` + night-mask diff is a sound approach for this bug.

m1ga:
Good points :+1: Both are changed and it still works fine. Changing between dark/light mode in an open app changes all tabs.

[Bildschirmaufnahme_20260509_113314.webm](https://github.com/user-attachments/assets/6bca6ff3-d28f-4f42-9672-25ab6ba5157a)

(using `adb shell "cmd uimode night yes"` in a console)

hansemannn (approved):
Thank you! Good to go now!

PR #14428, opened 2026-04-15
fix(ios): avoid reusing background completion handlers

### LINKED ISSUE #14418: Random crashes in iOS in TiApp performCompletionHandlerWithKey:andResult:removeAfterExecution:

### Description

I'm getting random crashes when running in iOS 26 due to "**Unbalanced call to dispatch_group_leave()**" in `[TiApp performCompletionHandlerWithKey:andResult:removeAfterExecution:]`
This is the stack:
```
Triggered by Thread: 0, Dispatch Queue: com.apple.main-thread

Exception Type:    EXC_BREAKPOINT (SIGTRAP)
Exception Codes:   0x0000000000000001, 0x00000001d88d3db8

Termination Reason:  Namespace SIGNAL, Code 5, Trace/BPT trap: 5
Terminating Process: exc handler [70621]

Application Specific Information:
BUG IN CLIENT OF LIBDISPATCH: Unbalanced call to dispatch_group_leave()

Thread 0 name:   Dispatch queue: com.apple.main-thread
Thread 0 Crashed:
0   libdispatch.dylib             	       0x1d88d3db8 dispatch_group_leave.cold.1 + 36
1   libdispatch.dylib             	       0x1d88a0c74 dispatch_group_leave + 139
2   TitaniumKit                   	       0x104c07788 -[TiApp performCompletionHandlerWithKey:andResult:removeAfterExecution:] + 80
3   Foundation                    	       0x19ba605c0 __NSFireTimer + 95
4   CoreFoundation                	       0x19e6cd4c0 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 31
5   CoreFoundation                	       0x19e6cd1b8 __CFRunLoopDoTimer + 979
6   CoreFoundation                	       0x19e6ccd2c __CFRunLoopDoTimers + 279
7   CoreFoundation                	       0x19e68e208 __CFRunLoopRun + 1815
8   CoreFoundation                	       0x19e68d1d0 _CFRunLoopRunSpecificWithOptions + 531
9   GraphicsServices              	       0x243bc3498 GSEventRunModal + 119
10  UIKitCore                     	       0x1a43512cc -[UIApplication _run] + 795
11  UIKitCore                     	       0x1a42bc158 UIApplicationMain + 331
12  Spike                         	       0x10406b1ac main + 400
13  dyld                          	       0x19b2a1c1c start + 6927
```
Searching for use of `pendingCompletionHandlers` I can say that ny app receives remote notifications so I suspect it has to do with `handleRemoteNotificationWithIdentifier:andUserInfo:responseInfo:completionHandler:` but I have no idea what the problem could be.
I don't see any reference to `dispatch_group` - only in MediaModule.m in the picker which I don't use.

### Expected Behavior

Should not crash

### Actual behavior

Random crashes while in the foreground.

### Reproducible sample

I have none.

### Steps to reproduce

No steps.

### Platform

iOS

### SDK version you are using

iOS 26

13.1.1.GA

hansemannn:
## Summary
- remove the internal state that allowed a background/silent-push completion handler to be invoked without being removed
- make `performCompletionHandlerWithKey:` one-shot again so a handled silent push cannot be flushed a second time on foreground
- keep the fix scoped to the existing iOS completion-handler lifecycle instead of the newer GCD/scene changes

## Issue
- fixes #14418

### DIFF
--- a/iphone/Classes/TiAppiOSProxy.m
+++ b/iphone/Classes/TiAppiOSProxy.m
@@ -977,7 +977,7 @@
   if ([handlerIdentifier rangeOfString:@"Session"].location != NSNotFound) {
     [[TiApp app] performCompletionHandlerForBackgroundTransferWithKey:handlerIdentifier];
   } else {
-    [[TiApp app] performCompletionHandlerWithKey:handlerIdentifier andResult:UIBackgroundFetchResultNoData removeAfterExecution:NO];
+    [[TiApp app] performCompletionHandlerWithKey:handlerIdentifier andResult:UIBackgroundFetchResultNoData];
   }
 }
 
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
@@ -297,7 +297,7 @@ TI_INLINE void waitForMemoryPanicCleared() // WARNING: This must never be run on
 - (void)registerBackgroundService:(TiProxy *)proxy;
 - (void)unregisterBackgroundService:(TiProxy *)proxy;
 - (void)stopBackgroundService:(TiProxy *)proxy;
-- (void)performCompletionHandlerWithKey:(NSString *)key andResult:(UIBackgroundFetchResult)result removeAfterExecution:(BOOL)removeAfterExecution;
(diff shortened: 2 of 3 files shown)


hansemannn:
Fix verified by the original reporter. Thanks everyone!

PR #14430, opened 2026-04-21
chore(android): replace ti.cloak with ti.crypt

m1ga:
For testing!

* I can build dev and store builds and run them on a phone (Android 13, Android 16). 
* bin files are generated and encrypted
* old ti.cloak so files are gone
* AI build this for me :smile: 
* renamed it to ti.crypt (there is already ti.crypto as a normal module)


cb1kenobi:
I wouldn't delete the old `ti.cloak`. Add a setting to the `tiapp.xml` that indicates which encryption "engine" to use. Some may see the black box as a feature.

m1ga:
good idea, will do that tomorrow

cb1kenobi:
Another idea is to not name your solution "ti.cloak". Then it makes it easy for people to select to encryption system by name.

android/cli/commands/_build.js
```diff
@@ -2889,12 +2889,23 @@ class AndroidBuilder extends Builder {
...
 						await fs.ensureDir(path.dirname(to));
 						this.unmarkBuildDirFile(to);
-						return await cloak.encryptFile(from, to);
+						return await crypt.encryptFile(from, to);
 					})
 				);
 
 				this.logger.info('Writing encryption key...');
-				await cloak.setKey('android', this.abis, path.join(this.buildAppMainDir, 'jniLibs'));
+				await crypt.setKey('android', this.abis, path.join(this.buildAppMainDir, 'jniLibs'));
```
prashantsaini1: This one seems write-only, no read usages now.

m1ga: good catch. I've commented it out for now with a FIXME note. We should use some kind of key in order to make it a little more robust. Just not sure how to would fit into the workflow so it will not be accessible right away in the app later

android/cli/commands/_build.js
```diff
@@ -2863,16 +2863,16 @@ class AndroidBuilder extends Builder {
 			return;
 		}
 
-		// ti.cloak's default export is jacked... there's nested default exports
-		let { default: Cloak } = await import('ti.cloak');
-		while (Cloak && typeof Cloak === 'object') {
-			Cloak = Cloak.default;
+		// ti.crypt's default export is jacked... there's nested default exports
```
cb1kenobi, line 2866: Love the new name!

android/cli/commands/_build.js
```diff
@@ -2863,16 +2863,16 @@ class AndroidBuilder extends Builder {
...
+		let { default: Crypt } = await import('../lib/ti.crypt/index.js');
```
cb1kenobi: Don't use default exports. Just export `Crypt` and then you can get rid of the while loop.

android/cli/lib/ti.crypt/crypt.js
```diff
@@ -0,0 +1,51 @@
+import fs from 'fs/promises';
+import crypto from 'crypto';
+import path from 'path';
+
+export default class Crypt {
```
cb1kenobi: As mentioned above, just export `Crypt`.

```suggestion
export class Crypt {
```

android/cli/lib/ti.crypt/crypt.js
```diff
@@ -0,0 +1,51 @@
+import fs from 'fs/promises';
+import crypto from 'crypto';
+import path from 'path';
```
cb1kenobi: Use the `node:` prefix. This will help out if we ever want to support Bun/Deno.

```suggestion
import fs from 'node:fs/promises';
import crypto from 'node:crypto';
import path from 'node:path';
```

android/cli/lib/ti.crypt/index.js
```diff
@@ -0,0 +1,9 @@
+import Crypt from './crypt.js';
```
cb1kenobi: I don't think you need this barrel file. Can we rename `crypt.js` -> `index.js` instead?

cb1kenobi (commented):
Right now, I feel pretty strongly that we should not remove `ti.cloak` and that we should introduce a new encryption library setting that is set in the `tiapp.xml`.

android/cli/lib/ti.crypt/index.js
m1ga: you are right, moved the file

android/cli/lib/ti.crypt/crypt.js
m1ga: done :+1:

android/cli/commands/_build.js
m1ga: thanks :-)

android/cli/commands/_build.js
m1ga: done :+1:

mbender74:
BTW: I build a toolchain to decompile an Android apk build with Titanium and can restore the full JS source code (from bin.js files). (needed this for a lost source project of me). The only thing not working is rebuild modules (for now)....

how ever, as long the "secret" is stored in the source code, it is a security issue

m1ga:
true, the old pre-ti.cloak encryption even made it to a BlackHat Asia 2015 talk :) 
I'll do some AI tests at the weekend to see if we can come up with something better.

mbender74:
I could provide you my toolchain, but I won´t make it public ;-)

mbender74:
I mailed you something ;-)

m1ga:
@mbender74 thanks, I'll have a look :+1:

cb1kenobi:
I agree. As long as the secret is embedded in the binary, it's not 100% secure. The old Appc verify supported calling a URL on first run to download the secret. I think some people would be interested in this feature. However, storing the secret in the binary can be obfuscated and extremely difficult to find. To this day, I don't think anyone has cracked iOS builds.

mbender74:
@cb1kenobi *cough* ... this not not longer true, about iOS ;-)

mbender74:
we will fix it together! (hopefully)
I think downloading the secret is not a good way, because, would the app then not be able to run in a complete offline mode situation?

cb1kenobi:
I agree, but then we're back to putting the key in the binary somewhere. I asked Claude a while ago how we could hide the key and it basically said no matter what, it can be defeated.

Downloading the key on first start and storing it in the keychain or something is probably the most secure thing we can offer. I think the decision to support offline mode should be up to the developer not us. We simply need to give people options.

m1ga:
yeah, came to the same conclusion. This we should over multiple ways at the end: online key (new), included key (like it is now), fixed key (something like this PR as should be the quickest because not crossing the bridge to decompile the files). I'll close this PR and I think we can continue in this feature request issue: https://github.com/tidev/titanium-sdk/issues/14432

PR #14433, opened 2026-04-25
feat(android): move old ActionBar to Toolbar

m1ga:
"The Toolbar is basically the advanced successor of the ActionBar" introduced in API 21 and since we've switched to minSdk 21 we can just use the Toolbar.

I also fixed the padding issue with the logo `<ActionBar platform="android" logo="appicon.png" />`.

13.2.0.GA:
Screenshot_20260425-123402 (1)

This PR:
Screenshot_20260425-161230

**Testing:**
Use KitchenSink and check the follwing examples:

- ListView Editable (add button, back button)
- ListView Multiselect (Menu, back button)
- Image View Zoom (transparent bar)
- Drawer Layout (Hamburger Menu)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIDrawerLayout.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIDrawerLayout.java
@@ -229,6 +229,7 @@ public class TiUIDrawerLayout extends TiUIView
 			return;
 		}
 		if (activity.getSupportActionBar() != null) {
+			activity.getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 			activity.getSupportActionBar().setHomeButtonEnabled(true);
 		}
 
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
@@ -6,10 +6,13 @@
  */
 package org.appcelerator.titanium.proxy;
 
+import android.content.Context;
 import android.graphics.drawable.Drawable;
+import android.graphics.drawable.InsetDrawable;
+import android.view.View;
 
-import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.Toolbar;
 
 import org.appcelerator.kroll.KrollProxy;
 import org.appcelerator.kroll.annotations.Kroll;
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
 	private static final String TAG = "ActionBarProxy";
 	private static final String ACTION_BAR_NOT_AVAILABLE_MESSAGE = "ActionBar is not enabled";
 
-	private final ActionBar actionBar;
+	private final Toolbar toolbar;
 	private boolean showTitleEnabled = true;
 
 	public ActionBarProxy(AppCompatActivity activity)
 	{
 		super();
-		actionBar = activity.getSupportActionBar();
-		// Guard against calls to ActionBar made before inflating the ActionBarView
-		if (actionBar != null) {
-			actionBar.setDisplayOptions(ActionBar.DISPLAY_USE_LOGO | ActionBar.DISPLAY_SHOW_HOME
-				| ActionBar.DISPLAY_SHOW_TITLE);
-		} else {
-			Log.w(TAG, "Trying to get a reference to ActionBar before its container was inflated.");
+		toolbar = findActionBarToolbar(activity);
+	}
+
+	private Toolbar findActionBarToolbar(AppCompatActivity activity)
+	{
+		if (activity == null) {
+			return null;
 		}
+		int actionBarId = 0;
+		try {
+			actionBarId = org.appcelerator.titanium.util.TiRHelper.getResource("id.action_bar");
+		} catch (Exception e) {
+			Log.d(TAG, "Could not find action_bar resource id");
+		}
+		if (actionBarId != 0) {
+			View view = activity.findViewById(actionBarId);
+			if (view instanceof Toolbar) {
+				return (Toolbar) view;
+			}
+		}
+		return null;
 	}
 
 	@Kroll.setProperty
 	public void setDisplayHomeAsUp(boolean showHomeAsUp)
 	{
-		if (actionBar != null) {
-			actionBar.setDisplayHomeAsUpEnabled(showHomeAsUp);
+		if (toolbar != null) {
+			toolbar.setNavigationIcon(getHomeAsUpIcon(showHomeAsUp));
 		} else {
 			Log.w(TAG, ACTION_BAR_NOT_AVAILABLE_MESSAGE);
 		}
 	}
 
+	private Drawable getHomeAsUpIcon(boolean showHomeAsUp)
+	{
+		if (!showHomeAsUp || toolbar == null) {
+			return null;
+		}
+		Context context = toolbar.getContext();
+		if (context == null) {
+			return null;
+		}
+		return context.getDrawable(androidx.appcompat.R.drawable.abc_ic_ab_back_material);
+	}
+
 	@Kroll.setProperty
 	public void setHomeAsUpIndicator(Object icon)
 	{
-		if (this.actionBar == null) {
+		if (this.toolbar == null) {
 			Log.w(TAG, ACTION_BAR_NOT_AVAILABLE_MESSAGE);
 			return;
 		}
-
 		if (icon instanceof Number) {
-			this.actionBar.setHomeAsUpIndicator(TiConvert.toInt(icon));
+			this.toolbar.setNavigationIcon(TiConvert.toInt(icon));
 		} else if (icon != null) {
-			this.actionBar.setHomeAsUpIndicator(TiUIHelper.getResourceDrawable(icon));
+			this.toolbar.setNavigationIcon(TiUIHelper.getResourceDrawable(icon));
 		} else {
-			this.actionBar.setHomeAsUpIndicator(null);
+			this.toolbar.setNavigationIcon(null);
 		}
 	}
 
 	@Kroll.setProperty
 	public void setHomeButtonEnabled(boolean homeButtonEnabled)
 	{
-		if (actionBar != null) {
-			actionBar.setHomeButtonEnabled(homeButtonEnabled);
+		if (toolbar != null) {
+			toolbar.setNavigationIcon(homeButtonEnabled ? getHomeAsUpIcon(true) : null);
 		} else {
 			Log.w(TAG, ACTION_BAR_NOT_AVAILABLE_MESSAGE);
 		}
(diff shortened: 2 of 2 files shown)


android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
...
+	}
+
+	private Toolbar findActionBarToolbar(AppCompatActivity activity)
+	{
+		if (activity == null) {
+			return null;
+		}
+		int actionBarId = 0;
+		try {
+			actionBarId = org.appcelerator.titanium.util.TiRHelper.getResource("id.action_bar");
```
hansemannn, line 48: Is there really a case where this could be missing? As it's a core resource thats always bundled.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
...
+		} catch (Exception e) {
+			Log.d(TAG, "Could not find action_bar resource id");
+		}
+		if (actionBarId != 0) {
+			View view = activity.findViewById(actionBarId);
```
hansemannn, line 53: Should always be a `Toolbar`.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
...
+		return null;
 	}
 
 	@Kroll.setProperty
 	public void setDisplayHomeAsUp(boolean showHomeAsUp)
 	{
-		if (actionBar != null) {
-			actionBar.setDisplayHomeAsUpEnabled(showHomeAsUp);
+		if (toolbar != null) {
+			toolbar.setNavigationIcon(getHomeAsUpIcon(showHomeAsUp));
```
hansemannn, line 65: **Nav-icon clobbering across many setters.**

`setDisplayHomeAsUp`, `setHomeAsUpIndicator`, `setIcon`, `setHomeButtonEnabled`, `setDisplayShowHomeEnabled`, and `onPropertyChanged(ON_HOME_ICON_ITEM_SELECTED)` all now write to `toolbar.setNavigationIcon(...)`. On `ActionBar` these mapped to independent state (home-as-up flag, home-button-enabled flag, home indicator, app icon). Now they overwrite each other in order-dependent ways, and they will also clobber the hamburger icon installed by `ActionBarDrawerToggle` in `TiUIDrawerLayout`.

Concrete repro of the drawer regression: open a window with a drawer (hamburger appears), then call `actionBar.displayHomeAsUp = true` from JS — the hamburger gets replaced with a back arrow.

Suggest keeping `getSupportActionBar()` for the `displayHomeAs*` family so these stay separable, or routing through `ActionBarDrawerToggle.setDrawerIndicatorEnabled(...)` when a drawer is present.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -112,35 +129,39 @@ public void setBackgroundImage(String url)
...
 		}
 	}
 
 	@Kroll.method
 	public void setDisplayShowTitleEnabled(boolean show)
 	{
-		if (actionBar != null) {
-			actionBar.setDisplayShowTitleEnabled(show);
+		if (toolbar != null) {
+			toolbar.setTitle(show ? toolbar.getTitle() : "");
```
hansemannn, line 145: **Lossy round-trip on `setDisplayShowTitleEnabled`.**

`toolbar.setTitle(show ? toolbar.getTitle() : "")` permanently destroys the title text on `(false)` — there is no separate "title set, but hidden" state on `Toolbar` like there was on `ActionBar.setDisplayShowTitleEnabled`. So the sequence:

```js
actionBar.title = "Inbox";
actionBar.setDisplayShowTitleEnabled(false); // toolbar title now ""
actionBar.setDisplayShowTitleEnabled(true);  // toolbar title still ""
```

silently breaks. Fix: cache the last user-set title in a field and restore it when `show=true`, instead of round-tripping through `toolbar.getTitle()`.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -168,27 +190,20 @@ public void setTitle(String title)
 	@Kroll.getProperty
 	public int getNavigationMode()
 	{
-		if (actionBar == null) {
-			return 0;
-		}
-		return actionBar.getNavigationMode();
+		return 0;
```
hansemannn, line 193: **`navigationMode` becomes a silent no-op; `customView` too (line 287).**

`getNavigationMode` now hard-codes `0`, `setNavigationMode` only logs, and `customView` (further down) just logs `"Use Ti.UI.Toolbar for custom view."` if the value happens to be a `TiViewProxy` (and silently does nothing for any other value). Meanwhile `apidoc/Titanium/Android/ActionBar.yml` still documents `navigationMode` (line ~210) and `customView` (line ~233) as functional properties.

Two ways to land this cleanly:
1. Mark both as `deprecated:` / `osver:` removed in the apidoc, add a `CHANGELOG.md` entry under breaking changes, and keep the warning logs.
2. Or actually re-implement: nav modes (tabs/list) are gone for good reason, but `customView` could be supported by `toolbar.addView(view)` since it’s still common in apps.

As-is, existing apps will appear to work and silently render wrong.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -226,57 +241,52 @@ public void setVisible(boolean value)
...
-			this.actionBar.setLogo(TiUIHelper.getResourceDrawable(image));
+		if (image != null) {
+			Drawable logo;
+			if (image instanceof Number) {
+				logo = TiUIHelper.getResourceDrawable(TiConvert.toInt(image));
+			} else {
+				logo = TiUIHelper.getResourceDrawable(image);
+			}
+			logo.setBounds(0, 0, logo.getIntrinsicWidth(), logo.getIntrinsicHeight());
+			InsetDrawable insetLogo = new InsetDrawable(logo, 24, 0, 14, 0);
```
hansemannn, line 256: **Hard-coded pixel insets, asymmetric, and content-inset side effect leaks.**

A few things stacked here:

1. `new InsetDrawable(logo, 24, 0, 14, 0)` — these are raw pixels. On a 3x density screen (`xxhdpi`) they are visually ~3× smaller than on a 1x screen. Should be dp converted via `TiUIHelper.getRawSize` / `TiConvert.toTiDimension(...).getAsPixels(...)` or similar.
2. The asymmetric `24` left vs `14` right inset has no comment explaining why — looks like it was tuned by eye for one specific test image. Worth a comment, or split into named constants.
3. `this.toolbar.setContentInsetsRelative(0, 0)` permanently zeros the content insets the first time a logo is set, and is **never restored** when `setLogo(null)` is called on the next line of the file. So the title position will stay shifted even after the logo is cleared.
4. If `TiUIHelper.getResourceDrawable(image)` returns `null` (it can — see how `setBackgroundImage` above falls back to `TiDrawableReference`), `logo.setBounds(...)` immediately NPEs on line 255. The original `setIcon` guarded for null; this should too.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -226,57 +241,52 @@ public void setVisible(boolean value)
...
 			Log.w(TAG, ACTION_BAR_NOT_AVAILABLE_MESSAGE);
 			return;
 		}
-
 		if (image instanceof Number) {
-			this.actionBar.setIcon(TiConvert.toInt(image));
+			this.toolbar.setNavigationIcon(TiConvert.toInt(image));
 		} else if (image != null) {
-			this.actionBar.setIcon(TiUIHelper.getResourceDrawable(image));
+			this.toolbar.setNavigationIcon(TiUIHelper.getResourceDrawable(image));
```
hansemannn, line 274: **`setIcon` semantically changed.**

Previously `setIcon` mapped to `actionBar.setIcon(...)` which set the **app icon** displayed alongside the title (DISPLAY_USE_LOGO/DISPLAY_SHOW_HOME). Now it sets `toolbar.setNavigationIcon(...)` — the back/up/hamburger slot. That makes `setIcon` and `setHomeAsUpIndicator` and `setLogo`-via-nav-slot all do the same thing, and breaks any existing app that relied on having both an app icon and a back arrow.

The closest behavioral match for the old API on `Toolbar` is `setLogo(...)` (display next to title). Suggest routing `setIcon(image)` to `toolbar.setLogo(...)` (or simply documenting it as an alias of `logo`), and reserving the navigation-icon slot for `displayHomeAsUp` / `homeAsUpIndicator`.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
...
+
+	private Toolbar findActionBarToolbar(AppCompatActivity activity)
+	{
+		if (activity == null) {
+			return null;
+		}
+		int actionBarId = 0;
+		try {
+			actionBarId = org.appcelerator.titanium.util.TiRHelper.getResource("id.action_bar");
+		} catch (Exception e) {
```
hansemannn, line 49: Nit: `catch (Exception e)` is too broad. `TiRHelper.getResource` only throws `TiRHelper.ResourceNotFoundException`; catching the narrower type avoids swallowing programmer errors (e.g. an NPE if the helper itself regresses). Also, dropping the missing-resource branch to `Log.d` is fine for the happy path, but in practice when this fires, `toolbar` ends up `null` and every subsequent setter logs `ACTION_BAR_NOT_AVAILABLE_MESSAGE` — worth at least a `Log.w` here so the cause shows up once, loudly, instead of as a flood of downstream warnings.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
```diff
@@ -26,54 +29,78 @@ public class ActionBarProxy extends KrollProxy
...
+	private Drawable getHomeAsUpIcon(boolean showHomeAsUp)
+	{
+		if (!showHomeAsUp || toolbar == null) {
+			return null;
+		}
+		Context context = toolbar.getContext();
+		if (context == null) {
+			return null;
+		}
+		return context.getDrawable(androidx.appcompat.R.drawable.abc_ic_ab_back_material);
```
hansemannn, line 80: Pulling `androidx.appcompat.R.drawable.abc_ic_ab_back_material` directly is a use of an internal-ish AppCompat resource. It works today, but it has tripped people up across AppCompat upgrades and it ignores the theme’s `homeAsUpIndicator` attribute (so a user theme that customizes the back arrow will be ignored when this code path runs).

A more robust pattern: resolve the `R.attr.homeAsUpIndicator` attribute from the toolbar context theme and use that drawable, falling back to `abc_ic_ab_back_material` only if unresolved. That way custom themes keep working.

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIDrawerLayout.java
```diff
@@ -229,6 +229,7 @@ private void initDrawerToggle()
 			return;
 		}
 		if (activity.getSupportActionBar() != null) {
+			activity.getSupportActionBar().setDisplayHomeAsUpEnabled(true);
```
hansemannn, line 232: Cross-file FYI: this is fine on its own, but combined with the `ActionBarProxy` rewrite, the back-arrow set here on the support action bar lives at the same nav-icon slot the new proxy code clobbers. If a JS app calls e.g. `actionBar.displayHomeAsUp = true` (or `actionBar.homeAsUpIndicator = ...`) at any point after the drawer opens/initializes, the hamburger drawn by `ActionBarDrawerToggle` will be replaced with a back arrow. Worth verifying explicitly with the "Drawer Layout" KitchenSink example after the proxy is opened from JS.

hansemannn (changes requested):

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
m1ga: `getResource` does throw a exception so I've put it into a try/catch. I'll reduce some lines in the next commit and import TiRHelper.

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
m1ga: They were deprecated in Android SDK 21 "Action bar navigation modes are deprecated and not supported by inline toolbar action bars. Consider using other common navigation patterns instead.". Have to look up what it actually does and how to replace it. Sounds like a bigger todo

android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
m1ga: correct, I was just eyeballed but the current default looked worse for many years now  :smile:

m1ga:
going through the other comments: think I'll close this one as it turns out to be much more work then the current PR is changing at the moment  :smile: 
Guess it will be quicker with some AI work, I just don't have any pro models. Not a big AI user myself and I only use opencode-ai with the free models and they don't do a good job here. Since it's not of any priority we can wait for someone else to pick it up and do the change.

PR #14439, opened 2026-05-07
feat(ios): remove ImageView remote image fade-in animation

m1ga:
Currently the iOS imageView will fade in a remote image with `defaultImage` for 0.5 seconds but removes the default image right away so it will have a blank screen in-between. This PR will remove that animation and just show the image.

**Test**

```js
	width: 200,
	height: 200,
	image: "/images/DefaultIcon.png",
	defaultImage: "/images/DefaultIcon.png",
	borderWidth: 1,
	borderColor: "red",
})
win.add(img);
setTimeout(function() {
	img.image = "https://picsum.photos/seed/"+Math.floor(Math.random()*1000)+"/2000"
}, 2000);
win.open();
```

### DIFF
--- a/iphone/Classes/TiUIImageView.m
+++ b/iphone/Classes/TiUIImageView.m
@@ -326,8 +326,9 @@ DEFINE_EXCEPTIONS
 
     // do a nice fade in animation to replace the new incoming image
     // with our placeholder
+    CGFloat duration = fadeInTimeSet ? fadeInTime : 0.5;
     [UIView beginAnimations:nil context:nil];
-    [UIView setAnimationDuration:0.5];
+    [UIView setAnimationDuration:duration];
     [UIView setAnimationDelegate:self];
     [UIView setAnimationDidStopSelector:@selector(animationCompleted:finished:context:)];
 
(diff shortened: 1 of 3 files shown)


iphone/Classes/TiUIImageView.m
```diff
@@ -326,8 +326,9 @@ - (void)setURLImageOnUIThread:(UIImage *)image
 
     // do a nice fade in animation to replace the new incoming image
     // with our placeholder
+    CGFloat duration = fadeInTimeSet ? fadeInTime : 0.5;
     [UIView beginAnimations:nil context:nil];
-    [UIView setAnimationDuration:0.5];
```
hansemannn, line 330: I would actually rather say we should remove the animation altogether - makes it much more clean. And iOS should offer something out of the box if it's really a common thing to do. WDYT?

hansemannn (changes requested):

iphone/Classes/TiUIImageView.m
m1ga: yeah, that sounds good too. Until know I thought that the fade-in was something iOS related but didn't know Ti was doing that 😄 Since the image is fully loaded at that point we can just show it. 

Will update the PR in a bit

m1ga: PR title and code are updated. Much leaner now

hansemannn (approved):
Nice!

PR #14443, opened 2026-05-12
perf(ios): incremental-build optimizations

hansemannn:
## Summary

Five targeted, low-risk perf changes inside `iphone/cli/commands/_build.js`. Each is a separate commit so they can be reverted or cherry-picked independently. Net diff: +141 / −64 in one file. No public API or hook contract changes.

Came out of a deep audit of the iOS build pipeline. A1/A2 (the bigger ones — `processJSFiles`/`gatherResources` worker-pool and wave merging) are intentionally **not** in this PR; they need more measurement and touch shared task infrastructure.

## Changes

| Commit | What |
|---|---|
| 528ce4e | `copyTitaniumiOSFiles` `beforeCopy` hook: if previous-build fingerprint (size+mtime) matches and dest still exists, reuse the cached entry and skip the source read, dest read, and hash. Skipped for ejs-rendered `appFiles` and once Frameworks has bulk-copied this build. |
| 733f164 | `createXcodeProject`: cache the parsed template `project.pbxproj` under `build/incremental/xcode-project/template-parsed.json`, keyed by sha1 of the source. The template only changes when the SDK changes, so subsequent builds skip `xcodeParser.parse()`. |
| c93cc7f | `createAppIconSet`: resolves the in-file `// FIXME: Do these in parallel`. Per-icon `fs.readFile` + `appc.image.pngInfo` now run via `Promise.all`. Validation/categorization stays serial over the resolved buffers (it mutates shared state). |
| f67db67 | `createAssetImageSets`: was awaiting one `writeAssetContentsFile` per directory level per image (the same namespace marker written many times). Dedupe parent directories in a `Set`, then write namespace markers + imageset manifests in one `Promise.all`. |
| c6b2bba | `processLaunchLogos`: hoist `LaunchLogo.png` stat+read+hash and the per-missing-logo `existsSync` probes into a single `Promise.all`. Same pattern applied to `writeAppIconSet`'s missing-icon `existsSync`+`readFileSync`+`pngInfo` loop. |

## Test plan

- [ ] `npm run test:iphone` (simulator) — clean build then incremental build, verify no regressions in resource copying, asset catalog, icon generation
- [ ] Manual: project with a custom `LaunchLogo.png` — verify the parallelized stat/read still yields correct missingLaunchLogos set
- [ ] Manual: project with a default icon that has alpha — flatten path in `createAppIconSet` still triggers
- [ ] Manual: incremental rebuild with no source changes — confirm `forceRebuild` stays false and xcodebuild is skipped
- [ ] Inspect `build/incremental/xcode-project/` after a build — both `template-src.sha1` and `template-parsed.json` present, second build logs "Reusing cached parsed pbxproj"

### DIFF
--- a/iphone/cli/commands/_build.js
+++ b/iphone/cli/commands/_build.js
@@ -5003,15 +5003,34 @@ class iOSBuilder extends Builder {
 						return null;
 					}
 
-					let contents = fs.readFileSync(srcFile),
-						changed = false;
 					const rel = srcFile.replace(path.dirname(this.titaniumSdkPath) + '/', ''),
-						destExists = fs.existsSync(destFile),
-						existingContent = destExists && fs.readFileSync(destFile),
-						srcHash = this.hash(contents),
-						srcMtime = JSON.parse(JSON.stringify(srcStat.mtime));
+						srcMtime = JSON.parse(JSON.stringify(srcStat.mtime)),
+						prev = this.previousBuildManifest.files && this.previousBuildManifest.files[rel],
+						destExists = fs.existsSync(destFile);
 
 					this.unmarkBuildDirFile(destFile);
+
+					// Fast path: if stat (size+mtime) matches the previous build and dest still exists,
+					// the source can't have changed, so reuse the cached fingerprint and skip read/hash/write.
+					// Skipped for appFiles (rendered through ejs each build) and once a Frameworks bulk
+					// copy has already run this build (the dest may have been replaced with fresh content).
+					if (
+						!appFiles[filename]
+						&& destExists
+						&& !nameChanged
+						&& prev
+						&& prev.size === srcStat.size
+						&& prev.mtime === srcMtime
+						&& !(dir === 'Frameworks' && !copyFrameworks)
+					) {
+						this.currentBuildManifest.files[rel] = prev;
+						return null;
+					}
+
+					let contents = fs.readFileSync(srcFile),
+						changed = false;
+					const existingContent = destExists && fs.readFileSync(destFile),
+						srcHash = this.hash(contents);
 
 					this.currentBuildManifest.files[rel] = {
 						hash: srcHash,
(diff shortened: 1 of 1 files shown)


hansemannn:
@cb1kenobi WDYT?

m1ga:
Did I understand it correctly that it should cache the icons and not regenerate them on each build? If so: at least the log still says it will regenerate all icons, even though I didn't change anything.

cb1kenobi:
@hansemannn gah, my bad, I'll take a peek this weekend.

hansemannn:
@cb1kenobi Quick ping

m1ga:
@hansemannn I let Claude check the changes and 4 of 5 look correct :+1: The one issue it found was:

---

The pbxproj cache gets deleted at the end of (almost) every build — commit 733f16422c, iphone/cli/commands/_build.js:3333. The cache files are written to buildDir/incremental/xcode-project/, but the base builder snapshots the entire build directory into buildDirFiles at build start (node-titanium-sdk/lib/builder.js:161), and removeFiles() at iphone/cli/commands/_build.js:7156 deletes every snapshotted file that was never unmarked. The existing incremental tasks survive this only because they explicitly opt out — see cli/lib/tasks/copy-resources-task.js:51: "don't let iOS build delete our incremental state files!" followed by unmarkBuildDirFiles(this.incrementalDirectory). Your new cache never unmarks itself, so: build N writes the cache (survives, since it wasn't in the snapshot), build N+1 hits the cache but then deletes it, build N+2 misses again. The cache only works on alternating builds. 
The fix is one line — unconditionally call this.unmarkBuildDirFiles(pbxprojCacheDir) right after the cache read/write block.

---

maybe you could check that.

Otherwise building iOS apps still works fine :+1:

hansemannn:
@m1ga Let me know if we're good now.

m1ga:
PR looks fine now :+1:  I'm currently running tests to check the improvements between 13.3.1 and this one but nothing that should stop this from getting merged. I'm just interested to see the changes :smile:

PR #14449, opened 2026-05-22
fix(iOS): textarea fix and feature enhancement

mbender74:
## Summary

  This PR implements several improvements to `Ti.UI.TextArea` on iOS to bring it closer to Android parity and fix longstanding layout issues.

  ### New Properties (iOS parity with Android)

  - **`scrollable`** (Boolean, default: `true`) — Controls whether the textarea can be scrolled. When `false`, the content is always visible and scrolling is disabled.
  - **`lines`** (Number, default: `1`) — Sets the initial height of the textarea by line count. A value of `-1` lets the textarea grow based on content.
  - **`maxLines`** (Number, default: `-1`) — Caps the maximum number of visible lines. Once reached, the textarea stops growing and scrolling is enabled. A value of `-1` means no limit.

  ### Bug Fixes

  - **Vertical alignment** — Fixed `verticalAlign` property (`top`, `center`, `bottom`) which was previously ignored by `UITextView`. Implemented using dynamic `textContainerInset` adjustment.
  - **Cursor visibility** — Fixed cursor displacement when typing. The cursor now stays visible on the correct line, especially when `maxLines` is set.
  - **Content shrinking** — Fixed textarea not shrinking when lines are deleted. The layout engine now correctly measures content height using base padding only.
  - **`maxLines` height calculation** — Fixed the max height to include text container insets (padding) and border, ensuring all `maxLines` lines are fully visible without scrolling.
  - **`lines` height calculation** — Fixed the initial height to include padding and border, so `lines:2` correctly produces a 2-line tall textarea.

  ### Technical Details

  - Vertical alignment is simulated by dynamically adjusting `textContainerInset` based on content height vs. bounds height.
  - During active editing, alignment is deferred to avoid cursor displacement. Alignment is applied after editing ends via `textViewDidEndEditing`.
  - The `scrollViewDidScroll` delegate now clamps the content offset to valid bounds instead of resetting to zero, preventing cursor displacement when scrolling is disabled.
  - Content height measurement uses `boundingRectWithSize:` during editing for accurate results, since `UITextView.contentSize` is often stale during active text input.

  ### Files Changed

  - `iphone/Classes/TiUITextArea.m` — Core implementation
  - `iphone/Classes/TiUITextArea.h` — Header with new instance variables and methods
  - `iphone/Classes/TiUITextAreaProxy.m` — Property definitions
  - `apidoc/Titanium/UI/TextArea.yml` — API documentation

```
var textArea = Ti.UI.createTextArea({
    top:8,
    bottom:8,
    left:15,
    right:8,
    autocorrect: false,
    editable:true,
    lines:1,
    maxLines:5,
    borderWidth: 1,
    borderColor: '#aaa',
    borderRadius: 16,
    color: '#000',
    backgroundColor: '#fff',
    font: {fontSize:16, fontWeight:'normal'},
    textAlign: 'left',
    value: '',
    width: '70%',
    padding:{left:4,right:4,top:8,bottom:8},
    height : Ti.UI.SIZE,
    suppressReturn:false
  });
```

  ### before
before

### after
after

### DIFF
--- a/apidoc/Titanium/UI/TextArea.yml
+++ b/apidoc/Titanium/UI/TextArea.yml
@@ -520,7 +520,7 @@ properties:
     summary: Vertical alignment within this text area.
     type: [Number, String]
     constants: Titanium.UI.TEXT_VERTICAL_ALIGNMENT_*
-    default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER>
+    default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_TOP>
     platforms: [android, iphone, ipad, macos]
 
 events:
--- a/apidoc/Titanium/UI/TextArea.yml
+++ b/apidoc/Titanium/UI/TextArea.yml
@@ -522,6 +522,7 @@ properties:
     constants: Titanium.UI.TEXT_VERTICAL_ALIGNMENT_*
     default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_TOP>
     platforms: [android, iphone, ipad, macos]
+    since: { android: "0.8", iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
 
 events:
   - name: blur
--- a/apidoc/Titanium/UI/TextArea.yml
+++ b/apidoc/Titanium/UI/TextArea.yml
@@ -413,8 +413,8 @@ properties:
 
   - name: lines
     summary: Number of lines tall the text area height will be, if set.
-    description: Sets the height of the text area by line height. Set to 0 to let the text area grow based on content.
-    default: 0
+    description: Sets the height of the text area by line height. Set to -1 to respect the view's height property instead.
+    default: 1
     type: Number
     since: { android: "7.5.0", iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
     platforms: [android, iphone, ipad, macos]
@@ -431,8 +431,8 @@ properties:
 
   - name: maxLines
     summary: Maximum line count of text field input.
-    description: Sets the maximum number of lines that the field will extend to when pressing `Return`. Once the maximum is reached, the text area stops growing and scrolling is enabled. A value of 0 means no limit.
-    default: 0
+    description: Sets the maximum number of lines that the field will extend to when pressing `Return`. Once the maximum is reached, the text area stops growing and scrolling is enabled. A value of `-1` means no limit.
+    default: -1
     type: Number
     since: { android: "7.5.0", iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
     platforms: [android, iphone, ipad, macos]
--- a/apidoc/Titanium/UI/TextArea.yml
+++ b/apidoc/Titanium/UI/TextArea.yml
@@ -413,11 +413,11 @@ properties:
 
   - name: lines
     summary: Number of lines tall the text area height will be, if set.
-    description: Sets the height of the text area by line height. Set to -1 to respect the view's height property instead.
-    default: 1
+    description: Sets the height of the text area by line height. Set to 0 to let the text area grow based on content.
+    default: 0
     type: Number
-    since: { android: "7.5.0" }
-    platforms: [android]
+    since: { android: "7.5.0", iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
+    platforms: [android, iphone, ipad, macos]
 
   - name: maxLength
     summary: Maximum length of text field input.
@@ -431,11 +431,11 @@ properties:
 
   - name: maxLines
     summary: Maximum line count of text field input.
-    description: Sets the maximum of lines that the field will be extended to when pressing `Return`.
-    default: -1
+    description: Sets the maximum number of lines that the field will extend to when pressing `Return`. Once the maximum is reached, the text area stops growing and scrolling is enabled. A value of 0 means no limit.
+    default: 0
     type: Number
-    since: { android: "7.5.0" }
-    platforms: [android]
+    since: { android: "7.5.0", iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
+    platforms: [android, iphone, ipad, macos]
 
   - name: padding
     summary: Sets the left and right padding of this TextArea. The text will always be vertically centered.
@@ -500,6 +500,7 @@ properties:
     summary: Determines whether this text area can be scrolled.
     type: Boolean
     default: true
+    since: { iphone: "13.3.0", ipad: "13.3.0", macos: "13.3.0" }
     platforms: [iphone, ipad, macos]
 
   - name: selection
@@ -519,8 +520,8 @@ properties:
     summary: Vertical alignment within this text area.
     type: [Number, String]
     constants: Titanium.UI.TEXT_VERTICAL_ALIGNMENT_*
-    default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_TOP>
-    platforms: [android]
+    default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER>
+    platforms: [android, iphone, ipad, macos]
 
 events:
   - name: blur
--- a/iphone/Classes/TiUITextArea.m
+++ b/iphone/Classes/TiUITextArea.m
@@ -110,6 +110,39 @@
 {
   [[self textWidgetView] sizeToFit];
   [super frameSizeChanged:frame bounds:bounds];
+  [self updateVerticalAlignment];
+}
+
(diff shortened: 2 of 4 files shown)


designbymind:
@mbender74 I've tested this with a patched version of TiSDK 13.2.0.GA and it's working great on my end!

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -413,11 +413,11 @@ properties:
 
   - name: lines
     summary: Number of lines tall the text area height will be, if set.
-    description: Sets the height of the text area by line height. Set to -1 to respect the view's height property instead.
-    default: 1
```
hansemannn, line 417: If the default value changed, we need to wait for a major release. Is there a way around it or do you prefer to wait for the major bump (likely August/September with iOS 27)

apidoc/Titanium/UI/TextArea.yml
```diff
@@ -520,7 +521,7 @@ properties:
     type: [Number, String]
     constants: Titanium.UI.TEXT_VERTICAL_ALIGNMENT_*
     default: <Titanium.UI.TEXT_VERTICAL_ALIGNMENT_TOP>
-    platforms: [android]
+    platforms: [android, iphone, ipad, macos]
```
hansemannn, line 524: Was this always available?

iphone/Classes/TiUITextArea.m
```diff
@@ -110,6 +110,39 @@ - (void)frameSizeChanged:(CGRect)frame bounds:(CGRect)bounds
...
   [super frameSizeChanged:frame bounds:bounds];
+  [self updateVerticalAlignment];
+}
+
+- (void)updateVerticalAlignment
+{
+  UITextView *tv = (UITextView *)[self textWidgetView];
+
+  id basePadding = [[self proxy] valueForUndefinedKey:@"padding"];
+  UIEdgeInsets baseInset = basePadding ? [TiUtils contentInsets:basePadding] : UIEdgeInsetsZero;
```
hansemannn: `contentInsets` should do the nil handling already

iphone/Classes/TiUITextArea.m
```diff
@@ -110,6 +110,39 @@ - (void)frameSizeChanged:(CGRect)frame bounds:(CGRect)bounds
...
+
+  // Always reset to base insets first so we start from a clean state
+  tv.textContainerInset = baseInset;
+  [tv layoutIfNeeded];
+
+  // Calculate the actual content height using base padding
+  CGFloat contentHeight = tv.contentSize.height;
+  CGFloat boundsHeight = self.bounds.size.height - 2 * self.layer.borderWidth;
```
hansemannn, line 129: Why `-2`?

iphone/Classes/TiUITextArea.m
```diff
@@ -130,6 +163,32 @@ - (void)frameSizeChanged:(CGRect)frame bounds:(CGRect)bounds
...
+    verticalAlign = UIControlContentVerticalAlignmentTop;
+    linesCount = 0;
+    maxRows = -1;
+
+    id padding = [[self proxy] valueForUndefinedKey:@"padding"];
+    if (padding) {
+      [self setPadding_:padding];
+    }
+
+    id va = [[self proxy] valueForUndefinedKey:@"verticalAlign"];
```
hansemannn: proper property names

iphone/Classes/TiUITextArea.m
```diff
@@ -130,6 +163,32 @@ - (void)frameSizeChanged:(CGRect)frame bounds:(CGRect)bounds
...
+
+    id linesProp = [[self proxy] valueForUndefinedKey:@"lines"];
+    if (linesProp) {
+      [self setLines_:linesProp];
+    }
+
+    id maxLinesProp = [[self proxy] valueForUndefinedKey:@"maxLines"];
+    if (maxLinesProp) {
+      [self setMaxLines_:maxLinesProp];
+    }
```
hansemannn, line 191: Proper types and property names

iphone/Classes/TiUITextArea.m
```diff
@@ -224,6 +283,14 @@ - (BOOL)checkLinkAttributeForString:(NSAttributedString *)theString atPoint:(CGP
 }
 #pragma mark Public APIs
 
+- (void)setValue_:(id)value
+{
+  [super setValue_:value];
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [self updateVerticalAlignment];
+  });
+}
```
hansemannn: TiThreadPerformOnMainThread (with wait=NO) should do the same

iphone/Classes/TiUITextArea.m
```diff
@@ -255,6 +322,26 @@ - (void)setScrollable_:(id)value
...
+- (void)setLines_:(id)value
+{
+  linesCount = [TiUtils intValue:value def:0];
+  [[self proxy] replaceValue:NUMINTEGER(linesCount) forKey:@"lines" notification:NO];
+  if (linesCount > 0) {
+    UITextView *tv = (UITextView *)[self textWidgetView];
+    CGFloat lineHeight = tv.font.lineHeight;
+    UIEdgeInsets padding = tv.textContainerInset;
+    CGRect frame = self.bounds;
+    frame.size.height = lineHeight * linesCount + padding.top + padding.bottom + 2 * self.layer.borderWidth;
```
hansemannn, line 336: The +2 again

hansemannn (changes requested):

apidoc/Titanium/UI/TextArea.yml
mbender74: nope, that was Android only ...
verticalAlign-API-Docs

iphone/Classes/TiUITextArea.m
mbender74: that is because else the calculation of the verticalAlignment is not correct if there are borders
- (2 * self.layer.borderWidth)  --> bracket not needed for correct math ;-)
the bound.size.height does not respect that

iphone/Classes/TiUITextArea.m
mbender74: I did change this, you are right

iphone/Classes/TiUITextArea.m
mbender74: also changed

iphone/Classes/TiUITextArea.m
mbender74: changed

iphone/Classes/TiUITextArea.m
mbender74: same here, the  + (2 * self.layer.borderWidth) is needed for correct sizing the TiUITextarea

iphone/Classes/TiUITextArea.m
mbender74: sure, this could be done more elegant, but I would keep this for now

apidoc/Titanium/UI/TextArea.yml
mbender74: I changed this to the correct behaviour again

iphone/Classes/TiUITextArea.m
mbender74: Here a more detailed explanation:

    The UITextView is a child view of the TiUITextArea. When the TextArea has a borderWidth, that border takes up space inside its own bounds:

     - Top edge: borderWidth pixels
     - Bottom edge: borderWidth pixels

    This means the available area for text is smaller than the full height of the view.

    Example:
     - self.bounds.size.height = 100
     - borderWidth = 2
     - Available height for text: 100 - 2*2 = 96

    Without this correction, you'd incorrectly assume the text has 100 pixels of space when it actually only has 96. This would cause vertical alignment to be calculated wrong —
    especially when checking if content fits (contentHeight < boundsHeight) and how much extra padding to add for Center/Bottom alignment.

apidoc/Titanium/UI/TextArea.yml
hansemannn: We should add a `since` tag in that case

iphone/Classes/TiUITextArea.m
hansemannn: Why that? you are doing logic thats 1:1 in `contentInsets:` already, incl. nil handling, so you can just do

```objc
  UIEdgeInsets baseInset = [TiUtils contentInsets:basePadding];
```

apidoc/Titanium/UI/TextArea.yml
mbender74: you are right, is added

iphone/Classes/TiUITextArea.m
mbender74: ok, is changed

hansemannn (approved):

PR #14450, opened 2026-05-22
fix(ios): tabgroup navigationwindow fix

mbender74:
Fix for https://github.com/tidev/titanium-sdk/issues/14021

## Summary

  Fixes `UIViewControllerHierarchyInconsistency` crash when a `Ti.UI.TabGroup` is wrapped in a `Ti.UI.NavigationWindow`.

  ### Before
  When a TabGroup is used inside a NavigationWindow, the app terminates immediately with:
  child view controller: should have parent view controller:
  but actual parent is:

  ### Root Cause
  `TiUITabGroup.open:` was using `TiApp.controller.topPresentedController` to find the parent view controller for `addChildViewController:`. This always returns `TiRootViewController`, which is correct for
  standalone TabGroups but wrong when the TabGroup is nested inside a NavigationWindow.

  The fix uses `[self.proxy windowHoldingController]` which correctly returns the `TiViewController` that wraps the TabGroup's window, regardless of whether it's standalone or nested.

  ### Changes

  - **`TiUITabGroup.open:`** — Use `windowHoldingController` instead of `topPresentedController` to get the correct parent `TiViewController`
  - **`TiUITabGroupProxy.viewWillAppear:`** — Add `parentViewController == nil` guard to prevent double `addChildViewController:` when the tabController was already added in `open:`

  ### Files Changed

  - `iphone/Classes/TiUITabGroup.m` — Fix parent controller selection in `open:`
  - `iphone/Classes/TiUITabGroupProxy.m` — Add parent guard in `viewWillAppear:`

### DIFF
--- a/iphone/Classes/TiUITabGroup.m
+++ b/iphone/Classes/TiUITabGroup.m
@@ -704,7 +704,7 @@ DEFINE_EXCEPTIONS
   TiThreadPerformOnMainThread(
       ^{
         isTabBarHidden = NO;
-        UIViewController *parentController = TiApp.controller.topPresentedController;
+        UIViewController *parentController = [self.proxy windowHoldingController];
         [parentController addChildViewController:self.tabController];
         [self addSubview:self.tabController.view];
         self.tabController.view.frame = self.bounds;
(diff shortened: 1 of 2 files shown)


m1ga:
Tested it with the demo code in the issue and it works fine. No crash when opening the tabgroup :+1:

hansemannn (approved):

PR #14456, opened 2026-06-03
fix(android): improve Window focus/blur events

m1ga:
Will fire blur/focus event when you drag down the statusbar on your phone.

**Note:** This will also fire when an alert is shown! That is currently not the case. So I'm not sure if we should use a different event for this as it may cause problems in existing apps.

```js
	title: "alert",
})

win.add(btn);
win.open();

win.addEventListener("focus", function(){
	console.log("win focus")
})
win.addEventListener("blur", function(){
	console.log("win blur")
})

btn.addEventListener("click", function(e) {
	alert(1);
})
```

I could change it to `windowblur` and `windowfocus` so you have a different event for these new blur/focus and it won't break any exisint `blur`/`focus` usage on a window

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -1456,6 +1456,9 @@ public abstract class TiBaseActivity extends AppCompatActivity implements TiActi
 				}
 			}
 		}
+		if (this.window != null) {
+			this.window.onWindowFocusChange(hasFocus);
+		}
 		super.onWindowFocusChanged(hasFocus);
 	}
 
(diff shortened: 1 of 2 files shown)


hansemannn:
How does iOS behave here, especially for alerts? IIRC, it also fires blur/focus events prior/after showing alerts.

m1ga:
need to check that on iOS. 

Since an alert would fire a blur event on Android after that change it might be better to give it a different event name in case iOS doesn't do that. Otherwise it might fire `blur` in cases you don't want it because it is just showing an alert.  But I'll test it on iOS next week and check how the events line up.

m1ga:
running the example code with the current main (not using this PR) on iOS and Android: 
* iOS:  blur/focus is fired already when opening/closing the alert
* Android: no events when opening/closing the alert

so this PR will actually make Android behave the same way as iOS is doing right now

PR #14459, opened 2026-06-06
feat(android): optimize Ti.Preferences

m1ga:
Optimizing the Ti.Preferences code a bit not to rely on try/catch.

**getPreference(String key)** Two issues here:
- systemProperties.get(key) throws JSONException for every non-tiapp.xml key
- preferences.getAll().get(key) loads the entire preference map into memory on every call, even when the key doesn't exist

**Fix:** Use has() instead of try/catch. Guard getAll() with contains().

**Test code**
```js
Ti.App.Properties.setString('givenName', 'Paul');
console.log('The value of the givenName property is: ' + Ti.App.Properties.getString('givenName'));

Ti.App.Properties.setBool('boolValue', true);
console.log('boolValue', Ti.App.Properties.getBool('boolValue', false));

console.log('boolValueDefault1', Ti.App.Properties.getBool('boolValueDefault1', false));
console.log('boolValueDefault2', Ti.App.Properties.getBool('boolValueDefault2'));

// get all
var props = Ti.App.Properties.listProperties();
for (var i = 0, ilen = props.length; i < ilen; i++) {
	var value = Ti.App.Properties.getString(props[i]);
	console.log(props[i] + ' = ' + value);
}
```

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiProperties.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiProperties.java
@@ -64,18 +64,13 @@ public class TiProperties
 
 	public Object getPreference(String key)
 	{
-		Object value = null;
-		if (systemProperties != null) {
-			try {
-				value = systemProperties.get(key);
-			} catch (JSONException e) {
-				value = preferences.getAll().get(key);
-			}
+		if (systemProperties != null && systemProperties.has(key)) {
+			return systemProperties.opt(key);
 		}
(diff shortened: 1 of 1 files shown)


hansemannn (approved):
Reviewed the current PR diff (TiProperties.java only). I do not see a blocking regression: system properties still take precedence, preference fallback remains for missing or mismatched typed values, and git diff --check is clean. I did not run the Android test suite.

PR #14462, opened 2026-06-12
fix: fix module packaging process

prashantsaini1:
After facing this build error randomly on my machine, I have had Codex to find the root cause and its fix.

**Problem:** SDK build process exits randomly at "package modules step" without any clear error. e.g.

```
....
Writing manifest.json
Including pre-packaged native modules...
Copying /var/folders/4x/p2x6d02x32v06h28tzl__6th0000gn/T/timob-build/ti.applesignin-iphone-3.1.2 to /Users/p.saini/Projects/titanium-sdk/dist/mobilesdk-13.3.0-osx
```

Sometimes here
```
Copying /var/folders/4x/p2x6d02x32v06h28tzl__6th0000gn/T/timob-build/com.appcelerator.urlSession-iphone-4.0.1 to /Users/p.saini/Projects/titanium-sdk/dist/mobilesdk-13.3.0-osx
```

**Codex's RCA:**
```
The break was in the packaging phase, not Xcode.

`npm run ios` was exiting during packaged module extraction inside `build/lib/utils.js (line 297)`. 

I reproduced it directly: the build stopped in `cacheUnzip()` while re-extracting cached module ZIPs such as `com.appcelerator.urlSession-iphone-4.0.1.zip`. 

The current node-appc unzip path never returned for those module ZIPs, so Node exited with a half-built dist/mobilesdk-13.3.0-osx tree and never reached zipPlatforms(), zip(), or install().
```

### DIFF
--- a/build/lib/utils.js
+++ b/build/lib/utils.js
@@ -12,7 +12,6 @@ import { fileURLToPath } from 'node:url';
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const titanium = path.resolve(__dirname, '..', '..', 'node_modules', 'titanium', 'bin', 'ti.js');
 const exec = util.promisify(childProcess.exec);
-const execFile = util.promisify(childProcess.execFile);
 
 const tempDir = os.tmpdir();
 
@@ -298,22 +297,12 @@ export async function cacheExtract(inFile, integrity, outDir, extractFunc) {
 */
 export function unzip(zipfile, dest) {
 	return fs.ensureDir(dest).then(async () => {
-		if (os.platform() === 'win32') {
-			await execFile('powershell.exe', [
-				'-NoProfile',
-				'-NonInteractive',
-				'-Command',
-				'Expand-Archive',
-				'-LiteralPath',
-				zipfile,
-				'-DestinationPath',
-				dest,
-				'-Force'
-			]);
+		if (os.platform() !== 'win32') {
+			await exec('unzip -q -o "' + zipfile + '" -d "' + dest + '"');
 			return;
 		}
 
-		await execFile('unzip', [ '-q', '-o', zipfile, '-d', dest ]);
+		await util.promisify(appc.zip.unzip)(zipfile, dest, null);
 	});
 }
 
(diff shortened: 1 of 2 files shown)


build/lib/utils.js
```diff
@@ -296,7 +297,24 @@ export async function cacheExtract(inFile, integrity, outDir, extractFunc) {
...
+				'-Command',
+				'Expand-Archive',
+				'-LiteralPath',
+				zipfile,
+				'-DestinationPath',
+				dest,
+				'-Force'
+			]);
+			return;
+		}
```
hansemannn: Is this needed? The Promise.all -> serial fix makes sense, but I don't see a specific change here that could affect your original issue

hansemannn (changes requested):

build/lib/utils.js
prashantsaini1: It's usually required since we also support building SDK on Windows OS. Though I cannot verify it at present, but Codex strongly suggests it despite instructing if it can be removed. I could resolve the issue without this `win32` related code on my Mac.

m1ga:
couldn't test the windows part but I had the issue that the generated iOS folder didn't contain all SDK files and the iOS fix here fixed it.

cb1kenobi:
I really don't like spawning powershell to unzip a file. I have reworked this code in node-titanium-sdk, so this PR is moot.

prashantsaini1:
@cb1kenobi Is the fix in `node-titanium-sdk` already live now? As long as we have a fix for this issue, it's fine to go with that route.

cb1kenobi:
@prashantsaini1 the fix is in a draft PR that isn't complete yet. Once the PR is ready, I will still have to do a handful of PRs against the SDK before we can introduce the new node-titanium-sdk version.

hansemannn:
@cb1kenobi @prashantsaini1 Can we revert the windows change and still ship it under that condition? I do see a lot of value in the other fixex of this PR.

prashantsaini1:
@hansemannn I have reverted the Windows-related changes for now, this still fixes the core issue. We can continue using this PR changes until the PR from @cb1kenobi is ready.

cb1kenobi (approved):

hansemannn (approved):

PR #14466, opened 2026-06-19
chore: github workflow

m1ga:
Fix for other workflow warnings:
```
AdoptOpenJDK has moved to Eclipse Temurin

Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: hendrikmuhs/ccache-action@v1.2
```

and upgrading to java 21 + macos-25

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -28,7 +28,7 @@ runs:
     - name: Setup Java
       uses: actions/setup-java@v5
       with:
-        distribution: 'adopt'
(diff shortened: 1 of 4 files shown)


m1ga:
it's still not producing the artifacts :shrug: the intermediate files are there but it won't copy everything into the zip files later on

cb1kenobi (approved):

m1ga:
also update the macos version to 25

cb1kenobi (approved):
Looks good, merge away!

PR #14469, opened 2026-06-25
feat(chore): new vitepress documentation

m1ga:
A first test to move to a new API documentation as the current API docs use a very old vitepress version and is spread out into multiple repositories to build the docs.

This PR will just add the vitepress to the SDK with the base layout. 

**Note:** This is only for the API docs, not the guides! So we could keep the guides as they are now (or include them here) and use this for the API reference. The modules are not included (yet).

Startpage:
Bildschirmfoto_20260626_000609

Api docs:
Bildschirmfoto_20260626_000617

search:
Bildschirmfoto_20260626_000631

Ti.UI.Button:
Bildschirmfoto_20260626_000641

**Testing:** 
* sync files: `npm i && npm run docs:vitepress:sync`
* build: `npm i && npm run docs:vitepress:build`
* dev server: `npm i && npm run docs:vitepress:dev`

**Note:**
I'm using vitepress 2 alpha now as that is in alpha for a very long time now and I don't want to start with 1 for a new project :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,5 +67,7 @@ junit_report.xml
 /vitepress/docs/api/
 /vitepress/docs/.vitepress/sidebar.mjs
 /vitepress/docs/.vitepress/dist/
+/vitepress/docs/.vitepress/.temp/
 /vitepress/docs/.vitepress/cache/
 /vitepress/docs/public/images/
+/vitepress/modules-cache/
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,11 @@ junit_report.xml
 
 # VS Code History Extension
 .history
+
+# VitePress docs
+/vitepress/node_modules/
+/vitepress/docs/api/
+/vitepress/docs/.vitepress/sidebar.mjs
+/vitepress/docs/.vitepress/dist/
+/vitepress/docs/.vitepress/cache/
+/vitepress/docs/public/images/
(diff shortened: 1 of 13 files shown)


cb1kenobi:
This is AWESOME!

m1ga:
Thanks! At the moment I'm playing around with it to see how far I can get this. E.g. external modules are included now too:

Bildschirmfoto_20260626_103858

m1ga:
Using a minimal theme to reduce the render parts in the converter script now. Also makes it easier to apply styling.

Examples:
Bildschirmfoto_20260626_112715

inline XML:
Bildschirmfoto_20260626_112820

cb1kenobi:
After some thinking, it feels like this should go in a new repo. I think we should decouple the SDK from the website.

Docs for released SDKs and modules are essentially static and I was thinking they could be baked and stored in a repo, then the website just need to pull that repo instead of the giant SDK repo. I'm happy to help create sites, repos, devops stuff, whatever. I just need a task list and I'll rock them out. :)

This work is so good and gets us so close to a new/better website, can we just make this the new website? Can we put all of the docs in `/docs` path, then make `/` being the titaniumsdk.com homepage? The homepage does not need to be super amazing. We don't need hero images and flashy transitions. Add a global header component with the links to the downloads, socials, github, etc and I think we're basically done.

m1ga:
I still think the actual page should be a bit more that shows people some interesting stuff like users and images of features (made this 2 years ago, posted it on slack multiple times: https://www.figma.com/design/eh6dVru2pixoOsY2D529YG/Untitled?node-id=0-1&t=cJzccxkmmbaEftTE-1 ). As with that page people decide if they want to use it or not and this is our competitor https://flutter.dev/
I know people don't like this, but it could easily be a WordPress page and adding a new interview/user story would be a simple form you fill out and it will be on the page (no compiling or deployment :smile: ) and the API pages can be vitepress (or something else) that is generated automatically.

Happy to move it to the titanium-docs repo as a PR :+1: Just wanted to start somewhere as I have many issues with the current page (not always searching when you search before it's done loading, some API pages are broken, ...) :smile:

m1ga:
moving it to titanium-docs: https://github.com/tidev/titanium-docs/pull/261

PR #14472, opened 2026-06-28
fix(ios): open the tab's root-window when the tab becomes active

prashantsaini1:
This PR brings iOS `TabGroup` behavior in line with Android by lazily opening a `Tab`’s root `Ti.UI.Window` only when that tab becomes active.

**Issue:** Previously on iOS, all tab root windows were opened during `TabGroup` initialisation, causing the root `Ti.UI.Window` opening for inactive tabs as well, and thus also firing their `open` event.

```
// iOS: both `win1` and `win2` open at same time, and so `onOpenWindowA` and `onOpenWindowB` are also fired immediately.
// Android: `win2` opens only when the `tab2` is selected.

<Alloy>
    <TabGroup backgroundColor="white" >
        <Tab id="tab1" title="Tab 1" icon="KS_nav_views.png">
            <Window id="win1" title="Tab 1" onOpen="onOpenWindowA">
                <Label>I am Window 1</Label>
            </Window>
        </Tab>
        <Tab id="tab2" title="Tab 2" icon="KS_nav_views.png">
            <Window id="win2" title="Tab 2" onOpen="onOpenWindowB">
                <Label>I am Window 2</Label>
            </Window>
        </Tab>
    </TabGroup>
</Alloy>
```


prashantsaini1:
@m1ga @cb1kenobi Isn't it surprising that this issue either went unnoticed for so long, or was intended?

I tested this fix on our internal apps, even opening some non root windows on inactive tabs. The user experience was natural and matched the Android behaviour.

m1ga:
can you provide an example so we can test this please. 
Not sure I understand it correctly: will it now open the window when you click the tab or only fire the open event later? I think for tabgroups most people use `selected` or `focus` to know when the window got active instead of the open event.

prashantsaini1:
@m1ga I just updated the description to be more clear. Currently, all windows in a `TabGroup` are opened immediately on iOS platform. This is the main issue here, as a Window is expected to only open when its host Tab becomes active. The code sample is straight forward, just notice that all Tabs window are opened at once.

m1ga:
Thanks :+1: I've used the example code from the docs and added an open event:
```js
var win1 = Ti.UI.createWindow({});
win1.addEventListener("open", () => {
	console.log("win1 open")
})
var win2 = Ti.UI.createWindow({});
win2.addEventListener("open", () => {
	console.log("win2 open")
})
var win3 = Ti.UI.createWindow({});
win3.addEventListener("open", () => {
	console.log("win3 open")
})

var tab1 = Ti.UI.createTab({
		window: win1,
    title: 'w1'
	}),
	tab2 = Ti.UI.createTab({
		window: win2,
    title: 'w2'
	}),
	tab3 = Ti.UI.createTab({
		window: win3,
    title: 'w3'
	}),
    tabs: [tab1, tab2, tab3],
	});
tabGroup.open();
```
so it's a bit quicker to test :wink: 

Works the same on both platforms now. Should this be considered a breaking change as it might affect e.g. API calls that you do in your third window in the open event (some people only have iOS apps so they don't know that Android is not doing that)?

BTW: wanted to install the generated artifact on my Mac and `ti sdk install file.zip` is broken :shrug: Does this work for you? Had to manually install it to test it. But it looks like a titanium-cli issue, just wanted to check if that works for you with this file. edit: [filed an issue here](https://github.com/tidev/titanium-sdk/issues/14539)

prashantsaini1:
@m1ga Even though this looks like a fix to me but as apps can be running in multiple scenarios and may have adjusted their code already to what they expect. I think 13.3.0.GA is still a good candidate, otherwise 14.x is next to take it.

Regarding the ti sdk install, I can check it later today.

mbender74:
I would prefer that the lazy window opening for tabsgroups should be an option not the default.
I do a lot in already opened windows on inactive tabs that will not be possible that way.

On Android I workaround that with special hacks, that I got multiple windows open, even when they are not active... but from my opinion maybe the android side could get simultanous opening too, instead deactivate it on iOS!?

Already having the window opened (hidden) and the complex UI in it rendered completely before showing the window is a huge win for some use cases....

prashantsaini1:
@mbender74 Thank you for sharing your use cases.

1. This could be included in the 14.x SDK release, as it would obviously affect the behaviour of existing apps.
2. Opening hidden windows when they are not visible is not good practice as it causes unnecessary delays to app launch times. Apps should avoid any practices that cause them to underperform.

I checked a few apps, such as the App Store, iTunes Store, Apple Music, WhatsApp, Google Drive, Slack and MS Teams, they all load windows lazily when a tab is selected.

I'd be happy to hear more opinions on this!

mbender74:
@prashantsaini1 

`opening hidden windows when they are not visible is not good practice as it causes unnecessary delays to app launch times. Apps should avoid any practices that cause them to underperform.`

if this is a good practice is the clearly a decision of the developer of the app. taking developers a choice or forcing them to a specific design pattern is not a good practice!

There are design patterns of keeping or open windows / views with full UI that are beneficial. Only having the choice for lazy opening restricts developers.

Therefore I agree that the lazy window opening should be there (which was a good finding of you), but as an option, not the only way.

mbender74:
@prashantsaini1 so please let us keep the old behaviour and the new one.
A property for lazyInit or something else would be nice, when not set or false, the current behaviour is maintained.
Else all developers need to check their code for the breaking change.
What do you mean? 😊

m1ga:
I would also prefer this as a property which makes it a none-breaking change and can be integrate earlier. Default as it currently is. And we could think about a way to do this on Android too (e.g. in the experimental BottomNavigation). 

If I remember correctly we already have some places where a default property is different on Android/iOS if you don't set it.  So doing nothing: same way as always. Setting the "preload" property to `true` will keep the current way on iOS and Android will start preloading windows, setting it to `false` will use the new way on iOS and keep the current way on Android. For now we would just gate your changes with a property and could add that feature to 13.4.0 (if we have that version)

prashantsaini1:
I understand how existing apps are designed. However, it's also undeniable that existing apps are not utilising the TabGroup's performance capabilities on iOS as they should be. Tab-based navigation is always intended to avoid loading things until they are shown to the user. Therefore, the Android behaviour is already optimal as it should be.

As there is no hurry to implement this significant change at this point, in order to maintain the current functionality of existing apps, I propose this:

1. The reason not to make it based on a property is that it would divert us from the optimal approach. The goal should be to make Titanium apps performant out of the box. 🙂 
2. We'll take it in SDK 14.x, as this will require a behavioural change. Existing apps will have enough time to make the necessary changes.
3. We can still keep the property, but it should be enabled by default to match Android behaviour. It has to be set to `false` for apps using the iOS-specific workaround.

hansemannn:
We have to have a unified experience on iOS and Android - that's what breaking changes are for. I am open to decide whether we take the Android behavior to iOS or the other way around, but the current state is not clean, so we should unify it.

mbender74:
> We have to have a unified experience on iOS and Android - that's what breaking changes are for. I am open to decide whether we take the Android behavior to iOS or the other way around, but the current state is not clean, so we should unify it.

I fully agree that as far as possible we always should get parity for features.
My point is that when we take a from my opinion nice feature, instead support it on Android too, the SDK will loose, also without the proposed optional property, we force maybe a lot developers to refactor their apps and this is work and costs money and raise a lot of frustration... I can say this from an enterprise company view and budged enough is sometimes not existing for reworking an app that actually just needs to be compiled with the newest SDK version to fit Google or Apples requirements. Then to find out why the app is not working as before is not nice... ;-)

Here are two screencasts that demonstrating what I mean.
The difference is clear, imagine you will prepare UI or prefill a listView/tableView already before the user opens the tab so that you do not see any delay in UI rendering.... many possibilities...

@prashantsaini1 is right that this is a difference between iOS and Android and the change should be that we support both variants on both OS versions (lazy and non lazy window opening for tabgroup)

The discussion what is the best design pattern is nonsense, because there are pros and cons for each variant and the startup time of the app is not notable affected if you do not do unnecessary stuff in the window open event, for this we got the window focus event... I´m pretty sure we all want the same, performance and outstanding possibilities. This is why I'm committed to Titanium since version 0.8

**iOS:**
iOS

**Android:**
Android


```
       backgroundColor: 'red',
       layout: 'vertical'
});
    text: 'counter',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
var label1a = Ti.UI.createLabel({
    text: 'counter focus',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
win1.add(label1);
win1.add(label1a);
win1.addEventListener('open', function() {
    console.log('win1 open');
    var count = 0;
    var timer1 = setInterval(function() {
        count++;
        label1.text = 'counter: ' + count;
    }, 1000);
});
var win1FocusInterval = false;
win1.addEventListener('focus', function() {
    console.log('win1 focus');
    if (!win1FocusInterval) {
        win1FocusInterval = true;
        var count = 0;
        var timer = setInterval(function() {
            count++;
            label1a.text = 'counter focus: ' + count;
        }, 1000);
    }
});


    backgroundColor: 'blue',
    layout: 'vertical'
});
var label2 = Ti.UI.createLabel({
    text: 'counter',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
var label2a = Ti.UI.createLabel({
    text: 'counter focus',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
win2.add(label2);
win2.add(label2a);
win2.addEventListener('open', function() {
    console.log('win2 open');
    var count = 0;
    var timer2 = setInterval(function() {
        count++;
        label2.text = 'counter: ' + count;
    }, 1000);
});
var win2FocusInterval = false;
win2.addEventListener('focus', function() {
    console.log('win2 focus');
    if (!win2FocusInterval) {
        win2FocusInterval = true;
        var count = 0;
        var timer = setInterval(function() {
            count++;
            label2a.text = 'counter focus: ' + count;
        }, 1000);
    }
});



var win3 = Ti.UI.createWindow({
    backgroundColor: 'green',
    layout: 'vertical'
});
var label3 = Ti.UI.createLabel({
    text: 'counter',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
var label3a = Ti.UI.createLabel({
    text: 'counter focus',
    color: '#fff',
    font: { fontSize: 24, fontWeight: 'bold' },
    textAlign: 'center',
    width: Ti.UI.FILL,
    height: Ti.UI.SIZE
});
win3.add(label3);
win3.add(label3a);
win3.addEventListener('open', function() {
    console.log('win3 open');
    var count = 0;
    var timer = setInterval(function() {
        count++;
        label3.text = 'counter: ' + count;
    }, 1000);
});
var win3FocusInterval = false;
win3.addEventListener('focus', function() {
    console.log('win3 focus');
    if (!win3FocusInterval) {
        win3FocusInterval = true;
        var count = 0;
        var timer = setInterval(function() {
            count++;
            label3a.text = 'counter focus: ' + count;
        }, 1000);
    }
});


var tab1 = Ti.UI.createTab({
        window: win1,
    title: 'w1'
    }),
    tab2 = Ti.UI.createTab({
        window: win2,
    title: 'w2'
    }),
    tab3 = Ti.UI.createTab({
        window: win3,
    title: 'w3'
    }),
    tabs: [tab1, tab2, tab3],
    });
tabGroup.open();
```

mbender74:
BTW: we got on iOS a background color problem on _tabGroup_ as you see in the example above

m1ga:
> We have to have a unified experience on iOS and Android - that's what breaking changes are for. I am open to decide whether we take the Android behavior to iOS or the other way around, but the current state is not clean, so we should unify it.

Which is fine and since we have the option with this PR to have it the same (optimized) way as Android has we can do this route. We just should keep the option to have the current way as apps used it for 10+ years like this and as Marc is saying: you have to rewrite a lot if you rely on this (e.g. an API call in a 2nd window that is now not called for people that only use iOS and didn't know that Android does it other ways).

And until this PR I didn't even know that iOS loads all content right away as it didn't have any big performance issue as far as I can tell :smile: 

**As mentioned before I would implement it like this:**
- add this feature but guard it with a property
- if the property is unset we keep it the same way as it is now
- maybe: add a note saying "Titanium 15: TabGroup Windows will only open when you click on them unless you set property xyz to false to keep it like it is now" (something like this)
- Titanium 15: set the unset value to `true`. 

That way nothing will change for a long time, users _can_ set it to `true` if they want the optimized behavior and Android parity. And that way they will know about the change in Ti 15 so they can set it to `false` during that 2 years (or so).

Then everyone is happy! You can have the optimized experience already (setting the property:true) or set it to `false` and never have to worry about it. If you don't do anything you will see the warning until Ti 15 and if you didn't change it during that time it will automatically set it to `true` at that point.

> BTW: we got on iOS a background color problem on tabGroup as you see in the example above

@mbender74 I know I've posted an issue about that already but I can't find it :smile: Think Hendrik worked on that too.

mbender74:
In a lot of my apps I used for Android a custom tabBar, that opens the windows without showing them also at app start time to match the iOS behaviour.  (because I like to prepare UI in the spare time the app does nothing without polluting the UI thread and get that way a very snappy UI experience and no visible delays, but this is just my taste of doing it)

The proposal of @m1ga looks to me a good way to go, so we could have the optimisation of @prashantsaini1 
with a wider time range for deprecation.

prashantsaini1:
So let's do it this way considering all inputs on this thread:
- Introduce a new iOS only property on `TabGroup lazyLoadingEnabled` (similarly on [ListView here](https://titaniumsdk.com/api/titanium/ui/listview.html#properties_lazyloadingenabled)).
- This property will be set to `true` by default in SDK 13.3.0.GA to keep the unified behaviour.
- It has to be set to `false` for apps intentionally loading windows without being visible. 

It would be a minuscule change for such apps (as with @mbender74 ) with no breaking changes at all.

m1ga:
I would only swap the defaults for a none-breaking update and it could make it into 13.3.0.

Then the property is there for people to use it and we can change the default after some time (e.g. Ti 14/15) with more room for people to adjust the setting or change their code.

prashantsaini1:
Done. 
- Added a new property `lazyLoadingEnabled` on `TabGroup`.
- Defaults to `false` and shows a log until SDK `14.0.0.GA`. 
- Tested in both scenarios, no changes required on existing apps unless manually set to `true`.

iphone/Classes/TiUITabGroupProxy.m
```diff
@@ -155,6 +161,9 @@ - (void)showTabBar:(id)args
 
 - (void)windowWillOpen
 {
+  if (![self lazyLoadingEnabled]) {
+    DebugLog(@"[WARN] Ti.UI.TabGroup.lazyLoadingEnabled is currently false by default on iOS, but will default to true starting with SDK 14.0.0.GA. Set lazyLoadingEnabled explicitly to preserve the current behavior.");
```
m1ga:
```suggestion
    DebugLog(@"[WARN] Ti.UI.TabGroup.lazyLoadingEnabled is currently false by default on iOS, but will default to true starting with SDK 14.0.0.GA. Set lazyLoadingEnabled explicitly to false to preserve the current behavior.");
```

m1ga:
nice @prashantsaini1 will test it later but sounds great :+1:

m1ga:
Test app works fine: default (not setting anything) shows the same open events as before. Setting it to `false` does the same. Setting it to `true` only showed the open event when clicking on a new tab.  :heavy_check_mark: 

Tested it with an existing app and all windows load correctly :heavy_check_mark: 

But for this app I think I prefer the `false` setting a bit more. There is a webview with a chart in page 3 and it was always loaded on iOS already and not it will show the preloader (same as Android). That makes it feel a bit slower on iOS now. But it's good we have that setting now so the dev can decide!

PR #14474, opened 2026-06-28
fix(android): fix padding unit in ScrollableView

m1ga:
This one is in the SDK for too long :-)
The left/right padding of a ScrollableView is applied not in the correct unit:

13.2.0:
Screenshot_20260628-215249
red box and padding should be `40`

This PR will apply the normal TiDimension conversion:
Screenshot_20260628-215221
red box and padding are `40`

**Example:**
```js

var view1 = Ti.UI.createView({ backgroundColor:'#123' });
var view2 = Ti.UI.createView({ backgroundColor:'#246' });
var view3 = Ti.UI.createView({ backgroundColor:'#48b' });
var view4 = Ti.UI.createView({ backgroundColor:'red', width: 40,height: 40 });
view1.add(view4);

// Common params
var params = {
  views:[view1,view2,view3],
  clipViews:false,
  showPagingControl:false,
	top: 0,
	padding: {
		left:40,
		right:40,
  }
};

var scrollableView = Ti.UI.createScrollableView(params);
win.add(scrollableView);
win.open();
```

**Note:**
* iOS uses `width`, so it just affects Android ([apidocs](https://titaniumsdk.com/api/titanium/ui/scrollableview.html#scrollable-view-with-multiple-visible-views))
* will be a breaking change as you have to remove your own dpi calculation if you use that

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollableView.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIScrollableView.java
@@ -17,6 +17,7 @@ import org.appcelerator.titanium.proxy.TiViewProxy;
 import org.appcelerator.titanium.util.TiConvert;
 import org.appcelerator.titanium.view.TiCompositeLayout;
 import org.appcelerator.titanium.view.TiUIView;
+import org.appcelerator.titanium.TiDimension;
 
 import ti.modules.titanium.ui.ScrollableViewProxy;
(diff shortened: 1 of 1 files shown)


prashantsaini1:
Yeah, has been too long in there. Is it required at any other places too just to be sure before merging?

m1ga:
for the ScrollableView it should just be this place. 

Wasn't there even a script in the SDK that found places with missing TiDimensions :thinking: Guess there might be one or two other places that we've missed

PR #14477, opened 2026-06-30
fix(android): validate sessionid in Chrome debug session

m1ga:
Fixing two FIXME parts in the code

* FIXME Use this UUID to store sessions and enforce it's used when a client is connecting
* FIXME: If we're not on API level 23+, shouldn't we reject/error?

The second one is just removed as we are minSDK 24 now!

**Testing:**
* build an app with `ti build -p android  -T device --debug-host localhost:8989`
* use the chrome link and change the ID value at the end

### DIFF
--- a/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
+++ b/android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
@@ -35,6 +35,9 @@ public final class JSDebugger
 	// Are we ready to continue? Has the debugger been connected and we've processed the first set of messages?
 	private final AtomicBoolean ready = new AtomicBoolean(false);
 
+	// The session UUID generated for this debugger instance, validated when a client connects
+	private String sessionId;
+
 	// Holding place for messages received from V8 intended for debugger
 	private final LinkedBlockingQueue<String> v8Messages = new LinkedBlockingQueue<>();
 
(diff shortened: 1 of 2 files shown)


android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
```diff
@@ -163,6 +165,12 @@ public InspectorAgent(int port) throws UnknownHostException
 		@Override
 		public void onOpen(WebSocket conn, ClientHandshake handshake)
 		{
+			String resourceDescriptor = conn.getResourceDescriptor();
+			if (resourceDescriptor == null || !resourceDescriptor.endsWith("/" + sessionId)) {
```
cb1kenobi, line 169: Does this need to be `this.sessionId`?

cb1kenobi (approved):

android/runtime/v8/src/java/org/appcelerator/kroll/runtime/v8/JSDebugger.java
m1ga: not in this case as it is inside `private class InspectorAgent extends WebSocketServer` and wouldn't find the outer class variable.

cb1kenobi: Gotcha, cool, merge away!

PR #14481, opened 2026-07-02
fix(ios): fix liveview restart

### LINKED ISSUE #14480: iOS: Liveview not working in 13.3.0

### Description

Liveview doesn't work on iOS with 13.3.0.GA. It will stop with
```
[INFO]  [LiveView] Reloading App
[INFO]  UI SHUTDOWN COMPLETE. TRYING TO RESUME RESTART
```

It looks like it is not working correctly with the MultiScene support

### Expected Behavior

reload the app

### Actual behavior

stops loading the app after saving

### Reproducible sample

```js

var view = Ti.UI.createView({
	width: 200,
	height: 60,
	backgroundColor: "white",
	borderColor: "red",
	borderWidth: 10,
	borderRadius: 20,
	top: 100
});
win.add(view);
win.open();
```

### Steps to reproduce

* run the app with liveview
* change a value and save the file

### Platform

iOS

### SDK version you are using

13.3.0.GA

_No response_

### AI analysis

Problem Analysis

The LiveView restart flow is broken because _resumeRestart: in AppModule.m calls:

[appDelegate application:app didFinishLaunchingWithOptions:[appDelegate launchOptions]];

Prior to the multi-scene support, didFinishLaunchingWithOptions: was responsible for:
 
1. Creating the UIWindow
2. Calling initController (creates TiRootViewController, sets root VC)
3. Calling boot (creates fresh KrollBridge, loads JS)
 
With the scene delegate architecture added, those three responsibilities were moved to scene:willConnectToSession:. Now didFinishLaunchingWithOptions: at TiApp.m:450-452 explicitly says:

// Do not boot here. With the scene lifecycle, booting happens in
// scene:willConnectToSession: — which is forwarded to this same (app-delegate)
// instance — so the window can be created against the connecting UIWindowScene.

So after _resumeRestart: tears down the UI and bridge, calling didFinishLaunchingWithOptions: does nothing to rebuild them — the app ends up with no window, no controller, and no JS bridge.

m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/14480

### Problem Analysis

The LiveView restart flow is broken because _resumeRestart: in AppModule.m calls:

[appDelegate application:app didFinishLaunchingWithOptions:[appDelegate launchOptions]];

Prior to the multi-scene support, didFinishLaunchingWithOptions: was responsible for:

1. Creating the UIWindow
2. Calling initController (creates TiRootViewController, sets root VC)
3. Calling boot (creates fresh KrollBridge, loads JS)

With the scene delegate architecture added, those three responsibilities were moved to scene:willConnectToSession:. Now didFinishLaunchingWithOptions: at TiApp.m:450-452 explicitly says:
```
// Do not boot here. With the scene lifecycle, booting happens in
// scene:willConnectToSession: — which is forwarded to this same (app-delegate)
// instance — so the window can be created against the connecting UIWindowScene.
```
So after _resumeRestart: tears down the UI and bridge, calling didFinishLaunchingWithOptions: does nothing to rebuild them — the app ends up with no window, no controller, and no JS bridge.

**Note:**
only tested it with 
```js

var view = Ti.UI.createView({
	width: 200,
	height: 60,
	backgroundColor: "white",
	borderColor: "red",
	borderWidth: 10,
	borderRadius: 20,
	top: 100
});
win.add(view);
win.open();
```
and liveview and it's fully AI generated, so I don't know if it is correct :smile: But liveview restart is working again. That's why it's draft only in case there is a better way to do it.

### DIFF
--- a/iphone/Classes/AppModule.m
+++ b/iphone/Classes/AppModule.m
@@ -74,8 +74,10 @@ extern NSString *const TI_APPLICATION_GUID;
   NSNotification *notification = [NSNotification notificationWithName:kTiContextShutdownNotification object:[appDelegate krollBridge]];
   [nc postNotification:notification];
 
+  /* Reboot via scene-aware method (creates new window, controller, and bridge) */
+  [appDelegate rebootApp];
+
   /* Begin foregrounding simulation */
-  [appDelegate application:app didFinishLaunchingWithOptions:[appDelegate launchOptions]];
   [appDelegate applicationWillEnterForeground:app];
   [appDelegate applicationDidBecomeActive:app];
   /* End foregrounding simulation */
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.h
@@ -294,4 +294,13 @@
 - (void)performCompletionHandlerForBackgroundTransferWithKey:(NSString *)key;
 - (void)watchKitExtensionRequestHandler:(id)key withUserInfo:(NSDictionary *)userInfo;
 
+/**
+ Re-initializes the UI and JS runtime against the existing UIWindowScene.
+
+ Used by LiveView (<Ti.App._restart>) to perform a hot restart of the application
+ without leaving the scene session. Tears down the old window, controller, and
+ KrollBridge, then creates fresh ones against the first connected UIWindowScene.
+ */
+- (void)rebootApp;
+
 @end
--- a/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/API/TiApp.m
@@ -1252,6 +1252,35 @@ extern void UIColorFlushCache(void);
   return kjsBridge;
 }
 
+- (void)rebootApp
+{
+  UIWindowScene *windowScene = nil;
+  for (UIScene *scene in [UIApplication sharedApplication].connectedScenes) {
+    if ([scene isKindOfClass:[UIWindowScene class]]) {
+      windowScene = (UIWindowScene *)scene;
+      break;
+    }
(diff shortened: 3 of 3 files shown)


hansemannn:
Why not change the existing `_restart` method? That's also used in LiveViw right now.

m1ga:
That I can't tell you :smile: the error `UI SHUTDOWN COMPLETE. TRYING TO RESUME RESTART` is from this place:
https://github.com/tidev/titanium-sdk/blob/main/iphone/TitaniumKit/TitaniumKit/Sources/API/TiRootViewController.m#L728-L734
where it is calling `_resumeRestart`.

That's the only hint I had and fed that into the AI

hansemannn:
If that works fine, I'm okay with the change. Is this ready for review then?

m1ga:
Liveview is working (starting up and restarting when I change values) and without liveview the app is starting too.

hansemannn (approved):
Excellent!

prashantsaini1:
@m1ga It's still broken. I could reproduce the issue on our apps since we already use `restartApp` method via [this module](https://github.com/rborn/tiapprestart), with exactly same implementation available in [iOS SDK Ti.App class](https://github.com/tidev/titanium-sdk/blob/6e6bf97c933cddec06971f1f60e947a695d63360/iphone/Classes/AppModule.m#L35) as well (wrapped for debug/developer mode, no idea why, but it's safe to enable it as a public API). Since this looks like a regression to me, I would try the suggestion from @hansemannn as it would fix both issues without requiring a new `reboot` method. I hope this is fine.

m1ga:
@prashantsaini1 feel free to generate a PR to fix that issue. As mentioned in the first post: I have no clue about this and this PR was 100% AI generated by giving it the regression info from the Slack user when Liveview was broken. I'm really not the best person to fix iOS stuff :smile:

PR #14483, opened 2026-07-02
chore: improve test suite

m1ga:
* move the timeouts to global constants
* commenting out `require('./ti.locale.test');` for now as it will restart the app on Android and stopping the test
* fix webview redirect links

Android should be something like this `4945 Total Tests: 4897 passed, 48 failed, 0 skipped.` at the end.

Use squash and merge :+1:


m1ga:
can be closed in favor of https://github.com/tidev/titanium-sdk/pull/14487 once that is ready for review

PR #14487, opened 2026-07-06
fix(all): fixes Android and iOS issues and tests

mbender74:
# Fix iOS 26 simulator and Android API 36 test failures

  ## Summary

**The work from @m1ga is strongly used from this PR: https://github.com/tidev/titanium-sdk/pull/14483**

  This branch resolves a large set of test-suite failures observed on **iOS 26 simulators** and **Android API 36 (Android 16)** devices, plus several long-standing
  flaky tests and test-infrastructure issues. The work spans native iOS, native Android, shared JavaScript, and the test harness itself.

  **Scope:** 66 commits · 145 files changed · +2,012 / −913 lines

  ---

  ## Android fixes

  ### `TiCompositeLayout` — TIMOB-23372 #4/#5 (`label.rect.y = -10`)
  - **Root cause:** For single-child horizontal layouts, `updateRowForHorizontalWrap()` was never called because of an `if (count > 1)` guard.
  `horizontalLayoutLineHeight` stayed at `0`, so `computePosition()` centered the child in a zero-height area and produced `label.rect.y = -measuredHeight/2` (e.g.
  `-10`).
  - **Fix:** Always call `updateRowForHorizontalWrap()` for the first child, even when there is only one child. Also moved the `horizontalLayoutLineHeight` tracking
   **before** the `rowWidth > maxRight` early return, so a child wider than the available space still contributes its height to the row.

  ### `TiContentFile` — `android.resource://` lookup on API 36
  - **Root cause:** On Android 16, `ContentResolver.openInputStream()` / `openAssetFileDescriptor()` / `openFileDescriptor()` can all fail for `android.resource://`
   URIs even when the resource exists, so `exists()` returned `false`.
  - **Fix:** Added a `Resources.getIdentifier()` fallback in `exists()` for `ContentResolver.SCHEME_ANDROID_RESOURCE` URIs. The fallback parses the URI path as
  `/<type>/<name>`, looks up the resource ID, and returns `true` if non-zero.
  - Also added a null-check on `getNativeFile()` before calling `.exists()`.

  ### `appicon` drawable resource
  - The test `android.resource://${Ti.App.id}/drawable/appicon` requires a real Android drawable named `appicon`. The template `appicon.png` only lived in
  `assets/Resources/`, not in `res/drawable/`, so `Resources.getIdentifier()` returned `0`. Added `tests/platform/android/res/drawable/appicon.png` so the build
  includes it as a proper Android resource.

  ### Other Android
  - `TiProperties`: hardened property parsing.
  - `IntentProxy`, `TiMimeTypeHelper`, `TiCameraXActivity`, `TiUIVideoView`, `TabGroupProxy`, `WindowProxy`: minor hardening / cleanup.
  - `JSDebugger`, `TiFileProvider`: dead-code cleanup.

  ---

  ## iOS fixes

  ### Clipboard (`Ti.UI.Clipboard`)
  - **Thread safety:** All mutators (`setText`, `clear`, etc.), `has*` probes, and named/unique pasteboard creation are now dispatched to the main thread.
  - **API surface:** `text` is now exposed as both a property and via `getText()` / `setText()` methods.
  - **UTI handling:** Plain-text UTIs are now handled explicitly in `mimeTypeToDataType()`.
  - **`hasText`:** Reads `UIPasteboard.string` rather than relying solely on `count`.

  ### `overrideUserInterfaceStyle`
  - The setter and getter now operate on the **same `UIWindow`** (the getter was reading from a different window than the setter wrote to, causing style reads to
  always return the default).
  - Style is now applied to **both** the window and its `rootViewController` so it propagates to presented view controllers.

  ### Safe area & NavigationWindow
  - The home-indicator bottom inset is now reported for windows inside a `NavigationWindow`.
  - `safeAreaPadding` is only set when `extendSafeArea` is `true`.

  ### Other iOS
  - `Ti2DMatrix` `apiName` corrected to `Ti.UI.Matrix2D`.
  - `CodecModule.nativeByteOrder` is now exposed.
  - `TiViewProxy` queues are released after `_destroy` in `dealloc`.
  - Hardened crash prevention and resource cleanup across several core modules.
  - `TiUITabGroup` / `TiUITabProxy` / `TiUITextArea`: various correctness fixes.
  - `TiApp+Addons`, `TiAppiOSProxy`, `TiMediaVideoPlayerProxy`, `TiUIWebView`, `TiUIListView`, `TiUITableView`: minor fixes.

  ### Re-entrant script errors
  - Guarded re-entrant script errors to prevent stack overflow when an error handler itself throws.

  ---

  ## Test-suite fixes

  ### iOS 26 simulator flakiness
  Skipped or hardened tests that cannot pass reliably on the iOS 26 simulator (hardware features not available in sim):
  - Geolocation: `getCurrentHeading`, `hasCompass`, `requestLocationPermissions`, `hasLocationPermissions`, `locationServicesEnabled`,
  `locationAccuracyAuthorization`, `locationServicesAuth`.
  - `Ti.UI.iOS.CollisionBehavior`, tertiary label override, `linkColor` override.
  - `VideoPlayer` timeout bumped.
  - `clearCookieUnaffectedCheck` switched to postman-echo.

  ### HTTPClient / network tests
  - Migrated remaining **httpbin.org** tests to **postman-echo.com** (httpbin is unreliable).
  - Fixed Content-Type header lookup to be **case-insensitive** (postman-echo lowercases header names per HTTP/2 convention).
  - Fixed multipart POST file assertion: postman-echo keys files by **uploaded filename** (e.g. `tixhr…png`) rather than the form field name (`attachment`), and
  reports MIME as `application/octet-stream`. The test now verifies the base64 payload is present regardless of the key or MIME type.
  - Fixed a single-character transcription typo in the `expectedBase64` constant (`2` instead of `y` at index 1014) that caused `containEql` to fail even though the
   blob was correct.
  - Hardened four flaky HTTPClient tests and fixed `largeFileWithRedirect` URL encoding / timeout mismatch.

  ### Error-stack tests (V8 CallSite)
  - V8 on newer Android returns `Error.stack` / `Error.nativeStack` as an **array of CallSite objects** rather than a formatted string. `error.test.js` now accepts
  both forms (string or array).

  ### Snapshot baselines
  - `matchImage` assertion now **skips** (generating a baseline) instead of **failing** when no platform-specific baseline image exists yet. This lets the suite
  pass on platforms without baselines while generating them for future comparison.
  - Added iPhone-specific `@3x~iphone.png` snapshots for ListView, TableView, and Window tests; removed the `OS_IOS ? maxPixelMismatch : 0` pattern now that
  device-specific baselines exist.

  ### Test reporter (`build/lib/test/test.js`)
  - **`handleTestContinuation`** rewritten to robustly parse split `!TEST_END:` JSON that is interleaved with ImeTracker / Choreographer / ActivityTaskManager log
  noise:
    - Strips ANSI codes via `stripAnsi` instead of magic `substring()` offsets (8/24/3/13) that assumed an exact prefix layout.
    - Strips `[LEVEL]` and `[device]` prefixes via regex / string match.
    - **Noise filter:** lines whose cleaned content doesn't start with a JSON continuation character (`" { [ } ] : ,` digit `-` / `true` / `false` / `null`) are
  skipped instead of being appended to `partialTestEnd`.
  - **`tryParsingTestResult`** now falls back to `extractBalancedJSON()` — a brace-matching scan that respects string literals and escapes — to salvage the
  outermost `{…}` object when the full string has trailing noise on the same logcat line.
  - Eliminates the `Unknown incomplete test` / `build/lib/test.js failed to parse reported test result` ghost failures.

  ### Other test fixes
  - Added missing `Timeout` require in `ti.android.service.test.js` (8 tests were failing with `Timeout is not defined`).
  - Skip `Notifications enabled by default` on Android API 33+ (POST_NOTIFICATIONS runtime permission not declared by the test app).
  - Clipboard tests made synchronous (replaced `setTimeout`-based waits).
  - WebView test uses `blacklistedURLs` for the deprecated `blacklisturl` event.
  - Hardened locale / date / util / stream tests.

  ---

```
npm run test:integration
5140 Total Tests: 5140 passed, 0 failed, 0 skipped.
5163 Total Tests: 5163 passed, 0 failed, 0 skipped.
```


android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
```diff
@@ -886,9 +886,13 @@ protected void onLayout(boolean changed, int l, int t, int r, int b)
...
-						if (count > 1) {
-							updateRowForHorizontalWrap(right, i);
-						}
+						// Always compute the row height, even with a single
+						// child. Without this, a one-child horizontal layout
+						// leaves horizontalLayoutLineHeight at 0, causing
+						// computePosition() to center the child in a zero-height
+						// area and produce a negative y (e.g. -measuredHeight/2).
+						// This is the root cause of TIMOB-23372 #4/#5.
+						updateRowForHorizontalWrap(right, i);
```
m1ga, line 895: Wow, going through old Jira tickets is crazy 🤯  Great find! My AI said in: https://github.com/tidev/titanium-sdk/pull/14476 No.4:
```
Add early-return guard in updateRowForHorizontalWrap() + skip initial forward scan when count <= 1
Avoids no-op forward scan when past last child; skips pre-scan entirely for single-child horizontal layouts
```
and remove that, so 14.0.0 is having this issue (code from the [JIRA ticket](https://jira-archive.titaniumsdk.com/TIMOB-23372)):

Image

My test cases worked but it looks like I didn't test all cases.

Well done 👍

PR #14488, opened 2026-07-06
fix(android): fix layout regression introduced in #14476

m1ga:
This only affects the current main branch:

In https://github.com/tidev/titanium-sdk/pull/14476 - Change 4 - the `count > 1` was added to "skips pre-scan entirely for single-child horizontal layouts". Turns out that there was an old JIRA ticket that this causes issues: https://jira-archive.titaniumsdk.com/TIMOB-23372 

617411216-43927c24-4443-4d1c-a675-1c0df0babcf5

And @mbender74 found this in his test-suite PR https://github.com/tidev/titanium-sdk/pull/14487

---
  ### `TiCompositeLayout` — TIMOB-23372  (`label.rect.y = -10`)
  - **Root cause:** For single-child horizontal layouts, `updateRowForHorizontalWrap()` was never called because of an `if (count > 1)` guard.
  `horizontalLayoutLineHeight` stayed at `0`, so `computePosition()` centered the child in a zero-height area and produced `label.rect.y = -measuredHeight/2` (e.g.
  `-10`).
  - **Fix:** Always call `updateRowForHorizontalWrap()` for the first child, even when there is only one child. Also moved the `horizontalLayoutLineHeight` tracking
   **before** the `rowWidth > maxRight` early return, so a child wider than the available space still contributes its height to the row.
---

I've made this into a separate PR so we have a clean main branch again that doesn't have this layout issue anymore

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/view/TiCompositeLayout.java
@@ -886,9 +886,13 @@ public class TiCompositeLayout extends ViewGroup implements OnHierarchyChangeLis
 						horizontalLayoutTopBuffer = 0;
 						horizontalLayoutLastIndexBeforeWrap = 0;
 						horizontalLayoutPreviousRight = 0;
-						if (count > 1) {
-							updateRowForHorizontalWrap(right, i);
-						}
+						// Always compute the row height, even with a single
+						// child. Without this, a one-child horizontal layout
+						// leaves horizontalLayoutLineHeight at 0, causing
+						// computePosition() to center the child in a zero-height
+						// area and produce a negative y (e.g. -measuredHeight/2).
+						// This is the root cause of TIMOB-23372 #4/#5.
+						updateRowForHorizontalWrap(right, i);
 					}
 					computeHorizontalLayoutPosition(params, childMeasuredWidth, childMeasuredHeight, right, top, bottom,
 													horizontal, vertical, i);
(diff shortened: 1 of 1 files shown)


prashantsaini1 (commented):
Looks fine but has to be tested in a ListView/TableView as well to ensure the measurement works during lazy loading too.

m1ga:
the first part is just putting the code back to the original state. So it will be the same as in 13.3.0.GA again.

The code from https://github.com/tidev/titanium-sdk/pull/14160#issue-2753986822 looks the same

my list look the same too:
Screenshot_20260706-143342

and we have 100% successful tests (there are a few layout tests in it)

Can setup a TableView test too to verify it.

m1ga:
TableView with a wrapping "horizontal" layout:
```js
var self = Ti.UI.createWindow();

function AddRows() {
	const rows = [];
	for (let i = 0; i < 50; i++) {
			title: "Row " + i,
			color: 'black',
			height: 100
		});
		row.add(createRowView());
		rows.push(row);
	}
	table.data = rows;
};

function createRowView() {
		layout: "horizontal",
		height: Ti.UI.SIZE,
		left: 20,
		right: 20,
		backgroundColor: "#900"
	})
		height: 80,
		width: 120,
		image: 'https://picsum.photos/200/300?' + Math.floor(Math.random() * 10)
	});
		text: "test",
		width: Ti.UI.SIZE,
		height: Ti.UI.SIZE,
		left: 10
	})
	view.add([imageView, label]);
	for (var i=0; i<10;++i) {
		const innerView = Ti.UI.createView({
			width: 20,
			height: 20,
			left: 10,
			backgroundColor: "red"
		})
		view.add(innerView)
	}
	return view;
}

self.add(table);
self.open();
AddRows();
```

13.3.0 and PR look the same:

Screenshot_20260706-145302


and using the broken code from the old JIRA ticket in every view renders fine too

hansemannn (approved):

PR #14490, opened 2026-07-06
fix(android): adjust paths for Windows builds

### LINKED ISSUE #14486: Windows CLI: invalid `file://C:/...` URI generated in Android module Gradle build, plus LiveView ESM URI error
## Summary

I am seeing Windows-specific URI/path issues after moving an Android native module project from Titanium SDK 13.0.x to Titanium SDK `13.3.0.GA`.

The same module builds correctly on Linux with Titanium SDK `13.3.0.GA`, but on Windows the generated Gradle project fails because a Windows path is emitted as:

```text
file://C:/ProgramData/Titanium/mobilesdk/win32/13.3.0.GA/android/m2repository
```

Gradle/Java does not seem to accept this URI as a valid file URI on Windows. It should probably be:

```text
file:///C:/ProgramData/Titanium/mobilesdk/win32/13.3.0.GA/android/m2repository
```

or generated using a proper path-to-URI conversion instead of manually concatenating `file://`.

There is also a separate Windows URI issue when running `ti clean`, apparently related to LiveView / ESM loading, where Node receives a raw `C:` path as an unsupported URL scheme.

## Environment

```text
OS: Windows 10
Titanium SDK: 13.3.0.GA
Previous SDK used by the project: 13.0.x
Project type: Android native module
SDK path: C:\ProgramData\Titanium\mobilesdk\win32\13.3.0.GA
```

The module builds successfully on Linux with Titanium SDK `13.3.0.GA`.

## Build command / context

The failure happens while building an Android native module.

## Actual result

The build fails during Gradle project evaluation.

```text
[INFO]  Building module
[INFO]  [GRADLE]
[INFO]  [GRADLE] > Configure project :module
[INFO]  [GRADLE] w: Deprecated 'org.jetbrains.kotlin.android' plugin usage
[INFO]  [GRADLE] The 'org.jetbrains.kotlin.android' plugin in project ':module' is no longer required for Kotlin support since AGP 9.0.
[ERROR] [GRADLE]
[ERROR] [GRADLE] FAILURE: Build completed with 2 failures.
[ERROR] [GRADLE]
[ERROR] [GRADLE] 1: Task failed with an exception.
[ERROR] [GRADLE] -----------
[ERROR] [GRADLE] * Where:
[ERROR] [GRADLE] Build file 'C:\Users\<user>\Documents\<project-folder>\Modules\<module-name>\android\build\module\build.gradle' line: 11
[ERROR] [GRADLE]
[ERROR] [GRADLE] A problem occurred evaluating project ':module'.
[ERROR] [GRADLE] > Cannot convert URI 'file://C:/ProgramData/Titanium/mobilesdk/win32/13.3.0.GA/android/m2repository' to a file.
```

A second Gradle failure is also reported:

```text
[ERROR] [GRADLE] 2: Task failed with an exception.
[ERROR] [GRADLE] -----------
[ERROR] [GRADLE] A problem occurred configuring project ':module'.
[ERROR] [GRADLE] > Android Gradle Plugin: project ':module' does not specify compileSdk in build.gradle
[ERROR] [GRADLE]   (C:\Users\<user>\Documents\<project-folder>\Modules\<module-name>\android\build\module\build.gradle).
```

I suspect the `compileSdk` error may be secondary/cascading, because the build file already fails earlier while evaluating the malformed `file://C:/...` repository URI.

Gradle also prints this Kotlin/AGP warning:

```text
Solution: Remove both android.builtInKotlin=true and android.newDsl=false from gradle.properties, then migrate to built-in Kotlin.

Problem found: Deprecated 'org.jetbrains.kotlin.android' plugin usage
The 'org.jetbrains.kotlin.android' plugin in project ':module' is no longer required for Kotlin support since AGP 9.0.
```

This may be unrelated to the URI failure, but I am including it because it appears in the same build output.

## Expected result

The generated Gradle build files should use a valid Windows file URI for the Titanium Android Maven repository, for example:

```text
file:///C:/ProgramData/Titanium/mobilesdk/win32/13.3.0.GA/android/m2repository
```

or an equivalent URI produced by a proper path-to-URI conversion.

The same module should build on Windows as it does on Linux.

## Temporary workaround / patch

As a temporary workaround, I patched the Titanium SDK files that generate the Android module Gradle files.

The patch changes generated `file://C:/...` style paths into `file:///C:/...`.

```powershell
param(
	[string]$TitaniumSdkRoot = "C:\ProgramData\Titanium\mobilesdk\win32\13.3.0.GA"
)

$ErrorActionPreference = "Stop"

$files = @(
	(Join-Path $TitaniumSdkRoot "android\cli\commands\_buildModule.js"),
	(Join-Path $TitaniumSdkRoot "android\cli\commands\_build.js")
)

foreach ($file in $files) {
	if (!(Test-Path -LiteralPath $file)) {
		Write-Host "SKIP missing $file"
		continue
	}

	$text = Get-Content -LiteralPath $file -Raw
	$next = $text

	$next = $next -replace "'file://'\s*\+", "'file:///' +"
	$next = $next -replace '"file://"\s*\+', '"file:///" +'
	$next = $next.Replace("file://C:/", "file:///C:/")

	if ($next -ne $text) {
		Copy-Item -LiteralPath $file -Destination "$file.bak" -Force
		Set-Content -LiteralPath $file -Value $next -NoNewline
		Write-Host "PATCHED $file"
	} else {
		Write-Host "OK no change needed $file"
	}
}

Write-Host "Done. Re-run Titanium build so it regenerates Gradle files with file:///C:/..."
```

This patch is not meant as a final fix, but it shows where the problem seems to be: the SDK appears to generate Windows file URIs by concatenating `file://` with an absolute Windows path.

On Windows, this produces:

```text
file://C:/...
```

while the valid file URI should be:

```text
file:///C:/...
```

A more robust fix would probably be to avoid manual string concatenation and use a proper file path to URI conversion.

## Separate issue: `ti clean` / LiveView ESM URI error

When I run `ti clean`, I get another URI-related error, apparently from LiveView and Node's ESM loader:

```text
[LiveView] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:195:11)
    at defaultLoadSync (node:internal/modules/esm/load:142:3)
    at #loadAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:776:12)
    at #loadSync (node:internal/modules/esm/loader:796:49)
    at ModuleLoader.load (node:internal/modules/esm/loader:762:26)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:504:31)
    at #getOrCreateModuleJobAfterResolve (node:internal/modules/esm/loader:555:36)
    at afterResolve (node:internal/modules/esm/loader:603:52)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:609:12)
    at node:internal/modules/esm/loader:628:32 {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```

This looks like a similar Windows path normalization issue, but I was not able to fix this one with the patch above.

It seems that somewhere in the LiveView / `ti clean` path, a raw Windows absolute path such as:

```text
C:\...
```

or

```text
C:/...
```

is being passed to the ESM loader as if it were a URL, so Node interprets `C:` as a URL scheme.

Node expects Windows absolute paths used as ESM URLs to be converted to a valid `file:///C:/...` URL first.

## Possible cause

There may be at least two Windows URI conversion problems in Titanium SDK `13.3.0.GA`:

1. Android module Gradle generation emits `file://C:/...` instead of `file:///C:/...`.
2. `ti clean` / LiveView may pass a raw Windows path directly to Node's ESM loader, causing `ERR_UNSUPPORTED_ESM_URL_SCHEME`.

## Request

Could you please check Windows path-to-URI handling in:

```text
android/cli/commands/_buildModule.js
android/cli/commands/_build.js
```

and also in the LiveView / `ti clean` code path?

The Android module build issue seems fixable by generating proper file URIs on Windows, ideally using a standard path-to-file-URL conversion rather than manual string concatenation.

m1ga:
fixes https://github.com/tidev/titanium-sdk/issues/14486

Using `pathToFileURL` to generate the maven paths.

Linux still works fine, windows works fine too: https://github.com/tidev/titanium-sdk/issues/14486#issuecomment-5008144493


android/cli/commands/_build.js
```diff
@@ -2161,8 +2161,7 @@ class AndroidBuilder extends Builder {
 		}
 
 		// Add a reference to the core Titanium library.
-		const tiMavenRepoUrl = 'file://' + path.join(this.platformPath, 'm2repository').replace(/\\/g, '/');
-		this.mavenRepositoryUrls.push(encodeURI(tiMavenRepoUrl));
+		this.mavenRepositoryUrls.push(pathToFileURL(path.join(this.platformPath, 'm2repository')).href);
```
cb1kenobi, line 2164: What's the difference between `.href` and `.toString()`?

m1ga: just less typing :smile: 
https://nodejs.org/api/url.html#urlhref
`Getting the value of the href property is equivalent to calling [url.toString()](https://nodejs.org/api/url.html#urltostring).`

cb1kenobi (approved):

PR #14497, opened 2026-07-10
fix(android): tableview content update fix

### LINKED ISSUE #13508: Android: Reasigning tableView data will lose click event

### Description

When reasigning `tableView.data` with the last data it will lose the click event.

### Expected Behavior

Keep the click event

### Actual behavior

-

### Reproducible sample

```js
const btn = Ti.UI.createButton({bottom: 0,title: "new data"});
const tableView = Ti.UI.createTableView({});
var tableData = [];
var t = 0;

function setData() {
	if (t > 0) {
		tableView.data = tableData;
		console.log("set data 2");
		return
	}

	t++;
	for (var i = 1; i <= 20; i++) {
		var entryRow = Ti.UI.createTableViewRow({height: 40});
		var entryNameLabel = Ti.UI.createLabel({text: "test"});
		entryRow.add(entryNameLabel);
		tableData.push(entryRow);
	}
	tableView.data = tableData;
	console.log("set data 1");
}


tableView.addEventListener("click", function(e) {
	console.log(e.index);
})
btn.addEventListener("click", function(e) {
	setData();
})

setData();
win.add([tableView, btn]);
win.open();
```

### Steps to reproduce

* run the code
* click on a row
* will output the index
* click the "new data" button
* no output when clicking the rows

### Platform

Android

### SDK version you are using

11.0.0

_No response_

m1ga:
reviving and fixing an old PR: https://github.com/tidev/titanium-sdk/pull/13826 to fix https://github.com/tidev/titanium-sdk/issues/13508

Changes:
* This prevents unnecessary cloning in TableViewSectionProxy.add() when the same row proxy objects are re-assigned. Without this, releaseViews() leaves stale parent references, causing rows to be cloned — and cloned proxies lose their JS event listeners.
* After releaseViews() destroys row views (including the GestureDetector), DiffUtil would incorrectly skip re-binding when it sees the same proxy objects. update(true) calls notifyDataSetChanged(), forcing the adapter to properly recreate and rebind all ViewHolders, which re-initializes touch handling.

## Test 1:

see https://github.com/tidev/titanium-sdk/pull/13826 for more details

```js
var menu = [];
var numrows = 0;

  if (e.index == numrows - 1)
    AddRows();
});

function AddRows() {
  if (menu.length == 0)
    row.add(lab);
  menu.push(row);
  numrows++;
  menu.push(row);
  numrows++;
  table.data = menu;
};

AddRows();
self.add(table);
self.open();
```

* start and click the last row a few times
* then click any other row and see if the click event still works

## Test 2
```js
var tableData = [];

for (var i=1; i<=10; i++){
    className: 'forumEvent', // used to improve table performance
	height: 50
  });
  var labelUserName = Ti.UI.createLabel({
    text: 'Fred Smith ' + i,
	left: 10
  });
  row.add(labelUserName);
  tableData.push(row);
}

  data: tableData
});

win.add(tableView);
win.open();
tableView.data[0].rows[0].children[0].text = "test"
```

* start and see if the first label says "test"

---

## Test 3
```js

function AddRows() {
	const rows = [];
			title: "Row " + i,
			color: 'black',
			height: 100
		});

		row.add(createRowView());
		
		rows.push(row);
	}
	
	table.data = rows;
};

function createRowView() {
		height: 80,
		width: 120
	});
	
	setTimeout(() => {
	}, 1000);

	return imageView;
}

self.add(table);
self.open();
AddRows();
```
* images should show up

---

## Test 4
```js
const label = Ti.UI.createLabel({ top: 0, height: 40, text: 'click a row' });
const button = Ti.UI.createButton({ bottom: 0, height: 50, title: 'reassign sections' });
const tableView = Ti.UI.createTableView({ top: 40, bottom: 50 });

const sections = [];
for (let s = 0; s < 3; s++) {
      const section = Ti.UI.createTableViewSection({ headerTitle: 'Section ' + s });
      for (let r = 0; r < 4; r++) {
              section.add(Ti.UI.createTableViewRow({ title: 'Section ' + s + ' Row ' + r, height: 44 }));
      }
      sections.push(section);
}
tableView.data = sections;

tableView.addEventListener('click', e => {
      label.text = 'index ' + e.index + ' / section ' + (e.section ? e.section.headerTitle : 'none');
      e.row.title = 'clicked';
});

button.addEventListener('click', () => {
      tableView.data = tableView.data;
      label.text = 'sections reassigned, click a row';
});

win.add([ label, tableView, button ]);
win.open();
```

* click on rows
* click on the reassign button
* click on rows again


cb1kenobi (commented):
The Android fix itself looks sound: unparenting rows before releaseViews avoids the clone that loses JS listeners, re-parenting on the section path keeps bubbling intact, and update(true) forces a full rebind past DiffUtil. The new regression test, however, asserts proxy identity with should().be.eql(), which is deep structural equality. A cloned row with the same title would still pass, so the test would not catch a regression of the very bug it guards. Switching those proxy assertions to .be.equal() fixes it.

—
Reviewed cf4731fb

tests/Resources/ti.ui.tableview.test.js
```diff
@@ -1090,6 +1090,53 @@ describe('Titanium.UI.TableView', function () {
...
+			clickCount++;
+		});
+
+		win = Ti.UI.createWindow();
+		win.addEventListener('focus', () => {
+			try {
+				// Re-assigning the same rows must keep the row proxies, not clone them,
+				// otherwise their event listeners would be lost. (See issue #13508)
+				tableView.data = rows;
+				should(tableView.sections[0].rows[0]).be.eql(rows[0]);
```
cb1kenobi: **Identity assertions use `eql`, so the regression test can't fail** — _Severity: medium (test-only)_

This test exists to prove the row proxies are **not cloned** when the same array is re-assigned, but `should(x).be.eql(y)` is deep structural equality, not identity. A cloned `TableViewRowProxy` created with the same `title` would compare equal, so if the unparenting fix in `TableViewProxy.setData()` regresses, this test still passes and #13508 returns silently. The same applies to the `rows[1]`, `parent`, `sections[0]`, and `section.rows[0].parent` assertions below.

**Fix:** use strict identity for every proxy comparison in this test, e.g. `should(tableView.sections[0].rows[0]).be.equal(rows[0]);`

—
Reviewed cf4731fb

m1ga: Fixed the test and ran it on my phone :+1:

cb1kenobi (commented):
The new commits fixed the one issue from the previous review: the regression test now asserts proxy identity with strict equal instead of deep eql, so a cloned row would actually fail it. The Android setData change (unparent rows before releaseViews, re-parent on the section path, force a full rebind with update(true)) is coherent and matches the bug being fixed. No blocking issues remain on the changed lines and all CI checks pass.

—
Reviewed 37c481c2

cb1kenobi (commented):
The Android setData fix unparents rows before releaseViews so re-assigned proxies are not cloned, re-parents rows on the section path, and forces a full rebind with update(true). The earlier review's one issue is fixed: the regression test now uses strict equal for proxy identity, so a cloned row would fail it. No blocking issues remain on the changed lines and CI is green.

—
Reviewed 37c481c2

PR #14499, opened 2026-07-10
chore: Github workflow for Android Mocha tests

m1ga:
**Android mocha test workflow**

* will run on demand, not on every push
* the "Titanium SDK Android Test" summary page contains an artifact with the full XML test report
* I've added a summary to the root node: `<testsuites name="Mocha Tests" passed="4868" failed="53" skipped="0">`
* runs around 18-24min
* `ti.locale.test` was commented out as that will currently restart the test app which will stop it
* note: https://github.com/tidev/titanium-sdk/pull/14487 will fix the failed tests

iOS version should be added as a separate workflow later so we can run them independently

### DIFF
--- a/.github/workflows/emulator-test.yml
+++ b/.github/workflows/emulator-test.yml
@@ -1,0 +1,70 @@
+name: Android Emulator Test
+on:
+  push:
+    branches:
+    - master
+    - main
+    - "[0-9]+_[0-9]+_X"
+  pull_request:
+
+jobs:
+  test:
+    runs-on: macos-26
+    name: Android Emulator Tests
+    env:
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
+      USE_CCACHE: 1
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          cache: 'npm'
+
+      - name: Setup Java
+        uses: actions/setup-java@v5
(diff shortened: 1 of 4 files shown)


hansemannn:
Do we have unlimited actions runner usage as an OSS project? If so, we should re-enable it at least for every commit into main.

m1ga:
https://docs.github.com/en/actions/reference/limits - no clue :smile: but it says "GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners" here: https://docs.github.com/en/actions/concepts/billing-and-usage?azure-portal=true&utm_source=chatgpt.com#about-billing-for-github-actions

We can enabled that once the other PR is fixing the tests. Maybe even something like:
```
on:
  pull_request_review:
    types: [submitted]

jobs:
  after-approval:
    if: github.event.review.state == 'approved'
```
to run when a PR was approved. Then it would run for potential PR that we want to have in main. Once it was approved and the tests are successful it would be ready to be merged.

hansemannn:
Excellent idea, lets do it

m1ga:
nice, so this is the logic now:


Review action | Workflow triggered | android job runs
-- | -- | --
Approve | Yes | **Yes**
Comment | Yes | No
Request changes | Yes | No

cb1kenobi:
Love this! Nice work!

hansemannn (approved):

m1ga:
nice, it worked and ran after approval (result: `passed="4873" failed="48" skipped="0"`) :+1:

PR #14502, opened 2026-07-13
chore: iOS test workflow

### LINKED ISSUE #13330: Test builds
We should implement this in the github actions:
https://github.com/tidev/titanium_mobile/blob/734c41b56cd63010b269ef6b6e6247da2f9d656c/Jenkinsfile#L72-L223

m1ga:
Fixes https://github.com/tidev/titanium-sdk/issues/13330 by adding the iOS workflow too. Then we have Android and iOS running on Github :+1: 

* renaming the android workflow file to `tests-android`
* adding `tests-ios.yml` to run the iOS test-suite
* guarding blob test and video test as they will currently crash the iOS simulator (is fixed in https://github.com/tidev/titanium-sdk/pull/14487 )


.github/workflows/tests-ios.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Install Titanium CLI
+        run: sudo npm i -g titanium && ti sdk install
+
+      - name: Run iOS simulator tests
+        run: npm run test:ios
+
+      - name: Upload iOS JUnit report
```
cb1kenobi, line 36: do we do anything with the JUnit report?

cb1kenobi (approved):
Awesome work!

.github/workflows/tests-ios.yml
m1ga: that contains the full result of the tests. At the moment we would need to download it and see if the tests succeded. Of course it would be better to have it uploaded and displayed somewhere to make it easier to check for errors. Not sure how it was with the old appc tests

m1ga:
was just about to write it is frozen but it's running the ios tests now. Locally the iOS tests ran faster than Android. Here it is the other way round. Let's see how long it takes...edit: 32 minutes

m1ga:
:facepalm: Used the wrong filename for the report. Fixed now

cb1kenobi (approved):

m1ga:
Hmmm...looks like the file is still different online and it didn't finish the first run so there was no file. But I see `JUnit test report written to /Users/runner/work/titanium-sdk/titanium-sdk/junit.ios.iphone.xml` now and the output is `5151 Total Tests: 5087 passed, 64 failed, 0 skipped.` at the moment.

PR #14519, opened 2026-07-16
fix(android): webview — loading property and file: URI NPE

mbender74:
- Exposes the WebView `loading` property so JS can read the current load state.
- Fixes a `file:` URI `NullPointerException` in the WebView load path.

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/WebViewProxy.java
@@ -587,6 +587,16 @@ public class WebViewProxy extends ViewProxy implements Handler.Callback, OnLifec
 		}
 	}
 
+	@Kroll.getProperty
+	public boolean getLoading()
+	{
+		TiUIView v = peekView();
+		if (v instanceof TiUIWebView) {
+			return ((TiUIWebView) v).isLoading();
+		}
+		return false;
+	}
+
 	public void clearBasicAuthentication()
 	{
 		fusername = null;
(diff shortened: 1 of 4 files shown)


m1ga:
nice! Please update the apidocs https://github.com/tidev/titanium-sdk/blob/main/apidoc/Titanium/UI/WebView.yml#L953-L956 too:

```
since: {android: "14.0.0", iphone: "0.8", ipad: "0.8", macos: "9.2.0"}
```

mbender74:
@m1ga done

m1ga:
working fine, test code:
```js
	top: 0,
	height: 200
});
var lbl = Ti.UI.createLabel({
	top: 250,
	text: "loading: false"
})

setInterval(function() {
	lbl.text = "loading: " + webview.loading;
}, 50)

	title: "load",
	bottom: 100
})
webview.addEventListener("beforeload" ,function(){
	console.log("beforeload")
})
webview.addEventListener("load" ,function(){
	console.log("load")
})

btn.addEventListener("click" ,function(){
	webview.url = "https://titaniumsdk.com"
})
window.add([webview, lbl, btn]);
window.open();
```

* press the button
* `.loading` is showing true/false

m1ga (approved):

m1ga (changes requested):
I had Claude check this PR and it came up with a clever way to check for iLoading() with less code:

### Recommended fix

1. Derive loading from the WebView instead of a hand-set flag. In TiUIWebView, replace the field, isLoading(), and setLoading() with:
```
public boolean isLoading()
{
      return (webView != null) && (webView.getProgress() < 100);
}
```
Chromium reports 100 when idle and below 100 from the start of a navigation until it stops, so this also covers stopLoading(), cancelled navigations, and the superseded-navigation case where a late onPageFinished for the old page would otherwise reset the flag early. The proxy getter can stay as it is.

2. Revert TiWebViewClient to the async fireEvent and delete the three setLoading calls and the comment block. The beforeload event goes back to the same FIFO queue as load, error, and progress.

PR #14525, opened 2026-07-16
fix(android): make Geolocation.headingFilter settable from JS; add accuracy default

mbender74:
- Fixes a bug where Ti.Geolocation.headingFilter could not be set from JavaScript on Android, and adds a missing default for Ti.Geolocation.accuracy.

### DIFF
--- a/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
+++ b/android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
@@ -106,13 +106,57 @@ import androidx.core.location.LocationManagerCompat;
  * accuracy, frequency properties or even changing modes are respected and kept but don't actually get applied on the OS until
  * the listener count is greater than 0.
  */
-@Kroll.module(propertyAccessors = { TiC.PROPERTY_ACCURACY })
+@Kroll.module(propertyAccessors = {
+	TiC.PROPERTY_ACCURACY,
+	"activityType",
+	"distanceFilter",
+	"headingFilter",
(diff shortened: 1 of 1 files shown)


android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
```diff
@@ -106,14 +106,58 @@
...
+	@Kroll.constant
+	public static final int ERROR_NETWORK = 2;
+	@Kroll.constant
+	public static final int ERROR_HEADING_FAILURE = 3;
+	@Kroll.constant
+	public static final int ERROR_REGION_MONITORING_DENIED = 4;
+	@Kroll.constant
+	public static final int ERROR_REGION_MONITORING_FAILURE = 5;
+	@Kroll.constant
+	public static final int ERROR_REGION_MONITORING_DELAYED = 6;
```
hansemannn: Where can these errors be thrown? I cannot see that from the current code. If you are just shimmering them is that an approach that we do not try to follow, as it adds false parity.

hansemannn (changes requested):

android/modules/geolocation/src/java/ti/modules/titanium/geolocation/GeolocationModule.java
mbender74: valid point! I removed the shimmering and now only the true

mbender74:
PR closed because change is minimal and not needed

hansemannn:
No worries!

PR #14528, opened 2026-07-20
feat(all): snapping for ListView/TableView

m1ga:
Rewrite of my old PR https://github.com/tidev/titanium-sdk/pull/13180 with iOS support.

```js
var btn = Ti.UI.createButton({title:"snap",bottom:0});
var snapping = true;
	snapping: snapping,
	templates: {
		test: {
			childTemplates: [{
				type: 'Ti.UI.View',
				childTemplates: [{
					type: 'Ti.UI.Label',
					bindId: 'label',
					properties: {
						color: 'black',
						bindId: 'label'
					}
				}],
				properties: {
					width: Ti.UI.FILL,
					height: 200,
					cardUseCompatPadding: true,
					backgroundColor: 'white',
				}
			}]
		}
	},
	defaultItemTemplate: 'test'
});
var items = [];

	items.push({
		label: {
			text: "item " + i
		},
		template: 'test'
	})
}



section.setItems(items);
btn.addEventListener('click', (e) => {
	snapping = !snapping;
	listView.snapping = snapping;
});
win.add(listView);
win.add(btn);
win.open();
```


android/modules/ui/src/java/ti/modules/titanium/ui/widget/tableview/TiTableView.java
```diff
@@ -432,6 +436,28 @@ public TiNestedRecyclerView getRecyclerView()
...
 
+	/**
+	 * Enable or disable snapping of rows to the nearest position after a scroll.
+	 *
+	 * @param value Set true to snap rows into place.
+	 */
+	public void setSnapping(boolean value)
+	{
+		if (value == (this.snapHelper != null)) {
+
```
hansemannn: Why the empty line?

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIListView.java
```diff
@@ -248,6 +248,12 @@ private void processProperty(String name, Object value)
 			this.listView.getRecyclerView().setVerticalScrollBarEnabled(TiConvert.toBoolean(value, true));
 		}
 
+		if (name.equals(TiC.PROPERTY_SNAPPING)) {
+
+			// Set item snapping.
```
hansemannn, line 252: Same here - please update your local AI settings and make sure to improve formatting to catch this

hansemannn (changes requested):

android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIListView.java
m1ga: we have those in the current file like https://github.com/tidev/titanium-sdk/blob/main/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIListView.java#L188-L193 or https://github.com/tidev/titanium-sdk/blob/main/android/modules/ui/src/java/ti/modules/titanium/ui/widget/TiUIListView.java#L136-L139 so it just used the same structure 😄

hansemannn: Thats not a very good reason to use it :) Then every file should be change. We should not copy-over bad style.

m1ga: no, of course not! Was more a note where I got it from. I only removed those in my changes now, did not touch any other code for now to make changes to old lines.
There might be a better way to do that perhaps with a lint rule that will remove empty lines between if and a comment.

hansemannn (approved):

PR #14530, opened 2026-07-22
feat(chore): remove danger files, move to native fetch/stream

m1ga:
* removing the unused danger files
* removing old & deprecated request/request-promise-native modules and replacing them with native calls


cb1kenobi (approved):
This is awesome! I think I have a branch somewhere that removed `request`, but this is way better. Thanks!

m1ga:
have to use all the Claude Max I can as long as I have it :smile:

PR #14532, opened 2026-07-23
feat(android): ti.crypt

m1ga:
This will now use ti.crypt instead of ti.cloak if you set `<encryption>ti.crypt</encryption>` in your tiapp.xml
Default is still `ti.cloak` if you don't set it or set it to a wrong value.

**How to test it:**
* build a dist android app
* open the APK as a zip and check the lib/*/ folder and see if there is `libti.cloak.so` or `libti.crypt.so`
* start the app and see if it runs (so you know it will encrypt the files)

a decompiled APK still contains file.js.bin files with encrypted content:
Bildschirmfoto_20260730_153746
so that looks the same :+1:


build/lib/packager.js
```diff
@@ -168,8 +168,9 @@ export class Packager {
 			await exec('node bin/download-all.js', { cwd: dir, stdio: 'inherit' });
 		}
 
-		// Include 'ti.cloak'
-		return unzip(path.join(ROOT_DIR, 'support', 'ti.cloak.zip'), path.join(this.zipSDKDir, 'node_modules'));
+		// Include the 'ti.cloak' and 'ti.crypt' encryption modules
+		await unzip(path.join(ROOT_DIR, 'support', 'ti.cloak.zip'), path.join(this.zipSDKDir, 'node_modules'));
+		return unzip(path.join(ROOT_DIR, 'support', 'ti.crypt.zip'), path.join(this.zipSDKDir, 'node_modules'));
```
cb1kenobi: Code looks wonky. I think we should be consistent and await both unzips.
```suggestion
		await unzip(path.join(ROOT_DIR, 'support', 'ti.crypt.zip'), path.join(this.zipSDKDir, 'node_modules'));
```

cb1kenobi (approved):
Code looks great! Tested on macOS 26.5.1, Node.js 24, emulator and device builds.

m1ga:
Pushed my first app with 14.0.0 + ti.crypt to the play store. App update worked fine, all code is loading. So I think we are good to go here for the first version of it (opt-in of course).

cb1kenobi (approved):
Sweet, code looks good. Ship it!

hansemannn (approved):

PR #14533, opened 2026-07-27
feat(android): new Material3 stopIndicator property for Ti.UI.ProgressBar

m1ga:
**stopIndicator**
If you use a Material3 Theme and a ProgressBar it looks like this by default:
Screenshot_20260727-205415
Notice the little "stop indicator" dot at the end. Currently there is no way to remove that so I've added a new property `stopIndicator` to fix that.

Screenshot_20260731-120824

**Bug fix:**
There is also a bug in the value calculation when you don't use min:0. The example code in 13.3.1 will show a full bar when you use min:50, max:100, value: 75 but it should be at 50%.

1. convertRange at TiUIProgressBar.java:134 ignores min in the numerator: (value / (max - min)) should be ((value - min) / (max - min)). With min: 50, max: 100, value: 75 the bar shows 150% (clamped to full) instead of 50%. It only works correctly when min is 0.
2. If min == max (both default to 0 when unset), the division yields NaN/Infinity — no crash, but the bar silently shows a nonsense value rather than guarding the degenerate range.

This is fixed in this PR too.

**Test code:**
```js
	theme: "Theme.Titanium.Material3.DayNight",
	layout:"vertical"
});
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 0, max: 100, value: 10,
}));
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 0, max: 100, value: 30,
	stopIndicator: false
}));
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 0, max: 100, value: 50,
	trackThickness: 10,
	stopIndicator: { enabled: true, size: 10}
}));
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 0, max: 100, value: 50,
	trackThickness: 10,
	stopIndicator: { enabled: true, size: 5}
}));
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 50, max: 100, value: 75,
	trackThickness: 20,
	stopIndicator: { size: 20}
}));
win.add(Ti.UI.createProgressBar({
	top: 10, width: 250, min: 50, max: 100, value: 95,
	stopIndicator: { enabled: false }
}));
win.open();
```

If you run this code without the M3 theme is will just show the progressbar as usual, `stopIndicator` is ignored

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/ProgressBarProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/ProgressBarProxy.java
@@ -25,6 +25,7 @@ import ti.modules.titanium.ui.widget.TiUIProgressBar;
 		TiC.PROPERTY_COLOR,
 		TiC.PROPERTY_TINT_COLOR,
 		TiC.PROPERTY_TRACK_TINT_COLOR,
+		TiC.PROPERTY_SHOW_STOP_INDICATOR,
 	})
 public class ProgressBarProxy extends TiViewProxy
 {
@@ -32,6 +33,7 @@ public class ProgressBarProxy extends TiViewProxy
 	{
 		super();
 		defaultValues.put(TiC.PROPERTY_ANIMATED, true);
+		defaultValues.put(TiC.PROPERTY_SHOW_STOP_INDICATOR, true);
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


hansemannn:
- Think of a more cross-platform naming convention if possible
- Do we have a shared plan on when and how to move MT2 components to MT3?

m1ga:
Will think about a better name :+1: I think `stopIndicator` would even be better already (that's how it is called on Android, not sure if iOS even has something like this).

Not sure when we should make M3 the default theme but we should start at some point. I'm using it for new apps as a custom theme and it works fine, even the old map crash in TabGroups is gone (think one internal Android library upgrade fixed it :smile: ). We still need to expose some properties (like the three new styles for the CardLayout so you can set the M2 layout again)

m1ga:
Have a better idea:
`stopIndicator: false` (shortcut to disable it) and `stopIndicator: {enabled: true|false , size: 2}` so we have all options to show/hide and configure the size.

PR #14534, opened 2026-07-29
fix(ios): install and launch device builds via devicectl

hansemannn:
This fixes a long-outstanding issue where device builds did not work since a while. Now they do again, with much more modern tooling! 🎉 

## Summary
- Fix device builds silently skipping the install step for iOS 17+ devices: they connect exclusively through CoreDevice/RemoteXPC and are invisible to the legacy MobileDevice framework (ioslib/node-ios-device), even over USB
- Replace the iOS install hook entirely with `xcrun devicectl` (ships since Xcode 15): device discovery, install, auto-launch, and console log streaming
- Wait for a locked device: the hook polls `devicectl device info lockState` and launches the app as soon as the device is unlocked
- Devices are matched by `pairingState` rather than tunnel state, since CoreDevice establishes the tunnel on demand (installs work over USB and Wi-Fi)

## Changes
- `iphone/cli/hooks/install.js`: drop ioslib/node-ios-device usage; discover devices via `devicectl list devices --json-output`, install via `devicectl device install app`, launch via `devicectl device process launch --console` with app output piped through the Titanium logger (same level parsing / CTRL-C behavior as the old log relay)

## Test Plan
- `ti build -p ios -T device` with an iPhone 17 Pro (iOS 26): verified build → install → auto-launch → live log streaming, over both USB and Wi-Fi
- Locked-device flow: verified the hook waits, then installs/launches immediately after unlocking
- `--build-only` still skips the install step

Note: pre-iOS-17 devices are no longer supported by this hook since CoreDevice cannot reach them.

### DIFF
--- a/iphone/cli/hooks/install.js
+++ b/iphone/cli/hooks/install.js
@@ -1,15 +1,67 @@
 /*
  * install.js: Titanium iOS CLI install hook
  *
+ * Installs and launches device builds via `xcrun devicectl` (CoreDevice).
+ * The legacy MobileDevice.framework path (ioslib/node-ios-device) is gone:
+ * iOS 17+ devices connect exclusively through CoreDevice/RemoteXPC and are
+ * invisible to the old API, even over USB.
+ *
  * Copyright TiDev, Inc. 04/07/2022-Present. All Rights Reserved.
  * See the LICENSE file for more information.
  */
 
 import appc from 'node-appc';
 import async from 'async';
-import ioslib from 'ioslib';
+import { execFile, spawn } from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
 
 export const cliVersion = '>=3.2';
+
+const EXEC_LIMIT = { maxBuffer: 10 * 1024 * 1024 };
+
+function listDevices(callback) {
+	const jsonFile = path.join(os.tmpdir(), `ti-devicectl-${process.pid}.json`);
+	execFile('xcrun', [ 'devicectl', 'list', 'devices', '--json-output', jsonFile ], EXEC_LIMIT, function (err, stdout, stderr) {
+		let devices = [];
+		if (err) {
+			return callback(new appc.exception('Failed to list devices via devicectl', (stderr || stdout || '').trim().split('\n')));
+		}
+		try {
+			devices = JSON.parse(fs.readFileSync(jsonFile, 'utf8')).result.devices;
+		} catch (e) {
+			return callback(new appc.exception(`Failed to parse devicectl output: ${e.message}`));
+		} finally {
+			try {
+				fs.unlinkSync(jsonFile);
+			} catch (e) {
+				// ignore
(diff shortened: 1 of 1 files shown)


cb1kenobi:
Oh, interesting!

hansemannn:
Yep - we may be able to drop node-ios-device in general and do everything using their top level CLIs moving forward!

hansemannn:
@cb1kenobi I am doing some more refactoring to get this into ioslib, so we can reuse parts of it for the device detection, which is the last node-ios-device usage. Then we can ditch it completely 😎

cb1kenobi:
@hansemannn I've been refactoring ioslib to TypeScript, modern JS, etc here: https://github.com/tidev/ioslib/tree/update-node-ios-device-v6. Go ahead and add what you will to v5 and I add it into the v6 branch.

I haven't had time to look into `devicectl`. I'm not sure how getting logs works or if it supports installing over Wi-Fi, both of which `node-ios-device` v6 supports. v6 also supports watching for devices being connected/disconnected, but the CLI doesn't need this today.

Aside from finding the free time to hack, I'm losing motivation on the ioslib refactor. It won't be long until I'm completely block by not having a TiDev Apple developer account to test certs, provisioning profile, and device builds. @joshualambert any update on this?

designbymind:
Awesome work! 📲

**Quick question:** is it currently possible to install a Ti app to device over Wi-Fi (specifically via VS Code extension)?

hansemannn:
@designbymind With this change, it could be possible, as it checks for remote devices as well. Give it a shot!

designbymind:
@hansemannn after applying the `install.js` changes, I'm able to build for both simulator and device (with Xcode 26.5/27) but I'm not able to see (or build to) devices connected via Network. Using Xcode 27's DeviceHub, I'm able to Launch my app via Network/Wi-Fi but I'm not seeing any devices in the device list (in VS Code extension) over the network, only when plugged in via USB(C). My device is paired for development. Any thoughts? 💭

designbymind:
@hansemannn okay, super cool! Even though I wasn't able to get the VS Code extension to perform builds over Wi-Fi network, I was able to successfully build over Wi-Fi (using Xcode 27 b4 on iOS 27 bX) with a simple `ti build -p ios -T device`. The app auto-launched and the logger is working perfectly. Just wanted to share as I've been wanting this for such a long time. **Well done man!** This is so great, thank you ✨

hansemannn:
@designbymind Thank you! @cb1kenobi I have the ioslib changes ready locally, but not much time to ship it this week. Will try to get something test-ready by end of next week! :-)

cb1kenobi:
@hansemannn Awesome! With this, we can get rid of node-ios-device, right?

hansemannn:
Once the ioslib change is in, yes!

hansemannn:
Superseded by https://github.com/tidev/ioslib/pull/261 - I will create another PR to change the calling code here once the ioslib update is released!

PR #14542, opened 2026-08-06
[14538] fix(ios): prevent Kroll reentrancy crash

### LINKED ISSUE #14538: iOS app crashes in certain race condition since SDK 13.2.0.GA

### Description

iOS app crashes in a certain specific race condition, starting since SDK 13.2.0.GA (possibly due to [Kroll core GCD PR](https://github.com/tidev/titanium-sdk/pull/14395)).

Conditions to crash the sample code:
1. Next window must have its `open` event.
2. Next window must be opened first before setting the HTML property on a label in previous window.

### Expected Behavior

App should not crash due to this race condition, and such race conditions can also occur in other variants.

### Actual behavior

App crashes when we click on the blue colored button.

### Reproducible sample

```
	backgroundColor: 'white'
});

	text: 'This is a Label',
	color: 'black',
	top: 100
});

	title: 'Click to set HTML on label',
	color: 'blue',
	bottom: 100
});

button.addEventListener('click', () => {
	// Crash happens only if we first open the window and set HTML next.
	openAnotherWindow();
	label.html = '<b>Dummy HTML</b><br/>The label was updated.';
});

window.add([label, button]);
window.open();

function openAnotherWindow() {
	var nextWindow = Ti.UI.createWindow({
		title: 'Navigation Window',
		backgroundColor: 'teal'
	});

	var navLabel = Ti.UI.createLabel({
		text: 'This is a Label on Nav Window',
		color: 'white',
	});

	navLabel.addEventListener('click', ()=> nextWindow.close());

	nextWindow.add(navLabel);

	nextWindow.addEventListener('open', function () {}); // Crashes only when the "open" event is present + the order based condition in button click event.
	nextWindow.open();
}

```

### Steps to reproduce

Run the sample code on iOS and click the blue colored button. The app would crash.

### Platform

iOS

### SDK version you are using

13.2.0.GA

_No response_

hansemannn:
## Summary
- Fix an iOS crash caused by queue-identity Kroll reentrancy detection during nested main run loops.
- Restore recursive mutex semantics so same-thread reentry and Objective-C exception propagation remain safe.
- Add deterministic integration coverage for opening a window while parsing label HTML.
- Fixes #14538.

## Changes
- Kroll locking: Replace the serial dispatch queue with an exception-safe recursive pthread mutex behind the existing helper.
- Tests: Exercise the button event, asynchronous window open, and label HTML nested-run-loop path.

## Test Plan
- [x] Reproduce the pre-fix `__DISPATCH_WAIT_FOR_QUEUE__` crash with the new regression test.
- [x] Run the focused Label regression test: 1 passed, 0 failed.
- [x] Run `Native exception surfaced`: 1 passed, 0 failed.
- [x] Run the complete Label suite: the regression passes; one unrelated `en-GB` localization assertion fails.
- [x] Run the broad iPhone integration suite without a process crash: 4,124 passed, 25 unrelated simulator/environment failures.
- [x] Run `npm run lint:objc`.
- [x] Run `npx oxlint tests/Resources/ti.ui.label.test.js`.
- [x] Run `npm run ios-sanity-check`.

### DIFF
--- a/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
+++ b/iphone/TitaniumKit/TitaniumKit/Sources/Kroll/KrollContext.m
@@ -13,18 +13,21 @@
 #import "TiUtils.h"
 
 #import "TiExceptionHandler.h"
+#include <pthread.h>
 
 static unsigned short KrollContextCount = 0;
 
-static dispatch_queue_t krollEntryQueue;
-static void *krollEntryQueueKey = &krollEntryQueueKey;
+static pthread_mutex_t KrollEntryLock;
 
+// Nested main run loops can re-enter Kroll on the owning thread under a different
+// dispatch queue identity, so Kroll entry must be recursive by thread.
 static inline void KrollEntryLockPerform(dispatch_block_t block)
 {
-  if (dispatch_get_specific(krollEntryQueueKey) == krollEntryQueueKey) {
-    block(); // Already on this queue — re-entrant call
-  } else {
-    dispatch_sync(krollEntryQueue, block);
+  pthread_mutex_lock(&KrollEntryLock);
+  @try {
+    block();
+  } @finally {
+    pthread_mutex_unlock(&KrollEntryLock);
   }
 }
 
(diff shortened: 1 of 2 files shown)


m1ga (approved):
Test app runs fine without a crash. 
Claude Code review is also happy with the changes.
Also tested two other apps (not using label.html) to see if they still work fine, which they do.

PR #14543, opened 2026-08-09
[14544] feat(ios): support Icon Composer app icons

abentley8:
**GitHub issue:** https://github.com/tidev/titanium-sdk/issues/14544

## Summary

Xcode 26 introduced Icon Composer documents (`.icon`), which describe an app icon as layers with Liquid Glass effects rather than a set of flattened PNGs. `actool` renders every size and appearance (default, dark, tinted) from one document, including the rasterized icons that older iOS versions still need.

Titanium currently has no way to use one — the iOS builder only ever generates an `AppIcon.appiconset` from `DefaultIcon.png`, so apps built with the SDK get the system's automatic fallback treatment on iOS 26 instead of a real Liquid Glass icon.

## What this adds

A project can drop a `DefaultIcon.icon` (or `DefaultIcon-ios.icon`) in its root, alongside the existing `DefaultIcon.png` convention. When one is found the builder:

- copies it into the build directory as `AppIcon.icon`, hashing its contents so an unchanged document doesn't force a rebuild
- adds it to the Xcode project's resources build phase as a `folder.iconcomposer.icon`, which is the input type `AssetCatalogCompiler.xcspec` accepts alongside `folder.assetcatalog`
- skips generating `AppIcon.appiconset` entirely, and removes a stale one from a previous build, so the asset catalog compiler is never handed two assets named "AppIcon"

`ASSETCATALOG_COMPILER_APPICON_NAME` is already `AppIcon` in `Titanium.xcodeproj`, so no build setting changes were needed.

Launch logos are still plain images. When a project has no `DefaultIcon.png`, the Icon Composer document is rendered to a 1024x1024 PNG using the `ictool` binary inside `Icon Composer.app` and that rendering becomes the source image, so an icon-only project still gets proper launch logos instead of the Titanium placeholder set.

Icon Composer documents are ignored with a warning on Xcode 25 and older, where `actool` doesn't understand them — those builds keep generating the app icon set from PNGs exactly as before. Projects without a `.icon` are completely unaffected.

## Testing

Verified against Xcode 26.6 / iOS 26.5 simulator with a classic app built by `ti create`, using an `.icon` produced by Icon Composer:

- `actool` receives both `Assets.xcassets` and `AppIcon.icon` with `--app-icon AppIcon`; the built `.app` contains `Assets.car`, the loose `AppIcon60x60@2x.png` / `AppIcon76x76@2x~ipad.png` legacy icons, and an `Info.plist` with `CFBundleIconName`
- the icon renders with Liquid Glass on the simulator home screen
- editing the document forces a rebuild; an unchanged document logs `No change, skipping` and reuses the copy in the build directory
- with `DefaultIcon.png` removed, the document is rendered via `ictool` and the launch logos are generated from it; a second build re-renders nothing

Also confirmed that `oxlint` reports nothing new and the existing PNG path is unchanged when no `.icon` is present.


iphone/cli/commands/_build.js
```diff
@@ -1906,6 +1910,12 @@ class iOSBuilder extends Builder {
 			// add the iOS specific default icon to the list of icons
 			this.defaultIcons.unshift(path.join(this.projectDir, 'DefaultIcon-ios.png'));
 
+			// the Icon Composer documents that, when present, replace the generated app icon set
+			this.defaultIconComposerIcons = [
+				path.join(this.projectDir, 'DefaultIcon-ios.icon'),
+				path.join(this.projectDir, 'DefaultIcon.icon')
+			];
```
hansemannn: We should likely only use `DefaultIcon-ios.icon`, as Android does not currently support this (and likely won't ever, as they have their own concept of layered icons).

iphone/cli/commands/_build.js
```diff
@@ -3706,6 +3716,41 @@ class iOSBuilder extends Builder {
 			return true;
 		}, this);
 
+		// if we have an Icon Composer app icon, add it to the project so `actool` compiles it
+		if (this.findIconComposerIcon()) {
```
hansemannn: Extract the find from the if-statement

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,187 @@ class iOSBuilder extends Builder {
...
+	 * found we skip generating the `AppIcon.appiconset` altogether.
+	 *
+	 * @returns {String|null} The path to the Icon Composer document or `null` if there isn't a usable one
+	 */
+	findIconComposerIcon() {
+		if (this.iconComposerIcon !== undefined) {
+			return this.iconComposerIcon;
+		}
+
+		const icon = (this.defaultIconComposerIcons || []).find(dir => fs.existsSync(dir) && fs.statSync(dir).isDirectory());
```
hansemannn: Why would `this.defaultIconComposerIcons` ever become `undefined`? I can't find a use case, yet. Also, check if `isDirectory` really needs the `existsSync` before or if it handles it already.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,187 @@ class iOSBuilder extends Builder {
...
+					walk(file, rel);
+				} else {
+					hash.update(rel);
+					hash.update(fs.readFileSync(file));
+				}
+			}
+		}(dir, ''));
+
+		return hash.digest('hex');
+	}
```
hansemannn: Check if we can use existing hashing utils, as we have many other assets that should do this already.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,187 @@ class iOSBuilder extends Builder {
...
+
+	/**
+	 * Renders an Icon Composer document to a 1024x1024 PNG using the `ictool` binary that ships
+	 * inside Icon Composer.app.
+	 *
+	 * @param {String} src The path to the Icon Composer document
+	 * @param {String} dest The path to write the rendered PNG to
+	 * @returns {Promise<void>}
+	 */
+	async renderIconComposerIcon(src, dest) {
```
hansemannn, line 6202: Make sure this is only called if the build is fresh, as this adds delays to the build time.

hansemannn (changes requested):
Great contribution, thank you! A first round of review attached, happy to see this land soon!

iphone/cli/commands/_build.js
abentley8: DefaultIcon-ios.icon only. Agreed on the Android reasoning; the cross-platform name would promise something the document can't deliver. FYI @designbymind your test above used DefaultIcon.icon, so this rename affects you when this is merged into a release 🙂

iphone/cli/commands/_build.js
abentley8: done

iphone/cli/commands/_build.js
abentley8: defaultIconComposerIcon is always set in validate(), which runs before every caller, so the guard was dead code. Gone. The separate existsSync is also gone: statSync(icon, { throwIfNoEntry: false })?.isDirectory() handles the missing-path case in one call.

iphone/cli/commands/_build.js
abentley8: Now uses the existing Builder.hash(), in the same hash→sort→join→hash shape as modulesNativeHash. Since .icon documents are directories, there's no single-file hash to reuse directly. The walk hashes each file and combines.

iphone/cli/commands/_build.js
abentley8: The ictool render was already gated on the document changing (or the rendering missing), and it only runs at all when the project has no DefaultIcon.png to build launch logos from. I've inverted the condition to make the skip explicit and added a No change, skipping trace so it's visible in build logs. On an unchanged incremental build, neither the copy nor the render executes

iphone/cli/commands/_build.js
designbymind: Good move on the naming conventions since it's an iOS-only feature. — I've tested the [latest version ](https://github.com/abentley8/titanium-sdk/blob/0c3fb348297d43383a622f01ff4c2be640029c6e/iphone/cli/commands/_build.js) (using `DefaultIcon-ios.icon`) and all is working great for both simulator and device (both iOS 27 latest betas). **Thanks again, this is great** 😃

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
+
+			if (!changed && alreadyRendered) {
+				// rendering shells out to `ictool`, so only pay for it when the document is new or
+				// has changed -- an unchanged document keeps the rendering from the previous build
+				this.logger.trace(`No change, skipping ${rendered.cyan}`);
+			} else {
+				await this.renderIconComposerIcon(dest, rendered);
+			}
+
+			if (fs.existsSync(rendered)) {
```
hansemannn: **Failed ictool render caches stale launch logos forever** (correctness, confirmed)

When ictool fails after the `.icon` document changed, the stale rendered `DefaultIcon.png` from the previous build is still used as the launch-logo source. Because `currentBuildManifest.files['AppIcon.icon']` was already updated with the new hash (line 6178), the next build sees `!changed && alreadyRendered` and never retries the render — the stale render is cached permanently.

Scenario: project has only `DefaultIcon-ios.icon`. Build 1 renders and generates launch logos. The user redesigns the `.icon`; in build 2 ictool exits non-zero, leaving build 1's PNG on disk. `fs.existsSync(rendered)` is true, so the old PNG is used; its hash/mtime match the previous manifest, so the old launch logos are kept while the warning falsely claims launch logos "will not be generated". Every subsequent build skips rendering entirely, so the app ships new app icons but a permanently outdated launch-screen logo with no error.

Suggestion: only record the `AppIcon.icon` manifest hash after the render succeeded (or clear it on failure).

iphone/cli/commands/_build.js
```diff
@@ -6096,6 +6309,16 @@ class iOSBuilder extends Builder {
 			};
 		}
 
+		// the Icon Composer document already covers every app icon size and appearance, so the
+		// only images left to produce are the launch logos
+		if (iconComposerIcon) {
+			const missingLaunchLogos = await this.processLaunchLogos(launchLogos, resourcesToCopy, defaultIcon, defaultIconChanged);
+			if (!missingLaunchLogos.length) {
```
hansemannn, line 6319: **Hand-tuned per-density icon PNGs silently dropped** (correctness, confirmed)

The early return removes the app-icon validation/copy loop and `writeAppIconSet` for Icon Composer projects, so user-supplied per-density icons in the `appIcons` map (`DefaultIcon-Dark.png`, `DefaultIcon-Tinted.png`, `DefaultIcon-60@2x.png`, …) are dropped silently — the old path validated, warned about, and copied each of them into the appiconset, and nothing in the new path even logs that they are being ignored.

Scenario: a project shipping hand-tuned dark/tinted variants adds a `DefaultIcon-ios.icon` document — the built app's dark/tinted icons silently switch to whatever actool renders from the `.icon`, and the developer only discovers the wrong icons on device or in App Store review.

Suggestion: at minimum log a warning when `appIcons` entries exist but are ignored because an Icon Composer document takes precedence.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
+			return this.iconComposerIcon;
+		}
+
+		const icon = this.defaultIconComposerIcon;
+
+		this.iconComposerIcon = null;
+
+		// Icon Composer documents are directories, so a stray file of the same name is not one
+		if (fs.statSync(icon, { throwIfNoEntry: false })?.isDirectory()) {
+			if (appc.version.lt(this.xcodeEnv.version, ICON_COMPOSER_MIN_XCODE_VER)) {
```
hansemannn, line 6102: **Xcode < 26 with an `.icon`-only project hard-fails the build** (correctness, confirmed)

On Xcode < 26, `findIconComposerIcon` returns null after only an "Ignoring …" warning. For a project whose sole icon source is the `.icon` document, `createAppIconSet` then falls through to the legacy path with no appicon PNGs and no `DefaultIcon`, and `generateAppIcons` hard-fails with "There are missing required app icons".

Scenario: a project adopts Icon Composer and deletes its `DefaultIcon.png`/`appicon-*.png` files; a teammate or CI runner on Xcode 25 gets a hard build failure whose warning never states the build will break or that `DefaultIcon.png` must be restored for older Xcode.

Suggestion: when the `.icon` is the only icon source and Xcode is too old, upgrade the warning to an actionable error (or explain the fallback requirement in the message).

iphone/cli/commands/_build.js
```diff
@@ -6096,6 +6309,16 @@ class iOSBuilder extends Builder {
 			};
 		}
 
+		// the Icon Composer document already covers every app icon size and appearance, so the
+		// only images left to produce are the launch logos
+		if (iconComposerIcon) {
```
hansemannn, line 6292: **Launch logos lose white-flattening for alpha-channel DefaultIcon.png** (correctness, confirmed)

The old path (previously lines ~6489-6505) flattened an alpha-channel `DefaultIcon.png` against white before `generateAppIcons` produced launch logos. The new Icon Composer path calls `generateAppIcons` directly on the unflattened source, and the new comment ("launch logos … are never flattened against a white background") misstates the prior behavior.

Scenario: an app with an alpha-channel `DefaultIcon.png` adds `DefaultIcon-ios.icon` on Xcode 26 — regenerated `LaunchLogo*.png` files now keep their transparency, so the launch screen shows the storyboard background through areas every previous build rendered as solid white, with no warning logged (the old path explicitly warned about the flattening).

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
+	 * @returns {String} A hash of the document's contents
+	 */
+	hashIconComposerIcon(dir) {
+		const hashes = [];
+
+		(function walk(dir, prefix) {
+			for (const name of fs.readdirSync(dir).sort()) {
+				const file = path.join(dir, name);
+				const rel = prefix ? prefix + '/' + name : name;
+				if (fs.statSync(file).isDirectory()) {
```
hansemannn: **Broken symlink inside the `.icon` document crashes the build** (correctness, confirmed)

`hashIconComposerIcon` calls `fs.statSync` on every directory entry without tolerating broken symlinks, so a dangling symlink inside the `.icon` document throws an unhandled ENOENT and aborts the build — even though `copyDirSync` (used to copy the same document at line 6182) explicitly skips broken symlinks, and `findIconComposerIcon` already uses `{ throwIfNoEntry: false }` at line 6119.

Suggestion: use `fs.statSync(file, { throwIfNoEntry: false })` (or lstat + skip) in the walk, matching `copyDirSync`'s tolerance.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
 	 * This may modify the list of resources to copy!
 	 * @param {Map.<String,FileInfo>} appIcons app icons to handle
 	 * @param {Map.<String,FileInfo>} launchLogos launch logos to process
 	 * @param {Map.<string,FileInfo>} resourcesToCopy plain files to handle
 	 */
 	async createAppIconSet(appIcons, launchLogos, resourcesToCopy) {
-		this.logger.info('Creating app icon set');
+		const iconComposerIcon = this.findIconComposerIcon();
+
+		if (iconComposerIcon) {
```
hansemannn, line 6246: **macOS targets skip required mac icon validation** (correctness, plausible)

`findIconComposerIcon` gates only on Xcode version, not build target, so for `macos`/`dist-macappstore` the early return skips the required mac icon lookup entries (16-512px, all `required: true`) and their `generateAppIcons` fallback that the old path always produced. The invariant "macOS targets always end up with the required mac icon sizes or the build errors" is not re-established anywhere in the composer path.

If actool does not derive the required mac icon sizes from the iOS-oriented `.icon` document, an archive could be submitted without required mac icons and fail App Store validation — worth verifying with a Mac Catalyst dist build.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
+
+		// an app icon set generated by a previous build would collide with the Icon Composer document
+		const appIconSetDir = path.join(this.buildDir, 'Assets.xcassets', 'AppIcon.appiconset');
+		if (await fs.exists(appIconSetDir)) {
+			this.logger.debug(`Removing ${appIconSetDir.cyan}`);
+			await fs.remove(appIconSetDir);
+		}
+
+		// the launch logos are generated by resizing a PNG, so make sure there's one to resize
+		if (!this.defaultIcons.some(icon => fs.existsSync(icon))) {
```
hansemannn: **ictool render runs even when the launch-screen storyboard is disabled** (correctness, confirmed)

`copyIconComposerIcon` renders the launch-logo source PNG via ictool without checking `enableLaunchScreenStoryboard`/`defaultLaunchScreenStoryboard`, even though `processLaunchLogos` immediately returns `[]` when the storyboard is disabled.

Scenario: a project with `<enable-launch-screen-storyboard>false</enable-launch-screen-storyboard>` (or a custom LaunchScreen.storyboard) and no `DefaultIcon.png` shells out to ictool on every build where the `.icon` changed; if ictool fails, the user gets misleading "Failed to render … Launch logos will not be generated" warnings for an artifact the build never uses.

Suggestion: skip the render when launch logos won't be generated.

iphone/cli/commands/_build.js
```diff
@@ -3706,6 +3715,42 @@ class iOSBuilder extends Builder {
 			return true;
 		}, this);
 
+		// if we have an Icon Composer app icon, add it to the project so `actool` compiles it
+		const iconComposerIcon = this.findIconComposerIcon();
+		if (iconComposerIcon) {
```
hansemannn: **pbxproj injection block duplicates the Settings.bundle block** (cleanup)

This `AppIcon.icon` injection block is a near-verbatim copy of the `Settings.bundle` block directly above (lines 3683-3713), differing only in name and `lastKnownFileType`. Consider extracting a shared helper (e.g. `addResourceFolderToXcodeProject(xobjs, resourcesBuildPhase, resourcesGroup, name, fileType)`) so the two 30-line PBXFileReference/PBXBuildFile/build-phase/group sequences don't have to be maintained in lockstep.

iphone/cli/commands/_build.js
```diff
@@ -6096,6 +6309,16 @@ class iOSBuilder extends Builder {
...
 		}
 
+		// the Icon Composer document already covers every app icon size and appearance, so the
+		// only images left to produce are the launch logos
+		if (iconComposerIcon) {
+			const missingLaunchLogos = await this.processLaunchLogos(launchLogos, resourcesToCopy, defaultIcon, defaultIconChanged);
+			if (!missingLaunchLogos.length) {
+				return;
+			}
+			return util.promisify(this.generateAppIcons).bind(this)(missingLaunchLogos);
```
hansemannn: **Duplicated `util.promisify(this.generateAppIcons).bind(this)` expression** (cleanup)

This inline expression duplicates the `boundGenerateAppIcons` binding created later in the same function (line 6513). Consider hoisting `const boundGenerateAppIcons = util.promisify(this.generateAppIcons).bind(this);` above the Icon Composer branch and reusing it in both paths, so the two invocation sites can't drift.

hansemannn (commented):
Additional (automated) review: 7 correctness findings and 3 cleanups, each independently verified against the diff. The correctness issues cluster around the incremental-build caching in `copyIconComposerIcon` (manifest hash recorded before the ictool render succeeds) and the early return in `createAppIconSet` (skips the legacy validation/copy path and its safety nets). Details inline.

iphone/cli/commands/_build.js
```diff
@@ -6055,27 +6100,195 @@ class iOSBuilder extends Builder {
...
+		}
+	}
+
 	/**
 	 * This may modify the list of resources to copy!
 	 * @param {Map.<String,FileInfo>} appIcons app icons to handle
 	 * @param {Map.<String,FileInfo>} launchLogos launch logos to process
 	 * @param {Map.<string,FileInfo>} resourcesToCopy plain files to handle
 	 */
 	async createAppIconSet(appIcons, launchLogos, resourcesToCopy) {
```
hansemannn, line 6243: **New Icon Composer pipeline ships with no tests** (test-coverage)

The diff touches only `_build.js` — no test file is added or updated, while AGENTS.md's testing guidelines call for tests with behavior changes. The new pipeline has several incremental-build paths worth covering (Xcode version gating in `findIconComposerIcon`, directory hashing/rebuild forcing in `copyIconComposerIcon`, appiconset deletion, ictool fallback rendering, and the reworked flatten logic in `createAppIconSet`), any of which could regress undetected.

iphone/cli/commands/_build.js
abentley8: Fixed, though the code moved: hashIconComposerIcon() is now hashIconComposerDocument() in cli/lib/icon-composer.js, extracted so it's unit-testable. Broken symlinks are skipped there via throwIfNoEntry: false, matching copyDirSync, with a test asserting it doesn't throw and the hash is unchanged.

iphone/cli/commands/_build.js
abentley8: Fixed in 0933cd18 — the AppIcon.icon manifest hash is now only recorded once the render succeeded, so a failed run is retried on the next build instead of being cached.

iphone/cli/commands/_build.js
abentley8: Fixed — now warns and lists them: Ignoring 2 app icons because DefaultIcon-ios.icon supplies every icon size and appearance: followed by each path.

iphone/cli/commands/_build.js
abentley8: Fixed — the render is skipped unless enableLaunchScreenStoryboard and defaultLaunchScreenStoryboard are both set, since processLaunchLogos() returns immediately otherwise.

iphone/cli/commands/_build.js
abentley8: Improved — when the document is ignored on older Xcode and no DefaultIcon.png exists, the warning now names the file to restore rather than letting it fail later with "missing required app icons".

iphone/cli/commands/_build.js
abentley8: Extracted into an addResourceFolder(name, fileType) helper used by both Settings.bundle and AppIcon.icon.

iphone/cli/commands/_build.js
abentley8: Hoisted above the Icon Composer branch and reused in both paths.

iphone/cli/commands/_build.js
abentley8: Added iphone/cli/tests/test-icon-composer.js — 8 tests. This needed the helpers extracted to cli/lib/icon-composer.js, since _build.js builds a CLI-bound instance at import time and can't be loaded standalone; that's presumably why there were no iOS CLI tests. Note two pre-existing failures in android/cli/tests/test-android-manifest.js on main — @xmldom/xmldom 0.9 requires a mimeType argument that createManifestXmlDomDocument() doesn't pass. Unrelated to this PR, happy to fix separately.

iphone/cli/commands/_build.js
abentley8: Partly verified. A -T macos build with a document succeeds and takes the Icon Composer path: actool emits AppIcon.icns into Contents/Resources alongside layered Assets.car renditions, with CFBundleIconName set, and no missing-icon errors. What I couldn't check is App Store validation — that needs a dist-macappstore build with a Mac App Store profile I don't have for a throwaway bundle id. So I haven't re-established the "required mac sizes or the build errors" invariant, I've only shown the build doesn't obviously break. If you'd rather be safe, I can keep the mac lookup entries active for macOS targets so the old validation still runs alongside the document.

iphone/cli/commands/_build.js
abentley8: You're right about the comment — it described the new behavior as if it were the old one. Fixed.

On the flattening itself I went the other way, and I want to lay out why, because I tried your suggestion first and it doesn't hold.

Restoring the flatten puts the build into a permanent loop. flattenIcon() rewrites build/iphone/DefaultIcon.png and records that file's stats in the build manifest. The rewrite changes its mtime, so the next build's defaultIconChanged comparison fails, which regenerates the launch logos, which triggers the flatten again. I measured it — three consecutive no-op builds each re-flattened and each logged Forcing rebuild: launch logos changed since last build. Moving the flatten before processLaunchLogos() doesn't help either; that only fixes the manifest entry the launch logos are compared against, not the mtime churn driving it.

The old path avoids this only because its flatten is gated on missingIcons.length !== 0, and those are missing app icons — they stop being missing once generated, so the loop terminates. In the Icon Composer path there are no app icons to generate, so the only thing in missingIcons is launch logos, and those never stop being missing while the manifest keeps getting invalidated. The gate isn't reusable here.

So the current behavior is: alpha is kept, and the build warns about it —

`[WARN] The default icon "DefaultIcon.png" contains an alpha channel, which is kept when generating launch logos
`
I'd argue this is also the better rendering. Flattening against white exists because App Store icons must be opaque; launch logos have no such requirement and composite over default-background-color, so white was a side effect of sharing the app-icon code path rather than a deliberate choice. Your scenario is real, though — an app that previously got white boxes will see the storyboard color instead, so it is a visible change and shouldn't be silent.

If you'd still prefer flattening preserved, the way to do it without the loop is to make the flatten idempotent: gate it on the source file's content hash rather than on defaultIconChanged, so an already-current flattened file is never rewritten. Happy to do that instead — just say which you'd rather have.

hansemannn (approved):
LGTM — both rounds of review feedback have been addressed (manifest-hash ordering, symlink tolerance, storyboard gating, icon-drop warning, extracted helpers, and tests). Remaining notes from the final pass are minor and non-blocking.

PR #14545, opened 2026-08-10
feat(ios): run device builds via CoreDevice (devicectl), drop node-ios-device

hansemannn:
## What this is

Rebuilds the entire iOS *device* workflow — detection, app install, launch and log streaming — on top of Apple's modern CoreDevice stack (`xcrun devicectl`, Xcode 15+), replacing the legacy MobileDevice.framework path that was driven by the native `node-ios-device` module. The heavy lifting lives in [ioslib 6.0.0](https://github.com/tidev/ioslib/pull/261); the SDK's install hook shrinks to Titanium-specific orchestration.

## Why this is a big deal

**It makes on-device development work again.** iOS 17+ devices connect exclusively through CoreDevice/RemoteXPC and are *invisible* to the old MobileDevice API — even over USB. On modern devices, `ti build -T device` could neither detect, install, nor stream logs. This PR restores the full workflow.

**It removes the last native binary from the CLI toolchain.** `node-ios-device` was a C++ addon shipped as prebuilt binaries per Node ABI. That meant:
- device builds silently broke whenever the installed Node version had no matching binary (the long-standing "empty `ti info` / no devices" failure class);
- the SDK packaging pipeline had to compile and bundle binaries for multiple Node versions;
- every new Node or macOS release was a potential breakage.

All of that is gone — ~120 packages dropped from the dependency tree, no compile step, no ABI matrix, smaller SDK zip.

**The developer experience got better on the way:**
- **Wireless device builds**: devices on the same network are detected, installed to and launched with live logs — no cable required.
- **Auto-launch with attached console**: previously you had to launch the app manually after install; now it launches by itself and streams `console.log()` / `Ti.API.*` output directly (leveled and colorized like simulator output).
- **Lock-aware**: if the device is locked, the CLI waits and launches as soon as you unlock.
- A failed device detection no longer nukes the entire `ti info` result (certs, Xcode, sims).

## Notes

- Requires **ioslib 6.0.0** and **Xcode 15+** command line tools (already the floor for current iOS development).
- Simulator and macOS workflows are untouched — they never used `node-ios-device`.
- The old TCP log relay (`logPort`) is superseded by devicectl's console capture; the log-server infrastructure in the app is unaffected.

## Testing

Verified live against a physical iPhone (iOS 27) over Wi-Fi: detection, lock state, install (~6s), auto-launch with full console capture (stdout + stderr/NSLog including multi-line messages), locked-device wait → unlock → launch, and clean app-quit propagation. See the [ioslib PR](https://github.com/tidev/ioslib/pull/261) for the library-level test protocol.


cb1kenobi (approved):
Tested on device, works AWESOME! Love the "Device <name> is locked, the app will launch as soon as it is unlocked" feature. Tested locked/unlocked state. Tested killing the app on device and CLI correctly exits. Tested log messages. This is pretty fantastic.

`ti info` looks good, both report and json output.

PR #14550, opened 2026-08-12
fix(android): make TabGroup opening lifecycle-safe

macCesar:
## Summary

Opening a `TabGroup` can get stuck on the splash screen when its owner Activity is recreated while `open()` is still pending. `KrollPromise.create()` dispatches the work asynchronously, so resolving the owner inside `handleOpen()` can return a replacement or an unrelated Activity, and the promise never settles.

This captures the owner when `open()` is invoked and then waits a bounded time for a usable replacement belonging to the same Titanium window. The existing root-Activity fallback stays. If the original owner is really closing, the open promise rejects instead of returning silently. Delayed work is generation-guarded and cancelled after success, failure or `close()`, and `TiSafeDisplay` now completes once and has a timeout, so an attach callback cannot leave the promise pending forever. Calling `close()` during a pending open settles both promises and lets the same proxy be opened again.

No public API or Window `close` event semantics change.

## Steps to reproduce

Use a `TabGroup` as the root UI and trigger an Activity recreation while its asynchronous `open()` is pending. Assigning a different interface style during startup is one way to reproduce it:

```js
Ti.UI.overrideUserInterfaceStyle = Ti.UI.USER_INTERFACE_STYLE_LIGHT;

const tabGroup = Ti.UI.createTabGroup();
const win = Ti.UI.createWindow({ title: 'Tab One' });
tabGroup.addTab(Ti.UI.createTab({ title: 'One', window: win }));
tabGroup.open();
```

**Expected:** the `TabGroup` opens and the promise settles.

**Actual:** the app can remain on the splash screen with the open promise pending.

## Test plan

- [x] Root `TabGroup` opens after its owner Activity is recreated
- [x] Opening from a finishing owner rejects instead of using an unrelated Activity
- [x] `open()` followed immediately by `close()` settles both promises, and the same proxy reopens
- [x] Verified on Android 10 (API 29), Android 13 (API 33) and a physical Android 16 (API 36)

### DIFF
--- a/android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
+++ b/android/modules/ui/src/java/ti/modules/titanium/ui/TabGroupProxy.java
@@ -26,6 +26,7 @@ import androidx.core.view.WindowCompat;
 import androidx.core.view.WindowInsetsControllerCompat;
 
 import org.appcelerator.kroll.KrollDict;
+import org.appcelerator.kroll.KrollPromise;
 import org.appcelerator.kroll.annotations.Kroll;
 import org.appcelerator.kroll.common.Log;
 import org.appcelerator.titanium.TiActivity;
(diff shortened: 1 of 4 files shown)


macCesar:
Ping on this one. I re-checked main today and the bug is still there: `TabGroupProxy.handleOpen()` resolves the owner with `getActivity()` inside the async handler (line 394) and returns silently when that Activity is finishing or destroyed, so the open promise never settles and the app stays on the splash screen.

The branch still merges cleanly and all checks are green. Happy to rebase or split it into smaller commits if that makes it easier to review.

PR #14551, opened 2026-08-12
fix(android): restore ActionBar up navigation after recreation

macCesar:
## Summary

An Activity recreation replaces the native ActionBar and its Activity proxy. Titanium keeps the Window proxy alive through that temporary destruction, but the ActionBar proxy was owned only by the old Activity proxy, so the replacement Activity consulted a new proxy with no up-navigation callback.

This moves ownership to the surviving Window. Each replacement Activity borrows the proxy and rebinds it to the new native ActionBar, and a real Window close releases it, so the up affordance and `onHomeIconItemSelected` keep referring to the same proxy. Other ActionBar properties and app-provided `supportToolbar` instances are outside this fix. There are no public API changes.

## Steps to reproduce

```js
const win = Ti.UI.createWindow({ title: 'Child' });

win.addEventListener('open', function once() {
	win.removeEventListener('open', once);
	win.activity.actionBar.displayHomeAsUp = true;
	win.activity.actionBar.onHomeIconItemSelected = () => {
		Ti.API.info('HOME ICON SELECTED');
	};
});

win.open();
```

With that window open, change the interface style:

```js
Ti.UI.overrideUserInterfaceStyle = Ti.UI.USER_INTERFACE_STYLE_DARK;
```

**Expected:** the up arrow remains visible and still invokes the callback.

**Actual:** the replacement Activity uses a different ActionBar proxy, so the arrow and native callback path are lost.

## Test plan

- [x] After a native `Activity.recreate()`, the replacement still shows the up affordance and still invokes `onHomeIconItemSelected`
- [x] The callback fires exactly once before and once after recreation, with no duplicate
- [x] A real Window close still releases the Window-owned ActionBar proxy
- [x] Verified on Android 10 (API 29), Android 13 (API 33) and a physical Android 16 (API 36)

The recreation check uses an isolated child Activity because changing the shared TiMocha root Activity resets the test runner. A Hyperloop harness invokes the native recreation without adding a test-only API to Titanium.

### DIFF
--- a/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/TiBaseActivity.java
@@ -957,6 +957,7 @@ public abstract class TiBaseActivity extends AppCompatActivity implements TiActi
 				finish();
 			}
 		}
+		activityProxy.restoreActionBar();
 	}
 
 	/**
--- a/android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
+++ b/android/titanium/src/java/org/appcelerator/titanium/proxy/ActionBarProxy.java
@@ -34,15 +34,25 @@ public class ActionBarProxy extends KrollProxy
 
 	// Weak references so a proxy kept alive by JS cannot leak a destroyed activity
 	// or its view hierarchy (the toolbar's context is the activity).
-	private final WeakReference<AppCompatActivity> activityRef;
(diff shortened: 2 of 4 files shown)


macCesar:
This one is still reproducible on main: `ActionBarProxy` is owned by the Activity proxy (`ActivityProxy.java:58`, created at line 292), so after a recreation the replacement Activity gets a fresh proxy with no up-navigation callback and the arrow stops working.

Checks are green and the branch merges cleanly. Let me know if you want it rebased or broken up.

PR #14555, opened 2026-08-22
chore: remove unused properties from tiapp.xml

### LINKED ISSUE #14554: Clean up tiapp.xml

### Description

Currently our default tiapp.xml has these fields:
https://github.com/tidev/titanium-sdk/blob/main/templates/app/default/template/tiapp.xml#L6-L9
```
	<publisher>not specified</publisher>
	<url>not specified</url>
	<description></description>
	<copyright>not specified</copyright>
```

that are not really used besides `Ti.App.url`, `Ti.App.description`,.. but not compiled into the apps itself or visible in the store. Not sure if those fields are really used inside apps by devs at all.

Should we still keep them or reduce the XML by removing them so people don't think they are used in the stores?

### Solution

remove the from the templates, deprecate the calls and return empty strings until we remove them (or remove them in 14.0.0)

### Alternatives

keep everything as it is. Just some XML entries that won't hurt.

### Platforms

Android & iOS

m1ga:
fixes #14554 

tiapp.xml fields description, url, publisher, copyright and not used in the app or store pages as some people thought. They are purely properties in your code. 

```js
console.log(Ti.App.description)
console.log(Ti.App.url)
console.log(Ti.App.publisher)
console.log(Ti.App.copyright)
```

outputs:
```
[WARN]  AppModule: (main) [65,65] Ti.App.description has been deprecated in 14.0.0 and always returns an empty string. The <description/> element in tiapp.xml is no longer used.
[INFO]  
[WARN]  AppModule: (main) [0,65] Ti.App.url has been deprecated in 14.0.0 and always returns an empty string. The <url/> element in tiapp.xml is no longer used.
[INFO]  
[WARN]  AppModule: (main) [0,65] Ti.App.publisher has been deprecated in 14.0.0 and always returns an empty string. The <publisher/> element in tiapp.xml is no longer used.
[INFO]  
[WARN]  AppModule: (main) [0,65] Ti.App.copyright has been deprecated in 14.0.0 and always returns an empty string. The <copyright/> element in tiapp.xml is no longer used.
[INFO]  
```

If people still want to manage them in their tiapp.xml they can use properties: `<property name="copyright" type="string">(c) 2026</property>` ( https://titaniumsdk.com/api/titanium/app/properties.html ) or Alloy.CFG


cb1kenobi (approved):
Wow, this is great. It touched more stuff than I would have guessed. :)

Merge when ready.

android/modules/app/src/java/ti/modules/titanium/app/AppModule.java
```diff
@@ -84,34 +84,47 @@ public String getVersion()
 		return appInfo.getVersion();
 	}
 
+	// TODO: remove in 14.2.0
```
hansemannn: We don't remove APIs in non-major versions. If we deprecate it in 14.0.0, we can remove it in 15.0.0.

hansemannn (changes requested):
In general: Is there a real reason to remove it? I don't see it adding much complexity or harm - there are much worse cases that do that.

android/modules/app/src/java/ti/modules/titanium/app/AppModule.java
m1ga: I can change the note. Some devs thought those fields would be used in the Store notes but that is not the case. It just looks too important to be there :smile:  So I'm not sure if they should be primary entries in the XML. 
And since we have custom properties already  people could use those if they really need them in the XML file.

hansemannn: Isn't it rather a documentation thing then? Letting them know what it does and what it doesn't.

m1ga: that would be another option. At the moment it just says e.g. `The copyright of this application.`
https://titaniumsdk.com/guide/Titanium_SDK/Titanium_SDK_Guide/Appendices/tiapp.xml_and_timodule.xml_Reference.html#copyright
or `The description of this application.` but not that it is not used anywhere in the build process or store description.

We can keep it as it is. This was my second option in the issue description :smile:  "keep everything as it is. Just some XML entries that won't hurt.". No issue in closing this and I make a PR to add some more context in the docs :+1:

PR #14558, opened 2026-08-24
feat(android): merge module build.gradle repositories

m1ga:
**Issue:**
Currently when you build a module that uses an external Maven repo like `maven { url = uri('https://jitpack.io') }` you have to ask the user to manually add those to the app too. Otherwise it will fail with something like this:

```
[ERROR] [GRADLE]
[ERROR] [GRADLE] FAILURE: Build failed with an exception.
[ERROR] [GRADLE]
[ERROR] [GRADLE] Configuration cache state could not be cached: field `__runtimeDependenciesNavigationFiles__` of task `:app:processDebugNavigationResources` of type `com.android.build.gradle.internal.tasks.ProcessNavigationXmlTask`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
[ERROR] [GRADLE] > Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
[ERROR] [GRADLE]    > Could not find com.github.pedroSG94.RootEncoder:library:2.7.5.
[ERROR] [GRADLE]      Searched in the following locations:
[ERROR] [GRADLE]        - https://dl.google.com/dl/android/maven2/com/github/pedroSG94/RootEncoder/library/2.7.5/library-2.7.5.pom
[ERROR] [GRADLE]        - https://repo.maven.apache.org/maven2/com/github/pedroSG94/RootEncoder/library/2.7.5/library-2.7.5.pom
[ERROR] [GRADLE]        - file:/home/miga/.titanium/mobilesdk/linux/13.4.0.GA/android/m2repository/com/github/pedroSG94/RootEncoder/library/2.7.5/library-2.7.5.pom
[ERROR] [GRADLE]        - file:/tmp/ti-android-module-build-2026724-115272-yos8ep.klq2m/ti.twitch/modules/android/ti.twitch/1.0.0/m2repository/com/github/pedroSG94/RootEncoder/library/2.7.5/library-2.7.5.pom
[ERROR] [GRADLE]      Required by:
[ERROR] [GRADLE]          project ':app' > ti:twitch:1.0.0
[ERROR] [GRADLE]
[ERROR] [GRADLE] * Try:
[ERROR] [GRADLE] > Run with --stacktrace option to get the stack trace.
[ERROR] [GRADLE] > Run with --info or --debug option to get more log output.
[ERROR] [GRADLE] > Run with --scan to get full insights from a Build Scan (powered by Develocity).
[ERROR] [GRADLE] > Get more help at https://help.gradle.org.
[ERROR] [GRADLE]
[ERROR] [GRADLE] BUILD FAILED in 1s
```

**Fix:**
This PR will automatically gather the repos from modules and add them during the build process.

You can add `<merge-module-maven-repositories>false</merge-module-maven-repositories>` to your `<android>` tiapp.xml node to disable that for security reasons. I've kept it opt-out because the benefits of not editing a build.gradle file when you add a module is easier for most people (nobody reads the module README files :smile: ).

This also helps when you build the `example/app.js` in your modules and you simply can run `ti build - android` without adding another example build.gradle file.

**Test**
Checkout https://github.com/m1ga/ti.twitch and go to the android folder and run `ti build -p android`. This will compile the module + the demo app. Without this PR it will fail. With `ti build -p android --sdk 14.0.0` it will start the app.

### DIFF
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2163,7 +2163,7 @@ class AndroidBuilder extends Builder {
 		await fs.writeFile(path.join(projectDirPath, 'build.gradle'), buildGradleContent);
 	}
 
-	async fetchMergeModuleMavenRepositoriesSetting() {
+	async shouldMergeModuleMavenRepositories() {
 		// Determines if Maven repository URLs defined by modules should be added to the app project.
 		// This is enabled by default. App developers can opt out via the below "tiapp.xml" setting:
 		//   <android><merge-module-maven-repositories>false</merge-module-maven-repositories></android>
@@ -2185,7 +2185,8 @@ class AndroidBuilder extends Builder {
 				return String(settingValue).trim().toLowerCase() !== 'false';
 			}
 		} catch (err) {
-			this.logger.warn(`Failed to read "merge-module-maven-repositories" setting from "tiapp.xml" file. Reason:\n${err}`);
+			this.logger.error(`Failed to read "merge-module-maven-repositories" setting from "tiapp.xml" file. Reason:\n${err}`);
+			throw err;
 		}
 		return true;
 	}
@@ -2208,7 +2209,7 @@ class AndroidBuilder extends Builder {
 		this.libDependencyStrings.push(`org.appcelerator:titanium:${this.titaniumSdkVersion}`);
 
 		// Check if the app developer opted out of adding module defined Maven repositories to the app project.
-		const isModuleRepoMergeEnabled = await this.fetchMergeModuleMavenRepositoriesSetting();
+		const isModuleRepoMergeEnabled = await this.shouldMergeModuleMavenRepositories();
 
 		// Process all Titanium modules referenced by the Titanium project.
 		for (const nextModule of this.modules) {
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2256,7 +2256,7 @@ class AndroidBuilder extends Builder {
 							`Skipping Maven repositories from module "${nextModule.manifest.moduleid}". `
 							+ `Was disabled via "merge-module-maven-repositories" setting in "tiapp.xml" file.`);
 					}
-				} else if (await fs.exists(mavenReposFilePath)) {
+				} else if (fs.existsSync(mavenReposFilePath)) {
 					try {
 						const repoUrls = JSON.parse(await fs.readFile(mavenReposFilePath, 'utf8'));
 						if (Array.isArray(repoUrls)) {
--- a/android/cli/commands/_build.js
+++ b/android/cli/commands/_build.js
@@ -2163,6 +2163,33 @@ class AndroidBuilder extends Builder {
 		await fs.writeFile(path.join(projectDirPath, 'build.gradle'), buildGradleContent);
 	}
 
+	async fetchMergeModuleMavenRepositoriesSetting() {
+		// Determines if Maven repository URLs defined by modules should be added to the app project.
+		// This is enabled by default. App developers can opt out via the below "tiapp.xml" setting:
+		//   <android><merge-module-maven-repositories>false</merge-module-maven-repositories></android>
+		// Note: Must be read from the XML directly since the tiapp parser only reads known <android/> tags.
+		try {
+			const fileContent = await fs.readFile(path.join(this.projectDir, 'tiapp.xml'), 'utf8');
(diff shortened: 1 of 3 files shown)


android/cli/commands/_build.js
```diff
@@ -2215,6 +2245,33 @@ class AndroidBuilder extends Builder {
 				// This supports dependency management to avoid library version conflicts.
 				this.mavenRepositoryUrls.push(pathToFileURL(repositoryDirPath).href);
 				this.libDependencyStrings.push(dependencyString);
+
+				// Add the module's custom Maven repository URLs (if any) to the app project so that the
+				// module's remote dependencies can be resolved without the app developer having to
+				// declare these repositories in the app's "platform/android/build.gradle" file manually.
+				const mavenReposFilePath = path.join(nextModule.modulePath, 'ti.maven.repos.json');
+				if (!isModuleRepoMergeEnabled) {
+					if (await fs.exists(mavenReposFilePath)) {
```
cb1kenobi: Use `existsSync`, it's faster plus I want to rip out `fs-extra`.

```suggestion
					if (fs.existsSync(mavenReposFilePath)) {
```

android/cli/commands/_build.js
```diff
@@ -2215,6 +2245,33 @@ class AndroidBuilder extends Builder {
...
+						this.logger.info(
+							`Skipping Maven repositories from module "${nextModule.manifest.moduleid}". `
+							+ `Was disabled via "merge-module-maven-repositories" setting in "tiapp.xml" file.`);
+					}
+				} else if (await fs.exists(mavenReposFilePath)) {
```
cb1kenobi: Same here:
```suggestion
				} else if (fs.existsSync(mavenReposFilePath)) {
```

android/cli/commands/_buildModule.js
```diff
@@ -718,6 +718,14 @@ export class AndroidModuleBuilder extends Builder {
 			dest.append(fs.createReadStream(filePath), { name: zipEntryName });
 		});
 
+		// Add module's custom Maven repository list JSON file to the archive if it exists.
+		// The app build system uses this to add the module's repositories (such as jitpack) to the
+		// app project so its remote dependencies can be resolved without manual app-side changes.
+		const mavenReposFilePath = path.join(this.buildModuleDir, 'build', 'outputs', 'ti.maven.repos.json');
+		if (await fs.exists(mavenReposFilePath)) {
```
cb1kenobi: Same here:

```suggestion
		if (fs.existsSync(mavenReposFilePath)) {
```

cb1kenobi (approved):

cb1kenobi (approved):

android/cli/commands/_build.js
```diff
@@ -2163,6 +2163,33 @@ class AndroidBuilder extends Builder {
...
+							settingValue = appc.xml.getValue(childNode);
+						}
+					});
+				}
+			});
+			if (settingValue !== null) {
+				return String(settingValue).trim().toLowerCase() !== 'false';
+			}
+		} catch (err) {
+			this.logger.warn(`Failed to read "merge-module-maven-repositories" setting from "tiapp.xml" file. Reason:\n${err}`);
```
hansemannn: Shouldn't this cause a failure state?

android/cli/commands/_build.js
```diff
@@ -2180,6 +2207,9 @@ class AndroidBuilder extends Builder {
 		this.mavenRepositoryUrls.push(pathToFileURL(path.join(this.platformPath, 'm2repository')).href);
 		this.libDependencyStrings.push(`org.appcelerator:titanium:${this.titaniumSdkVersion}`);
 
+		// Check if the app developer opted out of adding module defined Maven repositories to the app project.
+		const isModuleRepoMergeEnabled = await this.fetchMergeModuleMavenRepositoriesSetting();
```
hansemannn: Variable naming: You declare a boolean property, but the function suggests an action, not a check

hansemannn (changes requested):
Two small improvements!

android/cli/commands/_build.js
m1ga: good point. Since the XML should be correct at this point it should really fail if something is still wrong as the user clearly wants to use `merge-module-maven-repositories`, so an error might be better here instead of continuing.

android/cli/commands/_build.js
m1ga: changed it to `shouldMergeModuleMavenRepositories()`

hansemannn (approved):

PR #14564, opened 2026-08-28
Point the docs regen at titaniumsdk.com

cb1kenobi:
Part of the titaniumsdk.com rebuild. The API docs pipeline gains a second destination, `tidev/titanium-www`, while the existing one keeps running.

### What changes

- **Adds** a dispatch to `tidev/titanium-www` carrying this repo's name in the payload. The existing dispatch sent nothing, so the regen had to check out all 17 source repos at full history on every trigger to work out what had changed. The new one fetches only this repo, shallow, sparse to `apidoc/`.
- **Keeps** the existing dispatch to `tidev/titanium-docs`, unchanged, as a second job. Both pipelines run in parallel so the site being served today stays current while the replacement is proven; backing out is merging a revert.
- Fires on a push touching `apidoc/**` and on a published release, rather than only on a merged pull request — a direct push is no longer missed.
- The new dispatch lives in a reusable workflow in `titanium-www`, so the payload shape is defined once for all 17 source repos rather than copied into each.

### Why `notify` does not depend on `lint`

Gating the dispatch on the lint job means an unrelated tooling problem stops the docs updating and nothing says so. `ti.coremotion` cannot run `npm ci` at all today — its lockfile is out of sync with its `package.json` — and with the gate in place its docs would silently never regenerate.

Malformed apidoc still cannot ship: the compile in `titanium-www` fails on unresolved references and schema violations and commits nothing, which is the right place for that check since it can see the whole corpus.

### Cleanups

- `fetch-depth: 0` removed — a lint does not need full history.
- `if: steps.node-cache.outputs.cache-hit != 'true'` removed; there is no `node-cache` step in this file, so the condition never did anything. Replaced with real npm caching.
- Node moves to 24. Verified `npm ci` and `lint:docs` on a current Node against repos previously pinned to 16.
- Adds a `pull_request` trigger so `apidoc/` changes get linted before merge. `notify` is skipped on pull requests — a proposed change gets linted, not published.

`workflow_dispatch` and the `github.repository_owner == 'tidev'` guard are both kept.

The receiving end is verified — `titanium-www` has already compiled this repo's apidoc end to end (419 types) from a manual dispatch.

### DIFF
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -64,7 +64,7 @@ jobs:
     name: Notify titanium-docs (legacy)
     steps:
       - name: Repository dispatch
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@v4
         with:
           event-type: regen-api-docs
           repository: tidev/titanium-docs
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -34,9 +34,14 @@ jobs:
   # The dispatch itself lives in titanium-www so the payload shape is defined
   # once for all 17 source repos instead of copied into each.
   #
+  # Deliberately independent of the lint job. Gating the dispatch on it means an
+  # unrelated tooling problem stops the docs updating and nothing says so --
+  # ti.coremotion cannot run `npm ci` at all today because its lockfile is out of
+  # sync with its package.json. Genuinely malformed apidoc still cannot ship: the
+  # compile in titanium-www fails on it and commits nothing.
+  #
   # Skipped on pull_request: a proposed change gets linted, not published.
   notify:
-    needs: lint
     if: github.event_name != 'pull_request'
     uses: tidev/titanium-www/.github/workflows/notify-api-docs.yml@main
     secrets:
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -19,7 +19,7 @@ jobs:
     # Keeps forks from dispatching at the tidev sites.
     if: github.repository_owner == 'tidev'
     runs-on: ubuntu-latest
-    name: Lint
+    name: Trigger Regen
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
--- a/.github/workflows/regen-docs.yml
+++ b/.github/workflows/regen-docs.yml
@@ -1,41 +1,38 @@
 name: Regen Docs
+
+# Tells titaniumsdk.com to recompile the SDK API docs.
+#
+# Fires on a change to apidoc/ and on a published release, rather than on every
+# merged pull request, so a direct push to main is not missed.
 on:
   workflow_dispatch:
-  pull_request:
-    types: [ closed ]
-    paths: [ 'apidoc/**' ]
-    branches:
-    - master
-    - main
+  push:
+    branches: [main]
+    paths: ['apidoc/**']
+  release:
+    types: [published]
 
 jobs:
-  regen:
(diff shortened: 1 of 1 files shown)


.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
-      with:
-        fetch-depth: 0
+      - uses: actions/checkout@v6
 
-    - name: Setup node
-      uses: actions/setup-node@v6
-      with:
-        node-version: '24'
-        registry-url: 'https://registry.npmjs.org'
+      - uses: actions/setup-node@v4
```
cb1kenobi: Use the latest action:
```suggestion
      - name: Setup node
      - uses: actions/setup-node@v7
```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
+  push:
+    branches: [main]
+    paths: ['apidoc/**']
+  release:
+    types: [published]
 
 jobs:
-  regen:
-    if: github.repository_owner == 'tidev' && (github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch')
+  lint:
```
cb1kenobi: The name of the job is `regen`, not `lint`:

```suggestion
  regen:
```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
     runs-on: ubuntu-latest
-    name: Trigger Regen
-
+    name: Lint
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v6
-      with:
-        fetch-depth: 0
+      - uses: actions/checkout@v6
```
cb1kenobi: Use the latest action version and give it a name:
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v7
```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
 
-    - name: Setup node
-      uses: actions/setup-node@v6
-      with:
-        node-version: '24'
-        registry-url: 'https://registry.npmjs.org'
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '24'
```
cb1kenobi: Use the latest Node.js version:
```suggestion
          node-version: '26'
```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
+          cache: npm
 
-    - name: Install dependencies
-      run: npm ci
-      if: steps.node-cache.outputs.cache-hit != 'true'
+      - run: npm ci
```
cb1kenobi:
```suggestion
      - name: Install dependencies
        run: npm ci

```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
+      - run: npm run lint:docs
```
cb1kenobi:
```suggestion
      - name: Lint
        run: npm run lint:docs
```

.github/workflows/regen-docs.yml
```diff
@@ -1,41 +1,65 @@
...
 jobs:
-  regen:
-    if: github.repository_owner == 'tidev' && (github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch')
+  lint:
+    # Keeps forks from dispatching at the tidev sites.
+    if: github.repository_owner == 'tidev'
     runs-on: ubuntu-latest
-    name: Trigger Regen
-
+    name: Lint
```
cb1kenobi:
```suggestion
    name: Trigger Regen
```

PR #14556, opened 2026-08-22
feat: async command `validate()` and `run()`

cb1kenobi:
> [!IMPORTANT]
> This PR will require Titanium CLI 9.1.0, which could be considered a breaking change and thus, this PR MUST go in the SDK 14 release.

This PR does two major things:
1. Refactor command `validate()` and `run()` functions to be async/await instead of callback-based.
2. Move `Builder` class and some `titanium` namespace functions from `node-titanium-sdk` into the SDK:
  - `loadPlugins()`
  - `resolvePlatform()`
  - `validateModuleManifest()`
  - `validatePlatformOptions()`
  - and the SDK `manifest.json` loading

Making those function async/await unlocks the next wave of improvements which includes getting rid of the `async` dependency.

We are also looking to get rid of `node-titanium-sdk`, so starting to move some of the `titanium` namespace stuff to the SDK. There are still some stuff left to move over in the future.

PR #14568, opened 2026-09-03
refactor(cli): modernize the project command

cb1kenobi:
Modernizes the `project` command and moves two more helpers out of `node-titanium-sdk`, continuing the CLI migration from #14556.

## What changed

**`cli/commands/project.js`**
- `run()` is now `async run(logger, config, cli)`, matching `build` and `clean`; the `finished` callback is gone.
- The `switch (args.length)` block is split into `printProjectInfo()`, `getProjectKey()` and `setProjectKey()`.
- `var` chain at the top of `run()` replaced with block-scoped locals; `path` → `join`, `appc.fs.exists` → `existsSync`, `appc.string.rpad` → `padEnd`, `~indexOf` → `includes`.

**New `cli/lib/` helpers**
- `commonOptions()` and `validateProjectDir()` ported from `node-titanium-sdk`. `create`, `build` and `clean` still use the `ti.*` versions, so both exist for now.

## Behavior changes

- **`blackberry` is no longer a valid deployment target.** Validation now seeds from `sdkManifest.platforms` plus the `ipad` alias from `resolve-platform.js`, giving `android`, `iphone`, `ipad` — the same key set `ti create` writes. `ios` remains rejected: it is a build platform that maps to a device family, never a `deployment-targets` entry.
- **More than two arguments is now an error.** Previously `ti project a b c` matched no `switch` case and silently did nothing.

## Tests

New `cli/tests/test-project-run.js`, 21 tests, picked up by `npm run test:cli`:

- `report` / `json` / `text` output for the full project and for a single key
- `-o json` stays limited to `deployment-targets` plus the known property list, so `properties`, `modules` and platform sections do not leak into the output
- deployment target validation: iPhone-only, iPad-only and universal all reachable; `ios` and `blackberry` rejected
- the `ipad` → `iphone` alias actually copies the iphone template
- assertions run the written map through the same device family derivation as `iphone/cli/commands/_build.js`, so a regression in the target keys shows up as a wrong device family rather than just a wrong dict

One harness note worth knowing for future CLI tests: the suite installs the color getters on `String`, `Number` **and** `Boolean` prototypes, mirroring `assignColors()` in `titanium/src/cli.js`. Deployment target values parse as booleans, so a test that only loads the `colors` package (which patches `String.prototype` alone) will report a styling bug that does not exist at runtime.

Each fix was mutation-tested by reverting it and confirming the suite fails.

### DIFF
--- a/cli/commands/project.js
+++ b/cli/commands/project.js
@@ -5,10 +5,15 @@
  * See the LICENSE file for more information.
  */
 
-import path from 'node:path';
+import { join } from 'node:path';
 import ti from 'node-titanium-sdk';
 import appc from 'node-appc';
 import { loadPlugins } from '../lib/load-plugins.js';
+import { commonOptions } from '../lib/common-options.js';
+import { validateProjectDir } from '../lib/validate-project-dir.js';
+import { platformAliases, resolvePlatform } from '../lib/resolve-platform.js';
+import { existsSync } from 'node:fs';
+import { sdkManifest } from '../lib/sdk-manifest.js';
 
 export const cliVersion = '>=9.1.0';
 export const desc = 'get and set tiapp.xml settings';
@@ -40,7 +45,7 @@ export function config(logger, config) {
 				desc: 'the name of the project template to use',
 				default: 'default'
 			}
-		}, ti.commonOptions(logger, config)),
+		}, commonOptions(logger, config)),
 		args: [
 			{
 				name: 'key',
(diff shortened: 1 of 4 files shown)


m1ga (approved):
Looks good, output is the same as before 👍 
`tiapp.xml saving is currently not supported` 😄 didn't know that it could be use to change those properties at all. `project` is a parameter I never use as I just do `cat tiapp.xml` or open it

